Issue #11, opened 2016-03-03
KafkaMessageListenerContainer: MANUAL offsets concerns


artembilan:
Looks like `manualOffsets` are never cleared:
We add the `topic` and the Map for its offsets per partitions in the `updateManualOffset`:
```
if (!this.manualOffsets.containsKey(record.topic())) {
	this.manualOffsets.putIfAbsent(record.topic(), new ConcurrentHashMap<Integer, Long>());
}
this.manualOffsets.get(record.topic()).put(record.partition(), record.offset());
```
We call `commitAsync()` for them in the next `commitIfNecessary()` method. But we never remove those entries which have been commited already. 
The best way to fix is to add `iterator.remove();` into the `while` for the `iterator`.
Otherwise we commit them again on the next iteration.

PR #14, opened 2016-03-07
Fix MANUAL_IMMEDIATE

### LINKED ISSUE #12: KafkaMessageListenerContainer: MANUAL_IMMEDIATE isn't IMMEDIATE
We have this logs:
```
18:49:27.318 DEBUG [testMANUAL_IMMEDIATE-1-kafka-1][org.springframework.kafka.listener.KafkaMessageListenerContainer] Received: 2 records
18:49:27.318 DEBUG [testMANUAL_IMMEDIATE-0-kafka-1][org.springframework.kafka.listener.KafkaMessageListenerContainer] Received: 2 records
18:49:27.319 INFO  [testMANUAL_IMMEDIATE-1-kafka-1][org.springframework.kafka.listener.ConcurrentMessageListenerContainerTests] manual: ConsumerRecord(topic = testTopic5, partition = 1, offset = 0, key = 2, value = bar)
18:49:27.319 INFO  [testMANUAL_IMMEDIATE-0-kafka-1][org.springframework.kafka.listener.ConcurrentMessageListenerContainerTests] manual: ConsumerRecord(topic = testTopic5, partition = 0, offset = 0, key = 0, value = foo)
18:49:27.319 INFO  [testMANUAL_IMMEDIATE-1-kafka-1][org.springframework.kafka.listener.ConcurrentMessageListenerContainerTests] manual: ConsumerRecord(topic = testTopic5, partition = 1, offset = 1, key = 2, value = qux)
18:49:27.319 TRACE [testMANUAL_IMMEDIATE-1-kafka-1][org.springframework.kafka.listener.KafkaMessageListenerContainer] Polling...
18:49:27.320 TRACE [testMANUAL_IMMEDIATE-1-kafka-1][org.springframework.kafka.listener.KafkaMessageListenerContainer] Polling...
18:49:27.320 INFO  [testMANUAL_IMMEDIATE-0-kafka-1][org.springframework.kafka.listener.ConcurrentMessageListenerContainerTests] manual: ConsumerRecord(topic = testTopic5, partition = 0, offset = 1, key = 0, value = baz)
18:49:27.320 TRACE [testMANUAL_IMMEDIATE-0-kafka-1][org.springframework.kafka.listener.KafkaMessageListenerContainer] Polling...
18:49:27.320 TRACE [testMANUAL_IMMEDIATE-0-kafka-1][org.springframework.kafka.listener.KafkaMessageListenerContainer] Polling...
18:49:27.322 INFO  [main][org.springframework.kafka.listener.ConcurrentMessageListenerContainerTests] Stop MANUAL_IMMEDIATE
18:49:27.323 DEBUG [testMANUAL_IMMEDIATE-0-kafka-1][org.springframework.kafka.listener.KafkaMessageListenerContainer] Committing: {testTopic5-0=OffsetAndMetadata{offset=1, metadata=''}}
18:49:27.323 DEBUG [testMANUAL_IMMEDIATE-1-kafka-1][org.springframework.kafka.listener.KafkaMessageListenerContainer] Committing: {testTopic5-1=OffsetAndMetadata{offset=1, metadata=''}}
```

However with the `IMMEDIATE` I expect to see the commit exactly after each:
```
public void onMessage(ConsumerRecord<Integer, String> message, Acknowledgment ack) {
	logger.info("manual: " + message);
	ack.acknowledge();
	latch.countDown();
}
```

Moreover looks these offsets are placed to the regular auto offsets:
```
if (!ackMode.equals(AckMode.MANUAL)) {
	updatePendingOffsets(records);
}
```
From other they are in the `manualOffsets`, too:
```
else if (getAckMode().equals(AckMode.MANUAL_IMMEDIATE)) {
	updateManualOffset(record);
	consumer.wakeup();
}
```
Also I'm not sure in that last one `wakeup()`.
Its JavaDoc says:
```
/**
  * Wakeup the consumer. This method is thread-safe and is useful in particular to abort a long poll.
  * The thread which is blocking in an operation will throw {@link org.apache.kafka.common.errors.WakeupException}.
*/
```
Yes, we break the current `poll()`, but we commit offsets only after the `stop()`.

I'd say that we really should call `commitAsync()` on the consumer but only for that record directly in the `Acknowledgment` implementation without any `update...Offset`.

garyrussell:
Resolves #12

Execute the commit immediately, after asserting the callback is invoked on the consumer
thread.

Restructure consumer as an inner class.

__Given the restructuring, it might be easier to review this with `?w=1`__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -471,6 +471,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 						Entry<Integer, Long> offset = iterator.next();
 						commits.put(new TopicPartition(entry.getKey(), offset.getKey()),
 								new OffsetAndMetadata(offset.getValue() + 1));
+						iterator.remove();
 					}
 				}
 			}
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -54,7 +54,7 @@ import org.springframework.util.Assert;
  * @author Gary Russell
  *
  */
-public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListenerContainer<K, V> implements SchedulingAwareRunnable {
+public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListenerContainer<K, V> {
 
 	private final ConsumerFactory<K, V> consumerFactory;
 
@@ -64,15 +64,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 
 	private final TopicPartition[] partitions;
 
-	private final boolean autoCommit;
-
-	private final Map<String, Map<Integer, Long>> offsets = new HashMap<String, Map<Integer, Long>>();
-
-	private final ConcurrentMap<String, ConcurrentMap<Integer, Long>> manualOffsets = new ConcurrentHashMap<>();
-
-	private final CommitCallback callback = new CommitCallback();
-
-	private Consumer<K, V> consumer;
+	private ListenerConsumer listenerConsumer;
 
 	private ContainerOffsetResetStrategy resetStrategy = ContainerOffsetResetStrategy.NONE;
 
@@ -81,10 +73,6 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 	private MessageListener<K, V> listener;
 
 	private AcknowledgingMessageListener<K, V> acknowledgingMessageListener;
-
-	private volatile Collection<TopicPartition> definedPartitions;
-
-	private volatile Collection<TopicPartition> assignedPartitions;
 
 	/**
 	 * Construct an instance with the supplied configuration properties and specific
@@ -134,7 +122,6 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 		this.topicPattern = topicPattern;
 		this.partitions = topicPartitions == null ? null
 				: new LinkedHashSet<>(Arrays.asList(topicPartitions)).toArray(new TopicPartition[0]);
-		this.autoCommit = consumerFactory.isAutoCommit();
 	}
 
 	/**
@@ -166,11 +153,11 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 	 * either explicitly or by Kafka; may be null if not assigned yet.
 	 */
 	public Collection<TopicPartition> getAssignedPartitions() {
-		if (this.definedPartitions != null) {
-			return Collections.unmodifiableCollection(this.definedPartitions);
+		if (this.listenerConsumer.definedPartitions != null) {
+			return Collections.unmodifiableCollection(this.listenerConsumer.definedPartitions);
 		}
-		else if (this.assignedPartitions != null) {
-			return Collections.unmodifiableCollection(this.assignedPartitions);
+		else if (this.listenerConsumer.assignedPartitions != null) {
+			return Collections.unmodifiableCollection(this.listenerConsumer.assignedPartitions);
 		}
 		else {
 			return null;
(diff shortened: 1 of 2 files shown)


garyrussell:
Rebased; aded fix for #11

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
```diff
@@ -242,6 +242,8 @@ public void onMessage(ConsumerRecord<Integer, String> message) {
 	public void testManualCommit() throws Exception {
 		testManualCommitGuts(AckMode.MANUAL, topic4);
 		testManualCommitGuts(AckMode.MANUAL_IMMEDIATE, topic5);
+		testManualCommitGuts(AckMode.MANUAL, topic4);
+		testManualCommitGuts(AckMode.MANUAL_IMMEDIATE, topic5);
```
artembilan: What is the purpose of this repetition?
Looks like copy/paste, but for what?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -199,256 +183,314 @@ else if (messageListener instanceof MessageListener) {
...
+
+		private final ConcurrentMap<String, ConcurrentMap<Integer, Long>> manualOffsets = new ConcurrentHashMap<>();
+
+		private final Map<String, Map<Integer, Long>> offsets = new HashMap<String, Map<Integer, Long>>();
+
+		private final MessageListener<K, V> listener;
+
+		private final AcknowledgingMessageListener<K, V> acknowledgingMessageListener;
+
+		private final ContainerOffsetResetStrategy resetStrategy;
```
artembilan: Looks like this `ListenerConsumer` isn't `static`, so we can use properties from the outer class.
Why do you copy them here?

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
garyrussell: I wanted to be sure the commits worked ok so I ran the tests again and the second tests start at the committed offset.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -199,256 +183,314 @@ else if (messageListener instanceof MessageListener) {
...
+											new OffsetAndMetadata(record.offset() + 1));
+									if (logger.isDebugEnabled()) {
+										logger.debug("Committing: " + commits);
+									}
+									consumer.commitAsync(commits, callback);
+								}
+								else {
+									throw new IllegalStateException(
+											"With MANUAL_IMMEDIATE ack mode, acknowledget must be invoked on the "
+											+ "consumer thread");
```
artembilan: Would you mind to elaborate more. Why is it so important?
All other cases uses the same  `commitAsync()`.
With this your change I see no difference between `MANUAL_IMMEDIATE` and `RECORD`.

Thank you!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Because this class has setters.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
artembilan: Good. Let's add the comment on the matter for these lines!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: The consumer is not thread safe - the only thread-safe operation is `wakeup`; hence the commits have to happen on the consumer thread.

The difference between `RECORD` and `MANUAL_IMMEDIATE` is that, with the latter, the user gets to choose when the commit occurs.

For example `Foo, Foo, Foo, Bar, Foo, Bar` - the user might want to commit only when he gets a `Bar`.

artembilan: Ah! I got it now. The ` commitAsync()` should be invoked at the same Thread independently of our mode.
Therefore we can do only `poll()` or `commit()` at a time.

garyrussell: Right, that's why my previous (flawed) solution used wake.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: OK. Merging with the comment into the test-case and some other simple code polishing.
Those places where my IDEA doesn't like it. :smile:

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/df4a3a621a6a43a0f0d437c9d0ec7f46d5da0946

PR #15, opened 2016-03-07
GH-10: Add Sync Template Methods

garyrussell:
Resolves: #10


spring-kafka/src/test/java/org/springframework/kafka/utils/KafkaTestUtils.java
```diff
@@ -0,0 +1,199 @@
...
+
+import org.springframework.kafka.listener.ConcurrentMessageListenerContainer;
+import org.springframework.kafka.listener.KafkaMessageListenerContainer;
+import org.springframework.kafka.test.rule.KafkaEmbedded;
+
+/**
+ * @author Gary Russell
+ *
+ */
+public final class KafkaTestUtils {
```
artembilan: Why don't move this class to the `test` project as well?
Looks like enough boilerplate and can be customised for those `...Props`

garyrussell: Because it has dependencies on the listener container. I suppose we could split it into two classes.

artembilan: I'm fine with merge, but let me know if you are going to address this yourself or leave it to me on merge.

garyrussell: I think the matchers will be useful for users so I'll move them to `-test`

src/reference/asciidoc/testing.adoc
```diff
@@ -0,0 +1,131 @@
...
+		assertThat(received, hasPartition(0));
+		assertThat(received, hasValue("bar"));
+		template.syncConvertAndSend(TEMPLATE_TOPIC, 0, 2, "baz");
+		received = records.poll(10, TimeUnit.SECONDS);
+		assertThat(received, hasKey(2));
+		assertThat(received, hasPartition(0));
+		assertThat(received, hasValue("baz"));
+	}
+
+}
```
artembilan: Will fix `tabs -> whiltespaces` during merge together with all other doc vulnerabilities, like:
```
Content of the region-body on page 5 overflows the available area in block-progression dimension. (fo:page-sequence, location: 23/304)
```

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/47f1de132c25489e48016f73e9f4d282fe8fe164

Issue #8, opened 2016-03-02
Create a Shared Producer in the DefaultKafkaProducerFactory


artembilan:
See the comment on the matter: https://github.com/spring-projects/spring-integration-kafka/pull/106#discussion_r54483301

artembilan:
As a variant we can consider to implement `ProducerFactoryBean` for singleton and appropriate `Lifecycle` control.

PR #16, opened 2016-03-07
GH-8: Shared KafkaProducer

garyrussell:
Resovles #8

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
@@ -16,26 +16,128 @@
 package org.springframework.kafka.core;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -16,26 +16,128 @@
...
 	public DefaultKafkaProducerFactory(Map<String, Object> configs) {
 		this.configs = new HashMap<>(configs);
 	}
 
 	@Override
+	public void destroy() throws Exception {
+		Producer<K, V> producer = this.producer;
+		this.producer = null;
+		if (producer != null) {
+			producer.close();
```
artembilan: Is it expected?Looking into the `this.producer = new CloseSafeProducer<K, V>()`, and its `no-op` `close()` impl, I'd say that this is a dead code and we don't need this `start()`, `stop()` and `destroy()` "hell".

Or have you meant something like `producer.delegate.close();` ?

garyrussell: good catch!

garyrussell:
Pushed.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/c8b723fa072d0cc10cee46d825b3bc206dd00c64

PR #18, opened 2016-03-07
Polishing

### LINKED ISSUE #6: Restrict @TopicPartition just only for the annotation attributes.
Right now we can define `@TopicPartition` everywhere: on class and on method level, or even parameter.

Mark the annotation with:
```
@Target({})
@Retention(RUNTIME)
```
to restrict its usage.

See `@Poller` in the Spring Integration.

### LINKED ISSUE #7: Typo in the `AbstractKafkaListenerContainerFactory.initializeContainer` JavaDoc
The symptom is:
```
* @param instance the containe instance to configure.
```

There are other places in this class to polish, like redundant `counter` and `logger` properties.

Consider to polish other copy/paste artifacts like `@since 1.4` or similar.

garyrussell:
Resolves #6
Resolves #7


artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/e0914c5e98804a2c89e2b2296a4d95b9f415342e

Issue #3, opened 2016-03-02
ConcurrentMessageListenerContainerTests.testDefinedPartitions() fails on Windows


artembilan:
Adding `container.setResetStrategy(ContainerOffsetResetStrategy.EARLIEST);` makes test happy.
And I don't need a `Thread.sleep()` any more!

Any clues?

artembilan:
After some deep debuging I found that we have `auto.offset.reset=latest` for the `ConsumerConfig`:
```
.define(AUTO_OFFSET_RESET_CONFIG,
                                        Type.STRING,
                                        "latest",
                                        in("latest", "earliest", "none"),
                                        Importance.MEDIUM,
                                        AUTO_OFFSET_RESET_DOC)
```
So, our `ContainerOffsetResetStrategy.NONE` does not have effect.

For my slow SATA HDD the `Thread.sleep(1000);` isn't enough after `container1.start();` to be able to setup to Kafka Broker and pick up all required coordinators, balancers and, finally, the proper `0` offset for the partition.

From one hand we really can fix the test to wait for the offset assignments:
```
12:34:16.015 DEBUG [b2-0-kafka-1][org.apache.kafka.clients.NetworkClient] Completed connection to node 2147483647
12:34:16.015 DEBUG [b2-0-kafka-1][org.apache.kafka.clients.consumer.internals.ConsumerCoordinator] No committed offset for partition testTopic3-1
12:34:16.016 DEBUG [b2-0-kafka-1][org.apache.kafka.clients.consumer.internals.Fetcher] Resetting offset for partition testTopic3-1 to latest offset.
12:34:16.017 DEBUG [b2-0-kafka-1][org.apache.kafka.clients.consumer.internals.Fetcher] Fetched offset 0 for partition testTopic3-1
12:34:16.017 TRACE [b2-0-kafka-1][org.apache.kafka.clients.consumer.internals.Fetcher] Added fetch request for partition testTopic3-1 at offset 0
```
But from other side we may should be consistent with our `ContainerOffsetResetStrategy` and the existing `OffsetResetStrategy`.

garyrussell:
`NONE` simply means  the container takes no action but I see what you mean.

Since the consumer properties has this available, I think we should retain only `RECENT` (and remove the enum altogether).

Agree?

PR #20, opened 2016-03-07
GH-3:  Fix race condition in the partitions test 

artembilan:
Fixes GH-3 (https://github.com/spring-projects/spring-kafka/issues/3)


garyrussell:
I don't mind merging this but we really need to understand the problem so we should leave #3 open; this doesn't "fix" it.

artembilan:
Repushed with more robust fix.
See the commit comment.

artembilan:
The question about the `OffsetResetStrategy` with the standard automatic Consumer's `updateFetchPositions()` function instead of `ContainerOffsetResetStrategy` is still open.

Looks like we can achieve `RECENT` just with the `recentOffset > 0` and treat `auto.offset.reset` as a `latest` for this case.

PR #21, opened 2016-03-08
GH-13: Add checkstype and fix as much as possible

artembilan:
Fixes GH-13 (https://github.com/spring-projects/spring-kafka/issues/13)

Also upgrade to Gradle 2.11 to support the latest `checkstyle`


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -203,8 +205,6 @@ protected void doStop() {
 
 	private class ListenerConsumer implements SchedulingAwareRunnable {
 
-		private final Log logger = LogFactory.getLog(this.getClass());
-
 		private final CommitCallback callback = new CommitCallback();
```
garyrussell: I added a local logger explicitly to avoid referencing the enclosing class on each call.

artembilan: Yeah... I see.
What I don't like that I really fine with the `logger` without `this.`, but there is no `exclude` option for the `RequireThis` rule...

So, let's move it back and add `this.` because there is no choice! At least I did similar for the `KafkaTemplate` :smile:

garyrussell: ok; fixing during merge - also fixed some whitespace issues in the XML

artembilan: XML? Which XML?
I checked locally before pushing...

src/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,169 @@
...
+ 		<module name="StringLiteralEquality" />
+ 		<!--<module name="NestedForDepth">-->
+ 			<!--<property name="max" value="3" />-->
+ 		<!--</module>-->
+ 		<!--<module name="NestedIfDepth">-->
+ 			<!--<property name="max" value="3" />-->
+ 		<!--</module>-->
+ 		<!--<module name="NestedTryDepth">-->
+ 			<!--<property name="max" value="3" />-->
+ 		<!--</module>-->
```
artembilan: I'd remove these rules from the config altogether. We really have more deep `if...else` and they look good for me.
And we may have more complex structures.
WDYT?

garyrussell:
Rebased, merged as c7ad532fb2db7f431f061f685f82863f1d196eab

Thanks for doing the heavy lifting on this one!!

src/checkstyle/checkstyle.xml
garyrussell: Sorry - didn't see this before merging.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: `git diff --check`

There were many lines with `<sp>\t\t`.

Leading space before tab.

artembilan:
Yeah... That's not so hellish work...
Just take a look to the Checkstyle source code: https://github.com/checkstyle/checkstyle.
Gods are living there!

src/checkstyle/checkstyle.xml
garyrussell: I uncommented (in my other PR) and set the if depth to 4 - let's see how we like it going forward.

artembilan: OK. That's fine with me, too.

PR #26, opened 2016-03-09
Provide more Checkstyle rules and their fixes

artembilan:


artembilan:
Pay attention we don't use `AssertJ`, but we can, of course. Not so much to fix :smile: 

I can't make my IDEA to indent only with tabs, that's why that rule is commented, too.

garyrussell:
SInce it's a new project, we can consider switching to assertJ - not too many tests yet.

I recall that Phil tweeted about how much work it was to switch over.

spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
```diff
@@ -160,7 +160,7 @@ private void waitListening(String id) throws InterruptedException {
 		public Map<String, Object> consumerConfigs() {
 			Map<String, Object> props = new HashMap<>();
 			props.put("bootstrap.servers", embeddedKafka.getBrokersAsString());
-//			props.put("bootstrap.servers", "localhost:9092");
+			//props.put("bootstrap.servers", "localhost:9092");
```
garyrussell: I don't think I like this rule

garyrussell:
Polishing here: https://github.com/garyrussell/spring-kafka/commit/c0d0968d87155a9f0bbfa43c47e09778a06ff202

I think there's something wrong with your IDEA for this project - in several places you added a lot of leading spaces.

garyrussell:
Oh; I see; that's what you are talking about - there must be a way if boot has that rule; Stephane uses IDEA.

garyrussell:
More polish https://github.com/garyrussell/spring-kafka/commit/ba8a06fc57c6bf0063a865885f97a808c80802a0

Doesn't this help https://www.jetbrains.com/idea/help/changing-indentation.html ??

spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
artembilan: Yeah... Me too.
I guess you fixed that with the `BLOCK_COMMENT_BEGIN`.
Thanks.

artembilan:
Thank you for polishing.
I'm sure we can merge it.

And thank you for the article.
I was it before and it is just for basic usage, nothing specific about the count of tab.
My case is something like 6-7 tabs and then whitespaces.

Really need to go to Stephane for the help on the matter..

PR #46, opened 2016-03-30
GH-39-44: Template Changes

### LINKED ISSUE #39: Add KafkaTemplate Methods
There are not send methods for a partition with no key.

### LINKED ISSUE #44: Change Template convertAndSend... to just send...
The template's not doing any conversion; it's just a lightweight wrapper around the Kafka producer; the conversion is done by the Kafka Serializer.

garyrussell:
Resolves #39
Resolves #44

Remove `convert` from method names.
Add missing methods - no key.

Also remove log4j.properties from test jar.


artembilan:
LGTM, but the Reference Manual should be addressed as well.

garyrussell:
Oops - yes, of course :smile:

garyrussell:
Pushed - I'd appreciate this being merged ASAP - I have an s-i-k PR in the wings, waiting for this :smile: Thanks.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
```diff
@@ -73,7 +73,16 @@
 	 * @param data The data.
 	 * @return a Future for the {@link RecordMetadata}.
 	 */
-	Future<RecordMetadata> convertAndSend(String topic, K key, V data);
+	Future<RecordMetadata> send(String topic, K key, V data);
+
+	/**
+	 * Send the data to the provided topic with the provided partition and no key.
+	 * @param topic the topic.
+	 * @param partition the key.
```
artembilan: Typo: the s/key/partition 
:smile: 

Will fix on merge.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
```diff
@@ -143,7 +152,19 @@
...
 	 * @throws InterruptedException thread interrupted while awaiting result.
 	 */
-	RecordMetadata syncConvertAndSend(String topic, K key, V data) throws InterruptedException, ExecutionException;
+	RecordMetadata syncSend(String topic, K key, V data) throws InterruptedException, ExecutionException;
+
+	/**
+	 * Send the data to the provided topic with the provided partition and no key;
+	 * wait for result.
+	 * @param topic the topic.
+	 * @param partition the key.
```
artembilan: DITO

garyrussell: Good; thanks for finding my sloppiness.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -78,88 +79,102 @@ public void setDefaultTopic(String defaultTopic) {
...
 		flush();
 		return future.get();
 	}
 
 	@Override
-	public RecordMetadata syncConvertAndSend(String topic, int partition, K key, V data)
+	public RecordMetadata syncSend(String topic, int partition, V data)
+			throws InterruptedException, ExecutionException {
+		Future<RecordMetadata> future = send(topic, partition, data);
+		return future.get();
```
artembilan: Looking to all these `syncSend()`  implementation I'd say that we don't need them at all.
That is an end-user wish to block it on the `future.get()`.
The `Mono` discussion led me think this way...
WDYT?

artembilan: M-m-m. This method doesn't have `flush();` as all other `syncSend()`

garyrussell: Yeah, I was thinking it would be a convenience to save the user from 2 statements.

But I suppose he could use

template.send(...).get()

I'll add another commit.

garyrussell: Well, actually, no; the sync calls also `flush()` the producer (there's a bug in this one).

So without these, the user would have to 

    send()
    flush()
    get()

garyrussell: WDYT?

artembilan: OK. That's fine.
I think with the `Mono` we could do the `flush()` on demand, not directly like it is with `Future`.
However I'm not sure that it is a point to switch to Java 8 already...

Let me know if you are still going to add some commit!

garyrussell:
Can you just add the `flush()` and fix the two typos?

Thanks

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/dc19f0cd89d114b747c451280c530e1af474a563

PR #47, opened 2016-03-30
GH-45: Add RECEIVED... Headers

### LINKED ISSUE #45: Consider Adding KafkaHeaders.RECEIVED_* Headers
Separate inbound headers from outbound to prevent propagation by default - allows the elimination of the `enableHeaderRouting` property in the s-i-k `KPMH`.

garyrussell:
Resolves #45

Avoid propagation of received headers when sending `o.s.messaging.Message`s.

Also add `Message<?>` methods to the template.


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -211,9 +233,21 @@ public RecordMetadata syncSend(String topic, int partition, K key, V data)
...
-	public void flush() {
-		this.producer.flush();
+	private ProducerRecord<K, V> messageToProducerRecord(Message<?> message) {
+		MessageHeaders headers = message.getHeaders();
+		Object topic = headers.get(KafkaHeaders.TOPIC);
+		Object partition = headers.get(KafkaHeaders.PARTITION_ID);
+		Object key = headers.get(KafkaHeaders.MESSAGE_KEY);
+		Object payload = message.getPayload();
+		Assert.isTrue(topic == null || topic instanceof String, "topic header must be a String");
+		Assert.isTrue(partition == null || partition instanceof Integer, "partition header must be an Integer");
```
artembilan: Can't we use an overloaded version of `MessageHeaders.get()` with the `Class` to cast instead?

artembilan:
Here is a polishing how I assume them: https://github.com/artembilan/spring-kafka/commit/213cfcc3d9afaa3718e4102050af3409b4e3face

garyrussell:
Doh; thanks; pushed - also fixed those `send` methods.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/753d9bdb89ca5dcf22ee0a18b8da10303ed13c73

PR #50, opened 2016-03-31
GH-42: Ignore Bridged Methods

### LINKED ISSUE #42: Ignore Bridge Methods
See https://github.com/spring-projects/spring-amqp/pull/399

garyrussell:
Resolves #42

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -20,12 +20,17 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.Pattern;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 import org.springframework.aop.framework.Advised;
 import org.springframework.aop.support.AopUtils;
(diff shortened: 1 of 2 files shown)


garyrussell:
First commit solves it the same way as Spring AMQP; second commit converts to `MethodIntrospector`.

artembilan:
Here you are the polishing: https://github.com/artembilan/spring-kafka/commit/1bb48050365b30916b32908ebafbd2590ba8da0e

I was confused a bit at a glance why we need to return `null` for `false` matching.
Than I found another method which returns just `Set<Method>` and accepts a simple `boolean` filter.

Also, please, find there some polishing for tests. Especially the TX case.

garyrussell:
Polish looks good, although it's not clear why you added getters for the 2 latches - it's only a test so I don't think accessing the field is a big deal - but, if you find it offensive, why just those 2?

Also, I'm not sure the TX is necessary, `@Config` classes are already CGLIB - the original problem (dup id) was reproduced without the TX.

But, it doesn't hurt to leave it there.

artembilan:
I also don't mind about field access from test-cases.
Wee need those two because of `proxyTargetClass = true`.
So, `@Autowired public IfaceListenerImpl ifaceListener;` is a CGLIB proxy with those fields as `null`.

Your thoughts?

garyrussell:
OK; thanks - that explains - feel free to merge.

artembilan:
And I'm not sure why do you worry about config classes, because the issue on the target services which are CGLIB proxies, like this my TX tricks.

Thank you for the discussion. 
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/4cf24fbf28a0b8da390dba20c37a361df81cb0b4

garyrussell:
Yes; not really CGLIB, but it seems the bridge method problem exists even without TX.

artembilan:
Right we have it too, when the interface is generic and class is with the target type: http://docs.spring.io/spring-amqp/docs/1.6.0.BUILD-SNAPSHOT/reference/html/_reference.html#_proxy_rabbitlistener_and_generics

Issue #49, opened 2016-03-31
Add MessageConverter Support


garyrussell:
When using `o.s.messaging.Message`.

PR #52, opened 2016-03-31
GH-49: Add MessageConversion and JSON

garyrussell:
Resolves: #49

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,6 +73,7 @@ subprojects { subproject ->
 		avroVersion = '1.7.6'
 		gsCollectionsVersion = '5.0.0'
 		hamcrestVersion = '1.3'
+		jacksonVersion = '2.3.2'
 		junitVersion = '4.12'
 		kafkaVersion = '0.9.0.1'
 		log4jVersion = '1.2.17'
@@ -143,12 +144,9 @@ project ('spring-kafka') {
 
 	dependencies {
 		compile "org.springframework:spring-messaging:$springVersion"
-//		compile ("org.apache.avro:avro:$avroVersion", optional)
-//		compile ("org.apache.avro:avro-compiler:$avroVersion", optional)
-//		compile "com.goldmansachs:gs-collections:$gsCollectionsVersion"
-//		compile "io.projectreactor:reactor-core:$reactor2Version"
-
 		compile "org.apache.kafka:kafka-clients:$kafkaVersion"
+		compile ("com.fasterxml.jackson.core:jackson-core:$jacksonVersion", optional)
+		compile ("com.fasterxml.jackson.core:jackson-databind:$jacksonVersion", optional)
 
 		testCompile project (":spring-kafka-test")
 		testCompile "org.assertj:assertj-core:$assertjVersion"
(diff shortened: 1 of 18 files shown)


build.gradle
```diff
@@ -143,12 +144,9 @@ project ('spring-kafka') {
 
 	dependencies {
 		compile "org.springframework:spring-messaging:$springVersion"
-//		compile ("org.apache.avro:avro:$avroVersion", optional)
-//		compile ("org.apache.avro:avro-compiler:$avroVersion", optional)
-//		compile "com.goldmansachs:gs-collections:$gsCollectionsVersion"
-//		compile "io.projectreactor:reactor-core:$reactor2Version"
```
artembilan: If we remove the deps, we probably should remove their versions as well.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessageConverter.java
```diff
@@ -16,21 +16,36 @@
...
  *
  * @author Gary Russell
  */
-public interface MessageConverter<K, V> {
+public interface MessageConverter {
+
+	/**
+	 * Convert a {@link ConsumerRecord} to a {@link Message}.
+	 * @param record the record.
+	 * @param acknowledgment the acknowledment.
```
artembilan: Typo: acknowled**g**ment

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
```diff
@@ -37,7 +39,7 @@
  * @author Marius Bogoevici
  * @author Gary Russell
  */
-public class MessagingMessageConverter<K, V> implements MessageConverter<K, V> {
+public class MessagingMessageConverter<K, V> implements MessageConverter {
```
artembilan: My IDEA says me that the generics are redundant.
Investigating why you need them at all...

spring-kafka/src/main/java/org/springframework/kafka/support/converter/StringJsonMessageConverter.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * JSON Message converter - String on output, String or byte[] on input.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Gary Russell
+ *
+ */
+public class StringJsonMessageConverter<K, V> extends MessagingMessageConverter<K, V> {
```
artembilan: Well, looks like we are fine even without all these generics.
Am I missing anything ?

garyrussell: As I mentioned last night I was struggling with Generics yesterday; when I propagated `MessageConverter` through a bunch of APIs I was forced to add generics to lots of classes so I ended up removing them; I am not surprised there are some leftovers.

I'll take a look.

garyrussell:
Pushed

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/a087ce21b3e28eb3b199e0b01e4e8f246a62b4ae after rebase.

PR #57, opened 2016-04-04
GH-56: Change Template to return ListenableFuture

### LINKED ISSUE #56: KafkaTemplate - Return ListenableFuture; Remove syncSend methods

garyrussell:
Resolves #56

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -16,11 +16,11 @@
 
 package org.springframework.kafka.core;
 
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.kafka.clients.producer.Callback;
 import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.RecordMetadata;
@@ -28,9 +28,11 @@ import org.apache.kafka.clients.producer.RecordMetadata;
 import org.springframework.kafka.support.KafkaHeaders;
 import org.springframework.kafka.support.LoggingProducerListener;
 import org.springframework.kafka.support.ProducerListener;
-import org.springframework.kafka.support.ProducerListenerInvokingCallback;
+import org.springframework.kafka.support.SendResult;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHeaders;
+import org.springframework.util.concurrent.ListenableFuture;
+import org.springframework.util.concurrent.SettableListenableFuture;
 
 
 /**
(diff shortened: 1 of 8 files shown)


sdeleuze:
👍

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -91,112 +108,51 @@ public void setProducerListener(ProducerListener<K, V> producerListener) {
...
-	public Future<RecordMetadata> send(String topic, int partition, K key, V data) {
+	public ListenableFuture<SendResult<K, V>> send(String topic, int partition, K key, V data) {
 		ProducerRecord<K, V> producerRecord = new ProducerRecord<>(topic, partition, key, data);
 		return doSend(producerRecord);
 	}
 
 	@Override
-	public Future<RecordMetadata> send(Message<?> message) {
+	public ListenableFuture<SendResult<K, V>> send(Message<?> message) {
 		ProducerRecord<K, V> producerRecord = messageToProducerRecord(message);
```
artembilan: I think when we come to the Reactive solution this conversion should be done as a part of a stream - `.map(this::messageToProducerRecord)`.
And even that `new ProducerRecord<>()` above can be there as well.

WDYT, @sdeleuze ?

artembilan:
LGTM
And if there is no any concerns from your side, I can merge.

garyrussell:
:+1: Then, tomorrow, I'll update s-i-k; then we should release M2 (and M1 for s-i-k).

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/fac9ec6
after the "hard" conflicts fixing on rebase.

Bear in mind: Bamboo is still dead, so there is nothing to utilize from the SIK yet.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
sdeleuze: Indeed @artembilan, I also think this is a perfect use case for `.map(this::messageToProducerRecord)`.

Please notice that for more heavy transformations that could involve some latency, something like `.flatMap(this::fooMethod)` with `fooMethod` returning `Mono` or `Flux` would be a better choice in order to avoid to block too much the processing. That's why we have created [Reactive codec in Spring Reactive](https://github.com/spring-projects/spring-reactive/tree/master/src/main/java/org/springframework/core/codec) that are not specific to a protocol (they could be used for HTTP, messaging, Websocket, file, etc.). Most serialization libraries are still blocking so we stick to process small objects thanks to [JsonObjectDecoder](https://github.com/spring-projects/spring-reactive/blob/master/src/main/java/org/springframework/core/codec/support/JsonObjectDecoder.java) for example, but some are already asynchronous like Aalto XML, see [@poutsma PR about that](https://github.com/spring-projects/spring-reactive/pull/77).

artembilan: Yeah, thank you @sdeleuze !

Will try to study what you, guys, do there some days...

PR #65, opened 2016-05-07
GH 61: User defined serializer and deserializer injection

kvmkreddy:
Added feature to inject user defined Serializer and Deserializer implementations of key, value entities for both KafkaConsumer and KafkaProducer API.


spring-kafka/src/main/java/org/springframework/kafka/core/CompositeKafkaConsumerProducerStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+ * <p>
+ * This strategy will also produce {@link KafkaProducer} for provided {@link Map} {@code configs}, {@link Deserializer} {@code keyDeserializer}
+ * and {@link Deserializer} {@code valueDeserializer} on {@link #createKafkaProducer()} invocation.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Murali Reddy
+ */
+public class CompositeKafkaConsumerProducerStrategy<K, V> implements KafkaConsumerProducerStrategy<K, V> {
```
artembilan: Well, @kvmkreddy ,

Thank you for the contribution one more time, first of all!

I am a bit nervous with this mix of concerns. 
My application can use only producer or consumer. But with this I have to configure something for both of them.
That really would be more clear just to have setters on the `DefaultKafkaConsumerFactory`/`DefaultKafkaProducerFactory`.

Even if you are going to have the common `configs` map, that is your architectural choice for your application.
But I'd prefer to distribute responsibility and don't worry about other independent components.

Your purpose of such a solution?

src/checkstyle/checkstyle.xml
```diff
@@ -178,4 +178,4 @@
 		<module name="WhitespaceAround" />
 
 	</module>
-</module>
+</module>
```
artembilan: Not sure what led to make some changes in this file, but we definitely require a new line in the end of each file.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -100,13 +108,17 @@ public boolean isRunning() {
...
-					this.producer = new CloseSafeProducer<K, V>(new KafkaProducer<K, V>(this.configs));
+					this.producer = new CloseSafeProducer<K, V>(createKafkaProducer());
 				}
 			}
 		}
 		return this.producer;
 	}
 
+	protected KafkaProducer<K, V> createKafkaProducer() {
+		return this.strategy.createKafkaProducer();
```
artembilan: H-m. Look, you make `this.configs` useless.

spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
```diff
@@ -247,7 +260,8 @@ public MultiListenerBean multiListener() {
 
 		@Bean
 		public ProducerFactory<Integer, String> producerFactory() {
-			return new DefaultKafkaProducerFactory<>(producerConfigs());
+			Map<String, Object> configs = producerConfigs();
+			return new DefaultKafkaProducerFactory<>(configs, producerStrategy(configs));
```
artembilan: This configuration smells to make me duplicate `configs`.
Although finally the first one becomes useless (see my comment on `DefaultKafkaProducerFactory`).

So, I insist that we should just end up with a couple of setters on the `Factories` and don't introduce any "buggy" strategies.

spring-kafka/src/main/java/org/springframework/kafka/core/CompositeKafkaConsumerProducerStrategy.java
kvmkreddy: Let me re-implement the strategy contracts for both Consumer and Producer. In this way we can have clear separation and individual use-cases and/or clients can have their own implementations.

src/checkstyle/checkstyle.xml
kvmkreddy: This is my oversight. Will revert back to original.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
kvmkreddy: Basically we don't need `this.configs` once we have strategy implementation. Will clean it up and adjust constructors.

spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
kvmkreddy: May be you can take a look once again.  Even after this separation if you believe that this implementation is unnecessary, then I can think about simple Serializer/Deserializer injection directly into Factory classes.

artembilan:
Sorry, I still don't see reason and, therefore, don't understand whole the new `strategy` level of abstraction. The `ConsumerFactory` and `ProducerFactory` have been introduced exactly as strategies between target `container` object and user-specific requirements. 
With all those new `strategies` you eliminate the existing `factory` purpose and fully delegates its functionality to a new abstraction. 
My opinion is still just stay with the `Factory` abstraction and provide a couple setters for `(De)Serializer`s.

To be honest I would introduce anything, because you always ca implement your own `Consumer-` `ProducerFactory` and inject it into the target component :smile: 

Let me know if I am missing anything

garyrussell:
I tend to agree, why not just inject the (de)serializers into the default factories?

kvmkreddy:
The only reason why I tend to go for strategy instead of injecting (De)Serializer directly into `DefaultKafkaConsumerFactory` or `DefaultKafkaProducerFactory` is, 
 these factories are wrapping or have additional implementation on `KafkaConsumer` or `KafkaProducer`  instances respectively.

IMHO In this strategy way user's can inject their own implementation of  `KafkaConsumer` or `KafkaProducer` creation instead of implementing complete `DefaultKafkaConsumerFactory` or `DefaultKafkaProducerFactory`. In this model user can enjoy/benefit the in-built features of both factories.

artembilan:
OK, @kvmkreddy ,

I appreciate your wish to contribute anything to the project, but let's look at this from the perspective of the reason in face what we have already so far and what you would like to introduce.
(Thanks to the short code we can demonstrate it here in comments :smile: )
````java
public interface ConsumerFactory<K, V> {

	Consumer<K, V> createConsumer();

	boolean isAutoCommit();

}
````
````java
public interface KafkaConsumerStrategy<K, V> {

	KafkaConsumer<K, V> createKafkaConsumer();

}
````
So, semantically these two abstraction are fully similar. 
I really can reach your API variant with existing one.

````java
public class DefaultKafkaConsumerFactory<K, V> implements ConsumerFactory<K, V> {

	private final Map<String, Object> configs;

	public DefaultKafkaConsumerFactory(Map<String, Object> configs) {
	}

	@Override
	public Consumer<K, V> createConsumer() {
		return new KafkaConsumer<>(this.configs);
	}

	// Omit isAutoCommit() as non-relevant for discussion
}
````

````java
public class DefaultKafkaConsumerStrategy<K, V> implements KafkaConsumerStrategy<K, V> {

	private final Map<String, Object> configs;

	public DefaultKafkaConsumerStrategy(Map<String, Object> configs) {
	}

	@Override
	public KafkaConsumer<K, V> createKafkaConsumer() {
		return new KafkaConsumer<K, V>(this.configs);
	}

}
````
Yeah... Your implementation just duplicates an existing one, too.

I understand your argument
> user's can inject their own implementation of KafkaConsumer

But that really looks like you always can extend `DefaultKafkaConsumerFactory` and override its `createConsumer()` for your target purpose.

The same is fully orthogonal for the `Producer` part.

So, we still insist that the Framework don't need this extra strategy abstraction and those `(De)Serializer` options should be present on the `Default...Factory`. Enough straightforward and intuitive to have.
Any other custom cases can be implemented with the direct `ConsumerFactory` and `ProducerFactory` implementations. Or... as extensions of `Default` ones.

**Comment by artembilan** [comment]
??? You don't need to close the PR.
There is just enough to rework it according the initial requirements: 
provide the hooks to inject `(De)Serializer`s. For this purpose the `Default` factories are good candidates.
There is just no need to provide one more level of abstraction.

Sorry, if my comment mislead you...

File: spring-kafka/src/main/java/org/springframework/kafka/config/SimpleKafkaListenerContainerFactory.java [hunk 5]
```diff
@@ -43,6 +45,8 @@
 
 	private Long recentOffset;
 
+	private ConsumerRebalanceListener consumerRebalanceListener;
```
artembilan: You have to rebase your branch to the latest `master`.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
```diff
@@ -22,28 +23,53 @@
...
 	public DefaultKafkaConsumerFactory(Map<String, Object> configs) {
 		this.configs = new HashMap<>(configs);
 	}
 
+	public DefaultKafkaConsumerFactory(Map<String, Object> configs,
+			Deserializer<K> keyDeserializer,
+			Deserializer<V> valueDeserializer) {
+		this.configs = new HashMap<>(configs);
+		this.keyDeserializer = keyDeserializer;
+		this.valueDeserializer = valueDeserializer;
```
artembilan: No, they aren't required options, therefore has to be present as a setters.
And I even maybe interested only in `key` or `value` Deserializer but not both.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
```diff
@@ -22,28 +23,53 @@
...
+		return createKafkaConsumer();
+	}
+
+	protected KafkaConsumer<K, V> createKafkaConsumer() {
+		if (this.keyDeserializer != null &&
+				this.valueDeserializer != null) {
+			return new KafkaConsumer<K, V>(this.configs, this.keyDeserializer, this.valueDeserializer);
+		}
+		else {
+			return new KafkaConsumer<K, V>(this.configs);
```
artembilan: No need in this `if...else`. The `KafkaConsumer` takes care about those options properly:
```
if (keyDeserializer != null)
        newConfigs.put(KEY_DESERIALIZER_CLASS_CONFIG, keyDeserializer.getClass());
if (valueDeserializer != null)
        newConfigs.put(VALUE_DESERIALIZER_CLASS_CONFIG, valueDeserializer.getClass());
```
So, we can boldly rely only on one direct ctor invocation.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -100,13 +119,23 @@ public boolean isRunning() {
...
 		return this.producer;
 	}
 
+	protected KafkaProducer<K, V> createKafkaProducer() {
+		if (this.keySerializer != null &&
+				this.valueSerializer != null) {
+			return new KafkaProducer<K, V>(this.configs, this.keySerializer, this.valueSerializer);
+		}
+		else {
+			return new KafkaProducer<K, V>(this.configs);
```
artembilan: DITO for both concerns and in this class, too.

spring-kafka/src/main/java/org/springframework/kafka/config/SimpleKafkaListenerContainerFactory.java
kvmkreddy: I think I really screwed up my git flows. I am adding new PR link to here

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
kvmkreddy: `And I even maybe interested only in key or value Deserializer but not both.`
In that case , how user inject different (De)Serilaizer for key and value?

artembilan: Again: two setters. Separate for `key` and for `value`.

kvmkreddy: Any specific reason, why you don't want this overloaded constructor approach?

kvmkreddy: Anyway  please find the #68 for latest implementation.

artembilan: Just because they aren't required options.
It is some style of our code.
If the option is required it goes to the ctor argument, otherwise it is setter do not confuse end user.
And again: with this new ctor I am forced to specify both `Deserializer`s , like `null` one of them, if I'm not interested in both. E.g. I would rely just on the default `Deserializer` for `key`, but would provide some specific for `value`.
So, I don't see reason to overhead it with optional ctor.

artembilan:
Closed in favor of #68

PR #67, opened 2016-05-09
GH66: Added ConsumerRebalanceListener injection feature

kvmkreddy:
Added feature to inject user defined ConsumerRebalanceListener implementations to both KafkaMessageListenerContainer and ConcurrentMessageListenerContainer.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
@@ -23,6 +23,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.regex.Pattern;
 
+import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
 import org.apache.kafka.common.TopicPartition;
 
 import org.springframework.kafka.core.ConsumerFactory;
@@ -40,6 +41,7 @@ import org.springframework.util.Assert;
  *
  * @author Marius Bogoevici
  * @author Gary Russell
+ * @author Murali Reddy
  */
 public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageListenerContainer<K, V> {
 
@@ -56,6 +58,8 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 	private TopicPartition[] partitions;
 
 	private int concurrency = 1;
+
+	private ConsumerRebalanceListener consumerRebalanceListener;
 
 	/**
 	 * Construct an instance with the supplied configuration properties and specific
@@ -134,6 +138,15 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 	}
 
 	/**
+	 * Set the user defined {@link ConsumerRebalanceListener} implementation.
+	 *
+	 * @param consumerRebalanceListener the {@link ConsumerRebalanceListener} instance
+	 */
+	public void setConsumerRebalanceListener(ConsumerRebalanceListener consumerRebalanceListener) {
+		this.consumerRebalanceListener = consumerRebalanceListener;
+	}
+
+	/**
 	 * Return the list of {@link KafkaMessageListenerContainer}s created by
 	 * this container.
 	 * @return the list of {@link KafkaMessageListenerContainer}s created by
@@ -156,15 +169,16 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 				this.concurrency = this.partitions.length;
 			}
 			setRunning(true);
+			logger.info("Topics from ConcurrentMessageListenerContainer-->:" + (this.topics));
 			for (int i = 0; i < this.concurrency; i++) {
 				KafkaMessageListenerContainer<K, V> container;
 				if (this.partitions == null) {
-					container = new KafkaMessageListenerContainer<>(this.consumerFactory, this.topics,
-							this.topicPattern, this.partitions);
+					container = new KafkaMessageListenerContainer<>(this.consumerFactory, this.consumerRebalanceListener,
+							this.topics, this.topicPattern, this.partitions);
 				}
 				else {
-					container = new KafkaMessageListenerContainer<>(this.consumerFactory, this.topics,
-							this.topicPattern, partitionSubset(i));
+					container = new KafkaMessageListenerContainer<>(this.consumerFactory, this.consumerRebalanceListener,
+							this.topics, this.topicPattern, partitionSubset(i));
 				}
 				container.setAckMode(getAckMode());
 				container.setAckCount(getAckCount());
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -156,15 +169,16 @@ protected void doStart() {
 				this.concurrency = this.partitions.length;
 			}
 			setRunning(true);
+			logger.info("Topics from ConcurrentMessageListenerContainer-->:" + (this.topics));
```
artembilan: Thank you, @kvmkreddy, for contribution!

Everything looks good and really is ready for merge.
Although I have several simple concerns. 
So, let me express them with the hope that you'll share your mind.

What is the point of this `logger.info()` for `this.topics`?
I think we still can live without it. 
The `this.topics` may be `null` and even it is able to be changed at runtime, it does not make much sense to log that anyway. Just because this value isn't changed internally without an external end-user involvement.

More over we don't use `-->` formatting in the Framework logs, plus it looks like you should have a whitespace after colon :smile:

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -83,7 +86,20 @@
...
+
+	/**
+	 * Construct an instance with the supplied configuration properties, {@link ConsumerRebalanceListener} implementation
+	 * and specific topics/partitions - when using this constructor, {@link #setRecentOffset(long)
+	 * recentOffset} can be specified.
+	 * @param consumerFactory the consumer factory.
+	 * @param consumerRebalanceListener {@link ConsumerRebalanceListener} implementation.
+	 * @param topicPartitions the topics/partitions; duplicates are eliminated.
+	 */
+	public KafkaMessageListenerContainer(ConsumerFactory<K, V> consumerFactory, ConsumerRebalanceListener consumerRebalanceListener,
```
artembilan: We would like to avoid the code line length more than 120 symbols: https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-Framework-Code-Style.
Right, this is minor and can be fixed on merge.
Just want to share with you our code style rules with the hope that you may follow them in the future contribution to let us avoid extra work during review/merge.

artembilan:
That's all.
I could fix those simple concerns on merge, but I see more unaddressed options:
1. I think this `ConsumerRebalanceListener` must be present on the `SimpleKafkaListenerContainerFactory`. 
2. Something has to be mentioned in the Reference Manual (`kafka.adoc`) for this new option.

@kvmkreddy , we may address those issues in the separate tickets, but that really would be better that you do that within this PR.

Thank you anyway!

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
kvmkreddy: This is unnecessary and removed.

artembilan: For the future: If you agree with our comments, not reason to say "Agreed".
Just make a change silently and notify us about the push.
Otherwise we receive extra spam in our mail box. :smile:

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
kvmkreddy: As of now corrected this line and going forward will take care. Thanks for pointing me.

kvmkreddy:
@artembilan 
For #1, Even now `SimpleKafkaListenerContainerFactory` has `ConsumerRebalanceListener`. The only difference is users can add their own implementations through constructor instead of property injection.
 
For #2, Updated document with appropriate instructions.

artembilan:
> Even now `SimpleKafkaListenerContainerFactory` has `ConsumerRebalanceListener`

??? Please, open `SimpleKafkaListenerContainerFactory` and show me the code where you see  `ConsumerRebalanceListener` mentioning.

src/reference/asciidoc/kafka.adoc
```diff
@@ -174,6 +188,8 @@ public ConcurrentMessageListenerContainer(ConsumerFactory<K, V> consumerFactory,
 It also has a property `concurrency`, e.g. `container.setConcurrency(3)` will create 3
 `KafkaMessageListenerContainer` s.
 
+You can also configure rebalance listener by using `consumerRebalanceListener`
```
artembilan: ... property.
And dot in the end of the sentence is required. :smile:

kvmkreddy: For #1 ,
```

	private final ConsumerRebalanceListener consumerRebalanceListener;

	KafkaMessageListenerContainer(ConsumerFactory<K, V> consumerFactory,
					ConsumerRebalanceListener consumerRebalanceListener,
					String[] topics, Pattern topicPattern, TopicPartition[] topicPartitions) {
		this.consumerFactory = consumerFactory;
		this.consumerRebalanceListener = (consumerRebalanceListener == null) ? createConsumerRebalanceListener() :
			consumerRebalanceListener;
		this.topics = topics == null ? null : Arrays.asList(topics).toArray(new String[topics.length]);
		this.topicPattern = topicPattern;
		this.partitions = topicPartitions == null ? null : new LinkedHashSet<>(Arrays.asList(topicPartitions))
				.toArray(new TopicPartition[topicPartitions.length]);
	}
```

artembilan:
Please, don't use the `#1,#2` style, if you really aren't going to point to the particular issue/PR.
That would even better to quote other comment via `>` symbol in the beginning of the comment line.

src/reference/asciidoc/kafka.adoc
```diff
@@ -174,6 +188,8 @@ public ConcurrentMessageListenerContainer(ConsumerFactory<K, V> consumerFactory,
 It also has a property `concurrency`, e.g. `container.setConcurrency(3)` will create 3
 `KafkaMessageListenerContainer` s.
 
+You can also configure rebalance listener by using `consumerRebalanceListener` proerty.
```
artembilan: Typo: pro**p**erty.

artembilan:
@kvmkreddy , everything is good and I can fix the latest minor issues on merge, but I would like to know why you don't want to introduce the `ConsumerRebalanceListener` property into the `SimpleKafkaListenerContainerFactory`. This is used for the `@KafkaListener` POJO-method level abstraction. So, don't including the `ConsumerRebalanceListener` option there we are forced to implement our own `KafkaListenerContainerFactory` to accomplish the  `ConsumerRebalanceListener` requirements from the `@KafkaListener` level.

kvmkreddy:
I am not against to `ConsumerRebalanceListener` injection into the `SimpleKafkaListenerContainerFactor` . Now I added this implementation. Can you check once again.

artembilan:
I think I'm fine with the solution.
Merging...
Thank you!

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/e857c248eb88116bfd28ccf05671e5be13741f96

Issue #61, opened 2016-05-05
How do we inject user defined Key/Value serializer/deserializer objects?


kvmkreddy:

garyrussell:
Just add them to the consumer/producer properties - see [this test case](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java#L82).

kvmkreddy:
Thank you very much for your quick response.

How can we inject user defined decoders references instead of Kafka creating/stubbing object?
Example of user defined decoder:
```
 	<bean id="kafkaConfluentValueDecoder" class="com.my.own.KafkaAvroDecoder">
		<constructor-arg >
			<bean class="kafka.utils.VerifiableProperties">
				<constructor-arg>
					<props>
						<prop key="schema.registry.url">http://localhost:8081</prop>
						<prop key="something.else.key">something_else_value</prop>
					</props>
				</constructor-arg>
			</bean>
		</constructor-arg>
	</bean>
```

garyrussell:
The Kafka 0.9 client doesn't support it - it has no APIs for that.

The scala `Decoder` and `Encoder` APIs are not supported by the pure Java 0.9 API.

You need to implement `Serializer` (and `Deserializer`) and the producer/consumer properties object (in which the class is defined is passed into the `configure()` method.

garyrussell:
Please reopen if you have more concerns.

kvmkreddy:
Both kafka consumer and producer implementations are provided API where we can inject custom Serializer/Deserializer implementations.

Find the below consumer API for example:
```
    public KafkaConsumer(Map<String, Object> configs,
                         Deserializer<K> keyDeserializer,
                         Deserializer<V> valueDeserializer) {
        this(new ConsumerConfig(ConsumerConfig.addDeserializerToConfig(configs, keyDeserializer, valueDeserializer)),
            keyDeserializer,
            valueDeserializer);
    }
```

My thoughts on spring kafka API:

It should allow clients to inject their own serializer/deserializer implementations in addition to providing 
default behavior of kafka API to create these instances.

For example [DefaultKafkaConsumerFactory](https://github.com/kvmkreddy/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java) implementation will be like 
```
private KafkaConsumerProducerStrategy<K,V> strategy;

[...]

PR #68, opened 2016-05-10
GH-61-1: User defined serializer and deserializer injection

kvmkreddy:


spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
```diff
@@ -265,7 +272,8 @@ public MultiListenerBean multiListener() {
 
 		@Bean
 		public ProducerFactory<Integer, String> producerFactory() {
-			return new DefaultKafkaProducerFactory<>(producerConfigs());
+			Map<String, Object> configs = producerConfigs();
+			return new DefaultKafkaProducerFactory<>(configs);
```
artembilan: Technically I don't see difference with existing code and your change.
Just some unnecessary noise...
Will revert on merge.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
```diff
@@ -173,6 +174,48 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
 	}
 
 	@Test
+	public void testAutoCommitWithCompositeStrategy() throws Exception {
```
artembilan: We don't have such a `strategy` any more. Therefore the test case reflect some parallel irrelevant reality.
Will fix on merge.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
```diff
@@ -173,6 +174,48 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
...
+		logger.info("Start auto");
+		Map<String, Object> props = KafkaTestUtils.consumerProps("test1", "true", embeddedKafka);
+
+		DefaultKafkaConsumerFactory<Integer, String> cf =
+				new DefaultKafkaConsumerFactory<Integer, String>(props,
+						new org.apache.kafka.common.serialization.IntegerDeserializer(),
+						new org.apache.kafka.common.serialization.StringDeserializer());
```
artembilan: M-m-m. I think we don't need such a fully qualified class names.
Just `import` should be enough.
Will fix on merge.

src/checkstyle/checkstyle.xml
```diff
@@ -178,4 +178,4 @@
 		<module name="WhitespaceAround" />
 
 	</module>
-</module>
+</module>
```
artembilan: We are coming back to this issue.
Each file has to have a new line in the end.
The IDE has a particular setting on the matter.

Although it is fully unclear what do you do in these files if we don't see any changes, only this noise :smile:

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
```diff
@@ -22,28 +23,55 @@
...
 	public DefaultKafkaConsumerFactory(Map<String, Object> configs) {
 		this.configs = new HashMap<>(configs);
 	}
 
+	public DefaultKafkaConsumerFactory(Map<String, Object> configs,
+			Deserializer<K> keyDeserializer,
+			Deserializer<V> valueDeserializer) {
+		this.configs = new HashMap<>(configs);
+		this.keyDeserializer = keyDeserializer;
+		this.valueDeserializer = valueDeserializer;
```
artembilan: I understand that the `return new DefaultKafkaConsumerFactory(...)` looks good, but logically it is redundant extra noisy code.
But with respect to your contribution I leave it here and merge.
:crossed_flags:

src/checkstyle/checkstyle.xml
kvmkreddy: It seems some issues with my IDE which is corrupting this file now and then.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/48845e1de043ec9caf6493c7e2320966e7986a9f.

Thank you for the contribution!

Looking forward for more.

Hope the new PR won't be so tough on merge. :smile: 
The rebase and conflict fixing are really the branch author responsibility.

PR #77, opened 2016-05-13
GH-76: Resolve Template Ambiguity with String Keys

### LINKED ISSUE #76: Disambiguate KafkaTemplate Methods
When `<K, V>` are `<String, ?>`, `send(K key, V data)` and `send(String topic, V data)` are ambiguous.

garyrussell:
Resolves #76


src/reference/asciidoc/kafka.adoc
```diff
@@ -119,6 +123,13 @@ You may wish to invoke `flush()` before waiting or, for convenience, the templat
 parameter which will cause the template to `flush()` on each send.
 Note, however that flushing will likely significantly reduce performance.
 
+===== Ambiguous Methods
+
+As you can see, the template has generic types `K` and `V` for key and value types.
+Notice there is one method `sendKeyValue` - this is to resolve an ambiguity when the generic type `K` is `String`.
+The `send(K key, V data)` method is ambiguous with `send(String partition, V data)` when the key type is `String`;
```
mbogoevici: Should be "is ambiguous with `send(String topic, V data)`"

src/reference/asciidoc/kafka.adoc
```diff
@@ -12,6 +14,8 @@ ListenableFuture<SendResult<K, V>> send(V data);
 
 ListenableFuture<SendResult<K, V>> send(K key, V data);
 
+ListenableFuture<SendResult<K, V>> sendKeyValue(K key, V data);
```
mbogoevici: One alternative thought here - instead of introducing this method, how about we rename the other two as `sendToTopic` and `sendToPartition` - maybe make overload it to take a topic as argument as well?

spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java
```diff
@@ -157,6 +159,7 @@ protected void before() throws Exception { //NOSONAR
 		for (String topic : this.topics) {
 			AdminUtils.createTopic(zkUtils, topic, this.partitionsPerTopic, this.count, props);
 		}
+		System.setProperty(SPRING_EMBEDDED_KAFKA_BROKERS, getBrokersAsString());
```
artembilan: If we are expose it into the `System Properties` in the beginning, we have to clean up it in the end.

garyrussell:
@mbogoevici Yeah - I wasn't really happy with it; how about this instead (`sendDefault()` for those methods that send to the default topic)?

@artembilan Do you think it would be better to not set it at all?

mbogoevici:
@garyrussell I think `sendDefault` is better - and quite self-explanatory too. I like it!

artembilan:
> Do you think it would be better to not set it at all?

I'm fine with the option. I just worry that since we involve "external" resource for the test, we should free it in the end. Fortunately, we have the `after()`  there already.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
```diff
@@ -36,15 +36,19 @@
...
 	 */
-	ListenableFuture<SendResult<K, V>> send(V data);
+	ListenableFuture<SendResult<K, V>> sendDefault(V data);
 
 	/**
 	 * Send the data to the default topic with the provided key and no partition.
+	 * <p><em>Note: when the generic type K is {@link String}, this method is
+	 * ambiguous with {@code send(partition, data)}.
+	 * {@link #sendKeyValue(Object, Object)} is provided as an alternative
+	 * method to use in that situation</em>
```
mbogoevici: We can probably remove this javadoc now :)

garyrussell:
Pushed - I though I'd removed that javadoc - thanks.

artembilan:
I'm fine what we have here so far.

@mbogoevici , any objection do not merge yet?
Thank you for coming back to us anyway! :smile:

spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java
```diff
@@ -382,4 +391,31 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
...
 	}
 
+	/**
+	 * Subscribe a consumer to one of the embedded topics.
+	 * @param consumer the consumer.
+	 * @param topic the topic.
+	 * @throws Exception an exception.
+	 */
+	public void consumeFromAnEmbeddedTopic(Consumer<?, ?> consumer, String topic) throws Exception {
+		assertThat(Arrays.asList(this.topics).contains(topic)).as("topic is not in emebedded topic list").isTrue();
```
artembilan: Typo: em~~e~~bedded 

Will fix on merge.

artembilan: Plus I'd prefer this assertion:
```
assertThat(Arrays.asList(this.topics)).as("topic is not in embedded topic list").contains(topic);
```
Looks like it more useful when we fail:
```
java.lang.AssertionError: [topic is not in embedded topic list] 
Expecting:
 <["intKeyTopic", "stringKeyTopic"]>
to contain:
 <["foo"]>
but could not find:
 <["foo"]>
```

garyrussell: 👍

mbogoevici:
@artembilan Looks good to me!

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/da88336b8324589228ccf9c81a62cee493b7d395

Issue #79, opened 2016-05-14
Easy POJO to JSON serialization (and back)


stepio:
I've implemented (de)serialization for POJO messages as a separate project:
https://github.com/stepio/kafka-json

In my opinion it's too tiny to be supported as a separate functionality, but it would be nice as a part of Spring Kafka.
What do you think? If you agree on this, what package would you suggest?

garyrussell:
Thanks @stepio - while the project currently has JSON support vis the `StringJsonMessageConverter`, I agree it would be nice to include this in the project. I would say the new package should be `o.s.k.support.serializer`.

Please be sure to follow the [contribution guidelines](https://github.com/spring-projects/spring-kafka/blob/master/CONTRIBUTING.adoc), especially "signing" the CLA, "Spring Kafka" doesn't currently appear in the drop down, but you can use "Spring Integration". Also take note of the code formatting needed (license header etc).

Please note that we are planning to release the release candidate next Monday or Tuesday, so a quick turnaround would be very helpful. However, since this is just new code and not changing existing, we can add it after the candidate is released and still get it in the final release.

Thanks !!

stepio:
Looks good for me, thanks!

artembilan:
The PR #81 has been merged and closed

PR #81, opened 2016-05-14
GH-79: Easy POJO to JSON serialization (and back)

stepio:
Implementing the feature, discussed in scope of #79

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonWrapperException.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonWrapperException.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.support.serializer;
+
+import org.apache.kafka.common.KafkaException;
+
+/**
+ * Specific wrapper, extending {@link KafkaException}.
+ *
+ * @author Igor Stepanov
+ */
+public class JsonWrapperException extends KafkaException {
+
+    private static final long serialVersionUID = -4682389372221504570L;
+
+    public JsonWrapperException(Throwable cause) {
+        super(cause);
+    }
+}
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonWrapperException.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonWrapperException.java
@@ -25,9 +25,9 @@ import org.apache.kafka.common.KafkaException;
  */
 public class JsonWrapperException extends KafkaException {
 
-    private static final long serialVersionUID = -4682389372221504570L;
+	private static final long serialVersionUID = -4682389372221504570L;
 
-    public JsonWrapperException(Throwable cause) {
-        super(cause);
-    }
+	public JsonWrapperException(Throwable cause) {
+		super(cause);
+	}
 }
--- a/build.gradle
+++ b/build.gradle
@@ -71,9 +71,10 @@ subprojects { subproject ->
 	ext {
 		assertjVersion = '3.3.0'
 		hamcrestVersion = '1.3'
-		jacksonVersion = '2.3.2'
+		jacksonVersion = '2.6.5'
 		junitVersion = '4.12'
 		kafkaVersion = '0.9.0.1'
+		lombokVersion = '1.16.4'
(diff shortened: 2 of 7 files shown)


stepio:
Applied Spring formatting conventions & created an issue #82

stepio:
More questing on the process:
1. Where is the applied checkstyle configuration available?
2. Is it possible to modify the checkstyle rules? For example I'm not sure that it's wise to fix imports after they're optimised by Idea.

stepio:
Also why do you use gradle for this project, why not maven? Maven would have allowed to ease the configuration for all the checks and tests to run them locally.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonWrapperException.java
```diff
@@ -0,0 +1,33 @@
...
+package org.springframework.kafka.support.serializer;
+
+import org.apache.kafka.common.KafkaException;
+
+/**
+ * Specific wrapper, extending {@link KafkaException}.
+ *
+ * @author Igor Stepanov
+ */
+public class JsonWrapperException extends KafkaException {
```
stepio: Not sure about this - maybe I should use the already defined project-specific [KafkaException](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/KafkaException.java).

Pros:
 - it's not good to define a new Exception for each package;
Cons:
 - my exception extends KafkaException from 'kafka-client' which is good for such a low-level functionality.

garyrussell:
Thanks for the contribution, sorry for the pain with the initial setup to comply with the style/rules. 

I didn't receive a notification from the CLA system that you "signed" the CLA - please be sure to do so and confirm here, otherwise we can't look at your work.

To answer your specific questions:

- the checkstyle configuration is in `src/checkstyle`.
- IDEA can be configured to comply with the import rules - @artembilan can help you with that, I use eclipse/STS
- you can run `./gradlew check` to run the checkstyle rules locally

artembilan:
Hello, @stepio, again ! 

Thank you for contribution one more time, too!

As @garyrussell said already, we should receive a confirmation of the signed CLA for you before we start to take a look to code change.

Meanwhile, please, elaborate more:
> Maven would have allowed to ease the configuration for all the checks and tests to run them locally.

Really not sure what you mean. When I write code locally, I run tests from the IDE and don't worry about the management and build system on the project.
In the end, before pushing commit I run `./gradlew clean test --parallel`. I'm sure with Maven I'd do something similar.
So, that isn't clear what the problem you have with tests to run locally. 
Since you don't have (yet) appropriate code style settings that isn't surprise that your won't pass Checkstyle rules. We'll do that for your on merge, until we figure out what to do with automatic settings via project sources.

stepio:
> I didn't receive a notification from the CLA system that you "signed" the CLA - please be sure to do so and confirm here, otherwise we can't look at your work.
That's because I've signed it before (for Spring Boot), sorry for not mentioning this. Should I specify the Id here?

stepio:
Regarding the maven, you're right, now I see that gradle allows the same.

I'm just a little maven-addicted, so does not have much experience with gradle.
These hints were not mentioned on start page:
> you can run ./gradlew check to run the checkstyle rules locally
> ./gradlew clean test --parallel

With maven I've got used to running just "mvn clean install". This command does all the job, including checks, tests and everything else configured by project's owner.

Also maven is fully supported by IDEs - importing is easy both in Eclipse & IDEA.

And nice dependency management, of course - Spring Boot shows its advantages at full.

So... If you surprisingly think about switching to maven, I'd love to try implementing this. Just in case :)

artembilan:
> So... If you surprisingly think about switching to maven, I'd love to try implementing this. Just in case :)

Sorry, we don't . :wink: 

Thank you, of course, for sharing your Maven experience, but no. We are happy with Gradle but many reasons and won't back to Maven in any observable future. 

If you aren't familiar with Gradle, just don't worry about that and ask us which commands to use for each case. It is out of project scope to teach people how to use the build tool. Maven and Gradle are open source an well documented, so no reason to document which commands to run.

On the other hand, I'm sure that with Maven you even don't pay attention that we don't have any hints in the documentation. Just because you know how to use that tool! :smile: 

Gradle has good IDE integration as well, but that is different story, too.

Not sure about " nice dependency management" in Maven. Because in some big project we can't override the dependency provided by Spring Boot. In case of Gradle that works well just with explicit `version`. 

But, yes: as I mentioned to you in one discussion already, the documentation deserves more love. Especially in case we would like to have more and more contribution from the Community. 
But right now we have a different priorities to come up with as much as possible features in the project to make it very useful and lovely for the same Community.

garyrussell:
>Should I specify the Id here, or should I keep it private?

No need to share the number, we just need you to confirm you have signed, which you have now done.

Thanks again for the contribution.

stepio:
Thanks for sharing your experience!

No problem, gradle is Ok for me :)

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/db17d6620a24106bf87187a45b6345d9dfabf733
with subsequent polishing https://github.com/spring-projects/spring-kafka/commit/79250ff7a6b46c73f718067e4ad124d9c06e4618.

Sorry, we don't have enough time to review, comment and wait for your feedback properly, but thank you for your contribution anyway.
Waiting for more!

Some general comments anyway:
1. We don't use Lombok. It is enough easy to avoid such a noise and just generate setters/getters and equals/hashCode using IDE abilities.
2. Our code style is camel-case, not underlining.

Everything else you can see in my commit. So, feel free to comment there!

PR #83, opened 2016-05-16
GH-80: Add DeDuplication Listener Adapters

### LINKED ISSUE #80: Message duplication after ConsumerRebalancing
It's not a bug report, more like a consultancy request... maybe an idea for the feature.

While playing with kafka-console-consumer I've spotted that sometimes after consumer rebalancing same message is received the second time. Checked google, read the article:
https://cwiki.apache.org/confluence/display/KAFKA/Idempotent+Producer
As I understand, it's not a bug, but kind of limitation.

But we all know that in many cases message uniqueness is important & same message should not be processed second time. So is there any already implemented approach to achieve such uniqueness?

As of now, I see the next quite easy mechanism, which may help:
1. On producer side: each message should be wrapped into a "holder" object with unique id (UUID).
2. On consumer side: each message Id should be compared with the previous one (or ones) - if such an id is already known, the received message should be logged & discarded.

What do you think?

P.S.: This may be related to #66 & #79

garyrussell:
Resolves #80


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AbstractDeDuplicatingMessageListener.java
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.listener.adapter;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+
+/**
+ * A abstract message listener adapter that implements de-duplication logic
```
artembilan: Typo: A**n**

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/DeduplicatingAcknowledgingMessageListenerAdapter.java
```diff
@@ -0,0 +1,52 @@
...
+ * A {@link AcknowledgingMessageListener} adapter that implements de-duplication logic
+ * via a DeDuplicationStrategy.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Gary Russell
+ *
+ */
+public class DeduplicatingAcknowledgingMessageListenerAdapter<K, V> extends AbstractDeDuplicatingMessageListener<K, V>
```
artembilan: Why camel case style has been lost here? I mean why not De**D**uplicating ?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/DeduplicatingAcknowledgingMessageListenerAdapter.java
```diff
@@ -0,0 +1,52 @@
...
+
+package org.springframework.kafka.listener.adapter;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+
+import org.springframework.kafka.listener.AcknowledgingMessageListener;
+import org.springframework.kafka.support.Acknowledgment;
+
+/**
+ * A {@link AcknowledgingMessageListener} adapter that implements de-duplication logic
```
artembilan: `An` too ?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AbstractDeDuplicatingMessageListener.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author Gary Russell
+ *
+ */
+public abstract class AbstractDeDuplicatingMessageListener<K, V> {
+
+	private final DeDuplicationStrategy<K, V> deDupStrategy;
+
+	protected AbstractDeDuplicatingMessageListener(DeDuplicationStrategy<K, V> deDupStrategy) {
+		this.deDupStrategy = deDupStrategy;
```
artembilan: `Assert.notNull()` ? Otherwise this object does not make sense. 
Moreover you have an `NPE` below.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/DeduplicatingMessageListenerAdapter.java
```diff
@@ -0,0 +1,51 @@
...
+ * A {@link MessageListener} adapter that implements de-duplication logic
+ * via a DeDuplicationStrategy.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Gary Russell
+ *
+ */
+public class DeduplicatingMessageListenerAdapter<K, V> extends AbstractDeDuplicatingMessageListener<K, V>
```
artembilan: DITO for camel case style

spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
```diff
@@ -95,6 +96,9 @@
 	@Autowired
 	public KafkaListenerEndpointRegistry registry;
 
+	@Autowired
+	private DeDuper deDuper;
```
artembilan: Funny name, like the Russian phase from the street: `supper, puper, duper` :smile: 
But let it be, because I am not English specialist

src/reference/asciidoc/kafka.adoc
```diff
@@ -326,3 +326,23 @@ public void listen(@Payload String foo,
...
+
+In certain scenarios, such as rebalancing, a message may be redelivered that has already been processed.
+The framework cannot know whether or not such a message has been processed or not, that is an application-level
+function.
+This is known as the http://www.enterpriseintegrationpatterns.com/patterns/messaging/IdempotentReceiver.html[Idempotent
+Receiver] pattern and Spring Integration provides an http://docs.spring.io/spring-integration/reference/html/messaging-endpoints-chapter.html#idempotent-receiver
+[implementation thereof].
+
+The Spring for Apache Kafka project also provides some assistance by means of the `DeduplicatingMessageListenerAdapter`
+and `DeduplicatingAcknowledgingMessageListenerAdapter` classes, which can wrap your `MessageListener` or
```
artembilan: If you agree with me about camel case, these two mentionings should be addresses as well.

src/reference/asciidoc/kafka.adoc
```diff
@@ -326,3 +326,23 @@ public void listen(@Payload String foo,
...
+`AcknowledgingMessageLister` respectively.
+Each of these classes takes an implementation of `DeDuplicationStrategy` where
+you implement the `isDuplicate` method to signal that a message is a duplicate and should be discarded.
+
+Similarly, when using `@KafkaListener`, the `DeduplicationStrategy` can be injected into the container factory.
```
artembilan: This is typo: De**D**uplication

src/reference/asciidoc/kafka.adoc
```diff
@@ -326,3 +326,23 @@ public void listen(@Payload String foo,
...
+
+Of course, if you are using manual acknowledgments you should not use these classes because you would need to
+acknowledge the duplicated message.
```
artembilan: I don't understand this phrase.
We have `AcknowledgingMessageListener` which definitely assumes a `manual` acknowledgment. So, according this your conclusion the `DeduplicatingAcknowledgingMessageListenerAdapter` is a bad class.
Or we should : 
```
else {
	acknowledgment.acknowledge();
}
```
in case of duplicate anyway there...

garyrussell: Good point - I guess I should just remove the wrapper for the Acking listener.

garyrussell:
Pushed

artembilan:
I'm fine with the fix, but wonder if the Travis failure rings a bell for you.

Also. Anyway I'm still concerned why we can't manually acknowledge the duplicate in the particular `AcknowledgingMessageListener` ... m-m-m... automatically ?
What is your point on the matter?
Thanks

garyrussell:
I am not sure what the travis problem is; it builds fine for me - I restarted travis. Let's see.

Regarding your last comment - we can't make any assumptions about what the user's paradigm is for acknowledging - he might want to commit every 100 records and we'd be introducing some random "auto" ack when he's not expecting it. IMHO manual acking needs to be, er, always manual.

I suppose we could have another method on the interface that gets the record and the ack as arguments - so the user can choose, but I'm not sure I want to add more complexity. Such a user could take this wrapper/adapter approach and put whatever logic he wants regarding acking.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/5fa753cbb99057cd61a07c5ea85e2d32dc5db4ce.

Don't see reason to continue discussion, since everything sounds reasonable.
Will reconsider on demand from the community.

artembilan:
Sorry, forgot to mention. Upgraded to Gradle 2.13

PR #85, opened 2016-05-16
ISSUE-84: Pause consumer if message listener can't accept message

martindam:


martindam:
I have signed and agree to the terms of the SpringSource Individual Contributor License Agreement.

garyrussell:
Interesting technique; I was assuming this is something that would be pushed up to the listener (and perhaps in a listener adapter):

     String consumerId = container.pause(); // must be run on the listener thread

      ... 

     container.resume(consumerId); // queued to listener thread due to the consumer not being thread-safe

However, your technique has the added benefit of "automatically" resuming when the listener is freed up, but at the expense of retrying the delivery repeatedly until then.

@mbogoevici @artembilan WDYT?

I am not sure we want to use a general exception (MDE) to signal this behavior - maybe a more specific exception. OTOH, this approach means any other exception discards the message.

artembilan:
I'm not sure yet what we would like to address here, but let me quote `KafkaConsumer` JavaDocs on the matter:
````java
* <h4>Consumption Flow Control</h4>
 *
 * If a consumer is assigned multiple partitions to fetch data from, it will try to consume from all of them at the same time,
 * effectively giving these partitions the same priority for consumption. However in some cases consumers may want to
 * first focus on fetching from some subset of the assigned partitions at full speed, and only start fetching other partitions
 * when these partitions have few or no data to consume.
 *
 * <p>
 * One of such cases is stream processing, where processor fetches from two topics and performs the join on these two streams.
 * When one of the topic is long lagging behind the other, the processor would like to pause fetching from the ahead topic
 * in order to get the lagging stream to catch up. Another example is bootstraping upon consumer starting up where there are
 * a lot of history data to catch up, the applciations usually wants to get the latest data on some of the topics before consider
 * fetching other topics.
 *
 * <p>
 * Kafka supports dynamic controlling of consumption flows by using {@link #pause(TopicPartition...)} and {@link #resume(TopicPartition...)}
 * to pause the consumption on the specified assigned partitions and resume the consumption
 * on the specified paused partitions respectively in the future {@link #poll(long)} calls.
````
So, looks like Flow Control with `pause/resume` is for a bit different purpose...
But I may be wrong. 
Investigating...

mbogoevici:
The Kafka Consumer needs to send heartbeats in order not to be considered gone and lose its partitions.  The only way that can be done is via `poll` (the consumer is single threaded), so `pause`/`resume` allow the consumer to performe 'no-op' polls to signal that it is alive.

garyrussell:
We've been discussing this internally and we're not thrilled with using an exception to signal the pause.

A couple of us a traveling to Spring IO so will be off-air for much of this week; I have decided to delay the RC/GA so we can get this "right".

I haven't discussed this with the team, but here's one suggestion:

A new flag `enableFlowControl`. 

When `true`, instead of calling the listener on the consumer thread, hand off deliveries one at a time to another thread which calls the listener. Arm a timer (configurable) such that if the listener thread doesn't return, pause the consumer; resume it when the listener thread returns (and consumes the pending messages in the current batch).

martindam:
That is fair, we should get this thing right as it is the low level part that feeds into the rest of the Kafka implementations. I'll be happy to assist with these changes. 

Reg. why I did it like this, I wanted to integrate with Spring Integration and the standard channels while ensuring that I do not commit an offset before it has been delivered to the listener correctly. 

Having multiple threads would be a good idea as it would solve the issue that the processing off the entire batch has to be done in less then session.timeout.ms (which should be low) so if you hooked my implementation up to a DirectChannel you could still trigger the rebalance if you can't consume the batch within session.timeout.ms. I am currently using this implementation with a QueueChannel which "enables" the flow control.

artembilan:
Closed in favor of #91 

Thank you for the contribution anyway.
Looking forward for more!

Issue #88, opened 2016-05-18
MINOR: Replace "org.apache.commons.logging" with "org.slf4j"


stepio:
Is there any reason to use commons-logging instead of slf4j?

I can update this, if it's not a feature.

P.S.: Also could you please explain the reason of not making Logger static?

artembilan:
Hello, @stepio !

See here, please: https://github.com/spring-projects/spring-integration/issues/1782

Re. `static` VS non-`static`.
Well, it depends. Sometimes it is much better just let people to extend our classes and reuse super methods, as well as reuse our `protected` logger. So, all actions in the instance will be logged with the same logger.
See difference, please:
````java
private static final Log logger = LogFactory.getLog(Foo.class);
...
protected final Log logger = LogFactory.getLog(getClass());
````
So, in the second case `getClass()` is exactly a class of the target object.

stepio:
That explains much - thanks for details!
I've implemented backward change for my code here: f8e3e53592b87718613ca6823785e3f057bbddb8

May do the PR, although it's a minor polishing. Issue may be closed, case is clear.

artembilan:
Oh! I missed there that you use slf4j. :smile: 

Please, raise a PR on the matter with this issue as a subject.

Thank you for contribution and sane discussion!

artembilan:
I think we even should reconsider our dependencies excluding any transitive `slf4j` and that `compile log4j` to avoid exactly undesired cases like this in the `JsonDeserializer`.

PR #90, opened 2016-05-18
GH-88: Stop using slf4j instead of commons-logging to follow common approach

stepio:
Refers to issue #88

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -84,6 +84,7 @@ subprojects { subproject ->
 		log4jVersion = '1.2.17'
 		mockitoVersion = '1.9.5'
 		scalaVersion = '2.11'
+		slf4jVersionRuntime = '1.7.21'
 		springRetryVersion = '1.1.2.RELEASE'
 		springVersion = '4.2.6.RELEASE'
 
@@ -147,13 +148,17 @@ project ('spring-kafka') {
 
 	dependencies {
 		compile "org.springframework:spring-messaging:$springVersion"
-		compile "org.apache.kafka:kafka-clients:$kafkaVersion"
+		compile ("org.apache.kafka:kafka-clients:$kafkaVersion") {
+			exclude group: 'org.slf4j'
+		}
 		compile ("com.fasterxml.jackson.core:jackson-core:$jacksonVersion", optional)
 		compile ("com.fasterxml.jackson.core:jackson-databind:$jacksonVersion", optional)
 
 		testCompile project (":spring-kafka-test")
 		testCompile "org.assertj:assertj-core:$assertjVersion"
 		testCompile "org.springframework:spring-tx:$springVersion"
+
+		runtime ("org.slf4j:slf4j-jcl:$slf4jVersionRuntime", optional)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


artembilan:
Thank you, Игорь!
What do you think about my last comment in the issue: https://github.com/spring-projects/spring-kafka/issues/88#issuecomment-220067176 ?
Will you take care about that here or that would be better for you to leave that task for us?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -43,7 +43,7 @@
  */
 public class JsonDeserializer<T> implements Deserializer<T> {
 
-	protected final Logger logger = LoggerFactory.getLogger(getClass());
+	protected final Log logger = LogFactory.getLog(getClass());
```
artembilan: Well, there is no any usage of this property in this class any more.
So, it would be better to remove it altogether like I did in the `JsonSerializer`.
That was just my omission to leave it here.

Thank you for catching that anyway!

stepio:
That makes sense, let me try :)

stepio:
Done the change, Артем - please check now :)

build.gradle
```diff
@@ -147,7 +147,9 @@ project ('spring-kafka') {
 
 	dependencies {
 		compile "org.springframework:spring-messaging:$springVersion"
-		compile "org.apache.kafka:kafka-clients:$kafkaVersion"
+		compile ("org.apache.kafka:kafka-clients:$kafkaVersion") {
+			exclude group: 'org.slf4j'
```
artembilan: Well, exclude just `group` isn't enough for other our requirements like [Spring IO Platform](http://platform.spring.io/platform/). But that is already not your headache.
I'm merging it as is and think about dependencies offline.
Seems form me we should retain `slf4j` as `runtime` to let Kafka to log in our tests anyway.

artembilan:
BTW, @stepio , you should study to contribute from te top--level branch, not `master`: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc
Otherwise you have to `hard reset` to the `upstream`, not just `pull`.

Although I see some problems in Travis after your changes.
Have you tested the project locally?

stepio:
Studying the problem, sorry

build.gradle
stepio: My lack of experience with Gradle affects the result, sorry :(

build.gradle
```diff
@@ -84,6 +84,7 @@ subprojects { subproject ->
 		log4jVersion = '1.2.17'
 		mockitoVersion = '1.9.5'
 		scalaVersion = '2.11'
+		slf4jVersionRuntime = '1.7.21'
```
artembilan: It doesn't matter how we will use the variable. Our code convention here is just `...Version`
Will fix on merge.

build.gradle
```diff
@@ -147,13 +148,17 @@ project ('spring-kafka') {
...
+			exclude group: 'org.slf4j'
+		}
 		compile ("com.fasterxml.jackson.core:jackson-core:$jacksonVersion", optional)
 		compile ("com.fasterxml.jackson.core:jackson-databind:$jacksonVersion", optional)
 
 		testCompile project (":spring-kafka-test")
 		testCompile "org.assertj:assertj-core:$assertjVersion"
 		testCompile "org.springframework:spring-tx:$springVersion"
+
+		runtime ("org.slf4j:slf4j-jcl:$slf4jVersionRuntime", optional)
```
artembilan: This is `runtime` dependency, so it won't be included to the final POM for the artifact. Therefore `optional` is redundant.
Will fix on merge.

build.gradle
stepio: Okay.

stepio:
Latest travis build still failed although runs okay locally (did "clean", then "install", then "test"). Any hints?

build.gradle
stepio: Copy-paste, got the idea, thanks

artembilan:
That failure on the:
```
assertThat(this.ifaceListener.getLatch1().await(20, TimeUnit.SECONDS)).isTrue();
```
Isn't related to you. That is some kind of Kafka rebalancing bottleneck in the slow and restricted environment like Travis container or my local Windows :smile: 

Will take a look closely during merge.

artembilan:
Merged https://github.com/spring-projects/spring-kafka/commit/6a0813a5342d5763687ea93dcba8c079cbd130ed.
I changed it to use `log4j` for runtime, because exactly this one with you to log really. See `log4j.properties`.
And fixed all `slf4j` dependencies.

PR #91, opened 2016-05-24
GH-84: Pause/Resume

### LINKED ISSUE #84: Flow control support
This is more a feature request and an issue: While working with this project and the spring-integration-kafka project, I faced the problem that KafkaConsumer can fetch messages much faster than I can consume them. With the current implementation the batch of messages are read and each one is pushed downstream for processing. However, if it takes more than session.timeout.ms to process the batch (which can be thousands of messages if there is a large kafka lag) a rebalance will be triggered as KafkaConsumer was not able to send heart beats (it happens while calling poll()). This could be solved by placing a queueChannel after the KafkaMessageListenerContainer, but that just moves the problem as at some point that queue is going to fill up or I run out of memory. 

So my suggestion is this: Add support for some flow control in the KafkaMessageListenerContainer so I can start/pause the KafkaConsumer when my downstream processing pipe is backed up to prevent filling up all buffers. KafkaConsumer supports start/pause out of the box, so it allows to pause the consumer while it keeps sending heart beats (with poll()) but not receive more messages until it can be consumed downstream. This could then be wrapped in a integration component that could manage the flow control based on MessageDeliveryExceptions or similar.

I did an implementation like the above, but it required to copy the existing KafkaMessageListenerContainer to make it work due to most of the logic is in the inner class.

garyrussell:
Resolves #84

Slow consumers can cause a rebalance when not required.

Solution: 2 threads per consumer
 - beanName-kafka-consumer-n
 - beanName-kafka-listener-n

The first constantly polls Kafka; for each retrieved message then hands off
the message to the second thread (via a `ListenableFuture`).
If the future completes ok (`get()`, all is well and we move on to the next
message, if not, the consumer is paused and we continue to poll (heartbeat).

When the callback completes successfully, the consumer is awoken from the poll
and the next previously retrieved message is processed. When all such messages
are processed, the consumer is resumed.

If manual acks are in play, the ack is passed from the listener thread back to
the consumer thread via a `BlockingQueue` so the actual commit is done on the
right thread.

If the delayed execution fails, normally, the error is reported to the error
handler, the message is ack'd and we move on to the next as with a successful
completion.

However, to satisfy the use case reported by issue 84, if the exception is
assignable to the configured `pausedException` class, the consumer is
paused as with a timeout. In this case, after the next poll, the same
record is attempted to be delivered. Deliveries continue until successful
or some other exception is thrown.

Properties:

- `pauseWhenSlow` - enable pause/resume (default `true`)
- `pauseAfter` - how long to wait for the listener thread to return
- `pauseException` - an exception that will cause a pause

__Needs docs but hopefully the commit comment and javadocs explains things enough for review__


spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
```diff
@@ -83,11 +92,22 @@ public void setErrorHandler(ErrorHandler errorHandler) {
...
+	}
+
+	/**
+	 * Specify an {@link Executor} to use.
+	 * @param listenerTaskExecutor the {@link Executor} to use to invoke
+	 * the listener.
+	 * @see AbstractKafkaListenerContainerFactory#setListenerTaskExecutor
+	 */
+	public void setListenerTaskExecutor(AsyncListenableTaskExecutor listenerTaskExecutor) {
+		this.listenerTaskExecutor = listenerTaskExecutor;
```
artembilan: Missed propagation of this one to the `instance`.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -291,12 +302,70 @@ public void setErrorHandler(ErrorHandler errorHandler) {
...
+	public void setConsumerTaskExecutor(Executor consumerTaskExecutor) {
+		this.consumerTaskExecutor = consumerTaskExecutor;
+	}
+
+	protected AsyncListenableTaskExecutor getListenerTaskExecutor() {
+		return this.listenerTaskExecutor;
+	}
+
+	/**
+	 * Set the executor for threads that poll the consmer.
```
artembilan: Typo cons**u**mer.
Although need to determine if we should swap JavaDocs for these executors....

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -259,6 +271,10 @@ protected void doStop() {
 				container.stop();
 			}
 			this.containers.clear();
+			if (this.usingDefaultListenerExecutor) {
+				setListenerTaskExecutor(null); // in case the concurrency changes before the next start
+				this.usingDefaultListenerExecutor = false;
```
artembilan: Shouldn't we `ThreadPoolTaskExecutor.shutdown()` in case of `if (this.usingDefaultListenerExecutor)` ?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -276,13 +286,21 @@ else if (messageListener instanceof MessageListener) {
 			throw new IllegalStateException("messageListener must be 'MessageListener' "
 					+ "or 'AcknowledgingMessageListener', not " + messageListener.getClass().getName());
 		}
-		if (getTaskExecutor() == null) {
-			setTaskExecutor(
-					new SimpleAsyncTaskExecutor(getBeanName() == null ? "kafka-" : (getBeanName() + "-kafka-")));
+		if (getConsumerTaskExecutor() == null) {
+			SimpleAsyncTaskExecutor pooledExecutor = new SimpleAsyncTaskExecutor(
```
artembilan: The `SimpleAsyncTaskExecutor` isn't  `pooled`:
```
* <p><b>NOTE: This implementation does not reuse threads!</b> Consider a
* thread-pooling TaskExecutor implementation instead, in particular for
* executing a large number of short-lived tasks.
```
Maybe we really should reconsider in favor of `ThreadPoolTaskExecutor` ?

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -112,10 +115,18 @@
...
 	private volatile boolean running = false;
 
-	private Executor taskExecutor;
+	private Executor consumerTaskExecutor;
+
+	private AsyncListenableTaskExecutor listenerTaskExecutor;
 
 	private ErrorHandler errorHandler = new LoggingErrorHandler();
 
+	private long pauseAfter = DEFAULT_PAUSE_AFTER;
```
mbogoevici: I am wondering if we shouldn't calculate this based on the `heartbeat.interval.ms` property of the consumer (possibly lower than that), otherwise if the value is set too high (e.g. the 10s interval set by default is higher than the default `heartbeat.interval.ms` of 3s) this will not be effective.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: It shouldn't be a big deal since we're not starting/stopping consumers regularly - the listener executor is indeed pooled for this very reason.

I can change it if you insist.

artembilan: No, that's fine. Just change its variable name to avoid confuse. :smile:

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -444,85 +549,156 @@ public void run() {
...
+					}
+					else {
+						this.logger.error("Listener threw an exception and no error handler for " + record, e);
+					}
+				}
+			}
+			catch (TimeoutException e) {
+				if (this.pauseWhenSlow) {
+					this.shouldPause = true;
+					future.addCallback(new ListenableFutureCallback<Void>() {
```
artembilan: I don't understand how does it works.
We are here only if `future.get(this.pauseAfter, TimeUnit.MILLISECONDS);` throws `TimeoutException`, therefore failure. But I really see in debug that we step into `onSuccess()`!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -444,85 +549,156 @@ public void run() {
...
-								if (ListenerConsumer.this.ackMode.equals(AckMode.MANUAL_IMMEDIATE)) {
-									ListenerConsumer.this.consumer.commitAsync(commits,
-											ListenerConsumer.this.commitCallback);
+						@Override
+						public void onFailure(Throwable ex) {
+							if (ListenerConsumer.this.logger.isDebugEnabled()) {
+								ListenerConsumer.this.logger.debug(
+										"Delayed failure for " + record + ": " + ex.getMessage());
+							}
+							if (ListenerConsumer.this.pauseWhenSlow
```
artembilan: `ListenerConsumer.this.pauseWhenSlow` is redundant here. We are within the `if (this.pauseWhenSlow) {` for the `final` property.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: Ah! Got it!
This is the main trick how to `resume` from the slow listener.
Since we don't `cancel` the `Future` we still can get the result and we do that via this callback, meanwhile the main loop dance with the void heartbeats.
Doh...

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Change it to what?

artembilan: s/pooledExecutor/consumerTaskExecutor
sounds better for me. And with that name I'd even don't pay attention that it isn't `ThreadPoolTaskExecutor` :smile:

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -380,40 +438,87 @@ public boolean isLongLived() {
...
-						while (iterator.hasNext()) {
-							final ConsumerRecord<K, V> record = iterator.next();
+						if (!this.paused) {
+							currentRecords = records;
+							currentRecordsIterator = records != null ? records.iterator() : this.emptyIterator;
+						}
+					}
+					if (this.paused && !this.pausedForException && !this.listenerHasReturned) {
+						this.logger.trace("Paused consumer exited poll and listener has still not returned");
+						continue;
```
mbogoevici: Is there a circumstance under which this would cause subsequent poll() operations, overwriting `records`? E.g. if the consumer is paused, https://github.com/spring-projects/spring-kafka/pull/91/files/d005af1d3a6d268beabdbfefbace9c1747367938#diff-3dc07fec508b7a158b2ed0ba4b48e37cR454 is true, https://github.com/spring-projects/spring-kafka/pull/91/files/d005af1d3a6d268beabdbfefbace9c1747367938#diff-3dc07fec508b7a158b2ed0ba4b48e37cR455 gets executed repeatedly ?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -444,85 +549,162 @@ public void run() {
...
+				}
+			}
+			catch (TimeoutException e) {
+				if (this.pauseEnabled) {
+					/*
+					 * Getting the result from the consumer timed out, we will pause the consumer to
+					 * keep heartbeats going to the broker.
+					 */
+					this.shouldPause = true;
+				}
```
artembilan: Yeah... I think I'm fine with such a polishing.
Was going to ask about `getErrorHandler().handle()` in case of `this.pauseEnabled == false` :smile: 

I don't see any problem today yet.
Will come back tomorrow after all other polishing.

Meanwhile @mbogoevici may find anything else to discuss.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Not sure I understand - the poll won't return any records if the consumer if paused, we only continue the loop if we're paused and we don't need to resubmit the same record.

     *       ... Future calls to {@link #poll(long)} will not return
     * any records from these partitions until they have been resumed...

However, I just realized that the logic is flawed in that we wait for up to `pauseAfter` ms for __each record__ when in fact we need the reply for the whole batch within that time.

Polishing...

mbogoevici: Ah, right. Should have thought about it before asking - this is the heartbeat 'poll'.

mbogoevici: On that note, would it perhaps help making things simpler if we used a queue for storing fetched batches and we did an 'invokeListener' on the batch itself, pausing when the queue reaches max capacity?

garyrussell: I have just reworked it to pass the batch instead of individual records and it's much simpler.

Still using a Future for now, but could easily convert to a queue so we always have work "in hand" for the listener to process rather than polling only when the listener handles the current batch - good idea.

garyrussell:
Reworked as discussed - quite a bit simpler now - the exception use case is now handled by a suitably configured retry template (examples in tests).

Still needs docs; I suggest you read the commit comment on the last commit.

mbogoevici:
Great - this looks simpler indeed. I'll take a more thorough look tonight or tomorrow morning.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -443,89 +480,148 @@ public void run() {
...
-								}
-							}
-							else {
-								throw new IllegalStateException(
-										"With " + ListenerConsumer.this.ackMode.name()
-										+ " ack mode, acknowledge() must be invoked on the consumer thread");
-							}
-						}
+		private ConsumerRecords<K, V> checkPause(ConsumerRecords<K, V> unsent) {
+			if (this.paused && this.recordsToProcess.size() < this.queueDepth) {
```
martindam: Shouldn't you check if the unsent ConsumerRecords all fit in the queue? If not, the sendToListener could fail again and you loss the unsent records. Also, since you resumed it already, next poll would return new records you have to deal with. I would wait to resume until the unsent ConsumerRecords are all sent to the listener

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -318,9 +332,30 @@ protected void doStop() {
...
-		private Thread consumerThread;
+		private final long pauseAfter = getPauseAfter();
+
+		private final boolean pauseEnabled = isPauseEnabled();
+
+		private final BlockingQueue<ConsumerRecords<K, V>> recordsToProcess = new LinkedBlockingQueue<>(getQueueDepth());
+
+		private final int queueDepth = getQueueDepth();
+
+		private final BlockingQueue<ConsumerRecord<K, V>> acks = new LinkedBlockingQueue<>();
```
martindam: Shouldn't this be cleared when a rebalance happens? The application is not longer responsible for the partitions and should not be processing messages from it. If it isn't part of the standard flow here, the option to clear un-released messages should be exposed to external rebalance listeners

mbogoevici: Probably not this, since it's the acknowledgment queue for messages already processed, but we might want to do something about fetched messages waiting to process. I think that we do will be highly dependent on the acknowledgment mode. I think that the safest strategy here is to flush/process the remaining acks on `onPartitionRevoked` and gracefully stop the consumer, using `onPartitionsAssigned()` as a trigger to resume with the new partition set and use onPartitionAssigned() as a trigger to start consuming - for consistency, we may want to do that even on a normal start. (If we do that, we will need to have a different lifecycle management for manual assignments, but that may be ok, since manual assignments are not involved in the rebalancing process at all).

garyrussell: I don't think we need to stop the consumer, but we should check in the listener invoker that we are still assigned the partition before invoking - i.e. any messages in the pipeline for a partition that has been revoked should be dropped (logged).

mbogoevici: I've thought about that, too - and it seems more efficient as it doesn't require the application to re-fetch already retrieved messages. But there are two reasons why I believe this is unsafe:  

a) `onPartitionsRevoked()` does not provide the list of revoked partitions, but the list of currently allocated partitions (as counterintuitive as it sounds) - see https://kafka.apache.org/090/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html#onPartitionsRevoked(java.util.Collection), so it is not safe to continue processing locally stored messages - as we don't know which of them are still our responsibility. If we ignored `onPartitionsRevoked()` and relied only on `onPartitionsAssigned()`, we had no guarantee in which order do clients process the `onPartitionsAssigned()` calls, so it is perfectly possible that a client that loses partitions is notified of that only after another client has already started fetching from the same partitions - so we need to work based on the contract of `onPartitionsRevoked()` : the API guarantees that all calls to `onPartitionsRevoked()` (where we can bring all clients to a stable state) will be handled before any of the clients receive a `onPartitionsAssigned()` (where we can resume from a stable state). 

b) Even if we were able to somehow deal with a) correctly I think that there is at least one scenario where simple filtering can result in race conditions: considering a client that starts with all partitions, and another client that joins temporarily and leaves. The initial client will start filtering out messages, but when the other client leaves, it will regain all partitions again. If it didn't get to filter out all the messages from the previously lost partitions (that it has queued), it will start processing the leftover messages (as they now match the updated partition list)  - at a pretty arbitrary point - at the risk of either duplicating the processing done by the other client, or dropping messages, or processing them in the wrong order. It's not a very typical scenario, but points out why filtering would be unsafe.

So I think that being more conservative in the case of rebalancing would pay off, even if we were to re-fetch data in some clients - I think we should treat rebalancing as an infrequent event, and  allow it to be more expensive.  What we could do is to try to give a consumer that loses partitions a chance to complete its work, before returning from `onPartitionsRevoked()`.

martindam: Regarding b): A rebalance should restart all the affected partitions at the last committed offset (and I believe that is what the KafkaConsumer does internally) so keeping any data around after the partitions have been revoked would not be correct. You would actually get duplicates as your consumer would seek back to the latest committed offset and then refetch the messages you now have queued up. 

That actually brings up another issue, allowing the set the start offset during a rebalance from an external rebalance listener. This would be for the case where offsets are managed outside kafka and the application needs to set the offsets from another offset store (e.g. mysql). 

Also, a rebalance should be fairly snappy in my opinion. Consider the case of deploying a new version of your software. If you deploy it using a rolling update, you would trigger a lot of rebalances and to allow the system to make progress while updating, the rebalance would need to be fast.

mbogoevici: One note of what I meant by 'expensive' :) - I was referring to the cost of re-fetching the messages from partitions that remain assigned to clients, not necessarily to the work during rebalance taking a long time. Still, regardless of that, we must ensure that the client state is stable (i.e. all messages have been cleared out, either by processing them or by simply dumping them) and all the pending acks for messages that have already been processed have been committed. Anyway, I agree that it should be possible for transition process to be as fast as possible.

garyrussell: So it seems that on revoke, we should suspend the listener thread, commit any uncommitted acks, wait for the new assignment, restart the listener, which should then skip any queued messages that are for partitions that are no longer assigned to us and process any that are.

mbogoevici: My suggestion was more drastic - that we should drop queued messages and start fetching again from committed offsets in `onPartitionsAssigned()`.

garyrussell: Even simpler.

mbogoevici: I believe this is also @martindam 's suggestion, among other things :)

artembilan:
+1 for dropping in-flight queued messages.
Since we are in Kafka we don't lose data and just need to know from where to start fetch again- from commited offset :smile:

mbogoevici:
Cool. As discussed with @garyrussell I'll to sketch out an update for wiping out the in-flight messages as well as start/stop.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -239,6 +256,123 @@ public void setAutoStartup(boolean autoStartup) {
...
+	 * default the error handler is invoked.
+	 * @param recoveryCallback the recovery callback.
+	 */
+	public void setRecoveryCallback(RecoveryCallback<Void> recoveryCallback) {
+		this.recoveryCallback = recoveryCallback;
+	}
+
+
+	protected int getQueueDepth() {
+		return this.queueDepth > 0 ? this.queueDepth : 1;
```
artembilan: You say in the next JavaDocs that it is `Default 2`, but return `1`.
Is that intentional or omission ?

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -239,6 +256,123 @@ public void setAutoStartup(boolean autoStartup) {
...
+	 * Set a retry template to retry deliveries.
+	 * @param retryTemplate the retry template.
+	 */
+	public void setRetryTemplate(RetryTemplate retryTemplate) {
+		this.retryTemplate = retryTemplate;
+	}
+
+	protected RecoveryCallback<Void> getRecoveryCallback() {
+		return this.recoveryCallback != null ? this.recoveryCallback :
+			new RecoveryCallback<Void>() {
```
artembilan: Looks like this instance can be done only once on the property directly.
I know @mbogoevici prefers private specific classes on the matter :smile:

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -239,6 +256,123 @@ public void setAutoStartup(boolean autoStartup) {
...
+	protected int getQueueDepth() {
+		return this.queueDepth > 0 ? this.queueDepth : 1;
+	}
+
+	/**
+	 * Set the queue depth for handoffs from the consumer thread to the listener thread.
+	 * Default 2.
+	 * @param queueDepth the queue depth.
+	 */
+	public void setQueueDepth(int queueDepth) {
```
artembilan: Shouldn't we propagate `queueDepth` from the top the same way like we do, for example, for `recoveryCallback` etc. ?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -443,89 +480,148 @@ public void run() {
...
-						@Override
-						public String toString() {
-							return "Acknowledgment for " + record;
-						}
+		private boolean sendToListener(final ConsumerRecords<K, V> records) throws InterruptedException {
+			if (this.pauseEnabled) {
+				return !this.recordsToProcess.offer(records, this.pauseAfter, TimeUnit.MILLISECONDS);
+			}
+			else {
+				this.recordsToProcess.put(records);
```
artembilan: According `put()` JavaDocs:
```
* Inserts the specified element into this queue, waiting if necessary
* for space to become available.
```
So, if `pauseEnabled == false` we are blocked here in case of slow listener and therefore lose our `Consumer` because there is no heartbeat poll in the `run()` above.
Isn't it ?

garyrussell: That's the intent - a user might want to cause a rebalance if the consumer is slow, hence set the property to false.

artembilan: Sold.
Thank you!

mbogoevici:
Added a few additional changes (see commit message for details:) https://github.com/mbogoevici/spring-kafka/commit/53f070063b5fa25e69f85b0e704e20cbe92ba277

Still requires tests to prove that everything is handled correctly on rebalance. An additional possible change would be to do away with `assignedPartitions` and use `consumer.assginments()` directly instead (but that is minor)

garyrussell:
Added @mbogoevici 's commits - during a pairing session this morning, we decided to run the listner on the consumer thread if auto commit is enabled - this is to avoid premature commits.

However, it means that we can't avoid a repartition with slow consumer with auto commit enabled.

Bottom line is auto-commit is to be avoided if there's any chance your listener is slow.

mbogoevici:
> Bottom line is auto-commit is to be avoided if there's any chance your listener is slow.

I think that is ok as well if you're in a mode where you don't care much about partition affinity.

One thing I noticed with the new code is that the duration of the tests seems to have increased, probably due to waiting on `stop()` (which unfortunately is necessary in order to comply with the contract of `stop()`). Thinking on this, I could probably refactor this slightly and use a ListenableFuture and a runnable passed on `doStop()`. The behaviour inside `onPartitionsRevoked` might still need to be synchronous though, as we are supposed to return only after the consumer is stable (i.e. has finished work and flushed all transient stuff).

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -239,6 +260,145 @@ public void setAutoStartup(boolean autoStartup) {
...
+	 * Set the executor for threads that invoke the listener. If the instance set is not a
+	 * {@link AsyncTaskExecutor} it will be wrapped into one.
+	 * @param listenerTaskExecutor the executor.
+	 */
+	public void setListenerTaskExecutor(Executor listenerTaskExecutor) {
+		if (listenerTaskExecutor instanceof AsyncTaskExecutor) {
+			this.listenerTaskExecutor = (AsyncTaskExecutor) listenerTaskExecutor;
+		}
+		else {
+			this.listenerTaskExecutor = new ConcurrentTaskExecutor(listenerTaskExecutor);
```
artembilan: I don't see reason to do all this adaptation. If we really need `AsyncTaskExecutor` let's just request it from the configuration level!
All Spring target `Executor` implementations are `AsyncTaskExecutor`.
Otherwise it looks weird. 
You request from me `Executor`, but when I request `getListenerTaskExecutor()` I receive `AsyncTaskExecutor`.
If your bizarre `Executor` (e.g. from Reactor) isn't like `AsyncTaskExecutor` you always can wrap it into `ConcurrentTaskExecutor` on your own.
But I don't see reason to do that in our code when we can be more strict.

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
```diff
@@ -60,6 +65,14 @@
 
 	private DeDuplicationStrategy<K, V> deDuplicationStrategy;
 
+	private Boolean pauseEnabled;
+
+	private Long pauseAfter;
```
artembilan: I'd say that something like `pauseAfter > 0` is like and indicator from the top that `pause` is `Enabled`.
Just looks strange to require to configure both options.
Although we may configure it to `true`, when there is an explicit call to `pauseAfter`.
WDYT?

BTW is there some validation that `pauseAfter > 0`?
Or don't we need it ?

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -239,6 +260,145 @@ public void setAutoStartup(boolean autoStartup) {
...
+		this.recoveryCallback = recoveryCallback;
+	}
+
+	protected int getQueueDepth() {
+		return this.queueDepth > 0 ? this.queueDepth : 1;
+	}
+
+	/**
+	 * Set the queue depth for handoffs from the consumer thread to the listener thread.
+	 * Default 2.
```
artembilan: It is still `2` and `getQueueDepth()` still returns `1` for default.
What am I missing ?
Thanks

mbogoevici: I do believe this needs to be fixed yet :)

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
mbogoevici: Do we even use this value here? We have another one, with a default, below: https://github.com/spring-projects/spring-kafka/pull/91/files/fa9fd7286cd54c888774b65fcbef5eab5f92e493#diff-8ebdce1dc0deea04032bdfadbcd7bb4aR130

Wondering if we should externalize all configuration properties and pass the configured instance from the `Concurrent` class to the simple one. Every time we add property we need to copy it.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -318,37 +354,100 @@ protected void doStop() {
...
+		private final boolean isAnyManualAck = this.isManualAck || this.isManualImmediateAck;
+
+		private final boolean isRecordAck = this.ackMode.equals(AckMode.RECORD);
+
 		private final boolean syncCommits = KafkaMessageListenerContainer.this.syncCommits;
 
-		private Thread consumerThread;
+		private final long pauseAfter = getPauseAfter();
+
+		private final boolean pauseEnabled = isPauseEnabled();
```
artembilan: This is out of use already.
I won't mind either way, but method invocation style feels better for me...

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
artembilan: +1 for `ContainerProperties` abstraction

artembilan:
See the Travis report for some warn around generics and serialized state.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -607,14 +815,89 @@ private void commitIfNecessary() {
...
+
+		private final class ListenerInvoker implements SchedulingAwareRunnable {
+
+			private volatile boolean active = true;
+
+			@Override
+			public void run() {
+				while (this.active) {
+					try {
+						ConsumerRecords<K, V> records = ListenerConsumer.this.recordsToProcess.poll(1, TimeUnit.SECONDS);
```
artembilan: When we call `stopInvokerAndCommitManualAcks()` we may be blocked here maximum for 1 sec if there is no records before the next iteration to pick up `!this.active` state. 
That is a note for the increased tests execution time.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -443,89 +559,175 @@ public void run() {
...
+				this.logger.error("Invoker timed out while shutting down and will be canceled.");
+				this.listenerInvokerFuture.cancel(true);
+			}
+			finally {
+				this.listenerInvokerFuture = null;
+			}
+			// handle the last manual acks, after the listeners have closed
+			handleManualAcks();
+			if (this.offsets.size() > 0) {
+				// we always commit after stoping the invoker
```
artembilan: Typo: stop**p**ing

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
mbogoevici: Yeah, I think we should just interrupt the invoker's thread during stop() (as the invoker is discardable and is always associated with one and only one thread). Also, we can skip processing the records returned if the invoker is not active anymore.

artembilan: Yes, I also thought about `this.listenerInvokerFuture.cancel(true)`. Isn't it ?

artembilan: ... Instead of `this.listenerInvokerFuture.get(getShutdownTimeout(), TimeUnit.MILLISECONDS);`, of course.

mbogoevici: Yes, that was what I was thinking as well.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
```diff
@@ -169,23 +182,32 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
 		assertThat(latch.await(60, TimeUnit.SECONDS)).isTrue();
 		assertThat(rebalancePartitionsAssignedLatch.await(60, TimeUnit.SECONDS)).isTrue();
 		assertThat(rebalancePartitionsRevokedLatch.await(60, TimeUnit.SECONDS)).isTrue();
+		assertThat(listenerThreadNames).allMatch(new Predicate<String>() {
```
artembilan: We can use Java 8 language level in test! :smile:

mbogoevici: Ah, nice :). Will remember that.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
```diff
@@ -87,17 +90,18 @@
...
 	@Test
 	public void testAutoCommit() throws Exception {
-		logger.info("Start auto");
+		this.logger.info("Start auto");
 		Map<String, Object> props = KafkaTestUtils.consumerProps("test1", "true", embeddedKafka);
 		DefaultKafkaConsumerFactory<Integer, String> cf = new DefaultKafkaConsumerFactory<Integer, String>(props);
 		ConcurrentMessageListenerContainer<Integer, String> container =
 				new ConcurrentMessageListenerContainer<>(cf, topic1);
 		final CountDownLatch latch = new CountDownLatch(4);
+		final ConcurrentSkipListSet<String> listenerThreadNames = new ConcurrentSkipListSet<>();
```
artembilan: `ConcurrentSkipListSet` just for `CountDownLatch(4)`? :smile: 
Minor of course, but really it is curious how that was selected.
Thanks

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
artembilan: See our `build.gradle`:
```
	compileJava {
		sourceCompatibility = 1.7
		targetCompatibility = 1.7
	}

	compileTestJava {
		sourceCompatibility = 1.8
		targetCompatibility = 1.8
	}
```

Actually my IDEA suggests me to convert it into Lambda anyway :wink:

artembilan:
For the record, if you are still interested in the test performance:
```
ConcurrentMessageListenerContainerTests 	9 	0 	1 	41.503s 	100%
KafkaMessageListenerContainerTests 	        4 	0 	0 	58.543s 	100%
```
But that my, bloody hell, Windows if that! :smile:

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
mbogoevici: A `synchronizedSet` would be enough, I think. We don't need the sorting. Still needs to be thread-safe because it's modified by two separate threads.

artembilan: Well, according to your assertion that can be just simple (unsychronized) `ArrayList`.
OK. That's just nit-pick and fully isn't important in favor of all that hard work you did here.
Never mind

mbogoevici: That is a fair point.

mbogoevici:
Added a few more changes (see commit message) https://github.com/mbogoevici/spring-kafka/commit/b870a5c413862743d88a2ce8a7e92173c57c104f

garyrussell:
Pushed - sorry guys got tied up with something else. 

@mbogoevici I've given you push rights if you need to push anything more directly here. I'll probably not pick this up until Tuesday now - have a good holiday weekend all.

artembilan:
OK. Excellent!

I think from big height I'm fine with the solution.
Although I still don't see the full picture. The algorithm is very complicated. And maybe that would much much easier for me to catch up it eventually when we will have particular issues already :wink: .

I'll give it yet a couple review rounds from the next week, but now I give up.
Maybe need to pick the paper and pencil to draw this state machine :smile: 

Really have a great weekend!

mbogoevici:
Thanks! I'll be sending a few more cleanup updates until Tuesday, but I believe this is mostly done for now. 
Have a nice weekend as well!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: We should always use the boolean - I propagated these properties to the LC to avoid referencing the outer class on every access - in other words for performance reasons. I am ok and completely concur with using a properties container, though. But, it should be a hard final reference in the consumer, again to avoid accessing the container instance:

    isPauseEnabled();

    LINENUMBER 616 L0
    ALOAD 0
    GETFIELD org/springframework/kafka/listener/KafkaMessageListenerContainer$ListenerConsumer.this$0 : Lorg/springframework/kafka/listener/KafkaMessageListenerContainer;
    INVOKEVIRTUAL org/springframework/kafka/listener/KafkaMessageListenerContainer.isPauseEnabled ()Z
    IFEQ L1

    this.pauseEnabled:

    LINENUMBER 616 L0
    ALOAD 0
    GETFIELD org/springframework/kafka/listener/KafkaMessageListenerContainer$ListenerConsumer.pauseEnabled : Z
    IFEQ L1

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -561,18 +564,21 @@ public void run() {
 
 		private void stopInvokerAndCommitManualAcks() {
 			this.invoker.stop();
+			this.listenerInvokerFuture.cancel(true);
```
mbogoevici: Thinking this through again - `cancel` doesn't feel right, as it deprives us from the option to wait for normal completion of the invoker runnable - the runnable will stop, but will do this in the background, potentially processing some messages before it stops and keeping to send acks.

I believe that the correct thing to do (albeit complicated) is to interrupt the thread that executes the invoker directly during stop, wait for normal completion on `get()` - the most likely outcome - and cancel if we are dealing with a non-cooperative listener (whose acks we can discard).

mbogoevici:
Pushed

garyrussell:
Pushed docs for `ContainerProperties`.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/31541fe58e8b1265728301106027b250657aaeb2

Issue #87, opened 2016-05-17
Consider Adding autoCommitOnError


garyrussell:
See: https://github.com/spring-cloud/spring-cloud-stream/issues/542

Currently, the new container always commits regardless of whether an exception is caught.

PR #97, opened 2016-05-27
[Review only] GH-87 Add support for `ackOnError`

mbogoevici:
This PR will need to be merged (and probably slightly adjusted) after #91 as it modifies the way in which commits are processed (and the algorithm would beed to be slightly adapted to match the asynchronous model introduced by #91, esp. wrt. how batches are committed).

But we can still validated the approach.


mbogoevici:
Closed in favour of https://github.com/spring-projects/spring-kafka/pull/107

PR #98, opened 2016-05-27
ErrorHandler method is not set

jmirc:
When a KafkaMessageListenerContainer is created, the error handler method is not called. There is no way to set the error handler after it is created


artembilan:
I think that is good catch.
Thank you for contribution!

Would you mind to confirm that you have signed already some Spring projection [Contribution Agreement](https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc), e.g. Spring Boot?

And please, prove the issue with some test-case, e.g. in the `ConcurrentMessageListenerContainerTests`.

jmirc:
I signed the Contribution Agreement and I updated a test-case to prove it works.

artembilan:
Great! Would be glad to see your name in author list of changed classes and will merge tomorrow. Otherwise I can do that on merge, of course.

jmirc:
Thanks. When do you think you will do the next release? I am migrating our internal client we built based on 0.8 to spring-kafka and I would like to use the new version quickly. I added new feature like resilience that needs this bug.

artembilan:
The next week it will be RC1. Without any problems it will be GA over the week after.

garyrussell:
Merged as 903ef266220a31a07c9f43cbcf0f343be7855b3d

Thanks for the contribution!!

PR #102, opened 2016-06-01
GH-99 Idle Container Events

### LINKED ISSUE #99: How can I know when no records were consumed by Consumer.poll()?
Currently, when `Consumer.poll()` times out and no records were consumed, I have no way to know that.

See: https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L394

The `MessageListener.onMessage()` is called for each record consumed. What I would like is to also be notified when no records were consumed, like `MessageListener.onNoRecordsConsumed()`.

Why do I want that? My use case is a follows: if `Consumer.poll()` times out, it means that whatever was in the topic was consumed (the topic has a finite number of records), therefore I want to stop the consumer.

garyrussell:
Resolves #99

Also change `ContainerProperties` to use accessors

Also fix `stop(Runnable callback` logic for the container registry and
concurrent container.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/event/ListenerContainerIdleEvent.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/event/ListenerContainerIdleEvent.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.event;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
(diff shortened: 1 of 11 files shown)


spring-kafka/src/main/java/org/springframework/kafka/event/ListenerContainerIdleEvent.java
```diff
@@ -0,0 +1,81 @@
...
+	public long getIdleTime() {
+		return this.idleTime;
+	}
+
+	/**
+	 * The queues the container is listening to.
+	 * @return the queue names.
+	 */
+	public String[] getTopicPartitions() {
+		return this.topicPartitions.toArray(new String[this.topicPartitions.size()]);
```
artembilan: Can't we require exactly this value as a ctor arg? 
To avoid `TopicPartition` API dependency.
Not sure that it will work everywhere, because, I think we may end up with the ClassNotFoundException anyway on the `source`, if the event listener doesn't have appropriate classes in its CP.

garyrussell: Oops - copy/paste bug - should return `TopicPartition[]`.

I'm not sure what you mean about CNF - unless they ship the event to a different JVM.

The `source` is transient.

Maybe we can convert the topic/partitions to a map?

artembilan: Yeah... We even can live with `TopicPartition` because I might be just super scary with the `ApplicationEvent` purpose and its public API independence...

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/eff22d0922dc19116c96fbbeda6017a6c234833b

Issue #89, opened 2016-05-18
how can i set offset by myself


dragontree101:
now i find `recent-offset` can set offset from end message
if i want to consumer message from offset=100 but i don't know what is the lastest offset, so i can't use  `recent-offset`
but i doesn't find any way to set offset=100 use spring-kafka by myself.
how can i set offset by myself? thanks

garyrussell:
It's not currently supported - we can consider adding an option but the problem is you don't really know what the broker has retained. We thought the recent offset would be useful to get, say, the last message in a topic after starting, even if you've previously received it. We didn't envision a need for positioning `n` messages from the beginning of whatever the broker has retained.

Can you explain your use case, and we'll consider adding it.

[Contributions are always welcome too](https://github.com/spring-projects/spring-kafka/blob/master/CONTRIBUTING.adoc).

dragontree101:
we have this case: 
every day we have produce some order messages(related to money)
and next day we want to re consumption these order messages to reconciliation
we will recode offset to db at the begin of day, and a crontab to re consumption these order messages to reconciliation

PR #103, opened 2016-06-01
GH-89: Add support for `seek` from the beginning

artembilan:
Fixes GH-89 (https://github.com/spring-projects/spring-kafka/issues/89)

* Rename `recentOffset` just to `offset`
* Add logic for the `offset` property when it is *positive* seek the consumer from the offset beginning;
when the `offset` is negative, seek from the end

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
@@ -53,8 +53,8 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 
 	/**
 	 * Construct an instance with the supplied configuration properties and specific
-	 * topics/partitions - when using this constructor, {@link ContainerProperties#setRecentOffset(long)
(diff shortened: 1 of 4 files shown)


artembilan:
Rebased

garyrussell:
Hmmm - as I mentioned in skype yesterday, we talked about reinstating the OffsetManager - we need a way for a specific offset for each topic/partition to be set - and the values may be different for each.

garyrussell:
The recent offset is a different use case - e.g, re-send the last message I processed.

artembilan:
OK-OK. I understand that I'm wrong.
Sorry, I need more info to implement proper solution.
But which one?
Point me, please, what to read to figure out what and where should be implemented.
Thanks

garyrussell:
Something like this: https://github.com/spring-projects/spring-integration-kafka/blob/1.3.x/src/main/java/org/springframework/integration/kafka/listener/OffsetManager.java#L59

Where the user populates the initial offset for each topic/partition. Then, during startup, we iterate and seek.

artembilan:
Good. Than it is a subject of #9 and #89 should be closed in favor of that.

Closing this PR as irrelevant.

PR #104, opened 2016-06-01
GH-101: Rename DeDup Classes/Interfaces to Filter

### LINKED ISSUE #101: Question: Why DeDuplicationStrategy class is so specific?
I developed a filtering feature on top of Kafka using flatbuffer. The filter uses headers to filter or not the message. The goal is to deserialize the payload only when the consumer accepts the message. 
I finished yesterday to migrate my old spring-kafka library based on the version 0.8 to use this project and it works.

**My question is:**
A `DeDuplicationStrategy` has been created to filter the message is the message is a dupe. This strategy is great for the dedupe use case. However, this feature could be more generic to support many use cases.
My use case could fit perfectly in this strategy, however my use case is not a dedupe is a filter. So I decided to not use the `DeDuplicationStrategy` interface to handle this use case.

I would like to know if the `DeDuplicationStrategy` could be replaced by a generic interface that could support different strategies (dedupe, filtering or another use case) to not send the message to the listener.

I am hopping I am clear.

Thanks for your great work.

garyrussell:
Resolves #101


artembilan:
I think I'm fine with the solution.

@mbogoevici ?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringAcknowledgingMessageListenerAdapter.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @author Gary Russell
+ *
+ */
+public class FilteringAcknowledgingMessageListenerAdapter<K, V> extends AbstractFilteringMessageListener<K, V>
+		implements AcknowledgingMessageListener<K, V> {
+
+	private final AcknowledgingMessageListener<K, V> delegate;
+
+	private final boolean ackDiscarded;
```
artembilan: Shouldn't we add this option into the `MessagingMessageListenerAdapter` as well?

src/reference/asciidoc/kafka.adoc
```diff
@@ -339,15 +339,15 @@ Receiver] pattern and Spring Integration provides an
 http://docs.spring.io/spring-integration/reference/html/messaging-endpoints-chapter.html#idempotent-receiver
 [implementation thereof].
```
artembilan: Not related to this PR, but the divided link part brake it.
So the part in the `[]` should be moved to the same row as url.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringAcknowledgingMessageListenerAdapter.java
garyrussell: Yeah - I was hoping to avoid that 😄

mbogoevici: Trying to imagine a use case where not acknowledging a discarded message would be useful.

src/reference/asciidoc/kafka.adoc
```diff
@@ -339,15 +339,15 @@ Receiver] pattern and Spring Integration provides an
...
 
-The Spring for Apache Kafka project also provides some assistance by means of the `DeDuplicatingMessageListenerAdapter`
+The Spring for Apache Kafka project also provides some assistance by means of the `FilteringMessageListenerAdapter`
 classe, which can wrap your `MessageListener`.
```
artembilan: typo: class~~e~~

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringAcknowledgingMessageListenerAdapter.java
garyrussell: They might be acking manually on some condition (special marker record perhaps and not want random acks being done on their behalf.

I guess I am just a bit uncomfortable issuing automatic acks when they have specifically requested to use manual acks.

mbogoevici: Ah, I see. An option gives a choice then. Should we at least make this `true` by default - otherwise, in the case of manual acks, they would almost never have a chance to ack those messages and there's no chance that they would ack them in the future on replay - so unless there's a message that they're interested in this would keep redelivering messages on each restart.

garyrussell:
Pushed - @artembilan - it's a bit simpler now, if the container listener factory is configured - we wrap the MMLA with a filtering adapter.

garyrussell:
The trouble with making it true by default is it really can only be used of the AckMode is set to one of the MANUAL options.

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
```diff
@@ -226,7 +244,21 @@ public void setupListenerContainer(MessageListenerContainer listenerContainer, M
...
+				aml = new FilteringAcknowledgingMessageListenerAdapter<>(this.recordFilterStrategy, aml, this.ackDiscarded);
+				container.setupMessageListener(aml);
+			}
+			else {
+				messageListener = new FilteringMessageListenerAdapter<>(this.recordFilterStrategy, messageListener);
+				container.setupMessageListener(messageListener);
+			}
+		}
+		else {
+			container.setupMessageListener(messageListener);
```
artembilan: We can still have this line of code without duplication in each `if...else`.
Minor, of course.
Will polish on merge.

src/reference/asciidoc/kafka.adoc
```diff
@@ -329,25 +329,25 @@ public void listen(@Payload String foo,
...
 function.
 This is known as the http://www.enterpriseintegrationpatterns.com/patterns/messaging/IdempotentReceiver.html[Idempotent
 Receiver] pattern and Spring Integration provides an
-http://docs.spring.io/spring-integration/reference/html/messaging-endpoints-chapter.html#idempotent-receiver
-[implementation thereof].
+http://docs.spring.io/spring-integration/reference/html/messaging-endpoints-chapter.html#idempotent-receiver[implementation thereof].
 
-The Spring for Apache Kafka project also provides some assistance by means of the `DeDuplicatingMessageListenerAdapter`
+The Spring for Apache Kafka project also provides some assistance by means of the `FilteringMessageListenerAdapter`
 classe, which can wrap your `MessageListener`.
```
artembilan: The typo hasn't been fixed. Strange that my comment on the matter had gone.
NP. Will fix on merge.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/b1dc76e3a974f09b52c962f90351999eb5880af9

PR #107, opened 2016-06-02
Add `ackOnError` and fix some ack commit issues

### LINKED ISSUE #87: Consider Adding autoCommitOnError

mbogoevici:
Fixes #87

- allow for acking on both success and error (make the latter conditional upon `ackOnError`)
- fixes an issue where `processCommits()` wasn't handled on invoker stop, allowing for missed commits

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -470,6 +470,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 			}
 			// handle the last manual acks, after the listeners have closed
 			handleManualAcks();
+			processCommits();
 			if (this.offsets.size() > 0) {
 				// we always commit after stopping the invoker
 				commitIfNecessary();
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -470,6 +470,7 @@ private void stopInvokerAndCommitManualAcks() {
 			}
 			// handle the last manual acks, after the listeners have closed
 			handleManualAcks();
+			processCommits();
```
artembilan: Since we are adding this one here we don't need the next `commitIfNecessary()` with its `if`.
Minor, of course, because after `processCommits()` we will never reach that  `commitIfNecessary()`: `this.offsets` is empty already.

artembilan: OK. Never mind. With the `AckMode.RECORD` `processCommits()` doesn't perform `commitIfNecessary()`.
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/be8b438fdd4bfbabb1ca5551d5458b8af0519bab

PR #111, opened 2016-06-02
GH-89: Add support for `seek` from the beginning

artembilan:
Fixes GH-89 (https://github.com/spring-projects/spring-kafka/issues/89)

* Introduce `TopicPartitionInitialOffset`, where it utilize `TopicPartition` and `Long initialOffset`
* The `initialOffset` can be:
  - `null` - do nothing; rely on the `auto.offset.reset` of `Consumer` config;
  - positive (including `0`) - absolute offset from the begging of partition;
  - negative - the offset from the end of partition: `consumer.seekToEnd() - initialOffset`
* Rework everything around to rely on a new `TopicPartitionInitialOffset` abstraction
* The logic in the `KafkaMessageListenerContainer.ListenerConsumer.initPartitionsIfNeeded()` reworked to in favor of a new abstraction
* remove redundant `recentOffset`
* Reflect new `TopicPartitionInitialOffset` in the docs

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -452,12 +452,54 @@ public class KafkaListenerAnnotationBeanPostProcessor<K, V>
 				"topic in @TopicPartition must resolve to a String, not " + topic.getClass());
 		Assert.state(StringUtils.hasText((String) topic), "topic in @TopicPartition must not be empty");
 		String[] partitions = topicPartition.partitions();
-		Assert.state(partitions.length > 0,
-				"At least one partition required in @TopicPartition for topic '" + topic + "'");
+		PartitionOffset[] partitionOffsets = topicPartition.partitionOffsets();
+		Assert.state(partitions.length > 0 || partitionOffsets.length > 0,
+				"At least one 'partition' or 'partitionOffset' required in @TopicPartition for topic '" + topic + "'");
 		List<TopicPartitionInitialOffset> result = new ArrayList<>();
-		if (partitions.length > 0) {
-			for (int i = 0; i < partitions.length; i++) {
-				resolvePartitionAsInteger((String) topic, resolveExpression(partitions[i]), result);
+		for (int i = 0; i < partitions.length; i++) {
+			resolvePartitionAsInteger((String) topic, resolveExpression(partitions[i]), result);
+		}
+
+		for (PartitionOffset partitionOffset : partitionOffsets) {
+			Object partitionValue = resolveExpression(partitionOffset.partition());
+			Integer partition;
+			if (partitionValue instanceof String) {
+				Assert.state(StringUtils.hasText((String) partitionValue),
+						"partition in @PartitionOffset for topic '" + topic + "' cannot be empty");
+				partition = Integer.valueOf((String) partitionValue);
+			}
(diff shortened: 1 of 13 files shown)


artembilan:
If you are fine with the solution, I'd introduce for the `@TopicPartition` alongside with the `partitions` something like:
```
PartitionOffset[] partitionOffsets() default {};
```
annotation.

garyrussell:
Cool - works well; tested with a "real" application (SI Sample) (positive and negative)...

    		ContainerProperties properties = new ContainerProperties(
                                new TopicPartitionInitialOffset(this.topic, 0, 490L),
				new TopicPartitionInitialOffset(this.topic, 1, -3L));


Yes, I was going to suggest we add it to the annotation.

It would be very cool too if the evaluation context for the offset included `#topic` and `#partition` variables 😄

artembilan:
Well, we have there this one:
```
return this.resolver.evaluate(resolvedValue, this.expressionContext);
```
where `BeanExpressionContext expressionContext`, so any beans in the ctx are available via `@` operator. Therefore any complex `initialOffset` logic can be resolved somewhere there.

Or do you have something different in your mind ?

artembilan:
In my idea that looks like:
````java
public @interface TopicPartition {
        
    String topic();

	String[] partitions() default {};

	PartitionOffset[] partitionOffsets() default {};

}
````
where you can specify `partitions()` and `partitionOffsets()`. The last one has precedence though, if the same partition defined.
And that one is like:
````java
@Target({})
public @interface PartitionOffset {

	String partition();
	
	String initialOffset() default "";
	
}
````
So, not sure how your `"@bean.offsetFor(#topic, #partition)"` may be useful. 
Seems for me some much complexity...
Can't we just provide `ContainerProperties` into the `ConcurrentKafkaListenerContainerFactory` and populate the same `TopicPartitionInitialOffset[] topicPartitions` instead of that SpEL hell for annotations?

garyrussell:
>Can't we just...

No, because the factory is for multiple listeners; the topics/partitions are defined at the listener level.

>So, not sure how your `"@bean.offsetFor(#topic, #partition)"` may be useful. 

Well, a SpEL expression for an offset would need some context `@bean.getOffset()` is useless without knowing the topic/partition.

I suppose

    topic = "foo", partitionOffset = 
        @PartitionOffset(partition = "2", initialOffset = "@bean.getOffset('foo', 2)")

would be ok, too, if you think it's not worth the effort to add them as variables; I just thought it would be nicer to provide them as variables during the evaluation rather than having to repeat.

I guess we could consider that for a future release.

artembilan:
I see your point. Yes. But we are missing that in this annotation context we are only in the `BeanFactory` `EvaluationContext`. So we can do only something like `#{bean.getOffset(...)}`.
And we can't customize it (`BeanExpressionContext`) with variable for the current invocation...

I think we can do some `StandardEvaluationContext` manipulation to meat your requirements, but shouldn't than that sound like `initialOffsetExpression` as an alternative?

garyrussell:
I see what you mean; we'd have to add more SpEL infrastructure - let's defer that for now; after all they can do something like

    topic = "#{@foo.topics[0]}, partition = "#{@foo.parts[0]}, offset = "#{@foo.offs[0]}"

instead.

artembilan:
Pushed `@PartitionOffset` support.

garyrussell:
Polishing here https://github.com/garyrussell/spring-kafka/commit/1d54ec97363ad2377d5bf1a459819df4a541baf1 for your review.

garyrussell:
Merged as 06616f73eda4ea7f341c3d43cf211d6f4eb54211

PR #122, opened 2016-06-16
Fix race condition in the `KafkaMessageLCTests`

artembilan:
Having the async consumption nature we have to `spy` the `Consumer` before the real consumption.
Otherwise we can end up with the fact that we pass all desired phases in one thread, but another hasn't been spied yet to be verified in the end properly.

https://build.spring.io/browse/SK-SON-97/

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
@@ -19,6 +19,7 @@ package org.springframework.kafka.listener;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.atMost;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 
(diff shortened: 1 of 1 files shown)


mbogoevici:
LGTM, merging.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -237,7 +243,7 @@ public void testSlowConsumerCommitsAreProcessed() throws Exception {
 		// Verify that commitSync is called when paused
 		assertThat(latch.await(60, TimeUnit.SECONDS)).isTrue();
 		verify(consumer, atLeastOnce()).pause(any(TopicPartition.class), any(TopicPartition.class));
-		verify(consumer, atLeastOnce()).commitSync(any());
+		verify(consumer, atMost(2)).commitSync(any());
```
mbogoevici: Missed this on the first pass, unfortunately. Why are we doing this, as it completely changes the semantic of the test? It would pass if no commits are made, and why is there an upper bound of the number of commit operations?

artembilan: Doh... Good catch @mbogoevici !

This code has confused me:
```
if (foundSize > maxNumberOfInvocations) {
        new Reporter().wantedAtMostX(maxNumberOfInvocations, foundSize);
}
```

I haven't looks into the logic closely and somehow my English knowledges let me down :cry: 

I definitely meant `atLeast(2)`.

The previous `atLeastOnce()` isn't good for us because we always call it once in the `onPartitionsAssigned()` anyway.

Fee free to fix it in your branch! No need separate PR or even commits, like Boot guys requie :wink: 
I will understand you properly.

Issue #124, opened 2016-06-17
Consider providing some default values for `ackTime` and `ackCount`


mbogoevici:
Currently they're both set to 0, which essentially renders the default behaviour of `TIME`, `COUNT`,`COUNT_TIME` and `MANUAL` identical to `BATCH` (and `MANUAL_IMMEDIATE_SYNC` in the case of `MANUAL`).

It would be better if we provided some defaults, or at the very least, we validated the values (e.g. COUNT requires a non-zero count value, etc).

artembilan:
Any wisdom which *magic number* should we choose?

mbogoevici:
For time, we could, in principle, use the default settings of autocommit - the default for `auto.commit.interval.ms` is 5s.

For count, I'm not sure - this is very dependent on the nature of the application. Maybe we should just validate that a positive number is set in that case.

artembilan:
:+1: for both ideas.
We can reconsider in the future on demand.

PR #128, opened 2016-06-23
GH-124: Validate `ackCount` and `ackTime`

artembilan:
Fixes GH-124 (https://github.com/spring-projects/spring-kafka/issues/124)

That doesn't look reasonable to have void `commitIfNecessary()` if we exceed default `0` for `ackCount` or `ackTime`.

* Add requirement to have `ackCount` and `ackTime` `> 0` in case appropriate `ackMode`
* Make `ackTime` as 5 secs by default - similar to default for `auto.commit.interval.ms`

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -140,8 +140,19 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 		if (isRunning()) {
 			return;
 		}
-		setRunning(true);
-		Object messageListener = getContainerProperties().getMessageListener();
+		ContainerProperties containerProperties = getContainerProperties();
+
+		if (!this.consumerFactory.isAutoCommit()) {
+			AckMode ackMode = containerProperties.getAckMode();
+			if (ackMode.equals(AckMode.BATCH) || ackMode.equals(AckMode.COUNT) || ackMode.equals(AckMode.COUNT_TIME)) {
+				Assert.state(containerProperties.getAckCount() > 0, "'ackCount' must be > 0");
+			}
+			if (ackMode.equals(AckMode.TIME) || ackMode.equals(AckMode.COUNT_TIME)) {
+				Assert.state(containerProperties.getAckTime() > 0, "'ackTime' must be > 0");
+			}
+		}
+
+		Object messageListener = containerProperties.getMessageListener();
 		Assert.state(messageListener != null, "A MessageListener is required");
 		if (messageListener instanceof AcknowledgingMessageListener) {
 			this.acknowledgingMessageListener = (AcknowledgingMessageListener<K, V>) messageListener;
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -140,8 +140,19 @@ protected void doStart() {
...
 			return;
 		}
-		setRunning(true);
-		Object messageListener = getContainerProperties().getMessageListener();
+		ContainerProperties containerProperties = getContainerProperties();
+
+		if (!this.consumerFactory.isAutoCommit()) {
+			AckMode ackMode = containerProperties.getAckMode();
+			if (ackMode.equals(AckMode.BATCH) || ackMode.equals(AckMode.COUNT) || ackMode.equals(AckMode.COUNT_TIME)) {
+				Assert.state(containerProperties.getAckCount() > 0, "'ackCount' must be > 0");
```
mbogoevici: I don't think we should check this for BATCH - the contract is that it commits after the batch has been processed, regardless of count or time.

mbogoevici: The other question is what should we do for MANUAL ? It is in theory the same as COUNT_TIME, but using manual acks instead of container-managed.  Maybe we can just document that MANUAL is identical with MANUAL_IMMEDIATE_SYNC if no count or time are set.

artembilan: Well, I think I should study more about those `MANUAL`, but I don't understand your concern about `BATCH`.
Honestly it is used only from the same place as `COUNT` : 
```
if (ackMode.equals(AckMode.BATCH) || ackMode.equals(AckMode.COUNT) && countExceeded) {
	if (this.logger.isDebugEnabled()) {
		this.logger.debug("Committing in AckMode.COUNT because count " + this.count
					+ " exceeds configured limit of " + this.containerProperties.getAckCount());
	}
	commitIfNecessary();
	this.count = 0;
}
```
And, yes, `COUNT` isn't used anymore, too!

Don't we miss anything here?

mbogoevici: We should use some brackets to disambiguate ... `&&` has precedence over `||` so if ack mode is batch, it is always true, regardless of count.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -153,20 +164,21 @@ else if (messageListener instanceof MessageListener) {
...
 					(getBeanName() == null ? "" : getBeanName()) + "-kafka-listener-");
-			getContainerProperties().setListenerTaskExecutor(listenerExecutor);
+			containerProperties.setListenerTaskExecutor(listenerExecutor);
 		}
 		this.listenerConsumer = new ListenerConsumer(this.listener, this.acknowledgingMessageListener);
-		this.listenerConsumerFuture = getContainerProperties()
+		this.listenerConsumerFuture = containerProperties
 				.getConsumerTaskExecutor()
 				.submitListenable(this.listenerConsumer);
+		setRunning(true);
```
mbogoevici: We should place this before submitting the tasks, otherwise there's a race condition that makes it possible for the listener consumer to exit immediately.

mbogoevici:
LGTM.

mbogoevici:
Squashed and merged as https://github.com/spring-projects/spring-kafka/commit/3483387eb032c273b2084b2ba227ecc08f7c84c4

PR #131, opened 2016-06-30
GH-118: Rework Manual AckMode

### LINKED ISSUE #118: How MANUAL_IMMEDIATE[_SYNC] really works?
Hi,
As i understand this, manual commits are executed via `KafkaMessageListenerContainer.handleManualAcks()` that is called in `run()` in the same class after polling the records, but only in case any acks exist already.
But, after the first poll, no ack's in the queue exist as the listeners weren't invoked yet.
If there is only **one poll** with data (as happens in tests often), how exactly manual acks are sent?
Also, if i understand correctly, MANUAL_IMMEDIATE AckMode, doesn't guarantee that the ack really happened when the listener method called `acknowledge()`, right?

Thanks!

garyrussell:
Fixes #118

- Remove MANUAL_IMMEDIATE_SYNC - sync/async is controlled by the `syncCommits` property
- For MANUAL, wake the consumer thread at the end of the batch
- For MANUAL_IMMEDIATE, the consumer thread is woken directly from the Acknowledgement

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -241,12 +241,13 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 		private final boolean isManualAck = this.containerProperties.getAckMode().equals(AckMode.MANUAL);
 
 		private final boolean isManualImmediateAck =
-				this.containerProperties.getAckMode().equals(AckMode.MANUAL_IMMEDIATE)
-				|| this.containerProperties.getAckMode().equals(AckMode.MANUAL_IMMEDIATE_SYNC);
+				this.containerProperties.getAckMode().equals(AckMode.MANUAL_IMMEDIATE);
 
 		private final boolean isAnyManualAck = this.isManualAck || this.isManualImmediateAck;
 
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -625,8 +630,9 @@ private void processCommits() {
 					updatePendingOffsets();
 				}
 				boolean countExceeded = this.count >= this.containerProperties.getAckCount();
-				if (ackMode.equals(AckMode.BATCH) || (ackMode.equals(AckMode.COUNT) && countExceeded)) {
-					if (this.logger.isDebugEnabled()) {
+				if (this.isManualAck || ackMode.equals(AckMode.BATCH)
```
artembilan: Minor. We can already use that new `isBatchAck` here instead.

artembilan:
Gary, here is a polishing for your confirmation: https://github.com/artembilan/spring-kafka/commit/a89bcf3fe5ba55c3fd50fdbb1a16ae53614e1921

garyrussell:
Polish LGTM; let's see if Travis agrees 😄

artembilan:
Well, it's green already.
I guess I can merge.
Let me know your opinion.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/51af726693e698e23ea5af89c5e81fe04cd30cb3.

If there is something wrong, we always can push subsequent fix.

PR #136, opened 2016-07-05
GH-135: Fix Acks

### LINKED ISSUE #135: Kafka offsets not commited
Hi, we were trying to use new spring-kafka 1.0.0.RELEASE 'RECORD' strategy for committing offsets with ConcurrentKafkaListenerContainerFactory configuration:
- concurrency=1
- pollTimeout=10 000

In our scenario we encountered strange situation when offsets were not commited by ListenerConsumer.

It happened that ListenerInvoker thread processed the Kafka messages so quickly that all he was doing was to wake up ListenerConsumer thread (which was blocked on poll() method) every 1sec. Please note that there was no other messages on Kafka topic to consume so ListenerConsumer  will be blocked on poll() invocation for the entire 10seconds.

We think that the problem is because pollTimeout (ListenerConsumer ) is greater then 1 sec (ListenerInvoker) and due to that we were able to produce a live-lock on ListenerConsumer.

Please comment.

garyrussell:
Resolves #135

Properly transfer `ack`s from the listener thread to the consumer thread
for all ack modes.

Previously, record ack mode was not handled in `processCommits`.

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
@@ -24,10 +24,12 @@ import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
+import java.util.Arrays;
 import java.util.BitSet;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -436,6 +443,61 @@ public void onMessage(ConsumerRecord<Integer, String> message) {
...
+		});
+		containerProps.setSyncCommits(true);
+		containerProps.setAckMode(AckMode.RECORD);
+		containerProps.setAckOnError(false);
+		KafkaMessageListenerContainer<Integer, String> container = new KafkaMessageListenerContainer<>(cf,
+				containerProps);
+		container.setBeanName("testAckOnError");
+		container.start();
+		ContainerTestUtils.waitForAssignment(container, embeddedKafka.getPartitionsPerTopic());
+		KafkaConsumer<?, ?> containerConsumer = spy(
```
artembilan: This `spy()` and subsequent `DFA` can be replaced by just one `spyOnConsumer()`.
And it really would be better to make any stubbing before `waitForAssignment()`.
Otherwise there is no guaranty that all the environment is proper for further verifications.

Will fix on merge.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/677d135932047e6d453a3897d4ff29804bd04c13
and cherry-picked into `1.0.x`.

Issue #134, opened 2016-07-03
how to close temple when programmer is stop?


dragontree101:
i use template to send message, but i don't find any way to close template.
i found test case, use template is `template.send()` and `template.flush`
if I use `kafkaProducer`, I will `kafkaProducer.close` and print log `Closing the Kafka producer with timeoutMillis = 9223372036854775807 ms.` but when use template I am not see any close message.

Would template will close kafkaProducer auto?

stepio:
Hi @dragontree101 ,

To initialise `KafkaTemplate`, you have to specify `ProducerFactory`. And `DefaultKafkaProducerFactory` implements `DisposableBean`. So you may store the reference to your `DefaultKafkaProducerFactory` call `destroy()` when necessary:
```java
	@Override
	public void destroy() throws Exception { //NOSONAR
		CloseSafeProducer<K, V> producer = this.producer;
		this.producer = null;
		if (producer != null) {
			producer.delegate.close();
		}
	}
```
Or Spring will call it upon destroying the appropriate bean.

stepio:
@artembilan 

Is there any benefit in using Spring's `DisposableBean` in comparison to `AutoClosable` from Java 7 ?
Tried "googling", but have not found a good answer.

dragontree101:
thanks @stepio
before i have create DefaultKafkaProducerFactory this not as a bean 
after i add ```@Bean```  annotation i see the log print already

and seems if a bean implements ```Closeable``` interface, and spring will auto call ```close()``` method?

dragontree101:
and i haven another question.
i want to use ```@Scheduled``` to ```flush()``` producer, but if a producer has not ```send()``` and it's ```producer``` is null, and throw ```NullPointerException```
so i am not flush template by ```@Scheduled```?

stepio:
> and seems if a bean implements Closeable interface, and spring will auto call close() method?

Yes, Spring documentation states so:
http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html
> To disable destroy method inference for a particular @Bean, specify an empty string as the value, e.g. @Bean(destroyMethod=""). Note that the DisposableBean and the Closeable/AutoCloseable interfaces will nevertheless get detected and the corresponding destroy/close method invoked.

stepio:
> i want to use @Scheduled to flush() producer, but if a producer has not send() and it's producer is null, and throw NullPointerException
> so i am not flush template by @Scheduled?

[...]

PR #137, opened 2016-07-05
GH-134: Ignore flush() if 'producer' is not inititalized yet

stepio:
Fix for the issue:
https://github.com/spring-projects/spring-kafka/issues/134

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -52,7 +52,7 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V> {
 
 	private MessageConverter messageConverter = new MessagingMessageConverter();
 
-private final boolean autoFlush;
+	private final boolean autoFlush;
 	private volatile Producer<K, V> producer;
 
 	private volatile String defaultTopic;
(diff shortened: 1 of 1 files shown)


artembilan:
Yeah... `master` again?

Please, make a top-level branch for the issue, e.g. `GH-134` and commit and push from there.
Plus consider some simple test-case for better coverage.
And add your name to the `@author` list finally! :smile: 

Thank you for so quick turnaround!

stepio:
> Yeah... master again?

I know, I know - created a branch, but forgot to checkout into it - understood this only after pushing :)

I'm a bit lost regarding the change:
1. `IllegalStateException` - already discussed, but "disliked" by @garyrussell 
2. Just ignoring - I also thought this to be a good option originally + it's simpler (to need to compose a human-readable exception message).
3. Assert... Yeah, maybe, why not? But will they work if assertions are disabled?

> Plus consider some simple test-case for better coverage.

Sure, agree

artembilan:
You have to just add:
```
Assert.state(this.producer != null, "'producer' must not be null for flushing.")
```
before `this.producer.flush();`
Exactly that `state` throws `IllegalStateException`.
That was my point.

I don't see anything good in the promotion of the wrong way to use our API, if we just ignore the `null` fact and let to build undesired state machine in the target application.

Your arguments to ignore the `NPE`?

stepio:
I don't see any business case when this swallowing may be necessary and your reasoning is good enough, so Ok for me

PR #138, opened 2016-07-05
GH-134: Ignore flush() if 'producer' is not inititalized yet

stepio:
Reworked based on comments from https://github.com/spring-projects/spring-kafka/pull/137

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
@@ -195,4 +195,14 @@ public class KafkaTemplateTests {
 		pf.createProducer().close();
 	}
 
+	@Test(expected = IllegalStateException.class)
+	public void flushWithoutSend() throws Exception {
+		Map<String, Object> senderProps = KafkaTestUtils.producerProps(embeddedKafka);
+		DefaultKafkaProducerFactory<String, String> pf = new DefaultKafkaProducerFactory<String, String>(senderProps);
+		pf.setKeySerializer(new StringSerializer());
+		KafkaTemplate<String, String> template = new KafkaTemplate<>(pf);
+		template.setDefaultTopic(STRING_KEY_TOPIC);
+		template.flush();
+		throw new AssertionError("IllegalStateException must be thrown upon flushing");
+	}
 }
--- a/spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
@@ -17,10 +17,12 @@
 package org.springframework.kafka.core;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.springframework.kafka.test.assertj.KafkaConditions.key;
 import static org.springframework.kafka.test.assertj.KafkaConditions.partition;
 import static org.springframework.kafka.test.assertj.KafkaConditions.value;
 
+import java.util.Collections;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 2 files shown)


stepio:
@artembilan 

Did the proposed changes. Thanks for the hint with `Assert.state` - I haven't used it before, thought about throwing the exception "manually". Knowledge sharing is the best reason for contributing :)

spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
```diff
@@ -195,4 +195,14 @@ public void testTemplateDisambiguation() throws Exception {
...
 
+	@Test(expected = IllegalStateException.class)
+	public void flushWithoutSend() throws Exception {
+		Map<String, Object> senderProps = KafkaTestUtils.producerProps(embeddedKafka);
+		DefaultKafkaProducerFactory<String, String> pf = new DefaultKafkaProducerFactory<String, String>(senderProps);
+		pf.setKeySerializer(new StringSerializer());
+		KafkaTemplate<String, String> template = new KafkaTemplate<>(pf);
+		template.setDefaultTopic(STRING_KEY_TOPIC);
+		template.flush();
+		throw new AssertionError("IllegalStateException must be thrown upon flushing");
```
artembilan: Great! Merging with some code style polishing and your name for the affected classes.

Regarding this test-case some nit-picks. 
We don't need so much options for the `KafkaTemplate` since there is just enough to call its `flush()` and don't let any one after us to override the code without issues.
On the other hand we stopped with Gary rely on the `@Test(expected =` because there might be more places where such an exception can be thrown. That's why our pattern is like:

```
try {
      invocation_to_verify;
      fail("Some_Exception expceted");
}
catch {Exception e} {
      assertThat(e, instanceOf(Some_Exception.class));
      assertThat(e.getMessage(), containsString("Some part of the exception message to verify"));
}
```
Will fix on merge and share in this PR for future considerations.

stepio: Hmm... Good idea about exception's fragment.
I also prefer the approach with `instanceOf` - had some floating issues when `expected` did not work as expected. However I've chosen it here as it's "junit-native" - thought it's more suitable for the framework. But was not sure about checking the whole exception message - did not want to force updating the tests in case you decide to do some reorganisation in texts. But the point about identifying the exact needed exception is good.

And sorry for missing `@author` update, most probably it's really time to sleep for me :) Have a good day!

artembilan: You got it!
Have a good night and come back to us with more feedback any time :smile:

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/6a94ae2d7993c13aebc852f40b4bcf74f71f6462 and cherry-picked to `1.0.x`

spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
```diff
@@ -195,4 +197,12 @@ public void testTemplateDisambiguation() throws Exception {
...
 	}
 
+	@Test
+	public void flushWithoutSend() throws Exception {
+		DefaultKafkaProducerFactory<String, String> pf = new DefaultKafkaProducerFactory<String, String>(Collections.emptyMap());
+		KafkaTemplate<String, String> template = new KafkaTemplate<String, String>(pf);
+		assertThatExceptionOfType(IllegalStateException.class)
+				.isThrownBy(() -> template.flush())
+				.withMessageContaining("'producer' must not be null for flushing.");
+	}
```
artembilan: Wow! That's much better than my variant :smile: 

Accepting it even I merged already.

garyrussell:
> Knowledge sharing is the best reason for contributing :)

> Wow! That's much better than my variant

We all learn by collaboration - thanks @stepio

PR #139, opened 2016-07-06
GH-135: Fix BATCH Acks

abbadon123:
Hi,

In reference to #135 
Described scenario still exists for ack mode = BATCH

When:
- pollTimeout = 10s
- ack mode = BATCH

Then:
- records processed by MessageListener are not commited

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -448,6 +448,9 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 					}
 				}
 				catch (WakeupException e) {
+					if (!this.autoCommit) {
+						processCommits();
+					}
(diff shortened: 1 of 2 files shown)


artembilan:
Hello, @abbadon123 !

First of all thank you for the contribution.
Second: we don't accept any PRs without CLA signing confirmation: https://github.com/spring-projects/spring-integration/blob/master/CONTRIBUTING.adoc
Third: it would be better to start from GH issue to describe and confirm with the test-case.

The problem that I have copied/pasted your test-case from the PR into the current `master` and it passes. 

But on the other hand pay attention to the Travis report. I don't say that your fix is guilty, but it would be better to check locally on your side.

Thank you for your patience.

artembilan:
See #140

artembilan:
Closed in favor of #145 

@abbadon123 , we can't accept your PR because of lack of feedback.
Since we would like to release something very soon, we ended up with a bit different solution.
But anyway we borrowed your test-case, although with some improvement to confirm the problem.

Nevertheless thank you for contribution!
Definitely looking forward for more!

Issue #140, opened 2016-07-07
Kafka offsets are not commited when Container pollTimeout >> ListenerInvoker recordsPollTimeout


adamdec:
Hi,
In reference to #135 

Problem for live-lock still exists for version 1.0.1.RELEASE
I have tested it using your test case "testBatchAck" with ContainerProperties:
```
        containerProps.setSyncCommits(true);
        containerProps.setAckMode(AckMode.BATCH);
        containerProps.setPollTimeout(10000);
        containerProps.setAckOnError(false);
```
As abbadon123 already suggested the simple fix would be to add:
```
    catch (WakeupException e) {
		if (!this.autoCommit) {
		   processCommits();
		}
  		this.unsent = checkPause(this.unsent);		  			
    }
```
Regards

artembilan:
Thank you @adamdec , for having a look into this!

I'm appreciate all your feedback and let me describe the problem as I see it to steal your a bit more.

When the `this.consumer.poll(this.containerProperties.getPollTimeout())` is longer than listener's batch process and, therefore, there is no more records to proceed normally withing the main loop, we always end up in the `catch (WakeupException e)` after
```
if (this.isManualAck || this.isBatchAck) {
	this.consumer.wakeup();
}
```
initiation in the `invokeListener` process.
Since that  `catch (WakeupException e)` doesn't have `processCommits()` operation, we commit `BATCH` only in the container's `stop()` procedure. Or... after `pollTimeout`, of course. And that's why, I guess, @abbadon123 's `testBatchAck()`  test passes.

Correct me if I have missed something.

artembilan:
Well, after some hacking I'd say that it would be better to leave the main loop for `poll` as is and remove extra code from `ListenerInvoker`:
```
if (!ListenerConsumer.this.isManualImmediateAck && this.active) {
	ListenerConsumer.this.consumer.wakeup();
}
```
If we don't have records to process we don't have any chance for consumer to wait for them in the topic and just wake up it for nothing. As you noticed correctly each second:
```
ConsumerRecords<K, V> records = ListenerConsumer.this.recordsToProcess.poll(1,
									TimeUnit.SECONDS);
```

I don't see reason to break the poll process so often.

Although I won't mind if we move:
```
if (!this.autoCommit) {
	processCommits();
}
```
before the `this.consumer.poll()` task in the main loop just in case really to avoid so long time to wait after that poll.

[...]

PR #145, opened 2016-07-11
GH-140: Commit in the beginning of the main loop

artembilan:
Fixes GH-140 (https://github.com/spring-projects/spring-kafka/issues/140)

Previously with the pretty big `containerProperties.getPollTimeout()` we ended up with the issue of not acked commits for `BATCH` mode.
Just because the logic relies on the `consumer.wakeup()` which causes `WakeupException` breaking the main polling loop.

* Move `processCommits()` function to the beginning of main loop, before blocking on the `this.consumer.poll()`

**Cherry-pick to 1.0.x**

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
@@ -37,6 +37,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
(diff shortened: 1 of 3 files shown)


garyrussell:
s/begging/beginning/

Will fix on merge.

garyrussell:
travis failure

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -494,6 +497,62 @@ public void testRecordAck() throws Exception {
...
+				if (entry.getValue().offset() == 2) {
+					latch.countDown();
+				}
+			}
+			return invocation.callRealMethod();
+
+		}).given(containerConsumer)
+				.commitSync(any());
+
+		assertThat(latch.await(60, TimeUnit.SECONDS)).isTrue();
```
garyrussell: I think we need to verify the `commit()` is called within 10 seconds of the first message received.

garyrussell:
Merged as d04b53a7ab354f52dc4ab976832c1d0439199adc cherry-picked to 1.0.x

PR #147, opened 2016-07-12
GH-146: Fix TopicPartition Negative Reset

### LINKED ISSUE #146: Relative Initial Offset Needs Adjustment
`new TopicPartitionInitialOffset(this.topic, 0, -1000L)` results in

`java.lang.IllegalArgumentException: seek offset must not be a negative number`

Need to adjust seek to `Math.max(0, this.consumer.position(topicPartition) + offset)`.

garyrussell:
Fixes #146

When using reset relative to the current end, ensure the seek value is
not less than zero.

__cherry-pick to 1.0.x__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -691,12 +691,18 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 
 					if (offset < 0) {
 						this.consumer.seekToEnd(topicPartition);
-						newOffset = this.consumer.position(topicPartition) + offset;
+						newOffset = Math.max(0, this.consumer.position(topicPartition) + offset);
 					}
 
-					this.consumer.seek(topicPartition, newOffset);
-					if (this.logger.isDebugEnabled()) {
-						this.logger.debug("Reset " + topicPartition + " to offset " + newOffset);
+					try {
+						this.consumer.seek(topicPartition, newOffset);
+						if (this.logger.isDebugEnabled()) {
+							this.logger.debug("Reset " + topicPartition + " to offset " + newOffset);
+						}
+					}
+					catch (Exception e) {
+						logger.error("Failed to set initial offset for " + topicPartition
+								+ " at " + newOffset, e);
 					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -691,12 +691,18 @@ private void initPartitionsIfNeeded() {
...
-						this.logger.debug("Reset " + topicPartition + " to offset " + newOffset);
+					try {
+						this.consumer.seek(topicPartition, newOffset);
+						if (this.logger.isDebugEnabled()) {
+							this.logger.debug("Reset " + topicPartition + " to offset " + newOffset);
+						}
+					}
+					catch (Exception e) {
+						logger.error("Failed to set initial offset for " + topicPartition
+								+ " at " + newOffset, e);
```
artembilan: OK. Failed, logged. 
Really good catch. I'm definitely sure that there might be some cases when offset is changed in between `consumer.position()` and `consumer.seek()`.

But which one will be set then?
Shouldn't we mention the `consumer.position()` in this log, too?

garyrussell: I think it's ok to reposition from the results of the seek; the most likely is new records have been added - if someone else has done a seek with the same group-id - it's up to them,

I suppose we could add the position that to the log but we should capture it in a variable rather than calling it again.

Do you want to do during the merge, or do you want me to do it?

artembilan: Well, if you mean `this.consumer.position(topicPartition)` above as variable to use in this logs, I won't mind. 
Just it isn't clear to me by your comment. Sorry :cry:

artembilan: Looks like we can't extract variable so easy:
```
this.consumer.seekToEnd(topicPartition);
newOffset = Math.max(0, this.consumer.position(topicPartition) + offset);
```
That `position` definitely depends on the previous `seekToEnd`.
So, I think if we fail with consequent ``seek`, we should just call `this.consumer.position(topicPartition)` one more time in this `logger.error()` message.

Am I missing anything?

garyrussell: You are correct - ok; thanks; please make it so 😄

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/6d80e0e023aa86438d053881b33f61a254ed2bac and cherry-picked.

PR #149, opened 2016-07-12
GH-148: Initial Offsets Relative to Current

### LINKED ISSUE #148: Add Initial Offset Option Relative to the Current Committed Offset
Currently, a `TopicPartitionInitialOffset` with a negative initial offset means set the offset relative to the current end of the partition. A positive offset is an absolute offset.

Add a third option (new `boolean`?, subclass?) to indicate the initial offset (positive or negative) should be applied to the current committed offset (if any) for the consumer.

garyrussell:
Resolves #148

Provide an option to set the initial offset relative to the current position.

__no backport; will need rebase and/or fixup after #147 is merged__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/PartitionOffset.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/PartitionOffset.java
@@ -45,4 +45,13 @@ public @interface PartitionOffset {
 	 */
 	String initialOffset();
 
+	/**
+	 * By default, positive {@link #initialOffset()} is absolute, negative
+	 * is relative to the current topic end. When this is 'true', the
+	 * initial offset (positive or negative) is relative to the current
+	 * consumer position.
+	 * @return whether or not the offset is relative to the current position.
+	 */
+	String relativeToCurrent() default "false";
+
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/TopicPartitionInitialOffset.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/TopicPartitionInitialOffset.java
@@ -25,14 +25,19 @@ import org.apache.kafka.common.TopicPartition;
  * an initial offset for it. The initial offset can be:
  * <ul>
  * <li>{@code null} - do nothing;</li>
- * <li>positive (including {@code 0}) - seek to the absolute offset within the partition;
+ * <li>positive (including {@code 0}) - seek to EITHER the absolute offset within the
+ * partition or an offset relative to the current position for this consumer, depending
+ * on {@link #isRelativeToCurrent()}.
  * </li>
- * <li>negative - seek to the offset relative to the current last offset within the
- * partition: {@code consumer.seekToEnd() + initialOffset}.</li>
+ * <li>negative - seek to EITHER the offset relative to the current last offset within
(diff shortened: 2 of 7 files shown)


garyrussell:
However, since it's new optional behavior, I suppose it could go into 1.0.x. WDYT?

artembilan:
We can always backport on demand.
I guess more and more people wait now for Kafka 0.10 support from us :smile: .
So, we should release bug fixes and really start to concentrate on `1.1` work.

src/reference/asciidoc/kafka.adoc
```diff
@@ -169,7 +169,9 @@ public ContainerProperties(Pattern topicPattern)
 ----
 
 The first takes an array of `TopicPartitionInitialOffset` arguments to explicitly instruct the container which partitions to use
-(using the consumer `assign()` method), and with an optional initial offset: a positive value is an absolute offset; a negative value is relative to the current last offset within a partition.
+(using the consumer `assign()` method), and with an optional initial offset: a positive value is an absolute offset by default; a negative value is relative to the current last offset within a partition by default.
+A constructor for `TopicPartitionInitialOffset` is provided that takes an additional `boolean` argument.
+If this is `true`, the initial offets (positive or negative) are relative to the current position for this consumer.
```
artembilan: Typo: off**s**ets

spring-kafka/src/main/java/org/springframework/kafka/support/TopicPartitionInitialOffset.java
```diff
@@ -25,28 +25,64 @@
...
  * @author Artem Bilan
+ * @author Gary Russell
  */
 public class TopicPartitionInitialOffset {
 
 	private final TopicPartition topicPartition;
 
 	private final Long initialOffset;
 
+	private final boolean relativeToCurrent;
```
artembilan: Shouldn't we include this new option into `@PartitionOffset` as well ?

garyrussell: Yes, good point; ok; then that answers the backport question 😄 - I won't have time to do it today and we need to get 1.0.2 out.

artembilan: Yeah... That's good one argument. OK. Let's switch to your other PR, fix it merge and release :smile: 
Than I will have time to fight with `handleCancel`/`handleCancelOk` in SA :wink:

garyrussell:
Pushed, rebased.

spring-kafka/src/main/java/org/springframework/kafka/annotation/PartitionOffset.java
```diff
@@ -45,4 +45,13 @@
...
 	String initialOffset();
 
+	/**
+	 * By default, positive {@link #initialOffset()} is absolute, negative
+	 * is relative to the current topic end. When this is 'true', the
+	 * initial offset (positive or negative) is relative to the current
+	 * consumer position.
+	 * @return whether or not the offset is relative to the current position.
+	 */
+	String relativeToCurrent() default "false";
```
artembilan: `@since 1.1`
Will fix on merge.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/b2ec2b77efca3e3d9c6ed3b99e7db3fd5f58fa56

Issue #150, opened 2016-07-15
AdaptableMLA.onMessage(cr) Eats Exceptions; Should Throw to Container


garyrussell:
See https://github.com/spring-projects/spring-integration-kafka/issues/124

artembilan:
Also see https://jira.spring.io/browse/AMQP-623 and https://jira.spring.io/browse/SPR-14473

PR #151, opened 2016-07-15
GH-150: Don't eat errors in `AbstractAdaptableML`

artembilan:
Fixes: GH-150 (https://github.com/spring-projects/spring-kafka/issues/150)

In some corner cases the target `MessageListener` implementation may decide to invoke `AbstractAdaptableMessageListener.onMessage(message)` of its delegate.
In this case we can't receive any exceptions back into container for handling, because `handleListenerException()` just logs them.

* Deprecate `handleListenerException()` and don't wrap `onMessage(message, null)` call with `try...catch` to let exception bubble in container back,
like it is with regular `onMessage(message, null)` in the container.

(Consider to remove `handleListenerException()` altogether for current `1.1` `master`)

**Cherry-pick to 1.0.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AbstractAdaptableMessageListener.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AbstractAdaptableMessageListener.java
@@ -42,11 +42,10 @@ public abstract class AbstractAdaptableMessageListener<K, V> implements MessageL
 
 	protected final Log logger = LogFactory.getLog(getClass()); //NOSONAR
 
-
 	/**
 	 * Kafka {@link MessageListener} entry point.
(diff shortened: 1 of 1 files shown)


artembilan:
I have changed my mind a bit during the similar fix for AMQP (https://jira.spring.io/browse/AMQP-623).
But I won't mind if we remove that method altogether for Spring Kafka.

artembilan:
See the discussion with Juergen: https://jira.spring.io/browse/SPR-14473.
Looks like Spring Kafka doesn't need that `handleListenerException()` method.
Any exception should be thrown into container for handling. 
That is end-user decision to "eat" and handle listener exceptions on his own.

sobychacko:
LGTM - merging.

Also cherry-picked into 1.0.x branch.

PR #158, opened 2016-07-29
GH-157: Move ContainerTestUtils to s-k-test

### LINKED ISSUE #157: Move ContainerTestUtils to spring-kafka-test?
Would it be possible and wise to move the class ContainerTestUtils to the spring-kafka-test project? It appears, that the waitFor... methods could be useful in tests of projects, using spring-kafka. As the class is test-scoped in spring-kafka, it is not available int the using project though.

We were having some strange behavior with flakyness in our system-tests due to some assign / revoke condition. After a friendly hint from Artem Bilan, we managed to circumvent the issue, with code, close to the one in the aforementioned utils-class.

garyrussell:
Resolves #157 

__cherry-pick to 1.0.x__


artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/ad62b09732891797e02447e057bf6de7099b5d66
and cherry-picked.

PR #162, opened 2016-08-04
GH-161; Memory Leak with autoCommit

### LINKED ISSUE #161: The topic records shouldn't are added to "acks" queue when the "autoCommit" setted true in the class  KafkaMessageListenerContainer<K, V>
In the method run() of class ListenerConsumer of inner Class KafkaMessageListenerContainer<K, V>, it    invoke the method "invokeListener" ,which add the records  to "acks" queue when the "autoCommit" setted  true in the configuration.The size of "acks" queue will increase with the data increasement until beyond Integer.MAX_VALUE because the "acks" queue never poll when the "autoCommit" setted  true.
so ,I think the logic of handling message and putting the records to acks queue should separate into two part , and it should just invoke the logic of handling message  when  "autoCommit"  set true.
I don't know if I was clear.

garyrussell:
Fixes #161

We should not add to the `acks` collection when autoCommit is true.

__cherry-pick to 1.0.x__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -853,6 +853,9 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 			@Override
 			public void acknowledge() {
 				try {
+					if (ListenerConsumer.this.autoCommit) {
+						throw new IllegalStateException("Manual acks are not allowed when auto commit is used");
+					}
 					ListenerConsumer.this.acks.put(this.record);
 				}
 				catch (InterruptedException e) {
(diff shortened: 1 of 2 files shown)


artembilan:
It might not be related, but we have the code like this there as well:
```
public void acknowledge() {
    try {
		ListenerConsumer.this.acks.put(this.record);
    }
```
and looks like we still can simply use `AcknowledgingMessageListener` even if we are in `autoCommit == true` mode. And from here nothing stops end-user to call that `acknowledge()`.

Note: `@KafkaListener` provides such a listener unconditionally (`MethodKafkaListenerEndpoint`):
```
protected MessagingMessageListenerAdapter<K, V> createMessageListenerInstance() {
	return new MessagingMessageListenerAdapter<K, V>(this.method);
}
```

artembilan:
Otherwise fix looks good.

garyrussell:
Pushed

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/9e0e248ba54e91041028bdab6e97fb7c4303d519 and cherry-picked.

Issue #160, opened 2016-08-02
IllegalArgumentException: Payload must not be null


daroo:
Current implementation (1.0.2.RELEASE) of `KafkaListener` doesn't handle messages with null payload correctly. When a message with null payload comes in following exception is thrown:
```
java.lang.IllegalArgumentException: Payload must not be null
	at org.springframework.util.Assert.notNull(Assert.java:115)
	at org.springframework.messaging.support.MessageBuilder.createMessage(MessageBuilder.java:191)
	at org.springframework.kafka.support.converter.MessagingMessageConverter.toMessage(MessagingMessageConverter.java:78)
	at org.springframework.kafka.listener.adapter.MessagingMessageListenerAdapter.toMessagingMessage(MessagingMessageListenerAdapter.java:105)
	at org.springframework.kafka.listener.adapter.MessagingMessageListenerAdapter.onMessage(MessagingMessageListenerAdapter.java:97)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.invokeListener(KafkaMessageListenerContainer.java:597)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.access$1800(KafkaMessageListenerContainer.java:222)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer$ListenerInvoker.run(KafkaMessageListenerContainer.java:778)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:745)
```

This happens even if the payload is marked as a `@Payload(required = false)`. 

Currently the only way to handle such messages is to create and register own implementation of a message converter. 
However messages with null payload are completely valid in Kafka world (for example with a topic compaction enabled a message with a key and a null payload is treated as a delete from the log). Also `KafkaTemplate` support them correctly.

So, either remove `notNull` check for payload or use different converters depending on `@Payload(required = true/false)` flag... or at least provide a message converter implementation which accepts null payload.

garyrussell:
Thanks; this is a bug - we're all at SpringOne Platform this week.

[Contributions](https://github.com/spring-projects/spring-kafka/blob/master/CONTRIBUTING.adoc) are always welcome 😄

artembilan:
[...]

PR #163, opened 2016-08-04
GH-160: IllegalArgumentException: Payload must not be null

daroo:
Fixes GH-160 (https://github.com/spring-projects/spring-kafka/issues/160)


artembilan:
The idea of `KafkaMessage` is fine with me.
There is need to improve JavaDocs and maybe some other minor issues. We'll see.
At least right now we have something to discuss.
Of course a general idea would be better like `NullMessage` directly in the `spring-messaging` project to let all other get a gain. But as a quick solution to fix a bug it is fine right now. 

I'll try today to figure out some workaround because I don't like such a API stress in the minor version.
Anyway Spring Integration will fail with this approach because the same `MessageBuilder` is in use everywhere.

CC @garyrussell

garyrussell:
In SI (actually spring-messaging) 5.0, we will support `Optional<?>` payloads.

For this purpose, I would prefer a "special" payload (`KafkaNull`) or similar - because propagating a message with a `null` payload will cause problems with Spring Integration components.

garyrussell:
Obviously, at the `@KafkaListener` level, `KafkaNull` will need to be coerced to `null`.

artembilan:
@garyrussell , see my comment in the issue: https://github.com/spring-projects/spring-kafka/issues/160#issuecomment-237619616.

So, as a workaround we can consider to use empty String or empty `byte[]`.
And utilize such a logic exactly in the `MessagingMessageConverter.extractAndConvertValue()`.

Yes, that would be much better to deal with Java 8 `Optional<?>`, but that may be available only since SF-5.0. The same `PayloadArgumentResolver.resolveArgument()` can simply extract `Optional` on the matter.

Although IMO `NullMessage` from the SF-Messaging should be good as well. Just as a container to carry headers, like an HTTP `HEAD` method.

OTOH let's try `Optional<?>` on the `@KafkaListener` method instead!

garyrussell:
I think it must be a "special" value - an empty `String` or `byte[]` are valid payloads and might mean something other than `null` to an application.

daroo:
@artembilan, 
> OTOH let's try Optional<?> on the @KafkaListener method instead!

Together with `@Payload(required = true)` it could be a bit semantically inconsistent. Although, without the annotation (single arg listener method), `Optional<?>` support makes perfect sense.

@garyrussell 
> For this purpose, I would prefer a "special" payload (KafkaNull) or similar - because propagating a message with a null payload will cause problems with Spring Integration components.

But then Spring Integration would also need to know/support this new payload type (likewise the  `@KafkaListener`). Otherwise client's code will receive an unexpected payload type, right?

artembilan:
Forget an `Optional<?>` on `@KafkaListener`. At least until we don't support it from the Framework perspective.
Right now it doesn't work because we even can't overcome `MessagingMessageConverter` because of `MessageBuilder`. 

> an empty String or byte[] are valid payloads and might mean something other than null to an application.

Right. That's why I found that solution in the `PayloadArgumentResolver` as clumsy. Maybe `Optional.empty()` would be really good there as well. Will see. Need to discuss with SF team.

I see a workaround like:
````java
@Test
public void testNulls() throws Exception {
	kafkaJsonTemplate.send("annotated12", null, null);
	kafkaJsonTemplate.send("annotated12", null, "foo");
	assertThat(this.listener.latch8.await(60, TimeUnit.SECONDS)).isTrue();
}
...
private final CountDownLatch latch8 = new CountDownLatch(2);
...
@KafkaListener(id = "quux", topics = "annotated12", containerFactory = "nullPayloadListenerContainerFactory")
public void listen8(@Payload Optional<String> none) {
	this.latch8.countDown();
}
...
@Bean
public KafkaListenerContainerFactory<?> nullPayloadListenerContainerFactory() {
	ConcurrentKafkaListenerContainerFactory<Integer, String> factory =
	factory.setMessageConverter(new MessagingMessageConverter() {

		@Override
		protected Object extractAndConvertValue(ConsumerRecord<?, ?> record, Type type) {
			Object payload = super.extractAndConvertValue(record, type);
			return payload == null ? Optional.empty() : payload;
		}

	});
	return factory;
}
````

With that even Spring Integration will be happy.
Only the problem that we can't utilize it from Spring Kafka perspective yet, because the minimum is Java 7 still...

**Comment by garyrussell** [comment]
`Optional<?>` requires Java 8 so we can't use it yet. Right; that's why I am saying use a special value and coerce it to `null` when calling the `@KafkaListener`.

>Spring Integration would also need to know/support this new payload type 

SI is completely agnostic to payload type - it just can't be `null`.

From the application's perspective, if the converter emits a message with payload `KafkaNull`, it will become `null` when calling the `@KafkaListener` - which is what the application is expecting.

**Comment by artembilan** [comment]
OK. I think `KafkaNull` would be enough for now. At least until we figure out more general `null payload` solution from Spring Messaging perspective. 

@daroo , do you understand where to go from here?

I feel like there would be enough to add somehow simple `KafkaNullToObjectConverter` into `ConversionService` of the `DefaultMessageHandlerMethodFactory` in the `KafkaListenerAnnotationBeanPostProcessor.createDefaultMessageHandlerMethodFactory()`:
````java
KafkaNullToObjectConverter implement Converter<KafkaNull, Object> {

     public Object convert(KafkaNull source) {
            return null;
     }

}
````

**Comment by garyrussell** [comment]
One further issue is with a multi-method listener (class-level `@KafkaHandler` together with `@KafkaHandler` methods.

Since we use the post-conversion payload type to select the method, I don't see how we can support this use case (`null` payload) in that scenario - we won't be able to select the method to invoke.

**Comment by garyrussell** [comment]
Not saying we need a solution, but we should at least document it.

**Comment by artembilan** [comment]
>  we won't be able to select the method to invoke.

So, the `@KafkaHandler` has a limitation with its purpose if you expect nulls in topics.

But we may propose a fallback `@KafkaHandler` method if payload is null from the beginning.

**Comment by garyrussell** [comment]
Yes, maybe a no-payload handler method - actually, this whole scenario only makes sense if the listener has a `@Header(KafkaHeaders.RECEIVED_MESSAGE_KEY` argument - so the app knows which key to remove (was removed) when a `null` value is received.

**Comment by daroo** [comment]
> so the app knows which key to remove (was removed) when a null value is received.

and that's my case, but... Kafka allows also both key and payload to be null. I cannot figure out any useful use-case for such messages, so maybe you are right and it shouldn't be supported by spring-kafka.

**Comment by garyrussell** [comment]
I created a new issue #164 for the `@KafkaHandler` case - no need to address here.

File: spring-kafka/src/main/java/org/springframework/kafka/support/KafkaMessage.java [hunk 1]
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * An implementation of {@link Message} with a generic payload.
+ * Almost identical as {@link GenericMessage} but allows null payload.
+ *
+ * @param <T> the payload type.
+ *
+ * @author Dariusz Szablinski
+ * @see GenericMessage
+ */
+public class KafkaMessage<T> implements Message<T>, Serializable {
```
artembilan: Excellent!
But we don't need this class anymore :smile:

daroo: Right. Removed :smile:

artembilan:
LGTM.
Although I wonder if we should add some note into the Documentation about this new interesting `KafkaNull` hook. Even if `@KafkaListener` can simply deal with this via `Converter` many other application may not consult `ConversionService`, so they should know what that means for them.

From here looks like that even would be better to extract `KafkaNullToObjectConverter` to `public` class for example to be used from Spring Integration (`@IntegrationConverter`).

garyrussell:
@artembilan LGTM - I think we can defer the documentation until we resolve #164 - the converter is currently only used within the `@KafkaListener` infrastructure.

garyrussell:
However, we do need some code in `spring-integration-kafka` to coerce a `null` payload to a `KafkaNull` so we will need to document at some point.

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaNull.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.springframework.kafka.support;
+
+/**
+ * This class represents NULL Kafka payload.
+ *
+ * @author Dariusz Szablinski
+ */
+public enum KafkaNull {
```
daroo: This cannot be Enum. Enum values can be automatically converted to a String values ("INSTANCE").

spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
```diff
@@ -473,6 +481,11 @@ public void listen7(String foo) {
 			this.latch7.countDown();
 		}
 
+		@KafkaListener(id = "quux", topics = "annotated12")
+		public void listen8(@Payload(required = false) String none) {
+			this.latch8.countDown();
```
daroo: add assertion isNull(none)

daroo:
Unfortunately this code doesn't work correctly. The new converter is never called. I'm working on a fix

daroo:
Ok, I had to change `KafkaNull` to regular class. Also extract the converter to a class 
```java 
KafkaNullToObjectConverter implements GenericConverter
```
now conversion works and payload is converted to null, but I cannot pass the not null check in `org.springframework.messaging.handler.annotation.support.PayloadArgumentResolver#resolveArgument`
```java
payload = (this.converter instanceof SmartMessageConverter ?
		((SmartMessageConverter) this.converter).fromMessage(message, targetClass, parameter) :
		this.converter.fromMessage(message, targetClass));
if (payload == null) {
	throw new MessageConversionException(message,
			"No converter found to convert to " + targetClass + ", message=" + message);
```

artembilan:
Yep! The approach indeed doesn't work.
Look at the same `PayloadArgumentResolver`. Since our `KafkaNull` isn't `if (isEmptyPayload(payload)) {`, we go to the code like:
```
payload = (this.converter instanceof SmartMessageConverter ?
		((SmartMessageConverter) this.converter).fromMessage(message, targetClass, parameter) :
this.converter.fromMessage(message, targetClass));
if (payload == null) {
	throw new MessageConversionException(message,
		"No converter found to convert to " + targetClass + ", message=" + message);
}
```
Which isn't good solution again. 😢 

I reached that after some hacking for the `Converter`:
```
conversionService.addConverter(new GenericConverter() {

				@Override
				public Set<ConvertiblePair> getConvertibleTypes() {
					return Collections.singleton(new ConvertiblePair(KafkaNull.class, Object.class));
				}

				@Override
				public Object convert(Object source, TypeDescriptor sourceType, TypeDescriptor targetType) {
					return null;
				}

});
```

So, we again come back to the `Optional<?>` workaround and idea for some discussion with SF team.

In SI that should work because we don't deal with `PayloadArgumentResolver` there.

Right now I wonder if we can override it via this hook:
```
if (this.customArgumentResolvers != null) {
	resolvers.addAll(this.customArgumentResolvers);
}
resolvers.add(new PayloadArgumentResolver(this.messageConverter, this.validator));
```

daroo:
I've created pretty much the same converter :smile_cat: 
> So, we again come back to the Optional<?> workaround 
but `Optional<?>` is not an option :smile: because requires Java 8

BTW. I've found another problem: this `if (isEmptyPayload(payload))` in `PayloadArgumentResolver` it converts empty String or empty byte[] to null. That's not what I'd expect to get back from Kafka.

garyrussell:
Ugh; yes, that's wrong.

artembilan:
I found the solution like:
````java
private MessageHandlerMethodFactory createDefaultMessageHandlerMethodFactory() {
	DefaultMessageHandlerMethodFactory defaultFactory = new DefaultMessageHandlerMethodFactory();
		defaultFactory.setBeanFactory(KafkaListenerAnnotationBeanPostProcessor.this.beanFactory);
	ConfigurableBeanFactory cbf =
		(KafkaListenerAnnotationBeanPostProcessor.this.beanFactory instanceof ConfigurableBeanFactory ?
		(ConfigurableBeanFactory) KafkaListenerAnnotationBeanPostProcessor.this.beanFactory : null);
	DefaultFormattingConversionService conversionService = new DefaultFormattingConversionService();
	defaultFactory.setConversionService(conversionService);

	List<HandlerMethodArgumentResolver> argumentResolvers = new ArrayList<>();
	// Annotation-based argument resolution
	argumentResolvers.add(new HeaderMethodArgumentResolver(conversionService, cbf));
	argumentResolvers.add(new HeadersMethodArgumentResolver());

	// Type-based argument resolution
	argumentResolvers.add(new MessageMethodArgumentResolver());

	argumentResolvers.add(new PayloadArgumentResolver(new GenericMessageConverter(conversionService)) {

	@Override
	protected boolean isEmptyPayload(Object payload) {
		return payload instanceof KafkaNull || super.isEmptyPayload(payload);
	}

});

	defaultFactory.afterPropertiesSet();
	return defaultFactory;
}
````
it is exactly a copy/paste from the `DefaultMessageHandlerMethodFactory.initArgumentResolvers()`, which doesn't work in this case because:
```
if(this.argumentResolvers.getResolvers().isEmpty()) {              this.argumentResolvers.addResolvers(this.initArgumentResolvers());
}
```
isn't empty thanks to our override. And as you see I just override `isEmptyPayload()` for our specific type 😄

garyrussell:
👍

garyrussell:
However, `isEmptyPayload()` __must__ return false if the value is `""` or `byte[0]`.

garyrussell:
Forget that.

daroo:
@artembilan great solution :+1: 

@garyrussell 
>However, isEmptyPayload() must return false if the value is "" or byte[0].
>Forget that.

why?

garyrussell:
Doh - sorry; you are correct - I guess I am having a bad day.

However, `isEmptyPayload()` must return `false` if the value is `""` or `byte[0]` - in case that's the actual payload received from Kafka.

artembilan:
Agreed. So, we just shouldn't call `super.isEmptyPayload()` for that our case.
Only `null` and `KafkaNull` are valid for `null` to return to the target method argument.

garyrussell:
👍

daroo:
Cool. I'll submit updated PR shortly

artembilan:
Fine with me. Really cool. Ideal team work :smile: 

@garyrussell  ?

garyrussell:
LGTM

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/0f0ebed1d8c85134103188b321ce60f12aed4136 and cherry-picked to `master`.

@daroo , thank you for the contribution.
Looking forward for more!

daroo:
>@daroo , thank you for the contribution.
Looking forward for more!

No problem 😄  I will try to help as much as I can.

BTW. when are you going to release 1.0.3?

Issue #129, opened 2016-06-29
Consider Exposing Runtime Seek


garyrussell:
[Stack Overflow Question](http://stackoverflow.com/questions/38060283/kafka-with-manual-offset-management-spring-integration/38061015?noredirect=1#comment63624936_38061015).

jd73:
👍

PR #166, opened 2016-08-11
GH-142: Add Batch Listeners

### LINKED ISSUE #142: Support bulk processing of messages
Hi, 

Should there by an API in `MessageListener` to receive bulk records? Currently, once the call returns from `onMessage()`, then record is 'considered' processed, which simulates a blocking pattern. 
The acknowledgement variant of `onMessage()` does provide some relief in this matter, but it imposes a huge ordering requirement on the listener because `acknowledge()` would consider all records in that partition before the current one as processed, which may not be the case in a multi threaded scenario. 

So, I was thinking something like this would help: 
`List<ListenableFuture<?>> onMessage(List<ConsumerRecords<?,?> records)`, and then once the messages are done, (the list of futures is complete), the offsets can be committed (in the batch mode). 
In the TIME mode, on expiration of timer, we can iterate over the list and see which ones are complete and add them to `processedRecords`.

Or may be a variation like `void onMessage(List<ConsumerRecords<?,?> list, Acknowledgement ack)`, with 
`Acknowledgement` having another method `acknowledgement(ConsumerRecord<?,?> cr)`

What do you guys think?

garyrussell:
Resolves #142

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -197,8 +197,8 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	public final void start() {
 		synchronized (this.lifecycleMonitor) {
 			Assert.isTrue(
-					this.containerProperties.getMessageListener() instanceof MessageListener
-							|| this.containerProperties.getMessageListener() instanceof AcknowledgingMessageListener,
+					this.containerProperties.getMessageListener() instanceof IMessageListener
+							|| this.containerProperties.getMessageListener() instanceof IAcknowledgingMessageListener,
 					"Either a " + MessageListener.class.getName() + " or a "
 							+ AcknowledgingMessageListener.class.getName() + " must be provided");
 			doStart();
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -197,10 +197,8 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	public final void start() {
 		synchronized (this.lifecycleMonitor) {
 			Assert.isTrue(
-					this.containerProperties.getMessageListener() instanceof IMessageListener
-							|| this.containerProperties.getMessageListener() instanceof IAcknowledgingMessageListener,
-					"Either a " + MessageListener.class.getName() + " or a "
-							+ AcknowledgingMessageListener.class.getName() + " must be provided");
+					this.containerProperties.getMessageListener() instanceof KafkaDataListener,
+					"A " + KafkaDataListener.class.getName() + " implementation must be provided");
 			doStart();
 		}
 	}
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AcknowledgingMessageListener.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AcknowledgingMessageListener.java
@@ -18,8 +18,6 @@ package org.springframework.kafka.listener;
 
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 
-import org.springframework.kafka.support.Acknowledgment;
-
 /**
  * Listener for handling incoming Kafka messages, propagating an acknowledgment handle that recipients
  * can invoke when the message has been processed.
@@ -30,13 +28,6 @@ import org.springframework.kafka.support.Acknowledgment;
  * @author Marius Bogoevici
  * @author Gary Russell
  */
-public interface AcknowledgingMessageListener<K, V> {
-
-	/**
-	 * Executes when a Kafka message is received.
-	 * @param record the Kafka message to be processed
-	 * @param acknowledgment a handle for acknowledging the message processing
-	 */
-	void onMessage(ConsumerRecord<K, V> record, Acknowledgment acknowledgment);
+public interface AcknowledgingMessageListener<K, V> extends IAcknowledgingMessageListener<ConsumerRecord<K, V>> {
 
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AcknowledgingMessageListener.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AcknowledgingMessageListener.java
@@ -28,6 +28,6 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
  * @author Marius Bogoevici
  * @author Gary Russell
  */
-public interface AcknowledgingMessageListener<K, V> extends IAcknowledgingMessageListener<ConsumerRecord<K, V>> {
+public interface AcknowledgingMessageListener<K, V> extends GenericAcknowledgingMessageListener<ConsumerRecord<K, V>> {
 
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/IMessageListener.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/IMessageListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,17 +16,21 @@
 
 package org.springframework.kafka.listener;
 
-import org.apache.kafka.clients.consumer.ConsumerRecord;
-
 /**
- * Listener for handling individual incoming Kafka messages.
+ * Top level interface for listener types.
  *
- * @param <K> the key type.
- * @param <V> the value type.
+ * @param <T> the type.
  *
- * @author Marius Bogoevici
  * @author Gary Russell
+ * @since 1.1
+ *
  */
-public interface MessageListener<K, V> extends IMessageListener<ConsumerRecord<K, V>> {
+public interface IMessageListener<T> {
+
+	/**
+	 * Invoked with data from kafka.
+	 * @param data the data to be processed.
+	 */
+	void onMessage(T data);
 
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -83,9 +84,9 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 
 	private ListenableFuture<?> listenerConsumerFuture;
 
-	private MessageListener<K, V> listener;
+	private IMessageListener<?> listener;
 
-	private AcknowledgingMessageListener<K, V> acknowledgingMessageListener;
+	private IAcknowledgingMessageListener<?> acknowledgingMessageListener;
 
 	/**
 	 * Construct an instance with the supplied configuration properties.
@@ -136,7 +137,6 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 	}
 
 	@Override
-	@SuppressWarnings("unchecked")
 	protected void doStart() {
 		if (isRunning()) {
 			return;
@@ -156,11 +156,11 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 
 		Object messageListener = containerProperties.getMessageListener();
 		Assert.state(messageListener != null, "A MessageListener is required");
-		if (messageListener instanceof AcknowledgingMessageListener) {
-			this.acknowledgingMessageListener = (AcknowledgingMessageListener<K, V>) messageListener;
+		if (messageListener instanceof IAcknowledgingMessageListener) {
+			this.acknowledgingMessageListener = (IAcknowledgingMessageListener<?>) messageListener;
 		}
-		else if (messageListener instanceof MessageListener) {
-			this.listener = (MessageListener<K, V>) messageListener;
+		else if (messageListener instanceof IMessageListener) {
+			this.listener = (IMessageListener<?>) messageListener;
 		}
 		else {
 			throw new IllegalStateException("messageListener must be 'MessageListener' "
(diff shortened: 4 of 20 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/IMessageListener.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Top level interface for listener types.
+ *
+ * @param <T> the type.
+ *
+ * @author Gary Russell
+ * @since 1.1
+ *
+ */
+public interface IMessageListener<T> {
```
artembilan: How about something like `Generic*` instead of `I*`?
I don't mind against [C#](http://www.tutorialspoint.com/csharp/csharp_interfaces.htm), but at the same time I don't want to have their structures in our public Java API.
Just IMHO :smile:

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -197,8 +197,8 @@ public void setupMessageListener(Object messageListener) {
 	public final void start() {
 		synchronized (this.lifecycleMonitor) {
 			Assert.isTrue(
-					this.containerProperties.getMessageListener() instanceof MessageListener
-							|| this.containerProperties.getMessageListener() instanceof AcknowledgingMessageListener,
+					this.containerProperties.getMessageListener() instanceof IMessageListener
+							|| this.containerProperties.getMessageListener() instanceof IAcknowledgingMessageListener,
```
artembilan: Introducing one more level of abstraction for those generic variants, I won't mind if we add one more: `KafkaDataListener` - a marker interface to avoid similar asserts.
WDYT?

spring-kafka/src/main/java/org/springframework/kafka/listener/AcknowledgingMessageListener.java
```diff
@@ -30,13 +28,6 @@
...
  */
-public interface AcknowledgingMessageListener<K, V> {
-
-	/**
-	 * Executes when a Kafka message is received.
-	 * @param record the Kafka message to be processed
-	 * @param acknowledgment a handle for acknowledging the message processing
-	 */
-	void onMessage(ConsumerRecord<K, V> record, Acknowledgment acknowledgment);
+public interface AcknowledgingMessageListener<K, V> extends IAcknowledgingMessageListener<ConsumerRecord<K, V>> {
```
artembilan: I understand that it is late already, but, eh..., it isn't `MessageListener`. `ConsumerRecord` - right, but not for message...

spring-kafka/src/main/java/org/springframework/kafka/listener/IMessageListener.java
garyrussell: Yeah - I struggled with this too; `Generic` sounds good.

spring-kafka/src/main/java/org/springframework/kafka/listener/AcknowledgingMessageListener.java
garyrussell: You mean it should be named `ConsumerRecordListener`?

I guess, but we need some consistency with other technologies.

Perhaps we could have added a `Message` abstraction on top - but I didn't think it worth it since we won't have another "Kafka" implementation.

artembilan:
> I guess, but we need some consistency with other technologies.

Well, maybe `ConsumerRecordMessageListener`?..

But eh, I can live with that we have now. Just only the problem with naming when we try to expose more generic abstraction from already such a generic one :smile:

garyrussell:
Rebased, pushed.

artembilan:
Would you mind rebasing to `master` one more time?

garyrussell:
Rebased; pushed.

We should perhaps leave the travis debug for now - the project is small enough such that the debug output is not huge.

artembilan:
> We should perhaps leave the travis debug for now 

Well, not sure . See the current report. Seems for me it is in some mess (I mean there are so many DEBUG) and ends up wit the:
> The log length has exceeded the limit of 4 Megabytes 

We probably can increase that limit, but I'm not sure that it will help us for analyzing the target problem.
As for me that would be better to increase logging level for the test class and observe the problem on the Bamboo, like we do with other projects.

garyrussell:
ok; I was just trying to find out why that test fails - repushed without travis - if it fails again, I'll try with the single test.

artembilan:
No-no. I don't say that this your change has effect somehow. That test really used to fail sporadically.
Really some race condition in between send commit and retrieve an offset in a new consumer.

artembilan:
Different story though. Just wondered if you have something in mind.

I guess we can consider that in the separate issue.

garyrussell:
OK; tests passed this time - I will add a doc commit and we can merge, unless you have other issues with the PR.

garyrussell:
Pushed docs commit.

src/reference/asciidoc/kafka.adoc
```diff
@@ -369,6 +415,8 @@ In that case, the `ErrorHandler` will be invoked, if configured, or logged other
 
 When using `@KafkaListener`, set the `RetryTemplate` (and optionally `recoveryCallback`) on the container factory and the listener will be wrapped in the appropriate retrying adapter.
 
+A retry adapter is not provided for any of the batch <<message-listeners, message listeners>>.
```
artembilan: Sorry, I can't get "why?". Thanks

garyrussell: Because there's no easy way to determine the retry point - only the actual listener can determine that - he needs to implement retry internally.

Let's say the list has 10 records; 5 are successful - how would an adapter know that?

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -197,10 +197,8 @@ public void setupMessageListener(Object messageListener) {
 	public final void start() {
 		synchronized (this.lifecycleMonitor) {
 			Assert.isTrue(
-					this.containerProperties.getMessageListener() instanceof MessageListener
-							|| this.containerProperties.getMessageListener() instanceof AcknowledgingMessageListener,
-					"Either a " + MessageListener.class.getName() + " or a "
-							+ AcknowledgingMessageListener.class.getName() + " must be provided");
+					this.containerProperties.getMessageListener() instanceof KafkaDataListener,
+					"A " + KafkaDataListener.class.getName() + " implementation must be provided");
```
artembilan: Well, my idea was about making strong-typed `messageListener` property. WDYT ?

garyrussell: Doh; ok

garyrussell: Actually this is very painful to change - hundreds of references would need to be fixed (generics added to `ContainerProperties`, so my preference is STET).

artembilan: Not sure what is "STET", but eh, let it be. we can live without that coercion.
Let me come through the changes locally and I'll merge it if that.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -372,8 +379,41 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
 				consumer.assign(new ArrayList<>(this.definedPartitions.keySet()));
 			}
 			this.consumer = consumer;
-			this.listener = listener;
-			this.acknowledgingMessageListener = ackListener;
+			Object theListener = listener == null ? ackListener : listener;
```
artembilan: That might not make so much sense, but I'm changing this variable to the `KafkaDataListener<?>` for better code tracability.

garyrussell: OK

STET: verb
1.
let it stand (used as an instruction on a printed proof to indicate that a correction or alteration should be ignored).

😄

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -585,6 +625,54 @@ private void ackImmediate(ConsumerRecord<K, V> record) {
...
+							for (ConsumerRecord<K, V> record : recordList) {
+								this.acks.put(record);
+							}
+						}
+					}
+				}
+				catch (Exception e) {
+					if (this.containerProperties.isAckOnError() && !this.autoCommit) {
+						for (ConsumerRecord<K, V> record : recordList) {
+							this.acks.add(record);
```
artembilan: I don't remember if I asked similar question to @mbogoevici , when he had been doing the original `ackOnError` solution, but let me ask it again if that.

The main code uses `BlockingQueue.put()` - block/wait instead of this `IllegalState` `add()`.
I guess it doesn't make sense for the used `LinkedBlockingQueue`, but maybe it would be better to be consistent. At least to let me stop to remember such a case and don't ask this question again 😄

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -585,6 +625,54 @@ private void ackImmediate(ConsumerRecord<K, V> record) {
...
+						}
+					}
+					for (ConsumerRecord<K, V> record : recordList) {
+						if (this.containerProperties.getErrorHandler() != null) {
+							this.containerProperties.getErrorHandler().handle(e, record);
```
artembilan: This smells a bit. If we handle the list in the main listener, it feels like for symmetry we should do the same `batchErrorHandling`.
WDYT?

garyrussell: Yes, I felt a bit uncomfortable with this too. Will rework.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -87,9 +87,13 @@
 
 	private static String topic7 = "testTopic7";
 
+	private static String topic8 = "testTopic8";
+
+	private static String topic9 = "testTopic9";
```
artembilan: Looks like `topic9` is out of use, but I guess you will need it when you've done with `batchErrorHandling` 😄

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -560,6 +564,67 @@ public void testBatchAck() throws Exception {
...
+		template.sendDefault(1, 0, "bar");
+		template.sendDefault(1, 0, "qux");
+		template.flush();
+
+		Map<String, Object> props = KafkaTestUtils.consumerProps("test8", "false", embeddedKafka);
+		props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+		DefaultKafkaConsumerFactory<Integer, String> cf = new DefaultKafkaConsumerFactory<>(props);
+		ContainerProperties containerProps = new ContainerProperties(topic8);
+		containerProps.setMessageListener((BatchMessageListener<Integer, String>) message -> {
+			logger.info("batch listener: " + message);
```
artembilan: I'd say that test-case doesn't demonstrate `BatchMessageListener` power. 
Wouldn't it be more closer to reality with `MANUAL` ack, when we `ListenerConsumer.this.consumer.wakeup();` only once for the whole batch?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Both `put` and `add` can throw an `IllegalArgumentException` - it's based on the argument, not the queue state. If we change this to `put` we'd have to add an unnecessary try/catch here - it's un unbounded queue.

artembilan: My assumption has been done on this JavaDoc for `add()`:
```
 * Inserts the specified element into this queue if it is possible to do
     * so immediately without violating capacity restrictions, returning
     * {@code true} upon success and throwing an
     * {@code IllegalStateException} if no space is currently available.
     * When using a capacity-restricted queue, it is generally preferable to
     * use {@link #offer(Object) offer}.
```
That's why I say we don't need to worry about capacity for the `LinkedBlockingQueue`. I just wondered why a different method has been chosen...

src/reference/asciidoc/kafka.adoc
```diff
@@ -126,9 +126,57 @@ Note, however that flushing will likely significantly reduce performance.
...
 
+[[message-listeners]]
+===== Message Listeners
+
+When using a <<message-listener-container, Message Listener Container>> you must provide a listener to receive data.
+There are currently four supported interfaces for message listeners:
+
+[source, java]
+----
+public interface MessageListener<K, V> {} <1>
```
artembilan: That call-out link feature looks really cool 😄 !

OK. that's all for me.
Hope the next round with `batchErrorHandling` will be the last one to merge.

src/reference/asciidoc/kafka.adoc
```diff
@@ -126,9 +126,57 @@ Note, however that flushing will likely significantly reduce performance.
 
 ==== Receiving Messages
 
-Messages can be received by configuring a `MessageListenerContainer` and providing a `MessageListener`, or by
+Messages can be received by configuring a `MessageListenerContainer` and providing a <<message-listener, Message Listener>>, or by
```
artembilan: Well, not yet.

This link is a bug.
We have a tag as `message-listeners`, but OTOH it is just a next paragraph, so may be there is just no need in the link.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
garyrussell: That's why I added `topic9` but was too lazy to do a manual version of this test (which used `AckMode.BATCH`).

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: >I just wondered why a different method has been chosen...

Just to avoid another catch block.

garyrussell:
Pushed

ruanming:
Can seek the position when listener process failed?

garyrussell:
Not currently - we have an open issue to support seek: https://github.com/spring-projects/spring-kafka/issues/129

PR #169, opened 2016-08-16
GH-142: Add Batch Listeners

### LINKED ISSUE #142: Support bulk processing of messages

garyrussell:
Resolves #142 

Replaces #166 (accidentally deleted my branch on GH).

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchLoggingErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchLoggingErrorHandler.java
@@ -18,7 +18,8 @@ package org.springframework.kafka.listener;
 
 import java.util.Iterator;
 
-import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 
 /**
@@ -30,14 +31,16 @@ import org.apache.kafka.clients.consumer.ConsumerRecords;
  */
 public class BatchLoggingErrorHandler implements BatchErrorHandler {
 
-	private final LoggingErrorHandler recordHandler = new LoggingErrorHandler();
+	private static final Log log = LogFactory.getLog(BatchLoggingErrorHandler.class);
 
 	@Override
 	public void handle(Exception thrownException, ConsumerRecords<?, ?> data) {
 		Iterator<?> iterator = data.iterator();
+		StringBuilder message = new StringBuilder("Error while processing:\n");
 		while (iterator.hasNext()) {
-			this.recordHandler.handle(thrownException, (ConsumerRecord<?, ?>) iterator.next());
+			message.append(iterator.next()).append('\n');
 		}
+		log.error(message.substring(0, message.length() - 1), thrownException);
 	}
 
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchLoggingErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchLoggingErrorHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,14 +16,28 @@
 
 package org.springframework.kafka.listener;
 
+import java.util.Iterator;
+
 import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
 
 /**
- * Handles errors thrown during the execution of a {@link MessageListener}.
+ * Simple handler that invokes a {@link LoggingErrorHandler} for each record.
  *
- * @author Marius Bogoevici
  * @author Gary Russell
+ * @since 1.1
+ *
  */
-public interface ErrorHandler extends GenericErrorHandler<ConsumerRecord<?, ?>> {
+public class BatchLoggingErrorHandler implements BatchErrorHandler {
+
+	private final LoggingErrorHandler recordHandler = new LoggingErrorHandler();
+
+	@Override
+	public void handle(Exception thrownException, ConsumerRecords<?, ?> data) {
+		Iterator<?> iterator = data.iterator();
+		while (iterator.hasNext()) {
+			this.recordHandler.handle(thrownException, (ConsumerRecord<?, ?>) iterator.next());
+		}
+	}
 
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -83,9 +84,9 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 
 	private ListenableFuture<?> listenerConsumerFuture;
 
-	private MessageListener<K, V> listener;
+	private IMessageListener<?> listener;
 
-	private AcknowledgingMessageListener<K, V> acknowledgingMessageListener;
+	private IAcknowledgingMessageListener<?> acknowledgingMessageListener;
 
 	/**
 	 * Construct an instance with the supplied configuration properties.
@@ -136,7 +137,6 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 	}
 
 	@Override
-	@SuppressWarnings("unchecked")
 	protected void doStart() {
 		if (isRunning()) {
 			return;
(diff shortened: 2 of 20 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -874,6 +990,43 @@ public String toString() {
...
 
+		private final class ConsumerBatchAcknowledgment implements Acknowledgment {
+
+			private final List<ConsumerRecord<K, V>> records;
+
+			private final boolean immediate;
+
+			private ConsumerBatchAcknowledgment(List<ConsumerRecord<K, V>> records, boolean immediate) {
+				// make a copy in case the listener alters the list
+				this.records = new LinkedList<ConsumerRecord<K, V>>(records);
```
artembilan: That is precisely a case. See `FilteringBatchMessageListenerAdapter`.
But we have to ack those records anyway.
Correct?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchAcknowledgingMessageListenerAdapter.java
```diff
@@ -0,0 +1,67 @@
...
+	@Override
+	public void onMessage(List<ConsumerRecord<K, V>> consumerRecords, Acknowledgment acknowledgment) {
+		Iterator<ConsumerRecord<K, V>> iterator = consumerRecords.iterator();
+		while (iterator.hasNext()) {
+			if (filter(iterator.next())) {
+				iterator.remove();
+			}
+		}
+		if (consumerRecords.size() > 0) {
+			this.delegate.onMessage(consumerRecords, acknowledgment);
```
artembilan: Well, if we filter all the records and use `MANUAL` (it is a fact for this listener type), we never acknowledge them.
Should we send an empty list anyway?
Or just `acknowledge()` directly in the `else`?

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -560,6 +565,161 @@ public void testBatchAck() throws Exception {
...
+		props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+		DefaultKafkaConsumerFactory<Integer, String> cf = new DefaultKafkaConsumerFactory<>(props);
+		ContainerProperties containerProps = new ContainerProperties(topic9);
+		final CountDownLatch latch = new CountDownLatch(4);
+		containerProps.setMessageListener((BatchAcknowledgingMessageListener<Integer, String>) (messages, ack) -> {
+			logger.info("batch listener manual: " + messages);
+			for (int i = 0; i < messages.size(); i++) {
+				latch.countDown();
+			}
+			ack.acknowledge();
```
artembilan: This is definitely a race condition.
You `countDown()` to zero and let barrier to overcome, but you send commits only after that.
So, all those assertions in the end are going to fail from time to time.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -560,6 +565,161 @@ public void testBatchAck() throws Exception {
...
+
+		Map<String, Object> props = KafkaTestUtils.consumerProps("test9", "false", embeddedKafka);
+		props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+		DefaultKafkaConsumerFactory<Integer, String> cf = new DefaultKafkaConsumerFactory<>(props);
+		ContainerProperties containerProps = new ContainerProperties(topic10);
+		containerProps.setMessageListener((BatchMessageListener<Integer, String>) messages -> {
+			logger.info("batch listener errors: " + messages);
+			throw new RuntimeException("intentional");
+		});
+		containerProps.setSyncCommits(true);
```
artembilan: Well, I think we still have some race condition with such an approach.
Yes, the commit is sync, indeed. But it is done in the `ListenerConsumer` Thread, not our main to let us to be blocked until commit.
Therefore I think all those assertions in the end are going to fail from time to time.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Yes; filtered messages need to be ack'd.

spring-kafka/src/main/java/org/springframework/kafka/listener/BatchLoggingErrorHandler.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public class BatchLoggingErrorHandler implements BatchErrorHandler {
+
+	private final LoggingErrorHandler recordHandler = new LoggingErrorHandler();
+
+	@Override
+	public void handle(Exception thrownException, ConsumerRecords<?, ?> data) {
+		Iterator<?> iterator = data.iterator();
+		while (iterator.hasNext()) {
+			this.recordHandler.handle(thrownException, (ConsumerRecord<?, ?>) iterator.next());
```
artembilan: Why can't it just log an entire batch? And only once.
Otherwise we see a lot of StackTrace thanks to:
```
log.error("Error while processing: " + ObjectUtils.nullSafeToString(record), thrownException);
```
Isn't it?
(Sorry haven't tested)

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchAcknowledgingMessageListenerAdapter.java
garyrussell: Good catch

artembilan:
And a general comment. Maybe deserves a separate issue.

I wonder if we can come up with the solution for the case like:
```
@KafkaListener(topics = "foo")
public void handleBatch(List<String> batch) {
...
}
```
Or at least just for `List<ConsumerRecord>`...

garyrussell:
I thought about that, but it would be a lot of work so I decided to not do it and/or wait until someone asks for it.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchAcknowledgingMessageListenerAdapter.java
artembilan: An `ackDiscarded` then? See `FilteringAcknowledgingMessageListenerAdapter`.

garyrussell:
Pushed

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -661,8 +661,25 @@ public void testBatchListenerManual() throws Exception {
...
+
+			@SuppressWarnings({ "unchecked" })
+			Map<TopicPartition, OffsetAndMetadata> map = (Map<TopicPartition, OffsetAndMetadata>) invocation
+					.getArguments()[0];
+			for (Entry<TopicPartition, OffsetAndMetadata> entry : map.entrySet()) {
+				if (entry.getValue().offset() == 2) {
+					commitLatch.countDown();
+				}
+			}
+			return invocation.callRealMethod();
```
artembilan: This likely is still race condition. We overcome barrier before the real `commitSync()` is performed on the Kafka.
Can't we do `invocation.callRealMethod()` before `countDown()`?

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -700,6 +717,7 @@ public void testBatchListenerErrors() throws Exception {
 		containerProps.setAckOnError(true);
 		final CountDownLatch latch = new CountDownLatch(4);
 		containerProps.setErrorHandler((BatchErrorHandler) (t, messages) -> {
+			new BatchLoggingErrorHandler().handle(t, messages);
```
artembilan: Cool! :smile:

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
garyrussell: You are correct - but this was your code 😄 - copied from `testBatchAck()`.

artembilan: Well, I don't say that I was so smart that time. And I am still not sure that I am such right now or will be tomorrow.
There is really some magic around that makes some explosions in a head and after that it needs abundant firefighting. :smile:

artembilan:
Great!
I think I merge it with that simple fix for `e.printStackTrace()`, although I have found one more mock answer with similar race condition :smile: : 
```
willAnswer(invocation -> {

			latch.countDown();
			return invocation.callRealMethod();

		}).given(consumer)
				.commitSync(any());
```
Fixing as well...

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -540,10 +551,20 @@ public void testBatchAck() throws Exception {
...
+				result = invocation.callRealMethod();
+			}
+			finally {
+				for (Entry<TopicPartition, OffsetAndMetadata> entry : map.entrySet()) {
+					if (entry.getValue().offset() == 2) {
+						latch.countDown();
+					}
+				}
+			}
+			return result;
```
artembilan: Well, actually this can be simplified with direct `return` in the `try`. The `finally` is invoked before exit from the code block: http://www.programmerinterview.com/index.php/java-questions/will-finally-run-after-return/

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/1a86905bba1e962dc1544ef2d8741569938fa996

PR #171, opened 2016-08-22
Restore ErrorHandler Setter

garyrussell:
API change prevented an app compiled against 1.0 running with 1.1.


sobychacko:
Ran tests with new version. LGTM. Merged.

PR #173, opened 2016-08-23
GH-129: Implement Consumer Seek

### LINKED ISSUE #129: Consider Exposing Runtime Seek

garyrussell:
Resolves #129 

Review only @mbogoevici @artembilan 

TODO: docs, pass callback through adapters.


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -384,10 +390,9 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
 				consumer.assign(new ArrayList<>(this.definedPartitions.keySet()));
 			}
 			this.consumer = consumer;
-			Object theListener = listener == null ? ackListener : listener;
 			GenericErrorHandler<?> errHandler = this.containerProperties.getGenericErrorHandler();
-			if (theListener instanceof MessageListener) {
-				this.listener = (MessageListener<K, V>) theListener;
+			if (this.theListener instanceof MessageListener) {
+				this.listener = (MessageListener<K, V>) this.theListener;
 				this.batchListener = null;
```
stepio: As I see, this code is a part of constructor. In this case initializing object's field with `null` is unnecessary - java does this. So now it's just extra code, which makes the method longer and less readable.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -407,21 +412,21 @@ else if (theListener instanceof BatchMessageListener) {
...
 			}
-			else if (theListener instanceof AcknowledgingMessageListener) {
+			else if (this.theListener instanceof AcknowledgingMessageListener) {
 				this.listener = null;
-				this.acknowledgingMessageListener = (AcknowledgingMessageListener<K, V>) theListener;
+				this.acknowledgingMessageListener = (AcknowledgingMessageListener<K, V>) this.theListener;
 				this.batchListener = null;
 				this.batchAcknowledgingMessageListener = null;
 				this.isBatchListener = false;
 				validateErrorHandler(false);
```
stepio: This and next two lines can be moved into a separate if-else block below, depending on just `this.isBatchListener`

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: That is incorrect - since the field is marked `final`, the compiler requires it to be explicitly initialized.

stepio: Good point, missed `final`.

artembilan:
Must be rebased to the latest `master`.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -896,6 +921,11 @@ private void commitIfNecessary() {
 			}
 		}
 
+		@Override
+		public void seek(String topic, int partition, long offset) {
+			this.seeks.add(new TopicPartitionInitialOffset(topic, partition, offset));
```
artembilan: I wonder why we can't reach the same via just top-level container operation.
Moreover expose it to JMX.
What am I missing?
Thanks

artembilan:
The solution looks cool, but I'm curious what drove you away of the simple `seek()` on the `AbstractMessageListenerContainer` directly.

With good valuable argument I can merge it as is or let you go ahead with Docs :smile: 
Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -431,9 +436,12 @@ else if (theListener instanceof BatchAcknowledgingMessageListener) {
 			else {
 				throw new IllegalArgumentException("Listener must be one of 'MessageListener', "
 						+ "'BatchMessageListener', 'AcknowledgingMessageListener', "
-						+ "'BatchAcknowledgingMessageListener', not " + theListener.getClass().getName());
+						+ "'BatchAcknowledgingMessageListener', not " + this.theListener.getClass().getName());
 			}
 			Assert.state(!this.isBatchListener || !this.isRecordAck, "Cannot use AckMode.RECORD with a batch listener");
+			if (this.autoCommit && this.theListener instanceof ConsumerSeekAware) {
+				((ConsumerSeekAware) this.theListener).registerSeekCallback(this);
```
artembilan: Why do we need this one if we `registerSeekCallback()` any way in the `ListenerInvoker.run()` ?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: `KafkaMessageListenerContainer` has `1-1` relationship with the `ListenerConsumer`. So, I believe, the question is closed here.

The `ConcurrentMessageListenerContainer` builds a set of `containers` based on the `partitionSubset(containerProperties, i)`. Together with the `Collection<TopicPartition> getAssignedPartitions()`, it seems for me we can reach a uniqueness and correctly assign requested `seek` to the appropriate `KafkaMessageListenerContainer`.
No ?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Because with `autoCommit = true`, the listener is invoked on the consumer thread.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: My concern is that the topics/partitions can move when the broker does a rebalance of the assignment - I want to be sure that the seek is done in the context of the current state at the time the listener is called. That way, if the consumer no longer owns that partition the seek operation will fail.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Thanks - fixed this in [this PR](https://github.com/spring-projects/spring-kafka/pull/176).

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AbstractRetryingMessageListenerAdapter.java
```diff
@@ -32,34 +33,51 @@
  * @author Gary Russell
  *
  */
-public class AbstractRetryingMessageListenerAdapter<K, V> {
+public abstract class AbstractRetryingMessageListenerAdapter<K, V> implements ConsumerSeekAware {
 
 	protected final Log logger = LogFactory.getLog(this.getClass());
 
+	private final ConsumerSeekAware delegate;
```
artembilan: `seekAware`, too ?
See `AbstractFilteringMessageListener`

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchAcknowledgingMessageListenerAdapter.java
```diff
@@ -65,7 +65,7 @@ public FilteringBatchAcknowledgingMessageListenerAdapter(BatchAcknowledgingMessa
 	 */
 	public FilteringBatchAcknowledgingMessageListenerAdapter(BatchAcknowledgingMessageListener<K, V> delegate,
 			RecordFilterStrategy<K, V> recordFilterStrategy, boolean ackDiscarded) {
-		super(recordFilterStrategy);
+		super(delegate, recordFilterStrategy);
 		this.delegate = delegate;
```
artembilan: Maybe we can just store `delegate` in `super`?
Right, minor, but I don't know already what to complain about here.
Now I understand all that thread callback dance and just enjoy watching to the code :smile:

garyrussell:
Hmmm - no travis check - weird.

artembilan:
>  no travis check 

```
This branch has conflicts that must be resolved
```
Can't that be a reason?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: 1. I wasn't able to step in here in debug mode during testing. Maybe we just don't have any test on the matter?..
2. Having your condition, "the listener is invoked on the consumer thread."  and "listeners should store the callback in a `{@code ThreadLocal}`.", plus the test-case in the `EnableKafkaIntegrationTests` with exactly a `ThreadLocal`, I dug further and found this code in the `KafkaMessageListenerContainer.doStart()`:
```
this.listenerConsumer = new ListenerConsumer(this.listener, this.acknowledgingMessageListener);
		setRunning(true);
		this.listenerConsumerFuture = containerProperties
				.getConsumerTaskExecutor()
```
So, we create `ListenerConsumer` instance in the `KafkaMessageListenerContainer.doStart()` Thread, but `run()` it in a different Thread.

My concern that this `.registerSeekCallback(this)` for the `autoCommit = true` is done in a wrong Thread. Therefore I won't be able to get a proper `ThreadLocal` value, when I consume messages.

Am I missing anything?
Thanks

garyrussell: Yes, we don't have a test case; will add one - good catch - I had a similar problem with the other case - I called from the ctor instead of the run method in the `ListenerInvoker`. See the second commit where I moved that code.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AbstractMessageListenerAdapter.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Top level class for all listener adapters.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ * @param <T> the delegate type.
+ *
+ * @author Gary Russell
+ * @since 5.0
```
artembilan: `1.1` :smile:

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AbstractMessageListenerAdapter.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ */
+public abstract class AbstractMessageListenerAdapter<K, V, T> implements ConsumerSeekAware {
+
+	protected final T delegate; //NOSONAR
```
artembilan: I found Sonar as a stupid here.
It doesn't see the the property is `final`, so it is fine to have access to it.

artembilan:
Fine with me.

Do you still have something else to do here, e.g. Docs?
Otherwise I can merge.

garyrussell:
Will add a docs commit tomorrow.

spring-kafka/src/main/java/org/springframework/kafka/support/TopicPartitionCurrentOffset.java
```diff
@@ -0,0 +1,85 @@
...
+package org.springframework.kafka.support;
+
+import java.util.Objects;
+
+import org.apache.kafka.common.TopicPartition;
+
+/**
+ * Identifies the current position of a TopicPartition.
+ *
+ * @author Artem Bilan
```
artembilan: This is your class and `@since 1.1` 
:smile: 

Although I still try to understand if we need it...
Maybe simple `Map<TopicPartition, Long>` could be enough?

spring-kafka/src/main/java/org/springframework/kafka/support/TopicPartitionCurrentOffset.java
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	@Override
+	public int hashCode() {
+		return this.topicPartition.hashCode();
+	}
+
+	@Override
+	public String toString() {
+		return "TopicPartitionInitialOffset{" +
```
artembilan: Typo. Should be `TopicPartitionCurrentOffset`

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -870,9 +885,17 @@ public void registerSeekCallback(ConsumerSeekCallback callback) {
 		template.sendDefault(0, 0, "baz");
 		template.sendDefault(1, 0, "qux");
 		template.flush();
-		assertThat(latch.await(60, TimeUnit.SECONDS)).isTrue();
+		assertThat(latch.get().await(60, TimeUnit.SECONDS)).isTrue();
 		container.stop();
 		assertThat(messageListener.registerThread).isSameAs(messageListener.messageThread);
+
+		// Now test initial seek of assigned partitions.
+		latch.set(new CountDownLatch(2));
```
artembilan: Interesting trick.
Will take in into service for future tasks :smile:

src/reference/asciidoc/kafka.adoc
```diff
@@ -563,6 +502,110 @@ CAUTION: If you wish to use the idle event to stop the lister container, you sho
...
+
+[source, java]
+----
+void registerSeekCallback(ConsumerSeekCallback callback);
+
+void onPartionsAssigned(Collection<TopicPartitionCurrentOffset> assignments, ConsumerSeekCallback callback);
+----
+
+The first is called during initialization; this callback should be used when seeking at some arbitrary time after initialization.
+You should save a reference to the callback; if you are using the same listener in multiple containers (or in a `ConcurrentMessageListenerContainer`) you should store the callback in a `ThreadLocal` or some other structure keyed by the listener `Thread`.
```
artembilan: He-he!
> keyed by the listener `Thread`.

That leads only to the `ThreadLocal` logic:
```
 /* ThreadLocal values pertaining to this thread. This map is maintained
  * by the ThreadLocal class. */
ThreadLocal.ThreadLocalMap threadLocals = null;
```
I don't see reason to do some other storage based on `Thread` key, but, eh, let it be, it is end-user choice.

artembilan:
I think we need `What's New` note also.

artembilan:
Rebase is needed as well.

spring-kafka/src/main/java/org/springframework/kafka/support/TopicPartitionCurrentOffset.java
artembilan: Or maybe we can just rename `TopicPartitionInitialOffset` to `TopicPartitionOffset`...

artembilan:
That's all for this round.

garyrussell:
Needs rebase

spring-kafka/src/main/java/org/springframework/kafka/support/TopicPartitionCurrentOffset.java
garyrussell: I changed it to a Map.

artembilan:
I think I'm fine with the fix.

Polling locally for the last review round and probably merge...

spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerSeekAware.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @author Gary Russell
+ * @since 1.1
+ *
+ */
+public interface ConsumerSeekAware {
+
+	/**
+	 * Register the callback to use when seeking at some arbitrary time. When used with a
+	 * {@code ConcurrentMessageListenerContainer} or the same listener instance in multupe
```
artembilan: Typo: multiple

spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerSeekAware.java
```diff
@@ -0,0 +1,73 @@
...
+	 * @param callback the callback.
+	 */
+	void registerSeekCallback(ConsumerSeekCallback callback);
+
+	/**
+	 * When using group management, called when partition assignments change.
+	 * @param assignments the new assignments and their current offsets.
+	 * @param callback the callback to perform an initial seek after assignment.
+	 */
+	void onPartionsAssigned(Map<TopicPartition, Long> assignments, ConsumerSeekCallback callback);
```
artembilan: Critical typo: `onPartitions`

spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerSeekAware.java
```diff
@@ -0,0 +1,73 @@
...
+	 * when the container idle event is emitted - allowing a seek operation.
+	 * @param assignments the new assignments and their current offsets.
+	 * @param callback the callback to perform a seek.
+	 */
+	void onIdleContainer(Map<TopicPartition, Long> assignments, ConsumerSeekCallback callback);
+
+	/**
+	 * A callback that a listener can invoke to seek to a specific offset.
+	 */
+	public interface ConsumerSeekCallback {
```
artembilan: Any interface members are always `public`.
Any properties are `static final` also.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerSeekAware.java
garyrussell: Good catch; thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -437,6 +448,27 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
...
+		private void seekPartitions(Collection<TopicPartition> partitions, boolean idle) {
+			Map<TopicPartition, Long> current = new HashMap<>();
+			for (TopicPartition topicPartition : partitions) {
+				current.put(topicPartition,	ListenerConsumer.this.consumer.position(topicPartition));
+			}
+			ConsumerSeekCallback callback = new ConsumerSeekCallback() {
+
+				@Override
+				public void seek(String topic, int partition, long offset) {
+					ListenerConsumer.this.consumer.seek(new TopicPartition(topic, partition), offset);
```
artembilan: Why do we `seek` here directly instead of placing to the `seeks` queue ?

garyrussell: Because this is guaranteed to run on the consumer thread so we might as well immediately seek.

`invokeListener` runs on a different thread (unless `autoCommit` is set).

garyrussell: Plus, rebalance listener `onPartitionsAssigned()` is called from within the `poll()` so we need to seek before we actually poll for data. Otherwise we'd get data first and then seek.

artembilan: It doesn't relate to `invokeListener`. 
We preform `processSeeks()` in the `ListenerConsumer.run()`. And that is the same place for the `seekPartitions(getAssignedPartitions(), true);` in case of idle. Therefore the same `Thread`.
But, yeah, I see that in case of `seeks` we will wait for the next `while (isRunning()) {` round and won't block the listener's Thread. Maybe that is a point.
Looks like the same is applicable for the logic in the `onPartitionsAssigned()`:
```
* This callback will execute in the user thread as part of the {@link Consumer#poll(long) poll(long)} call whenever partition assignment changes.
```
from `ConsumerRebalanceListener` JavaDocs.

I won't mind either way. Just want to understand why there is a difference in logic...

artembilan: > Otherwise we'd get data first and then seek.

Not sure in your position. We have this code:
```
if (!this.autoCommit) {
						processCommits();
					}
					processSeeks();
					if (this.logger.isTraceEnabled()) {
						this.logger.trace("Polling (paused=" + this.paused + ")...");
					}
					ConsumerRecords<K, V> records = this.consumer.poll(this.containerProperties.getPollTimeout());
```

So, in the main loop we seek first and only after that poll.

artembilan: OK. I'm merging with minor polishing for all those typos.
And you can still comment here if that.

Technically there is no so much difference, so we can live with the current listener blocking solution during these special seeks.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/b18ef3e2866a08549dedde4724638074309f17aa

PR #176, opened 2016-08-25
GH-175: Fix Acknowledgment with @KafkaListener

### LINKED ISSUE #175: Batch Listener Broke @KafkaListener with Acknowledgment Parameter
The `MessagingMessageListenerAdapter` implements both `AcknowledgingMessageListener` and `MessageListener`.

Since the batch support was added, we check for ML before AML so the `Acknowledgment` will not be passed in.

Previously, we checked for AML before ML when deciding which method to call.

garyrussell:
Fixes #175

The `MessagingMessageListenerAdapter` implements both `AcknowledgingMessageListener`
and `MessageListener`.

Since the batch support was added, we check for ML before AML so the `Acknowledgment`
will not be passed in.

Previously, we checked for AML before ML when deciding which method to call.

Also, move the creation of the rebalance listener out of the constructor
to make it smaller.


artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/5599cebf842aef2d93b3b7a6ee7c83a9e5f5e554

Issue #174, opened 2016-08-25
Add support for @KafkaListener to handle batching 


ssherwood:
GH-142 has added batch support but I was also interested in seeing this capability exposed as part of the KafkaListener annotation.  In pull https://github.com/spring-projects/spring-kafka/pull/169 @artembilan and @garyrussell suggest this capability but were holding off on it until someone asks for it.  I guess that I am now asking for it.

As described, it could be something like this:
````
@KafkaListener(topics = "foo")
public void handleBatch(List<String> batch) {
...
}
````
Or
````
@KafkaListener(topics = "foo")
public void handleBatch(List<ConsumerRecord> batch) {
...
}
````
This could compliment the `max.poll.records` that was introduced in Kafka 0.10 as a way of defining the limit to be consumed in a single batch.  There probably should also be a maximum amount of wait time for each batch just in case the max records doesn't fill up so that it can still be consumed by the listener in a timely fashion.

I'm asking for this capability specifically for the ability to be able to produce a batch insert into a downstream database with each call to the "handleBatch".  In our use cases, there are some performance improvements that are achieved when handling Kafka records this way.

garyrussell:
It may not be as much work as I thought when I added that comment - I'll take a look.

garyrussell:
It did turn out to be a pain as I suspected but I basically got the first option (`List<String>` working) `List<ConsumerRecord>` is going to be a bit problematic but I have a couple of ideas.

I do populate some of the metadata in headers...

		@KafkaListener(id = "list2", topics = "annotated15", containerFactory = "batchFactory")
		public void listen11(List<String> list,
				@Header(KafkaHeaders.RECEIVED_MESSAGE_KEY) List<Integer> keys,
				@Header(KafkaHeaders.RECEIVED_PARTITION_ID) List<Integer> partitions,
				@Header(KafkaHeaders.RECEIVED_TOPIC) List<String> topics,
				@Header(KafkaHeaders.OFFSET) List<Long> offsets) {
			this.payload = list;
			this.keys = keys;
			this.partitions = partitions;
			this.topics = topics;
			this.offsets = offsets;
			this.latch11.countDown();
		}





Commit is on [this branch](https://github.com/garyrussell/spring-kafka/tree/GH-174).

artembilan:
M-m-m. Can't then support something like ` listen11(List<Message<?>> batch)` ?
(I haven't looked to your code yet).
[...]

PR #177, opened 2016-08-26
GH-174: Support Batch with @KafkaListener

garyrussell:
Resolves: #174


garyrussell:
Todo: `List<Message<?>>`.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessageConverter.java
```diff
@@ -16,36 +16,13 @@
...
-
-import org.springframework.kafka.support.Acknowledgment;
-import org.springframework.messaging.Message;
-
 /**
- * A Kafka-specific {@link Message} converter strategy.
+ * A top level interface for message converters.
  *
  * @author Gary Russell
+ * @since 5.0
```
artembilan: `1.1`

spring-kafka/src/main/java/org/springframework/kafka/support/converter/RecordMessageConverter.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.producer.ProducerRecord;
+
+import org.springframework.kafka.support.Acknowledgment;
+import org.springframework.messaging.Message;
+
+/**
+ * A Kafka-specific {@link Message} converter strategy.
+ *
+ * @author Gary Russell
```
artembilan: `@since 1.1`

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessageConverter.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.kafka.support.Acknowledgment;
+import org.springframework.messaging.Message;
+
+/**
+ * A Kafka-specific {@link Message} converter strategy.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Gary Russell
```
artembilan: `@since 1.1`.
And maybe say in JavaDocs that it is batch-based.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessageConverter.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public interface BatchMessageConverter<K, V> extends MessageConverter {
```
artembilan: `RecordMessageConverter` isn't generic somehow...

artembilan: I mean that it seems for me that this one can live without generics as well.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/BatchMessagingMessageListenerAdapter.java
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * A {@link org.springframework.kafka.listener.MessageListener MessageListener}
+ * adapter that invokes a configurable {@link HandlerAdapter}.
+ *
+ * <p>Wraps the incoming Kafka Message to Spring's {@link Message} abstraction.
+ *
+ * <p>The original {@link ConsumerRecord} and
+ * the {@link Acknowledgment} are provided as additional arguments so that these can
+ * be injected as method arguments if necessary.
```
artembilan: The JavaDocs must say that the class is about `batching` and `@since 1.1`

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessageConverter.java
garyrussell: I couldn't make it so - some limitation with generics and containers.

When I made it non-generic, for

```java
	protected Message<?> toMessagingMessage(List<ConsumerRecord<K, V>> records, Acknowledgment acknowledgment) {
		return getBatchMessageConverter().toMessage(records, acknowledgment, this.inferredType);
	}
```

I get 

    Description	Resource	Path	Location	Type
    The method toMessage(List<ConsumerRecord<?,?>>, Acknowledgment, Type) in the type BatchMessageConverter is not applicable for the arguments (List<ConsumerRecord<K,V>>, Acknowledgment, Type)	BatchMessagingMessageListenerAdapter.java	/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter	line 124	Java Problem

I'd be happy if you know a work-around other than making it generic.

artembilan: How about this:
```
@SuppressWarnings({"rawtypes", "unchecked"})
protected Message<?> toMessagingMessage(List records, Acknowledgment acknowledgment) {
	return getBatchMessageConverter().toMessage((List<ConsumerRecord<?, ?>>) records, acknowledgment, 
				this.inferredType);
}
```
in the `BatchMessagingMessageListenerAdapter` ?

garyrussell: Ugh - I hate `rawTypes` but I guess it's better than adding generics for no reason.

src/reference/asciidoc/kafka.adoc
```diff
@@ -380,6 +380,75 @@ public void listen(@Payload String foo,
...
+To configure the listener container factory to create batch listeners, set the `batchListener` property:
+
+[source, java]
+----
+@Bean
+public KafkaListenerContainerFactory<?> batchFactory() {
+    ConcurrentKafkaListenerContainerFactory<Integer, String> factory =
+            new ConcurrentKafkaListenerContainerFactory<>();
+    factory.setConsumerFactory(consumerFactory());
+    factory.setBatchListener(true);  // <<<<<<<<<<<<<<<<<<<<<<<<<
```
artembilan: I see that it isn't possible (yet) to determine batching or record-based target listener at runtime, although all that `MessagingMessageListenerAdapter.determineInferredType()` says opposite and looks like we can somehow find a combination of both `BatchMessagingMessageListenerAdapter` and `RecordMessagingMessageListenerAdapter`.

But let's, at least, add `batching = true/false` logic into the `@KafkaListener` annootation!
WDYT?

spring-kafka/src/main/java/org/springframework/kafka/config/MethodKafkaListenerEndpoint.java
```diff
@@ -114,10 +115,26 @@ protected HandlerAdapter configureListenerAdapter(MessagingMessageListenerAdapte
...
 	 * Create an empty {@link MessagingMessageListenerAdapter} instance.
+	 * @param messageConverter the converter (may be null).
 	 * @return the {@link MessagingMessageListenerAdapter} instance.
 	 */
-	protected MessagingMessageListenerAdapter<K, V> createMessageListenerInstance() {
-		return new MessagingMessageListenerAdapter<K, V>(this.method);
+	protected MessagingMessageListenerAdapter<K, V> createMessageListenerInstance(MessageConverter messageConverter) {
+		if (isBatchListener()) {
+			BatchMessagingMessageListenerAdapter<K, V> messageListener = new BatchMessagingMessageListenerAdapter<K, V>(
+					this.method);
```
artembilan: I still try to understand why we can't determine batching here at runtime basing on the `isConsumerRecordList` and `isMessageList` inferred states...

spring-kafka/src/main/java/org/springframework/kafka/config/SimpleKafkaListenerEndpoint.java
```diff
@@ -1,71 +0,0 @@
...
- * A {@link KafkaListenerEndpoint} simply providing the {@link MessageListener} to
- * invoke to process an incoming message for this endpoint.
- *
- * @param <K> the key type.
- * @param <V> the value type.
- *
- * @author Stephane Nicoll
- * @author Gary Russell
- */
-public class SimpleKafkaListenerEndpoint<K, V> extends AbstractKafkaListenerEndpoint<K, V> {
```
artembilan: What's wrong with this class? 
Why don't deprecate first of all?

spring-kafka/src/main/java/org/springframework/kafka/config/MethodKafkaListenerEndpoint.java
garyrussell: Because `List<ConsumerRecord<?,?>` works today (with only one entry of course).

How would we know whether `List<String>` is for a batch listener or not?

`List<String>` might the be the payload of a single message (deserialized JSON) so we need a flag (unless we add it to the annotation as you suggested). I suppose adding it to the annotation would mean the same factory could be used for batches and not batches. 

But I am inclined to keep it simple; we can always enhance it later if the community insists.

spring-kafka/src/main/java/org/springframework/kafka/config/SimpleKafkaListenerEndpoint.java
garyrussell: It's not used anywhere, not documented, and I don't see it adding any real value. Why go through conversion from `ConsumerRecord` to `Message<?>` and then immediately discard the conversion?

spring-kafka/src/main/java/org/springframework/kafka/config/MethodKafkaListenerEndpoint.java
artembilan: That sounds reasonable and like a gold compromise.
Let's leave as is! Even without annotation changes.

garyrussell: 👍

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
```diff
@@ -0,0 +1,147 @@
...
+	public List<ProducerRecord<?, ?>> fromMessage(Message<?> message, String defaultTopic) {
+		throw new UnsupportedOperationException();
+	}
+
+	/**
+	 * Subclasses can convert the payload; by default, it's sent unchanged to Kafka.
+	 * @param message the message.
+	 * @return the payload.
+	 */
+	protected Object convertPayload(Message<?> message) {
```
artembilan: Unused method.
Remove?

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	/**
+	 * Subclasses can convert the value; by default, it's returned as provided by Kafka.
+	 * @param record the record.
+	 * @param type the required type.
+	 * @return the value.
+	 */
+	protected Object extractAndConvertValue(ConsumerRecord<?, ?> record, Type type) {
+		return record.value() == null ? KafkaNull.INSTANCE : record.value();
```
artembilan: The `type` is out of use, but I guess it is just a copy/paste from the `MessagingMessageConverter`.
I wonder if we could use `ConversionService` to return the value for the desired `type`...

OTOH since we are here in the `batching` can't we bypass `KafkaNull.INSTANCE` and just return `null` as is for the `payloads` item?

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	@SuppressWarnings("serial")
+	private static class KafkaMessageHeaders extends MessageHeaders {
```
artembilan: Looks like this class can be moved to the package level and reused from both `MessageConverter` impls.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/BatchMessagingMessageListenerAdapter.java
artembilan: Hasn't been addressed yet.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
garyrussell: I think we should be consistent - if someone configures `Message<?> m` for not-batching, he'll get a `KafkaNull`.

Subclasses might want to use `type`.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/BatchMessagingMessageListenerAdapter.java
garyrussell: Ah - I added it to the converter, but not here.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/KafkaMessageHeaders.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Map;
+
+import org.springframework.messaging.MessageHeaders;
+
+/**
+ * Overload of message headers configurable for adding id and timestamp headers.
+ *
+ * @author Marius Bogoevici
+ * @author Gary Russell
+ * @since 5.0
```
artembilan: `1.1` :smile: 

That's all. Merging...

garyrussell: Doh - I need project-specific settings, like I have in spring-amqp.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/7d5cad274ebf3861c06089cd12149b6e5483ff4b

Issue #185, opened 2016-09-21
Suppress Acknowledgment for Non-Manual Modes


garyrussell:
While we don't allow acking when not manual...

	Assert.state(ListenerConsumer.this.isAnyManualAck,
				"A manual ackmode is required for an acknowledging listener");

...it would be better to not provide an Ack in that case (`null`) when the listener is an acking listener - the listener adapter implements both interfaces so always gets an ack,

PR #187, opened 2016-09-21
GH-185: Dont Populate Ack for Non-Manual Ack Modes - 1.0.x BackPort

garyrussell:
Fixes: https://github.com/spring-projects/spring-kafka/issues/185

If a listener is an `AcknowledgingMessageListener`, do not provide an ack if
the container is not configured for manual acks.


garyrussell:
@artembilan - we need to release 1.1.1 and 1.0.4 when these are merged.

PR #197, opened 2016-10-19
GH-196: Revert to assertJ 2.5.0 - IO Compatibility

### LINKED ISSUE #196: Platform Brussels build fails with NoSuchMethodErrors due to use of AssertJ 3.4
Spring Boot, and as a result the Platform, has dependency management for AssertJ. Due to Boot 1.5 and Platform Brussels still supporting Java 6 and 7, it uses AssertJ 2.x. This results in a no such method error in Spring Kafka's tests. For example:

```
java.lang.NoSuchMethodError: org.assertj.core.api.AbstractIterableAssert.allMatch(Ljava/util/function/Predicate;)Lorg/assertj/core/api/AbstractIterableAssert;
	at org.springframework.kafka.listener.ConcurrentMessageListenerContainerTests.testAfterListenCommit(ConcurrentMessageListenerContainerTests.java:234)
```

There are (at least) a couple of ways to fix this:

1. Downgrade Assert J 2.5 and change the assertions
2. Change you build.gradle to override the AssertJ version in to springIoTestRuntime configuration

I have a slight preference for 1 as it's the safer solution, but it's also more work and stops you from using any new goodies in Assert J 3.x. What would you prefer?

garyrussell:
Fixes #196

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,7 +76,7 @@ subprojects { subproject ->
 	}
 
 	ext {
-		assertjVersion = '3.4.1'
+		assertjVersion = '2.5.0'
 		hamcrestVersion = '1.3'
 		jacksonVersion = '2.6.7'
 		junitVersion = '4.12'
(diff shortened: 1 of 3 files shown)


artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/79fceba98bc683f82b2f52076f570596bfc318e7

Issue #199, opened 2016-11-01
Rebalance triggered under load


martindam:
Today I discovered an issue with the KafkaMessageListenerContainer (running 1.0.2) and manual acknowledgement that triggers rebalances and prevents offsets from being committed. 

What I believe is happening is the following sequence:
1) A batch is received on thread A and sends the batch to thread B and goes back to polling
2) Thread B invokes the listener for each record
3) When the record is processed the record is acknowledged which adds the offset to the ack queue
4) The batch is now fully processed and because it is manual ack, the consumer is woken up
5) Thread A throws the WakeupException and goes to commit the offset
6) Meanwhile thread B finished another batch and now calls wakeup again. This interrupts the commitSync() and the offset is not committed. 
7) Thread A goes back to polling, but now another batch is finished so the consumer is woken up again
8) and so on

Because the consumer is woken up all the time, it does not get to send its heartbeat nor commit its offset. This sequence of events can continue until the session timeouts and a rebalance happens. This is a fairly complicated sequence in order for an error to occur, but after have been running this in staging for some time, we see it happen fairly often with our workload. The image below is the "Time since last heartbeat" for the consumer which should be below 30s (the red area) and preferable below 10s (the yellow area). 

![image](https://cloud.githubusercontent.com/assets/10091249/19877382/f204c3b8-9f9b-11e6-98ef-1e2c03262cd2.png)

When I changed the consumer to remove the wakeup on line 623 (https://github.com/spring-projects/spring-kafka/blob/9ec3a780155e73a47aaf93807a51e4718523e777/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L623) the heartbeats succeeds, the commits are performed correctly and the overall performance is improved. Please have a look and get back to me - for now, I will be running it with my fix as we had 4-6 rebalances per hour per consumer before.

garyrussell:
Thanks for reporting this - I think we need to do a best-effort commit - i.e. perhaps a boolean indicating the consumer is in `poll()` and only wakeup if it's true. Not entirely reliable (we might miss a wake up in the small race condition between polls, but I think it's better than waiting for the poll to end to do the commits.

martindam:
I think that it should be an option whether the wakeup should be used or not. Using it will have a big impact of performance and is frankly not needed unless you are using an immediate manual commit.

garyrussell:
[...]

PR #200, opened 2016-11-03
GH-199: Do not wakeup consumer after queueing ack

martindam:
Waking up the consumer too often will trigger a rebalance and messages are not actually committed

Please patch this on the 1.0.x branch as well for kafka 0.9 users

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -772,9 +772,6 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 					if (!this.isAnyManualAck && !this.autoCommit) {
 						this.acks.add(record);
 					}
-					if (this.isRecordAck) {
-						this.consumer.wakeup();
-					}
 				}
 				catch (Exception e) {
 					if (this.containerProperties.isAckOnError() && !this.autoCommit) {
@@ -791,9 +788,6 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 						throw er;
 					}
 				}
-			}
-			if (this.isManualAck || this.isBatchAck) {
-				this.consumer.wakeup();
 			}
 		}
 
@@ -994,9 +988,6 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 							else {
 								ListenerConsumer.this.logger.debug("Interrupt ignored");
 							}
-						}
-						if (!ListenerConsumer.this.isManualImmediateAck && this.active) {
-							ListenerConsumer.this.consumer.wakeup();
 						}
 					}
 				}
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@martindam Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=200)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=200) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

pivotal-issuemaster:
@martindam Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=200)!

artembilan:
Hello, @martindam !

To be honest, I'm fine with your solution. It really looks better to have a short poll period instead of wake up, rather than don't have heartbeats and finish with always rebalance.

We thought with @garyrussell to have something like `immediate` flag to do existing logic for BATCH and RECORD modes. But that looks like switching on that option for your workload will lead you to the same bad system state. So, for me that smells weak and bug-aware anyway.

So, my position is as +1 to drop wakeup() where you do that already.
But in addition I'd add commitOffsets() call after `poll()` invocation one more time. Just because during `poll()` blocking we may have added new offsets for commit.

Although, tell me, please, if you receive the same records during the next poll, while we haven't committed yet because we don't wakeup() any more.

@garyrussell , your thoughts ?

martindam:
As long as you do not trigger a rebalance, you will not get duplicate messages from the poll() independent of you committed the offset or not. Committing the offset is committing it externally, but the consumer keeps track of the next offset to read internally anyway. 

You could add the commitOffsets() after the poll(), but you just did that before the poll and this loop should continue so I don't see the need. Unless you do auto commit you are just passing the batch of records to the queue and then calling the commitOffsets() so I don't think it is needed.

artembilan:
Thank you, @martindam , for the explanation!

Several more points to sanctify:

When we have several consumers in the same group, only one of them process single partition. 
And when happens rebalance a new partition consumer consults stored offset to start with. 
If that correct, then I won't worry a lot about non-committed offsets, since we always can reply them again and `filter` in a new consumer.

martindam:
Yes that is correct, after a rebalance it will clear its internal offsets and consult the external stored offsets and start from there even if the same consumer is reassigned a partition it already had.

garyrussell:
LGTM; thanks @martindam

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/c245cbbe8905af1334ce268ebb05420c3afb1265 and cherry-picked to 1.0.x.

Thank you for valuable contribution, @martindam ! As always :smile:

martindam:
You are welcome. What are your plans for releasing 1.0.5.RELEASE?

artembilan:
Well, @martindam ,

Frankly we have only one commit (your) there since the last `1.0.4` release.
Or everybody has moved to `1.1` already with Kafka 0.10 support, or they use some other configuration or don't use Spring Kafka at all :smile: 

Although we are flexible enough and can release `1.0.5` anyway. Even if it is useful only for you.

@garyrussell , you thoughts ?

garyrussell:
I have no issues making a 1.0.5 release, even with just 1 commit.

artembilan:
Good. Then it is fine with me to release just today!

If there are any objections, you have 10 min until I start the release process :smile:

artembilan:
@martindam ,

The version `1.0.5` has been released: https://repo.spring.io/release/org/springframework/kafka/spring-kafka/1.0.5.RELEASE/

Enjoy. And thank you for contribution!

PR #205, opened 2016-11-18
Missing item in documentation

ltalhouarne:
Channel 'received' was missing in the documentation even though it was referenced through a method call (line 158).

### DIFF
--- a/src/reference/asciidoc/si-kafka.adoc
+++ b/src/reference/asciidoc/si-kafka.adoc
@@ -173,6 +173,11 @@ public ConsumerFactory<String, String> consumerFactory() {
     // set more properties
     return new DefaultKafkaConsumerFactory<>(props);
 }
+
+@Bean
+public PollableChannel received() {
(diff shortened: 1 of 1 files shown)


artembilan:
Thank you for the contribution, of course, but to be honest I'd prefer to leave the code snippet in the Docs as short as possible and without any extra beans from other area.

Here we would like to concentrate an attention exactly on the Spring Integration adapters for Spring Kafka.
Everything rest assumes that you are already familiar with Spring Integration principles and something like `setOutputChannel(...)` should be obvious by its name.

Any other strong reason to merge this anyway?

garyrussell:
I don't mind either way; but the XML snippet is a bit weird too - `nullChannel`.

garyrussell:
Also, I am not sure we want to indicate a `PollableChannel` is a good choice here, since it could cause an OOM condition and message loss.

ltalhouarne:
That makes sense! Ill close the PR.

Issue #217, opened 2016-12-14
Incorrect Configuration in the SI Section


garyrussell:
http://stackoverflow.com/questions/41137034/spring-integration-kafka-message-driven-channel-adapter-receive-message

PR #218, opened 2016-12-14
GH-217: Fix obsolete docs and race condition

artembilan:
Fixes: GH-217 (https://github.com/spring-projects/spring-kafka/issues/217)

Now we have to provide `ContainerProperties` object instead of direct `topics` for the `KafkaMessageListenerContainer`

Fix race condition in the `KafkaMessageListenerContainerTests` where a `commitSync()` can be performed before spying a consumer.

Since `consumer` is available only after `container.start()` and several barriers to ensure that `consumer` is created but isn't executed yet to let the `spy` be ready for the subsequent `commitSync()` during `run()`


artembilan:
@garyrussell , not sure if this will fix https://build.spring.io/browse/SK-SON-285/ but at least there is one more race condition covered.
Maybe let's merge?
Thanks

Issue #232, opened 2017-02-03
KafkaTemplate: Don't Cache the Producer


garyrussell:
Since the factory serves up a singleton producer, the template should fetch it each time rather than caching it, so if the producer is closed, it will be re-created on demand.

See http://stackoverflow.com/questions/42003192/spring-kafka-shutting-down-producer-gracefully-during-runtime/42004516#comment71204987_42004516

PR #233, opened 2017-02-03
GH-232: Do not cache produce in the KafkaTemplate

artembilan:
Fixes GH-232  (https://github.com/spring-projects/spring-kafka/issues/232)

 Since the factory serves up a singleton producer, the template should fetch it each time rather than caching it, so if the producer is closed, it will be re-created on demand.

**Cherry-pick to 1.1.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


garyrussell:
We should do a close after each operation too (even though it's a no-op for the consumer served up by the default factory).

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -180,22 +179,46 @@ public void setMessageConverter(RecordMessageConverter messageConverter) {
...
+		try {
+			return callback.doInKafka(producer);
+		}
+		finally {
+			producer.close();
+		}
 	}
 
 	@Override
 	public void flush() {
```
garyrussell: We should add javadoc that this only makes sense if the `ProducerFactory` serves up a singleton `Producer`.

I will do so on merge.

artembilan: Yeah.... This caused some doubts for me as well.

Yep! Java doc would be great compromise
Thank you!

garyrussell:
Merged as 41d8c008b64cd30041069012320fa09f5503e386 and cherry-picked as c5bf0681c9f656ebd7682be930da7e5897f5bdda

Issue #230, opened 2017-02-02
Cannot make Batch receiver to automatically commit offsets.


dkaukov:
Hi,

This is my factory:
```
@Bean
	public ConcurrentKafkaListenerContainerFactory<IdAndRidKey, String> guestListenerContainerFactory() {
		ConcurrentKafkaListenerContainerFactory<IdAndRidKey, String> factory = new ConcurrentKafkaListenerContainerFactory<>();
		factory.setBatchListener(true);
		factory.setConcurrency(threadCount);
		factory.setConsumerFactory(new DefaultKafkaConsumerFactory<>(consumerProperties()));

		ContainerProperties containerProperties = factory.getContainerProperties();
		containerProperties.setAckMode(AbstractMessageListenerContainer.AckMode.BATCH);
		containerProperties.setAckOnError(true);
		containerProperties.setSyncCommits(true);

		return factory;
	}
```

Receiver:
```
@KafkaListener(topicPattern = "${kafka.topics.guests}", containerFactory = "guestListenerContainerFactory")
	public void receiveGuest(List<String> records) {
		log.debug("Kafka consumer processed {} record.", records.size());
	}
```

Kaffka property:
```
consumerProperties.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, false);
```

The problem is, that the offsets never got commited by this receiver. I did quick debugging and looks like KafkaMessageListenerContainer  creating BatchAcknowledgingMessageListener instead of BatchMessageListener here: 
https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L331

And ivokeBatchListener only add acks for BatchMessageListener:
https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L728

BTW: invokeRecordListener adds Acks regardless of the type of listener:
https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L772
This is probably why it works fine for Record listeners.

PR #235, opened 2017-02-03
GH-230: Commit batch acks properly

artembilan:
Fixes GH-230  (https://github.com/spring-projects/spring-kafka/issues/230)

The offsets never got commited by when we don't use MANUAL mode and rely on the `(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, false)`

* Move `this.acks.put(record)` outside of the `else` branch for the regular listener.
This way the acks will be placed for the commits even if we use `batchListener` and don't use `isAnyManualAck` mode-

**Cherry-pick to 1.1.x**


garyrussell:
Cherry-picked as bc8092216cbc887177cbf2e8017fd643f044069c

PR #238, opened 2017-02-09
Add KStreams support

artembilan:


artembilan:
@mbogoevici , @garyrussell ,

Sorry for plagiarism 😉 , I took a bit your old work and represented it via a bit different angle.

The main idea that there can be many different topologies and we would like to control their lifecycle independently.

Plus bear in mind that we can `join` streams or build `KTable`, so, I decided that expose builder result as a `@Bean` isn't bad idea, when in the end we can have many of them and `join` via Dependency Injection!

Won't mind if you just drop my effort.
Anyway that was useful exercise for me 😄

mbogoevici:
@artembilan great - thanks for doing this. I love the idea of exposing KStreams and KTables as beans!

artembilan:
Good. 
Thank you!

As you see i "hid" `KafkaStreams` lifecycle into the `KStreamBuilderFactoryBean` meanwhile it exposes exactly `KStreamBuilder`. Almost for nothing from the Spring perspective, but for the control of the future topoligies via its injection.

The same is applied for the `KStream` and `KTable` when they are represented as beans.

For the future I have some idea like some Kafka Stream processor implementation for the Spring Integration gateway proxy, when we can process Kafka data via SI flow and come back.

Or maybe we can come up with something like `toKStream(IntegrationFlow)` to be able to `.join()` to the existing `KStream`.

In all those cases we definitely need somehow be able to use dependency injection from the `KStream` definition.

mbogoevici:
Yes, hiding KafkaStreams is a good idea - I can say that because I've done something similar in my KStream binder POC ;) - https://github.com/mbogoevici/spring-cloud-stream-binder-kstream/blob/master/spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/kstream/config/KStreamLifecycle.java - but combining that with the FactoryBean from the other POC is an even better idea. 

A gateway for KStream would be very nice - I guess we can leverage that into the binder. With something like this, what we need on the binder side is to be able to abide by the conventions of `spring.cloud.stream.bindings.[input|output].etc` when creating the `KStream` references via the bindable proxy (destinations, contentType transformation, etc).

artembilan:
No, we can't build `KafkaStreams` just for raw `TopologyBuilder`.
We really have to build all our topologices in advance and only in the end `KafkaStreams` and start it.

Seems for me an oversight in the Apache Kafka API.
Since they provide `start()` API it really would be much better to parse `TopologyBuilder` already there.
No-op right now. 😢 
So, I'm creating `KafkaStreams` in the `KStreamBuilderFactoryBean`.

Another observation - once closed `KafkaStreams` can't be started again.
So, I just recreate it in the next `KStreamBuilderFactoryBean.start()`.

mbogoevici:
Yes, that's what I was mentioning on our call yesterday - any further invocations to the `TopologyBuilder` after being passed to the `KafkaStreams` constructor are ignored - `KafkaStreams` needs to be created after the topology is built.

spring-kafka/src/main/java/org/springframework/kafka/core/KStreamBuilderFactoryBean.java
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * An {@link AbstractFactoryBean} for the {@link KStreamBuilder} instance
+ * and lifecycle control for the internal {@link KafkaStreams} instance.
+ *
+ * @author Artem Bilan
+ *
+ * @since 2.0
+ */
+public class KStreamBuilderFactoryBean extends AbstractFactoryBean<KStreamBuilder> implements SmartLifecycle {
```
artembilan: Another trick here might be like `KafkaStreamBuilder extends KStreamBuilder implements SmartLifecycle` to make it easier for use in JavaConfig:
```
@Bean
KafkaStreamBuilder streamBuilder() {
    ...
}

@Bean
KStream<Integer, String> kStream() {
    streamBuilder().stream(...)
    ...
```

Really nothing critical to hide until `getObject()` lately...

artembilan: After some private discussion decided do not follow with extension practice because that may lead to unexpected breaking changes from the target Apache Kafka API.

artembilan:
I did just only a small effort for the automatic `KStreamBuilderFactoryBean` registration when we have a deafult `KafkaStreams` bean. The solution doesn't looks so far instead of just simple bean definition like:
```
@Bean
public KStreamBuilderFactoryBean kStreamBuilder() {
	return new KStreamBuilderFactoryBean(kStreamsConfigs());
}
```
So, I even not sure if we still should pursue the idea for default registration by some annotation option...

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaBootstrapConfiguration.java
```diff
@@ -53,4 +65,27 @@ public KafkaListenerEndpointRegistry defaultKafkaListenerEndpointRegistry() {
...
+
+		@Override
+		public void setBeanClassLoader(ClassLoader classLoader) {
+			this.classLoader = classLoader;
+		}
+
+		@Override
+		public void registerBeanDefinitions(AnnotationMetadata importingClassMetadata,
+				BeanDefinitionRegistry registry) {
+			if (ClassUtils.isPresent(org.apache.kafka.streams.kstream.KStreamBuilder.class.getName(),
```
mbogoevici: I suppose we should use the class name as a String, or else this would blow up if the class is not on the classpath?

markfisher: could `@ConditionalOnClass` be used on the inner class?

mbogoevici: `ConditionalOnClass` is a Boot thing, unfortunately.

markfisher: oh yeah... sorry. I've had the boots on so long I forget.

I guess you could implement a `Condition`, but relying on a String is simpler.

artembilan: Yeah... Really would be better to use just raw name.
I thought it would turn to the string constant, but not.
This is decompiled class:
```
 public void registerBeanDefinitions(AnnotationMetadata importingClassMetadata, BeanDefinitionRegistry registry) {
        if(ClassUtils.isPresent(KStreamBuilder.class.getName(), this.classLoader) && registry.containsBeanDefinition("defaultKafkaStreamsConfig")) {
```
So, it is even worse - the `import` is there to load a class.

I tried to implement `Condition`, but it doesn't see target beans somehow. Only a `@Configuration` on annotation and some common `BPP`s 😢 
Not sure how `@ConditionalOnMissingBean` works in Boot then...

mbogoevici: `ConditionalOnMissingBean` and `ConditionalOnBean` should only be used on Boot autoconfigurations b/c of their lifecycle. In any case, it feels overkill atm.

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableKafka.java
```diff
@@ -249,6 +250,6 @@
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-@Import(KafkaBootstrapConfiguration.class)
+@Import({KafkaBootstrapConfiguration.class, KafkaBootstrapConfiguration.KStreamsRegistrar.class})
```
mbogoevici: Maybe we should have a separate annotation for `@EnableKafkaStreams`? - it will rarely be a case that you'd have both a Kafka Stream and producer/consumers in the same app. Also, for regular Kafka apps I guess this also starts a Stream with an empty topology (probably wouldn't do much but I guess we can avoid it).

mbogoevici:
@artembilan I've added two comments - regarding the use of the class and the annotation. Re: https://github.com/spring-projects/spring-kafka/pull/238#issuecomment-283998346 - I suppose we can ditch that, although to me `@EnableKafkaStreams` feels much nicer (and lower barrier) than creating the FactoryBean.

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableKafka.java
artembilan: I thought about that in the beginning and that was our with you mind storm decision on the meeting last week.

But in the end I decided that  `@EnableKafka` is general annotation which says that the Spring Kafka infrastructure is switched on - feel free to to spill the code!

This our `enable KStream` case doesn't provide a lot of gain and it is conditional anyway.

I think it might be better to reconsider `KafkaListenerAnnotationBeanPostProcessor` registration if there is no `kafkaListenerContainerFactory`, but still have that very `@EnableKafka` and that's all.

mbogoevici: It does a little bit too much, that's all - and IMO it's a bit of an issue that we create a `KafkaStream` and start it transparently without the user requesting it (even if apparently it doesn't do anything now, that might change). Is it at least possible to enquire whether the topology is empty or not? 

If we do it this way it is a little hard to differentiate which of the two use cases the user actually wanted (plain `Kafka` app or `KafkaStreams` app). The counterargument is that none of that would happen if the Kafka Streams library is not on the CP, so YMMV - in my head, I still think that they're two different use cases.

artembilan:
Pushed `@EnableKStreams` solution.
Let me know if that satisfies your expectations.
Thanks

spring-kafka/src/test/java/org/springframework/kafka/kstreams/KStreamsTests.java
```diff
@@ -0,0 +1,182 @@
...
+
+		assertThat(result).isNotNull();
+
+		assertThat(result).isEqualTo(payload.toUpperCase() + payload2.toUpperCase());
+	}
+
+
+	@Configuration
+	@EnableKafka
+	@EnableKStreams
```
mbogoevici: I suppose we need to enable both here just because the test is using both clients.

artembilan: Yes, there is a `@KafkaListener` configuration to poll message from the topic after KStream "magic".

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableKStreams.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @author Artem Bilan
+ *
+ * @since 2.0
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Import(KStreamDefaultConfiguration.class)
+public @interface EnableKStreams {
```
mbogoevici: IMO the name here should be either `@EnableKafkaStreams` or `@EnableKStream`. Maybe the former would be more correct - given that `KStream` is the API type and 'Kafka Streams' is the official name of the feature.

spring-kafka/src/test/java/org/springframework/kafka/kstreams/KStreamsTests.java
mbogoevici: Cool, thanks - it was just a clarification question.

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableKStreams.java
artembilan: Make sense.
Thank you!

artembilan:
Addressed you concerns, @mbogoevici .
Meanwhile you are in review this last (?) round, I'm doing docs on the matter.

mbogoevici:
@artembilan Thanks - the code LGTM - I'll take a look at the docs and merge.

mbogoevici:
@artembilan some minor docs polishing here: https://github.com/mbogoevici/spring-kafka/commit/4ccce1b5806a624ee7a2ef3bdcb13b40c96e1ace

artembilan:
LGTM
Thank you @mbogoevici !

Those English articles are still my bottleneck 😄 

Ship it, by the way!

mbogoevici:
Rebased, squashed and merged as https://github.com/spring-projects/spring-kafka/commit/aa677a483b03598bceddbb86afceddcbecaf4f8b and cherry-picked to 1.1.x as https://github.com/spring-projects/spring-kafka/commit/ebbe00a3e3531419da48bec6c354713ae3f71a5c

Issue #227, opened 2017-01-22
add a KafkaTemplate send() method with argument for timestamp


dsw9742:
Since version 0.10.0.1, Apache Kafka offers a ProducerRecord constructor method that accepts a Long timestamp value. Reference JavaDocs presently at [https://kafka.apache.org/0100/javadoc/org/apache/kafka/clients/producer/ProducerRecord.html](https://kafka.apache.org/0100/javadoc/org/apache/kafka/clients/producer/ProducerRecord.html#ProducerRecord(java.lang.String,%20java.lang.Integer,%20java.lang.Long,%20K,%20V)) .

KafkaOperations and KafkaTemplate should offer additional send methods to support passing in a timestamp value, which could then be relayed to the ProducerRecord constructor.

artembilan:
`MessagingMessageConverter.fromMessage()` should populate this option as well.
Not sure if `MessageHeaders.TIMESTAMP` is sufficient, but at least we could add a new `KafkaHeaders` constant, since `ConsumerRecord` provides this property, too.

artembilan:
Fixed via #241

PR #241, opened 2017-02-21
Support for timestamps in Kafka Template

bijukunjummen:
@garyrussell, @artembilan - this is a Work in progress PR for #227, do provide any feedback and I will add it in. I will now work on @artembilan's comment about `MessagingMessageConverter.fromMessage()`

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
@@ -17,6 +17,7 @@
 package org.springframework.kafka.test.assertj;
 
 import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.common.record.TimestampType;
 import org.assertj.core.api.Condition;
 
 /**
@@ -47,6 +48,15 @@ public final class KafkaConditions {
 	 */
 	public static <V> Condition<ConsumerRecord<?, V>> value(V value) {
 		return new ConsumerRecordValueCondition<>(value);
+	}
+
+	/**
+	 * @param type the type of timestamp
+	 * @param value the timestamp.
+	 * @return a Condition that matches the timestamp value in a consumer record.
+	 */
+	public static Condition<ConsumerRecord<?, ?>> timestamp(TimestampType type, long value) {
+		return new ConsumerRecordTimestampCondition(type, value);
 	}
 
 	/**
@@ -86,6 +96,25 @@ public final class KafkaConditions {
 		@Override
 		public boolean matches(ConsumerRecord<?, V> value) {
 			return value != null && value.value().equals(this.payload);
+		}
+
+	}
+
+	public static class ConsumerRecordTimestampCondition extends Condition<ConsumerRecord<?, ?>> {
+
+		private TimestampType type;
+		private final long ts;
+
+		public ConsumerRecordTimestampCondition(TimestampType type, long ts) {
+			super("a ConsumerRecord with timestamp of type: " + type + " and timestamp: " + ts);
+			this.type = type;
+			this.ts = ts;
+		}
+
+		@Override
+		public boolean matches(ConsumerRecord<?, ?> value) {
+			return value != null &&
+					(value.timestampType() == this.type && value.timestamp() == this.ts);
 		}
 
 	}
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -56,7 +56,7 @@ public final class KafkaConditions {
 	 * @return a Condition that matches the timestamp value in a consumer record.
 	 */
 	public static Condition<ConsumerRecord<?, ?>> timestamp(long value) {
-		return new ConsumerRecordTimestampCondition(TimestampType.CREATE_TIME, value);
+		return timestamp(TimestampType.CREATE_TIME, value);
 	}
 
 	/**
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import org.assertj.core.api.Condition;
  *
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Biju Kunjummen
  */
 public final class KafkaConditions {
 
@@ -48,6 +49,14 @@ public final class KafkaConditions {
 	 */
 	public static <V> Condition<ConsumerRecord<?, V>> value(V value) {
 		return new ConsumerRecordValueCondition<>(value);
+	}
+
+	/**
+	 * @param value the timestamp.
+	 * @return a Condition that matches the timestamp value in a consumer record.
+	 */
+	public static Condition<ConsumerRecord<?, ?>> timestamp(long value) {
+		return new ConsumerRecordTimestampCondition(TimestampType.CREATE_TIME, value);
 	}
 
 	/**
@@ -102,7 +111,7 @@ public final class KafkaConditions {
 
 	public static class ConsumerRecordTimestampCondition extends Condition<ConsumerRecord<?, ?>> {
 
-		private TimestampType type;
+		private final TimestampType type;
 		private final long ts;
 
 		public ConsumerRecordTimestampCondition(TimestampType type, long ts) {
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
@@ -64,6 +64,16 @@ public interface KafkaOperations<K, V> {
 	ListenableFuture<SendResult<K, V>> sendDefault(int partition, K key, V data);
 
 	/**
+	 * Send the data to the default topic with the provided key and partition.
+	 * @param partition the partition.
+	 * @param timestamp the timestamp of the record.
+	 * @param key the key.
+	 * @param data the data.
+	 * @return a Future for the {@link SendResult}.
+	 */
+	ListenableFuture<SendResult<K, V>> sendDefault(int partition, Long timestamp, K key, V data);
+
+	/**
 	 * Send the data to the provided topic with no key or partition.
 	 * @param topic the topic.
 	 * @param data The data.
@@ -98,6 +108,17 @@ public interface KafkaOperations<K, V> {
 	 * @return a Future for the {@link SendResult}.
 	 */
 	ListenableFuture<SendResult<K, V>> send(String topic, int partition, K key, V data);
+
+	/**
+	 * Send the data to the provided topic with the provided key and partition.
+	 * @param topic the topic.
+	 * @param partition the partition.
+	 * @param timestamp the timestamp of the record.
+	 * @param key the key.
+	 * @param data the data.
+	 * @return a Future for the {@link SendResult}.
+	 */
+	ListenableFuture<SendResult<K, V>> send(String topic, int partition, Long timestamp, K key, V data);
 
 	/**
 	 * Send a message with routing information in message headers. The message payload
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -34,8 +34,17 @@ import org.springframework.util.concurrent.ListenableFuture;
  * @param <K> the key type.
  * @param <V> the value type.
  *
+ * If the Kafka topic is set with {@link org.apache.kafka.common.record.TimestampType#CREATE_TIME CreateTime}
+ * all send operations will use the user provided time if provided, else
+ * {@link org.apache.kafka.clients.producer.KafkaProducer} will generate one
+ *
+ * If the topic is set with {@link org.apache.kafka.common.record.TimestampType#LOG_APPEND_TIME LogAppendTime}
+ * then the user provided timestamp will be ignored and instead will be the
+ * Kafka broker local time when the message is appended
+ *
  * @author Marius Bogoevici
  * @author Gary Russell
+ * @author Biju Kunjummen
  */
 public interface KafkaOperations<K, V> {
 
@@ -61,7 +70,7 @@ public interface KafkaOperations<K, V> {
 	 * @param data the data.
 	 * @return a Future for the {@link SendResult}.
 	 */
-	ListenableFuture<SendResult<K, V>> sendDefault(int partition, K key, V data);
+	ListenableFuture<SendResult<K, V>> sendDefault(Integer partition, K key, V data);
 
 	/**
 	 * Send the data to the default topic with the provided key and partition.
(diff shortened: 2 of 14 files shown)


bijukunjummen:
Working on fixing checkstyle issues now

spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
```diff
@@ -50,6 +51,15 @@ private KafkaConditions() {
 	}
 
 	/**
+	 * @param type the type of timestamp
+	 * @param value the timestamp.
+	 * @return a Condition that matches the timestamp value in a consumer record.
+	 */
+	public static Condition<ConsumerRecord<?, ?>> timestamp(TimestampType type, long value) {
```
artembilan: I think we can add an overridden variant without `TimestampType`.
By default it is `message.timestamp.type=CreateTime`: https://cwiki.apache.org/confluence/display/KAFKA/KIP-32+-+Add+timestamps+to+Kafka+message#KIP-32-AddtimestampstoKafkamessage-Brokerconfigurationchange

But yeah, we can delegated to this one!

Also your name to the `@author`, plus we prefer to update Copyright to the actual year.

And I guess similar trick should be done for the `KafkaMatchers`.

That is good catch, @bijukunjummen , BTW!

We might have missed already a lot similar Kafka properties for testing matchers and conditions 😄

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
```diff
@@ -64,6 +64,16 @@
...
 
 	/**
+	 * Send the data to the default topic with the provided key and partition.
+	 * @param partition the partition.
+	 * @param timestamp the timestamp of the record.
+	 * @param key the key.
+	 * @param data the data.
+	 * @return a Future for the {@link SendResult}.
+	 */
+	ListenableFuture<SendResult<K, V>> sendDefault(int partition, Long timestamp, K key, V data);
```
artembilan: Your name to the `@author` list and copyright.

The general JavaDoc on the method should mention `timestamp` as well.

I see from where you have picked up the `Long` type wrapper, not primitive.
And it looks for me to be consistent we should do the same for `partition` argument.
But that will be already breaking change if we are going to backport.

Although, we won't backport this interface changes any way.
WDYT?

spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
```diff
@@ -132,6 +172,46 @@ public void testTemplate() throws Exception {
...
+
+		ConsumerRecord<Integer, String> r2 = KafkaTestUtils.getSingleRecord(consumer, INT_KEY_TOPIC);
+		assertThat(r2).has(value("foo-message-2"));
+		assertThat(r2).has(timestamp(TimestampType.CREATE_TIME, 1487694048615L));
+
+		MessagingMessageConverter messageConverter = new MessagingMessageConverter();
+
+		Message<?> recordToMessage = messageConverter.toMessage(r2, null, String.class);
+
+		assertThat(recordToMessage.getHeaders().get(KafkaHeaders.TIMESTAMP_TYPE)).isEqualTo("CreateTime");
```
artembilan: I think we can use `TimestampType.CREATE_TIME.name()` for comparison,
But, yeah! That was clever do not populate Kafka-specific types into the `MessageHeaders` 😄

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
bijukunjummen: Yes, got a little confused here actually @artembilan , looks like the ProducerRecord uses a Long type and the consumer the primitive long. Since we were dealing with `send()` and that creates a ProducerRecord I actually kept the wrapper type of `Long`. Missed the javadoc, will add that in.

bijukunjummen: Thinking a little more through it, I think the reason ProducerRecord timestamp is Long is to be able to mark it as null and then the system will generate it if null. With a ConsumerRecord the value will definitely be there and primitive long will work for such a scenario. Sounds like that should work for us too right?

spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
bijukunjummen: Awesome, yes, will add a variant with only timestamp and default it to `CreateTime`

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
artembilan: Ha! really, `ConsumerRecord` uses primitives for `timestamp` as well as for `partition`.
Well, I think I know why it is like that.
You can send a record with `null` `partition` and Kafka will assign one by the `key` (if present) or using round-robin algorithm otherwise.
Something similar happens for the timestamp if you send `null` - the broker assigns one using its local time.

My concern here was about inconsistency of our primitive type for the `partition` and how `ProducerRecord` expects wrapper.

So, my proposition is to change it into `Integer` as well to be consistent with Kafka API. 
Anyway this `timestamp` addition provides breaking change.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
```diff
@@ -74,6 +74,8 @@ public void setGenerateTimestamp(boolean generateTimestamp) {
 		rawHeaders.put(KafkaHeaders.RECEIVED_TOPIC, record.topic());
 		rawHeaders.put(KafkaHeaders.RECEIVED_PARTITION_ID, record.partition());
 		rawHeaders.put(KafkaHeaders.OFFSET, record.offset());
+		rawHeaders.put(KafkaHeaders.TIMESTAMP_TYPE, record.timestampType().name);
+		rawHeaders.put(KafkaHeaders.TIMESTAMP, record.timestamp());
```
artembilan: M-m-m, would you mind also taking a look into `BatchMessagingMessageConverter` for the same reason to populated these new headers for a batch of records?

artembilan:
And a general comment.

This new `timestamp` feature should be mention in the Reference Manual. I think as much as possible - on the `KafkaTemplate` and `MessageListenerContainer` as well.

Also I see that we have missed to a proper `What's new` for the current version `2.0`.
If you don't mind, @bijukunjummen , we will appreciate that kind of contribution, too!

Thank you so far! 😄

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
bijukunjummen: Yes, makes sense @artembilan, will make partition also an `Integer`. I see one problem because of this change though, there is ambiguity in these api's(if the key is an Integer type):

`ListenableFuture<SendResult<K, V>> send(String topic, Integer partition, V data);`
`ListenableFuture<SendResult<K, V>> send(String topic, K key, V data);`, 

How about eliminating `ListenableFuture<SendResult<K, V>> send(String topic, Integer partition, V data);` altogether and keeping `ListenableFuture<SendResult<K, V>> send(String topic, Integer partition, K key, V data);` - this maps with ProducerRecord constructors too

artembilan: Agree, since we are introducing the variant when you can use `null` for `partition` (via `Integer`) as you have wanted before with the `send(String topic, K key, V data)`, it should be fine to make this breaking change do not break minds in the future.

Your thoughts, @garyrussell ?

garyrussell: Sounds good, but I don't see how we can backport such a change (on the template).

artembilan: OK. We will backport only `timestamp` addition. The `partition` part should stay the same.

spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
bijukunjummen: Got mixed up between name and name() :-), updated PR coming

bijukunjummen:
All the changes should be in place now @artembilan, let me know if you need help with backporting to 1.1

spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
bijukunjummen: Looks like we are not using `KafkaMatchers` anywhere, may be safe to delete right?

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
```diff
@@ -34,8 +34,17 @@
  * @param <K> the key type.
  * @param <V> the value type.
  *
+ * If the Kafka topic is set with {@link org.apache.kafka.common.record.TimestampType#CREATE_TIME CreateTime}
+ * all send operations will use the user provided time if provided, else
+ * {@link org.apache.kafka.clients.producer.KafkaProducer} will generate one
```
artembilan: Ha! Good catch!
```
long timestamp = record.timestamp() == null ? time.milliseconds() : record.timestamp();
```
in the `KafkaProducer.doSend()`

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaders.java
```diff
@@ -67,4 +68,14 @@
...
 
+	/**
+	 * The header for holding the {@link org.apache.kafka.common.record.TimestampType type} of timestamp.
+	 */
+	public static final String TIMESTAMP_TYPE = PREFIX + "timestampType";
+
+	/**
+	 * The header for holding the timestamp of the record.
+	 */
+	public static final String TIMESTAMP = PREFIX + "timestamp";
```
artembilan: I think the clash `receive-and-send` is applied for `timestamp` as well.
So, we need one more header on the matter - `RECEIVED_TIMESTAMP`.
The same we have for `PARTITION_ID` and `TOPIC` and `KEY`.
I'm afraid you will have to redo converters logic appropriately.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -1,3 +1,11 @@
+=== What's new in 2.0
+
+==== Support for Kafka timestamps
+
+`KafkaTemplate` now supports API's to add records with timestamps. The behavior of the recorded timestamp is dependent on the
```
artembilan: There is a recommendation for makrdowns to right each sentence on its own line: http://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line.

Not critical right now, but for the future contribution.

Plus we typically make a short teaser sentences in the "What's new" with the link to an appropriate chapter with full description of the feature.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -1,3 +1,11 @@
...
+timestamp type configured on the Kafka topic. If the topic is configured to use CREATE_TIME and then the user specified timestamp will be recorded.
```
artembilan: I think these Kafka constants can be wrapped to the code.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -1,3 +1,11 @@
...
+If the topic is configured to use LOG_APPEND_TIME then the user specified timestamp will be ignored and broker will add in the local time.
+
 === What's new in 1.1 Since 1.0
```
artembilan: This What's new for `1.1` should gone to its own file and as a link from the `appendix.adoc` - `Change History`.

You can find a sample of that in Spring Integration: https://github.com/spring-projects/spring-integration/blob/master/src/reference/asciidoc/history.adoc

src/reference/asciidoc/kafka.adoc
```diff
@@ -45,7 +47,7 @@ interface ProducerCallback<K, V, T> {
 
 ----
 
-The first 3 methods require that a default topic has been provided to the template.
+The sendDefault api's require that a default topic has been provided to the template.
```
artembilan: I also would like to see new headers for the `timestamp` in the paragraph about `ListenerContainer`, more precisely in the sample code for the `@KafkaListener(id = "qux", topicPattern = "myTopic1")`

spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
artembilan: We might not use `KafkaMatchers` internally, but `spring-kafka-test` is a test utilities set exactly for end-users.
And there is no guaranty how they use AssertJ or stay with the Hamcrest.
Surely it is like you use `spring-integration-test` or well-known `spring-test` or new cool `spring-boot-test`.

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaders.java
bijukunjummen: Oh, no issues @artembilan, keep the review coming :-), will make this change and update the PR

src/reference/asciidoc/whats-new.adoc
bijukunjummen: Ah, I see, will try in an updated PR. Thanks

src/reference/asciidoc/whats-new.adoc
bijukunjummen: Sounds good, will pull check how spring-integration does it.

bijukunjummen:
If you can glance through the changes once more @artembilan, I think I have incorporated all of your review comments now.

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaders.java
```diff
@@ -67,4 +68,19 @@
...
+
+	/**
+	 * The header for holding the timestamp of the record.
```
artembilan: I think `consumer record` as opposite to `producer record` on the `TIMESTAMP`.
Just to let end-users to avoid confusing.

spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
```diff
@@ -50,6 +52,23 @@ private KafkaConditions() {
 	}
 
 	/**
+	 * @param value the timestamp.
+	 * @return a Condition that matches the timestamp value in a consumer record.
+	 */
+	public static Condition<ConsumerRecord<?, ?>> timestamp(long value) {
+		return new ConsumerRecordTimestampCondition(TimestampType.CREATE_TIME, value);
```
artembilan: I think we can delegate here the same way as you do in the `KafkaMatchers`

src/reference/asciidoc/whats-new.adoc
```diff
@@ -1,3 +1,9 @@
+=== What's new in 2.0
+
+==== Support for Kafka timestamps
+
+`KafkaTemplate` now supports API's to add and retrieve records with timestamps. See <<record-read-timestamps>> for more details.
+
 === What's new in 1.1 Since 1.0
```
artembilan: I think the Change history we can address in the separate PR/commit.

spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2017 the original author or authors.
```
artembilan: M-m-m. Sorry, if I was unclear.
We leave the first year as is and change (or add) only the current one.

bijukunjummen: Ah, okay, so something like 2016-2017?

artembilan: Exactly! Like you did here: https://github.com/spring-projects/spring-kafka/pull/241/files#diff-be360e97e458e9f0b6a7aea7890dbb8eR2

artembilan:
@bijukunjummen ,

no need to push anything else.
I'm in merge process already 😄

bijukunjummen:
Ouch, only fixed up the copyright year @artembilan :-), sorry did not realize that. Do you mind if I look over your shoulders, I am curious about how you backport.

artembilan:
NP. Just commented like this there: https://github.com/spring-projects/spring-kafka/pull/241#pullrequestreview-23245328

Re. backport... Well, I'm still not convinced that we should do that at all. 😄 

Let's come back to the topic if  there will be strict requirement!

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/8f43ad21335650a119b82061b08c7fab24ca34cc.

Thank you very much @bijukunjummen .

Looking forward for more!

PR #259, opened 2017-03-29
AckOnError Javadoc Polishing

garyrussell:
http://stackoverflow.com/questions/43085071/kafka-consumer-exception-and-offset-commits/43094537?noredirect=1#comment73287875_43094537

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/config/ContainerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/config/ContainerProperties.java
@@ -350,13 +350,17 @@ public class ContainerProperties {
 	}
 
 	/**
-	 * Set whether the container should ack messages that throw exceptions or not. This
-	 * works in conjunction with {@link #ackMode} and is effective only when auto ack is
-	 * false; it is not applicable to manual acks. When this property is set to
-	 * {@code true}, all messages handled will be acked. When set to {@code false}, acks
-	 * will be produced only for successful messages. This allows a component that starts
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/config/ContainerProperties.java
```diff
@@ -350,13 +350,17 @@ public void setIdleEventInterval(Long idleEventInterval) {
...
-	 * will be produced only for successful messages. This allows a component that starts
-	 * throwing exceptions consistently to resume from the last successfully processed
-	 * message. Manual acks will be always be applied.
+	 * Set whether or not the container should commit offsets (ack messages) where the
+	 * listener throws exceptions. This works in conjunction with {@link #ackMode} and is
+	 * effective only when the kafka property {@code enable.auto.commit} is {@code false};
+	 * it is not applicable to manual ack modes. When this property is set to {@code true}
+	 * (the default), all messages handled will have their offset committed. When set to
+	 * {@code false}, offsets will be committed only for successfully handled messages.
+	 * Manual acks will be always be applied. Bear in mind that, if the next message is
```
artembilan: Double `be`?

Maybe just: ` Manual acks will be always applied` ?

garyrussell: Hehe.... `will always be`.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/56c8d25ce66f6982188baffc73ac13526c86f5e1

Issue #262, opened 2017-03-30
BatchListener commits for every record in the batch instead of last one.


ltalhouarne:
When using the `BatchListenerAcknowledgment` and acknowledging the `Acknowledgement` given along with the batch with `MANUAL_IMMEDIATE` - it seems that the offset is committed for EACH record in the batch which is very chatty and hinders performance.

The below code in question seems to be the cause: 

```
@Override
public void acknowledge() {
	try {
		Assert.state(ListenerConsumer.this.isAnyManualAck,
		"A manual ackmode is required for an acknowledging listener");
		for (ConsumerRecord<K, V> record : this.records) {
			ListenerConsumer.this.acks.put(record);
		}
	} catch (InterruptedException e) {
		Thread.currentThread().interrupt();
		throw new KafkaException("Interrupted while queuing ack for " + this.records, e);
	}
	if (this.immediate) {
		ListenerConsumer.this.consumer.wakeup();
	}
}
```

I suggest we switch to adding the last record from the list for acknowledgment:

```
@Override
public void acknowledge() {
	try {
		Assert.state(ListenerConsumer.this.isAnyManualAck,
		"A manual ackmode is required for an acknowledging listener");
		if(!this.records.isEmtpy()) {
                     	ListenerConsumer.this.acks.put(this.records.get(this.records.length() - 1));
                }
	} catch (InterruptedException e) {
		Thread.currentThread().interrupt();
		throw new KafkaException("Interrupted while queuing ack for " + this.records, e);
	}
	if (this.immediate) {
		ListenerConsumer.this.consumer.wakeup();
	}
}
```

garyrussell:
That's by design; the manual methods are entirely in your domain - you have complete control of when the offset is committed. You should only acknowledge when you want the commit applied. If you want to commit at the end of each batch, let the container do the commits (`AckMode.BATCH`).

garyrussell:
Oops; sorry; misread the issue.

artembilan:
Oh! That is good catch!

Meanwhile I am fully agree with you that no reason to commit offsets for all the records in batch, I think we can't make an assumption that the last one has the highest offset.
So, my suggestion is to iterate throughout the `this.records` to find max `offset`.
I won't mind if there is really some provement that "the max is always in the end of batch" .

Contribution are welcome!

ltalhouarne:
Agreed. The highest offset needs to be committed, not necessarily the last item on the list. 
I would be happy to work on it and submit a PR.

[...]

PR #263, opened 2017-03-31
GH-262: In a batch only commit highest offset for each partition  Fix…

ltalhouarne:
…es GH-262 (https://github.com/spring-projects/spring-kafka/issues/262)  Committing each offset in a batch hinders the performance. Only the highest offset for a partition should be committed from a batch.

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
@@ -729,8 +729,12 @@ public class KafkaMessageListenerContainerTests {
 				return invocation.callRealMethod();
 			}
 			finally {
+				boolean smallOffsetCommitted = false;
 				for (Entry<TopicPartition, OffsetAndMetadata> entry : map.entrySet()) {
-					if (entry.getValue().offset() == 2) {
+					if (entry.getValue().offset() == 1) {
+						smallOffsetCommitted = true;
+						logger.error("The highest offset should be the only one committed.");
+					} else if (!smallOffsetCommitted && entry.getValue().offset() == 2) {
 						commitLatch.countDown();
 					}
 				}
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
@@ -734,7 +734,8 @@ public class KafkaMessageListenerContainerTests {
 					if (entry.getValue().offset() == 1) {
 						smallOffsetCommitted = true;
 						logger.error("The highest offset should be the only one committed.");
-					} else if (!smallOffsetCommitted && entry.getValue().offset() == 2) {
+					}
+					else if (!smallOffsetCommitted && entry.getValue().offset() == 2) {
 						commitLatch.countDown();
 					}
 				}
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -1084,7 +1084,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 					Map<Integer, ConsumerRecord<K, V>> highestOffsetMap = new HashMap<>();
 
 					for (ConsumerRecord<K, V> record : this.records) {
-						if(record != null) {
+						if (record != null) {
 							ConsumerRecord<K, V> consumerRecord = highestOffsetMap.get(record.partition());
 
 							if (consumerRecord == null || record.offset() > consumerRecord.offset()) {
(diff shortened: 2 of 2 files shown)


spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -729,8 +729,13 @@ public void testBatchListenerManual() throws Exception {
 				return invocation.callRealMethod();
 			}
 			finally {
+				boolean smallOffsetCommitted = false;
 				for (Entry<TopicPartition, OffsetAndMetadata> entry : map.entrySet()) {
-					if (entry.getValue().offset() == 2) {
+					if (entry.getValue().offset() == 1) {
+						smallOffsetCommitted = true;
+						logger.error("The highest offset should be the only one committed.");
```
artembilan: Well, it's really an `IllegalState` to be here.
According our new logic we we should never get lower offset for the `commit()`.
I'll fix on merge

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/1f1a54434fee5afc37ca8fc99cd3cc43d8f9413a and cherry-picked to `master`.

@ltalhouarne ,
Thank you for contribution.
Looking forward for more!

garyrussell:
Hmmm - sorry; I wanted to discuss this before merging.

It really should be consistent with what we do with non-manual acks - see `updatePendingOffsets()` and `addOffset()` which does assume the highest offset is last.

artembilan:
> I wanted to discuss this before merging.

Well, sorry to be in a rush, but I still think that we can discuss and provide another PR if that is needed.

The code is pretty simple to share here directly:
```
private void updatePendingOffsets() {
	ConsumerRecord<K, V> record = this.acks.poll();
	while (record != null) {
		addOffset(record);
		record = this.acks.poll();
	}
}

private void addOffset(ConsumerRecord<K, V> record) {
	if (!this.offsets.containsKey(record.topic())) {
		this.offsets.put(record.topic(), new HashMap<Integer, Long>());
	}
	this.offsets.get(record.topic()).put(record.partition(), record.offset());
}
...
Map<TopicPartition, OffsetAndMetadata> commits = new HashMap<>();
for (Entry<String, Map<Integer, Long>> entry : this.offsets.entrySet()) {
	for (Entry<Integer, Long> offset : entry.getValue().entrySet()) {
		commits.put(new TopicPartition(entry.getKey(), offset.getKey()),
				new OffsetAndMetadata(offset.getValue() + 1));
	}
}
```
So, we just re-map `acks` into `offsets`. That isn't obvious for me that highest is the last...

What do you mean, please and what is your suggestion for fixing the MANUAL mode?
Thanks

garyrussell:
>What do you mean, please and what is your suggestion for fixing the MANUAL mode?

I am just saying we should be consistent in the two places where we determine the actual offset to commit (manual and auto).

I think it's reasonable to assume the last is the highest but I don't mind the extra check.

artembilan:
Yeah... That's true, but we still don't have choice unless iterate batch to determine that highest offset for each partition in the batch.

Or is your idea that we should apply the same "extract highest offset" algorithm into the  non-manual acks for BATCH mode ?

garyrussell:
`addOffset()` already does that, but it assumes the last is the highest.

By consistent, I mean... consistent.

One inconsistency is we're doing the map/reduce in the container-ack case after loading up the acks whereas in the manual case, we're doing it before writing to acks; that's better because the reduce is done earlier.

I am just saying we should consider doing the same for the container-ack case - i.e. instead of loading up `acks` and then map/reduce, just add the ack to `this.offsets` (if it's higher).

In other words, use the same algo regardless of manual/auto.

artembilan:
Now I got it!

Thank you, Gary, for your patience!

@ltalhouarne , WDYT think about one more contribution?

ltalhouarne:
@artembilan I am fine with the aforementioned approach and would be happy to contribute. Lookout for another PR addressing the new approach.

ltalhouarne:
@artembilan @garyrussell After digging into the flow and moving the logic to the `addOffset` method, I realized that the new approach would only work with `AckMode.MANUAL`.   

`AckMode.MANUAL_IMMEDIATE` bypasses the `addOffset` method and commits for each record: 

```
if (ListenerConsumer.this.isManualImmediateAck) {
	try {
		ackImmediate(record);
	} catch (WakeupException e) {
		// ignore - not polling
	}
}
else {
	addOffset(record);
}
```

So I believe there is indeed a need for preventing all offsets in a partition to make it to `this.acks` as previously designed. 

Here is a diff of the work with the same test `testBatchListenerManual` failing:

https://github.com/spring-projects/spring-kafka/compare/1.1.x...ltalhouarne:hotfix/highest-offset-different-approach?expand=1

ltalhouarne:
@artembilan @garyrussell,

I have created the following PR: https://github.com/spring-projects/spring-kafka/pull/267

The work in the PR adds the highest offset per partition logic to non-manual ack batch listeners before adding to `acks`. I also took the liberty of adding the logic to the `addOffset` method to avoid any assumptions as far as the last record having the highest offset.
Finally I fixed an existing test that **would not** throw an exception if a smaller offset was committed.

PR #267, opened 2017-04-01
GH-262: In a batch only commit highest offset for each partition  Fix…

ltalhouarne:
…es GH-262 (https://github.com/spring-projects/spring-kafka/issues/262)

This commit adds the highest offset logic to batch listener with non-manual commits. It also adds the logic to the addOffset method to prevent any assumptions from being made.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -728,14 +728,14 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 						this.batchListener.onMessage(recordList);
 					}
 					if (!this.isAnyManualAck && !this.autoCommit) {
-						for (ConsumerRecord<K, V> record : recordList) {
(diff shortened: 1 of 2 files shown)


ltalhouarne:
@garyrussell Is this what you mind with your feedback on https://github.com/spring-projects/spring-kafka/pull/263 ? 

I would be happy to address this work if you had something different in mind.

garyrussell:
@ltalhouarne Yes, sorry; I have been busy on other things; this is what I mean; thanks!

When we forward port this stuff to master, we can simplify all this map/reduce stuff using streams since we're on Java 8 there.

ltalhouarne:
@garyrussell Great! Streams will make this simpler indeed.

Keep up the great work! Big advocate of this library.

garyrussell:
Merged as e41cf660f8ff5c9dda4a4761956e2534be6c5cd9 - @ltalhouarne - please limit commit headlines to 50 chars - per the [contribution guidelines](https://github.com/spring-projects/spring-kafka/blob/master/CONTRIBUTING.adoc).

Thanks for the contribution!

I will port to master, with the discussed stream changes.

PR #269, opened 2017-04-04
GH-268: Defer Manual Commits Until After Listener

### LINKED ISSUE #268: Manual Acks in RebalanceListener Are Not Committed
ERROR: KafkaMessageListenerContainer$ListenerConsumer - Container exception
    org.apache.kafka.clients.consumer.CommitFailedException: Commit cannot be completed since the group has already rebalanced and assigned the partitions to another member.

Ref: http://stackoverflow.com/questions/43164534/in-spring-kafka-which-method-does-the-same-thing-like-onpartitionsrevoked

While the internal rebalance listener will commit any pending acks before calling a user-defined `RebalanceListener`, any acks performed in that listener are not committed until later and you get the above error.

garyrussell:
Fixes #268

Previously, the container's internal rebalance listener committed any manual acks before
calling a user `RebalanceListener`. If that listener added more acks, they were not committed
until after the rebalance, which is an error.

Defer committing manual acks until after the user listener (if any) is called so all such
acks are committed.

Add a test case:
- two partitions; send two messages to each
- ack the first message in each partition in the listener
- start a second container to force a rebalance
- ack the second message in each partition in the rebalance listener
- assert that the offsets were committed correctly

__cherry-pick to 1.1.x__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -388,7 +388,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 									"and will stop the invoker.");
 						}
 						if (ListenerConsumer.this.listenerInvokerFuture != null) {
-							stopInvokerAndCommitManualAcks();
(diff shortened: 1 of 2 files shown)


artembilan:
... and cherry-picked to `1.1.x`

PR #271, opened 2017-04-04
GH-224: Add Seek to Beginning/End

### LINKED ISSUE #224: How can I seek the consumer to the end
Each time when I start the consumer I need latest messages only for a specific topic, On each consumer start I will call seekToEnd(), something like this. Is it supported in the current version.

garyrussell:
Resolves #224

Add specific methods for `seekToBeginning`, `seekToEnd`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerSeekAware.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerSeekAware.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -57,6 +57,7 @@ public interface ConsumerSeekAware {
 	/**
 	 * A callback that a listener can invoke to seek to a specific offset.
 	 */
+	@FunctionalInterface
 	interface ConsumerSeekCallback {
 
 		/**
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/TopicPartitionInitialOffset.java
```diff
@@ -80,9 +98,28 @@ public TopicPartitionInitialOffset(String topic, int partition, Long initialOffs
...
+	 * Construct an instance with the provided initial offset.
+	 * @param topic the topic.
+	 * @param partition the partition.
+	 * @param initialOffset the initial offset.
+	 * @param relativeToCurrent true for the initial offset to be relative to
+	 * the current consumer position, false for a positive initial offset to
+	 * be absolute and a negative offset relative to the current end of the
+	 * partition.
+	 * @param position {@link SeekPosition#BEGINNING} or {@link SeekPosition#END}
+	 * overrides initialOffset and relativeToCurrent.
```
artembilan: If that is a fact, why just don't encourage end-user to type those extra options?
In other words if they are so mutually exclusive, let's make a ctor without them!
```
public TopicPartitionInitialOffset(String topic, int partition, Long initialOffset, boolean relativeToCurrent)
...
public TopicPartitionInitialOffset(String topic, int partition, SeekPosition position) 
```

spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerSeekAware.java
```diff
@@ -68,6 +69,27 @@
...
 
+		/**
+		 * Queue a seekToBeginning operation to the consumer. The seek will occur after
+		 * any pending offset commits. The consumer must be currently assigned the
+		 * specified partition.
+		 * @param topic the topic.
+		 * @param partition the partition.
+		 */
+		default void seekToBeginning(String topic, int partition) {
+			seek(topic, partition, 0);
```
artembilan: I'm not sure in such a default implementation.
It doesn't reflect reality about `BEGINNING` and there is no intentions to expose this interface for end-users.
This is just an internal Framework hook to let end-users to perform seeks on the provided callback.
No body is going to implement it himself.

I will be pretty happy do not have this interface as a `@FunctionalInterface`.
Just IMHO...

spring-kafka/src/main/java/org/springframework/kafka/support/TopicPartitionInitialOffset.java
```diff
@@ -83,6 +101,26 @@ public TopicPartitionInitialOffset(String topic, int partition, Long initialOffs
...
 		this.relativeToCurrent = relativeToCurrent;
+		this.position = null;
+	}
+
+	/**
+	 * Construct an instance with the provided initial offset.
+	 * @param topic the topic.
+	 * @param partition the partition.
+	 * @param initialOffset the initial offset.
+	 * @param relativeToCurrent true for the initial offset to be relative to
```
artembilan: These couple params are not relevant any more.
Travis confirm that.
Will fix on merge...

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/79370858f18a68108553d8ec0cf19442ac9ded35

PR #275, opened 2017-04-06
Fix Stubbing in Tests

garyrussell:
Prevent the consumer thread from interacting with the mock before the stubbing is complete.

__cherry-pick to master, 1.1.x__


artembilan:
... and cherry picked to `master` and `1.1.x`

PR #277, opened 2017-04-07
Fix Executor Documentation

garyrussell:
The recent clarification of the default thread names incorrectly stated the
threads always end `-1` - the default executor is put into the container properties
while starting so subsequent starts will find an existing executor.

__cherry-pick to 1.2.x, 1.1.x__


artembilan:
... and cherry-picked to `1.2.x` and `1.1.x`.

Issue #276, opened 2017-04-07
Remove Listener Task Executor


garyrussell:
With the implementation of [KIP-62](https://cwiki.apache.org/confluence/display/KAFKA/KIP-62%3A+Allow+consumer+to+send+heartbeats+from+a+background+thread), hearbeats are now handled by the client on a separate thread. This means we can simplify the container by avoiding the thread-handoff (as we currently do with `enable.auto.commit=true`). We can always call the listener on the consumer thread.

#202, #246, #248

PR #279, opened 2017-04-07
GH-276: Rip out ListenerInvoker

garyrussell:
Resolves: https://github.com/spring-projects/spring-kafka/issues/276

Since the implementation if KIP-62, it is no longer necessary to run the listener
on a separate thread, enabling pause/resume on the consumer thread.

Remove all logic and properties related to handing off to a `ListenerInvoker` and remove all
"slow consumer" tests.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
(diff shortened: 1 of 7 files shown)


artembilan:
LGTM, but one question: 

Is that true that we don't have one `TaskExecutor` any more?
I mean the recent [doc](https://github.com/spring-projects/spring-kafka/commit/49e2cad975e4e0b2b9f25c35fb01a1de954cec0f) about threads in the `ListerContainer` must be polished with this PR, too.

garyrussell:
Sorry - yes; I meant to change that too - will push...

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/9d533988194ae0e50859b4772ad504ed9943817c.

Let me know if that should be cherry-picked to `1.2.x`!

Issue #249, opened 2017-03-07
Please add links to spring-docs resources


zgagnon:
When viewing the Spring Docs for spring-kafka (http://docs.spring.io/spring-kafka/docs/1.1.2.RELEASE/reference/html/_preface.html) it is very difficult to find this github, or the javadoc. It would be helpful if the documentation listed there contained links to both.

Thanks!

artembilan:
Does this work for you http://projects.spring.io/spring-kafka/ ?

artembilan:
I have just open even the latest Spring Framework doc and there is no such an information, too: http://docs.spring.io/spring/docs/5.0.0.BUILD-SNAPSHOT/spring-framework-reference/html/spring-introduction.html

But you really can find it in the main project page: https://spring.io/projects

zgagnon:
The spring.io page is great. Maybe a link back to it from the reference would be helpful tho? I arrived there through Google, not through the spring.io page.

PR #281, opened 2017-04-07
Explanation for MessageListener/MessageListenerContainer

zgagnon:
Adds text describing adding the MessagListener to a ContainerProperties object and creating the MessageListenerContainer using those properties.


zgagnon:
Resolves [#249](https://github.com/spring-projects/spring-kafka/issues/249)

Issue #280, opened 2017-04-07
Add ConsumerAwareMessageListener


garyrussell:
Allow the user to pause/resume the consumer (and interact with it - e.g. commit rather than using an `AcknowledgingMessageListener`.

We would have to invoke the listener with null `ConsumerRecord` to enable resuming a paused consumer.

consulthys:
Cool! Will this be backported to the 1.2 branch?

garyrussell:
No; this is a 2.0 feature; the first milestone for 2.0 (M1) is scheduled to be released later this week. It requires previous commits that rework in the threading model now that the client code handles heartbeats differently.

PR #282, opened 2017-04-10
GH-280: Add Consumer-Aware Listeners

garyrussell:
Resolves: https://github.com/spring-projects/spring-kafka/issues/280

- add Consumer-aware listeners (batch and record)
- refactor listener types into hierarchies, rooted on `GenericMessageListener<T>`
  - preserve `@FunctionalInterface` semantics at each level by using `default` and `@Override`s
- refactor Filter and Retry adapters to a single class in each case; update to suport consumer aware delegates
- add `DelegatingMessageListener<T>` so that the container can determine the ultimate listener type
  - avoids the need to create an `Acknowledgment` object if it will definitely be discarded
  - `@KafkaListener` gets access to the `Consumer<?, ?>` too
- Add test case to verify delegate listener type detection (multi-level) and correct method invocation
- Add test case to receive a message from all 8 listener types

__docs to follow after review__


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RetryingMessageListenerAdapter.java
```diff
@@ -63,17 +66,49 @@ public RetryingMessageListenerAdapter(MessageListener<K, V> messageListener, Ret
 	}
 
 	@Override
-	public void onMessage(final ConsumerRecord<K, V> record) {
+	public void onMessage(final ConsumerRecord<K, V> record, Acknowledgment acknowledgment, Consumer<?, ?> consumer) {
 		getRetryTemplate().execute(new RetryCallback<Void, KafkaException>() {
```
artembilan: Can be replaced with Lamda:

lambda-capable

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AbstractDelegatingMessageListenerAdapter.java
```diff
@@ -35,16 +38,20 @@
  * @since 1.1
  *
  */
-public abstract class AbstractMessageListenerAdapter<K, V, T> implements ConsumerSeekAware {
+public abstract class AbstractDelegatingMessageListenerAdapter<K, V, T>
```
artembilan: IDEA says me that we don't need `K, V`:
no_need_kv

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RetryingMessageListenerAdapter.java
```diff
@@ -37,7 +40,7 @@
  */
 public class RetryingMessageListenerAdapter<K, V>
```
artembilan: Looks like we are missing `RetryingBatchMessageListenerAdapter` and there is no way to retry the whole batch.
Or is that by design do not encourage end-users to retry on the whole batch ?
Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -152,22 +150,28 @@ protected void doStart() {
...
-		this.listenerConsumer = new ListenerConsumer(this.listener, this.acknowledgingMessageListener);
+		Assert.state(messageListener instanceof GenericMessageListener, "Listener must be a GenericListener");
+		this.listener = (GenericMessageListener<?>) messageListener;
+		ListenerType listenerType = ListenerUtils.determineListenerType(this.listener);
+		if (this.listener instanceof DelegatingMessageListener) {
+			Object delegating = this.listener;
+			while (delegating instanceof DelegatingMessageListener) {
+				Object delegate = ((DelegatingMessageListener<?>) delegating).getDelegate();
+				listenerType = ListenerUtils.determineListenerType(delegate);
+				if (delegate instanceof DelegatingMessageListener) {
```
artembilan: Looks like this `if...else` is fully covered with that `while()`:
```
while (delegating instanceof DelegatingMessageListener) {
	delegating = ((DelegatingMessageListener<?>) delegating).getDelegate();
	listenerType = ListenerUtils.determineListenerType(delegating);
}
```

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringMessageListenerAdapter.java
```diff
@@ -41,13 +47,69 @@
...
+				case ACKNOWLEDGING:
+					this.delegate.onMessage(consumerRecord, acknowledgment);
+					break;
+				case CONSUMER_AWARE:
+					this.delegate.onMessage(consumerRecord, consumer);
+					break;
+				case SIMPLE:
+					this.delegate.onMessage(consumerRecord);
+			}
+			switch (this.delegateType) {
```
artembilan: Missed `else` around this switch . 
Although I'd say we don't need switch at all. One `if` should be enough:
```
		else {
			if (this.ackDiscarded && acknowledgment != null &&
					(ListenerType.ACKNOWLEDGING_CONSUMER_AWARE.equals(this.delegateType) ||
							ListenerType.ACKNOWLEDGING.equals(this.delegateType))) {
				acknowledgment.acknowledge();
			}
		}

```

spring-kafka/src/main/java/org/springframework/kafka/listener/AcknowledgingConsumerAwareMessageListener.java
```diff
@@ -0,0 +1,50 @@
...
+	 * will detect that we are a consumer aware acknowledging listener.
+	 * @param data the data to be processed.
+	 * @param consumer the consumer.
+	 */
+	@Override
+	default void onMessage(ConsumerRecord<K, V> data, Consumer<?, ?> consumer) {
+		throw new UnsupportedOperationException("Container should never call this");
+	}
+
+	@Override
```
garyrussell: Inherited.

spring-kafka/src/main/java/org/springframework/kafka/listener/BatchAcknowledgingConsumerAwareMessageListener.java
```diff
@@ -0,0 +1,57 @@
...
+	 * will detect that we are a consumer aware acknowledging listener.
+	 * @param data the data to be processed.
+	 * @param consumer the consumer.
+	 */
+	@Override
+	default void onMessage(List<ConsumerRecord<K, V>> data, Consumer<?, ?> consumer) {
+		throw new UnsupportedOperationException("Container should never call this");
+	}
+
+	@Override
```
garyrussell: Inherited.

spring-kafka/src/main/java/org/springframework/kafka/listener/BatchConsumerAwareMessageListener.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Gary Russell
+ *
+ * @since 2.0
+ */
+@FunctionalInterface
+public interface BatchConsumerAwareMessageListener<K, V> extends BatchAcknowledgingMessageListener<K, V> {
```
garyrussell: Yes, I think I can do that with how things worked out in the end; this was an artifact of an earlier iteration.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RetryingMessageListenerAdapter.java
garyrussell: Yes, I believe there is something in the reference about it  - we don't know where in the batch to restart from; they are better off using a retry template within their listener.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RetryingMessageListenerAdapter.java
garyrussell: ok

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: No; we can't unconditionally cast the delegate's delegate to `DelegatingMessageListener` we have to test its type first.

artembilan: We don't cast unconditionally. That is what done by the `while()`

garyrussell: Huh?

`delegating = ((DelegatingMessageListener<?>) delegating).getDelegate();`

in your suggested "fix" unconditionally casts the delegate's delegate without testing it first.

garyrussell: Doh

spring-kafka/src/main/java/org/springframework/kafka/event/ListenerContainerIdleEvent.java
```diff
@@ -71,6 +75,15 @@ public String getListenerId() {
 		return this.listenerId;
 	}
 
+	/**
+	 * Retrieve the consumer. Only populated if the listener is consumer-aware.
+	 * Allows the listener to resume a paused consumer.
+	 * @return the consumer.
```
artembilan: `@since 2.0`

spring-kafka/src/main/java/org/springframework/kafka/listener/GenericMessageListener.java
```diff
@@ -17,20 +17,15 @@
...
  *
  */
+@FunctionalInterface
 public interface GenericMessageListener<T> extends KafkaDataListener<T> {
 
-	/**
-	 * Invoked with data from kafka.
-	 * @param data the data to be processed.
-	 */
-	void onMessage(T data);
```
artembilan: Hm. I wonder if we need this interface (or that `KafkaDataListener`) yet at all...

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+package org.springframework.kafka.listener;
+
+import org.springframework.util.Assert;
+
+/**
+ * Listener utilities.
+ *
+ * @author Gary Russell
+ * @since 5.0
```
artembilan: `@since 2.0`

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RetryingMessageListenerAdapter.java
artembilan: Well, there is : http://docs.spring.io/spring-kafka/docs/2.0.0.BUILD-SNAPSHOT/reference/html/_reference.html#_retrying_deliveries, but not so much:

> A retry adapter is not provided for any of the batch message listeners.

Would be great to elaborate.
Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/GenericMessageListener.java
garyrussell: Yes, we can remove it now (since it's a major release). This was needed before because we needed a common super-interface for `GenericAcknowledgingMessageListener` and `GenericMessageListener`.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/d94ac5c2dfdc21101fbddf5b7f5247c26e45f592

Issue #288, opened 2017-04-14
Consumer Factory Has the Wrong Default for isAutoCommit


garyrussell:
Results in the container committing which causes strange behavior with the offsets.

http://stackoverflow.com/questions/43405009/kafka-stop-consuming-message-from-new-assigned-patitions-after-rebalancing/43413899#43413899

wuhuanxing:
Hi Gary,
the work-around setting spring.kafka.consumer.enable-auto-commit=true will cause another issue:
the messages will be consumed twice or more times ( by original consumer and new assigned consumers after rebalancing)

garyrussell:
Auto-commit only commits every 5 seconds by default. If you use container commits (`AckMode.RECORD`) you should be ok.

PR #289, opened 2017-04-14
GH-288: Fix autoCommit Default Return Value

garyrussell:
Fixes: https://github.com/spring-projects/spring-kafka/issues/288

The `DefaultKafkaConsumerFactory` incorrectly returns `false` if no `enable.auto.commit`
property is present. This causes the container to issue commits, which causes strange
offset behavior. Unread partitions are set to the end so messages can be lost.

Work-around is to explicitly add the property.

See: http://stackoverflow.com/questions/43405009/kafka-stop-consuming-message-from-new-assigned-patitions-after-rebalancing/43413899#43413899

__cherry-pick to 1.2.x, 1.1.x, 1.0.x__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


artembilan:
... and cherry-picked.

wuhuanxing:
Hi there,

I have a multi-thread issue with spring-kafka:
http://stackoverflow.com/questions/43601004/spring-kafka-thorws-instancealreadyexistsexception-exception-after-setting-concu

can anyone kindly help?

PR #291, opened 2017-04-18
Change RecoveryCallback from <Void> to <Object>

garyrussell:
Actually <? extends Object> for backwards compatibility.

__cherry-pick to 1.2.x, 1.1.x__


artembilan:
... and cherry-picked to `1.1.x` and `1.2.x`.

Issue #194, opened 2016-10-13
replication-offset-checkpoint.tmp (No such file or directory)


MichalOrlowski:
We are using Embedded Kafka for Integration testing. 
What we're getting is `FileNotFoundException` (sometimes) on Jenkins. 
Question - is there a possibility to turn off `ReplicaManager` or so? Maybe there's another solution.

```
2016-10-13 07:53:35.161 ERROR 9469 --- [fka-scheduler-1] kafka.server.ReplicaManager              : [Replica Manager on Broker 0]: Error writing to highwatermark file: 

java.io.FileNotFoundException: /tmp/kafka-3266968904825284552/replication-offset-checkpoint.tmp (No such file or directory)
    at java.io.FileOutputStream.open0(Native Method)
    at java.io.FileOutputStream.open(FileOutputStream.java:270)
    at java.io.FileOutputStream.<init>(FileOutputStream.java:213)
    at java.io.FileOutputStream.<init>(FileOutputStream.java:162)
    at kafka.server.OffsetCheckpoint.write(OffsetCheckpoint.scala:37)
    at kafka.server.ReplicaManager$$anonfun$checkpointHighWatermarks$2.apply(ReplicaManager.scala:874)
    at kafka.server.ReplicaManager$$anonfun$checkpointHighWatermarks$2.apply(ReplicaManager.scala:871)
    at scala.collection.TraversableLike$WithFilter$$anonfun$foreach$1.apply(TraversableLike.scala:778)
    at scala.collection.immutable.Map$Map1.foreach(Map.scala:116)
    at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:777)
    at kafka.server.ReplicaManager.checkpointHighWatermarks(ReplicaManager.scala:871)
    at kafka.server.ReplicaManager$$anonfun$1.apply$mcV$sp(ReplicaManager.scala:153)
    at kafka.utils.KafkaScheduler$$anonfun$1.apply$mcV$sp(KafkaScheduler.scala:110)
    at kafka.utils.CoreUtils$$anon$1.run(CoreUtils.scala:60)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
[...]

PR #294, opened 2017-04-21
GH-194: Make `KafkaEmbedded` as a bean

artembilan:
Fixes: spring-projects/spring-kafka#194

To avoid concurrent calls of the shutdown hooks make an `KafkaEmbedded`
as a `Lifecycle` to let ApplicationContext to call `stop()` before
destroying itself

That way the Kafka servers and zookeeper and also working directory
are destroyed and removed before JMV shutdown hooks
and `FileNotFoundException /tmp/kafka-3266968904825284552/replication-offset-checkpoint.tmp`
should disappear

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -167,7 +167,7 @@ project ('spring-kafka-test') {
 	description = 'Spring Kafka Test Support'
 
 	dependencies {
-		compile "org.springframework:spring-beans:$springVersion"
+		compile "org.springframework:spring-context:$springVersion"
 		compile "org.springframework:spring-test:$springVersion"
 		compile "org.springframework.retry:spring-retry:$springRetryVersion"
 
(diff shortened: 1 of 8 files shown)


garyrussell:
Cool!

We should add some docs above [here](http://docs.spring.io/spring-kafka/reference/htmlsingle/#_hamcrest_matchers). Something like...

>It is generally recommended to use the rule as a `@ClassRule` to avoid starting/stopping the broker between tests (and use a different topic for each test). If you are using Spring's test application context caching, you can also declare a `KafkaEmbedded` bean, so a single broker can be used across multiple test classes. 

(plus some example configuration).

artembilan:
Yeah,,, I also thought about something like `@EmbeddedKafka` on the test class level, but since I'm not familiar with the Spring Test Framework caching magic, I'm not sure that registering bean from that level would work as expected...

Just need to get some feedback from @hoaz to go ahead!

artembilan:
Let's don't cherry-pick!
No issue from our side.
This feature can just be available from the `2.0` version

garyrussell:
Merged as 6b5c0878e440ff7a5f8a1b79afb754a42b51f226

PR #299, opened 2017-04-26
Move setting listenerType outside while loop

izeye:
No need to be inside while loop, so this PR moves it outside.


garyrussell:
Thanks; good catch.

PR #302, opened 2017-04-27
Add JsonSerde

### LINKED ISSUE #301: Add `Serde` support for serialization and deserialization
The Kafka Streams API provides a higher abstraction for both serializing and deserializing objects. Since we already provide support for serializing and deserializing JSON it would be useful to have a `JsonSerde` as well.

mbogoevici:
Fix #301

Adds a `JsonSerde` wrapper for `JsonSerializer` and
`JsonDeserializer`.

Increase commit frequency for `KafkaStream` tests to
make them publish updates more frequently (thus reducing
test duration)

(Would nice to have it backported to 1.1.x since it's not 2.0-dependent).

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerde.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerde.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


artembilan:
More polishing: https://github.com/spring-projects/spring-kafka/commit/fbffe60a529c638c33dcf47d947e7445e005c729

The generic type propagation doesn't work 😢

mbogoevici:
Thanks @artembilan - haven't thought of the generic type propagation :).

I was hoping to have separate tests for the Serde and the rest of the stream stuff, but it's fine having them all in one test I guess.

artembilan:
Well, looking to your `KafkaStreamsWithJsonPojoTests` I'd not say it is just about `Serde`. You really have there full KStream integration test. So, no reason to keep it separately since utilizing the functionality in the existing KStream scenario is fine. Although that hide a bit the `Serde` fact. Isn't so obvious at a glance where is JSON.

Any objections do not proceed with my suggestions?
Thanks

mbogoevici:
>Any objections do not proceed with my suggestions?

Not really :) As I said - thanks for fixing the generic type propagation and  I'm fine with having it all in one test.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/915b098c66fbacab38f268de1902e19df406232b and cherry-picked as https://github.com/spring-projects/spring-kafka/commit/77055ac44aaa5668e73bc28d43bde84e64256435 after resolving conflicts.

Issue #201, opened 2016-11-04
Move support for Kerberos from Spring Cloud Stream Kafka Binder to Spring Kafka


markpollack:
The functionality to configure security in Spring Cloud Stream Kafka Binder

http://docs.spring.io/spring-cloud-stream/docs/Brooklyn.SR1/reference/htmlsingle/#_using_spring_boot_properties

Should be moved into the Spring Kafka project as it is more general purpose.  Property name would have their 'stream' bits removed, so spring.kafka.jaas.loginModule.

The PR https://github.com/spring-projects/spring-boot/pull/6961 should then be updated to include these properties in Spring Boot autoconfiguration.

mbogoevici:
Maybe this is a good time to discuss about `spring-kafka-admin` as part of what's going on in the binder is tied to creating topics which requires Zookeeper permissions (which are another part of the binder setup, but are not required by `spring-kafka`).

PR #310, opened 2017-05-09
GH-201: Add Support for Kerberos Configuration

garyrussell:
Resolves: https://github.com/spring-projects/spring-kafka/issues/201

Port code from spring-cloud-stream-binder-kafka.

__Pretty much a straight port from SCSt__


garyrussell:
@mbogoevici - any suggestions for a real test?

garyrussell:
(hoping you'll say "no need" 😄 )

spring-kafka/src/main/java/org/springframework/kafka/security/jaas/JaasLoginModuleConfiguration.java
```diff
@@ -0,0 +1,89 @@
...
+		this.loginModule = loginModule;
+	}
+
+	public String getControlFlag() {
+		return this.controlFlag.toString();
+	}
+
+	public void setControlFlag(String controlFlag) {
+		Assert.notNull(controlFlag, "cannot be null");
+		if (AppConfigurationEntry.LoginModuleControlFlag.OPTIONAL.equals(controlFlag)) {
```
artembilan: How about to switch to `switch`?
http://docs.oracle.com/javase/7/docs/technotes/guides/language/strings-switch.html

spring-kafka/src/main/java/org/springframework/kafka/security/jaas/JaasInitializerListener.java
```diff
@@ -0,0 +1,117 @@
...
+			if (this.ignoreJavaLoginConfigParamSystemProperty && this.jaasLoginModuleConfiguration != null) {
+				Map<String, AppConfigurationEntry[]> configurationEntries = new HashMap<>();
+				AppConfigurationEntry kafkaClientConfigurationEntry = new AppConfigurationEntry(
+						this.jaasLoginModuleConfiguration.getLoginModule(),
+						this.jaasLoginModuleConfiguration.getControlFlagValue(),
+						this.jaasLoginModuleConfiguration.getOptions());
+				configurationEntries.put(JaasUtils.LOGIN_CONTEXT_CLIENT,
+						new AppConfigurationEntry[] { kafkaClientConfigurationEntry });
+				Configuration.setConfiguration(new InternalConfiguration(configurationEntries));
+				// Workaround for a 0.9 client issue where even if the Configuration is
```
artembilan: Maybe we don't need this workaround anymore since we are far away from `0.9` Client?
OTOH how about to switch to the `SmartInitializingSingleton` instead of `ApplicationListener` and rename the class to the `KafkaJaasInitializer`. Although, IMHO, we don't need this class at all and its functionality can be covered by the `JaasLoginModuleConfiguration`, but may be with more Kafka-specific name...

spring-kafka/src/test/java/org/springframework/kafka/security/jaas/JaasInitializerListenerTests.java
```diff
@@ -0,0 +1,85 @@
...
+
+import com.sun.security.auth.login.ConfigFile;
+
+/**
+ * @author Marius Bogoevici
+ * @author Gary Russell
+ *
+ * @since 2.0
+ */
+@SuppressWarnings("restriction")
```
artembilan: Just curious: what is this `restriction` about?
Thanks

garyrussell: I get this warning in STS...

>Access restriction: The type 'ConfigFile' is not API (restriction on required library '/Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home/jre/lib/rt.jar')	JaasInitializerListenerTests.java	/spring-kafka/src/test/java/org/springframework/kafka/security/jaas	line 36	Java Problem

artembilan: So, maybe follow with its JavaDocs recommendation and do this:
```
new sun.security.provider.ConfigFile.Spi(uri)
```

```
// NOTE: As of JDK 8, this class instantiates
// sun.security.provider.ConfigFile.Spi and forwards all methods to that
// implementation. All implementation fixes and enhancements should be made to
// sun.security.provider.ConfigFile.Spi and not this class.
// See JDK-8005117 for more information.

```

garyrussell: Doesn't help...

>Access restriction: The constructor 'ConfigFile.Spi(URI)' is not API (restriction on required library '/Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home/jre/lib/rt.jar')	KafkaJaasLoginModuleInitializerTests.java	/spring-kafka/src/test/java/org/springframework/kafka/security/jaas	line 49	Java Problem

garyrussell:
Pushed - I need to investigate whether that 0.9 client issue is resolved.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -16,17 +16,22 @@ You can now annotate `@KafkaListener` methods (and classes, and `@KafkaHandler`
...
-===== @EmbeddedKafka Annotation
+==== @EmbeddedKafka Annotation
 
 For convenience a test class level `@EmbeddedKafka` annotation is provided with the purpose to register `KafkaEmbedded` as a bean.
 See <<testing>> for more information.
+
+==== Kerberos Configuration
+
+Support for configuring Kerberos is now provided.
+See <<Kerberos>> for more information.
```
garyrussell: Need 2 fix case

artembilan: I don't merge yet.
I'm just fine with the current code.
Waiting for your confirmation about that `0.9` issue around system property.
Anyway we don't need to mention such a version in those comments since this code is already for higher versions.

garyrussell:
@artembilan I checked on the 0.9 client workaround; it looks like the original problem was fixed, but we still get a misleading DEBUG log message...

>10:16:45.889 DEBUG [main][org.apache.kafka.common.security.JaasUtils] System property 'java.security.auth.login.config' and Kafka SASL property 'sasl.jaas.config' are not set, using default JAAS configuration.

See `JaasUtils.defaultJaasConfig()`.

So my vote is to leave the workaround in place.

artembilan:
Fine with me.
Merging...

src/reference/asciidoc/whats-new.adoc
artembilan: Fixed on merge, BTW

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/33b10f46084e2f4543ac366783a00a26abaf2a0c

Issue #308, opened 2017-05-09
Expose some KafkaStreams setters delegation in the KStreamBuilderFactoryBean


artembilan:
For example `setStateListener(KafkaStreams.StateListener stateListener)`:

PR #311, opened 2017-05-10
GH-308: KStreamBuilderFB Polishing

garyrussell:
- null check on state listener and exception handler in case user has set manually
- use normal method order (ctor, accessors, other methods)

__cherry-pick to 1.2.x, 1.1.x (minus state listener)__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KStreamBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KStreamBuilderFactoryBean.java
@@ -84,12 +84,6 @@ public class KStreamBuilderFactoryBean extends AbstractFactoryBean<KStreamBuilde
 		return KStreamBuilder.class;
 	}
 
-	@Override
-	protected KStreamBuilder createInstance() throws Exception {
-		return new KStreamBuilder();
-	}
-
-
 	public void setAutoStartup(boolean autoStartup) {
 		this.autoStartup = autoStartup;
 	}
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KStreamBuilderFactoryBean.java
```diff
@@ -116,8 +130,12 @@ public synchronized void start() {
 		if (!this.running) {
 			try {
 				this.kafkaStreams = new KafkaStreams(getObject(), this.streamsConfig, this.clientSupplier);
-				this.kafkaStreams.setStateListener(this.stateListener);
-				this.kafkaStreams.setUncaughtExceptionHandler(this.exceptionHandler);
+				if (this.stateListener != null) {
```
artembilan: I didn't do that knowingly because `KafkaStreams` lets to have and configure them as null.

Although I won't mind if you insist to leave your fix.

garyrussell: Yes, I understand; I checked that logic before merging; however, I thought about it overnight and I am concerned about...

    fb.getKafkaStreams().setStateListener(myStateListener);
    fb.start();

I agree the setters are useful (for XML Spring config perhaps), but I think the null check is needed for the above reason.

artembilan: ??? The `fb.start()` creates a new `KafkaStreams` each time.
So, that `getKafkaStreams().setStateListener(myStateListener)` is irrelevant.
WDYT?

garyrussell:
Doh - that will teach me not to "think" while lying in bed.

Issue #313, opened 2017-05-14
Provide a Mechanism to set the group.id on @KafkaListener


garyrussell:
Override the group id in the listener container factory's consumer factory.

Something like `idIsGroup = true` meaning the the `id` attribute becomes the group id.

Need to clone the consumer factory - add getters for deserializers.

PR #314, opened 2017-05-14
GH-313: Configure group.id on @KafkaListener

garyrussell:
Resolves: https://github.com/spring-projects/spring-kafka/issues/313

Allow configuration of the `group.id` property on the annotation.


spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpoint.java
```diff
@@ -43,6 +43,14 @@
 	String getId();
 
 	/**
+	 * Return the groupId of this endpoint - if present, overrides the the
```
artembilan: Double `the`.
Will fix on merge if only concern

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/2ca52091c100aeda0fc41cefb70b2954156f0f94

PR #321, opened 2017-05-29
GH-319: Add Acknowledgment to Retry Context [1.x]

### LINKED ISSUE #319: How to acknowledge from RecoveryCallback
Or specifically, why does `RetryContext` get populated with `"record"`, but not `"acknowledgement"` in [RetryingAcknowledgingMessageListenerAdapter](https://github.com/spring-projects/spring-kafka/blob/1.2.x/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RetryingAcknowledgingMessageListenerAdapter.java)?

	@Override
	public void onMessage(final ConsumerRecord<K, V> record, final Acknowledgment acknowledgment) {
		getRetryTemplate().execute(new RetryCallback<Object, KafkaException>() {

			@Override
			public Void doWithRetry(RetryContext context) throws KafkaException {
				context.setAttribute("record", record);
				RetryingAcknowledgingMessageListenerAdapter.this.delegate.onMessage(record, acknowledgment);
				return null;
			}

		}, (RecoveryCallback<Object>) getRecoveryCallback());
	}

Now when I want to have X retries for processing record, and then in recovery callback, say, store record in some deadletters storage - and after that of course acknowledge it - I can't seem to do it in recovery callback.

P.S. In 2.x the RetryingAcknowledgingMessageListenerAdapter seems to be removed, so with 2.x I'm not even sure where to start...

garyrussell:
Fixes #319

Add the `Acknowledgment` to the retry context if the listener type warrants it.

__cherry-pick to 1.1.x, 1.0.x__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AbstractRetryingMessageListenerAdapter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AbstractRetryingMessageListenerAdapter.java
@@ -32,6 +32,17 @@ import org.springframework.util.Assert;
  */
 public abstract class AbstractRetryingMessageListenerAdapter<K, V, T> extends AbstractMessageListenerAdapter<K, V, T> {
 
+	/**
+	 * {@link org.springframework.retry.RetryContext} attribute key for an acknowledgment
+	 * if the listener is capable of acknowledging.
+	 */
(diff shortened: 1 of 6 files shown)


artembilan:
Cherry-picked to `1.1.x`.
Backported to `1.0.x` as https://github.com/spring-projects/spring-kafka/commit/2393cfea8a8f50561bacd091501940f8c7ed4245

PR #322, opened 2017-05-31
Reactive PoC Initial Commit [DO NOT MERGE]

garyrussell:
Initial PoC (sender side).

It doesn't add much value over the pure reactive client, although it does mean that the boot autoconfigured properties can be used.

cc/ @rajinisivaram @artembilan @mbogoevici

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
@@ -162,7 +162,7 @@ public class DefaultKafkaProducerFactory<K, V> implements ProducerFactory<K, V>,
 	@Override
 	public KafkaSender<K, V> createReactiveSender() {
 		if (this.sender == null) {
-			synchronized(this) {
+			synchronized (this) {
 				if (this.sender == null) {
 					this.sender = createSender();
 				}
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
@@ -38,6 +38,10 @@ import org.apache.kafka.common.serialization.Serializer;
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.context.Lifecycle;
 
+import reactor.kafka.sender.KafkaSender;
+import reactor.kafka.sender.SenderOptions;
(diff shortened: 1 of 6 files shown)


artembilan:
👍 Nothing to complain about!

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
```diff
@@ -163,6 +167,15 @@
 	void flush();
 
 	/**
+	 * Send a Flux of SenderRecords.
+	 * @param flux the flux.
+	 * @param <T> correlation metadata type.
+	 * @param t the class for the correlation metadata
+	 * @return a flux of SenderResults.
+	 */
+	<T> Flux<SenderResult<T>> send(Flux<SenderRecord<K, V, T>> flux, Class<T> t);
```
mbogoevici: Do we want to introduce a hard dependency in Spring Kafka on Reactor or should we provide this in a separate interface (e.g. `ReactiveKafkaOperations`).

spring-kafka/src/main/java/org/springframework/kafka/core/ProducerFactory.java
```diff
@@ -30,4 +32,6 @@
 
 	Producer<K, V> createProducer();
 
+	KafkaSender<K, V> createReactiveSender();
```
mbogoevici: Same question about the hard dependency on Reactor.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -148,6 +159,23 @@ public boolean isRunning() {
...
+					this.sender = createSender();
+				}
+			}
+		}
+		return this.sender;
+	}
+
+	protected CloseSafeKafkaSender<K, V> createSender() {
+		SenderOptions<K, V> options = SenderOptions.create(this.configs);
+		return new CloseSafeKafkaSender<>(reactor.kafka.sender.internals.ProducerFactory.INSTANCE, options);
```
mbogoevici: It would be fairly hacky, but wondering if we can't simply pass a `ProducerFactory` that returns `this.producer` (and maybe revise Reactor Kafka to allow a constructor that takes a Producer as argument). It seems like this creates two separate producers managed by this class (case in which this also opens a discussion about user expectations wrt ordering if mixing reactive/non-reactive sends, flush, etc) - perhaps it would be easier to have a separate reactive template after all.

PR #341, opened 2017-06-05
GH-335: Fix Class Cast with Batch Filter [1.x.x]

### LINKED ISSUE #335: Filtering Messages don't support batch listener when using @KafkaListener
A ClassCastException will be thrown if setRecordFilterStrategy to batch-listener-container-factory and using `@KafkaListener`. 

And doc said that, "Finally, FilteringBatchMessageListenerAdapter and FilteringBatchAcknowledgingMessageListenerAdapter are provided, for when using a batch message listener." But there is no FilteringBatchAcknowledgingMessageListenerAdapter class at all.

garyrussell:
Fixes #335

Failed to check `batchListener` field when casting listener to apply filter.

Conflicts:
	spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
	spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
Resolved.

__cherry-pick to all 1.x__


artembilan:
... and cherry-picked to `1.1.x`.
There is no *batch* support in the `1.0.x`, so, nothing to backport.

PR #354, opened 2017-06-24
GH-352: Add HeaderMapper for 0.11.0.0 Headers

### LINKED ISSUE #352: Add Support for Message Headers (KIP-82)
https://cwiki.apache.org/confluence/display/KAFKA/KIP-82+-+Add+Record+Headers

Map to `spring-messaging.Message<?>` headers for `@KafkaListener`.

garyrussell:
Resolves #352

Initial Commit

TODO: fix up JAAS stuff for changes
TODO: Docs

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,11 +35,11 @@ allprojects {
 	group = 'org.springframework.kafka'
 
 	repositories {
-		mavenLocal()
 		maven { url 'https://repo.spring.io/libs-milestone' }
 		if (version.endsWith('BUILD-SNAPSHOT') || project.hasProperty('platformVersion')) {
 			maven { url 'https://repo.spring.io/libs-snapshot' }
 		}
+		maven { url 'https://repository.apache.org/content/groups/staging' }
 	}
 
 }
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,6 @@ allprojects {
 		if (version.endsWith('BUILD-SNAPSHOT') || project.hasProperty('platformVersion')) {
 			maven { url 'https://repo.spring.io/libs-snapshot' }
 		}
-		maven { url 'https://repository.apache.org/content/groups/staging' }
 	}
 
 }
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,7 @@ allprojects {
 	group = 'org.springframework.kafka'
 
 	repositories {
+		mavenLocal()
 		maven { url 'https://repo.spring.io/libs-milestone' }
 		if (version.endsWith('BUILD-SNAPSHOT') || project.hasProperty('platformVersion')) {
 			maven { url 'https://repo.spring.io/libs-snapshot' }
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
@@ -22,8 +22,10 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -41,7 +43,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
- * Header mapper for kafka.
+ * Header mapper for Apache Kafka.
  *
  * @author Gary Russell
  * @since 2.0
@@ -51,11 +53,19 @@ public class DefaultKafkaHeaderMapper implements HeaderMapper<Headers> {
 
 	private static final Log logger = LogFactory.getLog(DefaultKafkaHeaderMapper.class);
 
+	private static final List<String> DEFAULT_TRUSTED_PACKAGES =
+			Arrays.asList(
+					"java.util",
+					"java.lang"
+			);
+
 	private static final String JSON_TYPES = "spring_json_header_types";
 
 	private final ObjectMapper objectMapper;
 
 	private final List<SimplePatternBasedHeaderMatcher> matchers = new ArrayList<>();
+
+	private final Set<String> trustedPackages = new LinkedHashSet<>(DEFAULT_TRUSTED_PACKAGES);
 
 	/**
 	 * Construct an instance with the default object mapper and header patterns for
@@ -94,6 +104,28 @@ public class DefaultKafkaHeaderMapper implements HeaderMapper<Headers> {
 		Assert.notNull(patterns, "'patterns' must not be null");
 		this.objectMapper = objectMapper;
 		patterns.forEach(pattern -> this.matchers.add(new SimplePatternBasedHeaderMatcher(pattern)));
+	}
+
+	/**
+	 * Add packages to the trusted packages list (default {@code java.util, java.lang}) used
+	 * when constructing objects from JSON.
+	 * If any of the supplied packages is {@code "*"}, all packages are trusted.
+	 * If a class for a non-trusted package is encountered, the header is returned to the
+	 * application with value of type {@link NonTrustedHeaderType}.
+	 * @param trustedPackages the packages to trust.
+	 */
+	public void addTrustedPackages(String... trustedPackages) {
+		if (trustedPackages != null) {
+			for (String whiteList : trustedPackages) {
+				if ("*".equals(whiteList)) {
+					this.trustedPackages.clear();
+					break;
+				}
+				else {
+					this.trustedPackages.add(whiteList);
+				}
+			}
+		}
 	}
 
 	@Override
@@ -154,6 +186,7 @@ public class DefaultKafkaHeaderMapper implements HeaderMapper<Headers> {
 				catch (IOException e) {
 					logger.error("Could not decode json types: " + new String(next.value()), e);
 				}
+				break;
 			}
 		}
 		final Map<String, String> jsonTypes = types;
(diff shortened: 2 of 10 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -0,0 +1,232 @@
...
+	/**
+	 * Construct an instance with the provided object mapper and default header patterns
+	 * for outbound headers; all inbound headers are mapped. The patterns are applied in
+	 * order, stopping on the first match (positive or negative). Patterns are negated by
+	 * preceding them with "!". The default pattern list is
+	 * {@code "!id", "!timestamp" and "*"}.
+	 * @param objectMapper the object mapper.
+	 * @see org.springframework.util.PatternMatchUtils#simpleMatch(String, String)
+	 */
+	public DefaultKafkaHeaderMapper(ObjectMapper objectMapper) {
```
artembilan: Any particular explanation why did you decide to go with the JSON for non-byte[] values?
Can't we be just strict as JMS and support only specific types:
```
private static List<Class<?>> SUPPORTED_PROPERTY_TYPES = Arrays.asList(new Class<?>[] {
			Boolean.class, Byte.class, Double.class, Float.class, Integer.class, Long.class, Short.class, String.class });
```
I don't have right now to much arguments for objection, but just don't feel it right to always send JSON, when other side might not expect and would be pretty happy to deserialize `Long` from simply `byte[]` for the timestamp header.
Or the same UUID conversion:
```
ByteBuffer bb = ByteBuffer.wrap(new byte[16]);
bb.putLong(uuid.getMostSignificantBits());
bb.putLong(uuid.getLeastSignificantBits());
return bb.array();
```
Thanks

garyrussell: Several reasons:

1. Simple
1. Easy to detect it was originated by a Spring app (presence of JSON_TYPES header)
1. Easy to document what a non-Spring app would need to do.
1. If we supported multiple types, we'd need some way to identify the type to create on the consumer side (and do all that work ourselves rather than letting Jackson do it).
1. There is no standard way to convey that type information.
1. The user can use his own scheme using messaging Message<?> headers with `byte[]` values, which are sent unchanged, or by writing his own mapper.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -0,0 +1,229 @@
...
+	@Override
+	public MessageHeaders toHeaders(Headers source) {
+		final Map<String, Object> headers = new HashMap<String, Object>();
+		Map<String, String> types = null;
+		Iterator<Header> iterator = source.iterator();
+		while (iterator.hasNext()) {
+			Header next = iterator.next();
+			if (next.key().equals(JSON_TYPES)) {
+				try {
+					types = this.objectMapper.readValue(next.value(), HashMap.class);
```
garyrussell: Yes, `break` is missing.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
artembilan: OK. Thank you!

As I said: I don't feel it, but at the same time don't have anything to object...
It looks like as smart as it is simple.
Maybe one more confuse for me that we don't use similar approach in many other places... 😄

build.gradle
```diff
@@ -39,6 +39,7 @@ allprojects {
 		if (version.endsWith('BUILD-SNAPSHOT') || project.hasProperty('platformVersion')) {
 			maven { url 'https://repo.spring.io/libs-snapshot' }
 		}
+		maven { url 'https://repository.apache.org/content/groups/staging' }
```
artembilan: Since `0.11.0` is already [GA](https://github.com/apache/kafka/releases/tag/0.11.0.0),  we don't need this repo any more.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
```diff
@@ -66,13 +71,23 @@ public void setGenerateTimestamp(boolean generateTimestamp) {
 		this.generateTimestamp = generateTimestamp;
 	}
 
+	/**
+	 * Set the header mapper to map headers.
+	 * @param headerMapper the mapper.
+	 * @since 2.1
```
artembilan: So, is the decision to go with Kafka `0.11.0` in `2.1` anyway?
Thanks

garyrussell: I can't see a way out of that. Can you?

artembilan: Fine with me either way. We have postponed SI release anyway for some reason.
On the other hand this feature will be shipped only in the SCSt 2.0 anyway...

We might consider something like `1.3` Java 7 based and with the Kafka 0.11 but only with headers and admin client support.

We can definitely deffer transaction management to `2.1`.

Just some thought to come up with compromise...

garyrussell: Yes - 1.3 with headers, admin, and new threading from 2.0 (remove listener thread) - that was an early commit and might not be too hard to backport - it's the ConsumerAware stuff that's hard to backport.

artembilan:
I think this PR deserves Docs and CheckStyle fixes and will be ready for merge:
```
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java:62:9: Missing a Javadoc comment. [JavadocVariable]

[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaderMapper.java:45: Unused Javadoc tag. [JavadocMethod]


```

garyrussell:
Rebased; fixed JAAS stuff; changed the target to master.

build.gradle
artembilan: One more time: we don't need this repo any more.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -0,0 +1,338 @@
...
+					return true;
+				}
+			}
+			return false;
+		}
+		return true;
+	}
+
+	/**
+	 * A pattern-based {@link HeaderMatcher} that matches if the specified
```
artembilan: There is no such a class. The simple wording should be fine.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -0,0 +1,338 @@
...
+	}
+
+	protected boolean matches(String header) {
+		for (SimplePatternBasedHeaderMatcher matcher : this.matchers) {
+			if (matcher.matchHeader(header)) {
+				return !matcher.isNegated();
+			}
+		}
+		if (logger.isDebugEnabled()) {
+			logger.debug(MessageFormat.format("headerName=[{0}] WILL NOT be mapped {1}; matched no patterns", header));
```
artembilan: There is no second arg for ` {1}` placeholder.

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaderMapper.java
```diff
@@ -0,0 +1,49 @@
...
+	 * @param headers the abstracted MessageHeaders.
+	 * @param target the native target message.
+	 */
+	void fromHeaders(MessageHeaders headers, Headers target);
+
+	/**
+	 * Map from the given target message to abstracted {@link MessageHeaders}.
+	 * @param source the native target message.
+	 * @param target the target headers.
+	 * @return the abstracted MessageHeaders.
```
artembilan: There is no `return` in the contract of this method.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -0,0 +1,338 @@
...
+	 * Construct an instance with the provided object mapper and default header patterns
+	 * for outbound headers; all inbound headers are mapped. The patterns are applied in
+	 * order, stopping on the first match (positive or negative). Patterns are negated by
+	 * preceding them with "!". The patterns will replace the default patterns; you generally
+	 * should not map the {@code "id" and "timestamp"} headers.
+	 * @param objectMapper the object mapper.
+	 * @param patterns the patterns.
+	 * @see org.springframework.util.PatternMatchUtils#simpleMatch(String, String)
+	 */
+	public DefaultKafkaHeaderMapper(ObjectMapper objectMapper, List<String> patterns) {
```
artembilan: Seems for me we talked with you about simpler Java config-friendly vararg version.
Also why there is no ctor with the default `ObjectMapper` but custom patterns?

Also we might need to document that there are a bunch of headers which are not mapped either way.
And maybe mention that the custom headers should be used if there is a requirement to carry them any way.

WDYT?

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
artembilan: The version has not been fixed

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
garyrussell: Yes (simpler ctor).

I don't understand your other comment about headers not mapped etc.

artembilan: For example: `new SimplePatternBasedHeaderMatcher("!" + KafkaHeaders.TOPIC)`
And we have here: ` this.matchers.add(new SimplePatternBasedHeaderMatcher(pattern))`.
So, looks like there is no way to override defaults and, therefore, if we strictly need to set those header to the Kafka message, we should remap them to the custom headers.

garyrussell:
I will add docs tomorrow.

src/reference/asciidoc/kafka.adoc
```diff
@@ -1010,6 +1010,85 @@ When using a `@KafkaListener`, the parameter type is provided to the message con
...
+public interface Header {
+
+    String key();
+
+    byte[] value();
+
+}
+----
+
+The `KafkaHeaders` interface has two methods:
```
artembilan: `KafkaHeaderMapper` I guess.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/6755f5fa9c4c4dab672b5f1f18579676fc312eb3

Issue #362, opened 2017-07-06
Add Support For Transactions


garyrussell:
- `template.executeInTransaction(KafkaOperationsCallback callback)`
- `PlatformTransactionManager`
- `DefaultProducerFactory` to hand out a new Producer instead of a singleton (when transactions are enabled).

jorgheymans:
will this include honoring @Transactional as well ?

artembilan:
I think so. Since `@Transactional` starts a TX in the current thread, that won't be surprise from the producer side to pick up that one and participate in it.

Looks like this issue duplicates #258

garyrussell:
@jorgheymans Yes - providing the implementation of `PlatformTransactionManager` provides that functionality.

cortexcompiler:
Is there a recommended way to set the "transactionIdPrefix" on the DefaultkafkaProducerFactory?  The documentation that was added does not appear to get into that detail.

artembilan:
A `transactionIdPrefix` is exactly about `ProducerConfig.TRANSACTIONAL_ID_CONFIG`: https://kafka.apache.org/documentation/#producerconfigs:

> The TransactionalId to use for transactional delivery. This enables reliability semantics which span multiple producer sessions since it allows the client to guarantee that transactions using the same TransactionalId have been completed prior to starting any new transactions. If no TransactionalId is provided, then the producer is limited to idempotent delivery. Note that enable.idempotence must be enabled if a TransactionalId is configured. The default is empty, which means transactions cannot be used.

So, if you familiar with that, it should not be so complicated to determine that this `transactionIdPrefix` is for distinguishing one transactional producer from another on the Broker level during troubleshooting.

Therefore you can set it into something like `myAwesomeProducer` on one `DefaultKafkaProducerFactory` and `theirBadProducer` on another. Or this can be based on the IP or other resource to determine the location of your producer. You can build any arbitrary string for this property.

Hope that helps

garyrussell:
I am not sure what you mean by 

>The documentation that was added does not appear to get into that detail.

See [the documentation](https://docs.spring.io/spring-kafka/docs/2.0.0.BUILD-SNAPSHOT/reference/html/_reference.html#transactions).

[...]

PR #363, opened 2017-07-06
GH-362: Add 0.11 Transaction Support

garyrussell:
Initial commit

Added transactions to container.

TODO: More tests.

See #362


build.gradle
```diff
@@ -141,6 +141,7 @@ project ('spring-kafka') {
 
 	dependencies {
 		compile "org.springframework:spring-messaging:$springVersion"
+		compile "org.springframework:spring-tx:$springVersion"
```
artembilan: `optional`? 
Maybe not everybody wants transactions...

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
```diff
@@ -158,6 +158,17 @@
 	<T> T execute(ProducerCallback<K, V, T> callback);
 
 	/**
+	 * Execute some arbitrary operation(s) on the operations and return the result.
+	 * The operations are invoked within a local transaction and do not participate
+	 * in a global transaction (if present).
+	 * @param callback the callback.
+	 * @param <T> the result type.
+	 * @return the result.
+	 * @since 1.1
```
artembilan: `@since 2.1` ? Or still `2.0` as you have on the `OperationsCallback` below

spring-kafka/src/main/java/org/springframework/kafka/transaction/KafkaTransactionManager.java
```diff
@@ -0,0 +1,225 @@
...
+ *
+ * <p>
+ * This local strategy is an alternative to executing Kafka operations within, and
+ * synchronized with, external transactions. This strategy is <i>not</i> able to provide
+ * XA transactions, for example in order to share transactions between messaging and
+ * database access.
+ *
+ * <p>
+ * Application code is required to retrieve the transactional Kafka resources via
+ * {@link TransactionUtils#getTransactionalResourceHolder(ProducerFactory, boolean)}.
```
artembilan: `ProducerFactoryUtils` ?

spring-kafka/src/main/java/org/springframework/kafka/transaction/KafkaTransactionManager.java
```diff
@@ -0,0 +1,225 @@
...
+	protected boolean isExistingTransaction(Object transaction) {
+		@SuppressWarnings("unchecked")
+		KafkaTransactionObject<K, V> txObject = (KafkaTransactionObject<K, V>) transaction;
+		return (txObject.getResourceHolder() != null);
+	}
+
+	@Override
+	protected void doBegin(Object transaction, TransactionDefinition definition) {
+		if (definition.getIsolationLevel() != TransactionDefinition.ISOLATION_DEFAULT) {
+			throw new InvalidIsolationLevelException("AMQP does not support an isolation level concept");
```
artembilan: s/AMQP/Apache Kafka ?

spring-kafka/src/main/java/org/springframework/kafka/transaction/KafkaTransactionManager.java
```diff
@@ -0,0 +1,225 @@
...
+		@SuppressWarnings("unchecked")
+		KafkaTransactionObject<K, V> txObject = (KafkaTransactionObject<K, V>) transaction;
+		txObject.setResourceHolder(null);
+		return TransactionSynchronizationManager.unbindResource(getProducerFactory());
+	}
+
+	@Override
+	protected void doResume(Object transaction, Object suspendedResources) {
+		@SuppressWarnings("unchecked")
+		KafkaResourceHolder<K, V> conHolder = (KafkaResourceHolder<K, V>) suspendedResources;
```
artembilan: Nit-pick, of course, but it's not about *connection*. So, just `holder` or `resourceHolder`

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -596,110 +617,208 @@ private void invokeListener(final ConsumerRecords<K, V> records) {
...
+			}
+		}
+
+		private void invokeBatchListenerInTx(final ConsumerRecords<K, V> records,
+				List<ConsumerRecord<K, V>> recordList, @SuppressWarnings("rawtypes") Producer producer) {
+			try {
+				this.transactionTemplate.execute(s -> {
+					RuntimeException aborted = doInvokeBatchListener(records, recordList, producer);
+					if (aborted != null) {
+						throw aborted;
```
artembilan: Can't we use `TransactionCallbackWithoutResult` instead and throw exceptions as is in the target `doInvokeBatchListener()` ?

src/reference/asciidoc/kafka.adoc
```diff
@@ -149,6 +149,55 @@ You may wish to invoke `flush()` before waiting or, for convenience, the templat
 parameter which will cause the template to `flush()` on each send.
 Note, however that flushing will likely significantly reduce performance.
 
+[[transactions]]
+===== Transactions
+
+The 0.11.0.0 client library added support for transactions.
+Spring for Apache Kafka adds support in several ways.
+
+- `KafkaTransactionManager` - used with normal Spring transaction support (`@Transactional`, `TransactionTemplate` etc).
```
jorgheymans: is the more "fancy" transactional stuff like PROPAGATION_REQUIRES_NEW and PROPAGATION_NESTED also supported ?

garyrussell: REQUIRES_NEW should work, but not NESTED.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: `TransactionCallbackWithoutResult` is not a functional interface.

artembilan: And why do you care about functional interface here, in the internals?
We don't provide any end-user API in this point, so, I'm pretty fine do not have `->` operator on this level.

Although to be honest I'm more curious why do you return `RuntimeException` but not throw it there as is?
Thanks

garyrussell: Because, when we're not running in a transaction, we don't want to throw the exception; I suppose I could add `if (producer != null)` and throw instead of returning; let me know if you prefer that.

artembilan: OK. Good. Thank you for explanation.
I will jump to this story thoroughly on Monday .

Although what you say about `producer` is legitimate and might be supplied with some comment regarding tx/non-tx behavior.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -603,101 +636,239 @@ private void invokeBatchListener(final ConsumerRecords<K, V> records) {
...
-							break;
-					}
-					if (!this.isAnyManualAck && !this.autoCommit) {
-						this.acks.add(record);
-					}
+					this.transactionTemplate.execute(s -> {
+						Producer producer = null;
+						if (this.kafkaTxManager != null) {
+							producer = ((KafkaResourceHolder) TransactionSynchronizationManager
+									.getResource(this.kafkaTxManager.getProducerFactory())).getProducer();
```
artembilan: How do transactions work in the container if we don't use `kafkaTxManager`?

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -283,8 +329,23 @@ public void onCompletion(RecordMetadata metadata, Exception exception) {
...
+
 	private Producer<K, V> getTheProducer() {
-		return this.producerFactory.createProducer();
+		if (this.transactional) {
+			Producer<K, V> producer = this.producers.get();
+			if (producer != null) {
+				return producer;
+			}
+			KafkaResourceHolder<K, V> holder = ProducerFactoryUtils
+					.getTransactionalResourceHolder(this.producerFactory);
```
garyrussell: I don't think we need to add that complication - the user can provide a separate producer factory and template for non-transactional work.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: The Kafka transaction is synchronized with the external TX.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -603,101 +636,239 @@ private void invokeBatchListener(final ConsumerRecords<K, V> records) {
...
+				}
 			}
+			return null;
+		}
+
+		@SuppressWarnings({ "unchecked", "rawtypes" })
+		protected void sendOffsetsToTransaction(Producer producer) {
+			handleAcks();
+			Map<TopicPartition, OffsetAndMetadata> commits = buildCommits();
+			producer.sendOffsetsToTransaction(commits, this.consumerGroupId);
```
garyrussell: That condition is not possible - we never invoke a listener with no records.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: They can either use a chained transaction manager or annotate the listener with `@Transactional(txm="ktm")` and the transaction will be sync'd.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: Well, I'm talking here about error use-case.
We send `acks` for commit only if we have `ackOnError`, but we call this method in case of exception unconditionally. Therefore when `ackOnError == false`, the `commits` is empty.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: Well, either way we don't have `producer` and, therefore, don't `sendOffsetsToTransaction()`.
So, I don't know what "sync'd" means here then...
Or am I missing anything else?

I think we can somehow reuse `ProducerFactoryUtils.getTransactionalResourceHolder()` for any other TX Managers.

But I agree that it might be postponed to separate issue.

garyrussell: Ah, yes; I see; I just thought it would "magically" work, but I was wrong (just ran a test).

I was thinking the user will be responsible to send the offset(s) to the transaction...

		props.setMessageListener((MessageListener<String, String>) m -> {
			transactionTemplate.execute(s -> {
				template.send("bar", "baz");
				Producer txProd = ((KafkaResourceHolder) TransactionSynchronizationManager
						.getResource(pf)).getProducer();
				txProd.sendOffsetsToTransaction(Collections.singletonMap(new TopicPartition(m.topic(), m.partition()),
						new OffsetAndMetadata(m.offset())), "group");
				return null;
			});
		});

But it doesn't work - anyway, yes; let's defer it to another PR.

artembilan:
Gary, I'm not sure what have been broken since your latest change, but that feels for me more like a race condition:
```
org.springframework.kafka.listener.TransactionalContainerTests > testConsumeAndProduceTransactionRollback FAILED

    org.mockito.exceptions.verification.VerificationInOrderFailure at TransactionalContainerTests.java:197
```

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/d2401f1a1a09931cfd69926488d7124882d3e233

Issue #373, opened 2017-07-17
Support External TX Manager and Transaction Synchronization


garyrussell:
https://github.com/spring-projects/spring-kafka/pull/363#discussion_r127779342

PR #375, opened 2017-07-18
GH-362: Transactional Container Polishing

garyrussell:
Avoid:

    17:24:19.706 WARN  [kafka-request-handler-0][kafka.coordinator.group.GroupMetadataManager]
    [Group Metadata Manager on Broker 1]: group: group with leader:
    consumer-1-7a4aa847-0ab2-409f-b022-87ff0766d8f7 has received offset commits from consumers
    as well as transactional producers.
    Mixing both types of offset commits will generally result in surprises and should be avoided.

- Commit the initial offset after partition assignment in a transaction.
- Use `TransactionCallbackWithoutResult`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -55,6 +55,8 @@ import org.springframework.kafka.support.TopicPartitionInitialOffset.SeekPositio
 import org.springframework.kafka.transaction.KafkaTransactionManager;
 import org.springframework.scheduling.SchedulingAwareRunnable;
 import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.TransactionStatus;
+import org.springframework.transaction.support.TransactionCallbackWithoutResult;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.springframework.transaction.support.TransactionTemplate;
 import org.springframework.util.Assert;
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -427,7 +429,21 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
...
 						}
-						if (KafkaMessageListenerContainer.this.getContainerProperties().isSyncCommits()) {
+						if (ListenerConsumer.this.transactionTemplate != null) {
+							ListenerConsumer.this.transactionTemplate.execute(new TransactionCallbackWithoutResult() {
+
+								@SuppressWarnings({ "unchecked", "rawtypes" })
+								@Override
+								protected void doInTransactionWithoutResult(TransactionStatus status) {
+									((KafkaResourceHolder) TransactionSynchronizationManager
+											.getResource(ListenerConsumer.this.kafkaTxManager.getProducerFactory()))
```
artembilan: The `ListenerConsumer.this.transactionTemplate` might not be null, but `kafkaTxManager` can be. Because we might use some different impl.
See #373

I think we should add `&& ListenerConsumer.this.kafkaTxManager != null` to the `if`.

Let me know and I'll fix on merge!

artembilan:
Do you know what is that?
```
18:08:45.857 ERROR [ReplicaFetcherThread-0-0][kafka.server.ReplicaFetcherThread] [ReplicaFetcherThread-0-0]: Error for partition [__transaction_state,2] to broker 0:org.apache.kafka.common.errors.UnknownTopicOrPartitionException: This server does not host this topic-partition.
```
I have a lot of this errors (and had before) in the `KafkaTemplateTransactionTests` and something similar in the `TransactionalContainerTests.`

Thanks

P.S. On Windows though

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: OK; thanks.

garyrussell:
Those errors appear while the 3 brokers are being shut down (we need at least 3 brokers for transactions). I have been ignoring them.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/51bebac6c3d0612677e4b6260491001180a8d29d

Issue #380, opened 2017-07-31
Adding a timeout parameter to record-getting test utils


hgwood:
Hello, 

I got my first end-to-end Kafka Stream test to run successfully today! Thanks a lot for making Spring Kafka! 💯 

I have a suggestion. `getRecords` and `getSingleRecord` from `KafkaTestUtils` currently enforce a timeout of 1 minute when polling the consumer. This means that when a test expects messages to be there but none are (because the tested code is not correct), then it hangs for 1 minute before it fails.

If a `timeout` parameter was added to these methods, a test could specify a shorter timeout and fail faster. To ensure backwards compatibility, the current signatures could remain as overloads of the new methods.

What do you think? I have PR ready if you are interested.

artembilan:
Ship it, @hgwood !

Thank you!

PR #381, opened 2017-07-31
GH-380: Add timeout parameter to record-getting test utils

hgwood:
This parameter is useful when the test expects messages to be there but they
are not because the tested code is not correct. In this case the test will
run for a minute waiting for a message before it fails. With a custom timeout,
a test can specify a shorter timeout to fail faster.

Closes GH-380.

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
@@ -108,30 +108,46 @@ public final class KafkaTestUtils {
 	}
 
 	/**
+	 * Same as {@link #getSingleRecord(Consumer, String, long)} with a timeout of 60 seconds.
+	 */
+	public static <K, V> ConsumerRecord<K, V> getSingleRecord(Consumer<K, V> consumer, String topic) {
+		return getSingleRecord(consumer, topic, 60000);
+	}
+
+	/**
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@hgwood Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=381)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=381) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@hgwood Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=381)!

spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
```diff
@@ -108,30 +108,46 @@ private KafkaTestUtils() {
 	}
 
 	/**
+	 * Same as {@link #getSingleRecord(Consumer, String, long)} with a timeout of 60 seconds.
```
artembilan: Well, JavaDocs are not arbitrary text.
We don't tell `Same as` from the beginning.
Would be better to copy/paste JavaDocs from existing methods and add `@see` in the and to refer user to those siblings.
And `@since 2.0` to the new methods.

Plus you name, please, into the `@author` list.

Issue #378, opened 2017-07-28
Reconsider consumer group generation


artembilan:
Right now we have an option on the `@KafkaListener`:
```
/**
 * When {@link #groupId() groupId} is not provided, use the {@link #id() id} (if
 * provided) as the {@code group.id} property for the consumer. Set to false, to use
 * the {@code group.id} from the consumer factory.
 * @return false to disable.
 * @since 2.0
 */
boolean idIsGroup() default true;
```

So, when don't change it and don't provide the `id()` we end up with this:
```
private String getEndpointId(KafkaListener kafkaListener) {
	if (StringUtils.hasText(kafkaListener.id())) {
		return resolve(kafkaListener.id());
	}
	else {
		return "org.springframework.kafka.KafkaListenerEndpointContainer#" + this.counter.getAndIncrement();
	}
}
```
A bit later we have this logic in the `DefaultKafkaConsumerFactory`:
```
if (groupId != null) {
	modifiedConfigs.put(ConsumerConfig.GROUP_ID_CONFIG, groupId);
}
```
From here we have the situation when we override provide `group.id` with the generate bean name.

When we have several consumer applications and would like to use different groups for the same topic, we end up with the single suspicious group and undesired rebalance.

This is a new behavior in the Spring Kafka `2.0` and it is a breaking change for existing apps. 
They have to specify `idIsGroup = false` to reinstate stable logic.

Consider to make `idIsGroup()` as `false` by default, since there is no strong requirement to rely on the `id`. 

Another option do not use bean name for group id if it is generate and not explicit on the `@KafkaListener#id()`.

PR #383, opened 2017-08-08
GH-378: @KafkaListener: Fix Group Id When No id

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/378

Don't use a generated container id as the group id if no `id` property provided.

Also fixes `sendToTransaction()` by always populating `consumerGroupId` regardless of source.

__cherry-pick to 1.3.x__

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java
@@ -74,7 +74,6 @@ import scala.collection.Set;
  * @author Artem Bilan
  * @author Gary Russell
  */
-@SuppressWarnings("serial")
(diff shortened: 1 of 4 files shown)


artembilan:
@garyrussell ,

LGTM, but I would like to understand your comment:

> Also fixes sendToTransaction() by always populating consumerGroupId regardless of source.

Looks like it always was like:
```
.getProducer().sendOffsetsToTransaction(offsets,															ListenerConsumer.this.consumerGroupId);
```

garyrussell:
The problem was that for listener with no `groupId` property (used the group if from the consumer configs), that field was `null` 😦 .

garyrussell:
Are you cherry-picking? Or shall I?

artembilan:
Cherry-picked to `1.3.x` as https://github.com/spring-projects/spring-kafka/commit/9573d3cda0298d1f76642bb8ab0823d7784ebaa2

Issue #342, opened 2017-06-05
Listener Error Handler When No @SendTo


garyrussell:
Currently, we enforce the presence of a `KafkaTemplate` if an error handler returns a result.

The assertion should only be performed if there is a `@SendTo` on the listener method.

We should log (warn? info?) and drop returned objects if the method is not configured for request/reply.

Docs still say returned objects are always ignored.

PR #385, opened 2017-08-08
GH-342: Allow KLErrorHandler to Return a Result

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/342

Allow no reply template when an error handler returns a result; the result is ignored.

Also add a note (and test) about a `void` method that has a `@SendTo` - which allows information
about the exception in a `void` method to be sent to a topic.

### DIFF
--- a/src/reference/asciidoc/kafka.adoc
+++ b/src/reference/asciidoc/kafka.adoc
@@ -716,7 +716,7 @@ public Collection<String> replyingBatchListener(List<String> in) {
     ...
 }
 
-@KafkaListener(topics = "annotated23")
+@KafkaListener(topics = "annotated23", errorHandler = "replyErrorHandler")
 @SendTo("annotated23reply") // static reply topic definition
 public String replyingListenerWithErrorHandler(String in) {
     ...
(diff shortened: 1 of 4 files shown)


src/reference/asciidoc/kafka.adoc
```diff
@@ -761,6 +761,28 @@ public KafkaTemplate<String, String> myReplyingTemplate() {
...
 
+NOTE: You can annotate a `@KafkaListener` method with `@SendTo` even if no result is returned.
+This is to allow the configuration of an `errorHandler` that can forward information about a failed message delivery to some topic.
+
+[source, java]
+----
+@KafkaListener(id = "voidListenerWithReplyingErrorHandler", topics = "someTopic",
+        errorHandler = "voidSendToErrorHandler")
+@SendTo("failures")
+public void voidListenerWithReplyingErrorHandler(String in) {
```
artembilan: This is a bit unusual behavior and isn't self-explanatory, but having the use-case description here in Docs, I'm fine to merge.

And I will hope no one will follow this pattern. 
That is really much organic to use `KafkaTemplate` in the `ErrorHandler` if we are going to send error result somewhere and our original `@KafkaListener` is just `void`. Just IMO 😄 

Merging...

garyrussell: I agree; I just realized while fixing the problem that it was possible; I saw no reason to disable it, though.

PR #387, opened 2017-08-09
GH-373: Transaction Synchronization

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/373

Implement Spring transaction synchronization.

__cherry-pick to 1.3.x__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -47,6 +47,7 @@ import org.apache.kafka.common.errors.WakeupException;
 import org.springframework.core.task.SimpleAsyncTaskExecutor;
 import org.springframework.kafka.core.ConsumerFactory;
 import org.springframework.kafka.core.KafkaResourceHolder;
+import org.springframework.kafka.core.ProducerFactoryUtils;
 import org.springframework.kafka.event.ListenerContainerIdleEvent;
 import org.springframework.kafka.listener.ConsumerSeekAware.ConsumerSeekCallback;
 import org.springframework.kafka.listener.config.ContainerProperties;
@@ -534,6 +535,9 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 			if (this.genericListener instanceof ConsumerSeekAware) {
 				((ConsumerSeekAware) this.genericListener).registerSeekCallback(this);
 			}
+			if (this.transactionManager != null) {
+				ProducerFactoryUtils.setContainerGroupId(this.consumerGroupId);
+			}
 			this.count = 0;
 			this.last = System.currentTimeMillis();
 			if (isRunning() && this.definedPartitions != null) {
@@ -589,6 +593,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 					}
 				}
 			}
+			ProducerFactoryUtils.clearContainerGroupId();
 			if (!this.fatalError) {
 				if (this.kafkaTxManager == null) {
 					commitPendingAcks();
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -534,6 +535,9 @@ public void run() {
 			if (this.genericListener instanceof ConsumerSeekAware) {
 				((ConsumerSeekAware) this.genericListener).registerSeekCallback(this);
 			}
+			if (this.transactionManager != null) {
+				ProducerFactoryUtils.setContainerGroupId(this.consumerGroupId);
```
artembilan: Why do you call it as `container` and use `consumer`? 
Looks like inconsistency and may confuse end-user since containers in the same group may have different consumer group.

spring-kafka/src/test/java/org/springframework/kafka/listener/TransactionalContainerTests.java
```diff
@@ -199,6 +203,65 @@ public void testConsumeAndProduceTransactionRollback() throws Exception {
...
+		final ProducerFactory pf = mock(ProducerFactory.class);
+		given(pf.transactionCapable()).willReturn(true);
+		given(pf.createProducer()).willReturn(producer);
+		ContainerProperties props = new ContainerProperties("foo");
+		props.setGroupId("group");
+		props.setTransactionManager(new SomeOtherTransactionManager());
+		final KafkaTemplate template = new KafkaTemplate(pf);
+		props.setMessageListener((MessageListener<String, String>) m -> {
+			template.send("bar", "baz");
+			template.sendOffsetsToTransaction(Collections.singletonMap(new TopicPartition(m.topic(), m.partition()),
```
artembilan: I wonder if we shouldn't require this hack from end-user and just continue to let to call `Acknowledgment.acknowledge()`. And already that one will be tied to the TX and responsible to the proper `sendOffsetsToTransaction`.

With that approach we even might not need that `groupIds` `ThreadLocal`...

WDYT?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Good point.

artembilan:
Do we still need to worry about distinguishing `KafkaTransactionManager` in the `ListenerContainer`? I don't see any changes on the matter there.
I thought after making the TX sync proper we won't need to worry about the TX Manager type.
No? What am I missing?
Thanks

garyrussell:
It depends on the answer to your other suggestion (which I need to think about) - we need to know whether or not to send the offsets.

spring-kafka/src/test/java/org/springframework/kafka/listener/TransactionalContainerTests.java
garyrussell: The problem is that when the TXM is not Kafka, we don't have a reference to the `ProducerFactory` which is the key for the bound resource. We need that resource to get access to the producer to send the offsets.

artembilan: OK. I see. Thanks.
Well, I guess I can live with this approach then.
Let me take a look into your description in Docs closely!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: OK. Everything is good. 
I get back my suggestion about `Acknowledge`.
Indeed we need that `Producer` resource for transactions...

Only this issue about container - consumer inconsistencty.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/36f836f07b20ceb7c94f6619c867c80488f2ce15

artembilan:
... and cherry-picked to `1.3.x`

Issue #388, opened 2017-08-09
Add KafkaAdmin


garyrussell:
- auto declare `NewTopic`s defined in the application context
- see [KIP-117](https://cwiki.apache.org/confluence/display/KAFKA/KIP-117%3A+Add+a+public+AdminClient+API+for+Kafka+admin+operations)

artembilan:
Reopen because of: https://github.com/spring-projects/spring-kafka/pull/389#discussion_r132443329

artembilan:
Fixed via: https://github.com/spring-projects/spring-kafka/pull/393

PR #389, opened 2017-08-09
GH-388: Add KafkaAdmin (Declarative Topic Create)

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/388

Initial Commit

TODO: Docs

/cc @sobychacko

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
@@ -1,0 +1,205 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.core;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.kafka.clients.admin.AdminClient;
+import org.apache.kafka.clients.admin.CreateTopicsResult;
+import org.apache.kafka.clients.admin.DescribeTopicsResult;
+import org.apache.kafka.clients.admin.NewTopic;
+import org.apache.kafka.clients.admin.TopicDescription;
+import org.apache.kafka.common.acl.AclBinding;
+
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.DisposableBean;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.context.SmartLifecycle;
+
+/**
+ * An admin that delegates to an {@link AdminClient} to create topics defined
+ * in the application context.
+ *
+ * @author Gary Russell
+ * @since 1.3
+ *
+ */
+public class KafkaAdmin implements ApplicationContextAware, SmartLifecycle, DisposableBean {
+
+	private static final int DEFAULT_CLOSE_TIMEOUT = 10;
+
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -0,0 +1,205 @@
...
+				}
+				catch (ExecutionException e) {
+					logger.error("Failed to create topics", e.getCause());
+				}
+			}
+		}
+	}
+
+	private void addAclBindingsIfNeeded(Collection<AclBinding> bindings) {
+		// TODO (or not)
```
artembilan: I would leave it without any ACL auto-binding. That looks like rare task and would be possible via `AdminClient`.
Creating topics by the application context is really something people like from Spring AMQP.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -0,0 +1,205 @@
...
+	 */
+	public void setCloseTimeout(int closeTimeout) {
+		this.closeTimeout = closeTimeout;
+	}
+
+	@Override
+	public synchronized void start() {
+		if (!this.running) {
+			if (this.applicationContext != null) {
+				addTopicsIfNeeded(this.applicationContext.getBeansOfType(NewTopic.class, false, false).values());
```
artembilan: Won't `SmartInitializingSingleton` work better here instead?
At least I don't understand the reason of the `setRunning()` API ...

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -0,0 +1,205 @@
...
+
+	@Override
+	public void destroy() throws Exception {
+		this.adminClient.close(this.closeTimeout, TimeUnit.SECONDS);
+	}
+
+	private void addTopicsIfNeeded(Collection<NewTopic> topics) {
+		if (topics.size() > 0) {
+			Map<String, NewTopic> topicNameToTopic = new HashMap<>();
+			topics.stream().forEach(t -> topicNameToTopic.compute(t.name(), (k, v) -> v = t));
```
garyrussell: It doesn't need to be for master; that's why I didn't say "cherry-pick"; I will create a 1.3.x PR after merge.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
garyrussell: Well; I thought `SmartLifecycle` is better because it will allow users to control it themselves. For example; if they want the context to start if the broker is not available, they can set `autoStartup` to `false` and `start()` themselves. With `SmartInitializingSingleton` the context will fail to start.

garyrussell: Actually; no; the admin is created in the ctor 😦 - need to think.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -10,6 +10,12 @@ See <<transactions>> for more information.
 The 0.11.0.0 client library added support for message headers; these can now be mapped to/from `spring-messaging` `MessageHeaders`.
 See <<headers>> for more information.
 
+==== Creating Topics
+
+The 0.11.0.0 client library provides an `AdminClient` which can be used to create topics.
+The `KafkaAdmin` uses this client to automatically add topics defined as `@Bean` s; see <<kafka-admin>> for more information.
```
artembilan: There is no `kafka-admin` section. 
Will fix on merge.

spring-kafka/src/test/java/org/springframework/kafka/core/KafkAdminTests.java
```diff
@@ -0,0 +1,87 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Gary Russell
+ * @since 1.3
+ *
+ */
+@RunWith(SpringRunner.class)
+@DirtiesContext
+public class KafkAdminTests {
```
artembilan: Typo: Kafk**a**AdminTests

src/reference/asciidoc/kafka.adoc
```diff
@@ -2,6 +2,38 @@
...
+public KafkaAdmin admin() {
+    Map<String, Object> configs = new HashMap<>();
+    configs.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG,
+            StringUtils.arrayToCommaDelimitedString(kafkaEmbedded().getBrokerAddresses()));
+    return new KafkaAdmin(configs);
+}
+
+@Bean
+public NewTopic topic1() {
+    return new NewTopic("foo", 10, (short) 2);
```
artembilan: This is silly to have such a high-level API requiring the cast to `short`, but I think we don't have choice and will live with this like with all the inconvenience in Kafka 😄

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/325d5bdf777dbdb62243ad59e9e509721bdb5acd

Issue #391, opened 2017-08-10
BatchMessagingMessageConverter Doesn't Map Kafka Headers


garyrussell:
Title says it all - I am not sure how to deal with it.

Issue #390, opened 2017-08-10
MessagingMessageConverter has a hard dependency on Jackson Via the Header Mapper


garyrussell:
Only instantiate `DefaultKafkaHeaderMapper` if Jackson is on the class path.

PR #392, opened 2017-08-10
GH-390: GH-391: Header Mapper Fixes

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/390
Fixes https://github.com/spring-projects/spring-kafka/issues/391

The `DefaultKafkaHeaderMapper` has a hard dependency on Jackson; don't use it as a default
if Jackson is not present.

Add header mapping for batch listeners.

__cherry-pick to 1.3.x__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaders.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaders.java
@@ -94,4 +94,17 @@ public abstract class KafkaHeaders {
 	 */
 	public static final String RECEIVED_TIMESTAMP = PREFIX + "receivedTimestamp";
 
+	/**
+	 * The header for holding the native headers of the consumer record; only provided
+	 * if no header mapper is present.
+	 */
+	public static final String NATIVE_HEADERS = PREFIX + "nativeHeaders";
+
+	/**
+	 * The header for a list of Maps of converted native Kafka headers. Used for batch
+	 * listeners; the map at a particular list position corresponds to the data in the
+	 * payload list position.
+	 */
+	public static final String BATCH_CONVERTED_HEADERS = PREFIX + "convertedHeaders";
+
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaders.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaders.java
@@ -105,6 +105,6 @@ public abstract class KafkaHeaders {
 	 * listeners; the map at a particular list position corresponds to the data in the
 	 * payload list position.
 	 */
-	public static final String BATCH_CONVERTED_HEADERS = PREFIX + "convertedHeaders";
+	public static final String BATCH_CONVERTED_HEADERS = PREFIX + "batchConvertedHeaders";
 
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
@@ -18,14 +18,21 @@ package org.springframework.kafka.support.converter;
 
 import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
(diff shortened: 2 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
```diff
@@ -106,6 +140,20 @@ public void setGenerateTimestamp(boolean generateTimestamp) {
 			offsets.add(record.offset());
 			timestampTypes.add(record.timestampType().name());
 			timestamps.add(record.timestamp());
+			if (this.headerMapper != null) {
+				Map<String, Object> converted = new HashMap<>();
+				this.headerMapper.toHeaders(record.headers(), converted);
+				convertedHeaders.add(converted);
+			}
+			else {
+				if (record.headers().iterator().hasNext() && logger.isWarnEnabled()) {
```
artembilan: So, we see the WARN only if we have entries in the `record.headers()`, but at the same time we populate the `KafkaHeaders.NATIVE_HEADERS` anyway, independently of the content.

Does it make sense to restrict warn then?

src/reference/asciidoc/kafka.adoc
```diff
@@ -1110,7 +1110,11 @@ You can trust other (or all) packages by adding trusted packages using the `addT
 If you are receiving messages from untrusted sources, you may wish to add just those packages that you trust.
 To trust all packages use `mapper.addTrustedPackages("*")`.
 
-The `DefaultKafkaHeaderMapper` is used in the `MessagingMessageConverter` by default.
+The `DefaultKafkaHeaderMapper` is used in the `MessagingMessageConverter` and `BatchMessagingMessageConverter` by default, as long as Jackson is on the class path.
+
+With the batch converter, the converted headers are available in the `KafkaHeaders.BATCH_CONVERTED_HEADERS` as a `List<Map<String, Object>` where the map in a position of the list corresponds to the data position in the payload.
```
artembilan: Missed closing `>` in the generics definition.

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaders.java
```diff
@@ -94,4 +94,17 @@
...
+	 * if no header mapper is present.
+	 */
+	public static final String NATIVE_HEADERS = PREFIX + "nativeHeaders";
+
+	/**
+	 * The header for a list of Maps of converted native Kafka headers. Used for batch
+	 * listeners; the map at a particular list position corresponds to the data in the
+	 * payload list position.
+	 */
+	public static final String BATCH_CONVERTED_HEADERS = PREFIX + "convertedHeaders";
```
garyrussell: Good catch; I changed this after deciding to add native headers when no mapper.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
garyrussell: Yes; we can probably make this DEBUG now - originally I had just the warn and dropped the headers.

artembilan: Yes, agree. Thinking even more: we might should consider to log it only once per batch.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
```diff
@@ -106,6 +141,21 @@ public void setGenerateTimestamp(boolean generateTimestamp) {
 			offsets.add(record.offset());
 			timestampTypes.add(record.timestampType().name());
 			timestamps.add(record.timestamp());
+			if (this.headerMapper != null) {
+				Map<String, Object> converted = new HashMap<>();
+				this.headerMapper.toHeaders(record.headers(), converted);
+				convertedHeaders.add(converted);
+			}
+			else {
+				if (record.headers().iterator().hasNext() && logger.isDebugEnabled() && !logged) {
```
artembilan: Gary, how about do not have this `record.headers().iterator().hasNext()`?
Since we call `natives.add(record.headers())` anyway we might log unconditionally, too

WDYT?

garyrussell: Makes sense; this code was also from before (when I warned about dropping headers).

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/1b284f685d43c292039e45c67da319f7899fa847 and cherry-picked to `1.3.x` as https://github.com/spring-projects/spring-kafka/commit/324c4842484a18a59077b013310e3efb13220045.

PR #394, opened 2017-08-11
GH-352: Add General send() with ProducerRecord

garyrussell:
Needed to support headers.

Adding overloaded send methods is getting out of hand, consider deprecating the parameterized sends.

__cherry-pick to 1.3.x__


artembilan:
LGTM.
Merging...

> Adding overloaded send methods is getting out of hand, 

Not sure where you see the problem. The `JdbcOperations` has at about 20 `query()` overloaded methods...

artembilan:
... and cherry-picked to `1.3.x`

Issue #395, opened 2017-08-14
Add Documentation for Async Template Usage


garyrussell:
SCSt needs the ability to correlate exceptions to generate an `ErrorMessage` for a sent message.

The current `ProducerListener` only provides key and data from the converterd `ProducerRecord`.

PR #396, opened 2017-08-14
GH-395: Document Async Template Usage

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/395

SCSt needs a reference to the sent message in order to generate an error message on a send failure.

### DIFF
--- a/src/reference/asciidoc/kafka.adoc
+++ b/src/reference/asciidoc/kafka.adoc
@@ -39,6 +39,8 @@ NOTE: The admin does not alter existing topics; it will log (INFO) if the number
 [[kafka-template]]
 ===== KafkaTemplate
 
+====== Overview
+
 The `KafkaTemplate` wraps a producer and provides convenience methods to send data to kafka topics.
 
 [source, java]
@@ -181,6 +183,50 @@ If you wish to block the sending thread, to await the result, you can invoke the
 You may wish to invoke `flush()` before waiting or, for convenience, the template has a constructor with an `autoFlush`
 parameter which will cause the template to `flush()` on each send.
 Note, however that flushing will likely significantly reduce performance.
+
+====== Examples
+
+.Non Blocking (Async)
+[source, java]
+----
+public void sendToKafka(final MyOutputData data) {
+    final ProducerRecord<String, String> record = createRecord(data);
+
+    ListenableFuture<SendResult<Integer, String>> future = template.send(record);
+    future.addCallback(new ListenableFutureCallback<SendResult<Integer, String>>() {
+
+        @Override
+        public void onSuccess(SendResult<Integer, String> result) {
+            handleSuccess(data);
+        }
+
+        @Override
+        public void onFailure(Throwable ex) {
+            handleFailure(data, record, ex);
+        }
+
+    });
+}
+----
+
+.Blocking (Sync)
+[source, java]
+----
+public void sendToKafka(final MyOutputData data) {
+    final ProducerRecord<String, String> record = createRecord(data);
+
+    try {
+        template.send(record).get(10, TimeUnit.seconds);
+        handleSuccess(data);
+    }
+    catch (ExecutionException e) {
+        handleFailure(data, record, e.getCause());
+    }
+    catch (TimeoutException | InterruptedException e) {
+        handleFailure(data, record, e);
+    }
+}
+----
 
 [[transactions]]
 ===== Transactions


garyrussell:
I believe we need to cherry-pick this back to all versions (to 1.1.x) - minus the `KafkaOperations` changes.

spring-kafka/src/main/java/org/springframework/kafka/support/CorrelatedProducerListener.java
```diff
@@ -0,0 +1,55 @@
...
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Marius Bogoevici
+ * @author Gary Russell
+ * @since 1.3
+ *
+ * @see org.apache.kafka.clients.producer.Callback
+ */
+public interface CorrelatedProducerListener<K, V> {
```
artembilan: Why do we need this at all since `send(ProducerRecord<K, V> record, ProducerListener<K, V> producerListener)` does exactly what we need for correlation request with the async response?

More over there is already mentioned `org.apache.kafka.clients.producer.Callback`...

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
```diff
@@ -134,6 +134,15 @@
 	ListenableFuture<SendResult<K, V>> send(ProducerRecord<K, V> record);
 
 	/**
+	 * Send the provided {@link ProducerRecord}.
+	 * @param record the record.
+	 * @param correlationData correlation data returned in a {@code FlexibleProducerListener}.
```
garyrussell: Typo; that was my original name.

spring-kafka/src/main/java/org/springframework/kafka/support/CorrelatedProducerListener.java
garyrussell: Because that's buried in the `KafkaTemplate` - we need a way for the template to call a user callback - e.g. with a `Message<?>`.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
garyrussell: Yes; I mean leave it on the template but don't change the interface.

garyrussell:
Now I see what you mean - the user provides the callback. OK; will rework.

spring-kafka/src/main/java/org/springframework/kafka/support/CorrelatedProducerListener.java
artembilan: > we need a way for the template to call a user callback

??? Isn't that `send(ProducerRecord<K, V> record, ProducerListener<K, V> producerListener)` per se?

src/reference/asciidoc/kafka.adoc
```diff
@@ -61,6 +61,8 @@ ListenableFuture<SendResult<K, V>> send(String topic, Integer partition, Long ti
 
 ListenableFuture<SendResult<K, V>> send(ProducerRecord<K, V> record);
 
+ListenableFuture<SendResult<K, V>> send(ProducerRecord<K, V> record, Object correlationData);
```
artembilan: This must be fixed for the `Callback`

src/reference/asciidoc/kafka.adoc
```diff
@@ -153,6 +155,18 @@ For convenience, the abstract `ProducerListenerAdapter` is provided in case you
 methods.
 It returns `false` for `isInterestedInSuccess`.
 
+Starting with version 1.3, the `CorrelatedProducerListener` is provided.
```
artembilan: Such a doc isn't relevant any more.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
```diff
@@ -134,6 +135,14 @@
 	ListenableFuture<SendResult<K, V>> send(ProducerRecord<K, V> record);
 
 	/**
+	 * Send the provided {@link ProducerRecord}.
+	 * @param record the record.
+	 * @param callback a {@link Callback} for the send operation.
+	 * @since 1.3
+	 */
+	void send(ProducerRecord<K, V> record, Callback callback);
```
garyrussell: Yes, it doesn't make sense to have a `ListenableFuture<?>` as well as a callback.

src/reference/asciidoc/kafka.adoc
```diff
@@ -61,6 +61,8 @@ ListenableFuture<SendResult<K, V>> send(String topic, Integer partition, Long ti
 
 ListenableFuture<SendResult<K, V>> send(ProducerRecord<K, V> record);
 
+ListenableFuture<SendResult<K, V>> send(ProducerRecord<K, V> record, Callback callback);
```
artembilan: I will change it to `void` during merge...

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -307,13 +314,25 @@ protected void closeProducer(Producer<K, V> producer, boolean inLocalTx) {
...
 		final Producer<K, V> producer = getTheProducer();
 		if (this.logger.isTraceEnabled()) {
 			this.logger.trace("Sending: " + producerRecord);
 		}
-		final SettableListenableFuture<SendResult<K, V>> future = new SettableListenableFuture<>();
+		final SettableListenableFuture<SendResult<K, V>> future = callback == null ? new SettableListenableFuture<>()
+				: null;
 		final boolean inLocalTx = inTransaction();
-		producer.send(producerRecord, new Callback() {
+		producer.send(producerRecord, callback != null ? callback : new Callback() {
```
artembilan: There is the block:
```
finally {
	if (KafkaTemplate.this.producers.get() == null) {
		closeProducer(producer, inLocalTx);
	}
}
```
In the end of `new Callback()` .
I think we just miss something there a bit and don't have choice unless wrap provided `callback` to that `new Callback()` any way. For consistency with other `send()` at least in case of TX...

WDYT?

spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
```diff
@@ -265,6 +267,29 @@ public boolean isInterestedInSuccess() {
 	}
 
 	@Test
+	public void withCallback() throws Exception {
```
artembilan: We might need to have similar test in the `KafkaTemplateTransactionTests` to be sure that TX behavior remains

spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
```diff
@@ -265,6 +267,29 @@ public boolean isInterestedInSuccess() {
...
+		template.flush();
+		assertThat(latch.await(10, TimeUnit.SECONDS)).isTrue();
+
+		//Drain the topic
+		KafkaTestUtils.getSingleRecord(consumer, INT_KEY_TOPIC);
+		pf.destroy();
+	}
+
+	@Test
 	public void testWithCallback() throws Exception {
```
artembilan: After closer review (sorry for late feedback), that looks like with this new `send()` with the `Callback` we just replicate the logic exists with the `ListenableFuture.addCallback()`.

Can not Success and Failure Message Channels logic be based on the `ListenableFuture.addCallback()`?
Looking into `producer.send()` usage in the `KafkaTemplate` we already fully meet the requirements.

WDYT?

garyrussell: Of course; good catch - should we revert the S-K change?

artembilan: Looks like we haven't merged anything yet to revert 😄 

But maybe we still can apply this PR with an appropriate comment to the documentation?

There is  a sample with the `future.addCallback()`, but I see no proper description...

garyrussell: Doh - sorry; thought you were commenting on SIK - I am juggling too many balls.

garyrussell: I am not sure why we would want to merge this, though; given that we can use the future.

artembilan: Well, my idea that we revert all the changes and present only the commit with the Docs polishing.

If we would have some comment in the doc before, you wouldn't came up with the current work, I think...

Otherwise, yes, we can just close this PR without any changes.

PR #402, opened 2017-08-25
Add API to access consumer metrics.

tsachev:
Here is my initial proposal to expose consumer metrics.
WDYT?

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
@@ -95,12 +95,12 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 	}
 
 	@Override
-	public Map<String, Map<MetricName, Metric>> metrics() {
-		Map<String, Map<MetricName, Metric>> metrics = new HashMap<>();
+	public Map<String, Map<MetricName, ? extends Metric>> metrics() {
+		Map<String, Map<MetricName, ? extends Metric>> metrics = new HashMap<>();
 		for (KafkaMessageListenerContainer<K, V> container : this.containers) {
 			metrics.putAll(container.metrics());
 		}
-		return metrics;
+		return Collections.unmodifiableMap(metrics);
 	}
 
 	/*
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
@@ -19,8 +19,13 @@ package org.springframework.kafka.listener;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
+
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
 
 import org.springframework.kafka.core.ConsumerFactory;
 import org.springframework.kafka.listener.config.ContainerProperties;
@@ -87,6 +92,15 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 	 */
 	public List<KafkaMessageListenerContainer<K, V>> getContainers() {
 		return Collections.unmodifiableList(this.containers);
+	}
+
+	@Override
+	public Map<String, Map<MetricName, Metric>> metrics() {
+		Map<String, Map<MetricName, Metric>> metrics = new HashMap<>();
+		for (KafkaMessageListenerContainer<K, V> container : this.containers) {
+			metrics.putAll(container.metrics());
+		}
+		return metrics;
 	}
 
 	/*
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -41,6 +41,8 @@ import org.apache.kafka.clients.consumer.NoOffsetForPartitionException;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.clients.consumer.OffsetCommitCallback;
 import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.WakeupException;
 
@@ -151,6 +153,22 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 		else {
 			return null;
 		}
+	}
+
+	@Override
+	public Map<String, Map<MetricName, Metric>> metrics() {
+		ListenerConsumer listenerConsumer = this.listenerConsumer;
+		Map<String, Map<MetricName, Metric>> metrics = new HashMap<>();
+		if (listenerConsumer != null) {
+			for (Map.Entry<MetricName, ? extends Metric> entry : listenerConsumer.consumer.metrics().entrySet()) {
+				MetricName metricName = entry.getKey();
+				Metric metric = entry.getValue();
+				String client = metricName.tags().get("client-id");
+				Map<MetricName, Metric> clientMetrics = metrics.computeIfAbsent(client, k -> new HashMap<>());
+				clientMetrics.put(metricName, metric);
+			}
+		}
+		return metrics;
 	}
 
 	@Override
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -156,23 +156,17 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 	}
 
 	@Override
-	public Map<String, Map<MetricName, Metric>> metrics() {
+	public Map<String, Map<MetricName, ? extends Metric>> metrics() {
 		ListenerConsumer listenerConsumer = this.listenerConsumer;
-		Map<String, Map<MetricName, Metric>> metrics = new HashMap<>();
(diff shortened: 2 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -89,6 +95,13 @@ public void setConcurrency(int concurrency) {
 		return Collections.unmodifiableList(this.containers);
 	}
 
+	@Override
+	public Map<MetricName, ? extends Metric> metrics() {
+		return this.containers.stream()
+				.flatMap(c -> c.metrics().values().stream())
+				.collect(Collectors.toMap(Metric::metricName, Function.identity()));
```
artembilan: @tsachev ,

thank you for contribution. The looks clear and straightforward!
Although I wonder if there is going to be some collision with metrics overrides if we have concurrency > 1. I mean different target `Consumer`s may have the same `MetricName`, so the value from the first will be overridden by the second.
That might not be a case though if `   * @param group       logical group name of the metrics to which this metric belongs` is different for different consumers, but I somehow would like to see some samples how it is in the reality.

Thanks

garyrussell: Perhaps we should change the signature to `public Map<Integer, Map<MetricName, ? extends Metric>> metrics()`, with the KMLC always returning a singleton `Map` and the concurrent container returning a map keyed by consumer number.

We need resolution on this by today or tomorrow since our release candidate is due Friday.

garyrussell: Or, perhaps, `public Map<String, Map<MetricName, ? extends Metric>> metrics()` where the key is the listener thread name - it might help with correlation.

artembilan: Well, in terms of Apache Kafka we might need just use `clientId` on the matter.
But on the other hand I would like to debug this and see if all these `MetricName` s are not crossed and let us really to distinguish the `consumer` they belong to.

artembilan: So, debug shows this:
```
key = {MetricName@4991} "MetricName [name=outgoing-byte-rate, group=consumer-node-metrics, description=, tags={client-id=consumer-2, node-id=node-2147483647}]"
 name = "outgoing-byte-rate"
 group = "consumer-node-metrics"
 description = ""
 tags = {LinkedHashMap@5307}  size = 2
  0 = {LinkedHashMap$Entry@5324} "client-id" -> "consumer-2"
  1 = {LinkedHashMap$Entry@5325} "node-id" -> "node-2147483647"
 hash = -260015487
```
Pay attention to that ` "client-id" -> "consumer-2"` tag. Therefore I was right. Even in this flatten mode from the `ConcurrentMessageListenerContainer` we get unique metrics.

However I agree that it would be much convenient to group them per consumer. From here the structure like `Map<String, Map<MetricName, ? extends Metric>> metrics()` is fine with me. However I still vote for the `client-id` as a main key.

tsachev: My initial idea was to group the metrics too, but I could not decide for myself if it should be a `client-id` or something else.
Note that there is no Kafka API we can use to get the _actual_ consumer client id, because they may genereate it in the constructor if not supplied. I  need to rely on the `client-id` metric tag key, which I doubt they would change, anyway.

I will update the PR in few minutes.

tsachev:
I have updated my PR based on the comments from the previous version.

If you decide to merge this are there any chances for backporting to `1.2` or `1.3`. I know its an api change, but 1.3 is not RC, yet?

garyrussell:
We can backport it to 1.3 without any problems; 1.3 will released on the same schedule as 2.0.

We can also backport it to 1.2, if necessary, but only on the concrete container classes (leaving the interface unchanged).

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -154,6 +156,22 @@ else if (this.listenerConsumer.assignedPartitions != null) {
...
 	@Override
+	public Map<String, Map<MetricName, Metric>> metrics() {
+		ListenerConsumer listenerConsumer = this.listenerConsumer;
+		Map<String, Map<MetricName, Metric>> metrics = new HashMap<>();
+		if (listenerConsumer != null) {
+			for (Map.Entry<MetricName, ? extends Metric> entry : listenerConsumer.consumer.metrics().entrySet()) {
+				MetricName metricName = entry.getKey();
+				Metric metric = entry.getValue();
+				String client = metricName.tags().get("client-id");
+				Map<MetricName, Metric> clientMetrics = metrics.computeIfAbsent(client, k -> new HashMap<>());
```
artembilan: If you want to have it backported, the code has to be Java 7 compatible.

And please, consider do not squash commits in the future. That way it easier to track the changes in the PR and we'd have a notification about a new commit. Right now I've just decided to take a look into the code one more time and realized there is a new fix! 
We squash during merge into the `master` anyway. Thanks.

On the other hand, I think the way you extract the `clientId` is fine. Here is the code from the `KafkaConsumer`:
```
String clientId = config.getString(ConsumerConfig.CLIENT_ID_CONFIG);
if (clientId.length() <= 0)
                clientId = "consumer-" + CONSUMER_CLIENT_ID_SEQUENCE.getAndIncrement();
this.clientId = clientId;
Map<String, String> metricsTags = Collections.singletonMap("client-id", clientId);
```
That's sad that there is no `getClientId()`, but we really can use that tag from metrics.

spring-kafka/src/main/java/org/springframework/kafka/listener/MessageListenerContainer.java
```diff
@@ -34,4 +39,11 @@
 	 */
 	void setupMessageListener(Object messageListener);
 
+	/**
+	 * See {@link org.apache.kafka.clients.consumer.Consumer#metrics()}.
```
artembilan: We typically use `@see` JavaDoc tag in the end. But here should be a general description what this method does. Even if it is just `Return the consumer(s) metrics grouped by {@code client-id}`. Right, the `@return` tag still must be there even if with the same sentence.

tsachev: Will fix that.
I was just following what's there for the producer metrics. https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java#L155-L160

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -163,8 +163,12 @@ else if (this.listenerConsumer.assignedPartitions != null) {
 			for (Map.Entry<MetricName, ? extends Metric> entry : listenerConsumer.consumer.metrics().entrySet()) {
 				MetricName metricName = entry.getKey();
 				Metric metric = entry.getValue();
-				String client = metricName.tags().get("client-id");
-				Map<MetricName, Metric> clientMetrics = metrics.computeIfAbsent(client, k -> new HashMap<>());
+				String clientId = metricName.tags().get("client-id");
+				Map<MetricName, Metric> clientMetrics = metrics.get(clientId);
+				if (clientMetrics == null) {
```
artembilan: Well, @garyrussell 's idea was to return here `Collections.singletonMap()`. So, the logic would be a bit different in this method.
Also I would like to see `Collections.unmodifableMap()` in the `ConcurrentMessageListenerContainer`. 
We really should not allow end-user to modify the result of the getter...

spring-kafka/src/main/java/org/springframework/kafka/listener/MessageListenerContainer.java
```diff
@@ -39,11 +39,13 @@
...
-	 * @return the consumer(s) metrics grouped by {@code client-id}
-	 * @since 2.0
-	 */
-	Map<String, Map<MetricName, Metric>> metrics();
+    /**
+     * Returns metrics for this container's consumer(s), grouped by {@code client-id}.
+     *
+     * @return the consumer(s) metrics grouped by {@code client-id}
+     * @see org.apache.kafka.clients.consumer.Consumer#metrics()
+     * @since 2.0
```
artembilan: Just nit-picking, but proper order that `@see` is the latest tag in the JavaDoc.
Also no blank lines in the method JavaDocs, please.
Plus the general sentence on the method JavaDoc should be in the imperative, 

The mentioned JavaDocs in the `KafkaOperations` must be fixed as well. Thanks. But that is already out of the scope your task.

artembilan:
The Travis report says:
```
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/MessageListenerContainer.java:49: Line has leading space characters; indentation should be performed with tabs only. [RegexpSinglelineJava]
```
The indents must always be only tabs.

spring-kafka/src/main/java/org/springframework/kafka/listener/MessageListenerContainer.java
tsachev: hmm, usually `@since`, `@serial` and `@deprecated` are after `@see`, but If you want it to be the last, I don't mind.

artembilan: Yeah...
This is our Checkstyle rule:
```
<module name="AtclauseOrder">
			<property name="target" value="METHOD_DEF, CTOR_DEF, VARIABLE_DEF"/>
		<property name="tagOrder" value="@param, @return, @throws, @since, @deprecated, @see"/>
</module>
```
This is our code style: https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style#javadoc-formatting

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/1945f29d5518e3c4a9950ba82135420dfb61e808 and cherry-picked to `1.3.x`.

I have polished JavaDocs a bit, add your name to the `@author` list and mentioned feature in the Docs.

Thank you for the contribution.
Looking forward for more!

Issue #405, opened 2017-08-30
Ack After Each Record with RECORD AckMode


garyrussell:
https://stackoverflow.com/questions/45949413/getting-high-number-of-duplicate-records-with-apache-kafka-consumer#comment78888930_45949413

PR #409, opened 2017-09-05
GH-405: Fix AckMode.RECORD

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/405

With `AckMode.RECORD`, now that the listener is running on the consumer thread
we should commit the offset immediately when the listener returns.

__requires backport to 1.3.x after review/merge__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -547,7 +547,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 			long lastAlertAt = lastReceive;
 			while (isRunning()) {
 				try {
-					if (!this.autoCommit) {
+					if (!this.autoCommit && !this.isRecordAck) {
 						processCommits();
 					}
 					processSeeks();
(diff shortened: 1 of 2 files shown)


artembilan:
Merging after fixing Checkstyle violations...

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -880,7 +880,19 @@ private RuntimeException doInvokeRecordListener(final ConsumerRecord<K, V> recor
 						this.listener.onMessage(record);
 						break;
 				}
-				if (!this.isAnyManualAck && !this.autoCommit) {
+				if (this.isRecordAck) {
```
artembilan: Hm. Shouldn't we apply the same immediate logic for the `this.isBatchAck` mode in the `doInvokeBatchListener()`?

Shouldn't we skip then both of them in the `processCommits()` afterwards?

garyrussell: Not necessary since we call `processCommits` anyway before the next poll, which will commit the batch acks.

I suppose we could now suppress the call to processCommits() if `isRecordAck` is true; will push.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/dc1d9d1cb22090624cf98dff4fcb39250787d3f8.

Are you going to PR the backport for `1.3.x` or do you let me to try to cherry-pick it?

garyrussell:
I will backport - it won't be clean.

garyrussell:
It was cleaner than I expected; cherry-picked and pushed to 1.3.x as 5818702bf47b21597dda8468734ee02a9f334896

PR #410, opened 2017-09-07
Allow An Explicit Port for EmbeddedKafka

garyrussell:
__cherry-pick to 1.3.x__


artembilan:
I don't have arguments against the fix, but just curious what drives you to do this?
Thanks

garyrussell:
>but just curious what drives you to do this?

@olivergierke suggestion so a broker can be exposed for communication between multiple JVMs.

artembilan:
... and cherry-picked as https://github.com/spring-projects/spring-kafka/commit/ab31a115586d02f58a937359897152bb51f94803

Issue #411, opened 2017-09-07
Change KafkaEmbedded to Use AdminClient for Topic Creation


garyrussell:
The 0.11 client supports topic creation directly instead of via Zookeeper.

PR #412, opened 2017-09-07
GH-411: Use AdminClient to Provision Test Topics

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/411

The 0.11 client has a new `AdminClient`; use it instead of interacting with
Zookeeper in the `EmbeddedKafka` when provisioning test topics.


artembilan:
The fix looks good, but what the benefit we get doing this way?
I don't see that we some how minimize our dependency of the Zookeeper there.
Thanks

garyrussell:
It's just that I prefer to remove as much Scala as possible 😄

Issue #418, opened 2017-09-15
KafkaAdmin - Don't Open Connection if No NewTopic Beans in Application Context


garyrussell:
Subject says it all.

However, if `fatalIfBrokerNotAvailable` is true; always attempt connection.

garyrussell:
Actually, the `fatalIfBrokerNotAvailable` doesn't work - we always get a client; the exception is thrown only when an operation is performed.

PR #420, opened 2017-09-15
GH-418: KafkaAdmin Improvements

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/418

Depending on the nature of the problem, `fatalIfBrokerNotAvailable` did not always work.

With a bad broker address, the client is not created.
With a good broker address, but the broker is down, the add fails.

Add more robust logic to the initialization code.

Also add the `autoCreate` flag, allowing users to suppress auto creation during
context initialization.

__cherry-pick to 1.3.x__

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
@@ -61,6 +61,10 @@ public class KafkaAdmin implements ApplicationContextAware, SmartInitializingSin
 
 	private boolean fatalIfBrokerNotAvailable;
 
+	private boolean autoCreate = true;
+
+	private boolean initializingContext;
+
 	/**
 	 * Create an instance with an {@link AdminClient} based on the supplied
 	 * configuration.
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -93,6 +97,15 @@ public void setFatalIfBrokerNotAvailable(boolean fatalIfBrokerNotAvailable) {
 	}
 
 	/**
+	 * Set to false to suppress auto creation of topics during context initialization.
+	 * @param autoCreate
```
artembilan: Missed param description

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -102,43 +115,59 @@ public void setFatalIfBrokerNotAvailable(boolean fatalIfBrokerNotAvailable) {
...
+			AdminClient adminClient = null;
 			try {
-				if (this.applicationContext != null) {
-					addTopicsIfNeeded(adminClient,
-							this.applicationContext.getBeansOfType(NewTopic.class, false, false).values());
+				adminClient = AdminClient.create(this.config);
+			}
+			catch (Exception e) {
+				if ((this.initializingContext && this.fatalIfBrokerNotAvailable)
+						|| !this.initializingContext) {
```
artembilan: IDEA suggests me to simplify:
```
if (!this.initializingContext || this.fatalIfBrokerNotAvailable) {
```

artembilan:
Merged and cherry-picked as https://github.com/spring-projects/spring-kafka/commit/b35843ae8e6df9918901c0b59c3a3f4feb0ba3f6

Issue #444, opened 2017-10-04
Support disabling `@EmbeddedKafka` in sub-classes


emetsger:
When `@EmbeddedKafka` is defined on a class, sub-classes inherit the annotation, as expected.  Sometimes this is not desired behavior, when there may be one class in the hierarchy which doesn't need a broker instantiated, or where instantiating a broker is in conflict in some form or fashion with the test.

This issue proposes to add a `disabled` attribute to `@EmbeddedKafka` which defaults to `false`.  When set to `true`, the logic in the `EmbeddedKafkaContextCustomizer` is bypassed, and no `KafkaEmbedded` instance is created.

This would allow a class in a hierarchy to disable the instantiation of `KafkaEmbedded` on a case-by-case basis, while still taking advantage of inheriting other Spring-related annotations.

emetsger:
Closed as won't fix.

PR #445, opened 2017-10-04
GH-444: Adds a `disabled` attribute to EmbeddedKafka

emetsger:
Allows classes to disable the instantiation of KafkaEmbedded on a case-by-case basis.

Resolves https://github.com/spring-projects/spring-kafka/issues/444

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
@@ -96,5 +96,14 @@ public @interface EmbeddedKafka {
 	 */
 	String[] brokerProperties() default { };
 
+	/**
+	 * When {@code true} no {@code KafkaEmbedded} instance is created, bypassing the {@link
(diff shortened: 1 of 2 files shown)


artembilan:
I think would be better to just remove `@Inherited` from the `@EmbeddedKafka` and let end-user to decide when and how to specify it in his/her hierarchy.

But good catch anyway!
Sad that we didn't do that before GA 😄

emetsger:
Yeah, it seems a bit awkward either way.  It's nice to be able to inherit it, but awkward to disable it.  If it isn't inherited, then you wind up potentially copying and pasting a bunch of boilerplate!  I'm open to whatever makes the most sense for the project!

artembilan:
Sorry, I would deny this `disable` approach and would leave it as `@Inherited`.
It's really a bit awkward to *enable* feature to *disable* it :shrug:.

That's definitely problem of the target hierarchy to inherit Kafka aware suit, but don't use it.
Smells like a flaw of the target design...

emetsger:
Understandable!

Issue #485, opened 2017-11-17
No method found for class when using @Nonnull annotation


Undestor:
When using @Nonnull (javax.annotation.Nonnull) on the parameter of the KafkaHandler, kafka can`t find the method: "org.springframework.kafka.KafkaException: No method found for class".
Same happens using the annotation @CheckForNull.

Without the annotation everything works fine.

```java
@KafkaListener(topics = "someTopic")
@Component
public class KafkaListener{ 

    @KafkaHandler
    public void receive(@Nonnull final SomeObject object) {
        //DoSomething
    }
}
```

garyrussell:
Interesting - it also fails the same way with `import org.springframework.lang.NonNull`.

It works if I move the listener to the method...

```java
	public class Listener {

		@KafkaListener(topics = "gh485", groupId = "gh485")
		public void listen (@NonNull final String in) {
			System.out.println(in);
		}

	}
```

So that might be a workaround - unless you need the multimethod listener feature.

garyrussell:
It's because we're explicitly looking for no annotation or `@Payload` to find the candidate handler method.

PR #486, opened 2017-11-17
GH-485: Fix multi listener method @KafkaListener

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/485

When looking for matching methods, we matched on no annotation or `@Payload`.

Match should have been no annotation(s) or not `@Header`.

__cherry-pick to 2.0.x, 1.3.x__


spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
```diff
@@ -82,6 +82,7 @@
 import org.springframework.kafka.support.converter.StringJsonMessageConverter;
 import org.springframework.kafka.test.rule.KafkaEmbedded;
 import org.springframework.kafka.test.utils.KafkaTestUtils;
+import org.springframework.lang.NonNull;
```
artembilan: This cannot be backported to `1.3.x` because it was introduced since Spring `5.0`.

However I'll replace it with some other Spring `4.3` compatible parameter level annotation during backport

artembilan:
... backported to `2.0.x` and as https://github.com/spring-projects/spring-kafka/commit/bf52586436a8db62c920d9acd3cdfb30e9f68bd3 into `1.3.x` after fixing versions compatibility.

Issue #492, opened 2017-11-24
StringJsonMessageConverter: Add type and/or type mapping information to Headers


garyrussell:
Now that Kafka supports headers, we can convey type information to allow deserializing multiple types in a single converter, similar to Spring AMQP's `Jackson2JsonMessageConverter`.

[SO Answer for reference](https://stackoverflow.com/questions/47456009/spring-kafka-multiple-serializers-and-consumer-container-factories/47456359#47456359).

garyrussell:
Also, enhance `Json(De)Serializer` to implement `Extended(De)Serializer`.

PR #494, opened 2017-11-27
GH-492: JSON Improvements

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/492

For `StringJsonMessageConverter` and `JsonSerializer`

- Convey type information in Headers using a Jackson type mapper
- Setting `addTypeInfo` to `false` disables this feature

For `StringJsonMessageConverter` (inbound)

- If configured with with the default type mapper, no changes - use the inferred type from the method signature
- If configured with a type mapper where the `TypePrecedence` is not `INFERRED`, use the headers if available.

For `JsonDeserializer`

- use type information from headers, falling back to a default if present

For `JsonSerializer` and `JsonDeserializer`

- provide configuration (e.g. default type) via kafka properties

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/converter/AbstractJavaTypeMapper.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/converter/AbstractJavaTypeMapper.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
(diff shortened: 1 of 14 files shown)


garyrussell:
Makes sense (future).

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/7cd53119d3841a0e964c6f42186a3affcf9a4ea2.

PR #496, opened 2017-11-28
Set default value for noPollThreshold

izeye:
Its Javadoc and the reference document say the default value for `noPollThreshold` is `3` but not set.

This PR sets default value for `noPollThreshold` and also adds constants for default values and uses them for Javadoc.


artembilan:
Merged and cherry-picked as https://github.com/spring-projects/spring-kafka/commit/11fd8637a2022c98f6649604f7b646957a262954

Thank you for contribution.
Looking forward for more!

Issue #451, opened 2017-10-08
Provide option for stopping consumption on error


mlueders:
First off, kudos for a great library!  At my company, we built a `common-kafka` library a while back and I'm going through it now and gutting it, replacing most of the functionality with spring-kafka (which also does a ton more), so thanks!

There's one particular feature of the old library for which I can't seem to find a non-hack solution.  spring-kafka does have support for retries via RetryTemplate but if all retries are exhausted, it sends the error to a handler and then continues going.  For our use case, this is bad.  We use kafka for state propagation, and a missed message equates to corrupted data.  Consider a temporary network failure which prevents communication with the database.  If the outage lasts less than some threshold, the system should recover.  But if it lasts longer, we don't want to just keep processing messages.  We want the container to halt and log an error, at which point monitoring kicks in and a developer gets paged.  If it's a dependency outage (e.g. database), we fix the dependency and restart the app which starts consuming at the last committed offset (which for us, is always immediatelyl after the last successfully processed message).  If it's a poison message, we fix the application and redeploy, but in no case do we drop a message.  I realize a DLQ can be used for this but that introduces it's own complexities so we've opted for the simpler approach of halt, fix and proceed.

I'm currently using `1.2.2.RELEASE` and the hack solution I have is to use a combination of throwing a Throwable from the ErrorHandler and https://projectlombok.org/features/SneakyThrows.  This effectively stops the ListenerInvoker by exiting out of invokeRecordListener/invokeBatchListener and eventually setting active to false in ListenerConsumer.  This is obviously not desirable.

Ideally, there would be a configurable property on ContainerProperties to control whether the container continued after an error or not.  Is there a reason this is not an option or is this open for contribution?

artembilan:
I wonder why don't you consider to use `container.stop()` from error handler or recovery callback.

It's undesirable to introduce such an option to stop after error. That's just doesn't fit to general purpose of messaging: one message failure doesn't effect processing for others.

mlueders:
I disagree that there is a "general purpose of messaging”.  There are use cases and there are solutions which fit those use cases.  In this instance, I am using kafka (messaging) to propagate state from one service to another, in some cases taking action based on that state.  If I lose messages, that state becomes corrupted.  In other words, in this instance, I don’t care about the processing of other messages - I care about the consistency of data.

[...]

PR #499, opened 2017-11-29
GH-451: Add Container Stopping Error Handlers

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/451

Add error handlers that stop the container.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerAwareBatchErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerAwareBatchErrorHandler.java
@@ -27,6 +27,7 @@ import org.apache.kafka.clients.consumer.ConsumerRecords;
  * @since 2.1
  *
  */
+@FunctionalInterface
 public interface ContainerAwareBatchErrorHandler extends ConsumerAwareBatchErrorHandler {
 
 	@Override
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerAwareBatchErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerAwareBatchErrorHandler.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.listener;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+
+/**
+ * An error handler that has access to the batch of records from the last poll the
+ * consumer, and the container.
+ *
+ * @author Gary Russell
+ * @since 2.1
+ *
+ */
+public interface ContainerAwareBatchErrorHandler extends ConsumerAwareBatchErrorHandler {
+
+	@Override
+	default void handle(Exception thrownException, ConsumerRecords<?, ?> data, Consumer<?, ?> consumer) {
+		throw new UnsupportedOperationException("Container should never call this");
+	}
+
+	/**
+	 * Handle the exception.
+	 * @param thrownException the exception.
+	 * @param data the consumer records.
+	 * @param consumer the consumer.
+	 * @param container the container.
+	 */
+	void handle(Exception thrownException, ConsumerRecords<?, ?> data, Consumer<?, ?> consumer,
+			MessageListenerContainer container);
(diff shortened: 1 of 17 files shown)


garyrussell:
Needs a "what's new" entry after the JSON PR is merged.

artembilan:
OK. This is much complicated than JSON one 😄 .

Too much classes anti-pattern, but I guess you have had a reason.
Or we should've never done "per-record" logic and always treat the result of the polling (`ConsumerRecords`) as a single entity...

Polling locally for review...

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerStoppingBatchErrorHandler.java
```diff
@@ -0,0 +1,67 @@
...
+		int n = 0;
+		while (container.isRunning() && n++ < 100) {
+			try {
+				Thread.sleep(100);
+			}
+			catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+				return;
+			}
+			throw new KafkaException("Stopped container");
```
artembilan: Something wrong is here.
You step into the loop (in most cases I believe), have some delay and ... throw exception.
You never go to the next cycle, therefore you don't wait for the container to be properly stopped.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerStoppingErrorHandler.java
```diff
@@ -0,0 +1,68 @@
...
+		while (container.isRunning() && n++ < 100) {
+			try {
+				Thread.sleep(100);
+			}
+			catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+				return;
+			}
+		}
+		throw new KafkaException("Stopped container");
```
artembilan: OK. And here the exception is thrown outside of the loop.
This is better than previous one, but I still can't compile why should we throw exception here anyway.
And if though why doesn't it include `thrownException` then?

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerAwareErrorHandler.java
```diff
@@ -0,0 +1,51 @@
...
+ * An error handler that has access to the unprocessed records from the last poll
+ * (including the failed record), the consumer, and the container.
+ * The records passed to the handler will not be passed to the listener
+ * (unless re-fetched if the handler performs seeks).
+ *
+ * @author Gary Russell
+ * @since 2.1
+ *
+ */
+public interface ContainerAwareErrorHandler extends RemainingRecordsErrorHandler {
```
artembilan: The `@FunctionalInterface` is missing.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerAwareBatchErrorHandler.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * An error handler that has access to the batch of records from the last poll the
+ * consumer, and the container.
+ *
+ * @author Gary Russell
+ * @since 2.1
+ *
+ */
+public interface ContainerAwareBatchErrorHandler extends ConsumerAwareBatchErrorHandler {
```
artembilan: The @FunctionalInterface is missing.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerStoppingErrorHandler.java
garyrussell: We have to throw an exception - when an error handler "handles" an error, further processing is performed (e.g. transaction commit) but, yes, it should include the cause.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -8,3 +8,9 @@ This version requires the 1.0.0 `kafka-clients` or higher.
 
 The `StringJsonMessageConverter` and `JsonSerializer` now add type information in `Headers`, allowing the converter and `JsonDeserializer` to create specific types on reception, based on the message itself rather than a fixed configured type.
 See <<serdes>> for more information.
+
+
+==== Container Stopping Error Handlers
+
+Container Error handlers are now provided for both record and batch listeners that treat any exceptions thrown by the listener as fatal; they stop the container.
```
artembilan: > ... and rethrow a `KafkaException` to prevent further processing.

Or something similar.

garyrussell: That's not really appropriate (too much detail) for a "what's new" entry. However, it raised an issue in my mind about the seek to current error handlers.

Polishing now...

artembilan: Oh! Really?
So, tell that in the main Doc then, please.

I think if we had it before I might not ask the question why we re-throw...

Thanks

garyrussell:
See commit comment in last commit - parts of it must be back ported.

The interface hierarchy is required for backwards compatibility; we can consider squashing it in 3.0.

artembilan:
> Parts of this commit will need to be back ported.

Good. Would you mind preparing an appropriate PR after merging this one?
Thanks

garyrussell:
>Would you mind preparing an appropriate PR after merging this one?

Yes, of course; that's what I meant.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/d2ce4aba1e68de9e97a63cb0052f7c83e90e40c2 after rebasing

Issue #501, opened 2017-11-30
Fix transactions with AckMode.RECORD


garyrussell:
With `AckMode.RECORD`, the container commits via the consumer instead of the transaction.

Already fixed on master by https://github.com/spring-projects/spring-kafka/pull/499

PR #502, opened 2017-11-30
GH-501: Fix transactions with AckMode.RECORD

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/501

With `AckMode.RECORD`, the container was committing the offsets on the
consumer instead of sending them to the transaction.

Also remove the restriction and allow the use of `RemainingRecordsAwareErrorHandler`
with transactions.
Throw an exception from the `SeekToCurrentErrorHandler` so the transaction rolls back.

__Separate PR upcoming for 1.3.x__

### DIFF
--- a/src/reference/asciidoc/kafka.adoc
+++ b/src/reference/asciidoc/kafka.adoc
@@ -1525,6 +1525,8 @@ The next `poll()` will return the 3 unprocessed records.
 
 If the `AckMode` was `BATCH`, the container commits the offsets for the first 2 partitions before calling the error handler.
 
+After seeking, an exception wrapping the ListenerExecutionFailedException is thrown. This is to cause the transaction to roll back (if transactions are enabled).
+
 [[kerberos]]
 ==== Kerberos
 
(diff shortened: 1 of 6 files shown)


src/reference/asciidoc/kafka.adoc
```diff
@@ -1525,6 +1525,8 @@ The next `poll()` will return the 3 unprocessed records.
 
 If the `AckMode` was `BATCH`, the container commits the offsets for the first 2 partitions before calling the error handler.
 
+After seeking, an exception wrapping the ListenerExecutionFailedException is thrown. This is to cause the transaction to roll back (if transactions are enabled).
```
artembilan: I prefer to wrap classes to the code and each sentence from new line.

Fixing on merge...

artembilan:
M-m-m. Not yet. 
Travis has failed with the:
```
org.springframework.kafka.annotation.EnableKafkaIntegrationTests > testBatch FAILED

    org.mockito.exceptions.misusing.UnfinishedStubbingException at EnableKafkaIntegrationTests.java:346
```

Does it say anything to you?

artembilan:
In other places we use `any(Map.class)` or just `any()`.
Look like it doesn't like `HashMap` for the case when we deal with the `Collections.singletonMap()`

garyrussell:
Seems to be unrelated; I'll create a separate PR to fix.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/1ad74fd2a16fc9e0916728133321bdee1d8703eb

PR #507, opened 2017-12-01
Introduced support for Spring Data projection based message binding.

olivergierke:
This commit introduces a ProjectingMessageConverter that supports binding `String`- and `byte[]`-backed JSON strings to Spring Data Projection interfaces. This allows very selective, and low-coupled bindings to data including the lookup of values from multiple places inside the JSON document. E.g. the following interface can be defined as message payload type:

```java
interface SomeSample {

  @JsonPath({ "$.username", "$.user.name" })
  String getUsername();
}
```

Accessor methods will be used to lookup the property name as field in the received JSON document by default. The `@JsonPath` expression allows to customize the value lookup and even to define multiple JSONPath expression to lookup values from multiple places until an expression returns an actual value.

If the type we're supposed to unmarshal into is a plain class the `ProjectingMessageConverter` delegates to the default `StringJsonMessageConverter`.

Added Spring Data and the transitively required Jayway JSONPath library as optional build dependencies.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/converter/ProjectingMessageConverter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/converter/ProjectingMessageConverter.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.support.converter;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+
+import org.springframework.core.ResolvableType;
+import org.springframework.data.projection.ProjectionFactory;
+import org.springframework.data.projection.SpelAwareProxyProjectionFactory;
+import org.springframework.data.web.JsonProjectingMethodInterceptorFactory;
+import org.springframework.kafka.support.KafkaNull;
+import org.springframework.util.Assert;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.jayway.jsonpath.spi.mapper.JacksonMappingProvider;
+
+/**
+ * A {@link MessageConverter} implementation that uses a Spring Data {@link ProjectionFactory} to bind incoming messages
+ * to projection interfaces.
+ *
+ * @author Oliver Gierke
+ * @since 2.2
+ */
+public class ProjectingMessageConverter extends MessagingMessageConverter {
+
+	private final ProjectionFactory projectionFactory;
+	private final MessagingMessageConverter delegate;
+
+	/**
+	 * Creates a new {@link ProjectingMessageConverter} using the given {@link ObjectMapper}.
+	 *
+	 * @param mapper must not be {@literal null}.
+	 */
+	public ProjectingMessageConverter(ObjectMapper mapper) {
+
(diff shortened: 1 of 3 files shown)


olivergierke:
Wasn't sure so I rather thought I'm conservative.

I am about to polish the PR to make sure the reverse pipes the potentially proxied instance through the Jackson serializer (which it currently doesn't). Stay tuned!

olivergierke:
PR updated.

artembilan:
@olivergierke ,

not related to this PR, but I think something on the matter any way.

Have you thought already about some general solution for the `org.springframework.messaging.converter.MessageConverter` implementation?

That might not be present in Spring Messaging since it can't depend on Spring Data, but we can settle it in the Spring Integration.

WDYT?

olivergierke:
I can certainly look into that. I think we could even have that implementation in Spring Data itself as Spring Messaging is a core Spring Framework module these days.

artembilan:
Ah, right! That's definitely makes sense 😄 

Also I guess we need similar solution for Spring AMQP - it has its own `MessagingMessageConverter` as well.

I don't tell that you have to do everything on the matter, just try to share as much potential I see here as possible.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/ProjectingMessageConverter.java
```diff
@@ -0,0 +1,121 @@
...
+
+		if (source instanceof String) {
+			return String.class.cast(source).getBytes(StandardCharsets.UTF_8);
+		}
+
+		if (source instanceof byte[]) {
+			return byte[].class.cast(source);
+		}
+
+		throw new ConversionException(String.format("Unsupported payload type %s!", source.getClass()), null);
```
artembilan: Shouldn't we say in this message what types do we support? Just to be as friendly as possible for end-users of our lib.

artembilan:
Just gently ping, @olivergierke , to be sure that we are on the same page.

olivergierke:
Sorry for the late reply. I'mm gonna polish up the exception message to include the supported payload types. Not planning to do anything beyond that for now. I've brought up the broader topic with Jürgen at SpringOne and we're gonna look into options to bring the projection facilities closer to the framework. But first things first I guess. 🙃

artembilan:
@olivergierke ,

I'm merging this PR with polishing we talked about before.
If you have any objections, you have at about 10 mins to raise your hand.
After that we are going to release this project.

olivergierke:
Please go ahead, Arten. Sorry, I didn’t get to it.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/ProjectingMessageConverter.java
```diff
@@ -0,0 +1,121 @@
...
+			return KafkaNull.INSTANCE;
+		}
+
+		Class<?> rawType = ResolvableType.forType(type).resolve(Object.class);
+
+		if (!rawType.isInterface()) {
+			return this.delegate.extractAndConvertValue(record, type);
+		}
+
+		InputStream inputStream = new ByteArrayInputStream(getAsByteArray(value));
```
artembilan: While you here, @olivergierke , I would like to ask on this line:

Is it OK to keep `InputStream` opened or would be better to do this:
```
try (InputStream inputStream = new ByteArrayInputStream(getAsByteArray(value))) {
	return this.projectionFactory.createProjection(rawType, inputStream);
}
catch (IOException e) {
	throw new ConversionException("Cannot project from the value", e);
}
```
?

Test are passed and I see that final object for method invocation is the `Map` based on the JSON parsing.
So, sounds like Jackson has finished its parsing process already and we have to close `InputStream`

Your thoughts?

Thanks.

artembilan: Well, actually we don't need to do that. The `JacksonJsonNodeJsonProvider` performs this:
```
public JsonNode readTree(Reader r) throws IOException {
       return _readTreeAndClose(_jsonFactory.createParser(r));
}
```

So, nevermind 😄 
Merging...

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/eb67289c7b840be22300a60fe0f4eb3e00ffb81d

spring-kafka/src/main/java/org/springframework/kafka/support/converter/ProjectingMessageConverter.java
olivergierke: A `ByteArrayInputStream` doesn't need to be closed explicitly as its `close()` method is a no-op anyway.

PR #509, opened 2017-12-05
Improve JsonDeserializer to keep compatibility

quaff:
Since version 2.1.0, spring-kafka add whitelist of trusted packages for security, It will break existing application, It must manual add generic type's package like this.addTrustedPackages(Trade.class.getPackage().getName()), We should add Trade's package as default for class TradeDeserializer extends JsonDeserializer<Trade>.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
@@ -93,6 +93,9 @@ public class JsonDeserializer<T> implements ExtendedDeserializer<T> {
 		if (targetType == null) {
 			targetType = (Class<T>) ResolvableType.forClass(getClass()).getSuperType().resolveGeneric(0);
 		}
+		if (targetType != null) {
+			this.addTrustedPackages(targetType.getPackage().getName());
+		}
 		this.targetType = targetType;
 	}
 
--- a/spring-kafka/src/test/java/org/springframework/kafka/support/serializer/JsonSerializationTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/support/serializer/JsonSerializationTests.java
@@ -24,11 +24,13 @@ import java.util.HashMap;
 import java.util.List;
 
 import org.apache.kafka.common.errors.SerializationException;
+import org.apache.kafka.common.header.Headers;
+import org.apache.kafka.common.header.internals.RecordHeaders;
 import org.apache.kafka.common.serialization.StringDeserializer;
 import org.apache.kafka.common.serialization.StringSerializer;
 import org.junit.Before;
 import org.junit.Test;
-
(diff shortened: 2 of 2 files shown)


garyrussell:
Nice! Thanks.

garyrussell:
A test case would be useful; thanks.

quaff:
Updated with test case.

garyrussell:
Thanks!

FYI, in general, we don't squash commits on PRs so we can see the deltas; we always do a final squash when merging.

Also, since the `targetType` can now come from consumer configs instead of via the CTOR, we need to do it there too.

Polishing here: https://github.com/garyrussell/spring-kafka/commit/32b66f6b143e954c313a239f3f71d27adc813d18 for your review/approval.

quaff:
Thanks, I'm waiting for release 2.1.1 .

artembilan:
Some Checkstyle failures:
```
:spring-kafka:checkstyleTest[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-kafka/spring-kafka/src/test/java/org/springframework/kafka/support/serializer/JsonSerializationTests.java:33: 'org.springframework.kafka.support.converter.AbstractJavaTypeMapper' should be separated from previous imports. [ImportOrder]

[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-kafka/spring-kafka/src/test/java/org/springframework/kafka/support/serializer/JsonSerializationTests.java:149: Line matches the illegal pattern 'Trailing whitespace'. [Regexp]
```

You can be sure locally before pushing via `check` Gradle command.

garyrussell:
@artembilan Sure, will do; but please review/merge #512 first; those test failures prevent this building for me.

garyrussell:
Merged as 36aabb40af1f37ec78895b1e1f573c5465b47063

@quaff Many thanks for the contribution - looking forward to more!

MartinX3:
So the consumer won't crash the application anymore, if there is an "untrusted" package on the kafka bus?

I get error messages about untrusted packages in a loop, but there is only ONE entry on the kafka bus.

garyrussell:
This change automatically "trusts" the package for the `Package` argument in the `listen` method in your example.

If the message contains a serialized `Package` object from a different `package`, you need to add a mapping from that package to the `Package` argument's package.

MartinX3:
But I have several microservices on the same topic on the same bus and not every microservice shares the same entities.
And I don't want to whitelist every package on every microservice which doesn't have the entity class.
I want that the unknown classes are ignored, but instead of ignoring it, it generates an endless loop of error messages and a CPU usage of 100%.

If someone evil or a application bug writes something on the kafka bus, it would crash every microservice.

garyrussell:
A deserializer can't filter records; it must return something.

You can subclass the deserializer, say pass some marker value to indicate the value was skipped, then add a `FilteringMessageListenerAdapter` to your listener container factory to filter out those records so they are not passed to your `listen` method.

If it's not clear how to do that, I can put together an example for you.

PR #512, opened 2017-12-10
Fix KafkaAdminBadContextTests

garyrussell:
Locally, I find the admin is sometimes created even with a bad host causing the test to fail.

Also, topic creation logged an error but did not throw an exception.

Fixed the test to expect one of two exceptions.

__cherry-pick to 2.0.x__


artembilan:
... and cherry-picked to `2.0.x` as https://github.com/spring-projects/spring-kafka/commit/dafea3284496ba3253ff03bfdb11f4aae28c9d93 after resolving conflicts.

Issue #511, opened 2017-12-10
Consumer committing old offsets on rebalance


tomvandenberge:
I'm seeing consumers committing offsets that are lower than the offset that it has committed before. This seems to happen during a rebalance.

I have a group of 4 consumers, reading from a single topic with 20 partitions, ack mode is RECORD (and enable.auto.commit=false), and no other non-default settings.

The 4 consumers are all happily consuming messages from their assigned partitions. Debug logging is enabled, which tells me exactly which offsets are committed for each partition. The consumer that is assigned partion 1 logs

```
Group g committed offset 2080124 for partition p-1
Group g committed offset 2080125 for partition p-1 
```

When one of the consumers is stopped, a rebalance is triggered on the remaining 3 consumers. The consumer that is assigned partition 1 shows the following sequence of log messages:

```
Revoking previously assigned partitions [p-0, p-3, p-4, p-1, p-2] for group g
Group g committed offset 2077901 for partition p-1
(Re-)joining group g
Group g committed offset 2077901 for partition p-1
Group g committed offset 2077902 for partition p-1
Performing assignment for group g using strategy range with subscriptions {...}
Finished assignment for group g: {...}
Successfully joined group g with generation 605
Setting newly assigned partitions [p-0, p-5, p-6, p-3, p-4, p-1, p-2] for group g
g fetching committed offsets for partitions: [p-0, p-5, p-6, p-3, p-4, p-1, p-2]
Group g committed offset 2028542 for partition p-0
Group g committed offset 2046396 for partition p-5
Group g committed offset 2065056 for partition p-6
Group g committed offset 2047834 for partition p-3
Group g committed offset 2055878 for partition p-4
Group g committed offset 2077901 for partition p-1
Group g committed offset 2058169 for partition p-2
partitions assigned:[p-0, p-5, p-6, p-3, p-4, p-1, p-2]

```

The committed offset of 2077901 on line 2 is lower than the last "regularly" committed offset of 2080125. It seems that the consumer somehow commits this incorrect offset as soon as the rebalance is started. It even commits two more incorrect offsets. After the consumer has joined the group, the offsets are fetched, including the incorrect one for p-1. After that, the consumer continues committing offsets starting from these fetched values, as expected.

This phenomenon happens almost every time a consumer is stopped, on a few partitions. I have not been able to reproduce it in a lab setting, unfortunately. Does any of you have an idea on what is happening here?

[...]

PR #515, opened 2017-12-11
Fix for consumer committing old offsets on rebalance

tomvandenberge:
This is a fix for the problem [raised here](https://github.com/spring-projects/spring-kafka/issues/511).

If record processing fails with a runtime exception, the actions that are taken are now (almost) the same as if it didn't fail.

The difference with the previous situation is that for record ack mode, the offset of the failed record is now immediately committed (and not added to `acks`). Also, for manual ack, the record is no longer added to `acks` anymore. This is also the case for successfully processed records, so I assume that the same approach should be taken for failed records.

I've added a test to show that the correct offsets are being committed.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -955,7 +955,20 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 			}
 			catch (RuntimeException e) {
 				if (this.containerProperties.isAckOnError() && !this.autoCommit && producer == null) {
-					this.acks.add(record);
+					if (this.isRecordAck) {
+						Map<TopicPartition, OffsetAndMetadata> offsetsToCommit =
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@tomvandenberge Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=515)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=515) to manually synchronize the status of this Pull Request.

artembilan:
Sounds good, @tomvandenberge , but would you mind to sign the CLA anyway. Otherwise we can't review and merge your solution.

Thank you!

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/d9782469b25ff1be76a5797942243d78b30b3756.

Cherry-picked to `2.0.x` and `master` as https://github.com/spring-projects/spring-kafka/commit/ebf7bb1d732ed4e94284d35fc439e610bbafbf1d after resolving minor conflicts.

Looking forward for more!

Issue #522, opened 2017-12-14
Nullpointer in spring-kafka


pradi:
we are using spring-kafka-2.0.1.RELEASE and get this error in production:

> exception java.lang.NullPointerException java.lang.NullPointerException: null at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.access$000(KafkaMessageListenerContainer.java:271) ~[spring-kafka-2.0.1.RELEASE.jar:2.0.1.RELEASE] at org.springframework.kafka.listener.KafkaMessageListenerContainer.getAssignedPartitions(KafkaMessageListenerContainer.java:152) ~[spring-kafka-2.0.1.RELEASE.jar:2.0.1.RELEASE] at

ref: https://stackoverflow.com/questions/47794253/possible-nullpointer-in-spring-kafka

pradi:
Hi,

is version spring-kafka 2.1.1 also compatible with kafka 0.11.x.x?

Thanks

artembilan:
Don't worry: we will back-port the fix up to `1.3.x`. So, `2.0.x` will be fixed as well.

garyrussell:
But, yes, the 1.0.0 kafka-clients will communicate ok with an 0.11.x.x broker.

garyrussell:
GitHub issues doesn't support multiple milestones so we just add the latest one.

PR #524, opened 2017-12-18
GH-522: Fix NPE in listener container

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/522

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -300,7 +300,8 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 	public String toString() {
 		return "KafkaMessageListenerContainer [id=" + getBeanName()
 				+ (this.clientIdSuffix != null ? ", clientIndex=" + this.clientIdSuffix : "")
-				+ ", topicPartitions=" + getAssignedPartitions()
+				+ ", topicPartitions="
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -179,11 +179,17 @@ public void setClientIdSuffix(String clientIdSuffix) {
...
+		ListenerConsumer listenerConsumer = this.listenerConsumer;
+		if (listenerConsumer != null) {
+			if (listenerConsumer.definedPartitions != null) {
+				return Collections.unmodifiableCollection(listenerConsumer.definedPartitions.keySet());
+			}
+			else if (listenerConsumer.assignedPartitions != null) {
+				return Collections.unmodifiableCollection(listenerConsumer.assignedPartitions);
+			}
+			else {
+				return null;
```
artembilan: I would suggest to fix `toString()` as well.
Right now it might finish with the `", topicPartitions=null"`, which might not be so useful.

Your thoughts?

Thanks

artembilan:
... and cherry-picked to `2.0.x` and `1.3.x`

pradi:
Hi, many thanks ... I still can not believe, that it happens so fast :)

pradi:
Hi, which will be the  spring-kafka-2.0.x version with the fix?

Thanks

garyrussell:
It will be in 2.0.3; we can release that this week.

Is there some reason you can't upgrade to 2.1.1, which includes the fix?

pradi:
Hi,
we will first update / fix to the smaller version and on the next release iteration to the next version

Issue #527, opened 2017-12-20
KafkaHandlerMethodFactoryAdapter: Replace DefaultFormattingConversionService with ConversionService Bean


zersetz-end:
I'd like to propose a change to the following lines of [KafkaListenerAnnotationBeanPostProcessor.java](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java)

```java
DefaultFormattingConversionService conversionService = new DefaultFormattingConversionService();
defaultFactory.setConversionService(conversionService);
```
Because although you can configure the MessageHandlerMethodFactory through an implementation of KafkaListenerConfigurer, you'd have to basically copy the code of KafkaListenerAnnotationBeanPostProcessor.KafkaHandlerMethodFactoryAdapter#createDefaultMessageHandlerMethodFactory to replace the ConversionService used for invocation of the annotated methods. So how about first trying to get the bean from the available BeanFactory and only using the DefaultFormattingConversionService only in case that fails. That way parameter conversion could be handled by simlpy implementing a Converter bean.

Example code:
```java
ConversionService conversionService;
try {
	conversionService =
			KafkaListenerAnnotationBeanPostProcessor.this.beanFactory.getBean(ConversionService.class);
}
catch (Exception e) {
	logger.warn("some warning", e);
	conversionService = new DefaultFormattingConversionService();
}
```

artembilan:
Sounds reasonable. Please, go ahead with the PR if you wish

PR #528, opened 2017-12-20
527: Replace DefaultFormattingConversionService with Bean

zersetz-end:
[Issue #527](https://github.com/spring-projects/spring-kafka/issues/527)

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2017 the original author or authors.
+ * Copyright 2014-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -50,6 +50,7 @@ import org.springframework.context.expression.StandardBeanExpressionResolver;
 import org.springframework.core.MethodIntrospector;
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.AnnotationUtils;
+import org.springframework.core.convert.ConversionService;
 import org.springframework.format.support.DefaultFormattingConversionService;
 import org.springframework.kafka.config.KafkaListenerConfigUtils;
 import org.springframework.kafka.config.KafkaListenerContainerFactory;
@@ -662,7 +663,17 @@ public class KafkaListenerAnnotationBeanPostProcessor<K, V>
 					(KafkaListenerAnnotationBeanPostProcessor.this.beanFactory instanceof ConfigurableBeanFactory ?
 							(ConfigurableBeanFactory) KafkaListenerAnnotationBeanPostProcessor.this.beanFactory : null);
 
-			DefaultFormattingConversionService conversionService = new DefaultFormattingConversionService();
+			ConversionService conversionService;
+			try {
+				conversionService =
+						KafkaListenerAnnotationBeanPostProcessor.this.beanFactory.getBean(ConversionService.class);
+			}
+			catch (Exception e) {
+				KafkaListenerAnnotationBeanPostProcessor.this.logger.warn(
+						"Failed to get bean for type ConversionService. "
+								+ "Using DefaultFormattingConversionService.", e);
+				conversionService = new DefaultFormattingConversionService();
+			}
 			defaultFactory.setConversionService(conversionService);
 
 			List<HandlerMethodArgumentResolver> argumentResolvers = new ArrayList<>();
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -50,7 +50,10 @@ import org.springframework.context.expression.StandardBeanExpressionResolver;
 import org.springframework.core.MethodIntrospector;
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.AnnotationUtils;
-import org.springframework.core.convert.ConversionService;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.core.convert.converter.GenericConverter;
+import org.springframework.format.Formatter;
+import org.springframework.format.FormatterRegistry;
 import org.springframework.format.support.DefaultFormattingConversionService;
 import org.springframework.kafka.config.KafkaListenerConfigUtils;
 import org.springframework.kafka.config.KafkaListenerContainerFactory;
@@ -99,6 +102,7 @@ import org.springframework.util.StringUtils;
  * @author Artem Bilan
  * @author Dariusz Szablinski
  * @author Venil Noronha
+ * @author Dimitri Penner
  *
  * @see KafkaListener
  * @see KafkaListenerErrorHandler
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@zersetz-end Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=528)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=528) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@zersetz-end Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=528)!

artembilan:
Thank you for contribution, @zersetz-end !

I didn't check the code yet, but that really would be better to have some test-case to confirm the requested functionality or demonstrate how it fixes some expectations.

spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
```diff
@@ -976,6 +1017,36 @@ public KafkaListenerErrorHandler voidSendToErrorHandler() {
...
+		}
+
+		@Bean
+		public ConversionService conversionService() {
+			ConversionServiceFactoryBean bean = new ConversionServiceFactoryBean();
+			bean.setConverters(Collections.singleton(
+					fooConverter()
+			));
+			bean.afterPropertiesSet();
+			return bean.getObject();
```
artembilan: You don't need to call `afterPropertiesSet()` in the `@Bean` definition method.
You don't need to call `FactoryBean.getObject()` in the `@Bean` definition method.
The Framework takes care about all the lifecycle callbacks and autowiring injections.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -664,7 +665,17 @@ private MessageHandlerMethodFactory createDefaultMessageHandlerMethodFactory() {
...
-			DefaultFormattingConversionService conversionService = new DefaultFormattingConversionService();
+			ConversionService conversionService;
+			try {
+				conversionService =
+						KafkaListenerAnnotationBeanPostProcessor.this.beanFactory.getBean(ConversionService.class);
+			}
+			catch (Exception e) {
+				KafkaListenerAnnotationBeanPostProcessor.this.logger.warn(
+						"Failed to get bean for type ConversionService. "
+								+ "Using DefaultFormattingConversionService.", e);
```
artembilan: I think we don't need this log message at all. At least on the `info` (or even `debug`) level...
And would be better to say:

> No ConversionService bean in the application context. Using DefaultFormattingConversionService as default instance.

zersetz-end:
I refactored the way Converter beans are added to the ConversionService after a little chat in spring-boot, because I wasn't really sure where the default ConversionService bean usually comes from. The title/issue content don't really match the implementation now, but I guess this is the better way of registering custom converters.

Not really sure where to document the feature though.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -690,6 +696,26 @@ protected boolean isEmptyPayload(Object payload) {
...
+				registry.addConverter(converter);
+			}
+			for (Formatter<?> formatter : getBeansOfType(Formatter.class)) {
+				registry.addFormatter(formatter);
+			}
+		}
+
+		private <T> Collection<T> getBeansOfType(Class<T> type) {
+			if (KafkaListenerAnnotationBeanPostProcessor.this.beanFactory instanceof ListableBeanFactory) {
+				return ((ListableBeanFactory) KafkaListenerAnnotationBeanPostProcessor.this.beanFactory).getBeansOfType(type).values();
```
artembilan: I checked the call stack and all of these `getBeansOfType()` is not good during `BeanPostProcessor.postProcessAfterInitialization()`. That is an "early BeanFactory initialization syndrome".

If we are going this `add*` way, we should consider to deffer their population until `afterSingletonsInstantiated()` phase.

zersetz-end:
I added documentation and changed the copyright dates. English isn't my native language, so if you have any remarks about the wording or the content, feel free to provide me an improved version. :-)

src/reference/asciidoc/kafka.adoc
```diff
@@ -1271,6 +1271,31 @@ public void listen(List<Foo> foos, @Header(KafkaHeaders.OFFSET) List<Long> offse
 
 Notice that you can still access the batch headers too.
 
+Starting with _versions 1.3.3_ the `org.springframework.core.convert.ConversionService` used by the default
```
artembilan: I don't think that we are going to back port it down into the `1.3.x`. (`2.1.1` should be good)
There is a workaround like `MessageHandlerMethodFactory` injection into the `KafkaListenerEndpointRegistrar` as you correctly mention below.

There should be `,` after version.

src/reference/asciidoc/kafka.adoc
```diff
@@ -1271,6 +1271,31 @@ public void listen(List<Foo> foos, @Header(KafkaHeaders.OFFSET) List<Long> offse
...
+`org.springframework.messaging.handler.annotation.support.MessageHandlerMethodFactory` to reslove parameters for the invocation
+of a listener method is provided all beans implementing any of the following interfaces:
```
artembilan: > ... is supplied with all beans ...

src/reference/asciidoc/kafka.adoc
```diff
@@ -1271,6 +1271,31 @@ public void listen(List<Foo> foos, @Header(KafkaHeaders.OFFSET) List<Long> offse
...
+
+[source, java]
+----
+@Bean
+public class JAXBFooConverter implements Converter<String, Foo> {
+    @Override
+    public Foo convert(String source) {
+        return JAXB.unmarshal(source, Foo.class);
+    }
+}
```
artembilan: This code sample just doesn't make sense. Please, remove it altogether.

src/reference/asciidoc/kafka.adoc
zersetz-end: I'm aware of the workaround, as i'm currently forced to use it. :-)

But I think the beauty of my solution is that unless you have to programatically configure any other properties of the listener handling that aren't confiurable through the `KafkaProperties`, you basically just have to put `@EnableKafka` somewhere and implement `Converter` for your incoming message types. Or maybe I'm misunderstanding the workaround, because this is what I have to do, to achieve the same behaviour:

``` java
@Component
public class KafkaConfig implements KafkaListenerConfigurer {

    @Autowired
    BeanFactory beanFactory;

    @Autowired
    ConversionService conversionService;

    @Override
    public void configureKafkaListeners(KafkaListenerEndpointRegistrar registrar) {
        registrar.setMessageHandlerMethodFactory(createDefaultMessageHandlerMethodFactory());
    }

    private MessageHandlerMethodFactory createDefaultMessageHandlerMethodFactory() {
        DefaultMessageHandlerMethodFactory defaultFactory = new DefaultMessageHandlerMethodFactory();
        defaultFactory.setBeanFactory(beanFactory);

        ConfigurableBeanFactory cbf =
                (beanFactory instanceof ConfigurableBeanFactory ?
                        (ConfigurableBeanFactory) beanFactory : null);

        defaultFactory.setConversionService(conversionService);

        List<HandlerMethodArgumentResolver> argumentResolvers = new ArrayList<>();

        argumentResolvers.add(new HeaderMethodArgumentResolver(conversionService, cbf));
        argumentResolvers.add(new HeadersMethodArgumentResolver());

        final GenericMessageConverter messageConverter = new GenericMessageConverter(conversionService);
        argumentResolvers.add(new MessageMethodArgumentResolver(messageConverter));
        argumentResolvers.add(new PayloadArgumentResolver(messageConverter) {

            @Override
            protected boolean isEmptyPayload(Object payload) {
                return payload == null || payload instanceof KafkaNull;
            }

        });

        defaultFactory.afterPropertiesSet();
        return defaultFactory;
    }

}
```

artembilan: Seems for me your workaround is overhead.
That should be like this:
```
@Override
public void configureKafkaListeners(KafkaListenerEndpointRegistrar registrar) {
        registrar.setMessageHandlerMethodFactory(defaultMessageHandlerMethodFactory());
}

@Bean
public MessageHandlerMethodFactory defaultMessageHandlerMethodFactory() {
    DefaultMessageHandlerMethodFactory defaultFactory = new DefaultMessageHandlerMethodFactory();
   defaultFactory.setConversionService(conversionService);
   return defaultFactory;
```

zersetz-end: Yes, I noticed that too. But this way it lacks the handling of `KafkaNull`. Which shouldn't cause any problems in my case. I'm also currently looking into adding a `Validator` to the `MessageHandlerMethodFactory'. Something you should consider adding in similar fashion some day. :-D

artembilan: Oh! I see what you mean about `KafkaNull`. Hard to read long code on GH 😄 .

Well, for that purpose we have `DefaultMessageHandlerMethodFactory.setCustomArgumentResolvers()`.

PR #534, opened 2018-01-14
Support local bean reference in @KafkaListener

garyrussell:
In the `@KafkaListener` bean post processor, allow property SpEL expressions to
access properties and methods on the bean being post-processed.

[See Stack Overflow](https://stackoverflow.com/questions/48162588/spring-expression-language-is-it-possible-to-use-this-in-sel/48171277#48171277).

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/annotation/BatchListenerConversionTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/annotation/BatchListenerConversionTests.java
@@ -70,18 +70,24 @@ public class BatchListenerConversionTests {
 	@Autowired
 	private KafkaTemplate<Integer, Foo> template;
 
-	@SuppressWarnings("unchecked")
 	@Test
 	public void testBatchOfPojos() throws Exception {
+		doTest(this.config.listener1(), "blc1");
+		doTest(this.config.listener2(), "blc2");
+	}
+
+	private void doTest(Listener listener, String topic) throws InterruptedException {
 		this.template.send(new GenericMessage<>(
-				new Foo("bar"), Collections.singletonMap(KafkaHeaders.TOPIC, "blc1")));
+				new Foo("bar"), Collections.singletonMap(KafkaHeaders.TOPIC, topic)));
 		this.template.send(new GenericMessage<>(
-				new Foo("baz"), Collections.singletonMap(KafkaHeaders.TOPIC, "blc1")));
-		assertThat(config.listener().latch1.await(10, TimeUnit.SECONDS)).isTrue();
-		assertThat(config.listener().received).isInstanceOf(List.class);
-		assertThat(((List<?>) config.listener().received).size()).isGreaterThan(0);
-		assertThat(((List<?>) config.listener().received).get(0)).isInstanceOf(Foo.class);
-		assertThat(((List<Foo>) config.listener().received).get(0).bar).isEqualTo("bar");
+				new Foo("baz"), Collections.singletonMap(KafkaHeaders.TOPIC, topic)));
+		assertThat(listener.latch1.await(10, TimeUnit.SECONDS)).isTrue();
+		assertThat(listener.latch2.await(10, TimeUnit.SECONDS)).isTrue();
+		assertThat(listener.received.size()).isGreaterThan(0);
+		assertThat(listener.received.get(0)).isInstanceOf(Foo.class);
+		assertThat(listener.received.get(0).bar).isEqualTo("bar");
+		assertThat((listener.receivedTopics).get(0)).isEqualTo(topic);
+		assertThat((listener.receivedPartitions).get(0)).isEqualTo(0);
 	}
 
 	@Configuration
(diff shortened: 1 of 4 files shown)


spring-kafka/src/test/java/org/springframework/kafka/annotation/BatchListenerConversionTests.java
```diff
@@ -134,24 +140,55 @@ public StringJsonMessageConverter converter() {
...
-		@KafkaListener(topics = "blc1")
-		public void listen(List<Foo> foos, @Header(KafkaHeaders.OFFSET) List<Long> offsets) {
-			this.received = foos;
+		private List<Integer> receivedPartitions;
+
+		public Listener(String topic) {
+			this.topic = topic;
+		}
+
+		@KafkaListener(topics = "#{__listenerBean__.topic}", groupId = "#{__listenerBean__.topic}.group")
```
artembilan: Can't advertise the feature like SpEL evaluation context variable and therefore register the outer object under a `#listener` name? More convenient to use and familiar syntax for SpEL users: https://docs.spring.io/spring/docs/5.0.2.RELEASE/spring-framework-reference/core.html#expressions-ref-variables ?

What I feel fragile for is that funky `beanRef` on the annotation...

Just my feeling from the end-user perspective.
If you are not agree with me, I'll merge it as you did.
Thanks

garyrussell: That was my initial plan, but we don't have access to the SpEL evaluation context - it is a single instance that's cached in the `StandardBeanExpressionResolver` - it would require replacing all that resolution infrastructure with our own.

artembilan: Yes, I checked the code there, bu I mean *pretend* like we have it as a SpEL variable , by the name of the object in scope you use already.

garyrussell: That works already, out of the box...

```java
@Bean
public Listener listener1() {
    return new Listener("topic1");
}
```

with

```java
    @KafkaListener(topics = "#{listener1.topic}",
        groupId = "#{listener1.topic}.group")
    public void listen(...) {
        ...
    }
```

But we're trying to solve a different problem, where we don't know the bean name while writing the code...

```java
@Bean
public Listener listener1() {
    return new Listener("topic1");
}

@Bean
public Listener listener2() {
    return new Listener("topic2");
}
```

We somehow need a reference to the current bean, but not by the bean name. Yes, ideally, a variable `#currentBean` would be best but, for the reason I specified above, we can't do that without copying a lot of code.

artembilan: OK. Let's go simpler discussion way!
I don't like the name for the object in `scope` as `__listenerBean__`. I would prefer `#listener` instead. This way it mimics the well-known SpEL variable syntax. 

Also I don't like that new `beanRef` annotation attribute. Don't know yet why, but my feeling like I would stay away of it. Although I agree about the argument "unlikely event that you have an actual bean called" especially in this point release. 

So, having all of this thinking in my head I only suggest to rename the default  `__listenerBean__` key to the `#listener`. And that's all.

Let me know if my point is still unclear!

garyrussell: We can't use `#listener` because SpEL will think it's a real variable and fail to find it.

garyrussell: I'd be ok with `_listener` or `_listener_` as the default.

artembilan: Ah! Good point!

OK. One more attempt. The `StandardBeanExpressionResolver` has this:
```
/**
 * Template method for customizing the expression evaluation context.
 * <p>The default implementation is empty.
 */
protected void customizeEvaluationContext(StandardEvaluationContext evalContext) {
}
```
Looks like this is the place where we could inject a SpEL variable for the bean in the method:
```
protected void processListener(MethodKafkaListenerEndpoint<?, ?> endpoint, KafkaListener kafkaListener, Object bean,
			Object adminTarget, String beanName) {
```
But I agree that is more coding, `ThreadLocal`, param propagation etc. And in the end technically we end up with the same logic for end user.

So, let me know if you would like to play with this or let's leave as is and merge!

artembilan: > I'd be ok with `_listener` or `_listener_` as the default.

Yeah... That really might be two `_` in the beginning, but I'm some lost to read dot after two `_`...

garyrussell: ok; I will change it; and do no more; I think we've spent enough time on this 😄

PR #538, opened 2018-01-21
Wait for partitions to be assigned when consuming from embedded topics

elliotkennedy:
Created issue #539

I think there is a bug in `EmbeddedKafka.consumeFromAnEmbeddedTopic(Consumer consumer, String topic)`

. It does not wait to be assigned partitions from the embedded topic as is the case with `EmbeddedKafka.onsumeFromAllEmbeddedTopics(Consumer consumer)`.

I created a test, which fails without this change [here](https://github.com/elliotkennedy/kafka-test-demo/blob/master/src/test/java/io/elken/kafkatestdemo/KafkaEmbeddedTest.java).

When a consumer is created with `EmbeddedKafka.consumeFromAllEmbeddedTopics(Consumer consumer)`, all the  messages are received.

When a consumer is created with `EmbeddedKafka.consumeFromAnEmbeddedTopic(Consumer consumer, String topic)`, the messages are not all received.


pivotal-issuemaster:
@elliotkennedy Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=538)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=538) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@elliotkennedy Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=538)!

elliotkennedy:
Thanks for the feedback @artembilan - I've updated as requested but I wasn't sure where to put the test. I dropped it in the tests project but that creates a dependency on the main project. Take a look and let me know if you'd like to move it somewhere else :+1: 

The test polling loop could be made smarter too, open to suggestions on that if you think it might be flaky.

Thanks,
Elliot

build.gradle
```diff
@@ -211,6 +211,8 @@ project ('spring-kafka-test') {
 
 		compile ("org.hamcrest:hamcrest-all:$hamcrestVersion", optional)
 		compile ("org.assertj:assertj-core:$assertjVersion", optional)
+
+        testCompile project(':spring-kafka')
```
artembilan: No, your test definitely should be in the `spring-kafka` module, in the `test` directory. And I think the `org.springframework.kafka.kstream` package should be good candidate for the place.

spring-kafka-test/src/test/java/org/springframework/kafka/test/rule/KafkaEmbeddedStreamTests.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.Map;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Elliot Kennedy
+ */
+@RunWith(SpringRunner.class)
+public class KafkaEmbeddedStreamTests {
```
artembilan: I think structure of the test-case is good enough.
We need `@DirtiesContext` on the class level to clear the application context after the test execution.
There is an `@EmbeddedKafka` instead of `@ClassRule`. Also I think that `kafkaTemplate` can be declared as a `@Bean` as well.

For the Kafka Streams we have an `@EnableKafkaStreams` as well.

Just consider to use all of them in your test or leave as is and we polish on merge!

But yeah... would be better to move the test to the appropriate module.

elliotkennedy:
Cool, thanks!

I'm slightly worried there is a race condition in ```KafkaTestUtils.getRecords(Consumer consumer)```, it doesn't receive all the messages in time when using the @ClassRule method, that's why I was looping in the first place. Tempted to add a ```getRecords(int numberOfRecords) function in another PR?

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/6b74d8ca05e6a8e5f090d2ec0c6899974215adfd and cherry-picked to `2.0.x` and `1.3.x`.

Regarding ` KafkaTestUtils.getRecords(Consumer consumer)`... Well, let's raise a separate issue and discuss the problem there with an appropriate test-case to reproduce!

Thank you for contribution here. Looking forward for more!

Issue #537, opened 2018-01-21
can i print some log when a message has done?  


dragontree101:
i want to log some handle message info (time and offse and so on) when message has done, 
but i only found `ErrorHandler` interface when consume message has error.

i don't whether other interface like `ErrorHandler` it can let me record message when consume has done in normal status, i will record some info?

i have read some source code , but not found

i found have commit log, when debug level is debug, but i think in pro env, log level info is the best, so i don't want to set debug level to debug. 

thanks!

garyrussell:
We could make that log level configurable (contributions are welcome - see [contributing](https://github.com/spring-projects/spring-kafka/blob/master/CONTRIBUTING.adoc).

In the meantime, you could do something like this...

```java
@SpringBootApplication
public class Kgh537Application {

	private static final Logger logger = LoggerFactory.getLogger(Kgh537Application.class);

	public static void main(String[] args) {
		SpringApplication.run(Kgh537Application.class, args);
	}

	@Bean
	public ApplicationRunner runner(KafkaTemplate<String, String> kafkaTemplate) {
		return args -> kafkaTemplate.send("kgh537", "foo");
	}

	@Bean
	public ConsumerFactory<Object, Object> consumerFactory(KafkaProperties properties) {
		return new DefaultKafkaConsumerFactory<Object, Object>(properties.buildConsumerProperties()) {

			@Override
			public Consumer<Object, Object> createConsumer() {
				return wrap(super.createConsumer());
			}

			@Override
			public Consumer<Object, Object> createConsumer(String groupId, String clientIdPrefix,
					String clientIdSuffix) {
				return wrap(super.createConsumer(groupId, clientIdPrefix, clientIdSuffix));
			}

			@Override
			public Consumer<Object, Object> createConsumer(String groupId, String clientIdSuffix) {
				return wrap(super.createConsumer(groupId, clientIdSuffix));
			}

			@Override
			public Consumer<Object, Object> createConsumer(String clientIdSuffix) {
				return wrap(super.createConsumer(clientIdSuffix));
			}

			@SuppressWarnings("unchecked")
			private Consumer<Object, Object> wrap(Consumer<Object, Object> createConsumer) {
				ProxyFactory pf = new ProxyFactory(createConsumer);
				pf.addAdvice(new MethodInterceptor() {

					@Override
					public Object invoke(MethodInvocation invocation) throws Throwable {
						Method method = invocation.getMethod();
[...]

PR #545, opened 2018-01-28
GH-537: Configurable log level for offset commits

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/537

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -533,7 +533,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 								return;
 							}
 						}
-						ListenerConsumer.this.commitLogger.log("Committing on assignment: " + offsets);
+						ListenerConsumer.this.commitLogger.log(() -> "Committing on assignment: " + offsets);
 						if (ListenerConsumer.this.transactionTemplate != null &&
 								ListenerConsumer.this.kafkaTxManager != null) {
 							ListenerConsumer.this.transactionTemplate.execute(new TransactionCallbackWithoutResult() {
@@ -757,7 +757,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 			Map<TopicPartition, OffsetAndMetadata> commits = Collections.singletonMap(
 					new TopicPartition(record.topic(), record.partition()),
 					new OffsetAndMetadata(record.offset() + 1));
-			this.commitLogger.log("Committing: " + commits);
+			this.commitLogger.log(() -> "Committing: " + commits);
 			if (this.containerProperties.isSyncCommits()) {
 				this.consumer.commitSync(commits);
 			}
@@ -994,7 +994,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 							Collections.singletonMap(new TopicPartition(record.topic(), record.partition()),
 									new OffsetAndMetadata(record.offset() + 1));
 					if (producer == null) {
-						this.commitLogger.log("Committing: " + offsetsToCommit);
+						this.commitLogger.log(() -> "Committing: " + offsetsToCommit);
 						if (this.containerProperties.isSyncCommits()) {
 							this.consumer.commitSync(offsetsToCommit);
 						}
@@ -1019,7 +1019,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 						Map<TopicPartition, OffsetAndMetadata> offsetsToCommit =
 								Collections.singletonMap(new TopicPartition(record.topic(), record.partition()),
 										new OffsetAndMetadata(record.offset() + 1));
-						this.commitLogger.log("Committing: " + offsetsToCommit);
+						this.commitLogger.log(() -> "Committing: " + offsetsToCommit);
 						if (this.containerProperties.isSyncCommits()) {
 							this.consumer.commitSync(offsetsToCommit);
 						}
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1209,8 +1225,8 @@ private void commitIfNecessary() {
 				this.logger.debug("Commit list: " + commits);
 			}
 			if (!commits.isEmpty()) {
-				if (this.logger.isDebugEnabled()) {
-					this.logger.debug("Committing: " + commits);
+				if (LogSupport.isEnabled(this.commitLogLevel, ListenerConsumer.this.logger)) {
```
artembilan: Looks like duplicated code. More over I see that `commitLogLevel` is `final`. So, the decision about `isEnabled` we can make once in the ctor as well. Are there any reason do not? For example we change the logging level at runtime, so with my suggestion the current `ListenerConsumer` won't pick up a fresh config - only a fresh one after restarting container.

Just share that to be sure what we prefer: optimization over convenience or flexibility in the end.
Thanks

spring-kafka/src/main/java/org/springframework/kafka/support/LogSupport.java
```diff
@@ -0,0 +1,135 @@
...
+			case DEBUG:
+				logger.debug(message);
+				break;
+			case TRACE:
+				logger.trace(message);
+				break;
+		}
+	}
+
+	public static void logAt(Level level, Log logger, String message, Throwable t) {
```
artembilan: Can't we move these two method to the `LogSupport.Level` to avoid extra argument for `static` methods and have a context and encapsulation with the polymorphism as they are stated in Java. 😄  

Plus pay attention to the logging method of `Log` contract. They are `Object`-based for the `message` argument. I would stay with that.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Hmmm - they can't change the `commitLogLevel` dynamically but, let's say it's INFO and the category is currently WARN - they can change the category to INFO at runtime via a boot actuator.

spring-kafka/src/main/java/org/springframework/kafka/support/LogSupport.java
garyrussell: Good points, both.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: Right, that was my concern do not do that once. So, convenience is more important than optimization. OK, let it be.

May we consider to move this `isEnabled()` to the `LogSupport.Level` as well?

garyrussell:
Squashed and re-pushed.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Yes; I did that too - made more sense once I removed the statics.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -529,9 +533,7 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
 								return;
 							}
 						}
-						if (ListenerConsumer.this.logger.isDebugEnabled()) {
-							ListenerConsumer.this.logger.debug("Committing on assignment: " + offsets);
-						}
+						ListenerConsumer.this.commitLogger.log("Committing on assignment: " + offsets);
```
artembilan: That's not good. We still have to check `isEnabled()` before concatenating strings.
Or what am I missing when we do that?

garyrussell: doh - you are correct - will rework

spring-kafka/src/main/java/org/springframework/kafka/support/LogIfLevelEnabled.java
```diff
@@ -0,0 +1,149 @@
...
+ */
+public final class LogIfLevelEnabled {
+
+	private final Log logger;
+
+	private final Level level;
+
+	public LogIfLevelEnabled(Log logger, Level level) {
+		this.logger = logger;
+		this.level = level;
```
artembilan: `Assert.notNull()` for both

src/reference/asciidoc/kafka.adoc
```diff
@@ -498,6 +499,10 @@ Refer to the JavaDocs for `ContainerProperties` for more information about the v
 
 Since version _2.1.1_, a new property `logContainerConfig` is available; when true, and INFO logging is enabled, each listener container will write a log message summarizing its configuration properties.
 
+By default, logging of topic offset commits is performed with the DEBUG logging level.
+Starting with _version 2.1.1_, there is a new property in `ContainerProperties` called `commitLogLevel` which allows you to specify the log level for these messages.
```
artembilan: `2.1.2` ?

src/reference/asciidoc/whats-new.adoc
```diff
@@ -22,6 +22,13 @@ Starting with _version 2.1.1_, it is now possible to set the `client.id` prefix
 Previously, to customize the client id, you would need a separate consumer factory (and container factory) per listener.
 The prefix is suffixed with `-n` to provide unique client ids when using concurrency.
 
+
+==== Logging Offset Commits
+
+By default, logging of topic offset commits is performed with the DEBUG logging level.
+Starting with _version 2.1.1_, there is a new property in `ContainerProperties` called `commitLogLevel` which allows you to specify the log level for these messages.
```
artembilan: `2.1.2` ?

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/b1781766d220f83a78d0cfb120e1da7d8328a503

Issue #548, opened 2018-01-30
Make StreamsConfig customizable in StreamsBuilderFactoryBean


sobychacko:
Currently, `StreamsBuilderFactoryBean` constructor requires a `StreamsConfig`. There are use cases in which it is required to set the `StreamsConfig` later than when the factory is created, but right before the  `start` method is called. Also, it will be beneficial if we can set the `phase` on the bean to `Integer.MAX_VALUE` as we need to ensure that the lifecycle for the object created by the factory is started only after all the other components are started.

garyrussell:
I generally prefer to leave some "breathing" room, in case some other bean needs to be started after this; so I would suggest `Integer.MAX_VALUE - 1000` or similar.

PR #551, opened 2018-02-01
GH-548: StreamsBuilderFactoryBean enhancements

sobychacko:
Make StreamsConfig customizable in StreamsBuilderFactoryBean
Set the phase on StreamsBuilderFactoryBean to Integer.MAX_VALUE - 1000
Adding tests

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
@@ -158,7 +158,9 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 
 	@Override
 	protected StreamsBuilder createInstance() throws Exception {
-		Assert.notNull(this.streamsConfig, "'streamsConfig' must not be null");
+		if (this.autoStartup) {
+			Assert.notNull(this.streamsConfig, "'streamsConfig' must not be null");
+		}
 		return new StreamsBuilder();
 	}
 
@@ -188,6 +190,7 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 	public synchronized void start() {
 		if (!this.running) {
 			try {
+				Assert.notNull(this.streamsConfig, "'streamsConfig' must not be null");
 				this.kafkaStreams = new KafkaStreams(getObject().build(), this.streamsConfig, this.clientSupplier);
 				this.kafkaStreams.setStateListener(this.stateListener);
 				this.kafkaStreams.setUncaughtExceptionHandler(this.exceptionHandler);
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
@@ -36,6 +36,7 @@ import org.springframework.util.Assert;
  *
  * @author Artem Bilan
  * @author Ivan Ursul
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
```diff
@@ -36,14 +36,15 @@
...
  * @since 1.1.4
  */
 public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilder> implements SmartLifecycle {
 
 	private static final int DEFAULT_CLOSE_TIMEOUT = 10;
 
-	private final StreamsConfig streamsConfig;
+	private StreamsConfig streamsConfig;
 
 	private final CleanupConfig cleanupConfig;
```
artembilan: Nit-pick: the fields order should be like:
`static final`
`static`
`final`
non-`volatile`
`volatile`

So, since the `streamsConfig` isn't `final` any more should be moved below `cleanupConfig`.

spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
```diff
@@ -63,6 +64,14 @@
 
 	private volatile boolean running;
 
+	/**
+	 * Default constructor.
```
artembilan: That's not enough.

We definitely need to say more in which state the object is created.

spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
```diff
@@ -107,6 +116,10 @@ public StreamsBuilderFactoryBean(Map<String, Object> streamsConfig, CleanupConfi
 		this.cleanupConfig = cleanupConfig;
 	}
 
+	public void setStreamsConfig(StreamsConfig streamsConfig) {
```
artembilan: Need JavaDocs together with `@since`

spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
```diff
@@ -166,6 +179,7 @@ public void stop(Runnable callback) {
 	public synchronized void start() {
 		if (!this.running) {
 			try {
+				Assert.notNull(this.streamsConfig, "'streamsConfig' must not be null");
```
artembilan: Would be better to have this in the `setStreamsConfig()` and in the `createInstance()`.

The last one will work from the application  start up to notify end-user that he/she hasn't called  `setStreamsConfig()` at all.

And the first one won't let to set `null` at all.

It's not good to check it from the `start()` because it may happen on a different thread and the failure might mislead a bit from the initialization phase or re-initialization at runtime.

spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
```diff
@@ -137,6 +158,7 @@ public void setCloseTimeout(int closeTimeout) {
 
 	@Override
 	protected StreamsBuilder createInstance() throws Exception {
+		Assert.notNull(this.streamsConfig, "'streamsConfig' must not be null");
```
artembilan: You know, Soby, looking at this one more time I retract my previous requirement.

Looks like we are indeed are good to live with the state without config until the first `start()` and that really may happen somewhere late by logic.

So, yes, let's check for the `streamsConfig` if `autoStartup == true` and one more check in the `start()` as you had before.

Make sense?

Sorry for the noise.

PR #552, opened 2018-02-01
Support stateful retry

garyrussell:


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RetryingMessageListenerAdapter.java
```diff
@@ -75,13 +84,38 @@ public RetryingMessageListenerAdapter(MessageListener<K, V> messageListener, Ret
...
+		this(messageListener, retryTemplate, recoveryCallback, false);
+	}
+
+	/**
+	 * Construct an instance with the provided template, callback and delegate.
+	 * @param messageListener the delegate listener.
+	 * @param retryTemplate the template.
+	 * @param recoveryCallback the recovery callback; if null, the exception will be
+	 * thrown to the container after retries are exhausted.
+	 * @param stateful true for stateful retry.
```
artembilan: `@since 2.1.3` ?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RetryingMessageListenerAdapter.java
```diff
@@ -75,13 +84,38 @@ public RetryingMessageListenerAdapter(MessageListener<K, V> messageListener, Ret
...
+		this.stateful = stateful;
 	}
 
 	@Override
 	public void onMessage(final ConsumerRecord<K, V> record, final Acknowledgment acknowledgment,
 			final Consumer<?, ?> consumer) {
+		RetryState retryState = null;
+		String recordKey = null;
+		if (this.stateful) {
+			recordKey = record.topic() + "-" + record.partition() + "-" + record.offset();
```
artembilan: I think we should mention somewhere in JavaDocs what is used for the stateful cache key.

Should we have a strategy on the matter, though?

garyrussell: I don't think we need a strategy; this is really the only logical way to uniquely identify a record.

src/reference/asciidoc/kafka.adoc
```diff
@@ -1073,6 +1074,18 @@ See its javadocs for more information.
 A retry adapter is not provided for any of the batch <<message-listeners, message listeners>> because the framework has no knowledge of where, in a batch, the failure occurred.
 Users wishing retry capabilities, when using a batch listener, are advised to use a `RetryTemplate` within the listener itself.
 
+[[stateful-retry]]
+===== Stateful Retry
+
+It is important to understand that the retry discussed above suspends the consumer thread (if a `BackOffPolicy` is used); there is no interaction with the broker during retry.
+For lengthy retry sequences, if the `max.poll.interval.ms` property is exceeded, the broker will think the consumer is dead and will revoke the assigned partitions and perform a rebalance.
```
maciekpapiez: I'd suggest rephrasing it in order to differentiate the `max.poll.interval.ms` timeout (the server says "you've spent too much time consuming, I'm giving the partition to someone else") and the `session.timeout.ms` (the server didn't receive any heartbeat from the consumer and considers it as a dead one).

Example:
```
For lengthy retry sequences, if the `max.poll.interval.ms` property is exceeded (which means that consumer is processing the record for too long), the broker will revoke the assigned partitions and perform a rebalance. This can cause another available consumer to consume the record, thus producing duplicates.
```

src/reference/asciidoc/kafka.adoc
```diff
@@ -1073,6 +1074,18 @@ See its javadocs for more information.
 A retry adapter is not provided for any of the batch <<message-listeners, message listeners>> because the framework has no knowledge of where, in a batch, the failure occurred.
 Users wishing retry capabilities, when using a batch listener, are advised to use a `RetryTemplate` within the listener itself.
 
+[[stateful-retry]]
+===== Stateful Retry
+
+It is important to understand that the retry discussed above suspends the consumer thread (if a `BackOffPolicy` is used); there is no interaction with the broker during retry.
```
maciekpapiez: `there is no interaction with the broker during retry`

During retry, the background heartbeat thread is alive and can interact with the broker. Does `there is no poll() call to the broker during retry` sound better?

src/reference/asciidoc/kafka.adoc
garyrussell: Thanks; yes, this assumes users already understand the difference between `max.poll.interval.ms` and `session.timeout.ms`.

PR #554, opened 2018-02-03
Replying template - add getter for reply topic(s)

garyrussell:
- supports the SIK outbound gateway to determine if there's a default reply topic.
- supports the SIK validating a reply topic header if present in the outbound message.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
@@ -18,8 +18,7 @@ package org.springframework.kafka.requestreply;
 
 import java.nio.ByteBuffer;
 import java.time.Instant;
-import java.util.Arrays;
-import java.util.Collections;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.UUID;
@@ -27,10 +26,10 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Consumer;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.internals.RecordHeader;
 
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
```diff
@@ -100,6 +107,50 @@ public void setReplyTimeout(long replyTimeout) {
...
+	public void setAutoStartup(boolean autoStartup) {
+		this.autoStartup = autoStartup;
+	}
+
+	/**
+	 * Return the topics configured for the replying listener container.
+	 * @param patternConsumer called if the container is subscribed to a pattern.
+	 * @return the topics or an empty list if the container is subscribed to a pattern.
+	 */
+	public List<String> obtainReplyTopics(Consumer<Pattern> patternConsumer) {
```
artembilan: I don't feel like this `Consumer` is better than just return a set of configured topics/patters and let end-user to filter or process them as it is convenient for them.

OTOH why can't we return a set of **assigned** topics from the:
```
/**
 * Return the {@link TopicPartition}s currently assigned to this container,
 * either explicitly or by Kafka; may be null if not assigned yet.
 * @return the {@link TopicPartition}s currently assigned to this container,
 * either explicitly or by Kafka; may be null if not assigned yet.
 */
public Collection<TopicPartition> getAssignedPartitions() {
```
?

Seems for me this way we don't need to worry about patterns at all.

WDYT?

garyrussell: AH - good catch; thanks; forgot about that.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
```diff
@@ -100,6 +104,38 @@ public void setReplyTimeout(long replyTimeout) {
...
+	public void setAutoStartup(boolean autoStartup) {
+		this.autoStartup = autoStartup;
+	}
+
+	/**
+	 * Return the topics configured for the replying listener container.
+	 * @param patternConsumer called if the container is subscribed to a pattern.
+	 * @return the topics or an empty list if the container is subscribed to a pattern.
+	 */
+	public Collection<TopicPartition> obtainReplyTopics(Consumer<Pattern> patternConsumer) {
```
artembilan: I think you are in a hurry.
The `patternConsumer` is redundant and misleading now.

Plus we have talked about returning *topic names* for the reply correlation logic (I guess).
But now you return full `TopicPartition`.

Well, that might be useful on the end-user side, indeed, but let's rename the method then!

PR #556, opened 2018-02-03
Resolve property placeholders in EmbeddedKafka topics

### LINKED ISSUE #555: Resolve property placeholders in EmbeddedKafka topics
I'm working on a project where we are configuring all of our topics in property files, so our tests have to know about the contents of the property files to configure the correct topics. The alternative to this is to use `@TestPropertySource` to override our topic properties, but we have a lot of topics, so it's not that convenient. 

Ideally the topic attribute of `@EmbeddedKafka` would be able to accept and resolve property placeholders.

Example:
```java
@EmbeddedKafka(topics = { "${kafka.topic.one}", "${kafka.topic.two}" })
```

zacholauson:
Fixes #555

* Update EmbeddedKafkaContextCustomizer to resolve property placeholders
* Update KafkaStreamsTests to use topic from property
* Add property resolving details to EmbeddedKafka javadoc and testing.adoc

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
@@ -51,6 +51,7 @@ import org.springframework.kafka.test.rule.KafkaEmbedded;
  *
  * @author Artem Bilan
  * @author Elliot Metsger
+ * @author Zach Olauson
  *
  * @since 1.3
  *
@@ -85,6 +86,8 @@ public @interface EmbeddedKafka {
 	int partitions() default 2;
 
 	/**
+	 * Topics that should be created
+	 * Topics may contain property placeholders, e.g. {@code topics = "${kafka.topic.one:topicOne}"}
 	 * @return the topics to create
 	 */
 	String[] topics() default { };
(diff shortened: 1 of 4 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
```diff
@@ -85,6 +85,8 @@
 	int partitions() default 2;
 
 	/**
+	 * Topics that should be created
+	 * Topics may contain property placeholders, e.g. {@code topics = "${kafka.topic.one:topicOne}"}
```
zacholauson: I wasn't sure if javadoc counted as a cosmetic change, so I didn't add an `@author` tag. Let me know if that needs to be added.

artembilan: That’s change anyway, so, please, add your name to the `@author`. Thanks

zacholauson: Added. Thanks @artembilan

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/94d4fce8930e5b90a860c87649c7d1634cb98343.

Thank you for contribution, @zacholauson .
Looking forward for more!

Issue #43, opened 2016-03-30
Provide a Reactive alternative to KafkaOperations and KafkaTemplate


sdeleuze:
[KafkaOperations](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java) and [KafkaTemplate](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java) does not currently provide a way to perform truly non-blocking operations since they are using [`java.util.concurrent.Future`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html) where the only way to retrieve the result is a [blocking `get()`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html#get%28%29) method.

Providing a Reactive alternative to these classes using [Reactive Streams](http://www.reactive-streams.org/) and [Reactor Core](https://spring.io/blog/2016/03/11/reactor-core-2-5-becomes-a-unified-reactive-foundation-on-java-8) types, like in our new [Reactive HTTP client](https://github.com/spring-projects/spring-reactive/issues/13#issuecomment-163975472) (which is a Reactive alternative to `RestTemplate`), in the Reactive Spring Data work that is about to start (see this [ReactiveMongoOperations](https://github.com/mp911de/spring-data-mongodb/blob/7c758318b09106b4d36bb0d4927606f6ad9f771b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ReactiveMongoOperations.java#L71) draft) or in the [new Cloud Foundry Java client](https://github.com/cloudfoundry/cf-java-client#usage) would enable truly async non-blocking and Reactive usage of `spring-kafka` and would be a huge improvement in term of scalability, especially when used with Reactive Spring Web or Reactive Spring Data support, or even with the [upcoming Reactive support in Spring Cloud Stream 1.1](https://github.com/spring-cloud/spring-cloud-stream/issues/458).

Based on all the work we have done on this topic on [Spring Reactive](https://github.com/spring-projects/spring-reactive) to prepare the Reactive support of Spring Framework 5, I would advise to use:
[...]

PR #559, opened 2018-02-06
Provide a Reactive alternative to KafkaOperations and KafkaTemplate

marknorkin:
Provides operations with reactive types on top of kafka template.
Opened for discussion, please share your thoughts
 
Related discussion #43.


artembilan:
@marknorkin ,

Thank you for the contribution of course, but unfortunately this is not what we are looking for.

This is not true Reactive solution (wrapping around `ListenableFuture`) and it would be much better to make it based on the existing Reactor team effort on the matter - **Reactor Kafka**: https://projectreactor.io/docs/kafka/release/reference/

marknorkin:
@artembilan thought this was the original intent of the issue discussion. Anyway I close the ticket for now, because the reactor kafka initiative absolutely is a way to go. Sorry for misunderstanding.

marknorkin:
Hello, @artembilan 
Based on your feedback I've looked at reactor-kafka and did a basic implementation for spring-kafka. Would you mind take a look into it ? If this is what project aiming for I could proceed with it then.
Thank you in advance

artembilan:
Thanks for the update, @marknorkin !

Unfortunately I still think that this feature is for the next `2.2` version and this is some kind of in low priority for me for review. I just mean that I will come to you for review, but somewhere late.

marknorkin:
@artembilan  That is fine with me and surely I understand that it maybe not the first priority feature for now. Could you maybe give me some hint on when the 2.2 planned to be released or when the work for this release/feature will start, please ?

artembilan:
I think somewhere after Spring Boot `2.0 GA`: https://github.com/spring-projects/spring-boot/milestone/52

hartmut-co-uk:
Hi @marknorkin thanks for the great work - started playing around with your branch for a little bit.
Quick question - if I got 10 partitions - how can I get a pool of 10 consumer threads based on the `ReactiveKafkaConsumerTemplate` to then work off that with e.g. `flatMapSequential` for max throughput? 
Or would that be the wrong approach? Can't get my head around fully yet... :-)

marknorkin:
Hi @hartmut-co-uk, thanks for checking this out. 

As for your case, I should say that reactive consumer template based on kafka receiver from reactor-kafka project and this receiver is associated with a single instance of kafka consumer and as underlying 
consumer is not thread safe, so the receiver and template aren't thread-safe too. 

So as a consequence in reactive way the mapping is one template/one consumer.

As for subscribing you probably might want to look at ReceiverOptions::subscription and ReceiverOptions::assignment.

hartmut-co-uk:
thanks, had more time to test - getting really good performance overall.

garyrussell:
I haven't had a chance to look at this (and not sure I could add much value, given my current lack of Reactor chops) but it sounds promising.

We should consider cutting a 2.1.x branch soon, and making master 2.2.x so we can start building milestones when you guys think we have reached sufficient stability.

garyrussell:
On the other hand, since this is pretty much all new code, and `reactor-kafka` is `optional`, I am not sure we should force a new minor version (2.2) for this; will discuss with @artembilan next week.

artembilan:
I'm not a fun of new features in the point releases, especially when we have a workaround, but if you insist, I'm along. However that still doesn't matter that I have a time and knowledge resources to review it now.

spring-kafka/src/test/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplateTransactionIntegrationTests.java
```diff
@@ -0,0 +1,288 @@
...
+		ProducerRecord<Integer, String> producerRecord =
+				new ProducerRecord<>(REACTIVE_INT_KEY_TOPIC, DEFAULT_PARTITION, DEFAULT_TIMESTAMP, DEFAULT_KEY, DEFAULT_VALUE);
+		long correlationMetadata = 1L;
+		SenderRecord<Integer, String, Long> senderRecord = SenderRecord.create(producerRecord, correlationMetadata);
+
+		Mono<SenderResult<Long>> publisher = reactiveKafkaProducerTemplate.sendTransactionally(senderRecord);
+		StepVerifier.create(publisher)
+				.assertNext(senderResult ->
+						Assertions.assertThat(senderResult.recordMetadata())
+								.has(match(recordMetadata -> REACTIVE_INT_KEY_TOPIC.equals(recordMetadata.topic())))
```
artembilan: I somehow believe that we can live without extra `match()` utility.
There is a `hasFieldOrPropertyWithValue()` and its siblings.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -209,12 +209,6 @@ public boolean isTransactional() {
 	@Override
 	public ListenableFuture<SendResult<K, V>> send(Message<?> message) {
 		ProducerRecord<?, ?> producerRecord = this.messageConverter.fromMessage(message, this.defaultTopic);
-		if (!producerRecord.headers().iterator().hasNext()) { // possibly no Jackson
```
artembilan: I think we can't remove such a logic from here: there is no guarantee that a `MessagingMessageConverter` is always going to be used.

I'm afraid we have to copy/paste such a logic into the `ReactiveKafkaProducerTemplate`.

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
```diff
@@ -0,0 +1,159 @@
...
+	public Flux<Flux<ConsumerRecord<K, V>>> receiveExactlyOnce(TransactionManager transactionManager) {
+		return this.kafkaReceiver.receiveExactlyOnce(transactionManager);
+	}
+
+	@Override
+	public <T> Mono<T> doOnConsumer(Function<Consumer<K, V>, ? extends T> function) {
+		return this.kafkaReceiver.doOnConsumer(function);
+	}
+
+	public Mono<Set<TopicPartition>> assignment() {
```
artembilan: This is somehow wrong from the reactive API perspective: if it is a `Collection` underneath let's consider to use `Flux.fromIterable()` instead of ` Mono<Set<?>>`!

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
```diff
@@ -0,0 +1,159 @@
...
+		this.kafkaReceiver = KafkaReceiver.create(receiverOptions);
+	}
+
+	@Override
+	public Flux<ReceiverRecord<K, V>> receive() {
+		return this.kafkaReceiver.receive();
+	}
+
+	@Override
+	public Flux<Flux<ConsumerRecord<K, V>>> receiveAutoAck() {
```
artembilan: I wonder how this `Flux<Flux<?>>` is useful on practice. Can't we just `flatMap()` it before returning?
Seems like API is claiming an `autoAck`, so, it doesn't bring me too much value to worry about and underneath "hard" logic.

Am I missing anything?

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerOperations.java
```diff
@@ -0,0 +1,52 @@
...
+ * Interface for executing reactive kafka operations on producer side.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Mark Norkin
+ */
+public interface ReactiveKafkaProducerOperations<K, V> {
+
+	<T> Flux<SenderResult<T>> send(Publisher<? extends SenderRecord<K, V, T>> records);
```
artembilan: We probably need do not expose such an API.
I would say a `ProducerRecord` and `org.springframework.kafka.support.SendResult` should be enough from Spring API perspective.

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerOperations.java
```diff
@@ -0,0 +1,52 @@
...
+
+	<T> Flux<Flux<SenderResult<T>>> sendTransactionally(Publisher<? extends Publisher<? extends SenderRecord<K, V, T>>> records);
```
artembilan: I might miss something, but looks like `Flux.concat()` does the trick for us:
```
/**
	 * Concatenate all sources provided as a vararg, forwarding elements emitted by the
	 * sources downstream.
	 * <p>
	 * Concatenation is achieved by sequentially subscribing to the first source then
	 * waiting for it to complete before subscribing to the next, and so on until the
	 * last source completes. Any error interrupts the sequence immediately and is
	 * forwarded downstream.
	 * <p>
	 * <img class="marble" src="https://raw.githubusercontent.com/reactor/reactor-core/v3.1.3.RELEASE/src/docs/marble/concat.png" alt="">
	 * <p>
	 * @param sources The {@link Publisher} of {@link Publisher} to concat
	 * @param <T> The type of values in both source and output sequences
	 *
	 * @return a new {@link Flux} concatenating all source sequences
	 */
	@SafeVarargs
	public static <T> Flux<T> concat(Publisher<? extends T>... sources) {
```

I mean it looks like just `Publisher<? extends ProducerRecord<K, V>>... records` would be enough, if there is a strong requirements to have a bunch of records in a single transaction.

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerOperations.java
```diff
@@ -0,0 +1,52 @@
...
+
+	TransactionManager transactionManager();
+
+	KafkaOutbound<K, V> createOutbound();
```
artembilan: We probably don't need this since we can do the stuff with the `Flux.concat()`

hartmut-co-uk:
note: https://twitter.com/OlehDokuka/status/1021499765593395201

marknorkin:
Hello @artembilan 

No worries and sorry for the delay on my side, was on holidays. I'm happy to go along and resume the work on this PR.
Also thank you for the review, I'll take a look more closely within the next few days.

spring-kafka/src/test/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplateTransactionIntegrationTests.java
marknorkin: Done. Replaced by `extracting+containsExactly` and `allSatisfy`.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
marknorkin: Good point. I'll revert this logic and at the moment will duplicate it in reactive template. On this note should we create internal decorator which will encapsulate provided message converter, this would resolve code duplication for both templates. WDYT ?

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
marknorkin: Done. 
Also did similar change to `ReactiveKafkaProducerTemplate::partitionsFromProducerFor` and for several api's on consumer template side.

Also there are several api's left with Mono over Map, for example there is `ReactiveKafkaProducerTemplate::metricsFromProducer` with return type `Mono<? extends Map<MetricName, ? extends Metric>>`, should we consider changing it to `Flux<? extends Map.Entry<MetricName, ? extends Metric>>` and perform these change for similar api's, WDYT ?

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
marknorkin: Yes, this would simplify the api a bit.
However we have to assume that client is not interest in receiving each batch separately as stated in KafkaReceiver doc 
> Returns a {@link Flux} containing each batch of consumer records returned by {@link Consumer#poll(long)}

Do you know a use case which requires proccessing each autoAck batch separately, if so should we give the client an option then ?

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerOperations.java
marknorkin: it's a bit unclear what client should do if he/she wants matching the response to its corresponding request with correlation metadata. Or your point is that if needed they could check the headers of SendResult.getProducerRecord() ?

Another concern is that SenderResult caries the exception if records sent unsuccessfully, should we extend SendResult with exception then ? 

Also if exposing reactor-kafka types is a concern there is also `public Flux<ReceiverRecord<K, V>> receive()`, however not exposing it would mean there will be no ability to ack the messages manually

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerOperations.java
marknorkin: Done. This would also align with `<T> Flux<SenderResult<T>> send(Publisher<? extends SenderRecord<K, V, T>> records)`

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerOperations.java
marknorkin: Done.

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Reactive kafka consumer operations implementation.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Mark Norkin
+ */
+public class ReactiveKafkaConsumerTemplate<K, V> implements ReactiveKafkaConsumerOperations<K, V> {
```
artembilan: What benefit does this bring over Reactor Kafka?
Also I am a bit reluctant to make stateful operation in the template pattern...
However I see that we may consider it as a `JdbcTemplate`  which is really based on some `DataSource`, but I still don't see too much gain with this template over  standard Reactor Kafka `KafkaReceiver`...

Of course we can give it a shot and see how it makes sense with the target Spring Kafka users.

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerOperations.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @author Mark Norkin
+ */
+public interface ReactiveKafkaProducerOperations<K, V> {
+
+	<T> Flux<SenderResult<T>> send(Publisher<? extends SenderRecord<K, V, T>> records);
+
+	<T> Flux<SenderResult<T>> sendTransactionally(Publisher<? extends SenderRecord<K, V, T>> records);
+
+	TransactionManager transactionManager();
```
artembilan: Why do we need to expose this option if we have that `sendTransactionally()` API?
Does that mean that `Receiver` is fully based on the `TransactionManager` from the sender only?

Thanks

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerOperations.java
artembilan: I see what you mean.
I'm afraid we can't change `SendResult` because it is tied with the `ListenableFuture` returned from the `KafkaTemplate.send()`.

I don't see yet much cleaner solution, so I won't mind to give it a shot and see how it is attractive for end-users.

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplate.java
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	public ReactiveKafkaProducerTemplate(SenderOptions<K, V> senderOptions, RecordMessageConverter messageConverter) {
+		this.sender = KafkaSender.create(senderOptions);
+		this.messageConverter = messageConverter;
+	}
+
+	@Override
+	public <T> Flux<SenderResult<T>> sendTransactionally(Publisher<? extends SenderRecord<K, V, T>> records) {
+		Flux<Flux<SenderResult<T>>> sendTransactionally = this.sender.sendTransactionally(Flux.just(records));
```
artembilan: Don't you mean here `Flux.from(records)`?

build.gradle
```diff
@@ -82,7 +82,8 @@ subprojects { subproject ->
 		springRetryVersion = '1.2.2.RELEASE'
 		springVersion = '5.1.0.RC1'
 		springDataCommonsVersion = '2.1.0.RC1'
-
+        reactorKafkaVersion = '1.0.0.RELEASE'
```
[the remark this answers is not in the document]
hartmut-co-uk: bump `1.0.1.RELEASE` has been released with bugfixes, ... - in case it slipped through above

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplate.java
marknorkin: Well, Flux.from can't be used here as  `sendTransactionally` accepts publisher of publishers e.g. `Publisher<? extends Publisher<? extends SenderRecord<K, V, T>>> records`, but Flux.from would return `Flux<? extends SenderRecord<K, V, T>>` instead of `Flux<? extends Publisher<? extends SenderRecord<K, V, T>>>`

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerOperations.java
marknorkin: I think it could be useful for manual control over transaction, so for example if the user has some custom logic it could use transaction manager directly

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
marknorkin: Sorry, I didn't get the point about stateful operation in the template, can you elaborate please ?

As for usefulness of consumer template. At the moment it's mainly reactor-kafka receiver with a bit more friendlier api with a shortcuts, but I think that incorporating it into the framework also will give users familiar template abstraction.

build.gradle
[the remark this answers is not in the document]
marknorkin: Done

marknorkin:
@artembilan gentle ping for feedback

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerOperations.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Interface for executing reactive kafka operations on consumer side.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Mark Norkin
+ */
+public interface ReactiveKafkaConsumerOperations<K, V> {
```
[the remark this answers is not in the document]
artembilan: Well, that’s not good. When someone has a custom implementation, it is going to be a breaking change pulling new contract methods later when we really can do that now. Or as I said in one comments: do we need this interface at all?

build.gradle
```diff
@@ -83,6 +83,8 @@ subprojects { subproject ->
 		springRetryVersion = '1.2.2.RELEASE'
 		springVersion = '5.1.0.BUILD-SNAPSHOT'
 		springDataCommonsVersion = '2.1.0.BUILD-SNAPSHOT'
+		reactorKafkaVersion = '1.0.1.RELEASE'
+		reactorTestVersion = '3.1.8.RELEASE'
```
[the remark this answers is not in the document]
artembilan: Well, the problem that we have to be aligned with the stack and there is not going to be several Reactor versions in the Spring Boot. That’s why would be better to upgrade this project to Reactor 3.2 to be sure that we are compatible with target platform which is going to consume our 2.2 version. You shouldn’t look into Maven central on the matter: it is in Milestone indeed

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	public Mono<Void> seek(TopicPartition partition, long offset) {
+		return doOnConsumer(consumer -> {
+			consumer.seek(partition, offset);
+			return null;
+		});
+	}
+
+	public Mono<Void> seekToBeginning(Collection<TopicPartition> partitions) {
```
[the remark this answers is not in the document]
artembilan: If it’s not vararg, you require from me to create a collection object to wrap my several partition (even single) to an extra collection object. 

Even if Spring Data Mongo has a collection-based API, that doesn’t mean it is perfect.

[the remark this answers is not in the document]
marknorkin: Done. Changed to var args in all places where Collection of TopicPartition was passed previously as parameter.

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerOperations.java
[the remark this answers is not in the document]
artembilan: The problem that you didn't provide a `default` methods, but at the same time there are a lot of `public` API on the template impl.

I'm fully fine to live without interfaces for now. We can extract them later if necessary...

build.gradle
```diff
@@ -83,6 +83,9 @@ subprojects { subproject ->
 		springRetryVersion = '1.2.2.RELEASE'
 		springVersion = '5.1.0.BUILD-SNAPSHOT'
 		springDataCommonsVersion = '2.1.0.BUILD-SNAPSHOT'
+		reactorCoreVersion = '3.2.0.M4'
+		reactorKafkaVersion = '1.0.1.RELEASE'
+		reactorTestVersion = '3.2.0.M4'
```
artembilan: This is redundant property. We are fully fine just use a single `reactorVersion = '3.2.0.M4'`

spring-kafka/src/test/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplateTransactionIntegrationTests.java
```diff
@@ -0,0 +1,216 @@
...
+	private Flux<SenderRecord<Integer, String, Integer>> generateSenderRecords(int recordsCount, int seed) {
+		return Flux.range(1, recordsCount)
+				.map(i -> {
+					int correlationMetadata = i * (int) Math.pow(10, seed);
+					return SenderRecord.create(new ProducerRecord<>(REACTIVE_INT_KEY_TOPIC, DEFAULT_PARTITION, DEFAULT_TIMESTAMP, DEFAULT_KEY, DEFAULT_VALUE + correlationMetadata), correlationMetadata);
+				});
+	}
+
+	@Test//todo
+	@Ignore
```
artembilan: Any chances to Un`@Ignore` ?

Thanks

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
```diff
@@ -0,0 +1,158 @@
...
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Mark Norkin
+ */
+public class ReactiveKafkaConsumerTemplate<K, V> {
+	private final KafkaReceiver<K, V> kafkaReceiver;
+
+	public ReactiveKafkaConsumerTemplate(ReceiverOptions<K, V> receiverOptions) {
+		this.kafkaReceiver = KafkaReceiver.create(receiverOptions);
```
artembilan: I didn't get an answer about thread safety in this component: https://github.com/spring-projects/spring-kafka/pull/559#discussion_r210743133

Thanks

build.gradle
marknorkin: Done

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
marknorkin: Yes, now see what you mean. Looking at the code of KafkaReceiver I see that it operates on a single kafka consumer, which is not thread safe and having the logic like
```
        if (!isActive.compareAndSet(false, true))
            throw new IllegalStateException("Multiple subscribers are not supported for KafkaReceiver flux");
``` 
probably gives the answer that it's shouldn't be used in that way, so the pattern as you prescribed would be to subscribe once to the flow and then layer other processing/sinking logic similar to what apache spark and other stream processing frameworks do.

Having template as a `factory` of receivers (if I understand you correctly) probably would be confusing, as each receiver bound to one kafka consumer, so creating a consumer each time would not work as expected for users.

As templates are based on KafkaReceiver/KafkaSender, I think it's better to align with reactor-kafka design for them.

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
marknorkin: Yes, sorry overlooked it, thank you.

artembilan:
@marknorkin ,

Thanks for heads up!

I think it is going to be merged when we switch `master` to the version `2.3.0`.

I'll ping you one more time for rebase then!

Feel free to provide more features meanwhile or just raise issues which should be addressed as additions for this functionality afterwards!

marknorkin:
@artembilan 
Cool, thanks. 
Sure, ping me if needed.

hartmut-co-uk:
note latest reactor kafka release: `reactorKafkaVersion = '1.1.0.RELEASE'`

hartmut-co-uk:
Hi, don't want to hijack the PR - but I was just playing/testing the current state and came across literally the same issue as mentioned here: https://github.com/reactor/reactor-kafka/issues/62

So essentially being able to setup a `@KafkaListener` ~> maybe a `@ReactiveKafkaListener`?
Which would be a permanent running Consumer which should be able to handle repartitioning, connection drops, ..

Where would be a good starting point or what interfaces would need to be implemented to use the existing non-reactive code that's there @artembilan?

artembilan:
That's interesting point, @hartmut-co-uk , but I suppose we need to think about that already after merging this PR.

hartmut-co-uk:
@artembilan any update in regards to merging this PR + follow-up TODOs / task planning for sth like a `@ReactiveKafkaListener`?

artembilan:
@hartmut-co-uk ,

We are going to come back to this PR in a new year already. When we will start `2.2` version.

Thanks for heads up anyway!

garyrussell:
>When we will start 2.2 version.

That's 2.3.

build.gradle
```diff
@@ -83,6 +83,8 @@ subprojects { subproject ->
 		springRetryVersion = '1.2.2.RELEASE'
 		springVersion = '5.1.0.RC2'
 		springDataCommonsVersion = '2.1.0.RC2'
+		reactorVersion = '3.2.0.M4'
+		reactorKafkaVersion = '1.0.1.RELEASE'
```
artembilan: Let's consider to update these dependencies to see if the API is better now!

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
artembilan: `Assert.notNull(receiverOptions)`

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
```diff
@@ -0,0 +1,158 @@
...
+
+	public Flux<ConsumerRecord<K, V>> receiveAutoAck() {
+		return this.kafkaReceiver.receiveAutoAck().flatMap(Function.identity());
+	}
+
+	public Flux<ConsumerRecord<K, V>> receiveAtmostOnce() {
+		return this.kafkaReceiver.receiveAtmostOnce();
+	}
+
+	public Flux<Flux<ConsumerRecord<K, V>>> receiveExactlyOnce(TransactionManager transactionManager) {
```
artembilan: I don't remember if we discussed this, but why `Flux` of `Flux`es?
Can't we flatten them? 
Or if that is some Reactor Kafka specific type for records per partition, then let's return that one maybe!

Just explain me one more time "Why?"

Thanks

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	public Mono<Void> resume(TopicPartition... partitions) {
+		return doOnConsumer(c -> {
+			c.resume(Arrays.asList(partitions));
+			return null;
+		});
+	}
+
+	public Flux<Map.Entry<MetricName, ? extends Metric>> metricsFromConsumer() {
```
artembilan: I would for the `Tuple` from Reactor is better, than `Map.Entry` to have here.

build.gradle
marknorkin: Done

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
marknorkin: Done

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
marknorkin: Done. Also changed for other api's on producer side that return Map.Entry.

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
marknorkin: Well, looking at the documentation of KafkaReceiver, particularly:

> Returns a {@link Flux} of consumer record batches that may be used for exactly once delivery semantics. 
> ...
> A new transaction is started for each inner Flux and it is the responsibility of the consuming application to commit or abort the transaction
> ...
> The next batch of consumer records will be delivered only after the previous flux terminates.

My understanding of it is that we can't flatten them.
It results in having one flattened Flux which will never terminate and so is not clear where transaction demarcation happens/should happen.

marknorkin: Done

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/9763e637cd338dc2c7dcf6c31d724b19913ad1e7 after some minor polishing.

Thank you very much, @marknorkin , for contribution; looking forward for more!

for example outstanding task is there about documenting this new feature 😄

marknorkin:
Hello, @artembilan. Glad to help with contribution. Will try to add document the feature if time permits.

hartmut-co-uk:
👏 💪

hartmut-co-uk:
Any chance we could address https://github.com/spring-projects/spring-kafka/pull/559#issuecomment-439868452 as well, in scope for spring-kafka 2.3 / spring-boot 2.2?

I think that would be key to actually make use of the consumer.

artembilan:
@hartmut-co-uk 

> Any chance we could address


Please, raise a fresh GH issue with much more details.
Contribution is always welcome though!

dnitzan:
As part of this pull request, is there a also a plan to implement a fully reactive interface specifically for the functionality provided by ReplyingKafkaTemplate? This would be super helpful for reactive web services that interact with Kafka in a request/response manner.

artembilan:
This PR is already closed being merge before.

What you are asking is already fully new feature.
Please, consider to raise an new issue with enough information to understand and weight a feature request.

PR #560, opened 2018-02-06
More Spring Integration Kafka Support

garyrussell:
- don't map message channel headers, unless they are String
- tolerate `byte[]` in topic header.


artembilan:
@garyrussell ,

Does it say anything to you:
```
org.springframework.kafka.core.KafkaTemplateTests > testTemplate FAILED
    java.lang.NullPointerException at KafkaTemplateTests.java:133
```
?

See Travis report.

Thanks

PR #564, opened 2018-02-10
rename constant DEFAULT_VALUE_TYPE to keep the same with DEFAULT_KEY_TYPE

qct:
because DEFAULT_KEY_TYPE is "spring.json.key.default.type",  but DEFAULT_VALUE_TYPE is "spring.json.default.value.type", seems like wrong word order, we should keep them the same.

from
```
public static final String DEFAULT_KEY_TYPE = "spring.json.key.default.type";
public static final String DEFAULT_VALUE_TYPE = "spring.json.default.value.type";
```
to 
```
public static final String DEFAULT_KEY_TYPE = "spring.json.key.default.type";
public static final String DEFAULT_VALUE_TYPE = "spring.json.value.default.type";
```

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
@@ -54,7 +54,6 @@ public class JsonDeserializer<T> implements ExtendedDeserializer<T> {
 
 	/**
 	 * Kafka config property for the default key type if no header.
-	 *
 	 * @deprecated in favor of {@link #KEY_DEFAULT_TYPE}
 	 */
 	@Deprecated
@@ -62,8 +61,9 @@ public class JsonDeserializer<T> implements ExtendedDeserializer<T> {
 
 	/**
 	 * Kafka config property for the default value type if no header.
-	 * <p>Will be removed in the future, use {@link #VALUE_DEFAULT_TYPE} instead.
+	 * @deprecated in favor of {@link #VALUE_DEFAULT_TYPE}
 	 */
+	@Deprecated
 	public static final String DEFAULT_VALUE_TYPE = "spring.json.default.value.type";
 
 	/**
@@ -162,15 +162,16 @@ public class JsonDeserializer<T> implements ExtendedDeserializer<T> {
 				}
 			}
 			// TODO don't forget to remove these code after DEFAULT_VALUE_TYPE being removed.
-			else if (!isKey && configs.containsKey(DEFAULT_VALUE_TYPE)) {
-				if (configs.get(DEFAULT_VALUE_TYPE) instanceof Class) {
-					this.targetType = (Class<T>) configs.get(DEFAULT_VALUE_TYPE);
+			else if (!isKey && configs.containsKey("spring.json.default.value.type")) {
+				if (configs.get("spring.json.default.value.type") instanceof Class) {
+					this.targetType = (Class<T>) configs.get("spring.json.default.value.type");
 				}
-				else if (configs.get(DEFAULT_VALUE_TYPE) instanceof String) {
-					this.targetType = (Class<T>) ClassUtils.forName((String) configs.get(DEFAULT_VALUE_TYPE), null);
+				else if (configs.get("spring.json.default.value.type") instanceof String) {
+					this.targetType = (Class<T>) ClassUtils
+						.forName((String) configs.get("spring.json.default.value.type"), null);
 				}
 				else {
-					throw new IllegalStateException(DEFAULT_VALUE_TYPE + " must be Class or String");
+					throw new IllegalStateException("spring.json.default.value.type must be Class or String");
 				}
 			}
 			else if (!isKey && configs.containsKey(VALUE_DEFAULT_TYPE)) {
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
@@ -53,15 +53,17 @@ import com.fasterxml.jackson.databind.ObjectReader;
 public class JsonDeserializer<T> implements ExtendedDeserializer<T> {
 
 	/**
+	 * Kafka config property for the default key type if no header.
+	 *
 	 * @deprecated in favor of {@link #KEY_DEFAULT_TYPE}
 	 */
 	@Deprecated
 	public static final String DEFAULT_KEY_TYPE = "spring.json.key.default.type";
 
 	/**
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@qct Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=564)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=564) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@qct Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=564)!

artembilan:
This is good catch, indeed. And really from the consistency perspective it is confusing a bit.
We really accept this change, but only concern that this is a public constant and might be used from the external configurations. Therefore we can merge such a breaking change only in the next point release.
Or even much stronger: to follow the backward compatibility policy we can do that only in the next `3.0` major version.

Meanwhile we definitely should introduce a new constant and support both of them some time. The current one should be `@Deprecated` to notify end-users to switch to a new, correct value as earlier as possible.

@garyrussell ,
WDYT ?

Thanks

garyrussell:
Good catch but, unfortunately, we can't change this easily without breaking compatibility; it will break applications for people doing rolling upgrades.

We'll have to do it in stages - in 2.1.x handle both values (error if both are present). 3.0.x change to the new value.

garyrussell:
Hehe - we answered at the same time 😄

artembilan:
@qct ,

any news on the matter?
We are going to release today, but we can't merge your fix as is - need more work. See our comments above.

@garyrussell ,

We definitely need to come back to this issue some day if not today at all before releasing the current `2.1.3`. I can jump into this and PR something in the afternoon though...

Thanks

garyrussell:
@artembilan I don't consider this critical; while it's unfortunate we missed it, it's just cosmetic, and not an issue at all if apps use the constant.

qct:
@artembilan I agree with you and @garyrussell , because it's a breaking change, we should consider compatibility, I will change pr to handle both value and mark old one `@Deprecated`, so it can be merged in the next release.

qct:
hi, guys, I have marked the old pair of `DEFAULT_KEY_TYPE` and `DEFAULT_VALUE_TYPE` deprecated, introduced a new pair:  `KEY_DEFAULT_TYPE` and `VALUE_DEFAULT_TYPE`,  this change will keep the compatibility, so we can publish to users next patch version and totally change it next major version.

artembilan:
Checkstyle violation:
```
:spring-kafka:checkstyleMain[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java:55: Javadoc has empty description section. [JavadocStyle]
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java:61: Javadoc has empty description section. [JavadocStyle]
```

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -136,15 +148,15 @@ public void setUseTypeMapperForKey(boolean isKey) {
...
-					this.targetType = (Class<T>) ClassUtils.forName((String) configs.get(DEFAULT_KEY_TYPE), null);
+				else if (configs.get(KEY_DEFAULT_TYPE) instanceof String) {
+					this.targetType = (Class<T>) ClassUtils.forName((String) configs.get(KEY_DEFAULT_TYPE), null);
 				}
 				else {
-					throw new IllegalStateException(DEFAULT_KEY_TYPE + " must be Class or String");
+					throw new IllegalStateException(KEY_DEFAULT_TYPE + " must be Class or String");
 				}
 			}
 			else if (!isKey && configs.containsKey(DEFAULT_VALUE_TYPE)) {
```
artembilan: This causes a deprecation warning.

Also there is a `KafkaMessageListenerContainerTests` which uses constant as well with similar deprecation warning. Should not use it any more.

Please, fix these concerns, run `gradlew clean check`. 

And also, please, take a look into the Reference Manual (`.adoc` files) if we use these constant there. And, therefore, fix the doc on the matter.

Otherwise looks good.

:+1:

qct: keep the old DEFAULT_VALUE_TYPE here because of the compatibility, I think we can't remove it in next patch version, but can be removed in next major version, how about adding a `@SuppressWarnings("deprecation")` on the method?

artembilan: Right, we will remove that code in the next major (or maybe even minor) version. However I’m not a fan of `@SuppressWarnings("deprecation")`. Would be better to use literals here instead of deprecated constants. However we will need TODO comment here do not forget to remove the obsolete code

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -53,15 +53,17 @@
 public class JsonDeserializer<T> implements ExtendedDeserializer<T> {
 
 	/**
+	 * Kafka config property for the default key type if no header.
+	 *
```
artembilan: No blank lines in the method or properties JavaDocs, please

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -53,15 +53,17 @@
...
 	 */
 	@Deprecated
 	public static final String DEFAULT_KEY_TYPE = "spring.json.key.default.type";
 
 	/**
-	 * @deprecated in favor of {@link #VALUE_DEFAULT_TYPE}
+	 * Kafka config property for the default value type if no header.
+	 * <p>Will be removed in the future, use {@link #VALUE_DEFAULT_TYPE} instead.
 	 */
-	@Deprecated
```
artembilan: No, you didn't understand me. 
We definitely have to deprecate it to encourage end-users to migrate to a new value.
My point was to use `"spring.json.default.value.type"` literal instead of this constant in the `else if` below.

artembilan:
@qct ,

thank you very much for the contribution.
Looking forward for more!

qct:
thank you, it's my honor.

Issue #561, opened 2018-02-07
Cannot use JsonSerde for both key and value serialization in a stream


elliotkennedy:
Am I correct in thinking I should be able to use a ```JsonSerde``` for both the key and value in a stream?

Deserialization fails when using ```JsonSerde``` to serialize both the key and value of a topic.

The key and value are both given the same header value here:

https://github.com/spring-projects/spring-kafka/blob/31431fac4661dc9db9e93bb6ad14f7e926c3c8b4/spring-kafka/src/main/java/org/springframework/kafka/support/converter/DefaultJackson2JavaTypeMapper.java#L177

So when deserializing, the key and value header chosen is simply the first header in the list. This results in attempting to deserialize the value as the key type.

https://github.com/spring-projects/spring-kafka/blob/31431fac4661dc9db9e93bb6ad14f7e926c3c8b4/spring-kafka/src/main/java/org/springframework/kafka/support/converter/AbstractJavaTypeMapper.java#L116

I've created a test to demonstrate the issue here:

https://github.com/elliotkennedy/spring-kafka/commit/ff157158d12d06102ef988765c16c87b4ef6b10b

I might be able to attempt a fix using different headers soon if people think that would be appropriate?

artembilan:
That's definitely good catch, indeed!

I think as a workaround we have to use an extension of the `DefaultJackson2JavaTypeMapper` and store `key` or `value` under different header name overriding the:
```
public String getClassIdFieldName() {
	return DEFAULT_CLASSID_FIELD_NAME;
}
```

To fix the issue I guess we need to take into account `isKey` argument in the:

```
    /**
     * Configure this class.
     * @param configs configs in key/value pairs
     * @param isKey whether is for key or value
     */
    void configure(Map<String, ?> configs, boolean isKey);
```

And configure different `classIdFieldName`.

In addition that looks like we have to require `isKey` as a ctor argument for the cases like:
```
Produced.with(Serdes.Integer(), new JsonSerde<Foo>() {})
```

Feel free to go ahead with the Contribution.

Thank you for the report!

elliotkennedy:
Thanks @artembilan, I've attempted to configure the Serdes - please let me know what you think.

The problem is that when you run the tests now, ```void configure(Map<String, ?> configs, boolean isKey);``` is never called on the JsonDeserializer. This feels like a bug in Kafka itself - I'd expect the topology builder to configure the serde at some point. What do you think?

[...]

PR #567, opened 2018-02-13
#561 -  Cannot use JsonSerde for both key and value serialization in a stream

elliotkennedy:
GH-561 Add a test to json serialize both a key and value in a stream.

GH-561 - Configure Jackson2JavaKeyTypeMapper in JsonDeserializer and JsonSerializer when the isKey flag is present.

GH-561 - Add collection json serialization tests.

GH-561 - Use isKey to set json headers.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/converter/AbstractJavaTypeMapper.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/converter/AbstractJavaTypeMapper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2017-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -37,6 +37,7 @@ import org.springframework.util.ClassUtils;
  * @author Sam Nelson
  * @author Andreas Asplund
  * @author Gary Russell
+ * @author Elliot Kennedy
  *
  * @since 2.1
  */
@@ -57,22 +58,70 @@ public abstract class AbstractJavaTypeMapper implements BeanClassLoaderAware {
 	 */
 	public static final String DEFAULT_KEY_CLASSID_FIELD_NAME = "__KeyTypeId__";
 
+	/**
+	 * Default header name for key type information.
+	 */
+	public static final String KEY_DEFAULT_CLASSID_FIELD_NAME = "__Key_TypeId__";
+
+	/**
+	 * Default header name for key container object contents type information.
+	 */
+	public static final String KEY_DEFAULT_CONTENT_CLASSID_FIELD_NAME = "__Key_ContentTypeId__";
+
+	/**
+	 * Default header name for key map key type information.
+	 */
+	public static final String KEY_DEFAULT_KEY_CLASSID_FIELD_NAME = "__Key_KeyTypeId__";
+
 	private final Map<String, Class<?>> idClassMapping = new HashMap<String, Class<?>>();
 
 	private final Map<Class<?>, byte[]> classIdMapping = new HashMap<Class<?>, byte[]>();
 
+	private String classIdFieldName = DEFAULT_CLASSID_FIELD_NAME;
+
+	private String contentClassIdFieldName = DEFAULT_CONTENT_CLASSID_FIELD_NAME;
+
+	private String keyClassIdFieldName = DEFAULT_KEY_CLASSID_FIELD_NAME;
+
 	private ClassLoader classLoader = ClassUtils.getDefaultClassLoader();
 
 	public String getClassIdFieldName() {
-		return DEFAULT_CLASSID_FIELD_NAME;
+		return this.classIdFieldName;
+	}
+
+	/**
+	 * Configure header name for type information.
+	 * @param classIdFieldName the header name.
+	 * @since 2.1.3
+	 */
+	public void setClassIdFieldName(String classIdFieldName) {
+		this.classIdFieldName = classIdFieldName;
 	}
 
 	public String getContentClassIdFieldName() {
-		return DEFAULT_CONTENT_CLASSID_FIELD_NAME;
+		return this.contentClassIdFieldName;
+	}
+
+	/**
+	 * Configure header name for container object contents type information.
+	 * @param contentClassIdFieldName the header name.
+	 * @since 2.1.3
+	 */
+	public void setContentClassIdFieldName(String contentClassIdFieldName) {
+		this.contentClassIdFieldName = contentClassIdFieldName;
 	}
 
 	public String getKeyClassIdFieldName() {
-		return DEFAULT_KEY_CLASSID_FIELD_NAME;
+		return this.keyClassIdFieldName;
+	}
+
+	/**
+	 * Configure header name for map key type information.
+	 * @param keyClassIdFieldName the header name.
+	 * @since 2.1.3
+	 */
+	public void setKeyClassIdFieldName(String keyClassIdFieldName) {
+		this.keyClassIdFieldName = keyClassIdFieldName;
 	}
 
 	public void setIdClassMapping(Map<String, Class<?>> idClassMapping) {
@@ -131,6 +180,19 @@ public abstract class AbstractJavaTypeMapper implements BeanClassLoaderAware {
 
 	public Map<String, Class<?>> getIdClassMapping() {
 		return Collections.unmodifiableMap(this.idClassMapping);
+	}
+
+	/**
+	 * Configure the TypeMapper to use default key type class.
+	 * @param isKey Use key type headers if true
+	 * @since 2.1.3
+	 */
+	public void setUseForKey(boolean isKey) {
+		if (isKey) {
+			setClassIdFieldName(AbstractJavaTypeMapper.KEY_DEFAULT_CLASSID_FIELD_NAME);
+			setContentClassIdFieldName(AbstractJavaTypeMapper.KEY_DEFAULT_CONTENT_CLASSID_FIELD_NAME);
+			setKeyClassIdFieldName(AbstractJavaTypeMapper.KEY_DEFAULT_KEY_CLASSID_FIELD_NAME);
+		}
 	}
 
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2017 the original author or authors.
+ * Copyright 2015-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/converter/Jackson2JavaKeyTypeMapper.java
```diff
@@ -0,0 +1,57 @@
...
+package org.springframework.kafka.support.converter;
+
+/**
+ * Jackson 2 type mapper for record keys.
+ *
+ * @author Elliot Kennedy
+ *
+ * @since 2.1
+ */
+public class Jackson2JavaKeyTypeMapper extends DefaultJackson2JavaTypeMapper {
```
artembilan: Well, actually that's not what we have talked about.

The idea is like make the types properties *configurable* in the `DefaultJackson2JavaTypeMapper`.

Introducing this new class you are mixing concerns and say explicitly what it is about and at the same time narrow its functionality.

So, my point is to provide setters for those field names.

This way we still have the same `DefaultJackson2JavaTypeMapper`, but now it is more flexible.

Does it make sense to you?

elliotkennedy: Thanks @artembilan - I'm still not sure about the isKey constructor argument. Should it configure the headers as I have done or is that overkill?

Please take a look

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -112,9 +115,22 @@ public void setTypeMapper(Jackson2JavaTypeMapper typeMapper) {
 		this.typeMapper = typeMapper;
 	}
 
+	/**
+	 * Configure the JsonDeserializer to use key type headers.
+	 * @param isKey Use key type headers if true
+	 */
+	public void configureTypeMapper(boolean isKey) {
```
artembilan: How about instead of those new ctors and extra arg there just rename this to the setter - `setUseTypeMapperForKey(boolean)`  and encourage end-users to call it for the particular reason in the class JavaDocs?

And right: still call it from the `configure()`.

How is this approach with you?

We will have a new functionality and won't have any breaking changes.

I'm open form any other suggestions.
But I'm against the breaking change to the `JsonDeserializer(Class<T> targetType, ObjectMapper objectMapper)` ctor. We can't afford it in the current point release.

elliotkennedy: That sounds better, thanks - do you think the ```JsonSerde``` constructor should still have an ```isKey``` param, in addition to the original ctors?

In order to configure streams like this:
```
.stream(OBJECT_INPUT_TOPIC, Consumed.with(
new JsonSerde<>(JsonObjectKey.class, true), 
new JsonSerde<>(JsonObjectValue.class)));
```
So I think the jsonSerde ctor is ok?

It would be great is spring just intercepted the serde and didn't need a constructor arg though ;)

artembilan: 😄 OK. Agreed. Only the concern to keep backward compatibility.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/Jackson2JavaTypeMapper.java
```diff
@@ -50,4 +51,22 @@
 
 	void addTrustedPackages(String... packages);
 
+	/**
+	 * Configure header name for type information.
+	 * @param classIdFieldName the header name.
+	 */
+	void setClassIdFieldName(String classIdFieldName);
```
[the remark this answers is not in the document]
elliotkennedy: Cool, I've made the default methods no-op, hope that's ok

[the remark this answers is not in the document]
artembilan: Well, actually I have though one more time yesterday, but forgot to comment in the morning: I don't think now that it is good idea to make setters as a contract. That is indeed a part of the target implementation - `AbstractJavaTypeMapper`.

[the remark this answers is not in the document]
elliotkennedy: Agreed I didn't like adding those to the interface - but how should I configure the ```Jackson2JavaTypeMapper``` in ```setUseTypeMapperForKey``` then? Instance of and then cast feels smelly - create a new instance and configure it?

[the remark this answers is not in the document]
artembilan: No, casting is good (with checking type). Also I'd prefer to have flag like `typeMapperExplicitlySet`  and don't reconfigure it if it is an external one.

This is really just only our `Jackson2JavaTypeMapper` implementation responsibility. End-user can provide any other exotic solution and we shouldn't not interfere either way.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -110,11 +114,28 @@ public Jackson2JavaTypeMapper getTypeMapper() {
...
 		Assert.notNull(typeMapper, "'typeMapper' cannot be null");
 		this.typeMapper = typeMapper;
+		this.typeMapperExplicitlySet = true;
+	}
+
+	/**
+	 * Configure the JsonDeserializer to use key type headers.
+	 * @param isKey Use key type headers if true
+	 */
+	public void setUseTypeMapperForKey(boolean isKey) {
```
elliotkennedy: Do you think this setting should be moved to ```AbstractJavaTypeMapper```? I'm repeating myself in ```JsonSerializer``` & ```JsonSerializer```

artembilan: Right, sounds good.
Delegating is the best choice here do not repeate 😄 

Although maybe with a slightly different name - `setUseForKey(boolean)`. Since the class is `TypeMapper` per se already.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/AbstractJavaTypeMapper.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2018 the original author or authors.
```
artembilan: Well, you know we have to supply here a range of years with the current in the end.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/AbstractJavaTypeMapper.java
```diff
@@ -57,22 +58,55 @@
...
+	public void setContentClassIdFieldName(String contentClassIdFieldName) {
+		this.contentClassIdFieldName = contentClassIdFieldName;
 	}
 
 	public String getKeyClassIdFieldName() {
-		return DEFAULT_KEY_CLASSID_FIELD_NAME;
+		return this.keyClassIdFieldName;
+	}
+
+	public void setKeyClassIdFieldName(String keyClassIdFieldName) {
```
artembilan: Yeah... Any new `public` API must go with JavaDocs and `@since`

spring-kafka/src/main/java/org/springframework/kafka/support/converter/Jackson2JavaTypeMapper.java
```diff
@@ -29,6 +29,7 @@
  * @author Sam Nelson
  * @author Andreas Asplund
  * @author Gary Russell
+ * @author Elliot Kennedy
```
artembilan: There is no any changes in this class, so, that is already speculation to add your name for nothing 😄

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
artembilan: Need `@since 2.1.3` in JavaDocs, too.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerde.java
```diff
@@ -49,29 +50,43 @@
...
+		this(null, objectMapper, false);
+	}
+
+	public JsonSerde(ObjectMapper objectMapper, boolean isKey) {
+		this(null, objectMapper, isKey);
 	}
 
 	@SuppressWarnings("unchecked")
-	public JsonSerde(Class<T> targetType, ObjectMapper objectMapper) {
+	public JsonSerde(Class<T> targetType, ObjectMapper objectMapper, boolean isKey) {
```
artembilan: I think we are loosing here ctor for backward compatibility.

You know, having such a complexity I'd prefer to come to the setter idea. The setter can be like this:
```
public JsonSerde setUseTypeMapperForKey(boolean)
```
Therefore we still can use it as a fluent API for a sample you've shown before.

spring-kafka/src/test/java/org/springframework/kafka/kstream/KafkaStreamsJsonSerializationTests.java
```diff
@@ -0,0 +1,210 @@
+/*
+ * Copyright 2017-2018 the original author or authors.
```
artembilan: And this one is new, therefore should be only with the current year.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerde.java
elliotkennedy: Ok sounds good!

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerializer.java
```diff
@@ -135,4 +155,11 @@ public void close() {
 		// No-op
 	}
 
+	private boolean isTypeMapperExplicitlySet() {
+		return this.typeMapperExplicitlySet;
+	}
+
+	private void setTypeMapperExplicitlySet(boolean typeMapperExplicitlySet) {
```
artembilan: Sorry, not clear why do we need these two method?
You didn't introduce them in the `JsonDeserializer`, why here then?
Thanks

elliotkennedy: Sorry, my mistake - removed

artembilan:
@garyrussell ,

please, take a look here. I wonder if it is related somehow to your resolution in the #578 

Thanks

garyrussell:
Looks OK to me, I imagine most use cases use simple types in keys that can be handled by standard SerDes, but it won't hurt to provide more complex types this way; I don't think it's worth adding logic to the message converters, though.

spring-kafka/src/test/java/org/springframework/kafka/kstream/KafkaStreamsJsonSerializationTests.java
izeye: In regards to copyright years this project looks to have a different policy with Spring Boot. I’m just curious which one is recommended legally.

See https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MeterRegistryPostProcessor.java

garyrussell: I can't speak for boot's policy, but we were advised by legal that the copyright should initially just have the year the file was created.

izeye: @garyrussell Thanks for the quick response!

That's weird as Spring Boot looks to use the year the project has been created as the starting year and I thought the members in the two projects are employed by the same company, hence having the same legal team.

/cc @philwebb

PR #584, opened 2018-02-21
SIKGH-198: Add pause/resume to Listener Containers

garyrussell:
See: https://github.com/spring-projects/spring-integration-kafka/issues/198

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -391,6 +391,8 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 
 		private boolean taskSchedulerExplicitlySet;
 
+		private boolean consumerPaused;
+
 		@SuppressWarnings("unchecked")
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -391,6 +391,8 @@ public String toString() {
 
 		private boolean taskSchedulerExplicitlySet;
 
+		private boolean consumerPaused;
```
artembilan: I think this should be `volatile`. Even if that looks like we manipulate with this variable only in the `run()`, but when we call `stop()/start()` on the container, we lose that thread context and may run in some other one.

garyrussell: It's not necessary; this is only ever accessed in `run()` by one thread (the consumer thread) - so there can't be any problems with cached values. `AMLC.paused` is volatile, though (and has to be).

artembilan: Yeah... Agreed. Now I see this in the `doStart()`:
```
this.listenerConsumer = new ListenerConsumer(this.listener, listenerType);
```

So, everything is perfect - merging...

Issue #599, opened 2018-03-05
Initial Seek when group management is not used is broken


garyrussell:
There is no logic to perform the initial seeks when a list of `TopicPartitionInitialOffset` is provided with `SeekPosition` (works with a numeric initial offset).

PR #600, opened 2018-03-06
GH-599: Fix initial seek

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/599

Previously, initial seeks using `TopicPartitionInitialOffset`s only worked with
a provided `offset`. The `SeekPosition` field was ignored, and only used for
subsequent seek operations.

`initPartitionsIfNeeded()` now processes both styles of initial offset.


garyrussell:
Thanks; I don't think so; the `seek(0)` and `seek(Long.MAXVALUE)` is a workaround.

See https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L451

PR #605, opened 2018-03-19
Polish

izeye:
This PR fixes some typos and polishes trivial stuff.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
@@ -121,18 +121,16 @@ public class DefaultKafkaConsumerFactory<K, V> implements ConsumerFactory<K, V> 
 		if (groupId == null && !shouldModifyClientId) {
 			return createKafkaConsumer();
 		}
-		else {
-			Map<String, Object> modifiedConfigs = new HashMap<>(this.configs);
-			if (groupId != null) {
-				modifiedConfigs.put(ConsumerConfig.GROUP_ID_CONFIG, groupId);
-			}
-			if (shouldModifyClientId) {
-				modifiedConfigs.put(ConsumerConfig.CLIENT_ID_CONFIG,
-					(overrideClientIdPrefix ? clientIdPrefix
-							: modifiedConfigs.get(ConsumerConfig.CLIENT_ID_CONFIG)) + clientIdSuffix);
-			}
-			return createKafkaConsumer(modifiedConfigs);
+		Map<String, Object> modifiedConfigs = new HashMap<>(this.configs);
+		if (groupId != null) {
+			modifiedConfigs.put(ConsumerConfig.GROUP_ID_CONFIG, groupId);
 		}
+		if (shouldModifyClientId) {
+			modifiedConfigs.put(ConsumerConfig.CLIENT_ID_CONFIG,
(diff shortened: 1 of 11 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
```diff
@@ -121,18 +121,16 @@ public void setValueDeserializer(Deserializer<V> valueDeserializer) {
...
+		Map<String, Object> modifiedConfigs = new HashMap<>(this.configs);
+		if (groupId != null) {
+			modifiedConfigs.put(ConsumerConfig.GROUP_ID_CONFIG, groupId);
 		}
+		if (shouldModifyClientId) {
+			modifiedConfigs.put(ConsumerConfig.CLIENT_ID_CONFIG,
+				(overrideClientIdPrefix ? clientIdPrefix
+						: modifiedConfigs.get(ConsumerConfig.CLIENT_ID_CONFIG)) + clientIdSuffix);
+		}
+		return createKafkaConsumer(modifiedConfigs);
```
garyrussell: What is the purpose of this change - seems like just a style preference.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -99,14 +101,9 @@ public void setConcurrency(int concurrency) {
...
-		this.containers.forEach(c -> {
-			Collection<TopicPartition> assignedPartitions = c.getAssignedPartitions();
-			if (assignedPartitions != null) {
-				assigned.addAll(assignedPartitions);
-			}
-		});
-		return assigned;
+		return this.containers.stream()
+				.map(KafkaMessageListenerContainer::getAssignedPartitions).filter(Objects::nonNull)
+				.flatMap(assignedPartitions -> assignedPartitions.stream()).collect(Collectors.toList());
```
garyrussell: we prefer
    
    .oneOperation()
    .perLine()
    .style()

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -192,16 +192,11 @@ public void setClientIdSuffix(String clientIdSuffix) {
...
 			}
-			else {
-				return null;
-			}
-		}
-		else {
-			return null;
 		}
+		return null;
 	}
```
garyrussell: Again; style only.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
izeye: @garyrussell I think `if` block with `return` statement makes the following `else` block redundant and removing it reduces a block depth. WDYT?

artembilan: I can't find the rule now, but we really prefer `if...else` style even if there is a `return` operator.

izeye:
@garyrussell @artembilan Thanks for the feedback!

The requested changes have been applied in b9b54f7.

garyrussell:
Sounds good; thanks - and thanks to @izeye

Issue #566, opened 2018-02-12
KafkaMLC - make addOffsets() thread-safe


garyrussell:
https://stackoverflow.com/questions/47652651/spring-kafka-is-acknowledgement-acknowledge-thread-safe/47663459#comment84501434_47663459

lhsm:
Hi, seems still not thread safe. Please, take a look into KafkaMessageListenerContainer.ListenerConsumer#buildCommits. While iterating offsets new offset may be added
The same in KafkaMessageListenerContainer.ListenerConsumer#commitPendingAcks

garyrussell:
Good catch; thanks.

PR #619, opened 2018-03-22
GH-566: Ack Concurrency Issue

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/566

If `Acknowledgment.acknowledge()` is called on a foreign thread, there is a concurrency
problem with the `offsets` field; the consumer thread might `clear()` unprocessed acks.

Furthermore, `AckMode.MANUAL_IMMEDIATE` cannot use the `Consumer` object if the ack
is called on a foreign thread - the `Consumer` is not thread-safe.

- Revert `offsets` to simple `HashMap`s
- Only reference `offsets` on the consumer thread
- enqueue foreign acks into the `acks` queue (even "immediate" acks)

**Cherry pick to 2.0.x (and 1.3.x, fixing the lambda in the test and the diamond operators).**


artembilan:
... and cherry-picked to `2.0.x`.
Back-ported to `1.3.x` as: https://github.com/spring-projects/spring-kafka/commit/5e877a8003ef7fc39ff3e1390def1e669ae32273

Issue #618, opened 2018-03-22
Expose application.id / StreamsConfig from StreamsBuilderFactoryBean


zacholauson:
I'm working on a project where we are managing multiple Kafka Streams applications using `StreamsBuilderFactoryBean` and I'm looking to get access to some of the properties stored in the StreamsConfig, specifically `application.id`.

I was planning to open a PR to expose the `application.id` since it's a required property and a (seemingly) good identifier for the streams instance, but I figured it would probably make more sense to run this by you all beforehand. 

Does this seem like a reasonable addition to `StreamsBuilderFactoryBean` and if so, should just `application.id` be exposed, or would it make more sense to expose the whole `StreamsConfig`?

artembilan:
Well, the `StreamsBuilderFactoryBean` is fully based on the injected `StreamsConfig`.
While there is no `getStreamsConfig()` yet you can consider to use those external `StreamsConfig` beans to get desired information.

zacholauson:
Sounds good, I'll plan to open a PR tonight @garyrussell.

That's a good point @artembilan, unfortunately in our case I don't think it will work. We currently have a few streams applications in our service, each with slightly different configs (mostly just application.id), so we would have several pairs of StreamsConfigs and StreamsBuilderFactoryBean, without any connection between them.

PR #620, opened 2018-03-23
GH-618: Add StreamsBuilderFactoryBean#getStreamsConfig

zacholauson:
@garyrussell mentioned returning an `UnmodifiableMap` for the `StreamsConfig` in #618 , but after looking through the `StreamsConfig` code it doesn't look like it's mutable anyways (unless I'm missing something), as it makes a copy of the original configs when querying a `StreamsConfig` instance. With that said, let me know if you would still prefer an `UnmodifiableMap`, I'm happy to change it. As it stands right now, `StreamsConfig` has the benefit of knowing which keys are valid, the expected types of each value which helps for validation when retrieving configs. 

I also marked the return value as `Nullable`, as the `StreamsBuilderFactoryBean` allows configuration after instantiation. I also considering wrapping this in an `Optional`, but I didn't see any usage of `Optionals` in this project, so I wasn't sure if there was a project convention around them.


garyrussell:
Agreed; thanks for the contribution.

Issue #621, opened 2018-03-25
Add StateRestoreListener customization.


nyilmaz:
Would you consider `StateRestoreListener` customization through `StreamsBuilderFactoryBean`? So it can be passed via `KafkaStreams#setGlobalStateRestoreListener`. Extending the factory bean does not work because `KafkaStreams` has private access.

Maybe a better solution would be providing an interface like `KafkaStreamsCustomizer` such that; 

`StateListener`, `UncaughtExceptionHandler`, `cleanupOnStart` and `cleanupOnStop` can be configured through.

artembilan:
Works either way for me.

I even think that both: `setStateRestoreListener()` and `KafkaStreamsCustomizer` are good candidates for improvement.

Contribution is welcome!

nyilmaz:
Signed CLA

PR #625, opened 2018-03-27
Add KafkaStreams customizers and wrapper configurer for StreamsBuilderFactoryBean.

nyilmaz:
Addresses https://github.com/spring-projects/spring-kafka/issues/621


pivotal-issuemaster:
@nyilmaz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=625)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=625) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@nyilmaz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=625)!

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaStreamsDefaultConfiguration.java
```diff
@@ -54,10 +63,13 @@
 
 	@Bean(name = DEFAULT_STREAMS_BUILDER_BEAN_NAME)
 	public StreamsBuilderFactoryBean defaultKafkaStreamsBuilder(
-			@Qualifier(DEFAULT_STREAMS_CONFIG_BEAN_NAME) ObjectProvider<StreamsConfig> streamsConfigProvider) {
+			@Qualifier(DEFAULT_STREAMS_CONFIG_BEAN_NAME) ObjectProvider<StreamsConfig> streamsConfigProvider,
+			Optional<List<KafkaStreamsCustomizer>> kafkaStreamsCustomizers) {
```
artembilan: Do we really need here an `Optional` wrapper? 
Won't we have an empty `List` if there is no `KafkaStreamsCustomizer` beans in the context?

On the other hand I don't think that this is good approach if we are going to manage several `KafkaStreams` in the application. This way you are going to consume all the `KafkaStreamsCustomizer`s where you might need just only one.

I wouldn't allow to inject a `List<KafkaStreamsCustomizer>` at all. A single `KafkaStreamsCustomizer` shouldn't be fully enough. And here, on the `@EnableKafkaStreams` level we would use really some conventional bean name do not clash with any other possible `KafkaStreamsCustomizer` in the application context.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaStreamsConfigurer.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.kafka.streams.KafkaStreams;
+import org.apache.kafka.streams.processor.StateRestoreListener;
+
+/**
+ * Wrapper class of configurable properties of {@link KafkaStreams}. Note that
+ * {@code setters} of this configurer has precedence over {@link KafkaStreamsCustomizer}s
+ *
+ * @author Nurettin Yilmaz
+ */
+class KafkaStreamsConfigurer {
```
artembilan: I don't see reason in this class. Sounds like `CompositeKafkaStreamsCustomizer` would be better solution.

spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
```diff
@@ -140,12 +144,21 @@ public void setClientSupplier(KafkaClientSupplier clientSupplier) {
 		this.clientSupplier = clientSupplier; // NOSONAR (sync)
 	}
 
+	public void addKafkaStreamsCustomizers(List<KafkaStreamsCustomizer> kafkaStreamsCustomizers) {
```
artembilan: This must be setter. Remember XML `<bean>`. 
And as I said before: only single `KafkaStreamsCustomizer` should be injected here. The list-like behavior we can reach with the `CompositeKafkaStreamsCustomizer` implementation.

spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
```diff
@@ -197,22 +209,27 @@ public void stop(Runnable callback) {
...
+				logger.debug(topology.describe());
+				doStart(topology);
 			}
 			catch (Exception e) {
 				throw new KafkaException("Could not start stream: ", e);
 			}
 		}
 	}
 
+	private void doStart(Topology topology) {
```
artembilan: I'm not sure that we need to extract this method. There is no too much complexity and no any exception handling.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaStreamsDefaultConfiguration.java
nyilmaz: Completely right, fixing.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaStreamsConfigurer.java
nyilmaz: Could you eloborate on this a little bit? I though that separating settable fields like `stateListener`, `stateRestoreListener` and `uncaughtExceptionHandler` and configure logic from `StreamsBuilderFactoryBean` is more readable.

spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
nyilmaz: Adding
```
Topology topology = getObject().build();
logger.debug(topology.describe());
```
increases its complexity a little.

artembilan: Well, that's great that you took my attention to this `logger.debug()`.
To be honest I don't want to have it, at least not without `logger.isDebugEnabled()`.

You just moved 8 lines of code to the method without any value in the future. 
That new `doStart()` method doesn't bring any distribution of responsibility.
That's how I would like to live without it. 
More over it's not going to be called from any other place.

Right, we might think different way for code design, but our style (at least mine) is slightly different and I'm not convinced yet in this new method.

Thank you for understanding

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaStreamsCustomizer.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.springframework.kafka.core;
+
+import org.apache.kafka.streams.KafkaStreams;
+
+/**
+ * Callback interface that can be used to configure {@link KafkaStreams} directly.
+ *
+ * @author Nurettin Yilmaz
```
artembilan: `@since 2.1.5`

spring-kafka/src/main/java/org/springframework/kafka/core/CompositeKafkaStreamsCustomizer.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.kafka.streams.KafkaStreams;
+
+/**
+ * Composite {@code KafkaStreamsCustomizer} customizes {@link KafkaStreams} by delegating.
+ *
+ * @author Nurettin Yilmaz
```
artembilan: `@since 2.1.5`

Also sounds like a wrong package for this API.
Would be better to have it in the `config`, but then we will have a package tangle with the `core` via `StreamsBuilderFactoryBean`.

I think we don't have choice now, but we can reconsider the place for these Kafka Streams related classes in the next `2.2` version.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaStreamsDefaultConfiguration.java
```diff
@@ -31,10 +33,14 @@
  * {@link KafkaStreamsDefaultConfiguration#DEFAULT_STREAMS_CONFIG_BEAN_NAME} is present
  * in the application context. Otherwise a {@link UnsatisfiedDependencyException} is thrown.
  *
+ * Also collects {@link KafkaStreamsCustomizer}s to be applied just after {@link KafkaStreams}
+ * created and before {@link KafkaStreams#start()}.
```
artembilan: Sounds like something is missed in this class.
However we discussed with you about some conventional `KafkaStreamsCustomizer`. Similar to what we have now for the `StreamsConfig` with the `DEFAULT_STREAMS_CONFIG_BEAN_NAME`.

Then again: this JavaDoc is not related anymore and has to be changed respectively

spring-kafka/src/main/java/org/springframework/kafka/core/CompositeKafkaStreamsCustomizer.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2017-2018 the original author or authors.
```
artembilan: This class has been created in this `2018`, so, no any past in the Copyright, please.
Just the actual truth.

spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
```diff
@@ -140,12 +150,21 @@ public void setClientSupplier(KafkaClientSupplier clientSupplier) {
 		this.clientSupplier = clientSupplier; // NOSONAR (sync)
 	}
 
+	public void setKafkaStreamsCustomizer(KafkaStreamsCustomizer kafkaStreamsCustomizer) {
+		Assert.notNull(kafkaStreamsCustomizer, "'kafkaStreamsCustomizer' must not be null");
+		this.kafkaStreamsCustomizer = kafkaStreamsCustomizer; // NOSONAR (sync)
+	}
+
 	public void setStateListener(KafkaStreams.StateListener stateListener) {
-		this.stateListener = stateListener; // NOSONAR (sync)
```
artembilan: I think you need to reinstate that Sonar related comment.

spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
```diff
@@ -52,15 +58,19 @@
 
 	private KafkaStreams kafkaStreams;
 
+	private KafkaStreamsCustomizer kafkaStreamsCustomizer = new CompositeKafkaStreamsCustomizer();
```
artembilan: This instance for nothing doesn't have too much value.
Only what it safes for you is just simple `if (this.kafkaStreamsCustomizer != null)`.
I'd prefer such an `if`, instead of extra method call.

spring-kafka/src/main/java/org/springframework/kafka/core/StreamsBuilderFactoryBean.java
```diff
@@ -198,9 +216,15 @@ public synchronized void start() {
...
+				Topology topology = getObject().build();
+				if (logger.isDebugEnabled()) {
+					logger.debug(topology.describe());
+				}
+				this.kafkaStreams = new KafkaStreams(topology, this.streamsConfig, this.clientSupplier);
 				this.kafkaStreams.setStateListener(this.stateListener);
-				this.kafkaStreams.setUncaughtExceptionHandler(this.exceptionHandler);
+				this.kafkaStreams.setGlobalStateRestoreListener(this.stateRestoreListener);
+				this.kafkaStreams.setUncaughtExceptionHandler(this.uncaughtExceptionHandler);
+				this.kafkaStreamsCustomizer.customize(this.kafkaStreams);
```
nyilmaz: @artembilan here one may be overwriting `stateRestoreListener`, `uncaughtExceptionHandler` or `stateListener` via customizer api. Do you think we should warn users about that?

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/e77e1efbc5ef9c27d6aa50d8a7c1e9fd9962b516.

@nyilmaz ,

Thank you for contribution.
Looking forward for more!

PR #633, opened 2018-04-02
Change log-compaction url

neetkee:
Changed log-compaction url. The old is not available, page deleted.


garyrussell:
Thanks; cherry-picked to 2.0.x, 1.3.x

Issue #646, opened 2018-04-13
Detect that a send() operation is being performed when no transaction exists


garyrussell:
It causes an uncommitted transactional producer to be returned to the cache.

See #645

PR #648, opened 2018-04-13
GH-646: Avoid sync when no transaction

garyrussell:
Unfortunately, I had this local change that I failed to push to the PR #647 - it avoids synchronization when there transactions are not being used.

@artembilan I will do the merge/cherry-picking if you like, after your review.


artembilan:
OK. I've merged.
Feel free to cherry-pick.

During the previous one I noticed that we didn't have `2.1.x` CI plan.
Created, ran...

PR #651, opened 2018-04-16
KBGH-356: Fix thread naming with no bean name

garyrussell:
See https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/356

Use token 'consumer' for thread prefix when no bean name.

**cherry-pick to 2.0.x, 1.3.x**


artembilan:
Do we back-port this to `2.1.x` ?

artembilan:
... and cherry-picked to `2.0.x` and `1.3.x`.

The question about `2.1.x` is still opened.
Not sure though why wouldn't we cherry-pick there though...

artembilan:
OK. I took a responsibility and cherry-picked to `2.1.x` as well.
We always can revert the change if that was wrong...

garyrussell:
I keep forgetting we already moved to 2.2 - too many versions 😦

artembilan:
Yeah... That's why I'm still reluctant to merge SIK into SI...

Thanks for merging my PR there though: https://github.com/spring-projects/spring-integration-kafka/pull/202 ! :smile:

Issue #652, opened 2018-04-18
NullPointerException in AbstractJavaTypeMapper


mohamnag:
I'm using the spring-kafka 2.1.5 with kafka-streams 1.0.1 and I'm facing a null pointer with following stack:

```
java.lang.NullPointerException: null
	at org.springframework.kafka.support.converter.AbstractJavaTypeMapper.addHeader(AbstractJavaTypeMapper.java:146)
	at org.springframework.kafka.support.converter.DefaultJackson2JavaTypeMapper.fromJavaType(DefaultJackson2JavaTypeMapper.java:177)
	at org.springframework.kafka.support.serializer.JsonSerializer.serialize(JsonSerializer.java:134)
	at org.apache.kafka.streams.kstream.internals.ChangedSerializer.serialize(ChangedSerializer.java:66)
	at org.apache.kafka.streams.kstream.internals.ChangedSerializer.serialize(ChangedSerializer.java:83)
	at org.apache.kafka.streams.kstream.internals.ChangedSerializer.serialize(ChangedSerializer.java:29)
	at org.apache.kafka.streams.processor.internals.RecordCollectorImpl.send(RecordCollectorImpl.java:91)
	at org.apache.kafka.streams.processor.internals.RecordCollectorImpl.send(RecordCollectorImpl.java:78)
	at org.apache.kafka.streams.processor.internals.SinkNode.process(SinkNode.java:87)
	at org.apache.kafka.streams.processor.internals.ProcessorContextImpl.forward(ProcessorContextImpl.java:85)
	at org.apache.kafka.streams.kstream.internals.KTableRepartitionMap$KTableMapProcessor.process(KTableRepartitionMap.java:94)
	at org.apache.kafka.streams.kstream.internals.KTableRepartitionMap$KTableMapProcessor.process(KTableRepartitionMap.java:72)
	at org.apache.kafka.streams.processor.internals.ProcessorNode$1.run(ProcessorNode.java:46)
	at org.apache.kafka.streams.processor.internals.StreamsMetricsImpl.measureLatencyNs(StreamsMetricsImpl.java:208)
	at org.apache.kafka.streams.processor.internals.ProcessorNode.process(ProcessorNode.java:124)
	at org.apache.kafka.streams.processor.internals.ProcessorContextImpl.forward(ProcessorContextImpl.java:85)
	at org.apache.kafka.streams.kstream.internals.KTableMapValues$KTableMapValuesProcessor.process(KTableMapValues.java:106)
[...]

PR #654, opened 2018-04-18
GH-652: JsonSerializer add null check for Headers

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/652

Kafka Streams `ChangedSerializer` calls the `Serde` with `null` in `headers`.

**cherry-pick to 2.1.x, 2.0.x, 1.3.x**


artembilan:

I'm afraid we can't back-port this to `2.0.x` and `1.3.x`: the `JsonSerializer implements ExtendedSerializer<T>`  exactly since `2.1.0`: #492

garyrussell:
>I'm afraid

Nothing to be afraid of 😄 - it means the problem doesn't exist in those versions 😄

PR #655, opened 2018-04-18
Upgrade to Gradle 4.7 and others, build polishing

artembilan:
* Upgrade all the dependencies to their latest version
* Remove Spring IO plugin stuff as it isn't supported any more
* Remove redundant dependencies since they are polled transitively


build.gradle
```diff
@@ -103,12 +88,11 @@ subprojects { subproject ->
 		testRuntime "org.junit.platform:junit-platform-launcher:$junitPlatformVersion"
 
 		// To support JUnit 4 tests
-		testRuntime "org.junit.vintage:junit-vintage-engine:$junitVintageVersion"
+		testRuntime "org.junit.vintage:junit-vintage-engine:$junitJupiterVersion"
```
garyrussell: Is there some guarantee that these versions will always be aligned?

I based this on JUnit's sample.

artembilan: At least that is what I see in their release page: https://github.com/junit-team/junit5/releases

Strange versioning for Milestone though, so, I decided do not upgrade to that one yet.
Although I can reconsider if you insist.

Issue #657, opened 2018-04-19
KafkaMessageListenerContainer.doInvokeBatchListener()


Gunju-Ko:
Please understand that I'm not good at english.

Even though the transaction is rolled back in the doInvokeBatchListener, the offset can be committed.

Below code is a doInvokeBatchListener's catch
```
catch (RuntimeException e) {
	if (this.containerProperties.isAckOnError() && !this.autoCommit) {
		for (ConsumerRecord<K, V> record : getHighestOffsetRecords(recordList)) {
			this.acks.add(record);
		}
	}
	if (this.batchErrorHandler == null) {
		throw e;
	}
	.... skip
}		
```

If ackOnError is true and autoCommit is false, Add ConsumerRecord to `acks` for committing offset.

If batchErrorHandler is null, exception is throw. So Kafka Transaction will be rollback. (If KafkaMessageListenerContainer is provided KafkaTransactionManager)

problem is that before call KafkaConsumer.poll method() processsCommits() can be called.

below is code
```
while (isRunning()) {
	try {
		if (!this.autoCommit && !this.isRecordAck) {
			processCommits();
		}
``` 

As a result even through Kafka Transaction is rollbacked but offset is committed.

I think before add ConsumerRecord to `acks`, checks producer and batchErrorHandler like below code

```
if (producer != null && this.batchErrorHandler == null) {
	throw e;
}
if (this.containerProperties.isAckOnError() && !this.autoCommit) {
	for (ConsumerRecord<K, V> record : getHighestOffsetRecords(recordList)) {
		this.acks.add(record);
	}
}
if (this.batchErrorHandler == null) {
	throw e;
}
```

Thanks

garyrussell:
Yes, I noticed this yesterday; we should never add to acks when we are using transactions.

I will address it when fixing this: https://github.com/spring-projects/spring-kafka/issues/656

Issue #656, opened 2018-04-18
Transaction Improvements


garyrussell:
`invokeRecordListenerInTx` only seeks the current failed record - it should seek all unprocessed records.

Also, some strategy is needed to discard the poison record and continue. See https://github.com/spring-projects/spring-kafka/issues/653#issuecomment-382489640

PR #659, opened 2018-04-19
GH-656: Fix seek on rollback [Backport]

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/656
Fixes https://github.com/spring-projects/spring-kafka/issues/657

Previously, after a rollback, we only performed a `seek` on the failed record.
We need to seek for all unprocessed records.

Also, when no error handler was provided, and using a batch listener, the
offsets were added to `acks` and incorrectly committed. (#657).

Also, if a `ContainerAwareErrorHandler` "handles" the error, the offsets weren't
committed.

Enhance the tests to verify full seeks.
Add a new test to verify the batch listener doesn't commit after a roll back.

**cherry-pick to 2.1.x, 2.0.x** I will backport to 1.3.x after review.

* Some simple polishing

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AfterRollbackProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AfterRollbackProcessor.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.listener;
+
+import java.util.List;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+
+/**
+ * Invoked by a listener container with remaining, unprocessed, records
+ * (including the failed record). Implementations should seek the desired
+ * topics/partitions so that records will be re-fetched on the next
+ * poll. When used with a batch listener, the entire batch of records is
+ * provided.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Gary Russell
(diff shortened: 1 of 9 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/AfterRollbackProcessor.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Gary Russell
+ *
+ * @since 1.3.5
+ *
+ */
+@FunctionalInterface
```
artembilan: This is not going to work on Java 7, I guess.

Or will be ignored at runtime...

spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
```diff
@@ -0,0 +1,69 @@
...
+		Map<TopicPartition, Long> seekOffsets = new HashMap<>();
+		Iterator<ConsumerRecord<K, V>> iterator = records.iterator();
+		while (iterator.hasNext()) {
+			ConsumerRecord<K, V> record = iterator.next();
+			TopicPartition topicPartition = new TopicPartition(record.topic(), record.partition());
+			if (!seekOffsets.containsKey(topicPartition)) {
+				seekOffsets.put(topicPartition, record.offset());
+			}
+		}
+		for (Entry<TopicPartition, Long> entry : seekOffsets.entrySet()) {
```
artembilan: Why do we need this extra loop if we can just perform `seek` during the first one skipping already processed partitions ?

garyrussell: It seems much more complicated to me to keep track of which ones we've already seeked.

garyrussell: I don't see a big difference and I doubt the performance will be any better - it has to calculate hash codes instead performing the second loop - but feel free to change it if you insist.

artembilan: >  it has to calculate hash codes 

???

Don't you do that already with the current `HashMap` and `seekOffsets.containsKey()` ?

garyrussell: Doh, of course; ok. Feel free to change it.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/e826a587821d967be380d33ac7695f0ddb7ba21b

Issue #661, opened 2018-04-23
Invalid transition attempted from state IN_TRANSACTION to state IN_TRANSACTION after broker bounce


garyrussell:
While testing #660 ...

```kotlin
@RestController
class UserController {

    @Autowired
    private lateinit var kafka: KafkaTemplate<String, String>

    @Transactional
    @RequestMapping(value = ["/user"], method = [RequestMethod.POST])
    fun user(@RequestBody user: String): String {

        kafka.send("user_updated", user)

        println("Hit enter to continue")
        readLine();

        return user
    }
}
```

I stopped the broker, hit enter, and restarted the broker.

After that, all subsequent POSTs got 

```
{"timestamp":"2018-04-23T16:47:10.101+0000","status":500,"error":"Internal Server Error","message":"Could not create Kafka transaction; nested exception is org.apache.kafka.common.KafkaException: TransactionalId txId0: Invalid transition attempted from state IN_TRANSACTION to state IN_TRANSACTION","path":"/user"}
```

PR #662, opened 2018-04-23
GH-661: Close Producer after beginTransaction fail

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/661

I could not reproduce the problem that I reported in the issue; so, if any exception occurs
on `beginTransaction()`, throw the exception to the caller after closing the
producer and prevent its return to the cache.

Also, reading the javadocs for `ProducerFencedException`, we should have been closing
the producer if that exception occurred anyway.

```
/**
 * This fatal exception indicates that another producer with the same <code>transactional.id</code> has been
 * started. It is only possible to have one producer instance with a <code>transactional.id</code> at any
 * given time, and the latest one to be started "fences" the previous instances so that they can no longer
 * make transactional requests. When you encounter this exception, you must close the producer instance.
 */
```

Cherry-pick to master, 2.0.x, 1.3.x.


artembilan:
... and cherry-picked to mentioned version after fixing for proper Mockito classes and not using JUnit 5.

PR #663, opened 2018-04-27
Add Kotlin test case

garyrussell:


build.gradle
```diff
@@ -49,14 +51,22 @@ subprojects { subproject ->
 	apply plugin: 'idea'
 	apply plugin: 'jacoco'
 	apply plugin: 'checkstyle'
+	apply plugin: 'kotlin'
```
artembilan: Let's consider to add `kotlin-spring` as well?
https://github.com/spring-projects/spring-integration/pull/2432

Issue #667, opened 2018-05-04
KafkaMessageListenerContainer.handleConsumerError will never use batchErrorHandler


tengvig:
The handleConsumerError method introduced in pull request #615  will never use batchErrorHandler. It will always use the default errorHandler:

`				if (this.errorHandler != null) {
					this.errorHandler.handle(e, Collections.emptyList(), this.consumer,
							KafkaMessageListenerContainer.this);
				}
				else if (this.batchErrorHandler != null) {
`

When the KafkaMessageListenerContainer instance is created,  a LoggingErrorHandler will always be assigned as errorHandler in addition to a batchErrorHandler. Thus, the LoggingErrorHandler will always be used to handle consumer errors, e.g. quite annoying SerializationErrors.

garyrussell:
I think you meant https://github.com/spring-projects/spring-kafka/pull/616 but, yes, this is a bug.

tengvig:
Damn, that was fast! You guys are awesome :)

PR #669, opened 2018-05-04
GH-667: Fix BatchErrorHandler for generic errors [2.1.x]

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/667

When catching exceptions outside of the listener (e.g. commit errors), the
`BatchErrorHandler` was never called since the record error handler gets a
logging handler.

Check listener type before invoking the error handler for consumer errors.


garyrussell:
Note: the problem did not exist prior to 2.1.5 so no further backports needed.

Issue #666, opened 2018-05-03
Single shared KafkaEmbedded is all JUnit tests


evgeniy-khist:
It seems like there is no convenient way to define single shared KafkaEmbedded that will be used in all JUnit tests.

There might be multiple tests that require `KafkaEmbedded`.
When `@KafkaEmbedded` is used with `@ClassRule`, Kafka server is started and stopped for every test class. 
It may take a lot of time. Better to have embedded Kafka server started only once for all test classes.

Straightforward approach is to add to test sources configuration class defining a `KafkaEmbedded` bean:

```java
@Configuration
public class KafkaEmbeddedConfig {

  @Bean
  public KafkaEmbedded kafkaEmbedded() {
    KafkaEmbedded kafkaEmbedded = new KafkaEmbedded(1, false, 1, "test_topic");
    kafkaEmbedded.setKafkaPorts(9092);
    return kafkaEmbedded;
  }
}
```

The problem is that there might tests creating multiple test contexts:

```java
@TestPropertySource("classpath:test.properties")
@RunWith(SpringRunner.class)
@SpringBootTest
public class KafkaListenerTest {
...
```

and

```java
@RunWith(SpringRunner.class)
@SpringBootTest
@AutoConfigureMockMvc
public class RestControllerTest {
...
```

In this example Kafka server will be started twice. Second executed test class will fail with the exception 

```
Caused by: kafka.common.KafkaException: Socket server failed to bind to localhost:9092: Address already in use.
```

if port is specified or new Kafka server will be started if random port is used.

Moreover, if `@DirtiesContext` is used, together with the refresh of the test context Kafka server will be restarted.

As a workaround `KafkaEmbedded` can be defined as a constant in some base test class that test classes will extend:

```java
@ActiveProfiles("kafka")
public class KafkaTestBase {

  public static final KafkaEmbedded KAFKA_EMBEDDED = createKafkaEmbedded();

  private static KafkaEmbedded createKafkaEmbedded() {
    AnnotationConfigApplicationContext context =
        new AnnotationConfigApplicationContext(KafkaEmbeddedConfig.class);
    KafkaEmbedded kafkaEmbedded = context.getBean(KafkaEmbedded.class);
    Runtime.getRuntime().addShutdownHook(new Thread(context::close));
    return kafkaEmbedded;
  }
}
```

```java
@TestPropertySource("classpath:test.properties")
@RunWith(SpringRunner.class)
@SpringBootTest
[...]

PR #671, opened 2018-05-04
GH-666: Refactor KafkaEmbedded to expose admin and addTopics()

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/666

Allow a single `KafkaEmbedded` to be used across multiple test classes.


garyrussell:
No; you are not missing anything; there are no hooks; the broker only stops when the test JVM shuts down

garyrussell:
Maybe we should just do the refactoring (`addTopics()`) and not institutionalize the holder, but the refactoring would allow users to do it if they understand the limitations.

artembilan:
That's going to be a middle ground, Gary.
I also thought why we don't provide `addTopics()` API there...
:smile:

garyrussell:
**cherry-pick to 2.1**

artembilan:

Issue #676, opened 2018-05-14
KafkaEmbeddedHolder - topic created but not added.


nklmish:
NOTE: In order to reproduce this issue locally please clone https://github.com/nklmish/spring-kafka-issue-676 and execute `./gradlew clean test`.

I ll like to use the same broker(s) for multiple test classes (https://github.com/garyrussell/spring-kafka/commit/24c399ec8992707a35297207d6cc91dcc09fc0f9), and saw this example mentioned in [docs](https://docs.spring.io/spring-kafka/reference/htmlsingle/#_using_the_same_broker_s_for_multiple_test_classes). However, it seems that topic(s) are created in Kafka (I have verified this) but never appended inside [KafkaEmbedded#topics](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java#L121) as a result when we try to access a newly created topic, e.g.
`  embeddedKafka.consumeFromAnEmbeddedTopic(consumer, SOME_TOPIC);`
An assertion error gets thrown due to [this](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java#L481) .  

We should update the topic array after this [call](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java#L255). I can also submit a PR if the above suggestions sounds reasonable.

Thanks!

garyrussell:
Agreed; thanks for the offer of a PR. We should not add them when called from before() though; maybe change the array to a Set? That will also save iterating over the array in `consumeFromEmbeddedTopics()`.

nklmish:
Thank you, set sounds a good idea, however once we change `String[]` -> `Set`, we need to choose one of the following paths:
a). Don't change method signature and perform an internal conversion from `String[]` -> `Set` (it introduce a penalty for converting `list` to `set` in `addTopics` & `consumeFromEmbeddedTopics`)
[...]

PR #677, opened 2018-05-14
GH-676: KafkaEmbeddedHolder - creating and adding topics.

nklmish:
Fixes https://github.com/spring-projects/spring-kafka/issues/676. we have also migrated `topics` from `array` to `set`.

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java
@@ -475,7 +475,7 @@ public class KafkaEmbedded extends ExternalResource implements KafkaRule, Initia
 	 * @throws Exception an exception.
 	 */
 	public void consumeFromAllEmbeddedTopics(Consumer<?, ?> consumer) throws Exception {
-		consumeFromEmbeddedTopics(consumer, this.topics.toArray(new String[this.topics.size()]));
+		consumeFromEmbeddedTopics(consumer, this.topics.toArray(new String[0]));
 	}
 
 	/**
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java
@@ -26,9 +26,11 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java
```diff
@@ -235,31 +238,43 @@ public void before() throws Exception { //NOSONAR
...
 	}
 
+
+	/**
+	 * Add topics to the existing broker(s) using the configured number of partitions.
+	 * @param topics the topics.
+	 * @since 2.1
+	 */
+	public void addTopics(String... topics) {
+		createKafkaTopics(new HashSet<>(Arrays.asList(topics)), this.topics::addAll);
```
artembilan: I don't think we need such a `callback` overhead. You simply can call this `addAll()` in the end of this `addTopics()` method. More over in other place the callback is just empty, so, it really doesn't bring too much value in the logic of this class.

spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java
```diff
@@ -477,9 +496,7 @@ public void consumeFromAnEmbeddedTopic(Consumer<?, ?> consumer, String topic) th
 	 * @throws Exception an exception.
 	 */
 	public void consumeFromEmbeddedTopics(Consumer<?, ?> consumer, String... topics) throws Exception {
-		for (String topic : topics) {
-			assertThat(this.topics).as("topic '" + topic + "' is not in embedded topic list").contains(topic);
-		}
+		assertThat(this.topics).as("topic(s):'" + Arrays.toString(topics) + "' are not in embedded topic list").containsAll(new HashSet<>(Arrays.asList(topics)));
```
artembilan: Well. I would prefer to assert all, but report only wrong topics - don't tell that the whole set is wrong.

spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/KafkaEmbedded.java
```diff
@@ -457,7 +475,7 @@ public boolean isEmbedded() {
 	 * @throws Exception an exception.
 	 */
 	public void consumeFromAllEmbeddedTopics(Consumer<?, ?> consumer) throws Exception {
-		consumeFromEmbeddedTopics(consumer, this.topics);
+		consumeFromEmbeddedTopics(consumer, this.topics.toArray(new String[this.topics.size()]));
```
artembilan: My latest IDEA doesn't recommend me to use arrays with static size for this operation:
image

nklmish: Thank you, updated my intellij and saw the same warning. Should be better now.

garyrussell:
No concerns; LGTM; @nklmish thanks for the contribution.

artembilan:
Sure thing, @nklmish !

Looking forward for more!

nklmish:
@garyrussell @artembilan Thank you very much

artembilan:

PR #679, opened 2018-05-16
Container factory doc polishing

garyrussell:


src/reference/asciidoc/kafka.adoc
```diff
@@ -1381,9 +1381,25 @@ To arbitrarily seek at runtime, use the callback reference from the `registerSee
...
+
+    ConcurrentMessageListenerContainer<String, String> container =
+        factory.createContainer("topic1", "topci2");
+    container.setMessageListener(m -> { ... } );
+    return container;
+}
+----
+
+IMPORTANT: Containers created this way are not added to the endpoint registry.
+They should be created as `@Bean` s or used with the Java DSL so that they will be registered with the application context.
```
artembilan: `Spring Integration Java DSL` and link. This project knows nothing about that Java DSL.

src/reference/asciidoc/si-kafka.adoc
```diff
@@ -277,6 +277,22 @@ In most cases, this will be an `ErrorMessageSendingRecoverer` which will send th
 When building `ErrorMessage` (for use in the `error-channel` or `recovery-callback`), you can customize the error message using the `error-message-strategy` property.
 By default, a `RawRecordHeaderErrorMessageStrategy` is used; providing access to the converted message as well as the raw `ConsumerRecord`.
 
+Starting with _Spring for Apache Kafka version 2.2_ (_Spring Integration Kafka 3.1_), the container factory used for `@KafkaListener` annotations can also be used to create `ConcurrentMessageListenerContainer` s for other purposes.
+See <<container-factory>> for an example.
+
+With the Java DSL, the container does not have to be configured as a `@Bean` because the DSL will register the container as a bean.
```
artembilan: I think the same concern here as well.

src/reference/asciidoc/si-kafka.adoc
```diff
@@ -277,6 +277,22 @@ In most cases, this will be an `ErrorMessageSendingRecoverer` which will send th
...
+----
+@Bean
+public IntegrationFlow topic2ListenerFromKafkaFlow() {
+    return IntegrationFlows
+            .from(Kafka.messageDrivenChannelAdapter(kafkaListenerContainerFactory().createContainer(TEST_TOPIC2),
+                        KafkaMessageDrivenChannelAdapter.ListenerMode.record))
+            ...
+            get();
+}
+----
```
artembilan: Would be great in the end to explain what bean name that container gets.
And also add an `.id()` to the `messageDrivenChannelAdapter()` definition for the complete sample and proper explanation.

Thanks for understanding.

src/reference/asciidoc/si-kafka.adoc
garyrussell: ? Scroll up a few lines...

>When using Spring Integration Java DSL:

I think we can just remove the reference to the DSL in the pure Kafka section.

PR #681, opened 2018-05-18
Add methods receiving ProducerRecord to ProducerListener

endSly:
Also added default implementations to ProducerRecord methods and deprecated the ProducerRecordAdapter

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -354,8 +354,7 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V> {
 						future.set(new SendResult<>(producerRecord, metadata));
 						if (KafkaTemplate.this.producerListener != null
 								&& KafkaTemplate.this.producerListener.isInterestedInSuccess()) {
-							KafkaTemplate.this.producerListener.onSuccess(producerRecord.topic(),
-									producerRecord.partition(), producerRecord.key(), producerRecord.value(), metadata);
+							KafkaTemplate.this.producerListener.onSuccess(producerRecord, metadata);
 						}
 						if (KafkaTemplate.this.logger.isTraceEnabled()) {
 							KafkaTemplate.this.logger.trace("Sent ok: " + producerRecord + ", metadata: " + metadata);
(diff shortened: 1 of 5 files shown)


pivotal-issuemaster:
@endSly Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=681)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=681) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@endSly Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=681)!

artembilan:
@endSly ,

Thank you for the contribution, but would you mind to provide some explanation for reason to have this change. What the change we can see in your code, but "why?" is still a question.

Thank you for understanding.

endSly:
Hi @artembilan ,
I need to read the record headers from the listener, and, with the current callbacks, this is not possible.

Thanks!

garyrussell:
I think it's a valid change, but we won't be able to back-port it as-is to 2.1 because it's a breaking change. If you need this in a 2.1.x release it would have to be implemented in a non-breaking way.

endSly:
@garyrussell, correct me if I'm wrong, but, it should be backward compatible. Adding new methods to `ProducerListener` with a default implementation does not break the previous contract.

garyrussell:
That's true, the contract is ok, but you've changed the template to unconditionally call the new method.

garyrussell:
Doh; sorry - hadn't looked at the default impls.

garyrussell:
@endSly If you would like to go ahead and deprecate that method (something like `* @deprecated it is no longer necessary to implement this method`), and add your name to the `@author` tags for the affected classes, we will go ahead and merge/cherry-pick this PR).

Thanks for the contribution.

endSly:
@garyrussell branch updated.

Thanks!

garyrussell:
Merged as 7a1653b5dd059006c6e4b6508f7cbb89be6baad6 and cherry-picked to 2.1.x as 221ead50fae291dc749af4436990549396ac7de3

nicklaseklund:
@endSly @garyrussell 

In the current documentation: https://docs.spring.io/spring-kafka/docs/2.4.1.RELEASE/reference/html/ it says that "isInterestedInSuccess" should be set if we want to get the onSuccess callback.

But in this commit it has been removed and deprecated. 

Please update the documentation.

garyrussell:
Thanks - https://github.com/spring-projects/spring-kafka/pull/1364

Issue #683, opened 2018-05-21
Transaction not getting committed when error handler is invoked


mawri:
I need to have a simple application doing the following - 
1. consuming from kafka topic
2.1 `doSomeWork()` which involves writing to a different output topic
2.2 In case of `RuntimeException` write the message to another failure topic

This is my code and config

````java
@Test
    public void simpleTransactionTest1() throws InterruptedException {
        String testCaseId = "simpleTransactionTest";
        final CountDownLatch latch = new CountDownLatch(1);
        
        /**
         * producer properties
         */
        Map<String, Object> producerProps = new HashMap<>();
        producerProps.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG,
                "localhost:9092");
        producerProps.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG,
                StringSerializer.class);
        producerProps.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG,
                StringSerializer.class);
        producerProps.put(ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG, "true");
        /**
         * send to test topic
         */
        Producer<String, String> producer = new KafkaProducer<>(producerProps);
        ProducerRecord<String, String> pr = new ProducerRecord<String, String>(
                "test", "simpleTransactionTest");
        producer.send(pr);
        producer.close();

        /**
         * producer factory
         */
        DefaultKafkaProducerFactory<String, String> pf = new DefaultKafkaProducerFactory<>(
                producerProps);
        pf.setTransactionIdPrefix("my.transaction");
        KafkaTemplate<String, String> kafkaTemplate = new KafkaTemplate<>(pf);
        
        /**
         * consumer properties
         */
        Map<String, Object> consumerProps = kafkaConfig.consumerProps();
        consumerProps.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG,
                "localhost:9092");
        consumerProps.put(ConsumerConfig.GROUP_ID_CONFIG, "group");
        consumerProps.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG,
                StringDeserializer.class);
        consumerProps.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG,
                StringDeserializer.class);
        consumerProps.put(ConsumerConfig.ISOLATION_LEVEL_CONFIG,
                "read_committed");
[...]

PR #684, opened 2018-05-21
GH-683: Fix Transactions with ErrorHandler

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/683

If the error handler handles the error, the offset should be sent to the tx.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -1112,12 +1112,7 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 						this.errorHandler.handle(e, record, this.consumer);
 					}
 					if (producer != null) {
-						try {
-							sendOffsetsToTransaction(producer);
-						}
-						catch (Exception e1) {
-							this.logger.error("Send offsets to transaction failed", e1);
-						}
+						ackCurrent(record, producer);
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1112,12 +1112,7 @@ private RuntimeException doInvokeRecordListener(final ConsumerRecord<K, V> recor
 						this.errorHandler.handle(e, record, this.consumer);
 					}
 					if (producer != null) {
-						try {
-							sendOffsetsToTransaction(producer);
-						}
-						catch (Exception e1) {
-							this.logger.error("Send offsets to transaction failed", e1);
-						}
+						ackCurrent(record, producer);
```
artembilan: I wonder why the block in the beginning of the `catch` doesn't work:
```
if (this.containerProperties.isAckOnError() && !this.autoCommit && producer == null) {
		ackCurrent(record, producer);
}
```

That's one of the option the reporter in the #683 uses.

May you think what's wrong is there?

Thanks

garyrussell: Because the producer is not null.

garyrussell: We can't do this up there because we need to wait for the error handler to process before we decide to commit or roll back.

PR #687, opened 2018-05-22
Emit WARN log for @SendTo with void return

garyrussell:
Typo in doc.

**cherry-pick to 2.1.x**


artembilan:
Well, looks like I have merged this PR prematurely.
https://build.spring.io/browse/SK-MAS-401/

So, we have a test-case like: `testVoidListenerWithReplyingErrorHandler()`. Therefore it light be OK to have a `@SendTo` and `void` return, but the result of the `errorHandler` is going to be used for sending to.

Right now it looks like you have broker the logic in the `MethodKafkaListenerEndpoint`.

WDYT now ?

Thanks

Issue #691, opened 2018-05-24
ProducerListener: add CompositeProducerListener or list of ProducerListener


mayras:
Typical situation when there is the need of list of listeners, but there can be set up one ProducerListener for KafkaTemplate only. Pls add CompositeProducerListener with list of inner listeners or 
enhance KafkaTemplate to addProducerListener method.

artembilan:
I think this is fully valid request and `CompositeProducerListener` sounds good to me.

I guess you can simply implement one for your project yourself and when you're done, feel free to contribute it back to the Framework!

mayras:
public class CompositeProducerListener<K, V> implements ProducerListener<K, V> {

    private List<ProducerListener<K,V>> producerListeners = newArrayList();
    private boolean interestedInSuccess;

    public CompositeProducerListener(List<ProducerListener<K,V>> producerListeners, boolean interestedInSuccess) {
        this.producerListeners = producerListeners;
        this.interestedInSuccess = interestedInSuccess;
    }

    @Override
    public void onSuccess(String topic, Integer partition, K key, V value, RecordMetadata recordMetadata) {
        producerListeners.forEach(c -> c.onSuccess(topic, partition, key, value, recordMetadata));
    }

    @Override
    public void onError(String topic, Integer partition, K key, V value, Exception exception) {
        producerListeners.forEach(c -> c.onError(topic, partition, key, value, exception));
    }

    @Override
    public boolean isInterestedInSuccess() {
        return interestedInSuccess;
    }
}

PR #695, opened 2018-05-28
GH-691: Add CompositeProducerListener

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/691

**cherry-pick to 2.1.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/CompositeProducerListener.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/CompositeProducerListener.java
@@ -65,7 +65,7 @@ public class CompositeProducerListener<K, V> implements ProducerListener<K, V> {
 
 	@Override
 	public void onSuccess(ProducerRecord<K, V> producerRecord, RecordMetadata recordMetadata) {
-		this.delegates.forEach(d -> d.onSuccess(null, recordMetadata));
+		this.delegates.forEach(d -> d.onSuccess(producerRecord, recordMetadata));
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/CompositeProducerListener.java
```diff
@@ -0,0 +1,76 @@
...
+		this.delegates.add(delegate);
+	}
+
+	public boolean removeDelegate(ProducerListener<K, V> delegate) {
+		return this.delegates.remove(delegate);
+	}
+
+	@Override
+	public void onSuccess(ProducerRecord<K, V> producerRecord, RecordMetadata recordMetadata) {
+		this.delegates.forEach(d -> d.onSuccess(null, recordMetadata));
```
artembilan: Why `null`? What am I missing?
Thanks

spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
```diff
@@ -242,16 +243,19 @@ public void withListener() throws Exception {
...
-		final CountDownLatch latch = new CountDownLatch(1);
-		template.setProducerListener(new ProducerListener<Integer, String>() {
+		final CountDownLatch latch = new CountDownLatch(2);
+		class PL implements ProducerListener<Integer, String> {
 
 			@Override
-			public void onSuccess(String topic, Integer partition, Integer key, String value,
-					RecordMetadata recordMetadata) {
+			public void onSuccess(ProducerRecord<Integer, String> record, RecordMetadata recordMetadata) {
 				latch.countDown();
```
artembilan: Sounds like test is not full - no `record` and `recordMetadata` assertions.

spring-kafka/src/main/java/org/springframework/kafka/support/CompositeProducerListener.java
garyrussell: I have no idea where that came from 😄

artembilan:

Issue #701, opened 2018-06-01
Enable Batch KafkaListener to allow ConsumerRecords as a parameter


zach-schoenberger:
The current implementation of batch listener combines all records returned into a list. This loses a lot of information that is provided when the data is returned as a ConsumerRecords. Most importantly, the records are already grouped by topic and partition at this point, which can be useful when implementing processing logic. I would suggest adding additional processing logic to allow the KafkaListener to pass ConsumerRecords as a parameter when batch processing is enabled.

garyrussell:
Good point; the reasoning for converting to a list was so that a `FilteringBatchMessageListenerAdapter` could wrap the listener, but I agree it should be an option.

PR #703, opened 2018-06-08
GH-701: Support ConsumerRecords<?, ?> (@Klistener)

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/701

Support passing the entire `ConsumerRecords<?, ?>` object to the listener.
Record filtering is not applied in this case.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -360,6 +360,8 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 
 		private final boolean isBatchListener;
 
+		private final boolean wantsFullRecords;
+
 		private final boolean autoCommit = KafkaMessageListenerContainer.this.consumerFactory.isAutoCommit();
 
 		private final boolean isManualAck = this.containerProperties.getAckMode().equals(AckMode.MANUAL);
@@ -458,11 +460,13 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 				this.listener = null;
 				this.batchListener = (BatchMessageListener<K, V>) listener;
 				this.isBatchListener = true;
+				this.wantsFullRecords = this.batchListener.wantsPollResult();
 			}
 			else if (listener instanceof MessageListener) {
 				this.listener = (MessageListener<K, V>) listener;
 				this.batchListener = null;
 				this.isBatchListener = false;
+				this.wantsFullRecords = false;
 			}
 			else {
 				throw new IllegalArgumentException("Listener must be one of 'MessageListener', "
@@ -872,11 +876,13 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 
 		private void invokeBatchListener(final ConsumerRecords<K, V> records) {
 			List<ConsumerRecord<K, V>> recordList = new LinkedList<ConsumerRecord<K, V>>();
-			Iterator<ConsumerRecord<K, V>> iterator = records.iterator();
-			while (iterator.hasNext()) {
-				recordList.add(iterator.next());
+			if (!this.wantsFullRecords) {
+				Iterator<ConsumerRecord<K, V>> iterator = records.iterator();
+				while (iterator.hasNext()) {
+					recordList.add(iterator.next());
+				}
 			}
-			if (recordList.size() > 0) {
+			if (recordList.size() > 0 || this.wantsFullRecords) {
 				if (this.transactionTemplate != null) {
 					invokeBatchListenerInTx(records, recordList);
 				}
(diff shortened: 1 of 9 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1358,6 +1386,17 @@ private void commitIfNecessary() {
 			return highestOffsetMap.values();
 		}
 
+		private Collection<ConsumerRecord<K, V>> getHighestOffsetRecords(ConsumerRecords<K, V> records) {
+			final Map<TopicPartition, ConsumerRecord<K, V>> highestOffsetMap = new HashMap<>();
+			Iterator<ConsumerRecord<K, V>> iterator = records.iterator();
+			while (iterator.hasNext()) {
+				ConsumerRecord<K, V> r = iterator.next();
+				highestOffsetMap.compute(new TopicPartition(r.topic(), r.partition()),
```
artembilan: Can't really `ConsumerRecords.partitions()` help us here somehow?

garyrussell: We'd still have to iterate but, yes, it would avoid having to create new TPs.

Good catch!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1449,34 +1425,10 @@ public String toString() {
...
-
-		}
-
-		private final class ConsumerRecordsAcknowledgment implements Acknowledgment {
-
 			private final ConsumerRecords<K, V> records;
 
-			ConsumerRecordsAcknowledgment(ConsumerRecords<K, V> records) {
+			ConsumerBatchAcknowledgment(ConsumerRecords<K, V> records) {
+				// make a copy in case the listener alters the list
```
artembilan: ??? But you don't make copy here.
I guess this is a leftover from the previous version.

Anyway let's think what really makes sense to to do here!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -872,11 +876,13 @@ private void invokeListener(final ConsumerRecords<K, V> records) {
 
 		private void invokeBatchListener(final ConsumerRecords<K, V> records) {
 			List<ConsumerRecord<K, V>> recordList = new LinkedList<ConsumerRecord<K, V>>();
```
artembilan: May we consider do not initialize this variable if we are not interested in the list, but the whole `records` ?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1349,11 +1363,11 @@ private void commitIfNecessary() {
 			return commits;
 		}
 
-		private Collection<ConsumerRecord<K, V>> getHighestOffsetRecords(List<ConsumerRecord<K, V>> records) {
+		private Collection<ConsumerRecord<K, V>> getHighestOffsetRecords(ConsumerRecords<K, V> records) {
 			final Map<TopicPartition, ConsumerRecord<K, V>> highestOffsetMap = new HashMap<>();
-			records.forEach(r -> {
-				highestOffsetMap.compute(new TopicPartition(r.topic(), r.partition()),
-						(k, v) -> v == null ? r : r.offset() > v.offset() ? r : v);
+			records.partitions().forEach(tp -> {
```
artembilan: `records.partitions().stream()...collect(Collectors.toMap())` instead?

spring-kafka/src/main/java/org/springframework/kafka/listener/SeekToCurrentBatchErrorHandler.java
```diff
@@ -39,7 +39,8 @@
 	public void handle(Exception thrownException, ConsumerRecords<?, ?> data, Consumer<?, ?> consumer,
 			MessageListenerContainer container) {
 		Map<TopicPartition, Long> offsets = new LinkedHashMap<>();
-		data.forEach(r -> offsets.computeIfAbsent(new TopicPartition(r.topic(), r.partition()), k -> r.offset()));
+		data.partitions()
+			.forEach(tp -> offsets.put(tp, data.records(tp).get(0).offset()));
```
artembilan: Similar stream approach here?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: I don't think that will be efficient because we'd have to iterate over the lists to get the highest rather than indexing directly.

garyrussell: Correction - I figured it out.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/91e0f62908f2270647df16818e18b19d0e08f19e

Issue #686, opened 2018-05-22
NPE in JsonSerializer on template.send if key is null


Shabin:
I just upgraded from 1.3.4.RELEASE to 2.1.6.RELEASE and on sending a message without any key I am getting NPE.
```java
-java.lang.NullPointerException
	at org.springframework.kafka.support.serializer.JsonSerializer.serialize(JsonSerializer.java:134)
	at org.apache.kafka.clients.producer.KafkaProducer.doSend(KafkaProducer.java:783)
	at org.apache.kafka.clients.producer.KafkaProducer.send(KafkaProducer.java:768)
	at org.springframework.kafka.core.DefaultKafkaProducerFactory$CloseSafeProducer.send(DefaultKafkaProducerFactory.java:285)
	at org.springframework.kafka.core.KafkaTemplate.doSend(KafkaTemplate.java:348)
	at org.springframework.kafka.core.KafkaTemplate.send(KafkaTemplate.java:181)
	at com.itest.event.distributer.kafka.KafkaEventDistributor.notify(KafkaEventDistributor.java:226)
```

I am calling the default send() method.
`template.send(topicName, eventdata);`

I coudn't find any migration steps related to this.

My kafka template configuration:
```xml
<bean id="kafkaTemplate" class="org.springframework.kafka.core.KafkaTemplate">
		<constructor-arg>
			<bean class="org.springframework.kafka.core.DefaultKafkaProducerFactory">
				<constructor-arg>
					<map>
						<entry key="bootstrap.servers" value="${event.kafka.url}" />
						<entry key="key.serializer"
							value="org.springframework.kafka.support.serializer.JsonSerializer" />
						<entry key="value.serializer"
							value="com.itest.event.kafka.EventJsonSerializer" />
					</map>
				</constructor-arg>
			</bean>
		</constructor-arg>
	</bean>
```

artembilan:
I think this is a good catch and our problem that we don't have this there:
```
 if (data == null)
            return null;
```

I mean that is is normal in Kafka world to send `null` as for `value`, as well as for `key`.

Therefore this `NPE` has to be fixed.

Feel free to contribute it, @Shabin !

Thank you!

PR #707, opened 2018-06-13
GH-686: Handle tombstones in Json Ser/Deser

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/686

**Cherry-pick to 2.1.x, 2.0.x, 1.3.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerializer.java
@@ -130,6 +130,9 @@ public class JsonSerializer<T> implements ExtendedSerializer<T> {
 
 	@Override
 	public byte[] serialize(String topic, Headers headers, T data) {
+		if (data == null) {
+			return null;
+		}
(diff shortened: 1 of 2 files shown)


artembilan:
... and cherry-picked to all mentioned versions after resolving conflicts about missed `headers` variant in the version < `2.1`.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerializer.java
```diff
@@ -138,6 +141,9 @@ else if (config instanceof String) {
 
 	@Override
 	public byte[] serialize(String topic, T data) {
+		if (data == null) {
+			return null;
+		}
 		try {
 			byte[] result = null;
 			if (data != null) {
```
tsachev: this check for not null is now redundant

artembilan: 😄 

Good catch! 
Feel free to provide a contribution on the matter!

Issue #716, opened 2018-06-26
KafkaNull and custom deserializer confusing stacktrace (documentation)


jorgheymans:
I have a custom deserializer that returns null when the payload is somehow not what we expect it to be. spring-kafka converts the `null` to `KafkaNull` and I guess then tries to match the @KafkaListener annotated method with this object. This then prints below confusing stacktrace about not being able to invoke my listener methods:

```
org.springframework.kafka.listener.ListenerExecutionFailedException: Listener method could not be invoked with the incoming message
Endpoint handler details:
Method [public void MyService.onCommand(my.CCMessage)]
Bean [MyService@2689b752]; nested exception is org.springframework.messaging.handler.annotation.support.MethodArgumentNotValidException: Could not resolve method parameter at index 0 in public void MyService.onCommand(my.CCMessage): 1 error(s): [Error in object 'command': codes []; arguments []; default message [Payload value must not be empty]] , failedMessage=GenericMessage [payload=org.springframework.kafka.support.KafkaNull@70b6d3ac, headers={kafka_offset=353, kafka_consumer=brave.kafka.clients.TracingConsumer@75b04b84, kafka_timestampType=CREATE_TIME, kafka_receivedMessageKey=null, kafka_receivedPartitionId=4, kafka_receivedTopic=command, kafka_receivedTimestamp=1530011220100}]
	at org.springframework.kafka.listener.adapter.MessagingMessageListenerAdapter.invokeHandler(MessagingMessageListenerAdapter.java:262) ~[spring-kafka-2.1.7.RELEASE.jar:2.1.7.RELEASE]
	at org.springframework.kafka.listener.adapter.RecordMessagingMessageListenerAdapter.onMessage(RecordMessagingMessageListenerAdapter.java:80) ~[spring-kafka-2.1.7.RELEASE.jar:2.1.7.RELEASE]
	at org.springframework.kafka.listener.adapter.RecordMessagingMessageListenerAdapter.onMessage(RecordMessagingMessageListenerAdapter.java:51) ~[spring-kafka-2.1.7.RELEASE.jar:2.1.7.RELEASE]
	at org.springframework.kafka.listener.adapter.RecordMessagingMessageListenerAdapter$$FastClassBySpringCGLIB$$cde8c01d.invoke(<generated>) ~[spring-kafka-2.1.7.RELEASE.jar:2.1.7.RELEASE]
[...]

PR #717, opened 2018-06-26
GH-716: Null Payload Doc Improvements

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/716

Improve documentation - `null` is not necessarily just for tombstone records.

__cherry-pick to 2.1.x__


artembilan:

Issue #719, opened 2018-06-29
Spring 5 support for spring-kafka-test


fabienmifsud:
Hi all,

Can spring-kafka-test support spring 5 in the next release ?

I can help on that subject if needed.

Thx !

artembilan:
Would you mind being more precise?
Spring Kafka project is based on Spring 5 since version `2.0`: https://docs.spring.io/spring-kafka/docs/2.0.0.RELEASE/reference/html/_introduction.html#compatibility

fabienmifsud:
spring-kafka-test has spring-retry dependency wich refeers to a spring 4.x version and that leads my tests to have some class not found..

I did make a PR on it https://github.com/spring-projects/spring-retry/pull/117 to upgrade it.

artembilan:
Well, that's not so easy to move Spring Retry to Spring 5, because the last one requires Java 8 minimum and for that reason we would need to start a version `2.0`.
That even might be a bad idea, but we don't have themes for `2.0` version yet.
On the other hand we haven't done that yet just because the current Spring Retry version is fully compatible with Spring 5 - we just only need to have proper dependencies in the classpath!

To fix the problem in your project you need to `exclude` Spring 4 transitive dependencies from the spring-retry dependency:

```
compile ("org.springframework.retry:spring-retry:1.2.2.RELEASE") {
    exclude group: 'org.springframework'
}
```

I would say that exactly this fix has to be done here in this Spring Kafka project.

fabienmifsud:
I'm not sure to well understand why you don't want to upgrade versions, but
ok. I did try to exclude spring-retry and spring-core from transitive test
dependencies and it hasn't been enough. I'll try to exclude the whole
groupId and let you know. Thanks for your time

Le ven. 29 juin 2018 à 17:07, Artem Bilan <notifications@github.com> a
écrit :

artembilan:
Just bumping version is not enough to start a new major version.
We can't do what you suggest in your PR in the current minor version: it is going to be a big breaking change for our users and we are very reluctant to make breaking change.
Another argument do not start a new major version I've explained before: Spring Retry works well with the latest Spring 5.0 - just need to adjust dependencies in your project a bit!
Otherwise we would need to support two major versions in parallel. 

Nevertheless, there is no reason to start a new major version because everything is compatible in the current state.

I didn't tell you to exclude `spring-retry`. Its Spring transitive dependencies must be excluded.
[...]

PR #721, opened 2018-07-02
Exclude Spring 4.X dependency

fabienmifsud:
Exclusion of Spring 4.X transitive dependencies from Spring-Retry
#719

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -190,7 +190,7 @@ project ('spring-kafka') {
 		compile "org.springframework:spring-messaging:$springVersion"
 		compile "org.springframework:spring-tx:$springVersion"
 		compile ("org.springframework.retry:spring-retry:$springRetryVersion") {
-			exclude group: 'org.springframework'
+			exclude group: 'org.springframework', module: 'spring-core'
 		}
 		compile "org.apache.kafka:kafka-clients:$kafkaVersion"
 		compile ("org.apache.kafka:kafka-streams:$kafkaVersion", optional)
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -189,7 +189,9 @@ project ('spring-kafka') {
 		compile "org.springframework:spring-context:$springVersion"
 		compile "org.springframework:spring-messaging:$springVersion"
 		compile "org.springframework:spring-tx:$springVersion"
-		compile "org.springframework.retry:spring-retry:$springRetryVersion"
+		compile ("org.springframework.retry:spring-retry:$springRetryVersion") {
+			exclude group: 'org.springframework'
```
artembilan: I suggest to do like this `exclude group: 'org.springframework', module: 'spring-core'` and back-port it into the `2.0.x` and `2.1.x` as well:
1. Those versions suffer the same dependency issue
2. We need to be compatible with the Spring IO Platform there.

Otherwise - LGTM!

fabienmifsud: Hi @artembilan ,
I'll modify the way you suggest. Should I close this PR and open two new ones linked to 2.0.X and 2.1.X ?

artembilan: No, you don't need to do that - we will simply `cherry-pick` your commit to those versions.

Thanks

build.gradle
```diff
@@ -190,7 +190,7 @@ project ('spring-kafka') {
 		compile "org.springframework:spring-messaging:$springVersion"
 		compile "org.springframework:spring-tx:$springVersion"
 		compile ("org.springframework.retry:spring-retry:$springRetryVersion") {
-			exclude group: 'org.springframework'
+			exclude group: 'org.springframework', module: 'spring-core'
```
artembilan: Just to clarify a bit: we don't need to exclude any other Spring dependencies since only this one is not `optional` or `test` in that project: https://github.com/spring-projects/spring-retry/blob/master/pom.xml#L244

artembilan:
... and cherry-picked to `2.1.x` and `2.0.x`.

@fabienmifsud ,

Thank you for the contribution!

Issue #726, opened 2018-07-05
Allow empty @SendTo with Message<?> return type


garyrussell:
Since the topic is taken from the message headers, the `@SendTo` does not need a value.

See https://github.com/spring-projects/spring-kafka/issues/725

Issue #725, opened 2018-07-05
Support `@SendTo` (replying) for log-compacted topics


nyilmaz:
Currently `MessagingMessageListenerAdapter#sendResponse(Object result, String topic, @Nullable Object source)` method is not setting key before sending and causes `CorruptRecordException` as stated [here](https://stackoverflow.com/questions/49098274/kafka-stream-get-corruptrecordexception).

Workaround is to write a `ProducerInterceptor` and add it to producer config like;

`extended.put(ProducerConfig.INTERCEPTOR_CLASSES_CONFIG, List.of(KeySetterProducerInterceptor.class));`

Planning to support this?

garyrussell:
The problem is there is only a single result; we'd need some expression to evaluate to determine what to set the key to.

We'd need another `@SendTo` annotation since the current one is owned by spring-messaging (or we'd need a change there).

There currently is another workaround - return a message...

```java
@SpringBootApplication
public class Kgh725Application {

	public static void main(String[] args) {
		SpringApplication.run(Kgh725Application.class, args);
	}

	@Bean
	public ApplicationRunner runner(KafkaTemplate<String, String> template) {
		return args -> {
			template.send("kgh725-1", "foo", "bar");
		};
	}

	@KafkaListener(id = "one", topics = "kgh725-1")
	@SendTo("kgh725-2")
	public Message<?> listen(String in) {
		return MessageBuilder.withPayload(in.toUpperCase())
				.setHeader(KafkaHeaders.TOPIC, "kgh725-2")
				.setHeader(KafkaHeaders.MESSAGE_KEY, "baz")
				.build();
	}

	@KafkaListener(id = "two", topics = "kgh725-2")
	public void listen(ConsumerRecord<String, String> in) {
		System.out.println(in);
	}

	@Bean
	public NewTopic topic1() {
		return new NewTopic("kgh725-1", 1, (short) 1);
	}

	@Bean
	public NewTopic topic2() {
		return new NewTopic("kgh725-2", 1, (short) 1);
	}

}
```

(Note that you have to populate the topic header too for this work (it's also needed in the `@SendTo`).

Result:

```
ConsumerRecord(topic = kgh725-2, partition = 0, offset = 0, CreateTime = 1530812067965, serialized key size = 3, serialized value size = 3, headers = RecordHeaders(headers = [], isReadOnly = false), 
key = baz, value = BAR)
```

nyilmaz:
Is this approach valid for batch message listeners?

[...]

PR #729, opened 2018-07-06
GH-725: Fix return Collection<Message<?>>

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/725
Fixes https://github.com/spring-projects/spring-kafka/issues/726

Handle return type `Collection<Message<?>>`.
Allow empty `@SendTo` when the return type is `Message<?>` (or a collection thereof).

### DIFF
--- a/src/reference/asciidoc/kafka.adoc
+++ b/src/reference/asciidoc/kafka.adoc
@@ -1171,6 +1171,22 @@ public KafkaTemplate<String, String> myReplyingTemplate() {
 }
 ----
 
+IMPORTANT: If the listener method returns `Message<?>` or `Collection<Messge<?>>` the listener method is responsible for setting up the message headers for the reply; for example, when handling a request from a `ReplyingKafkaTemplate`, you might do the following:
+
+[source, java]
+----
+@KafkaListener(id = "messageReturned", topics = "someTopic")
+public Message<?> listen(String in, @Header(KafkaHeaders.REPLY_TOPIC) byte[] replyTo,
+        @Header(KafkaHeaders.CORRELATION_ID) byte[] correlation) {
+    return MessageBuilder.withPayload(in.toUpperCase())
+            .setHeader(KafkaHeaders.TOPIC, replyTo)
+            .setHeader(KafkaHeaders.MESSAGE_KEY, 42)
+            .setHeader(KafkaHeaders.CORRELATION_ID, correlation)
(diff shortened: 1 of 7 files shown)


src/reference/asciidoc/kafka.adoc
```diff
@@ -1171,6 +1171,22 @@ public KafkaTemplate<String, String> myReplyingTemplate() {
 }
 ----
 
+IMPORTANT: If the listener method returns `Message<?>` or `Collection<Messge<?>>` the listener method is responsible for setting up the message headers for the reply; for example, when handling a request from a `ReplyingKafkaTemplate`, you might do the following:
```
artembilan: When `if`, `when` etc is in the first part of the sentence, it must be separated with a comma : https://www.ego4u.com/en/cram-up/writing/comma?10 😉 

I mean in this case we need comma before `the listener method`.

src/reference/asciidoc/kafka.adoc
```diff
@@ -1171,6 +1171,22 @@ public KafkaTemplate<String, String> myReplyingTemplate() {
...
+
+[source, java]
+----
+@KafkaListener(id = "messageReturned", topics = "someTopic")
+public Message<?> listen(String in, @Header(KafkaHeaders.REPLY_TOPIC) byte[] replyTo,
```
artembilan: Are you missing `@SendTo` here?
Or it doesn't matter and non-`void` return is enough to produce reply?

garyrussell: Interesting; yes; I guess if the return type is `Message<?>` the `@SendTo` is entirely optional - we could enforce it, I suppose, but maybe it's ok. 

WDYT?

src/reference/asciidoc/kafka.adoc
garyrussell: Hehe - thanks, yes, please fix on merge (unless you want me to do something with your other comment).

src/reference/asciidoc/kafka.adoc
artembilan: Ah! Yes, it's really OK: your sentence really implies that.
So, yeah, merging...
Thanks

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/d9c23e3c32e2687b9022ce6d1f72ce2955ecce0a

Issue #731, opened 2018-07-08
ReplyingKafkaTemplate scalability restrictions


b-liberman:
Hello, 
is it possible to remove the restrictions of ReplyingKafkaTemplate that every client instance requires its own topic or partition?

According to the reference 
> If you have multiple client instances, each will need a dedicated reply topic for each instance. An alternative is to set the KafkaHeaders.REPLY_PARTITION and use a dedicated partition for each instance

This makes an easy (auto-)scaling cumbersome. 
Looking at the code, my understanding is that this is only required to produce an error log entry: 
```java
RequestReplyFuture<K, V, R> future = this.futures.remove(correlationId);
if (future == null) {
	this.logger.error("No pending reply: " + record + " with correlationId: "
		+ correlationId + ", perhaps timed out");
}
else {
	if (this.logger.isDebugEnabled()) {
		this.logger.debug("Received: " + record + " with correlationId: " + correlationId);
	}
	future.set(record);
}
```

Is not it better to 

1. allow multiple (client) instances listen to the same topic and filter their own correlation ids

1. in case a timeout occurs to move the entry (future) with that correlation ID into another map for a certain period of time and report the "perhaps timed out" based on that map. If nothing found there - do not report anything.

I also suggest to improve the implementation and set the reply header automatically if possible, i.e. remove the need for 
```java
record.headers().add(new RecordHeader(KafkaHeaders.REPLY_TOPIC, "kReplies".getBytes()));
```
and set the header in `ReplyingKafkaTemplate` based on the reply container configuration as they need to match anyway.

Regards

Boris

garyrussell:
>Is not it better to

"Better" is rather subjective here. If it's ok to waste broker, network and CPU resources receiving and discarding unexpected replies then using a common reply topic is ok.

There is no actual restriction - if each instance uses a unique `group.id` in the reply container, all instances will get all replies, which will be discarded by the uninterested instances; you can suppress the ERROR log with appropriate configuration of your logging subsystem. We could consider making that log message level configurable.

I don't think there is a need for a "second map" at all; the "right" template will either find, or not, the `Future<?>` and the others will never find it. The timeout is logged as a warning elsewhere.

I agree that the template can set the reply topic header, if the container has simple configuration (i.e. it is configured to listen to a single topic and not a topic pattern).

b-liberman:
thanks a lot! 

[...]

PR #733, opened 2018-07-09
GH-731: Replying template improvements

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/731

- Auto detect reply topic/partition
- Improve docs to explain how to use a shared reply topic
- Add boolean to reduce log level for unexpected responses

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
@@ -17,6 +17,7 @@
 package org.springframework.kafka.requestreply;
 
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.Collection;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
```diff
@@ -74,7 +77,13 @@
...
 	private long replyTimeout = DEFAULT_REPLY_TIMEOUT;
 
-	private volatile boolean schedulerSet;
+	private boolean schedulerSet;
+
+	private boolean sharedReplyTopic;
+
+	private String replyTopic;
+
+	private byte[] replyPartition;
```
artembilan: Can't they both be as `final`?
Plus can't we convert `topic` to `byte[]` immediately in the ctor instead of on each request?

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
```diff
@@ -240,8 +287,16 @@ public void onMessage(List<ConsumerRecord<K, R>> data) {
...
-							+ correlationId + ", perhaps timed out");
+					if (this.sharedReplyTopic) {
+						if (this.logger.isDebugEnabled()) {
+							this.logger.debug("No pending reply: " + record + " with correlationId: "
+									+ correlationId + ", perhaps timed out, or using a shared reply topic");
+						}
+					}
+					else if (this.logger.isErrorEnabled()) {
+						this.logger.error("No pending reply: " + record + " with correlationId: "
+								+ correlationId + ", perhaps timed out, or using a shared reply topic");
```
artembilan: Looks like message is the same for both logging.
Can't we extract it into the `private` method?

artembilan:
Looks like there is some conflict with the `master`.

Issue #732, opened 2018-07-09
Add support for JUnit 5 extension.


NeveHanter:
Currently `spring-kafka` allow using `KafkaEmbedded` using `@ClassRule` or using spring's injection thanks to provided context customizer.

Is there any reason why there's no JUnit 5 extension provided or are there any plans on supporting this?

For now I'm using custom written JUnit 5 extension which implements both replication manager shutdown and high watermark _hacks_ from #194 but it will be best to have this done properly by spring-kafka itself.

garyrussell:
We didn't write a JUnit5 extension because we already have the `@EmbeddedKafka` annotation, which can be used with JUnit5.

[Documentation here](https://docs.spring.io/spring-kafka/reference/html/_reference.html#__embeddedkafka_annotation).

Feel free to create a PR for your extension and we'll consider merging it if it adds significant value.

Woodham:
@garyrussell As far as I can see, the `@EmbeddedKafka` still requires junit4 (or junit-jupiter-vintage) on the classpath because `KafkaEmbedded.java` imports `org.junit.rules.ExternalResource`.

Is there any way to use `KafkaEmbedded` without having junit 4 on the classpath?

artembilan:
Yep! That's good catch, indeed!

I think we need a JUnit 5 variant of the `KafkaEmbedded` and that `@EmbeddedKafka` should be able to select the proper implementation according classpath state.
Although we definitely should extract a common logic in the separate class, free from JUnit at all and provide appropriate wrappers.
Something like `KafkaEmbeddedJunit4Rule` and `KafkaEmbeddedJunit5Extension` ?

d35h:
Does it work for you now with JUnit4?
I have the following when running ```./gradlew build```:
```
Supertypes of the following classes cannot be resolved. Please make sure you have the required  dependencies in the classpath:
    class org.springframework.kafka.test.rule.EmbeddedKafkaRule, unresolved supertypes: org.junit.rules.ExternalResource, org.junit.rules.TestRule
    class org.springframework.kafka.test.rule.KafkaRule, unresolved supertypes: org.junit.rules.TestRule
```

Those are my dependencies:
```
testImplementation("org.springframework.kafka:spring-kafka-test")
testImplementation("org.junit.jupiter:junit-jupiter-api:$junitVersion")
testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:$junitVersion")
[...]

PR #743, opened 2018-07-17
GH-732: Refactor KafkaEmbedded for JUnit 5

artembilan:
Fixes https://github.com/spring-projects/spring-kafka/issues/732

The current `KafkaEmbedded` is fully tied with the JUnit 4 and can't be
used with just plain JUnit 5.

* Extract `EmbeddedKafkaBroker.java` class to the core level and add
an `EmbeddedKafkaRule` wrapper
* The `@EmbeddedKafka` now creates a `EmbeddedKafkaBroker` bean which is
still good for JUnit 5 support with Spring

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -219,14 +219,12 @@ project ('spring-kafka-test') {
 		compile "org.apache.kafka:kafka_$scalaVersion:$kafkaVersion"
 		compile "org.apache.kafka:kafka_$scalaVersion:$kafkaVersion:test"
 
-		compile ("junit:junit:$junit4Version") {
-			exclude group: 'org.hamcrest', module: 'hamcrest-core'
-		}
-		compile ("org.mockito:mockito-core:$mockitoVersion") {
-			exclude group: 'org.hamcrest', module: 'hamcrest-core'
-		}
-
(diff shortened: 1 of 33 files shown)


artembilan:
I marked it with the `DO NOT MERGE YET`, because we have chosen an unfortunate name for the JUnit 4 rule.
So, I have decided to make some breaking change in that part.
Therefore this might be considered for only `3.0` version.
Or... we can close our eyes and ship it with the current `2.2`: it is just a testing utility and there is no too much code change in the target tests though...

artembilan:
@garyrussell ,

Thank you for your feedback!
Please, take a look: I've tried to follow your suggestions.

artembilan:
FYI, this can be reviewed for a while already.
Any feedback welcome!

artembilan:
Pushed Docs. 
Thanks for the reminder!

garyrussell:
Minor polishing for your review: https://github.com/garyrussell/spring-kafka/commit/0edaaaf0db748a4991bd80ca438a9a336cc2535c

garyrussell:
Merged as 63472ce3dccb53341d488977913a26087726ae34

PR #760, opened 2018-08-02
Add wrapper task

izeye:
This PR adds `wrapper` task for easy Gradle Wrapper upgrades.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -359,3 +359,7 @@ task dist(dependsOn: assemble) {
 	group = 'Distribution'
 	description = 'Builds -dist, -docs distribution archives.'
 }
+
+wrapper {
+	gradleVersion = '4.9'
+}


build.gradle
```diff
@@ -359,3 +359,7 @@ task dist(dependsOn: assemble) {
 	group = 'Distribution'
 	description = 'Builds -dist, -docs distribution archives.'
 }
+
+wrapper {
+	gradleVersion = '4.9'
```
artembilan: We don't need this.
What we do is like:
1. we go into the `/gradle/wrapper` directory
2. open `gradle-wrapper.properties`
3. Modify the version in the `distributionUrl` property
4. Call `gradlew wrapper` task

That's all. No need in anything else. Also this way we keep Gradle version only in a single place.

garyrussell:
Actually, I do 1, 2, 3 with

    ./gradlew wrapper --gradle-version=4.9 --distribution-type=bin

then

    ./gradlew wrapper

artembilan:
Well, it is going to be much easier for you, Gary: the `distribution-type` is a `bin` by default 😉 

https://docs.gradle.org/current/userguide/gradle_wrapper.html

I'm terrible with typing in the CLI, so that is easier for me to edit file from IDE 😄

izeye:
@artembilan @garyrussell Thanks for the quick feedback!

With the `wrapper` task in place, I usually upgrades Gradle Wrapper as follows:

- Modify the version in the `wrapper` task.
- Run `./gradlew wrapper`.
- Run `./gradlew clean check` to ensure the upgraded Gradle Wrapper works.

I just thought this is easy as there's nothing to memorize but feel free to close this if you disagree.

artembilan:
Well, the problem with upgrade that we will have several places where things for the version are going to be changed in the commit.
Therefore I decided to go a simpler way. I mean the `gradle-wrapper.properties` are going to be changed anyway. So, what is the point to change something else yet?

Thanks for your feedback.

Closing as `Works as Designed`

izeye:
@artembilan I usually consider the `./gradle` directory as generated stuff via the `wrapper` task but I got your point without the task. Thanks for the feedback!

Issue #763, opened 2018-08-03
Add Stateful AfterRollbackProcessor


garyrussell:
Don't re-seek the failing record after some number of failures and provide some strategy that is invoked after retries are exhausted.

PR #764, opened 2018-08-03
GH-763: DefaultAfterRollbackProcessor, SeekToCurrrentErrorHandler: recovery

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/763

Add recovery capability to the default ARP.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AfterRollbackProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AfterRollbackProcessor.java
@@ -40,10 +40,25 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 public interface AfterRollbackProcessor<K, V> {
 
 	/**
-	 * Process the remaining records.
+	 * Process the remaining records. Recoverable will be true if the container is
+	 * processing individual records; this allows the processor to recover (skip) the
+	 * failed record rather than re-seeking it. This is not possible with a batch listener
+	 * since only the listener itself knows which record in the batch keeps failing.
 	 * @param records the records.
 	 * @param consumer the consumer.
+	 * @param exception the exception
+	 * @param recoverable the recoverable.
+	 * @since 2.2
 	 */
-	void process(List<ConsumerRecord<K, V>> records, Consumer<K, V> consumer);
+	void process(List<ConsumerRecord<K, V>> records, Consumer<K, V> consumer, Exception exception, boolean recoverable);
+
+	/**
+	 * Optional method to clear thread state; will be called just before a consumer
+	 * thread terminates.
+	 * @since 2.2
+	 */
+	default void clearThreadState() {
+		// NOSONAR
+	}
 
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
@@ -30,7 +30,10 @@ import org.springframework.lang.Nullable;
 /**
  * Default implementation of {@link AfterRollbackProcessor}. Seeks all
  * topic/partitions so the records will be re-fetched, including the failed
- * record.
+ * record. Starting with version 2.2 after a configurable number of failures
+ * for the same topic/partition/offset, that record will be skipped after
+ * calling a {@link BiConsumer} recoverer. The default recoverer simply logs
+ * the failed record.
  *
  * @param <K> the key type.
  * @param <V> the value type.
@@ -74,7 +77,8 @@ public class DefaultAfterRollbackProcessor<K, V> implements AfterRollbackProcess
 	 * @param maxFailures the maxFailures.
 	 * @since 2.2
 	 */
-	public DefaultAfterRollbackProcessor(@Nullable BiConsumer<ConsumerRecord<?, ?>, Exception> recoverer, int maxFailures) {
+	public DefaultAfterRollbackProcessor(@Nullable BiConsumer<ConsumerRecord<?, ?>, Exception> recoverer,
+			int maxFailures) {
 		this.failureTracker = new FailedRecordTracker(recoverer, maxFailures, logger);
 	}
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
@@ -19,12 +19,16 @@ package org.springframework.kafka.listener;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.BiConsumer;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.common.TopicPartition;
+
+import org.springframework.lang.Nullable;
 
 /**
  * Default implementation of {@link AfterRollbackProcessor}. Seeks all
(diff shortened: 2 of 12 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
```diff
@@ -41,13 +45,68 @@
...
+	/**
+	 * Construct an instance with the provided recoverer which will be called after
+	 * maxFailures have occurred for a topic/partition/offset.
+	 * @param recoverer the recoverer; if null, the default (logging) recoverer is used.
+	 * @param maxFailures the maxFailures.
+	 * @since 2.2
+	 */
+	public DefaultAfterRollbackProcessor(@Nullable BiConsumer<ConsumerRecord<K, V>, Exception> recoverer, int maxFailures) {
+		if (recoverer == null) {
+			this.recoverer = (r, t) -> logger.error("Max failures (" + maxFailures + ") reached for:" + r, t);
```
artembilan: We need a whitespace after colon.

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+public class FailedRecordTracker<K, V> {
+
+	private final ThreadLocal<FailedRecord> failures = new ThreadLocal<>(); // intentionally not static
+
+	private final BiConsumer<ConsumerRecord<K, V>, Exception> recoverer;
+
+	private final int maxFailures;
+
+	public FailedRecordTracker(BiConsumer<ConsumerRecord<K, V>, Exception> recoverer, int maxFailures, Log logger) {
```
artembilan: `@Nullable` ?

Also: does this component make sense to be `public`?

spring-kafka/src/main/java/org/springframework/kafka/listener/SeekUtils.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.kafka.common.TopicPartition;
+
+/**
+ * Seek utilities.
+ *
+ * @author Gary Russell
+ * @since 2.2
+ *
+ */
+public final class SeekUtils {
```
artembilan: Can't this class go to `support` package?

spring-kafka/src/main/java/org/springframework/kafka/support/SeekUtils.java
```diff
@@ -0,0 +1,108 @@
...
+	/**
+	 * Seek records to earliest position, optionally skipping the first.
+	 * @param records the records.
+	 * @param consumer the consumer.
+	 * @param exception the exception
+	 * @param recoverable true if skipping the first record is allowed.
+	 * @param skipper function to determine whether or not to skip seeking the first.
+	 * @param logger a {@link Log} for seek errors.
+	 */
+	public static void doSeeks(List<ConsumerRecord<?, ?>> records, Consumer<?, ?> consumer, Exception exception,
```
garyrussell: I don't really like having 2 methods here; the problem is one client has generic types, the other doesn't.

One solution would be to remove the generic types from `AfterRollbackPostProcessor` but I wonder if that's too much of a breaking change? Perhaps not.

WDYT?

spring-kafka/src/main/java/org/springframework/kafka/listener/SeekToCurrentErrorHandler.java
```diff
@@ -37,14 +39,52 @@
...
+	}
+
 	@Override
 	public void handle(Exception thrownException, List<ConsumerRecord<?, ?>> records,
 			Consumer<?, ?> consumer, MessageListenerContainer container) {
-		Map<TopicPartition, Long> offsets = new LinkedHashMap<>();
-		records.forEach(r ->
-			offsets.computeIfAbsent(new TopicPartition(r.topic(), r.partition()), k -> r.offset()));
-		offsets.forEach(consumer::seek);
+		SeekUtils.doSeeks(records, consumer, thrownException, true, this.failureTracker::skip, logger);
```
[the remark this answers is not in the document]
garyrussell: It would be a big change to add generics to error handlers; I found a solution.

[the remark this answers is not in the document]
artembilan: Well, I meant only one which suffers of such a generic util, but I’m you got a solution, so never mind.

spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
```diff
@@ -43,19 +45,50 @@
...
+	 * @since 2.2
+	 */
+	public DefaultAfterRollbackProcessor(@Nullable BiConsumer<ConsumerRecord<?, ?>, Exception> recoverer, int maxFailures) {
+		this.failureTracker = new FailedRecordTracker(recoverer, maxFailures, logger);
+	}
+
+	@Override
+	public void process(List<ConsumerRecord<K, V>> records, Consumer<K, V> consumer, Exception exception,
+			boolean recoverable) {
+		SeekUtils.doSeeks(records.stream().collect(Collectors.toList()),
```
artembilan: Well, if you prefer this, let it be. I other place I do something like "double cast" trick:
```
(List<?, ?>)(List) records
```

😄

garyrussell: Ha - that works - but STS removes the second cast for me on save. Pushed.

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Track record processing failure counts.
+ *
+ * @author Gary Russell
+ * @since 2.2
+ *
+ */
+class FailedRecordTracker {
+
+	private final ThreadLocal<FailedRecord> failures = new ThreadLocal<>(); // intentionally not static
```
artembilan: Gary, remind me, please, when we need it to be `static` and when not.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
```diff
@@ -0,0 +1,106 @@
...
+		else {
+			this.recoverer = recoverer;
+		}
+		this.maxFailures = maxFailures;
+	}
+
+	boolean skip(ConsumerRecord<?, ?> record, Exception exception) {
+		FailedRecord failedRecord = this.failures.get();
+		if (failedRecord == null || !failedRecord.getTopic().equals(record.topic())
+				|| failedRecord.getPartition() != record.partition() || failedRecord.getOffset() != record.offset()) {
```
artembilan: Can't we have `equals()` implementation in the `FailedRecord` instead?

spring-kafka/src/main/java/org/springframework/kafka/listener/AfterRollbackProcessor.java
```diff
@@ -40,10 +40,25 @@
...
+	 * failed record rather than re-seeking it. This is not possible with a batch listener
+	 * since only the listener itself knows which record in the batch keeps failing.
 	 * @param records the records.
 	 * @param consumer the consumer.
+	 * @param exception the exception
+	 * @param recoverable the recoverable.
+	 * @since 2.2
 	 */
-	void process(List<ConsumerRecord<K, V>> records, Consumer<K, V> consumer);
+	void process(List<ConsumerRecord<K, V>> records, Consumer<K, V> consumer, Exception exception, boolean recoverable);
```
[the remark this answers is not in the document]
garyrussell: I don't think it's a big breaking change - as long as we add it to the migration guide - if I do the `default` dance, it can no longer be a `@FunctionalInterface`, so that would be a breaking change too.

I think custom implementations will be rare.

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
garyrussell: Per thread Vs. Per thread/instance: https://stackoverflow.com/questions/2784009/why-should-java-threadlocal-variables-be-static

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
garyrussell: That would required creating a new instance even if we don't need one.

spring-kafka/src/main/java/org/springframework/kafka/listener/AfterRollbackProcessor.java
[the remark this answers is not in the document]
garyrussell: If you have strong objections, I suppose I could create a new sub-interface.

Oops - I see you merged already.

Issue #769, opened 2018-08-05
Add Type Mapping Info Properties for Json(De)Serializer


garyrussell:
e.g. `spring.json.type.mapping=foo:com.foo.Foo,bar:com.foo.Bar`

- Outbound: class names mapped to keys
- Inbound: keys mapped to class names

Issue #768, opened 2018-08-05
Support Type Precedence in the JSON Deserializer


garyrussell:
Allow the use of target type over headers. See https://stackoverflow.com/questions/51688924/spring-kafka-the-class-is-not-in-the-trusted-packages/51695630#comment90356211_51695630

PR #770, opened 2018-08-06
GH-769: Add JSON Type Mapping Property

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/769

Allow configuration of type mapping using Kafka properties.

Resolves https://github.com/spring-projects/spring-kafka/issues/768

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
@@ -164,9 +164,6 @@ public class JsonDeserializer<T> implements ExtendedDeserializer<T> {
 			if (this.targetType != null) {
 				this.reader = this.objectMapper.readerFor(this.targetType);
 			}
-			else {
-				this.reader = null;
-			}
 		}
 		else {
 			this.targetType = targetType;
@@ -215,11 +212,9 @@ public class JsonDeserializer<T> implements ExtendedDeserializer<T> {
 			if (isKey && configs.containsKey(KEY_DEFAULT_TYPE)) {
 				if (configs.get(KEY_DEFAULT_TYPE) instanceof Class) {
 					this.targetType = (Class<T>) configs.get(KEY_DEFAULT_TYPE);
-					this.reader = this.objectMapper.readerFor(this.targetType);
 				}
 				else if (configs.get(KEY_DEFAULT_TYPE) instanceof String) {
 					this.targetType = (Class<T>) ClassUtils.forName((String) configs.get(KEY_DEFAULT_TYPE), null);
-					this.reader = this.objectMapper.readerFor(this.targetType);
 				}
 				else {
 					throw new IllegalStateException(KEY_DEFAULT_TYPE + " must be Class or String");
@@ -229,12 +224,10 @@ public class JsonDeserializer<T> implements ExtendedDeserializer<T> {
 			else if (!isKey && configs.containsKey("spring.json.default.value.type")) {
 				if (configs.get("spring.json.default.value.type") instanceof Class) {
 					this.targetType = (Class<T>) configs.get("spring.json.default.value.type");
-					this.reader = this.objectMapper.readerFor(this.targetType);
 				}
 				else if (configs.get("spring.json.default.value.type") instanceof String) {
 					this.targetType = (Class<T>) ClassUtils
 						.forName((String) configs.get("spring.json.default.value.type"), null);
-					this.reader = this.objectMapper.readerFor(this.targetType);
 				}
 				else {
 					throw new IllegalStateException("spring.json.default.value.type must be Class or String");
@@ -252,6 +245,9 @@ public class JsonDeserializer<T> implements ExtendedDeserializer<T> {
 				else {
 					throw new IllegalStateException(VALUE_DEFAULT_TYPE + " must be Class or String");
 				}
+			}
+			if (this.targetType != null) {
+				this.reader = this.objectMapper.readerFor(this.targetType);
 			}
 			addTargetPackageToTrusted();
 		}
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
@@ -103,7 +103,7 @@ public class JsonDeserializer<T> implements ExtendedDeserializer<T> {
 	 * Construct an instance with a default {@link ObjectMapper}.
 	 */
 	public JsonDeserializer() {
-		this((Class<T>) null, true);
+		this(null, true);
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -91,29 +99,74 @@
...
+	 * @param targetType the target type.
+	 * @param objectMapper the mapper.
+	 * @param useHeadersIfPresent true to use headers if present and fall back to target
+	 * type if not.
+	 * @since 2.2
+	 */
+	@SuppressWarnings("unchecked")
+	public JsonDeserializer(@Nullable Class<T> targetType, ObjectMapper objectMapper, boolean useHeadersIfPresent) {
 		Assert.notNull(objectMapper, "'objectMapper' must not be null.");
+		Assert.isTrue(targetType != null || useHeadersIfPresent,
```
artembilan: This one has to go after `resolveGeneric(0)` a bit bellow.
It can be `null` here, but we may resolve a target type from the generics.

WDYT?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -91,29 +99,74 @@
 
 	private boolean typeMapperExplicitlySet = false;
 
+	/**
+	 * Construct an instance with a default {@link ObjectMapper}.
+	 */
 	public JsonDeserializer() {
-		this((Class<T>) null);
+		this((Class<T>) null, true);
```
artembilan: We don't need this cast anymore

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -196,6 +249,10 @@ else if (configs.get(VALUE_DEFAULT_TYPE) instanceof String) {
 						StringUtils.commaDelimitedListToStringArray((String) configs.get(TRUSTED_PACKAGES)));
 			}
 		}
+		if (configs.containsKey(TYPE_MAPPINGS)) {
```
artembilan: The `JsonSerializer` has more condition in similar place.
Why is it missed here?

Thanks

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -218,19 +275,26 @@ public T deserialize(String topic, Headers headers, byte[] data) {
...
+			if (javaType != null) {
+				reader = this.objectMapper.readerFor(javaType);
 			}
-			catch (IOException e) {
-				throw new SerializationException("Can't deserialize data [" + Arrays.toString(data) +
-						"] from topic [" + topic + "]", e);
+		}
+		if (reader == null && this.targetType != null) {
+			if (this.reader == null) {
+				this.reader = this.objectMapper.readerFor(this.targetType);
```
artembilan: Why this can't be initialized only one in the ctor?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerializer.java
```diff
@@ -126,6 +135,27 @@ else if (config instanceof String) {
 				throw new IllegalStateException(ADD_TYPE_INFO_HEADERS + " must be Boolean or String");
 			}
 		}
+		if (configs.containsKey(TYPE_MAPPINGS) && !this.typeMapperExplicitlySet && this.typeMapper instanceof AbstractJavaTypeMapper) {
+			((AbstractJavaTypeMapper) this.typeMapper)
+					.setIdClassMapping(createMappings((String) configs.get(TYPE_MAPPINGS)));
+		}
+	}
+
+	public static Map<String, Class<?>> createMappings(String mappings) {
```
artembilan: I think maximum this method must be `protected`.

src/reference/asciidoc/kafka.adoc
```diff
@@ -1610,9 +1610,42 @@ Starting with _version 2.1_, type information can be conveyed in record `Headers
 In addition, the serializer/deserializer can be configured using Kafka properties.
 
 - `JsonSerializer.ADD_TYPE_INFO_HEADERS` (default `true`); set to `false` to disable this feature on the `JsonSerializer` (sets the `addTypeInfo` property).
+- `JsonSerializer.TYPE_MAPPINGS` (default `empty`); see below
```
artembilan: Maybe we should add period in the end of sentence?
At least that is what I see in other bullets here...

src/reference/asciidoc/kafka.adoc
```diff
@@ -1610,9 +1610,42 @@ Starting with _version 2.1_, type information can be conveyed in record `Headers
...
+senderProps.put(JsonSerializer.TYPE_MAPPINGS, "foo:com.myfoo.Foo, bar:com.mybar.bar");
+...
+consumerProps.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, JsonDeserializer.class);
+consumerProps.put(JsonDeSerializer.TYPE_MAPPINGS, "foo:com.yourfoo.Foo, bar:com.yourbar.bar");
+----
+====
+
+Of course, the corresponding objects must be compatible.
+
+When using https://docs.spring.io/spring-boot/docs/2.0.4.RELEASE/reference/htmlsingle/#boot-features-kafka-extra-props[Spring Boot], these properties can be provided in the `application.properties` (or yaml) file:
```
artembilan: I would say that this link would be better: https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-messaging.html#boot-features-kafka

First of all it is version-agnostic. Another its feature it is pointing directly the Kafka page there and doesn't load the whole Boot reference, which is pretty big and therefore slow.

WDYT?

src/reference/asciidoc/kafka.adoc
```diff
@@ -1610,9 +1610,42 @@ Starting with _version 2.1_, type information can be conveyed in record `Headers
...
+
+====
+[source]
+----
+spring.kafka.producer.value-serializer=org.springframework.kafka.support.serializer.JsonSerializer
+spring.kafka.producer.properties.spring.json.type.mapping=foo:com.myfoo.Foo,bar=com.mybar.Bar
```
artembilan: You say they have to be `colon-separate`, but here you use `=` for `bar`. I think this is typo.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
garyrussell: Good catch

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -158,13 +158,22 @@ public JsonDeserializer(Class<T> targetType, ObjectMapper objectMapper) {
...
+			if (this.targetType != null) {
+				this.reader = this.objectMapper.readerFor(this.targetType);
+			}
+			else {
+				this.reader = null;
+			}
+		}
+		else {
+			this.targetType = targetType;
+			this.reader = this.objectMapper.readerFor(this.targetType);
```
artembilan: I think we can have this line of code only once: after that `Assert.isTrue()` and within `if...else` around `this.targetType`.
Just a nit-pick do keep DRY principle...

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -206,9 +215,11 @@ public void configure(Map<String, ?> configs, boolean isKey) {
 			if (isKey && configs.containsKey(KEY_DEFAULT_TYPE)) {
 				if (configs.get(KEY_DEFAULT_TYPE) instanceof Class) {
 					this.targetType = (Class<T>) configs.get(KEY_DEFAULT_TYPE);
+					this.reader = this.objectMapper.readerFor(this.targetType);
```
artembilan: Hm. Looks like we don't need `this.reader = null;` in the ctor at all! 😃

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -230,9 +243,11 @@ else if (configs.get("spring.json.default.value.type") instanceof String) {
 			else if (!isKey && configs.containsKey(VALUE_DEFAULT_TYPE)) {
 				if (configs.get(VALUE_DEFAULT_TYPE) instanceof Class) {
 					this.targetType = (Class<T>) configs.get(VALUE_DEFAULT_TYPE);
+					this.reader = this.objectMapper.readerFor(this.targetType);
 				}
 				else if (configs.get(VALUE_DEFAULT_TYPE) instanceof String) {
 					this.targetType = (Class<T>) ClassUtils.forName((String) configs.get(VALUE_DEFAULT_TYPE), null);
+					this.reader = this.objectMapper.readerFor(this.targetType);
```
artembilan: Hm. Can't all these `readerFor()` calls go to the end of this method, e.g. before `addTargetPackageToTrusted();` ?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
garyrussell: Yeah - I had it final but tests failed because of this mode.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/9ae454a86ebcff5c427693512074135d5f9dc996

PR #776, opened 2018-08-09
Add KafkaStreamsConfiguration

garyrussell:
See https://github.com/spring-projects/spring-boot/pull/14021#discussion_r208929624

Avoid using a bean of type `Properties`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/KafkaStreamsConfiguration.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/KafkaStreamsConfiguration.java
@@ -26,8 +26,7 @@ import org.apache.kafka.streams.StreamsBuilder;
 import org.springframework.util.Assert;
 
 /**
- * Wrapper for {@link StreamsBuilder} properties; used to avoid auto-configuring
- * a generic {@link Properties} object in Boot.
+ * Wrapper for {@link StreamsBuilder} properties.
  *
  * @author Gary Russell
  * @since 2.2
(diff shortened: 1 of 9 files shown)


artembilan:
Well, with this I believe we need to change Spring Integration global properties from the `Properties` bean as well...

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaStreamsConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Map;
+import java.util.Properties;
+
+import org.apache.kafka.streams.StreamsBuilder;
+
+import org.springframework.util.Assert;
+
+/**
+ * Wrapper for {@link StreamsBuilder} properties; used to avoid auto-configuring
+ * a generic {@link Properties} object in Boot.
```
artembilan: I don't think this is a good idea to mentioned Spring Boot here.
More over in its non-official name.

I believe we can find some generic message to say here.

Thanks

garyrussell:
>Well, with this I believe we need to change Spring Integration global properties from the Properties bean as well...

Yes, then I guess we can expose the props as boot auto config properties.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaStreamsConfiguration.java
garyrussell: Good catch on the missed ctor.

Issue #777, opened 2018-08-14
DefaultKafkaProducerFactory.stop() doesn't change running state


quaff:
https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java#L201

stop method should change running to false, it's better to use `AtomicBoolean.compareAndSet()` instead of `volatile boolean`

quaff:
I have created PR#778

PR #778, opened 2018-08-14
GH-777: Change DefaultKafkaProducerFactory running state to false when stop it

quaff:


spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -92,7 +93,7 @@
 
 	private String transactionIdPrefix;
 
-	private volatile boolean running;
+	private AtomicBoolean running = new AtomicBoolean(true);
```
artembilan: This must be `final` and it can't be `true` by default. What if I'm not going to start this component?
Also this is a behavior change.

Please, reconsider this logic for proper expectations.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -193,28 +194,32 @@ public void destroy() throws Exception { //NOSONAR
 
 	@Override
 	public void start() {
-		this.running = true;
+		this.running.compareAndSet(false, true);
```
artembilan: I believe just `this.running.set(true);` is fully enough here. Looks like logic in this method is fully idempotent. Well, just because we do nothing more 😄

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -193,28 +194,32 @@ public void destroy() throws Exception { //NOSONAR
...
 
 	@Override
 	public boolean isRunning() {
-		return this.running;
+		return this.running.get();
 	}
 
 	@Override
 	public Producer<K, V> createProducer() {
+		if (!isRunning())
```
artembilan: I don't think we need it here. Even if we don't start this component, we are going to clean up all the states in the `destroy()`, which is only a logic in the `stop()` implementation above.

I even wonder from here if we need the `Lifecycle` implementation here at all...

I don't tell that we need to change the logic right now, but let's figure out if it is obsolete at all!

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
quaff: I set it with `true` by default because running state checking added in `createProducer()`, make it auto started not throwing exception at first time, keep behavior consistency.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
quaff: I'm also confused why it implemented `Lifecycle`, and if we implemented it, we must check running state before any operation.

garyrussell:
The reason we need `Lifecycle` rather than a method such as `resetProducers()` is because the application context can be stopped and started independently of being destroyed/closed and we need to close the `Producer` (s) whenever the container is stopped.

artembilan:
Gary, can't we then rely on the `ContextStoppedEvent` instead? I really don't see any reason to maintain the whole `Lifecycle` contract since we don't get it right, plus we are in doubt where we have to fully support it.

Thanks

garyrussell:
Sure, that's another option, but it seems more complex, and I really don't see what the big deal is.

But, I don't mind either way.

artembilan:
Right. That may make sense, too.
So, we just need to remove a `running` property and comment in the body of `start()` and `stop()` what and how is going on there.
This indeed is going to be cleaner solution after your explanation of the reason! 😄 
Although robust one would be really based on the `ContextStoppedEvent` listener impl...

And as I said before: there is nothing to back-port since what we are chasing here is only clear code and no bugs to fix.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
quaff: I agree to remove `Lifecycle`

PR #785, opened 2018-08-20
Polish KafkaTemplate.executeInTransaction()

izeye:
This PR polishes `KafkaTemplate.executeInTransaction()`.


artembilan:
Can we have more explanation for the reason of change?
Thanks

garyrussell:
Can you explain the rationale for this?

With this change, we'll no longer close the producer or remove it from `this.producers`.

izeye:
@artembilan @garyrussell Thanks for the quick feedback! I just collapsed the `finally` clause for brevity as I thought it's equivalent with the current implementation. Am I missing something?

garyrussell:
The `abortTransaction` after an exception needs to be in a try/catch too.

garyrussell:
Doh - I see - the `finally` handles it.

PR #803, opened 2018-09-11
Fix Package Tangle

garyrussell:
The introduction of `KafkaStreamsConfiguration` introduced a tangle
between `core` and `config`.

Move the `StreamsBuilderFactoryBean` to `config`.


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaStreamsCustomizer.java
```diff
@@ -18,6 +18,8 @@
 
 import org.apache.kafka.streams.KafkaStreams;
 
+import org.springframework.kafka.config.StreamsBuilderFactoryBean;
```
artembilan: I would say this `KafkaStreamsCustomizer` is part of the configuration as well.
Feels organic to move it into the `config` package as well.

spring-kafka/src/test/java/org/springframework/kafka/core/KafkaStreamsCustomizerTests.java
```diff
@@ -33,6 +33,7 @@
 import org.springframework.kafka.annotation.EnableKafkaStreams;
 import org.springframework.kafka.annotation.KafkaStreamsDefaultConfiguration;
 import org.springframework.kafka.config.KafkaStreamsConfiguration;
+import org.springframework.kafka.config.StreamsBuilderFactoryBean;
```
artembilan: It's not so critical, but may be would be better to move relevant tests to the `config` package, too.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaStreamsCustomizer.java
garyrussell: Agreed; it crossed my mind too. Will do.

PR #818, opened 2018-10-02
KafkaStreamsConfiguration Conversion

garyrussell:
Add `ConversionService` to convert map values to String.
Also use `Class.getName()`.


garyrussell:
Do not merge. Must not mutate shared instance

Issue #813, opened 2018-09-26
Rebalancing while KafkaMessageListenerContainer is paused prevent it from being paused again after the rebalance


jffou17:
Kafka spring version : 2.1.5.RELEASSE
Kafka client version : 1.0.0

If a rebalance occurs while a KafkaMessageListenerContainer is paused, consumption of records will resume and it will not be possible to pause the container again.

This is because the private variable "consumerPaused" is not reinitialised during a rebalance. It while be stuck on "true" even though the consumer is not paused anymore. From this moment on, it will not be possible to pause the container again and records will be fetched/consumed on each poll. To pause the container again you have to resume it, let it poll once, then you can pause it again. 

This can be reproduce with a KafkaMessageListenerContainer configured as follows :

```
ContainerProperties containerProperties = new ContainerProperties("topic-test");
containerProperties.setGroupId("group-test");
containerProperties.setAckMode(AbstractMessageListenerContainer.AckMode.MANUAL_IMMEDIATE);
containerProperties.setMessageListener(new TestListener());
containerProperties.setErrorHandler(new TestErrorHandler());
KafkaMessageListenerContainer<K, V> messageListenerContainer = new KafkaMessageListenerContainer<>(consumerFactory, containerProperties);

```

```
public class TestListener implements AcknowledgingMessageListener<String, String> {

    private static final Logger log = LoggerFactory.getLogger(TestListener.class);

    @Override
    public void onMessage(ConsumerRecord<String, String> record, Acknowledgment ack) {
        log.info("Consuming " + record.partition() + ":" + record.offset());

        throw new IllegalStateException("Test");
    }
}
```

```
public class TestErrorHandler implements ContainerAwareErrorHandler {

    private static final Logger log = LoggerFactory.getLogger(TestErrorHandler.class);

    @Override
    public void handle(Exception thrownException, List<ConsumerRecord<?, ?>> records, Consumer<?, ?> consumer, MessageListenerContainer container) {

        seekToCurrent(records, consumer);

        container.pause();

        log.info("Consumer paused in Error handler");
    }

    private void seekToCurrent(List<ConsumerRecord<?, ?>> records, Consumer<?, ?> consumer) {
        Map<TopicPartition, Long> offsets = new LinkedHashMap<>();
        records.forEach(r -> offsets.computeIfAbsent(new TopicPartition(r.topic(), r.partition()), k -> r.offset()));
        offsets.forEach(consumer::seek);
    }
}
```

[...]

PR #823, opened 2018-10-11
GH-813: Re-pause a paused consumer after rebalance

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/813

Kafka resumes the consumer(s) after a rebalance, but since the
`ListenerConsumer.consumerPaused` is still true, the container starts
consuming again and can't be paused without first issuing a resume.

On a rebalance, reset the boolean; the consumer will re-pause  before
the next poll, if the container's `isPaused()` is still true after
the rebalance listeners exit.

Tested with a stand-alone Boot app (see the referenced issue).

**cherry-pick to 2.1.x**


artembilan:

PR #825, opened 2018-10-12
@KafkaListener Add BytesToStringConverter

garyrussell:
Currently a `byte[]` gets converted to a String using an
`ArrayToStringConverter`, which is not very useful.

Add a `BytesToStringConverter`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -17,6 +17,8 @@
 package org.springframework.kafka.annotation;
 
 import java.lang.reflect.Method;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -204,6 +208,20 @@ public void setBeanFactory(BeanFactory beanFactory) {
...
+	 * Default UTF-8.
+	 * @param charset the charset.
+	 * @since 2.2
+	 */
+	public void setCharset(Charset charset) {
+		Assert.notNull(charset, "'charset' cannot be null");
+		this.charset = charset;
+	}
+
+	private Charset getCharset() {
```
artembilan: What is the purpose of this `private` method?
I don't say that it is harmful somehow, but it is not useful as well...

garyrussell: Artifact of an earlier iteration - will remove.

Issue #811, opened 2018-09-21
The headers are not sent back by @SendTo 


heerdev:
For https://docs.spring.io/spring-kafka/reference/html/_reference.html#annotation-send-to
the @SendTo, does not seems to send back the header it received except correlation_id.
We can see this on Kafka Broker, the message key is empty for the reply message.
We can send our custom header , if the reply type of the method is Message<T>.
In this case, consumer will decide what header he wants to send to producer.
But , as per replyTemplate, the producer should be able to decide what headers it set.

artembilan:
@heerdev ,

What do you suggest has to be as a fix for this issue?
Any chances that we can see a Pull Request from you on the matter?

Thanks

garyrussell:
I was thinking we can add a `Predicate<String> shouldCopyHeader` to the listener adapter/container factory.

PR #826, opened 2018-10-15
GH-811: Configurable Reply Headers

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/811

Support copying and/or modifying reply headers.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
@@ -25,6 +25,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -54,11 +55,13 @@ import org.springframework.kafka.support.converter.MessagingMessageConverter;
 import org.springframework.kafka.support.converter.RecordMessageConverter;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
 import org.springframework.messaging.MessagingException;
 import org.springframework.messaging.converter.MessageConversionException;
 import org.springframework.messaging.handler.annotation.Payload;
 import org.springframework.messaging.support.MessageBuilder;
 import org.springframework.util.Assert;
+import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 
 /**
@@ -107,6 +110,8 @@ public abstract class MessagingMessageListenerAdapter<K, V> implements ConsumerS
 	private boolean hasAckParameter;
 
 	private boolean messageReturnType;
+
+	private ReplyHeadersConfigurer replyHeadersConfigurer;
 
 	public MessagingMessageListenerAdapter(Object bean, Method method) {
 		this.bean = bean;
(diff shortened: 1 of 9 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -384,6 +408,24 @@ else if (result instanceof Message) {
...
+										&& !key.startsWith(KafkaHeaders.RECEIVED);
+							})
+							.filter(e -> this.replyHeadersConfigurer.shouldCopy(e.getKey(), e.getValue()))
+							.collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()));
+						if (headersToCopy.size() > 0) {
+							builder.copyHeaders(headersToCopy);
+						}
+						headersToCopy = this.replyHeadersConfigurer.additionalHeaders();
+						if (!ObjectUtils.isEmpty(headersToCopy)) {
+							builder.copyHeaders(headersToCopy);
```
artembilan: maybe `copyHeadersIfAbsent()` ?

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaders.java
```diff
@@ -26,7 +26,15 @@
...
-	private static final String PREFIX = "kafka_";
+	/**
+	 * The prefix for Kafka headers.
+	 */
+	public static final String PREFIX = "kafka_";
+
+	/**
+	 * The prefix for Kafka headers containing 'received' values.
+	 */
+	public static final String RECEIVED = PREFIX + "received";
```
artembilan: Shouldn't we have an underscore in the end of this prefix as well?

garyrussell: That would change the header values.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
garyrussell: I don't think so; this is a new `MessageBuilder` so there won't be any existing headers - except for those that passed the filter above (plus the correlation id and topic).

What concerns you?

artembilan: Exactly! That’s my point that end-user May shot his/her own leg when the same header passed the filter and it is in the extra. Maybe just add a JavaDoc note that these additional always win?..

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaHeaders.java
artembilan: Doh. Now I see that this is just a refactoring for existing code without anything semantically new.

Thanks and never mind

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ReplyHeadersConfigurer.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ */
+@FunctionalInterface
+public interface ReplyHeadersConfigurer {
+
+	/**
+	 * Return true if the header should be copied to the reply message.
+	 * {@link KafkaHeaders#CORRELATION_ID} will not be offered; it is always copied.
+	 * {@link MessageHeaders#ID} and {@link MessageHeaders#TIMESTAMP} are never copied.
+	 * {@code KafkaHeaders.RECEIVED_*} headers are never copied.
```
artembilan: I think this underscore is confusing here.

PR #828, opened 2018-10-18
Doc polishing, remove connectionFactory

neetkee:
Looks like connectionFactory from docs doesn't exist anymore


artembilan:

PR #833, opened 2018-10-25
Fix DEBUG error message about reply topics

garyrussell:
**cherry-pick to 2.1.x**


artembilan:
@garyrussell ,

Doesn't look like we have something about `replyTopic` in version `2.1.x`: https://github.com/spring-projects/spring-kafka/blob/2.1.x/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java#L86

What do we do?

Thanks

garyrussell:
Oops - right in 2.1, you always had to set the reply topic header, so no cherry-pick; sorry about that.

https://github.com/spring-projects/spring-kafka/commit/82322d4b5dd27583c56f07c528c19c1768db343e

Issue #843, opened 2018-10-30
ErrorHandlingDeserializer and types


garyrussell:
Should be `implements ExtendedDeserializer<? extends Object>`.

https://stackoverflow.com/questions/53026701/transactional-behavior-in-spring-kafka/53033273#comment93030711_53033273

PR #844, opened 2018-10-30
GH-843: Make `ErrorHandlingDeserializer` as generic one

artembilan:
Fixes https://github.com/spring-projects/spring-kafka/issues/843

* To ensure a good end-user experience with the `ErrorHandlingDeserializer`
make it as generic one.
* Implement a `private deserializationException()` method to overcome
a generic type during compilation with the `AtomicReference` wrapper
* Expect a regular `Deserializer` for delegate and wrap it into the
`ExtendedDeserializer` if necessary
* Ensure a proper logic in the `configure()` to instantiate an
`ExtendedDeserializer` properly
* Demonstrate a proper compilation with new `ErrorHandlingDeserializer`
generics trick in the `KafkaMessageListenerContainerTests`
* Polishing for the `KafkaMessageListenerContainerTests`


artembilan:
Also fixes: https://github.com/spring-projects/spring-kafka/issues/845

garyrussell:
Hehe - looks like we both worked on 845 - I have stopped work, but I did update the test case... https://github.com/garyrussell/spring-kafka/commit/6b07874c9cc218300e5ddd4b659f11578bd44626

artembilan:
😃 OK. I'll copy-paste your test here.

Thank you!

Woodham:
It's unfortunate that this merge didn't make it into the 2.2.0 release, as I think without the changes here the `ErrorHandlingDeserializer` can't be used to wrap non-spring deserializer (i.e. one that doesn't extend `ExtendedDeserializer`).

Is this likely to make it into a 2.2.1 release anytime soon?

Woodham:
Sorry I didn't mean non-spring, just non `ExtendedDeserializer` implementation.

garyrussell:
@Woodham We are planning on releasing 2.2.1 later this week; however, also see https://github.com/spring-projects/spring-kafka/pull/857 (which will also be included).

PR #851, opened 2018-11-02
Log TX Exceptions

garyrussell:
- Log exceptions on commit/abort transaction
- Don't attempt to abort if commit fails, otherwise we see

`Abort failed`
`org.apache.kafka.common.KafkaException: Cannot execute transactional method because we are in an error state`

**cherry-pick to 2.1.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -274,15 +274,15 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V> {
 		}
 
 		this.producers.set(producer);
-		boolean commitFailed = false;
+		boolean callbackSuccessful = false;
 		try {
 			T result = callback.doInOperations(this);
-			commitFailed = true;
+			callbackSuccessful = true;
 			producer.commitTransaction();
 			return result;
 		}
 		catch (Exception e) {
-			if (!commitFailed) {
+			if (!callbackSuccessful) {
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -274,13 +274,17 @@ public boolean isTransactional() {
 		}
 
 		this.producers.set(producer);
+		boolean commitFailed = false;
 		try {
 			T result = callback.doInOperations(this);
+			commitFailed = true;
```
artembilan: Huh?

The variable name somehow doesn't reflect reality.
Isn't there a special exception to catch after `producer.commitTransaction()` failure?

garyrussell: What do you mean; if we get an exception after this point, it's the commit that failed. I don't know all the possible exceptions from a commit.

artembilan: Yes, I see the logic and I agree about exceptions in the client.
Only what I don't feel comfort is the name of the variable. 
Since it is going to be `true` even if commit is not failed.

garyrussell: Seems ok to me 😄 

Would you prefer `callbackWasSuccessful = true` and `if (!callbackWasSuccessful)` ?

artembilan:

PR #856, opened 2018-11-05
Add Samples

garyrussell:


samples/sample-01/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<groupId>com.example</groupId>
+	<artifactId>kafka-sample-01</artifactId>
+	<version>0.0.1-SNAPSHOT</version>
+	<packaging>jar</packaging>
+
+	<name>confluentBlog</name>
```
artembilan: I don't think this is the right name for our samples...

samples/sample-01/src/test/java/com/example/ApplicationTests.java
```diff
@@ -0,0 +1,16 @@
+package com.example;
```
artembilan: Missed Copyright

Issue #853, opened 2018-11-05
Type safety with ErrorHandlingDeserializer


martin-mucha:
I think, that newly added ErrorHandlingDeserializer creates typesafety issues. Why? Because the correctly declared deserializer says, what type it returns, and then, ErrorHandlingDeserializer returns something else. Consider this code, and check gotcha in record filter and bizarre syntax just to be able to compile:

```
@Bean
public ConcurrentKafkaListenerContainerFactory<String, Predictor> avroKafkaListenerContainerFactory(
 String bootstrapServers,
 String groupId){

    ConcurrentKafkaListenerContainerFactory<String, Predictor> factory=
    new ConcurrentKafkaListenerContainerFactory<>();

    //noinspection unchecked    //spring-kafka currently has incorrectly defined types in ErrorHandlingDeserializer
    ConsumerFactory<String, Predictor> consumerFactory=new DefaultKafkaConsumerFactory(
    createKafkaConsumerFactoryProperties(bootstrapServers,groupId),
    new StringDeserializer(),
    new ErrorHandlingDeserializer(new AvroDeserializer(Predictor.class)));

    factory.setConsumerFactory(consumerFactory);
    factory.setBatchListener(true);

    //is always false? Well, not really. Since spring-kafka team breaks type safety.
    //noinspection ConstantConditions
    factory.setRecordFilterStrategy(consumerRecord->((Object)consumerRecord.value())instanceof DeserializationException);

    return factory;
    }
```

So what we actually need is deserializer which does not lie about return type. The only way how to correctly achieve that is to return Object, which can represent either valid or invalid result. This might be potentially intensive for garbage collection, so use could be also allowed to provide some constant for representation of errorneous deserialization, which can be then compared using == in recordfilter for example. So here is what I'd consider better approach. Please consider this as a suggestion of one way how to handle current issues.

parent (to allow simple code + chain of responsibility:

```
public abstract class NotConfigurableExtendedDeserializer<T> implements ExtendedDeserializer<T> {

    @Override
    public void configure(Map<String, ?> configs, boolean isKey) {
        throw new UnsupportedOperationException(
                "We do apologize, but this Deserializer does not support configuration via properties. " +
                        "This approach severely limits reuse and make decorator pattern approach impossible. " +
                        "Thus we're not following this design.");
    }

}
```

Typesafe error handler:
```
public abstract class ErrorObjectDeserializerWrapper<DELEGATETYPE, TRANSFORMEDTYPE>
[...]

PR #857, opened 2018-11-05
GH-853: Type Safe ErrorHandlingDeserializer

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/853

Since a `null` key is common, we only check for the exception header
if we detect that the error handling deserializer is configured.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -16,6 +16,9 @@
 
 package org.springframework.kafka.listener;
 
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
 import java.lang.reflect.Type;
 import java.time.Duration;
 import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1197,6 +1226,27 @@ private RuntimeException doInvokeRecordListener(final ConsumerRecord<K, V> recor
...
 
+		public void checkDeser(final ConsumerRecord<K, V> record, String headerName) {
+			Header header = record.headers().lastHeader(headerName);
+			if (header != null) {
+				try {
+					DeserializationException ex = (DeserializationException) new ObjectInputStream(
+							new ByteArrayInputStream(header.value())).readObject();
+					Headers headers = new RecordHeaders();
+					record.headers().forEach(h -> {
+						if (!h.key().startsWith(ErrorHandlingDeserializer2.KEY_DESERIALIZER_EXCEPTION_HEADER_PREFIX)) {
```
artembilan: Looks like you want to filter out the `ErrorHandlingDeserializer2.KEY_DESERIALIZER_EXCEPTION_HEADER_PREFIX`.
Can't we do such a logic with Java Stream?
Looks like `RecordHeaders` has a ctor like this `RecordHeaders(Iterable<Header> headers)` which can be used in the `collect(Collector<? super T, A, R> collector)` of the original stream.

garyrussell:
The reason the current solution doesn't work is with a batch listener.

```java
@KafkaListener(id = "deser", topics = "blc.2.1")
public void listen1(List<Foo> foos) {
    foos.forEach(...)
}
```

Crashes with a `ClassCastException` when it encounters the `DeserializationException`.

However, this still doesn't work with the new solutions because `null` is represented by a `KafkaNull` payload so we still get a CCE.

I have a new solution; pushing after my build passes.

Issue #855, opened 2018-11-05
DeadLetterPublishingRecoverer and transaction synchronization


garyrussell:
The `DeadLetterPublishingRecoverer` should not use `executeInTransaction` if there is an existing Kafka transaction.

PR #858, opened 2018-11-05
GH-855: Dead letter publisher and transactions

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/855

`DeadLetterPublishingRecoverer` should participate in the current transaction,
if present, or start a transaction otherwise (if the template is transactional).

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTransactionTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTransactionTests.java
@@ -56,6 +56,7 @@ import org.mockito.InOrder;
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.kafka.listener.DeadLetterPublishingRecoverer;
 import org.springframework.kafka.support.TransactionSupport;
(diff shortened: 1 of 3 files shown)


spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTransactionTests.java
```diff
@@ -269,6 +270,37 @@ public void testTransactionSynchronizationExceptionOnCommit() {
...
+			new DeadLetterPublishingRecoverer(template).accept(
+					new ConsumerRecord<>(STRING_KEY_TOPIC, 0, 0L, "key", "foo"),
+					new RuntimeException("foo"));
+			return null;
+		});
+
+		verify(producer1).beginTransaction();
+		verify(producer1).commitTransaction();
+		verify(producer1).close();
+		verify(producer2, never()).beginTransaction();
```
artembilan: The `producer2` is really never used. Not sure why we need it at all in the test.
Just because you have only this:
```
given(pf.createProducer()).willReturn(producer1);
```

On the other hand since we really talk here about `executeInTransaction()` from the `DeadLetterPublishingRecoverer`, then it would be better to verify exactly this method call. Maybe really with such a `never()`...

artembilan:
Gary, there are some conflicts with current `master`.

Would you mind to resolve them before merging this one?

Thanks

Issue #859, opened 2018-11-05
Nested transactions broken


garyrussell:
`KafkaTemplate.executeInTransaction()` when running on a transactional container thread is supposed to start a new transaction.

This no longer happens since adding the `producerPerConsumerPartition` to solve zombie fencing.

We need to restore the previous behavior in this case.

PR #860, opened 2018-11-05
GH-859: Fix nested transactions

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/859

When using `executeInTransaction` on a transactional container thread,
we cannot use the existing transaction - clear the TL to allow a new
Producer to be allocated.

Invalid state transition (in-trans to in-trans).

**cherry-pick to 2.1.x, 2.0.x, backport needed for 1.3.x**


artembilan:
... and cherry-picked to `2.1.x` and `2.0.x`.

I believe we can't just back-port it into `1.3.x` because of `default` method on the `ProducerFactory`.

Do you have anything in mind how to proceed?

Thanks

garyrussell:
Right - we will have to check that it's a DKPF and cast - I will do a backport PR.

Issue #842, opened 2018-10-30
NPE that hides real error when trying to create IllegalStateException on annotation processing


thomas-hugon:
Hello,

There are multiple possible NPE in class org.springframework.kafka.annotation.KafkaListenerAnnotationBeanPostProcessor
 on methods resolveExpressionAs* in the else statements because if the resolved value is null, resolved.getClass() that is used to create context for the thrown exception will throw a NPE that will hide the real error.

exemple:
https://github.com/spring-projects/spring-kafka/blob/c92f1bd3a36a30fd70eff065772d2b9910318620/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java#L663
https://github.com/spring-projects/spring-kafka/blob/c92f1bd3a36a30fd70eff065772d2b9910318620/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java#L522
https://github.com/spring-projects/spring-kafka/blob/c92f1bd3a36a30fd70eff065772d2b9910318620/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java#L694
https://github.com/spring-projects/spring-kafka/blob/c92f1bd3a36a30fd70eff065772d2b9910318620/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java#L678

artembilan:
@thomas-hugon ,

Please, go ahead and contribute the fix on the matter.
Even if an input value can't be `null` or empty String, the `resolveEmbeddedValue()` and `BeanExpressionResolver.evaluate()` have a `@Nullable` contract for their return value.

Thank you for catching that!

PR #864, opened 2018-11-06
GH-842: Fix NPEs in KafkaListenerAnnotationBPP

artembilan:
Fixes spring-projects/spring-kafka#842

**Cherry-pick to 2.1.x, 2.0.x and 1.3.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -130,8 +130,7 @@ public class KafkaListenerAnnotationBeanPostProcessor<K, V>
 	 */
 	public static final String DEFAULT_KAFKA_LISTENER_CONTAINER_FACTORY_BEAN_NAME = "kafkaListenerContainerFactory";
 
-	private final Set<Class<?>> nonAnnotatedClasses =
-			Collections.newSetFromMap(new ConcurrentHashMap<Class<?>, Boolean>(64));
+	private final Set<Class<?>> nonAnnotatedClasses = Collections.newSetFromMap(new ConcurrentHashMap<>(64));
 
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -517,6 +517,9 @@ private Pattern resolvePattern(KafkaListener kafkaListener) {
 			else if (resolved instanceof String) {
 				pattern = Pattern.compile((String) resolved);
 			}
+			else if (resolved == null) {
+				return null;
+			}
 			else {
```
garyrussell: Why not `else (if resolved != null)` rather than an additional branch?

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -517,10 +517,10 @@ private Pattern resolvePattern(KafkaListener kafkaListener) {
 			else if (resolved instanceof String) {
 				pattern = Pattern.compile((String) resolved);
 			}
-			else if (resolved == null) {
-				return null;
-			}
 			else {
```
garyrussell: You misunderstood me; we don't need an extra branch at all...

```java
	private Pattern resolvePattern(KafkaListener kafkaListener) {
		Pattern pattern = null;
		String text = kafkaListener.topicPattern();
		if (StringUtils.hasText(text)) {
			Object resolved = resolveExpression(text);
			if (resolved instanceof Pattern) {
				pattern = (Pattern) resolved;
			}
			else if (resolved instanceof String) {
				pattern = Pattern.compile((String) resolved);
			}
			else if (resolved != null) {
				throw new IllegalStateException(
						"topicPattern must resolve to a Pattern or String, not " + resolved.getClass());
			}
		}
		return pattern;
	}
```

Issue #869, opened 2018-11-09
JsonDeserializer raises NPE if targetType is an array


oranje42:
Spring-Kafka version: 2.2.0

The following code snippet produces an NPE (the error occurs with any array type, `String[]` is just used as an example here):
```
JsonDeserializer<String[]> deserializer = new JsonDeserializer<>(String[].class);
```
The stacktrace is as follows:
```
Exception in thread "main" java.lang.NullPointerException
	at org.springframework.kafka.support.serializer.JsonDeserializer.addTargetPackageToTrusted(JsonDeserializer.java:299)
	at org.springframework.kafka.support.serializer.JsonDeserializer.<init>(JsonDeserializer.java:180)
	at org.springframework.kafka.support.serializer.JsonDeserializer.<init>(JsonDeserializer.java:142)
	at org.springframework.kafka.support.serializer.JsonDeserializer.<init>(JsonDeserializer.java:130)
```
The error is caused in method `addTargetPackageToTrusted` because `this.targetType.getPackage()` returns `null` for array types:
```
private void addTargetPackageToTrusted() {
    if (this.targetType != null) {
        addTrustedPackages(this.targetType.getPackage().getName());
    }
}
```
A possible fix could be as follows (this one would also work for arrays of primitive types, like `int[]`):
```
private void addTargetPackageToTrusted() {
    if (this.targetType != null) {
        Package targetPackage = resolvePackage(this.targetType);
        if (targetPackage != null) {
            addTrustedPackages(targetPackage.getName());
        }
    }
}

// this method might still return null if the provided type represents a primitive array type, as int[]
private Package resolvePackage(Class<?> type) {
    return (type.isArray() ? type.getComponentType() : type).getPackage();
}
```

artembilan:
Uh! So, bad that we have missed this...

I wonder if we should use `org.springframework.utilClassUtils.getPackageName()` instead:
```
	/**
	 * Determine the name of the package of the given class,
	 * e.g. "java.lang" for the {@code java.lang.String} class.
	 * @param clazz the class
	 * @return the package name, or the empty String if the class
	 * is defined in the default package
	 */
	public static String getPackageName(Class<?> clazz) {
```

The `first-timers-only` issue?

oranje42:
@garyrussell I could possibly provide a pull request beginning of next week. However, if anyone else is able to provide a fix before, please go ahead.

@artembilan: Using `ClassUtils.getPackageName()` could of course be an option.

[...]

PR #874, opened 2018-11-12
GH-869: fixed NPE in case of array target types

oranje42:
This PR fixes the NPE that is thrown when an array targetType is passed to one of the `JsonDeserializer` constructors.

I tried to keep the implementation in sync with `DefaultJackson2JavaTypeMapper.isTrustedPackage(String requestedType)` by using the same method to get the target package name:

`ClassUtils.getPackageName(this.targetType).replaceFirst("\\[L", "")`

However, this method seems to only support single-dimensional arrays, indicated by a single square bracket at the beginning of the returned package name. If support for multi-dimensional arrays is also of interest (e.g. by switching to `replaceFirst("\\[+L", "")`), it should be added via another PR.


oranje42:
Cool, thanks! And get well soon :wink:

Issue #881, opened 2018-11-19
NPE in container.closeProducers()


garyrussell:
See https://gitter.im/spring-projects/spring-kafka?at=5bf2adbc1ad4155d3aeb29b5

PR #882, opened 2018-11-19
GH-881: Null check in closeProducers()

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/881

`getAssignedPartitions()` can return `null`.

__cherry-pick to 2.1.x, 2.0.x, 1.3.x__


artembilan:
and cherry-picked to `2.1.x` and `2.0.x` after fixing conflicts.

Back-ported to `1.3.x` without `@Nullable`.

PR #891, opened 2018-11-28
Fix NPE in InvocationResult's toString

neetkee:
Calling toString() in
 `org.springframework.kafka.listener.adapter.InvocationResultt` raises NPE, if `sendTo` is null. 
It can be checked by adding
 `<Logger name="org.springframework.kafka.listener.adapter" level="debug" />` and running `org.springframework.kafka.annotation.EnableKafkaIntegrationTests#testMultiJson` for example.


garyrussell:
Thanks; now I see what Sonar was complaining about. Merged as 84eec92cbf76f3e3ca0d26659343d51a10efa712

Issue #892, opened 2018-11-29
Reinstate MimeTypeJsonDeserializer


garyrussell:
See https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/509

For backward compatibility with 2.1 and earlier.

PR #893, opened 2018-11-29
GH-892: Reinstate `MimeTypeJsonDeserializer`

artembilan:
Fixes spring-projects/spring-kafka#892

For compatibility with older Spring Kafka producer, it is better to keep
a `MimeTypeJsonDeserializer` when they can send `MimeType` headers
as a serialized version, not string representation


garyrussell:
We should coordinate a bit better 😄 

Please grab the test from my version: https://github.com/garyrussell/spring-kafka/commit/0ebeb9de2a479ecc0057151642fbe03468dc129f

garyrussell:
Merged as 56b0e937e15b6177eef8cc31a6f20e93e75b041c with a bit of polish and a test.

Issue #900, opened 2018-12-10
Initial Commit in onPartitionsAssigned has wrong transactional id


garyrussell:
See https://stackoverflow.com/questions/53706988/spring-kafka-transaction-id-is-wrong-at-start

Need to invoke `TransactionSupport.setTransactionIdSuffix`

PR #901, opened 2018-12-10
GH-900: Use correct txId for initial commit

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/900

Use the proper zombie-fencing `transational.id` when committing initial
offsets after assignment.

**cherry-pick to all, 1.3.x will probable need a backport**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -1656,14 +1656,14 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 							ListenerConsumer.this.kafkaTxManager != null) {
 						try {
 							offsetsToCommit.forEach((partition, offsetAndMetadata) -> {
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -525,12 +531,6 @@ else if (listener instanceof MessageListener) {
 				this.batchErrorHandler = new BatchLoggingErrorHandler();
 			}
 			Assert.state(!this.isBatchListener || !this.isRecordAck, "Cannot use AckMode.RECORD with a batch listener");
-			if (this.transactionManager != null) {
-				this.transactionTemplate = new TransactionTemplate(this.transactionManager);
-			}
-			else {
-				this.transactionTemplate = null;
-			}
```
garyrussell: FYI, this move would not be needed for production since `onPartitionsAssigned()` won't be called until we `poll()`.

It is, needed, however, for mocked tests since it is called directly downstream of `subscribeOrAssignTopics()`.

PR #913, opened 2018-12-19
NonTrustedHeaderType header propagation

garyrussell:
Consider `->app1->app2` when `app1` receives a header with a non-trusted type.

This will be decoded as a `NonTrustedHeaderType` containing the raw json and the type.

If `app1` then propagates that header to `app2`; there, since NTHT is also not truested,
the header remains as NTHT, but the original type is lost (the type is set to NTHT).

Always trust NTHT and make it jackson-friendly so that the propagaged header is restored
as-is, retaining the original non-trusted type.

If the original type is trusted by app2, then decode it properly.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
@@ -264,7 +264,21 @@ public class DefaultKafkaHeaderMapper extends AbstractKafkaHeaderMapper {
 					}
 					if (trusted) {
 						try {
-							headers.put(h.key(), getObjectMapper().readValue(h.value(), type));
+							Object value = getObjectMapper().readValue(h.value(), type);
+							if (type.equals(NonTrustedHeaderType.class)) {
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -264,7 +264,21 @@ public void toHeaders(Headers source, final Map<String, Object> headers) {
...
 						try {
-							headers.put(h.key(), getObjectMapper().readValue(h.value(), type));
+							Object value = getObjectMapper().readValue(h.value(), type);
+							if (type.equals(NonTrustedHeaderType.class)) {
+								// Upstream NTHT propagated; may be trusted here...
+								NonTrustedHeaderType nth = (NonTrustedHeaderType) value;
+								if (trusted(nth.getUntrustedType())) {
+									try {
+										type = ClassUtils.forName(nth.getUntrustedType(), null);
+										value = getObjectMapper().readValue(nth.getHeaderValue(), type);
```
artembilan: Can we extract a local variable for the `ObjectMapper`?

PR #924, opened 2019-01-03
Added custom serializer example

RemusRD:
Hi!, first PR here, hope it all works, just tell me if anything has to be fixed, thanks!


samples/sample-04/.gitignore
```diff
@@ -0,0 +1,29 @@
+# Created by .ignore support plugin (hsz.mobi)
```
artembilan: Please, copy/paste content from one existing `.ignore` file, e.g. : https://github.com/spring-projects/spring-kafka/blob/master/samples/sample-01/.gitignore

samples/sample-04/README.adoc
```diff
@@ -0,0 +1,10 @@
+== Sample 4
+
+This sample uses a custom kafka serializer and deserializer in order to consume and send events using the jdk8 LocalDatetime
+and Kotlin data classes.
```
artembilan: Please, follow *One sentence per line rule*: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

samples/sample-04/build.gradle
```diff
@@ -0,0 +1,85 @@
...
+    }
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath("org.springframework.boot:spring-boot-gradle-plugin:${springBootVersion}")
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${kotlinVersion}")
+        classpath("org.jetbrains.kotlin:kotlin-allopen:${kotlinVersion}")
+        classpath("org.jetbrains.kotlin:kotlin-noarg:${kotlinVersion}")
+        classpath("io.spring.gradle:dependency-management-plugin:1.0.2.RELEASE")
```
artembilan: Looks like there is a newer version of this plugin: https://github.com/spring-gradle-plugins/dependency-management-plugin/releases

samples/sample-04/build.gradle
```diff
@@ -0,0 +1,85 @@
+buildscript {
+    ext {
+        kotlinVersion = "1.3.10"
+        springBootVersion = "2.1.1.RELEASE"
+        arrow_version = "0.8.1"
```
artembilan: What is this about?
Doesn't look like this property is used somewhere in the `buildscript` definition

samples/sample-04/build.gradle
```diff
@@ -0,0 +1,85 @@
...
+apply plugin: "kotlin"
+apply plugin: "kotlin-spring"
+apply plugin: "org.springframework.boot"
+apply plugin: "io.spring.dependency-management"
+apply plugin: "kotlin-allopen"
+apply plugin: "kotlin-noarg"
+apply plugin: "kotlin-jpa"
+
+
+group "com.remusrd"
```
artembilan: That's wrong group if you contribute into this project.
See all other samples here on the matter.

samples/sample-04/build.gradle
```diff
@@ -0,0 +1,85 @@
...
+dependencies {
+    // Kotlin
+    implementation "org.jetbrains.kotlin:kotlin-stdlib"
+    implementation "org.jetbrains.kotlin:kotlin-reflect"
+    implementation "com.fasterxml.jackson.module:jackson-module-kotlin"
+    implementation "io.arrow-kt:arrow-core:$arrow_version"
+    implementation "io.arrow-kt:arrow-data:$arrow_version"
+
+    // Spring Boot
+    implementation "org.springframework.boot:spring-boot-starter-web:$springBootVersion"
```
artembilan: I don't think that we need an explicit version for Spring Boot, if we use an appropriate Spring Boot plugin

samples/sample-04/build.gradle
```diff
@@ -0,0 +1,85 @@
...
+    implementation "org.springframework.kafka:spring-kafka"
+    implementation "com.fasterxml.jackson.module:jackson-module-kotlin:2.9.7"
+
+    // BBDD
+    implementation "mysql:mysql-connector-java:8.0.13"
+    implementation "com.h2database:h2:1.4.197"
+
+
+    // Test
+    testImplementation "junit:junit:4.12"
```
artembilan: I believe we fully don't need this dependency if we use `spring-boot-starter-test`

samples/sample-04/gradle.properties
```diff
@@ -0,0 +1 @@
+kotlin.code.style=official
```
artembilan: All files must finish with blank like.
You can configure your IDE to do that for you on save.

On the other hand: why do you use `official` value for this property?
Isn't it really official by default? 😕

samples/sample-04/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.0-all.zip
```
artembilan: Looks like there is already Gradle `5.1`. Also no point to force everyone to download `all`. The `bin` us fully enough for this kind of projects.

samples/sample-04/src/main/kotlin/com/remusrd/notesample/configuration/KafkaConfig.kt
```diff
@@ -0,0 +1,57 @@
+package com.remusrd.notesample.configuration
```
artembilan: Need Copyright header in all the classes.

samples/sample-04/src/main/kotlin/com/remusrd/notesample/configuration/KafkaConfig.kt
```diff
@@ -0,0 +1,57 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.autoconfigure.kafka.KafkaProperties
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder
+import org.springframework.kafka.core.*
```
artembilan: No asterisk imports, please.

samples/sample-04/src/main/kotlin/com/remusrd/notesample/configuration/KafkaConfig.kt
```diff
@@ -0,0 +1,57 @@
...
+@Configuration
+class KafkaConfig {
+    @Autowired
+    lateinit var jackson2ObjectMapperBuilder: Jackson2ObjectMapperBuilder
+    @Autowired
+    lateinit var kafkaProperties: KafkaProperties
+
+
+    @Bean
+    fun kafkaTemplate(): KafkaTemplate<Any, Any> {
```
artembilan: If we rely on Spring Boot, I don't see reason to declare this bean.

samples/sample-04/src/main/kotlin/com/remusrd/notesample/configuration/KafkaConfig.kt
```diff
@@ -0,0 +1,57 @@
...
+        return KafkaTemplate<Any, Any>(defaultKafkaProducerFactory())
+    }
+
+
+    @Bean
+    fun defaultKafkaConsumerFactory(): ConsumerFactory<Any, Any> {
```
artembilan: Why do we need to override auto-configured beans?
This one and the next - `defaultKafkaProducerFactory()`.

samples/sample-04/src/main/kotlin/com/remusrd/notesample/domain/annotation/NoArg.kt
```diff
@@ -0,0 +1,5 @@
+package com.remusrd.notesample.domain.annotation
+
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.SOURCE)
+annotation class NoArg
```
artembilan: Why do we need this annotation and the next `Open` one?
Aren't there something already explicit in Kotlin world?
Like I see those in the Gradle config:
```
apply plugin: "kotlin-allopen"
apply plugin: "kotlin-noarg"
```

samples/sample-04/src/main/resources/application.yml
```diff
@@ -0,0 +1,24 @@
...
+
+logging.level.org.springframework: DEBUG
+spring:
+  datasource:
+    driver-class-name: com.mysql.cj.jdbc.Driver
+
+spring.kafka.bootstrap-servers: localhost:9092
+spring.kafka.consumer.group-id: noteGroup
+spring.kafka.consumer.auto-offset-reset: earliest
+spring.kafka.consumer.properties.spring.json.trusted.packages: com.remusrd.notesample.domain.event
```
artembilan: Can you reformat this file to really true YAML style?

RemusRD: Thanks a lot for the feedback, will fix it in the next days!

samples/sample-04/src/main/kotlin/com/remusrd/notesample/configuration/KafkaConfig.kt
RemusRD: This was because IntelliJ says there are no beans of type KafkaTemplate, I guess it is a bug of some type

samples/sample-04/src/main/kotlin/com/remusrd/notesample/configuration/KafkaConfig.kt
RemusRD: What should I do instead?, creating my own beans without replacing the default ones?

samples/sample-04/gradle.properties
RemusRD: Oh, it was necessary before [Kotlin 1.3](https://kotlinlang.org/docs/reference/code-style-migration-guide.html), thanks for pointing out!

samples/sample-04/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+    implementation "org.jetbrains.kotlin:kotlin-reflect"
+    implementation "com.fasterxml.jackson.module:jackson-module-kotlin"
+    implementation "io.arrow-kt:arrow-core:$arrow_version"
+    implementation "io.arrow-kt:arrow-data:$arrow_version"
+
+    // Spring Boot
+    implementation "org.springframework.boot:spring-boot-starter-web"
+    implementation "org.springframework.boot:spring-boot-starter-data-jpa"
+    implementation "org.springframework.kafka:spring-kafka"
+    implementation "com.fasterxml.jackson.module:jackson-module-kotlin:2.9.7"
```
artembilan: We don't need an explicit version for this dependency as well. It is managed by the `jackson-bom`: https://github.com/FasterXML/jackson-bom/blob/master/pom.xml#L258

samples/sample-04/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+
+    // BBDD
+    implementation "mysql:mysql-connector-java:8.0.13"
+    implementation "com.h2database:h2:1.4.197"
```
artembilan: These two are managed by Spring Boot: we don't need their versions.

samples/sample-04/src/main/kotlin/com/example/controller/NoteController.kt
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package com.example.controller
+
+import arrow.core.Option
+import com.example.domain.Note
+import com.example.service.NoteService
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.web.bind.annotation.*
```
artembilan: Please, revise all the classes for avoiding asterisk imports.

samples/sample-04/src/main/kotlin/com/example/domain/event/NoteEvent.kt
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package com.example.domain.event
+
+import com.example.domain.Note
+import com.example.domain.annotation.NoArg
+import com.example.domain.annotation.Open
+
+@NoArg
+@Open
```
artembilan: Why do we need these annotations?
Why the mentioned plugins not enough?

samples/sample-04/src/main/kotlin/com/example/sampleApp.kt
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.example
+
+import org.springframework.boot.SpringApplication
+import org.springframework.boot.autoconfigure.SpringBootApplication
+
+@SpringBootApplication
+class NoteSampleApplication
```
artembilan: Why file name doesn't match class name?

samples/sample-04/src/main/resources/application.yml
```diff
@@ -0,0 +1,30 @@
...
+      bootstrap-mode: default
+  kafka:
+    bootstrap-servers: localhost:9092
+    consumer:
+      group-id: noteGroup
+      auto-offset-reset: earliest
+      properties:
+        spring:
+          json:
+            trusted:packages: com.remusrd.notesample.domain.event
```
artembilan: What is this package?
Doesn't look like legitimate package name in the public project.

samples/sample-04/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+
+repositories {
+    mavenCentral()
+    maven { url 'http://repo.spring.io/milestone' }
+}
+noArg{
+    annotation("com.remusrd.notesample.domain.annotation.NoArg")
+}
+allOpen{
+    annotation("com.remusrd.notesample.domain.annotation.Open")
```
artembilan: I feel like we can just remove these two declarations.
First of all - wrong package.
Secondly if we are here, then project even works without these two annotations.

I'm confused, so everybody else is going to be to. 🤷‍♂️

samples/sample-04/src/main/kotlin/com/example/configuration/KafkaConfig.kt
```diff
@@ -0,0 +1,70 @@
...
+    lateinit var kafkaProperties: KafkaProperties
+
+
+    @Bean
+    fun defaultKafkaConsumerFactory(): ConsumerFactory<Any, Any> {
+        val objectMapper = jackson2ObjectMapperBuilder.build() as ObjectMapper
+        val jsonDeserializer = JsonDeserializer<Any>(objectMapper)
+        jsonDeserializer.configure(kafkaProperties.buildConsumerProperties(), false)
+        val kafkaConsumerFactory = DefaultKafkaConsumerFactory<Any, Any>(
+                kafkaProperties.buildConsumerProperties(),
```
garyrussell, line 46: Unfortunately, the Boot team does not consider `KafkaProperties` a public API and we would be conflicting with their position if we put code like this in an official Spring repo, promoting its use. They asked us to remove all such references from the documentation.

artembilan:
I'm closing this one as `not fully Kafka related`: too much Kotlin, too much JPA.
The sample should be as simple as possible to give a reader a chance to learn something.
Plus there is no reporter reaction for our review for a long time.

Thanks for understanding.

Issue #926, opened 2019-01-08
ErrorHandlingJsonSerializer2 configuration shortcoming


dipeti:
Quote from documentation of 2.2.2.RELEASE:
_If the delegate fails to deserialize the record content, the ErrorHandlingDeserializer2 returns a null value and a DeserializationException in a header, containing the cause and raw bytes. When using a record-level MessageListener, if either the key or value contains a DeserializationException header, the container’s ErrorHandler is called with the failed ConsumerRecord; the record is not passed to the listener._

This all works perfect if I configure my `ConsumerFactory` as described in the docs: 
```
public ConsumerFactory<?, ?> consumerFactoryFromDocs() {
        Map<String, Object> props = new HashMap<>()
        props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, ErrorHandlingDeserializer2.class);
        props.put(ErrorHandlingDeserializer.KEY_DESERIALIZER_CLASS, JsonDeserializer.class);
        return new DefaultKafkaConsumerFactory<>(props);
}
```

However, I have configured my application in a slightly different way, which should arguably lead to the same settings but it does not. 
```
public ConsumerFactory<?, ?> consumerFactoryMyWay() {
        return new DefaultKafkaConsumerFactory<>(new HashMap<>(), new StringDeserializer(), 
new ErrorHandlingDeserializer2<>(new JsonDeserializer<>(Foo.class)));
}
```

Using the latter setup for the `ConsumerFactory`, the `ListenerConsumer` fails to find the `value.deserializer` entry in the properties map and ends up setting the `checkNullValueForExceptions` field to `false`. Checking _"if either the key or value contains a DeserializationException header"_ occurs only if `checkNullValueForExceptions` is set to `true`.
```
if (record.value() == null && this.checkNullValueForExceptions) {
   checkDeser(record, ErrorHandlingDeserializer2.VALUE_DESERIALIZER_EXCEPTION_HEADER);
}
if (record.key() == null && this.checkNullKeyForExceptions) {
   checkDeser(record, ErrorHandlingDeserializer2.KEY_DESERIALIZER_EXCEPTION_HEADER);
}
```

Can you explain what is the reason behind this?

garyrussell:
Good catch - we don't check if the deserializer was configured manually.

As a work around, you should be able to set the property as well as constructing your own (which will take precedence over the property, but we'll still  set the boolean based on the property).

PR #927, opened 2019-01-08
GH-926: Always Detect ErrorHandlingDeserializer2

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/926

Previously, the container only detected the presence of the deserializer
when it was configured using Kafka properties.

The container now detects it (and subclasses) when configured directly
into the consumer factory and properly routes exceptions to the error
handler.

Also add `setKey` to designate the deserializer as a key deserializer.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -553,14 +553,14 @@ public class KafkaMessageListenerContainer<K, V> extends AbstractMessageListener
 		}
 
 		private Object findDeserializerClass(Map<String, Object> props, String config) {
-			Object deserializerClass = KafkaMessageListenerContainer.this.consumerFactory.getKeyDeserializer();
-			if (deserializerClass == null) {
-				deserializerClass = props.get(config);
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -546,9 +546,21 @@ else if (listener instanceof MessageListener) {
...
-			this.checkNullKeyForExceptions = checkDeserializer(props.get(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG));
+			this.checkNullKeyForExceptions = checkDeserializer(
+					findDeserializerClass(props, ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG));
 			this.checkNullValueForExceptions = checkDeserializer(
-					props.get(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG));
+					findDeserializerClass(props, ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG));
+		}
+
+		private Object findDeserializerClass(Map<String, Object> props, String config) {
+			Object deserializerClass = KafkaMessageListenerContainer.this.consumerFactory.getKeyDeserializer();
```
artembilan: Essentially a result of the `getKeyDeserializer()` is not a `Class<?>`...

Issue #928, opened 2019-01-09
EmbeddedKafkaRule does not delete temporary directories


apanasevich:
Hi, 

I've faced the problem that EmbeddedKafkaRule does not delete temporary directories. Here is a simple example https://github.com/apanasevich/embeded-kafka-rule

Every run of EmbededKafkaTest#example on Windows leaves non-empty directory with zookeeper log file in user's temporary directory.

artembilan:
We do in the `after()` implementation of the rule something like this:
```
CoreUtils.delete(kafkaServer.config().logDirs());
...
this.zookeeper.shutdown();
```

So, if you say that dirs are still there in the tmp user dir, then it is not what we can control from Spring level.

I would recommend to consult with Apache Kafka on the matter if you find this as a problem.

I use Windows as well and the latest Kafka client works pretty stable for me here. I might just not noticed those dirs in my tmp though, but according the code it doesn't look like we can do anything else to clean up them properly.

garyrussell:
I just looked at NiFi's implementation

```java
        if (this.started) {
            logger.info("Shutting down Kafka server");
            this.kafkaServer.shutdown();
            this.kafkaServer.awaitShutdown();
            logger.info("Shutting down Zookeeper server");
            this.shutdownZookeeper();
            logger.info("Embedded Kafka is shut down.");
            this.cleanupKafkaWorkDir();
            this.started = false;
        }
```

https://github.com/apache/nifi/blob/master/nifi-nar-bundles/nifi-kafka-bundle/nifi-kafka-0-8-processors/src/test/java/org/apache/nifi/processors/kafka/test/EmbeddedKafka.java

apanasevich:
I've created [an issue](https://issues.apache.org/jira/browse/ZOOKEEPER-3248) in Apache tracker.

FNaughton:
Looks like might be related to / has same root cause as - https://issues.apache.org/jira/browse/KAFKA-6647

artembilan:
Closed as *Won't Fix*.
See mentioned above Apache Kafka JIRAs.
Should be addressed automatically after upgrade into a Kafka Client version with the fix.

PR #932, opened 2019-01-12
GH-928: ZooKeeperServer with an auto deletable data directory

apanasevich:
Hi,

here is a fix for [Issue 928](https://github.com/spring-projects/spring-kafka/issues/928). I've created a class to start a ZooKeeper with an auto deletable data dicrectory.

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/ZooKeeperServer.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/ZooKeeperServer.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.test;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.kafka.test.TestUtils;
+import org.apache.zookeeper.server.ServerConfig;
+import org.apache.zookeeper.server.ZooKeeperServerMain;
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@apanasevich Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=932)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=932) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@apanasevich Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=932)!

apanasevich:
Shall I fix codestyle issues manually or there is a codestyle.xml file for IDEA?

spring-kafka-test/src/main/java/org/springframework/kafka/test/ZooKeeperServer.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.zookeeper.server.ServerConfig;
+import org.apache.zookeeper.server.ZooKeeperServerMain;
+import org.apache.zookeeper.server.quorum.QuorumPeerConfig;
+
+/**
+ * A zookeeper server to use with unit tests.
+ *
+ * @author Dmitrii Apanasevich
+ */
+public final class ZooKeeperServer {
```
artembilan: We can't accept such a breaking change. On the other hand it looks like in the end you are pursuing a `cnxnFactory.shutdown()`, which, looks like is accessed via:
```
this.zookeeper.zookeeper().shutdown();
this.zookeeper.zookeeper().getServerCnxnFactory().shutdown();
```
in the `EmbeddedKafkaBroker`.

I agree about other polishing around magic numbers and missed logs for swallowed exception, but I don't see reason in this class overhead which may lag with updates in the future Apache Kafka versions.

So, WDYT about my suggestion?

apanasevich: I have no other options, so I've created [an issue](https://issues.apache.org/jira/browse/ZOOKEEPER-3248) in Apache tracker.

artembilan:
Looks like the issue with regular Kafka Consumer and Produce usage has disappeared with the latest Kafka Client.

We still an issue in Kafka Streams tests, but that's going to be addressed on Apache Kafka side: https://issues.apache.org/jira/browse/KAFKA-6647

So, closed as *Won't Fix*.

Thank you for the effort anyway!

Issue #917, opened 2018-12-28
Specify the port with annotation


JBertaux:
Hello,
With the class rule `EmbeddedKafkaRule` this is possible to specify the port like that:
```java
public static EmbeddedKafkaRule embeddedKafka = new EmbeddedKafkaRule(1, true, "topic").kafkaPorts(9092);
```
How can I achieve the same with the annotation ? (I use Junit 5)
```java
@EmbeddedKafka(controlledShutdown = true, topics = {"topic"})
```

artembilan:
I think this is not too bad to support on the annotation level:
```
/**
 * @return ports for brokers.
 */
int[] ports() default { };
```

So, treating this as an  improvement.
And that's not too late yet to implement in the current version.

PR #935, opened 2019-01-14
Specify the port with annotation #917

SLourenco:
Added ports attribute to EmbeddedKafka annotation.
It has a default value of 0 to be consistent with the count attribute.

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 the original author or authors.
+ * Copyright 2017-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -53,8 +53,9 @@ import org.springframework.core.annotation.AliasFor;
  * @author Elliot Metsger
  * @author Zach Olauson
  * @author Gary Russell
+ * @author Sergio Lourenco
  *
- * @since 1.3
+ * @since 2.24
  *
  * @see org.springframework.kafka.test.EmbeddedKafkaBroker
  */
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
@@ -81,7 +81,7 @@ public @interface EmbeddedKafka {
 	 * @return passed into {@code kafka.utils.TestUtils.createBrokerConfig()}.
 	 */
 	boolean controlledShutdown() default false;
-	
+
 	/**
 	 * Set explicit ports on which the kafka brokers will listen. Useful when running an
 	 * embedded broker that you want to access from other processes.
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
@@ -80,6 +80,11 @@ public @interface EmbeddedKafka {
 	 * @return passed into {@code kafka.utils.TestUtils.createBrokerConfig()}.
 	 */
 	boolean controlledShutdown() default false;
+	
+	/**
+	 * @return ports for brokers.
+	 */
+	int[] ports() default {0};
 
 	/**
 	 * @return partitions per topic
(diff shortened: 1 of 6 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
```diff
@@ -80,6 +80,11 @@
 	 * @return passed into {@code kafka.utils.TestUtils.createBrokerConfig()}.
 	 */
 	boolean controlledShutdown() default false;
+	
+	/**
+	 * @return ports for brokers.
```
artembilan: Please, add `@since 2.2.4` into this Javadoc.
Or do we switch to the `2.3.0` already, @garyrussell ?

garyrussell: No; this can go into 2.2.4.

spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
```diff
@@ -53,8 +53,9 @@
  * @author Elliot Metsger
  * @author Zach Olauson
  * @author Gary Russell
+ * @author Sergio Lourenco
  *
- * @since 1.3
+ * @since 2.24
```
artembilan: No, you can't change a `@since`. You just have to add it into the method Javadoc as I mentioned on the line.

spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafkaContextCustomizer.java
```diff
@@ -41,8 +41,9 @@
  * @author Elliot Metsger
  * @author Zach Olauson
  * @author Oleg Artyomov
+ * @author Sergio Lourenco
  *
- * @since 1.3
+ * @since 2.2.4
```
artembilan: You don't change this one at all. 
And you don't introduce any public API in this class, so only your name and Copyright.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultKafkaConsumerFactoryTests.java
```diff
@@ -45,13 +45,16 @@
...
+ * 
+ * @since 2.2.4
  *
  */
 @EmbeddedKafka(topics = { "txCache1", "txCache2", "txCacheSendFromListener" },
 		brokerProperties = {
 				"transaction.state.log.replication.factor=1",
-				"transaction.state.log.min.isr=1"}
+				"transaction.state.log.min.isr=1"},
+		ports= {8085}
```
artembilan: I believe this can work even with braces. Just because you use a single value.

src/reference/asciidoc/testing.adoc
```diff
@@ -224,8 +224,9 @@ public class KafkaStreamsTests {
 
 }
 ----
+Starting with _version 2.2.4_ the `@EmbeddedKafka` annotation can also be used to specify the kafka ports property.
```
artembilan: No wrapping to the italic, please, and comma before `the`.

src/reference/asciidoc/whats-new.adoc
```diff
@@ -72,6 +72,7 @@ See <<headers>> for more information.
 The `KafkaEmbedded` class and its `KafkaRule` interface have need deprecated in favor of the `EmbeddedKafkaBroker` and its JUnit 4 `EmbeddedKafkaRule` wrapper.
 The `@EmbeddedKafka` annotation now populates an `EmbeddedKafkaBroker` bean instead of the deprecated `KafkaEmbedded`.
 This allows the use of `@EmbeddedKafka` in JUnit 5 tests.
+The `@EmbeddedKafka` annotation now has the attribute `ports` to specify the port which will populate the `EmbeddedKafkaBroker` (since 2.2.4).
```
artembilan: No reason to mention  a version here in this file.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultKafkaConsumerFactoryTests.java
```diff
@@ -152,6 +155,15 @@ public void testContainerTxProducerIsNotCached() throws Exception {
 			pfTx.destroy();
 		}
 	}
+	
+	@Test
+	public void testKafkaEmbedded() {
+		assertThat(embeddedKafka.getBrokersAsString()).isEqualTo("127.0.0.1:" + 8085);
```
artembilan: I somehow don't like this test against an explicit port and an attempt for real Kafka broker on it.

Can't we somehow have a mocking test do not open this port in the reality?
I mean there is just enough to be sure that `ports` is populated properly into the `EmbeddedKafkaBroker.kafkaPorts` without starting a broker(s).

spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
artembilan: I also want to see in this Javadoc an explanation that number of ports provided must be equal to the `count` value.

artembilan:
@SLourenco ,

Thank you for the contribution.
Looking forward for more!

PR #936, opened 2019-01-15
A method-chaining way to build branches for Streams API topology

inponomarev:
(Maybe this one has to be posted directly to Kafka project instead of here. However, spring-kafka already contains a number of convenience classes).

`KStream.branch` method uses varargs to supply predicates and returns array of streams ('[Each stream in the result array corresponds position-wise (index) to the predicate in the supplied predicates](http://home.apache.org/~ijuma/kafka-0.10.0.1-rc1/javadoc/org/apache/kafka/streams/kstream/KStream.html)'). 

This is poor API design that makes building branches very inconvenient because of 'impedance mismatch' between arrays and generics in Java language.

 * In general, the code  have low cohesion: we need to define predicates in one place, and respective stream processors in another place of code. 
 
 * If the number of predicates is predefined, this method forces us to use 'magic numbers' to extract the right branch from the result (see examples [here](https://stackoverflow.com/questions/48950580/kafka-streams-send-on-different-topics-depending-on-streams-data)).
 
 * If we need to build branches dynamically (e. g. one branch per enum value) we inevitably have to deal with 'generic arrays' and 'unchecked typecasts'.
 
The proposed class `KafkaStreamBrancher` introduces new standard way to build branches on top of KStream. 

Instead of 

```java
KStream<String, String> source_o365_user_activity = builder.stream("source");
KStream<String, String>[] branches = source_o365_user_activity.branch( 
      (key, value) -> value.contains("A"),
      (key, value) -> value.contains("B"),
      (key, value) -> true
     );

branches[0].to("A");
branches[1].to("B");
branches[2].to("C");
```

we can use

```java
new KafkaStreamsBrancher<String, String>()
   .addBranch((key, value) -> value.contains("A"), ks->ks.to("A"))
   .addBranch((key, value) -> value.contains("B"), ks->ks.to("B"))
   //default branch should not necessarily be defined in the end!
   .addDefaultBranch(ks->ks.to("C"))
   .onTopOf(builder.stream("source"))
```

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/KafkaStreamBrancher.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/KafkaStreamBrancher.java
@@ -32,10 +32,10 @@ import org.apache.kafka.streams.kstream.Predicate;
  * <pre>
  * {@code
  * new KafkaStreamsBrancher<String, String>()
- *    .addBranch((key, value) -> value.contains("A"), ks->ks.to("A"))
- *    .addBranch((key, value) -> value.contains("B"), ks->ks.to("B"))
+ *    .branch((key, value) -> value.contains("A"), ks->ks.to("A"))
+ *    .branch((key, value) -> value.contains("B"), ks->ks.to("B"))
  *    //default branch should not necessarily be defined in the end
- *    .addDefaultBranch(ks->ks.to("C"))
+ *    .defaultBranch(ks->ks.to("C"))
  *    .onTopOf(builder.stream("source"))
  * }
  * </pre>
@@ -58,7 +58,7 @@ public final class KafkaStreamBrancher<K, V> {
 	 * @param consumer  The consumer of this branch's {@code KStream}
 	 * @return {@code this}
 	 */
-	public KafkaStreamBrancher<K, V> addBranch(Predicate<? super K, ? super V> predicate,
+	public KafkaStreamBrancher<K, V> branch(Predicate<? super K, ? super V> predicate,
 											Consumer<? super KStream<K, V>> consumer) {
 		this.predicateList.add(Objects.requireNonNull(predicate));
 		this.consumerList.add(Objects.requireNonNull(consumer));
@@ -66,14 +66,14 @@ public final class KafkaStreamBrancher<K, V> {
 	}
 
 	/**
-	 * Defines a default branch. To this stream will be directed all the messages that
-	 * were not dispatched to other branches. This method should not necessarily be called in the end
+	 * Defines a default branch. All the messages that were not dispatched to other branches will be directed
+	 * to this stream. This method should not necessarily be called in the end
 	 * of chain.
 	 *
 	 * @param consumer The consumer of this branch's {@code KStream}
 	 * @return {@code this}
 	 */
-	public KafkaStreamBrancher<K, V> addDefaultBranch(Consumer<? super KStream<K, V>> consumer) {
+	public KafkaStreamBrancher<K, V> defaultBranch(Consumer<? super KStream<K, V>> consumer) {
 		this.defaultConsumer = Objects.requireNonNull(consumer);
 		return this;
 	}
@@ -82,8 +82,9 @@ public final class KafkaStreamBrancher<K, V> {
 	 * Terminating method that builds branches on top of given {@code KStream}.
 	 *
 	 * @param stream {@code KStream} to split
+	 * @return the provided stream
 	 */
-	public void onTopOf(KStream<K, V> stream) {
+	public KStream<K, V> onTopOf(KStream<K, V> stream) {
 		if (this.defaultConsumer != null) {
 			this.predicateList.add((k, v) -> true);
 			this.consumerList.add(this.defaultConsumer);
@@ -94,5 +95,6 @@ public final class KafkaStreamBrancher<K, V> {
 		for (int i = 0; i < this.consumerList.size(); i++) {
 			this.consumerList.get(i).accept(result[i]);
 		}
+		return stream;
 	}
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/KafkaStreamBrancher.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/KafkaStreamBrancher.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


artembilan:
Spring Cloud Stream Kafka Streams Binder provides a bit different solution on the subject: https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/docs/src/main/asciidoc/kafka-streams.adoc#multiple-output-bindings-aka-branching, but I definitely see your point with the custom processing in the particular branch via that `ks->ks` lambda.

So, let me review what you have so far and let's hope we can merge this !

artembilan:
OK. That's all interesting, but doesn't look like the code belongs to Spring project. There is just nothing Spring-related in your fix!
Have you tried to contact with Apache Kafka to propose this feature over there?

Thanks

inponomarev:
Yes, that's what I said in the first place: ideally, Apache Kafka project itself is a better place for this change.

The only reason to post it here is a much better chance to see this class in Maven in reasonable time: complicated [process](https://cwiki.apache.org/confluence/display/KAFKA/Contributing+Code+Changes) and [550 open PRs since 2015](https://github.com/apache/kafka/pulls) is a bit scaring for me :-)

Alright, @artembilan I'm glad that you liked the idea in general. 

Let's just postpone/close this PR for a while. I'll try to post this change to Kafka Project and we will see if they are going to do something with this.

artembilan:
Well, we even can merge this into our code base and deprecate/remove it from here whenever it is ready on Apache Kafka side.

@garyrussell , WDYT?

garyrussell:
Yes, fine with me, but I think we need some mention of it and an example in the reference manual (`src/reference/asciidoc`) and the what's new section.

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaStreamBrancher.java
```diff
@@ -0,0 +1,98 @@
...
+		this.defaultConsumer = Objects.requireNonNull(consumer);
+		return this;
+	}
+
+	/**
+	 * Terminating method that builds branches on top of given {@code KStream}.
+	 *
+	 * @param stream {@code KStream} to split
+	 */
+	public void onTopOf(KStream<K, V> stream) {
```
artembilan: I would suggest to have it as a method chain and return provided `stream`.
I mean for our Spring use-case like this:
```
		@Bean
		public KStream<String, String> trueFalseStream(StreamsBuilder streamsBuilder) {
			KStream<String, String> trueFalseStream = streamsBuilder
					.stream(TRUE_FALSE_INPUT_TOPIC, Consumed.with(Serdes.String(), Serdes.String()));

			return new KafkaStreamBrancher()
				.addBranch((key, value) -> String.valueOf(true).equals(value),
						ks ->ks.to(TRUE_TOPIC, Produced.with(Serdes.String(), Serdes.String())))
				.addBranch((key, value) -> String.valueOf(false).equals(value)),
						ks -> ks.to(FALSE_TOPIC, Produced.with(Serdes.String(), Serdes.String())))
				.onTopOf(trueFalseStream);
		}
```

See `KafkaStreamsBranchTests` which you can modify respectively.

inponomarev:
There is even [KAFKA-5488](https://issues.apache.org/jira/browse/KAFKA-5488) issue from 2017-- I'm not the first one who don't want to deal with current `branch` API :-) Ok, I posted a [PR#6164](https://github.com/apache/kafka/pull/6164) to Apache Kafka. Meanwhile,  I will add commits here to finalize this PR: documentation and `onTopOf` method.

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaStreamBrancher.java
```diff
@@ -0,0 +1,98 @@
...
+	private Consumer<? super KStream<K, V>> defaultConsumer;
+
+	/**
+	 * Defines a new branch.
+	 *
+	 * @param predicate {@link Predicate} instance
+	 * @param consumer  The consumer of this branch's {@code KStream}
+	 * @return {@code this}
+	 */
+	public KafkaStreamBrancher<K, V> addBranch(Predicate<? super K, ? super V> predicate,
```
artembilan: How about just have it like `branch()` as well as `defaultBranch()`?
Not too much value in the `add` verb 😄

spring-kafka/src/test/java/org/springframework/kafka/support/KafkaStreamsBrancherTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.apache.kafka.streams.kstream.KStream;
+import org.apache.kafka.streams.kstream.Predicate;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+class KafkaStreamsBrancherTest {
```
artembilan: We prefer `*Tests` suffix.

spring-kafka/src/test/java/org/springframework/kafka/support/KafkaStreamsBrancherTest.java
```diff
@@ -0,0 +1,59 @@
...
+	@Test
+	@SuppressWarnings({"unchecked", "rawtypes"})
+	void correctConsumersAreCalled() {
+		Predicate p1 = Mockito.mock(Predicate.class);
+		Predicate p2 = Mockito.mock(Predicate.class);
+		KStream input = Mockito.mock(KStream.class);
```
artembilan: You can just `import static` for `Mockito` methods.

spring-kafka/src/test/java/org/springframework/kafka/support/KafkaStreamsBrancherTest.java
```diff
@@ -0,0 +1,59 @@
...
+		KStream[] result = new KStream[]{Mockito.mock(KStream.class),
+				Mockito.mock(KStream.class), Mockito.mock(KStream.class)};
+		Mockito.when(input.branch(Mockito.eq(p1), Mockito.eq(p2), Mockito.any()))
+				.thenReturn(result);
+		AtomicInteger invocations = new AtomicInteger(0);
+		Assertions.assertSame(input, new KafkaStreamBrancher()
```
artembilan: I think JUnit 5 `Assertions` ca be statically imported as well.

spring-kafka/src/test/java/org/springframework/kafka/support/KafkaStreamsBrancherTest.java
inponomarev: I know, but static imports are prohibited by Checkstyle here  :smiley:

spring-kafka/src/test/java/org/springframework/kafka/support/KafkaStreamsBrancherTest.java
inponomarev: Me too, but Checktyle thinks different 😄

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/ab344af41a7a00167f3258711d8e0ae1400afaf5

@inponomarev,

Looking forward for more, but, please, consider in the future to raise pull requests against top level task-specific branches, not from the `master`.
Right now you have to `git reset --hard` your local `master` to pick up all the changes from the `upstream master`.

Issue #942, opened 2019-01-18
TopicExistsException printed - causing lots of noise in logs


magaton:
Latest change in KafkaAdmin (due to sonar warnings) where the Exception is thrown and printed in catch introduces lots of noise in logs in case that topic already exists.

garyrussell:
@magaton Can you provide an example? I don't see any new log messages when creating topics; there are existing INFO logs produced when the actual partition count doesn't match, but those are not new.

magaton:
I have 3 instances of monolith app starting at the same time and trying to create topic "events". 1 succeeds, 2 fail with the exception below.


KakaAdmin is declared as bean
`@Bean
	public KafkaAdmin admin() {
		Map<String, Object> configs = new HashMap<>();
		configs.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaBrokerProperties.getBootstrapServers());
		return new KafkaAdmin(configs);
	}`
The topic "events" is configured as 
	`@Bean
	public NewTopic inboundEvent() {
         // ...
        }`


Jan 18 12:44:42 bi-test-3 java[2847]: 12:44:42.101 monolith [main] ERROR o.s.kafka.core.KafkaAdmin - Failed to create topics
Jan 18 12:44:42 bi-test-3 java[2847]: org.apache.kafka.common.errors.TopicExistsException: Topic 'events' already exists.
Jan 18 12:44:42 bi-test-3 java[2847]: 12:44:42.102 monolith [main] ERROR o.s.kafka.core.KafkaAdmin - Could not configure topics
Jan 18 12:44:42 bi-test-3 java[2847]: org.springframework.kafka.KafkaException: Failed to create topics; nested exception is org.apache.kafka.common.errors.TopicExistsException: Topic 'events' alr
Jan 18 12:44:42 bi-test-3 java[2847]:         at org.springframework.kafka.core.KafkaAdmin.addTopics(KafkaAdmin.java:258)
Jan 18 12:44:42 bi-test-3 java[2847]:         at org.springframework.kafka.core.KafkaAdmin.addTopicsIfNeeded(KafkaAdmin.java:199)
Jan 18 12:44:42 bi-test-3 java[2847]:         at org.springframework.kafka.core.KafkaAdmin.initialize(KafkaAdmin.java:167)
[...]

PR #943, opened 2019-01-18
GH-942: KafkaAdmin Improvements

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/942

Suppress topic/partition already exist exceptions - possible race when multiple
instances of the same app start concurrently.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 the original author or authors.
+ * Copyright 2017-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -254,8 +257,13 @@ private void addTopics(AdminClient adminClient, List<NewTopic> topicsToAdd) {
...
 		}
 		catch (ExecutionException e) {
-			logger.error("Failed to create topics", e.getCause());
-			throw new KafkaException("Failed to create topics", e.getCause()); // NOSONAR
+			if (e.getCause() instanceof TopicExistsException) { // Possible race with another app instance
+				logger.debug("Failed to create topics", e.getCause());
+			}
+			else {
+				logger.error("Failed to create topics", e.getCause());
+				throw new KafkaException("Failed to create topics", e.getCause()); // NOSONAR
```
artembilan: I think Sonar is not going to like a repeated String.
Let's consider to extract it as a variable!
As well as for next one about partitions.

garyrussell: I think the threshold is 4 😄 

Will fix if he complains.

PR #955, opened 2019-01-31
Update to the build file to modernize the docs

Buzzardo:
This commit throws out Docbook and gives us the modern look and feel for the docs. It uses Asciidoctor to generate the PDF.


artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/f6caf5bb1efc13273dfa0f01179115462a580dca.

@Buzzardo ,

thank you for you contribution!

Issue #960, opened 2019-02-12
SeekToCurrentErrorHandler - commit offset of recovered records


garyrussell:
See https://stackoverflow.com/questions/53781327/spring-kafka-listener-using-deadletterpublishingrecoverer-and-manual-ack-mode/53781959#comment96087649_53781959

PR #961, opened 2019-02-12
GH-960: Seek-to-Current - commit recovered offset

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/960

When the container is configured with `AckMode.MANUAL_IMMEDIATE`, the
`SeekToCurrentErrorHandler` can be configured to commit the offset of
a recovered record.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/SeekToCurrentErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/SeekToCurrentErrorHandler.java
@@ -18,6 +18,7 @@ package org.springframework.kafka.listener;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.function.BiConsumer;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/SeekToCurrentErrorHandler.java
```diff
@@ -82,16 +91,58 @@ public SeekToCurrentErrorHandler(BiConsumer<ConsumerRecord<?, ?>, Exception> rec
...
+					consumer.commitSync(Collections.singletonMap(new TopicPartition(record.topic(), record.partition()),
+							new OffsetAndMetadata(record.offset() + 1)));
+				}
+				else {
+					OffsetCommitCallback commitCallback = container.getContainerProperties().getCommitCallback();
+					if (commitCallback == null) {
+						commitCallback = LOGGING_COMMIT_CALLBACK;
+					}
+					consumer.commitAsync(Collections.singletonMap(new TopicPartition(record.topic(), record.partition()),
+							new OffsetAndMetadata(record.offset() + 1)), commitCallback);
```
artembilan: Looks like `TopicPartition` and `OffsetAndMetadata` could be extracted as variables before `if..else`. We are going to use them in both cases anyway.

PR #962, opened 2019-02-12
Fix ErrorHandler for Consumer Errors

garyrussell:
When a container-aware error handler is called from a
`ConcurrentMessageListenerContainer`, when the consumer throws an
exception (rather than the listener), the wrong container was
passed in the container argument. The child container was passed
in instead of the concurrent container.

__cherry-pick to 2.1.x__


artembilan:

Issue #950, opened 2019-01-29
Resolve TODO in KMLC


garyrussell:
See https://github.com/spring-projects/spring-kafka/issues/948

Wrap all exceptions in an LEFE.

PR #966, opened 2019-02-15
GH-950: Always wrap listener exception in LEFE

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/950

Provide access to the `group.id` for all listener types, including
simple listeners.


artembilan:
Hm, i have just pushed that "What's New" change: https://github.com/spring-projects/spring-kafka/commit/acdc2686203e634d6ecbcc3bf40e973124989179
Didn't think that you are going to tackle it so fast 😄 

Please, consider to rebase and sorry for confuse with docs work...

garyrussell:
Yeah - I just figured I'd do it on the first PR that needs a doc change. np; fixing...

garyrussell:
Need to fix tests

PR #967, opened 2019-02-16
Fix some typos.

neetkee:


artembilan:
Great catches, Artem!

Please, compile the project via Gradle before before pushing commits here.
Or just run its `updateCopyrights` task directly! This way all your modified files are going to be updated in the Copyright part.
Although you need to revert the commit fist of all, so the files are not going to be treated as stashed. Otherwise the update is not going to happen.
Would be great to always run at at least `check` task locally before pushing to PRs.

Thanks for understand!

neetkee:
@artembilan pushed again, is it correct now?

artembilan:
👍 

Thank you very much; looking forward for more!

Issue #970, opened 2019-02-22
Add getGroupId() to MessageListenerContainer


joaomiguelsoares:
Hi, I want to create an API which receives the consumer group Id and I want to start/stop the corresponding listener container from reading messages from kafka.
How can I get all my listener container and it's consumer group for an app? Retrieving all the containers I can stop/start them, but until now I cannot find a way to retrieve all the Listener containers, unless my listeners are created through annotation **@KafkaListener**.

Can you help me?

**Affects Version(s):** 2.2.2.RELEASE

garyrussell:
Did you not read the issue template? GitHub issues are NOT for asking questions; use Stack Overflow for that (tagged spring-kafka). Issues are for reporting bugs or asking for new features.

You can get all containers created for `@KafkaListener` using the `KafkaListenerEndpointRegistry` bean.

You can get all containers defined as `@Bean`s from the `ApplicationContext` (`getBeansOfType`).

Once you have the container(s), to find the group id:

- `container.getContainerProperties().getGroupId()` (if a groupId was set on the container - or the `@KafkaListener` has an `id` or `groupId`).

If the groupId came from the consumer factory (it was not specifically set on the container) there is currently no way to get the group or the consumer factory properties from the container.

joaomiguelsoares:
Sorry for the inconvenience! 
I looked up of your implementation and I realized that the only way I can get the the groupId of a message listener is either by setting on the container Properties or by setting on the  `@KafkaListener`  annotation. If not, I could not match a message listener with a consumer group.
Another important feature that you might consider is: the `KafkaListenerEndpointRegistry` bean could, not only, register the listeners created by annotation, but  register the listeners created through @Bean.
On your project roadmap, do you have intentions to implement those features?

garyrussell:
We can certainly expose the actual group for the container as a property on the container itself (and we **will** do that in the next release scheduled for next month.

No, we can't change the registry to manage all containers, but we could, I suppose, add a convenience method, something like `getAllContainers()` which can present a merged collection of containers it manages together with the containers managed by the Spring application context.

joaomiguelsoares:
Nice, I'll wait for the next release to match all the containers with it's consumer group.

Yes, It was I nice to have if we can get all the containers on the registry.
Now, I will go for your suggestion, retrieve all containers by the applicationContext and by the registry, until the registry expose all the containers.
Do you have the intention to implement this on the next release?

Thanks in advance.

[...]

PR #971, opened 2019-02-22
GH-970: Add getGroupId() to containers

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/970

- allow retrieval of the `group.id`, even if not set on the container properties
- also add `getAllListenerContainers` to the `RLERegistry` as a convenience
- also add `getListenerId` to return the id or bean name of the container

__cherry-pick to 2.2.x__


artembilan:
... and cherry-picked to `2.2.x`.

Issue #781, opened 2018-08-14
Add commitSyncDuration


garyrussell:
From the Kafka documentation:

>Overloads have also been added for other KafkaConsumer methods like partitionsFor, listTopics, offsetsForTimes, beginningOffsets, endOffsets and close that take in a Duration.

artembilan:
"Non-blocking"? How do they mean?
```
public void close(Duration timeout) {
        if (timeout.toMillis() < 0)
            throw new IllegalArgumentException("The timeout cannot be negative.");
        acquire();
        try {
            if (!closed) {
                closed = true;
                close(timeout.toMillis(), false);
            }
        } finally {
            release();
        }
    }

...
  try {
            if (coordinator != null)
                coordinator.close(Math.min(timeoutMs, requestTimeoutMs));
        } catch (Throwable t) {
            firstException.compareAndSet(null, t);
            log.error("Failed to close coordinator", t);
        }
        ClientUtils.closeQuietly(fetcher, "fetcher", firstException);
        ClientUtils.closeQuietly(interceptors, "consumer interceptors", firstException);
        ClientUtils.closeQuietly(metrics, "consumer metrics", firstException);
        ClientUtils.closeQuietly(client, "consumer network client", firstException);
        ClientUtils.closeQuietly(keyDeserializer, "consumer key deserializer", firstException);
        ClientUtils.closeQuietly(valueDeserializer, "consumer value deserializer", firstException);
        AppInfoParser.unregisterAppInfo(JMX_PREFIX, clientId, metrics);
        log.debug("Kafka consumer has been closed");
```

I don't see any async interaction...

More over looks like all those non-`Duration` APIs just delegate to the `Duration`-based:
```
public List<PartitionInfo> partitionsFor(String topic) {
       return partitionsFor(topic, Duration.ofMillis(defaultApiTimeoutMs));
}
```

What am I missing? Why doesn't that look for me as non-blocking?

Thanks

garyrussell:
We should also decide whether a `TimeoutException` is fatal for `start()` when checking for the existence of the partitions.

artembilan:
OK. All sounds good. Although I don't think that `org.apache.kafka.common.errors.TimeoutException` should be considered in boundaries of this issue.
But I'm not so good in guessing what you have in mind about Kafka 😄 

Thanks for sharing your thoughts!

PR #972, opened 2019-02-22
GH-781: Add commitSyncTimeout container property

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/781

Also fix updateCopyright gradle task.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -463,6 +463,10 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 
 		private final boolean checkNullValueForExceptions;
 
+		private final boolean syncCommits = this.containerProperties.isSyncCommits();
+
+		private final Duration syncCommitTimeout;
+
 		private Map<TopicPartition, OffsetMetadata> definedPartitions;
 
 		private volatile Collection<TopicPartition> assignedPartitions;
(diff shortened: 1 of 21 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -557,6 +561,34 @@ else if (listener instanceof MessageListener) {
...
+							.get(ConsumerConfig.DEFAULT_API_TIMEOUT_MS_CONFIG);
+				}
+				if (timeout instanceof Number) {
+					return Duration.ofMillis(((Number) timeout).longValue());
+				}
+				else if (timeout instanceof String) {
+					return Duration.ofMillis(Long.parseLong((String) timeout));
+				}
+				else {
+					return Duration.ofSeconds(60);
```
artembilan: Can the `timeout` from properties be as a `Duration` as well? 
Why there is no such an `else if` then?
Why do you default it to `60 secs` and no any WARN that  `timeout` is not of expected type?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1711,12 +1743,16 @@ protected void doInTransactionWithoutResult(TransactionStatus status) {
...
-					else if (KafkaMessageListenerContainer.this.getContainerProperties().isSyncCommits()) {
-						ListenerConsumer.this.consumer.commitSync(offsetsToCommit);
-					}
 					else {
-						ListenerConsumer.this.consumer.commitAsync(offsetsToCommit,
-								KafkaMessageListenerContainer.this.getContainerProperties().getCommitCallback());
+						ContainerProperties containerProps = KafkaMessageListenerContainer.this.getContainerProperties();
+						if (containerProps.isSyncCommits()) {
+							ListenerConsumer.this.consumer.commitSync(offsetsToCommit,
+									containerProps.getSyncCommitTimeout());
```
artembilan: Why don't you call `determineSyncCommitTimeout()` from here?

spring-kafka/src/main/java/org/springframework/kafka/listener/SeekToCurrentErrorHandler.java
```diff
@@ -130,7 +130,7 @@ else if (this.commitRecovered) {
 						new TopicPartition(record.topic(), record.partition()),
 						new OffsetAndMetadata(record.offset() + 1));
 				if (container.getContainerProperties().isSyncCommits()) {
-					consumer.commitSync(offsetToCommit);
+					consumer.commitSync(offsetToCommit, container.getContainerProperties().getSyncCommitTimeout());
```
artembilan: Can't we call `determineSyncCommitTimeout()` here as well?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Good point; I default it to 60 seconds because that's what the client defaults it to if it is not provided. But, I agree, it should only be that if it's `null`.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Because I set the property if it was null.

garyrussell: Because I set the property if it was null - see line 565.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Actually, I don't think Kafka will accept `Duration` there, but I'll add the code anyway.

artembilan:
```
#### Conflicting files
`spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java`
```

And will be ready for merge!

PR #973, opened 2019-02-23
Fix Consumer Property Overrides

garyrussell:
If the consumer properties object contains `default` properties,
they were ignored. `Properties.forEach()` does not return default
properties.

Use `stringPropertyNames()` instead.

**cherry-pick to 2.2.x**


artembilan:

Issue #990, opened 2019-03-08
Option to commit offsets when publishing to dead letter topic.


miklesw:
**Affects Version(s):**  latest

When max retries is reached, the DefaultAfterRollbackProcessor uses seek to skip the failing record. This means that the record will be processed when the app is restarted.

To work around this I extended the DeadLetterPublishingRecoverer to also commit the offset.

Alternatively the DefaultAfterRollbackProcessor could be configured to commit the offsets after rolling back the main transaction.

https://stackoverflow.com/questions/55023117/kafkatemplate-in-transactional-kafkalistener-publishes-messages-when-transaction

```
@Transactional(KAFKA_TRANSACTION_MANAGER)
class OffsetCommittingAndDeadLetterPublishingRecoverer(val template: KafkaTemplate<Any, Any>) :
    DeadLetterPublishingRecoverer(template) {

    override fun accept(record: ConsumerRecord<*, *>, exception: Exception) {
        super.accept(record, exception)

        val topicPartition = TopicPartition(record.topic(), record.partition())
        val offsetAndMetadata = OffsetAndMetadata(record.offset() +1)

        template.sendOffsetsToTransaction(
                mapOf(topicPartition to offsetAndMetadata)
            )
        }
}
```

anschnapp:
Sorry to write to this already closed ticket, if you think i should open a new one please let me know.

But my problem fits perfectly to the title of this issue. So i want to have an explicit commit after a dead letter was send. I want to avoid sending a dlq duplicate (or more if the are in a row)  after a consumer restart.

I'm not sure if the solution inside of DefaultAfterRollbackProcessor works for me.
Because it seems that this solution is only for circumstances where kafka transactions are used.
Which is not the case in my scenario. I have a quite simple setup (but with manual direct offset) where i want commit after each processed entry (even on dlq handled ones).

garyrussell:
Commenting on old, closed, issues is not a good idea; especially when unrelated since you are not using transactions.

The default behavior in supported versions is to commit the offset after an error handler successfully "handles" the error - see `CommonErrorHandler.ackAfterHandle()`. 

If you are seeing different behavior, open a new issue, providing a minimal, complete, reproducible example.

anschnapp:
Thanks for answering @garyrussell 

You are right that this is not the best place.
Yes the topic is slightly different but the title was a complete fit and also most of the description (until the example appears ;))

But if i might ask you one more question:

[...]

PR #991, opened 2019-03-08
GH-990: Run AfterRollbackProcessor in Tx

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/990

Provide a mechanism to start a new transaction within which to invoke
the processor, so if it recovers the failed record, its offset can
be sent to the transaction.

**cherry-pick to 2.2.x**


artembilan:
... and cherry-picked to `2.2.x` as https://github.com/spring-projects/spring-kafka/commit/c226c9ee8bda4f28ee238917b5ab71f92f81c358 after fixing conflicts.

Issue #995, opened 2019-03-13
Memory leak in ErrorHandlingDeserializer2 ?


rahmanusta:
Hi,

I'm using Spring-Kafka 2.2.2, why ObjectOutputStream is not closed here? https://github.com/spring-projects/spring-kafka/blob/9dccd38b1f75f974a9162e6bb7581b9564bea4ad/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/ErrorHandlingDeserializer2.java#L203  

May it cause a memory leak?

garyrussell:
We should probably close it, but why do you think it will cause a memory leak?

The underlying `ByteArrayOutputStream.close()` does nothing.

garyrussell:
`writeObject` drains the `BlockDataOutputStream` to the underlying BAOS...

```java
        /**
         * Writes all buffered data from this stream to the underlying stream,
         * but does not flush underlying stream.
         */
        void drain() throws IOException {
 ```

`flush()` and `close()` are no-ops for BAOS so closing really has no effect.

rahmanusta:
Yes, it looks safe to not close. Thanks for the details. I'm closing this issue.

PR #996, opened 2019-03-13
GH-995: Close ObjectOutputStream in EHD2

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/995

Doesn't really make a difference because `writeObject()` drains
all data to the underlying output stream and `flush()` and `close()`
are no-ops for `ByteArrayOutputStream`.

But, technically, correct.

**cherry-pick to 2.2.x**


artembilan:

PR #997, opened 2019-03-13
Improve container test coverage

garyrussell:
From 78.5% to 83.8% and fix bugs found:

- Deprecate unused ctor
- Fix check for proper error handler for listener type

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/ErrorHandlingDeserializerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/ErrorHandlingDeserializerTests.java
@@ -171,7 +171,7 @@ public class ErrorHandlingDeserializerTests {
 		public ConsumerFactory<String, String> cf() {
 			Map<String, Object> props = KafkaTestUtils.consumerProps(TOPIC + ".g1", "false", embeddedKafka());
 			props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
-			props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, ErrorHandlingDeserializer2.class);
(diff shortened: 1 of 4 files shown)


spring-kafka/src/test/java/org/springframework/kafka/listener/ErrorHandlingDeserializerTests.java
```diff
@@ -171,7 +171,7 @@ else if (r.key() == null && t.getCause() instanceof DeserializationException) {
 		public ConsumerFactory<String, String> cf() {
 			Map<String, Object> props = KafkaTestUtils.consumerProps(TOPIC + ".g1", "false", embeddedKafka());
 			props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
-			props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, ErrorHandlingDeserializer2.class);
+			props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, ErrorHandlingDeserializer2.class.getName());
 			props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, ErrorHandlingDeserializer2.class);
 			props.put(ErrorHandlingDeserializer2.KEY_DESERIALIZER_CLASS, FailSometimesDeserializer.class);
```
artembilan: I wonder why we don't need to `getName()` for the `key` deserializer class...

garyrussell: Coverage:

```java
		private boolean checkDeserializer(Object deser) {
			return deser instanceof Class
					? ErrorHandlingDeserializer2.class.isAssignableFrom((Class<?>) deser)
					: deser instanceof String && deser.equals(ErrorHandlingDeserializer2.class.getName());
		}
```

We didn't have coverage for the String branch.

Issue #1026, opened 2019-03-23
Add an option to enhance the Headers to contain the consumer's group id


garyrussell:
When the same listener is used for multiple groups, it would be useful if the application could have access to which group a record came from.

By adding the group to the `ConsumerRecord.headers()` it will be mapped, and available in the `@KafkaListener` as ` @Header` parameter.

PR #1030, opened 2019-03-25
GH-1026: Support KafkaHeaders.GROUP_ID

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1026

Also improve `DefaultKafkaHeaderMapper` with a `NeverMatchHeaderMatcher`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListener.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListener.java
@@ -241,4 +241,13 @@ public @interface KafkaListener {
 	 */
 	String[] properties() default {};
 
+	/**
+	 * When set to true, enable the listener code to access the consumer's {@link code
+	 * group.id} property using a {@code @Header(KafkaHeaders.GROUP_ID} String groupId)
+	 * method parameter. Useful if you use the same listener code in multiple containers.
+	 * @return true to provide the header.
+	 * @since 2.3
+	 */
+	String exposeGroupId() default "false";
+
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListener.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListener.java
@@ -241,13 +241,4 @@ public @interface KafkaListener {
 	 */
 	String[] properties() default {};
 
-	/**
-	 * When set to true, enable the listener code to access the consumer's {@link code
-	 * group.id} property using a {@code @Header(KafkaHeaders.GROUP_ID} String groupId)
-	 * method parameter. Useful if you use the same listener code in multiple containers.
-	 * @return true to provide the header.
-	 * @since 2.3
-	 */
-	String exposeGroupId() default "false";
-
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -56,7 +56,6 @@ import org.springframework.kafka.KafkaException;
 import org.springframework.kafka.core.ConsumerFactory;
 import org.springframework.kafka.core.KafkaResourceHolder;
 import org.springframework.kafka.core.ProducerFactory;
-import org.springframework.kafka.core.ProducerFactoryUtils;
 import org.springframework.kafka.event.ConsumerPausedEvent;
 import org.springframework.kafka.event.ConsumerResumedEvent;
 import org.springframework.kafka.event.ConsumerStoppedEvent;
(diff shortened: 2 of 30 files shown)


spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListener.java
```diff
@@ -241,4 +241,13 @@
...
 	String[] properties() default {};
 
+	/**
+	 * When set to true, enable the listener code to access the consumer's {@link code
+	 * group.id} property using a {@code @Header(KafkaHeaders.GROUP_ID} String groupId)
+	 * method parameter. Useful if you use the same listener code in multiple containers.
+	 * @return true to provide the header.
+	 * @since 2.3
+	 */
+	String exposeGroupId() default "false";
```
artembilan: I wonder why would one doesn't want to map this property to header?..
Why don't we do that unconditionally like for many other Kafka records properties?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -743,9 +743,7 @@ public void run() {
 			if (this.genericListener instanceof ConsumerSeekAware) {
 				((ConsumerSeekAware) this.genericListener).registerSeekCallback(this);
 			}
-			if (this.transactionManager != null) {
-				ProducerFactoryUtils.setConsumerGroupId(this.consumerGroupId);
-			}
+			KafkaUtils.setConsumerGroupId(this.consumerGroupId);
```
artembilan: Looks like there is no a conditions to not set this value into the `ThreadLocal` any more...

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -44,26 +47,29 @@
...
-			new SimplePatternBasedHeaderMatcher("!" + KafkaHeaders.TIMESTAMP),
-			new SimplePatternBasedHeaderMatcher("!" + KafkaHeaders.TIMESTAMP_TYPE),
-			new SimplePatternBasedHeaderMatcher("!" + KafkaHeaders.BATCH_CONVERTED_HEADERS),
-			new SimplePatternBasedHeaderMatcher("!" + KafkaHeaders.NATIVE_HEADERS),
-			new SimplePatternBasedHeaderMatcher("!" + KafkaHeaders.TOPIC));
-
-	protected final List<SimplePatternBasedHeaderMatcher> matchers = new ArrayList<>(NEVER_MAPPED); // NOSONAR
+	protected final Log LOGGER = LogFactory.getLog(getClass()); // NOSONAR
+
+	private static final List<HeaderMatcher> NEVER_MAPPED = Collections.singletonList(
```
artembilan: Doesn't look like this constant is used any where.
Can't we just make it inline in the `matchers` bellow?

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -44,26 +47,29 @@
...
+			new NeverMatchHeaderMatcher(Arrays.stream(new String[] {
```
artembilan: I would move `toSet()` into the `NeverMatchHeaderMatcher` ctor having its argument as a varargs variant to even avoid an explicit array here.

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -200,13 +206,58 @@ private String mapRawIn(String header, byte[] value) {
 	}
 
 
+	/**
+	 * A matcher for headers.
+	 * @since 2.3
+	 */
+	protected interface HeaderMatcher {
```
artembilan: So, the idea to let have any custom matcher only if we extend the whole class.
 No way to reuse default one, but insert custom matchers.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/support/JavaUtils.java
```diff
@@ -0,0 +1,175 @@
...
+		}
+		return this;
+	}
+
+	/**
+	 * Invoke {@link Consumer#accept(Object)} with the value if it is not null or empty.
+	 * @param value the value.
+	 * @param consumer the consumer.
+	 * @return this.
+	 * @since 5.2
```
artembilan: I think all these `@since` on methods are just wrong for this new class in this project.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListener.java
garyrussell: Doh - good point 😄

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
garyrussell: I simply realized that having a pattern matcher for header names without a wildcard is much less efficient than looking up the header in a set.

The interface is so I can have 2 different types of matcher.

spring-kafka/src/main/java/org/springframework/kafka/support/JavaUtils.java
garyrussell: Doh - copy/paste.

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
artembilan: But my point that it is `protected` and therefore not available if we don't extend.
The implementation is fine though, only concern is about convenience of end-users.

PR #1054, opened 2019-03-29
Add EnhancedLogFactory

garyrussell:
- add `Supplier<Object>` log methods
- add tracing to `CloseSafeProducer`

If you agree with the approach, I'll change all logging.


garyrussell:
On hold - see https://github.com/spring-projects/spring-framework/issues/22726

PR #1061, opened 2019-04-05
Fix Sonar issue

garyrussell:


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -575,6 +568,24 @@ else if (parameterizedType.getRawType().equals(List.class)
...
+	}
+
+	private boolean isListOfConsumerRecordUpperBounded(Type paramType) {
+		return isWIldCardWithUpperBound(paramType)
+			&& ((WildcardType) paramType).getUpperBounds()[0] instanceof ParameterizedType
+			&& ((ParameterizedType) ((WildcardType) paramType).getUpperBounds()[0])
+						.getRawType().equals(ConsumerRecord.class);
+	}
+
+	private boolean isWIldCardWithUpperBound(Type paramType) {
```
artembilan: Looks like typo: W**i**ld ?

Issue #1008, opened 2019-03-19
ProducerConfig.TRANSACTIONAL_ID_CONFIG and ProducerFactory.transactionIdPrefix overlapping ?


jorgheymans:
From gitter:

Q: when i have this in the producer configuration  `props.put(TRANSACTIONAL_ID_CONFIG, "my-service");` , why do i still need to put a `transactionIdPrefix` on the producer factory ?
```java
DefaultKafkaProducerFactory<String, byte[]> producerFactory = new DefaultKafkaProducerFactory<>(producerConfig());
// otherwise transactions don't work
producerFactory.setTransactionIdPrefix("my-service");
```
Suggestion: ProducerFactory could source the prefix property from the ProducerConfig if not specified directly. The fact that one is a prefix and the other not is not so relevant from an end user perspective.

Gary Russell @garyrussell Mar 15 16:45
@jorgheymans Because we have to manage a pool of producers when using transactions, for concurrency; producers have to have unique transaction ids; hence it's a prefix.
Furthermore, when publishing on a consumer thread, to prevent zombie fencing, the transaction id has to include the group/topic/partition. If you don't care about zombie fencing (needed for exactly once semantics), you can set producerPerConsumerPartition to false, but we still need multiple txids for concurrency.
We could add some code to detect concurrent use of a single producer to allow that configuration, but what is the objection to using the factory property instead of the producer config?

Jorg Heymans @jorgheymans Mar 16 08:55
@garyrussell no objection, it's just confusing to me that both need to be set. From a user perspective it looks to be the same thing. Why not just take TRANSACTIONAL_ID_CONFIG as the default prefix for example?

garyrussell:
OK; if we detect a property but the setter is not called, we'll use the property as the prefix (and log a warning that it will get suffixes).

PR #1065, opened 2019-04-11
GH-1008: Use `transactional.id` config as a prefix

artembilan:
Fixes https://github.com/spring-projects/spring-kafka/issues/1008

* If provided `configs` for the `DefaultKafkaProducerFactory` contains a
`ProducerConfig.TRANSACTIONAL_ID_CONFIG` value, treat it as a `prefix`
and override for the target `Producer` configs.
Also log an `INFO` that this value is going to be prefix the target
`Producer`
* An explicit `setTransactionIdPrefix()` will override this value
any way
* Polishing for `KafkaTemplateTransactionTests` together with an
explicit `ProducerConfig.TRANSACTIONAL_ID_CONFIG` config in the
`testLocalTransaction()` test

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
@@ -55,6 +55,7 @@ import org.springframework.context.event.ContextStoppedEvent;
 import org.springframework.kafka.support.TransactionSupport;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * The {@link ProducerFactory} implementation for a {@code singleton} shared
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -126,16 +127,30 @@ public DefaultKafkaProducerFactory(Map<String, Object> configs) {
...
 			@Nullable Serializer<K> keySerializer,
 			@Nullable Serializer<V> valueSerializer) {
+
 		this.configs = new HashMap<>(configs);
 		this.keySerializer = keySerializer;
 		this.valueSerializer = valueSerializer;
+
+		String txId = (String) this.configs.get(ProducerConfig.TRANSACTIONAL_ID_CONFIG);
+		if (StringUtils.hasText(txId)) {
+			setTransactionIdPrefix(txId);
```
garyrussell: Needs to be made final when calling from a CTOR. Perhaps add a protected getter if a subclass wants to sniff the value.

artembilan: It is there. See the change on line 186.

I'm OK to add getter.

garyrussell: Doh, I missed it because of all the other noise 😦 sorry.

PR #1069, opened 2019-04-15
Use empty array in toArray call.

neetkee:
Seems like more performant version, at least in newer JDK versions.
See: https://github.com/spring-projects/spring-boot/pull/12153


artembilan:
Thank you!

Those have slipped somehow from our eyes 😄

Issue #998, opened 2019-03-14
Handle multiple responses with ReplyingKafkaTemplate


gkohen:
🎁 Enhancement
Continued thread from [Stackoverflow](https://stackoverflow.com/questions/55168639/handle-multiple-responses-with-replyingkafkatemplate).
I'm trying to implement a reply response pattern in which I publish a message to a topic listened to by several consumer groups. That means that they will all get the message as well as submit the response in the reply topic.

The problem is because they all respond to the same message, only the first received message in the reply topic shall be answered. The others will be discarded. Given that I know how many responses I should be getting on the reply topic(call that number-`n`), how can I make `ReplyingKafkaTemplate` wait for n responses and then resolve the answer? I've tried inferring from the Spring Kafka documentation but could not quite figure it out
 
There were few options suggested amongst one is to extend `ReplyingKafkaTemplate` to support the feature or implement a timer after which you can return partial responses. 
At this point `ReplyingKafkaTemplate` is not really built for extensions and might require almost wholesale copy of the whole class rather than extending it.

garyrussell:
I thought about it some more, and wonder why something like this wouldn't work...

```java
public class AggregatingReplyingKafkaTemplate<K, V, R> extends ReplyingKafkaTemplate<K, V, Collection<R>> {

	public AggregatingReplyingKafkaTemplate(ProducerFactory<K, V> producerFactory,
			GenericMessageListenerContainer<K, Collection<R>> replyContainer) {

		super(producerFactory, replyContainer);
	}

	@Override
	public void onMessage(List<ConsumerRecord<K, Collection<R>>> data) {
		List<ConsumerRecord<K, Collection<R>>> completed = new ArrayList<>();
		data.forEach(record -> {
			// We should move the following to a helper method
			Iterator<Header> iterator = record.headers().iterator();
			CorrelationKey correlationId = null;
			while (correlationId == null && iterator.hasNext()) {
				Header next = iterator.next();
				if (next.key().equals(KafkaHeaders.CORRELATION_ID)) {
					correlationId = new CorrelationKey(next.value());
				}
			}
			if (correlationId == null) {
				this.logger.error("No correlationId found in reply: " + record
						+ " - to use request/reply semantics, the responding server must return the correlation id "
						+ " in the '" + KafkaHeaders.CORRELATION_ID + "' header");
			}
			// end of helper method

			else {
				List<R> list = addToCollection(record, correlationId);
				if (list.size() == 5) {
[...]

PR #1072, opened 2019-04-19
GH-998: AggregatingReplyingKafkaTemplate

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/998

It abuses generics a type erasure a little, but works.

@artembilan Docs to follow if you are ok with it.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.requestreply;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.header.Header;
+import org.apache.kafka.common.header.internals.RecordHeader;
+
+import org.springframework.kafka.core.ProducerFactory;
+import org.springframework.kafka.listener.BatchConsumerAwareMessageListener;
+import org.springframework.kafka.listener.ContainerProperties.AckMode;
+import org.springframework.kafka.listener.GenericMessageListenerContainer;
+import org.springframework.kafka.support.KafkaHeaders;
+import org.springframework.util.Assert;
+
+/**
+ * A replying template that aggregates multiple replies with the same correlation id.
+ *
+ * @param <K> the key type.
+ * @param <V> the outbound data type.
+ * @param <R> the reply data type.
+ *
+ * @author Gary Russell
+ * @since 2.3
+ *
+ */
+public class AggregatingReplyingKafkaTemplate<K, V, R>
+		extends ReplyingKafkaTemplate<K, V, Collection<ConsumerRecord<K, R>>>
+		implements BatchConsumerAwareMessageListener<K, Collection<ConsumerRecord<K, R>>> {
+
+	private final ConcurrentMap<CorrelationKey, List<ConsumerRecord<K, R>>> pending = new ConcurrentHashMap<>();
+
+	private final Map<TopicPartition, Long> offsets = new HashMap<>();
+
+	private final Function<Collection<ConsumerRecord<K, R>>, Boolean> releaseStrategy;
+
+	private Duration commitTimeout = Duration.ofSeconds(30);
+
+	private boolean returnPartialOnTimeout;
+
+	/**
+	 * Construct an instance using the provided parameter arguments. The releaseStrategy
+	 * is consulted to determine when a collection is "complete".
+	 * @param producerFactory the producer factory.
+	 * @param replyContainer the reply container.
+	 * @param releaseStrategy the release strategy.
+	 */
+	public AggregatingReplyingKafkaTemplate(ProducerFactory<K, V> producerFactory,
+			GenericMessageListenerContainer<K, Collection<ConsumerRecord<K, R>>> replyContainer,
+			Function<Collection<ConsumerRecord<K, R>>, Boolean> releaseStrategy) {
+
+		super(producerFactory, replyContainer);
+		Assert.notNull(releaseStrategy, "'releaseStrategy' cannot be null");
+		AckMode ackMode = replyContainer.getContainerProperties().getAckMode();
+		Assert.isTrue(ackMode.equals(AckMode.MANUAL) || ackMode.equals(AckMode.MANUAL_IMMEDIATE),
+				"The reply container must have a MANUAL or MANUAL_IMMEDIATE AckMode");
+		this.releaseStrategy = releaseStrategy;
+	}
+
+	/**
+	 * Set the timeout to use when committing offsets.
+	 * @param commitTimeout the timeout.
+	 */
+	public void setCommitTimeout(Duration commitTimeout) {
+		Assert.notNull(commitTimeout, "'commitTimeout' cannot be null");
+		this.commitTimeout = commitTimeout;
+	}
+
+	/**
+	 * Set to true to return a partial result when a request times out.
+	 * @param returnPartialOnTimeout true to return a partial result.
+	 */
+	public void setReturnPartialOnTimeout(boolean returnPartialOnTimeout) {
+		this.returnPartialOnTimeout = returnPartialOnTimeout;
+	}
+
+	@Override
+	public void onMessage(List<ConsumerRecord<K, Collection<ConsumerRecord<K, R>>>> data, Consumer<?, ?> consumer) {
+		List<ConsumerRecord<K, Collection<ConsumerRecord<K, R>>>> completed = new ArrayList<>();
+		data.forEach(record -> {
+			Header correlation = record.headers().lastHeader(KafkaHeaders.CORRELATION_ID);
+			if (correlation == null) {
+				this.logger.error("No correlationId found in reply: " + record
+						+ " - to use request/reply semantics, the responding server must return the correlation id "
+						+ " in the '" + KafkaHeaders.CORRELATION_ID + "' header");
+			}
+			else {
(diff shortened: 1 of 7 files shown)


garyrussell:
Actually, I think I can avoid the generics abuse by making the RKT a delegate instead of a superclass. Will rework on Monday.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
```diff
@@ -0,0 +1,180 @@
...
+		});
+		if (completed.size() > 0) {
+			super.onMessage(completed);
+		}
+	}
+
+	private synchronized void checkOffsetCommits(List<ConsumerRecord<K, R>> list, Consumer<?, ?> consumer) {
+		list.forEach(record -> this.offsets.compute(
+			new TopicPartition(record.topic(), record.partition()), (k, v) -> record.offset()));
+		if (this.pending.isEmpty()) {
```
artembilan: Can we explain this in Docs that we commit offsets only when there is no pending any more?

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
```diff
@@ -0,0 +1,180 @@
...
+			consumer.commitSync(this.offsets.entrySet().stream()
+					.collect(Collectors.toMap(
+											entry -> entry.getKey(),
```
artembilan:
```suggestion
											Map.Entry::getKey,
```
?

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
garyrussell: Good point; while doing this, I realized we must weed out duplicate deliveries too.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
```diff
@@ -0,0 +1,281 @@
...
+			.collect(Collectors.toMap(entry -> entry.getKey(), entry -> entry.getValue()
+					.stream()
+					.map(holder -> holder.getRecord())
+					.collect(Collectors.toList())));
+		if (logger.isDebugEnabled()) {
+			logger.debug("Discarding " + orphaned + " most likely a partial redelivery of an already released group");
+		}
+		orphaned.values()
+			.forEach(list -> checkOffsets(list));
+		orphaned.keySet().forEach(corr -> this.pending.remove(corr));
```
artembilan: I would iterate here only once, over `entrySet()`

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
```diff
@@ -85,15 +86,20 @@
 
 	private boolean sharedReplyTopic;
 
+	private Function<ProducerRecord<K, V>, CorrelationKey> correlationStrategy =
```
artembilan: Not related to your change, but can we already resolve a TODO in this class on the `TemplateRequestReplyFuture` ?

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
```diff
@@ -0,0 +1,281 @@
...
+
+	private final Map<TopicPartition, Long> offsets = new HashMap<>();
+
+	private final Predicate<Collection<ConsumerRecord<K, R>>> releaseStrategy;
+
+	private Duration commitTimeout = Duration.ofSeconds(30);
+
+	private boolean returnPartialOnTimeout;
+
+	private volatile long lastOrphanCheck = System.currentTimeMillis();
```
artembilan: Doesn't look like this property is updated anywhere...

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
```diff
@@ -0,0 +1,281 @@
...
+	 * discard from time-to-time.
+	 * @param consumer the consumer.
+	 */
+	private void weedOrphans(Consumer<?, ?> consumer) {
+		Map<CorrelationKey, List<ConsumerRecord<K, R>>> orphaned = this.pending.entrySet()
+			.stream()
+			.filter(entry -> entry.getValue()
+					.stream()
+					.allMatch(holder -> holder.getTimestamp() > this.lastOrphanCheck))
+			.collect(Collectors.toMap(entry -> entry.getKey(), entry -> entry.getValue()
```
artembilan: Looking into this one more time, I would say we don't need to collect at all: there is just enough to *filter* and then `forEach`, without intermediate `List` and `Map`.

Also it might be even better to rethink this Java Streams usage and just use an `Iterator` from `this..pending.entrySet()`

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
```diff
@@ -0,0 +1,281 @@
...
+	 */
+	public void setReturnPartialOnTimeout(boolean returnPartialOnTimeout) {
+		this.returnPartialOnTimeout = returnPartialOnTimeout;
+	}
+
+	@Override
+	public void onMessage(List<ConsumerRecord<K, Collection<ConsumerRecord<K, R>>>> data, Consumer<?, ?> consumer) {
+		long now = System.currentTimeMillis();
+		if (now - this.lastOrphanCheck > getReplyTimeout() * 10) { // NOSONAR magic #
+			weedOrphans(consumer);
```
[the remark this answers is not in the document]
garyrussell: What would that add?

If we get a real timeout, `handleTimeout()` removes the orphaned replies for that specific request (optionally returning a partial result).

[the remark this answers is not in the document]
artembilan: Oh! Right. So, then the opposite story question: why do we need to check for orphans in the `onMessage` then?

Won't that `handleTimeout` for all the orphans cover us?

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
garyrussell: I collected them so I could log the discarded records. Or do you mean log them one correlation key at-a-time?

How would I do the `allMatch()` test if I use a simple iterator? I'd have to iterate over each list.

garyrussell: Pushed the other changes while we discuss this further.

artembilan: > Pushed the other changes

Looks like not yet.

OK. If the logging is the point, then never mind. Let's live with collecting!

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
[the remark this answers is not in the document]
garyrussell: No; consider this case - request with 2 replies...

- request1
- request2
- request3
- reply1a
- reply2a
- reply3a
- reply1b (complete group)
- request2 times out (and clears out reply2a)
- rebalance

Since we had a pending request3, we didn't commit the offsets for the replies to request1.

After the rebalance, we get

- reply1a
- reply2a
- reply3a
- reply1b (complete group)
- reply3b (complete group)

request1 replies will be passed to super.onMessage, but that's ok, it will log 

```java
"No pending reply: " + record + WITH_CORRELATION_ID
				+ correlationId + ", perhaps timed out, or using a shared reply topic";
```

reply3b will complete request3.

But the reply2a redelivery will stay forever because request2 already timed out so `handleTimeout()` won't be called again.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
garyrussell: Sorry; committed but didn't push.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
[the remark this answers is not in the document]
artembilan: Huh? Why should we worry about `reply2a` then if there is no `request2` for it any more?

We should log "late reply" and discard. What else am I missing?

Your explanation is good, but I'm not sure what is the point to deal with poison replies...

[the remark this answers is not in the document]
garyrussell: Ah, of course! - I already have `isPending(correlationId)` later in `onMessage()` - I should check it earlier 😄

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
```diff
@@ -0,0 +1,247 @@
...
+				}
+			}
+		});
+		if (completed.size() > 0) {
+			super.onMessage(completed);
+		}
+	}
+
+	@Override
+	protected synchronized boolean handleTimeout(CorrelationKey correlationId,
```
artembilan: Why is this? 
Looks like such this method is called only from a single scheduled thread.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
```diff
@@ -0,0 +1,247 @@
...
+											entry -> new OffsetAndMetadata(entry.getValue()))),
+								this.commitTimeout);
+			this.offsets.clear();
+		}
+	}
+
+	@SuppressWarnings({ "rawtypes", "unchecked" })
+	private Set<RecordHolder<K, R>> addToCollection(ConsumerRecord record, CorrelationKey correlationId) {
+		Set<RecordHolder<K, R>> set = this.pending.computeIfAbsent(correlationId, id -> new LinkedHashSet<>());
+		set.add(new RecordHolder<>(record));
```
artembilan: Looks like we don't need `RecordHolder` any more...

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
garyrussell: `pending` needs to be protected.

Consider this race...

Thread in `onMessage()` calls `isPending(correlationId)` (true)
Timeout thread fires and removes correlationId from `pending`
`onMessage()` thread adds new delivery to `pending`.

Result: orphan

spring-kafka/src/main/java/org/springframework/kafka/requestreply/AggregatingReplyingKafkaTemplate.java
garyrussell: We need `RecordHolder` because we avoid duplicates via a `Set`. `ConsumerRecord` does not implement `hashCode()` and `equals()`. `RecordHolder` `hashCode()` and `equals()` are based on topic/partition/offset.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/f27a3ae7dac3b83f90857ea9df1d813180f8491e after some minor polishing.

PR #1073, opened 2019-04-20
typo correction

javatutorials2016:


javatutorials2016:
minor correction on the sub-heading text

src/reference/asciidoc/quick-tour.adoc
```diff
@@ -122,7 +122,7 @@ private Map<String, Object> senderProps() {
 ----
 ====
 
-==== With Java Configuration
+==== With Spring Configuration in Java
```
garyrussell: It's not clear why this is needed; it's certainly not a typo. "Spring" is implied since this is a Spring project.

javatutorials2016: sure, The example followed is using spring configuration, so I thought it be a more relevant title.

artembilan: Well, we definitely talk in this document only about Spring. If the story is about something else, we are more specific.

It is really called `Java Configuration` in Spring world: https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#beans-java

I don't think we are going to change the text flow.

So, sorry, but closed as `Invalid`. There is just no typo in this sentence.

artembilan:
Closed as `Invalid`: The *Java Configuration* is a strong term for the `@Configuration` and `@Bean` combinations in Spring Framework.

PR #1078, opened 2019-04-26
Use LogAccessor for logging

garyrussell:
- use lambdas instead of testing logging level.
- add tracing to `CloseSafeProducer`


igouss:
I'm curious why?

Issue #1082, opened 2019-05-02
Add Kafka Stream exception handlers that push to DLQ


adrien-ben:
Hi,

It would be nice to have implementations of Kafka Streams' `DeserializationExceptionHandler` and `ProductionExceptionHandler` that push to dead-letter queue and enhance the record the way `DeadLetterPublishingRecoverer` does.

garyrussell:
This turns out to be a little bit involved because you can't pass live objects in `StreamsConfig` properties.

I created a `DeadLetterPublishingRecoverer` as a `@Bean` so the `DeserializationExceptionHandler` implementation can invoke it but by the time we get to `DeserializationExceptionHandler.configure()`, the property has had a `toString()` operation performed on it 😦 .

It has to be a live object because it needs a template to do the publishing; it would be a pain to have to construct a template from properties in `configure()`.

I will see if I can come up with a reasonable work-around.

sobychacko:
@garyrussell I actually hit some problems when implementing custom deserialization handlers in the Kafka Streams binder. I ended up doing some manual commits in the DLQ handler that the binder provides. Here are some details about the issues I ran into:
https://stackoverflow.com/questions/48470899/kafka-streams-deserialization-handler
which lead to the creation of this issue in the Kafka jira: https://issues.apache.org/jira/browse/KAFKA-6502

garyrussell:
@adrien-ben I am not sure what we can do on the producer side. What do you have in mind?

adrien-ben:
@garyrussell Sorry, I'm no sure I understand what you mean by "what we can do on the producer side".

garyrussell:
@adrien-ben 

>It would be nice to have implementations of Kafka Streams' `DeserializationExceptionHandler` and `ProductionExceptionHandler` that push to dead-letter queue 

If we get a `SerializationException` on the producer side (handled by a `ProductionExceptionHandler`), how can we "push to a dead-letter queue"; we would need to serialize the bad data somehow.

Thanks.

adrien-ben:
Oh yeah sorry, I though you were talking about the producer that pushes to the DLQ.

Actually I'm not even sure that the `ProductionExceptionHandler` catches serialization exception since the `handle` method already provides the serialized record. It's not clear in the documentation, I'll dig a bit in the Kafka Stream source and come back to you.

garyrussell:
[...]

PR #1083, opened 2019-05-02
GH-1082: Dead Letter Streams Deserialization

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1082

Needs docs.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
@@ -187,7 +187,7 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
-		this.contexts.put(applicationContext.getId(), applicationContext);
+		contexts.put(applicationContext.getId(), applicationContext);
 	}
 
 	/**
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
@@ -19,7 +19,6 @@ package org.springframework.kafka.config;
 import java.time.Duration;
 import java.util.Map;
 import java.util.Properties;
-import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.streams.KafkaClientSupplier;
@@ -30,9 +29,7 @@ import org.apache.kafka.streams.Topology;
 import org.apache.kafka.streams.processor.StateRestoreListener;
 import org.apache.kafka.streams.processor.internals.DefaultKafkaClientSupplier;
 
-import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.config.AbstractFactoryBean;
-import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.SmartLifecycle;
 import org.springframework.core.log.LogAccessor;
@@ -70,8 +67,6 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 	private static final String STREAMS_CONFIG_MUST_NOT_BE_NULL = "'streamsConfig' must not be null";
 
 	private static final String CLEANUP_CONFIG_MUST_NOT_BE_NULL = "'cleanupConfig' must not be null";
-
-	private static final Map<String, ApplicationContext> contexts = new ConcurrentHashMap<>();
 
 	private KafkaClientSupplier clientSupplier = new DefaultKafkaClientSupplier();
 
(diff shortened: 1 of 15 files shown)


spring-kafka/src/test/java/org/springframework/kafka/streams/RecoveringDeserializationExceptionHandlerTests.java
```diff
@@ -0,0 +1,280 @@
...
+			props.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, this.brokerAddresses);
+			props.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.ByteArraySerde.class);
+			props.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, FailSerde.class);
+			props.put(StreamsConfig.DEFAULT_TIMESTAMP_EXTRACTOR_CLASS_CONFIG,
+					WallclockTimestampExtractor.class.getName());
+			props.put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, "100");
+			props.put(StreamsConfig.DEFAULT_DESERIALIZATION_EXCEPTION_HANDLER_CLASS_CONFIG,
+					RecoveringDeserializationExceptionHandler.class);
+			props.put(RecoveringDeserializationExceptionHandler.KSTREAM_DESERIALIZATION_CONTEXT_ID, context.getId());
+			props.put(RecoveringDeserializationExceptionHandler.KSTREAM_DESERIALIZATION_RECOVERER, "recoverer");
```
artembilan: And why we can't just store a `recoverer()` instance in the properties to configure streams?

spring-kafka/src/main/java/org/springframework/kafka/streams/RecoveringDeserializationExceptionHandler.java
```diff
@@ -0,0 +1,113 @@
...
+	public static final String KSTREAM_DESERIALIZATION_CONTEXT_ID = "spring.deserialization.context.id";
+
+	/**
+	 * Property name for the bean name of the recoverer.
+	 */
+	public static final String KSTREAM_DESERIALIZATION_RECOVERER = "spring.deserialization.recoverer";
+
+	private static final Log LOGGER = LogFactory.getLog(RecoveringDeserializationExceptionHandler.class);
+
+	private ConsumerRecordRecoverer recoverer;
```
artembilan: `final` and `@Nullable` and also initialized as a `null` in the ctor.

spring-kafka/src/test/java/org/springframework/kafka/streams/RecoveringDeserializationExceptionHandlerTests.java
garyrussell: It doesn't work - see my comment in the issue.

spring-kafka/src/main/java/org/springframework/kafka/streams/RecoveringDeserializationExceptionHandler.java
garyrussell: ? It's assigned in `configure()`.

artembilan: Ah! That's all about `toString()` on objects in that case.

And there is no way to pass object at all:
```
public DeserializationExceptionHandler defaultDeserializationExceptionHandler() {
        return getConfiguredInstance(DEFAULT_DESERIALIZATION_EXCEPTION_HANDLER_CLASS_CONFIG, DeserializationExceptionHandler.class);
    }
```

Then OK: let's live with what you have so far. Until Apache Kafka will come up with the setters on the `KafkaStreams`...

garyrussell: Actually, the problem is ours - `KafkaStreamsConfiguration.asProperties()` - we are doing conversion there. However, they are using the `Properties` as a general Map, with any object type. It's a bit of an abuse but I think we can change that method to just `putAll()`.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -66,6 +71,8 @@
 
 	private static final String CLEANUP_CONFIG_MUST_NOT_BE_NULL = "'cleanupConfig' must not be null";
 
+	private static final Map<String, ApplicationContext> contexts = new ConcurrentHashMap<>();
```
artembilan: Look like we don't this logic any more.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -52,7 +53,8 @@
  *
  * @since 1.1.4
  */
-public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilder> implements SmartLifecycle {
+public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilder>
+		implements SmartLifecycle, ApplicationContextAware {
```
artembilan: This is not an `ApplicationContextAware` any more...

PR #1089, opened 2019-05-06
HeaderEnricher: Add test using stream test utils

garyrussell:
Looks like I pushed this a few seconds too late 😦


artembilan:
Np. Waiting for green and then merge.

PR #1090, opened 2019-05-06
KafkaStreams <-> Spring Messaging

garyrussell:


spring-kafka/src/main/java/org/springframework/kafka/streams/messaging/MessagingTransformer.java
```diff
@@ -0,0 +1,95 @@
...
+
+	private final MessagingFunction function;
+
+	private final MessagingMessageConverter converter;
+
+	private ProcessorContext processorContext;
+
+	public MessagingTransformer(MessagingFunction function, MessagingMessageConverter converter) {
+		this.function = function;
+		this.converter = converter;
```
artembilan: `Assert.notNull()`? For both?

src/reference/asciidoc/si-kafka.adoc
```diff
@@ -681,6 +681,47 @@ public void in(@Header(KafkaHeaders.RECEIVED_MESSAGE_KEY) String key,
...
+
+    stream.print(Printed.toSysOut());
+
+    return stream;
+}
+
+@Bean
+@DependsOn("flow")
+public MessagingTransformer<String, String, String> transformer)
+        transformer(MessagingFunction function) {
```
artembilan: Looks like typo with duplicated `transformer`.

An interface in the `IntegrationFlows.from()` becomes as a bean with the name of the flow plus `.gateway` suffix.
so, it can be used in the `@Qualifier` for such an autowired parameter.

garyrussell:
Thanks; it can be merged.

PR #1098, opened 2019-05-16
Doc: Add Tx Synchronization Example

garyrussell:


src/reference/asciidoc/whats-new.adoc
```diff
@@ -4,6 +4,11 @@ This section covers the changes made from version 2.2 to version 2.3.
 
 Also see <<new-in-sik>>.
 
+==== Tips, Tricks and Examples
+
+A new chapter <<tips-n-tricks>> has been added.
+Please submit GitHub issues and/or pull requests for additional entries in that chapter.
```
artembilan: Do we mentioned `samples` somewhere in the Docs: https://github.com/spring-projects/spring-kafka/tree/master/samples ?

How about to make you JDBC sample in the doc as a whole sample project in the repo?

garyrussell: Yes, I should add it there too and we should reference those samples here.

I don't think it hurts to show the code here as well, though.

artembilan: OK. Let's address that in the separate issue!

Issue #1107, opened 2019-05-30
Provide a JUnit 5 extension equivalent of EmbeddedKafkaRule


wilkinsona:
For those that wish to use JUnit 5, it would be helpful to have a JUnit 5 extension that is the equivalent of `EmbeddedKafkaRule`. Spring Boot 2.2 is switching to using JUnit 5 by default in `spring-boot-starter-test` while including the Vintage Engine to allow JUnit 4 to be used during a migration period. This means that Boot 2.2 users will be able to use `EmbeddedKafkaRule` but it may prevent a complete migration to JUnit 5.

garyrussell:
JUnit5 is already supported via the `@EmbeddedKafka` test annotation.

artembilan:
I think our `@EmbeddedKafka` meets all the requirements for JUnit 5 with Spring:

```
@SpringJUnitConfig
@DirtiesContext
@EmbeddedKafka
public class KafkaStreamsCustomizerTests {
...
}
```

Or is your idea to really have a fully blown JUnit 5 extension, so we don't need requiring Spring Test environment?

wilkinsona:
I was looking for a replacement for this usage of `EmbeddedKafkaRule`: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfigurationIntegrationTests.java#L55-L57. I don't think `@EmbeddedKafka` fits the bill there, does it?

garyrussell:
All `@EmbeddedKafka` does is add a `@Bean` to the test config.

You can simply add a `EmbeddedKafkaBroker` `@Bean` to `KafkaConfig` there.

But, I agree, it would be more convenient to add more formal JUnit support.

garyrussell:
@wilkinsona https://github.com/spring-projects/spring-boot/pull/17034

PR #1108, opened 2019-05-31
GH-1107: Improved JUnit5 Support

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1107


artembilan:
That `ExecutionCondition` is too complicated for implementation, but I guess that's way how JUnit 5 works with all those contexts and parameter resolutions.

👍

Issue #1111, opened 2019-06-03
Starting paused `ConcurrentMessageListenerContainer` results in messages being consumed


bgK:
**Affects Version(s):** 2.2.6 (probably 2.3.x as well)

**Use case:**
I need to manually control the start offset for topic partitions while still using topic subscriptions; with multiple instances of the application sharing the same groupId.

The solution I plan to use is as follows:
* Start all the instances of the application but without starting the CMLCs.
* Have an external system transmit the desired start offsets to all the instances. Pause all the CMLCs. Start them all. In a `ConsumerSeekAware` message listener, seek the topic partitions in the `onPartitionsAssigned` callback.
* Once the consumer group is stabilized, resume all the CMLCs to start message consumption.

This works mostly fine but I have encountered two issues:

**The paused state is not propagated to KafkaMessageListenerContainer on start:**
```java
ConcurrentMessageListenerContainer cmlc = ...;
cmlc.pause();
cmlc.start();
```

After executing this code, `ConcurrentMessageListenerContainer` is paused, but the created `KafkaMessageListenerContainer` are not. 

It's seemingly possible to fix this issue by applying the following modification in `ConcurrentMessageListenerContainer`:

```java
protected void doStart() {
...
	KafkaMessageListenerContainer<K, V> container = new KafkaMessageListenerContainer<>();
...
	if (isPaused()) {
		container.pause();
	}
...
	container.start();
	this.containers.add(container);
}
```

**The first call to KafkaConsumer#poll still returns messages:**
`KafkaMessageListenerContainer.ListenerConsumer.ListenerConsumerRebalanceListener#onPartitionsAssigned` is indirectly called by `KafkaMessageListenerContainer.ListenerConsumer#pollAndInvoke`, via `consumer.poll`.

In my case, the KMLC is paused when the rebalance occurs and thus this code is executed:
```java
	ListenerConsumer.this.consumerPaused = false;
	ListenerConsumer.this.logger.warn("Paused consumer resumed by Kafka due to rebalance; "
		+ "the container will pause again before polling, unless the container's "
		+ "'paused' property is reset by a custom rebalance listener");
```

However, the call to `consumer.poll` that caused the rebalance listener to be called still returns messages.

As a quick test, I changed the code above to:
```java
	ListenerConsumer.this.consumer.pause(partitions);
```

This yields the desired behavior. The call to `consumer.poll` does not return messages.
There may be other implications to this change I don't understand ...

[...]

PR #1112, opened 2019-06-03
Fix re-pausing consumer after a rebalance

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/1111

`pause()` the consumer in the rebalance listener so that the Consumer
will discard any fetched records that would otherwise be returned by
the initial poll.

**cherry-pick to 2.1.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -1739,10 +1739,9 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			@Override
 			public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
 				if (ListenerConsumer.this.consumerPaused) {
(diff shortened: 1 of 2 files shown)


artembilan:
Do we still cherry-pick to `2.2.x`?
Thanks

artembilan:
Well, it's not cherry-pick'able into `2.1.x` - need a separate PR.
Or we just don't back-port it into `2.1.x` and `2.2.x` one is enough.
Nevertheless let me know what is in your mind!
Thanks

artembilan:
OK. Looks like according the issue state, we really talk about back-porting only to `2.2.x`, not `2.1.x`.

Nevermind. let's consider that as a typo in your comment! 😉

garyrussell:
I always seem to do that (switching between Kafka and AMQP). Yes, I meant 2.2.x - updated.

artembilan:
Cherry-picked to `2.2.x` as https://github.com/spring-projects/spring-kafka/commit/a5bfe1b7987c8d114ebc3c4f3cfa50e3205d5a94 after fixing conflicts in the `KafkaMessageListenerContainerTests`.

Issue #1080, opened 2019-04-27
DefaultConsumerFactory designed to use single deserializer instances shared across multiple consumers


christopher-gilbert:
This relates to an issue I posted on [StackOverflow](https://stackoverflow.com/questions/55782540/should-abstractlistenercontainerfactory-really-close-the-consumer-it-uses-to-che). Essentially, [DefaultKafkaConsumerFactory ](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java) has a single instance of each deserializer (key and value) which is shared between all consumers that it creates. The problem is that consumers have a lifecycle which cascades down to the deserializers - if a consumer is closed then it calls close on the deserializers, and as a consequence a consumer can be subsequently created by the factory with a closed deserializer.

My initial thought for addressing this in a backward compatible way would be to move the bulk of code from [DefaultKafkaConsumerFactory](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java) into a new class which allows clients to set a `Supplier<Deserializer>` rather than a Deserializer. [DefaultKafkaConsumerFactory ](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java)would retain the same interface and behaviour by delegating to the new class passing a Supplier that just returns the single Deserializer instance. 

Gary Russell commented on StackOverflow with a different approach involving a factory for DeSerializers/Serializers and SerDes which is a bigger solution than that required for the single issue raised, but would obviously have a wider applicability and be potentially more useful. If I submit a contribution, I will carefully consider the pros and cons of each approach - any comments on the subject are very welcome.

garyrussell:
I am fine with your suggestion of a simple delegating facade, but I believe it applies to the `Serializer`s too.

garyrussell:
Further to my original suggestion, if we go with a single factory, I think it should only be for the key/value serializer/deserializer (don't include the streams `SerDe`s because `kafka-streams` is only an optional dependency and we don't want to make it a hard dep.

christopher-gilbert:
Hi Gary,

[...]

PR #1117, opened 2019-06-09
Addressing issue GH-1080

christopher-gilbert:
I caused myself a bit of a merge headache by leaving this a while, but all merged in now. Note that while the DefaultKafkaConsumerFactory can delegate to the KafkaConsumeFactory class (which contains the original DefaultKafkaConsumerFactory functionality), the DefaultProducerFactory has to extend instead in order to retain backward compatibility with the protected access to the CloseSafeProducer.

Although there are quite a few apparent changes in this pull request, most of the changes to existing code are just moves. I did refactor the client id prefixing and suffixing a bit to make it clearer in KafkaConsumerFactory, and extended the unit test coverage to ensure the refactored code was honouring the Javadoc in ConsumerFactory.

This is my first PR so please call out as many things as you can. All usage of DefaultKafkaConsumerFactory and DefaultKafkaProducerFactory should be completely unaffected by these changes, including any protected methods (which are probably only used by unit tests, but I think at one time or another we've all accessed protected members of external libraries - not to be recommended!!

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
@@ -36,6 +36,7 @@ import org.springframework.kafka.listener.BatchErrorHandler;
 import org.springframework.kafka.listener.ContainerProperties;
 import org.springframework.kafka.listener.ErrorHandler;
 import org.springframework.kafka.listener.GenericErrorHandler;
+import org.springframework.kafka.listener.RecordInterceptor;
 import org.springframework.kafka.listener.adapter.RecordFilterStrategy;
 import org.springframework.kafka.listener.adapter.ReplyHeadersConfigurer;
 import org.springframework.kafka.requestreply.ReplyingKafkaOperations;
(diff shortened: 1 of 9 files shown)


christopher-gilbert:
IntelliJ annoyingly inserts spaces to line up code even when indentation is with tabs. I'll go through and get rid of the checkstyle failures.

christopher-gilbert:
test to fix - didn't run locally because of problems pulling down jacoco agent behind work proxy

artembilan:
Such a name of the Pull Request sounds like a spam.

Please, consider to learn how commit message and PR description is important in the whole contribution flow: https://chris.beams.io/posts/git-commit/

The point is that we may merge your contribution as is and then eventually we won't understand what was the reason of these changes.

Thanks for understanding.

artembilan:
> didn't run locally because of problems pulling down jacoco agent behind work proxy

We are fully against such an approach. It is not clear why would one need to deal with Jacoco when testing some Spring Kafka code.

christopher-gilbert:
Apologies for the title - I tried to keep commit messages clear and meaningful, but didn't do the same with the PR

christopher-gilbert:
The issue with Jacoco is an issue with building locally, and what I mean is that the PR is effectively premature - should not have been raised until completely successful build - my apologies for this

PR #1128, opened 2019-06-19
GH-1080 allow supplier functions for Serializers and Deserializers

christopher-gilbert:
Provides an alternative way of ensuring each consumer/producer has its own instances if configuration is not an option.

Also extend DefaultKafkaConsumerFactory unit test coverage to ensure compliance with the ConsumerFactory documented behaviour.

see https://github.com/spring-projects/spring-kafka/issues/1080 for more detail

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
@@ -20,6 +20,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
+import java.util.function.Supplier;
 
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
@@ -30,10 +31,20 @@ import org.springframework.lang.Nullable;
 import org.springframework.util.StringUtils;
 
 /**
- * The {@link ConsumerFactory} implementation to produce a new {@link Consumer} instance
- * for provided {@link Map} {@code configs} and optional {@link Deserializer} {@code keyDeserializer},
- * {@code valueDeserializer} implementations on each {@link #createConsumer()}
+ * The {@link ConsumerFactory} implementation to produce new {@link Consumer} instances
+ * for provided {@link Map} {@code configs} and optional {@link Deserializer}s on each {@link #createConsumer()}
  * invocation.
+ * <p>
+ * If you are using {@link Deserializer}s that have no-arg constructors and require no setup, then simplest to
+ * specify {@link Deserializer} classes in spring.kafka.consumer configuration and create the
+ * {@link DefaultKafkaConsumerFactory} with just a map of configs.
+ * <p>
+ * If that is not possible, but you are using {@link Deserializer}s that may be shared between all {@link Consumer}
+ * instances (and specifically that their close() method is a no-op), then you can pass in {@link Deserializer}
+ * instances for one or both of the key and value deserializers.
+ * <p>
+ * If neither of the above is true then you may provide a {@link Supplier} function for one or both {@link Deserializer}s
+ * which will be used to obtain {@link Deserializer}(s) each time a {@link Consumer} is created by the factory.
  *
  * @param <K> the key type.
  * @param <V> the value type.
@@ -41,21 +52,22 @@ import org.springframework.util.StringUtils;
  * @author Gary Russell
  * @author Murali Reddy
  * @author Artem Bilan
+ * @author Chris Gilbert
  */
 public class DefaultKafkaConsumerFactory<K, V> implements ConsumerFactory<K, V> {
 
 	private final Map<String, Object> configs;
 
-	private Deserializer<K> keyDeserializer;
+	private Supplier<Deserializer<K>> keyDeserializerSupplier;
 
-	private Deserializer<V> valueDeserializer;
+	private Supplier<Deserializer<V>> valueDeserializerSupplier;
 
 	/**
 	 * Construct a factory with the provided configuration.
 	 * @param configs the configuration.
 	 */
 	public DefaultKafkaConsumerFactory(Map<String, Object> configs) {
-		this(configs, null, null);
+		this(configs, () -> null, () -> null);
 	}
 
 	/**
@@ -67,17 +79,29 @@ public class DefaultKafkaConsumerFactory<K, V> implements ConsumerFactory<K, V> 
 	public DefaultKafkaConsumerFactory(Map<String, Object> configs,
 			@Nullable Deserializer<K> keyDeserializer,
 			@Nullable Deserializer<V> valueDeserializer) {
+		this(configs, () -> keyDeserializer, () -> valueDeserializer);
+	}
+
+	/**
+	 * Construct a factory with the provided configuration and deserializer suppliers.
+	 * @param configs the configuration.
+	 * @param keyDeserializerSupplier   the key {@link Deserializer} supplier function.
+	 * @param valueDeserializerSupplier the value {@link Deserializer} supplier function.
+	 */
+	public DefaultKafkaConsumerFactory(Map<String, Object> configs,
+									@Nullable Supplier<Deserializer<K>> keyDeserializerSupplier,
+									@Nullable Supplier<Deserializer<V>> valueDeserializerSupplier) {
 		this.configs = new HashMap<>(configs);
-		this.keyDeserializer = keyDeserializer;
-		this.valueDeserializer = valueDeserializer;
+		this.keyDeserializerSupplier = keyDeserializerSupplier == null ? () -> null : keyDeserializerSupplier;
+		this.valueDeserializerSupplier = valueDeserializerSupplier == null ? () -> null : valueDeserializerSupplier;
 	}
 
 	public void setKeyDeserializer(@Nullable Deserializer<K> keyDeserializer) {
-		this.keyDeserializer = keyDeserializer;
+		this.keyDeserializerSupplier = () -> keyDeserializer;
 	}
 
 	public void setValueDeserializer(@Nullable Deserializer<V> valueDeserializer) {
-		this.valueDeserializer = valueDeserializer;
+		this.valueDeserializerSupplier = () -> valueDeserializer;
 	}
 
 	@Override
@@ -87,12 +111,12 @@ public class DefaultKafkaConsumerFactory<K, V> implements ConsumerFactory<K, V> 
 
 	@Override
 	public Deserializer<K> getKeyDeserializer() {
-		return this.keyDeserializer;
+		return this.keyDeserializerSupplier.get();
 	}
 
 	@Override
 	public Deserializer<V> getValueDeserializer() {
-		return this.valueDeserializer;
+		return this.valueDeserializerSupplier.get();
 	}
 
 	@Override
@@ -161,7 +185,7 @@ public class DefaultKafkaConsumerFactory<K, V> implements ConsumerFactory<K, V> 
 	}
 
 	protected KafkaConsumer<K, V> createKafkaConsumer(Map<String, Object> configProps) {
-		return new KafkaConsumer<>(configProps, this.keyDeserializer, this.valueDeserializer);
+		return new KafkaConsumer<>(configProps, this.keyDeserializerSupplier.get(), this.valueDeserializerSupplier.get());
 	}
 
 	@Override
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
@@ -36,8 +36,8 @@ import org.springframework.util.StringUtils;
  * invocation.
  * <p>
  * If you are using {@link Deserializer}s that have no-arg constructors and require no setup, then simplest to
- * specify {@link Deserializer} classes against {@value ConsumerConfig#KEY_DESERIALIZER_CLASS_CONFIG} and
- * {@value ConsumerConfig#VALUE_DESERIALIZER_CLASS_CONFIG} keys in the {@code configs} passed to the
+ * specify {@link Deserializer} classes against {@link ConsumerConfig#KEY_DESERIALIZER_CLASS_CONFIG} and
+ * {@link ConsumerConfig#VALUE_DESERIALIZER_CLASS_CONFIG} keys in the {@code configs} passed to the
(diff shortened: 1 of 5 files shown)


christopher-gilbert:
Ah yes, that's true - I should really have added those in a separate commit. Basically there was a lot of existing untested behaviour around manipulation of the consumer and group ids at the point of creating a consumer, so I added the tests in the spirit of leaving the campsite cleaner. 

I will add to the adoc and commit tomorrow (UK time)

garyrussell:
LGTM too; thanks!! and thanks for all the new tests.

christopher-gilbert:
Hi, I have added some adoc, and I hope it is clear, and in the most appropriate places but if not please do change it directly or let me know any necessary corrections.

Many thanks, Chris

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
```diff
@@ -30,32 +31,43 @@
...
 /**
- * The {@link ConsumerFactory} implementation to produce a new {@link Consumer} instance
- * for provided {@link Map} {@code configs} and optional {@link Deserializer} {@code keyDeserializer},
- * {@code valueDeserializer} implementations on each {@link #createConsumer()}
+ * The {@link ConsumerFactory} implementation to produce new {@link Consumer} instances
+ * for provided {@link Map} {@code configs} and optional {@link Deserializer}s on each {@link #createConsumer()}
  * invocation.
+ * <p>
+ * If you are using {@link Deserializer}s that have no-arg constructors and require no setup, then simplest to
+ * specify {@link Deserializer} classes in spring.kafka.consumer configuration and create the
```
artembilan: First of all the `spring.kafka.consumer` should be wrapped to `{@code}` to make it more visible.
However on the other hand it is wrong to say about this property since it is part of Spring Boot. 
I think we need to be more specific here and really say about a native Apache Kafka property we are aiming to - `ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG` and/or `ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG`.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -64,8 +65,24 @@
  * <p>
  * This implementation will return the same {@link Producer} instance (if transactions are
  * not enabled) for the provided {@link Map} {@code configs} and optional {@link Serializer}
- * {@code keySerializer}, {@code valueSerializer} implementations on each
- * {@link #createProducer()} invocation.
+ * implementations on each {@link #createProducer()} invocation.
+ * <p>
+ * If you are using {@link Serializer}s that have no-arg constructors and require no setup, then simplest to
+ * specify {@link Serializer} classes in spring.kafka.producer configuration and create the
```
artembilan: Same here about wrong `spring.kafka.producer` reference where we need to mention appropriate Apache Kafka `ProducerConfig` props.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultKafkaConsumerFactoryTests.java
```diff
@@ -59,18 +65,149 @@
...
+		}, null);
+		assertThat(target.getKeyDeserializer()).isNotSameAs(target.getKeyDeserializer());
+	}
+
+	@Test
+	public void testNoOverrides() {
+		Map<String, Object> originalConfig = Collections.emptyMap();
+		new DefaultKafkaConsumerFactory<String, String>(originalConfig) {
+			protected KafkaConsumer<String, String> createKafkaConsumer(Map<String, Object> configProps) {
+				assertThat(configProps).isEqualTo(originalConfig);
```
artembilan: I don't think this is correct place to perform assertions. We really need to have a code flow like `given, when, then`.

I think you can store a `config` in some `AtomicReference` to reach it in the end of test.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultKafkaConsumerFactoryTests.java
```diff
@@ -59,18 +65,149 @@
...
+	}
+
+	@Test
+	public void testGroupIdOnPropertyOverride() {
+		Map<String, Object> originalConfig = Collections.singletonMap(ConsumerConfig.GROUP_ID_CONFIG, "original");
+		Properties overrides = new Properties();
+		overrides.setProperty(ConsumerConfig.GROUP_ID_CONFIG, "property-overridden");
+		new DefaultKafkaConsumerFactory<String, String>(originalConfig) {
+			protected KafkaConsumer<String, String> createKafkaConsumer(Map<String, Object> configProps) {
+				assertThat(configProps.get(ConsumerConfig.GROUP_ID_CONFIG)).isEqualTo("overridden");
```
artembilan: Same test code flow concern in all your new tests.

src/reference/asciidoc/kafka.adoc
```diff
@@ -295,6 +315,27 @@ IMPORTANT: When `producerPerThread` is `true`, user code **must** call `closeThr
...
+or `Serializer` instances may be passed to the `DefaultKafkaProducerFactory` constructor (in which case all Producers share the same instances).
+Alternatively you can provide Suppliers of Serializers that will be used to obtain separate `Serializer` instances for each `Producer`:
+
+====
+[source, java]
+----
+
+@Bean
+public ProducerFactory<Integer, CustomValue> producerFactory() {
+    return new DefaultKafkaProducerFactory<>(producerConfigs(), null, () -> new CustomValueSerializer();
```
artembilan: Missed closed bracket.

src/reference/asciidoc/kafka.adoc
```diff
@@ -164,6 +164,26 @@ public KafkaTemplate<Integer, String> kafkaTemplate() {
 ----
 ====
 
+or if you are using Spring Boot, you can make use of the `KafkaProperties` ConfigurationProperties class to pull in application properties, rather than manually adding them to a map:
```
artembilan: This doesn't say anything  valuable to me.
Why would we mention Spring Boot in this case?

src/reference/asciidoc/kafka.adoc
```diff
@@ -295,6 +315,27 @@ IMPORTANT: When `producerPerThread` is `true`, user code **must** call `closeThr
 This will physically close the producer and remove it from the `ThreadLocal`.
 Calling `reset()` or `destroy()` will not clean up these producers.
 
+When creating a `DefaultKafkaProducerFactory`, key and/or value `Serializer` classes can be picked up from configuration by calling the constructor that only takes in a Map of properties (see example in <<kafka-template>>),
+or `Serializer` instances may be passed to the `DefaultKafkaProducerFactory` constructor (in which case all Producers share the same instances).
+Alternatively you can provide Suppliers of Serializers that will be used to obtain separate `Serializer` instances for each `Producer`:
```
artembilan: Don't "`Serializer`s" and "`Serializer`s" work here?

src/reference/asciidoc/kafka.adoc
```diff
@@ -295,6 +315,27 @@ IMPORTANT: When `producerPerThread` is `true`, user code **must** call `closeThr
 This will physically close the producer and remove it from the `ThreadLocal`.
 Calling `reset()` or `destroy()` will not clean up these producers.
 
+When creating a `DefaultKafkaProducerFactory`, key and/or value `Serializer` classes can be picked up from configuration by calling the constructor that only takes in a Map of properties (see example in <<kafka-template>>),
+or `Serializer` instances may be passed to the `DefaultKafkaProducerFactory` constructor (in which case all Producers share the same instances).
```
artembilan: *One sentence per line*, please - https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

src/reference/asciidoc/whats-new.adoc
```diff
@@ -14,11 +14,16 @@ Please submit GitHub issues and/or pull requests for additional entries in that
...
-==== Producer Factory Changes
+==== Producer and Consumer Factory Changes
 
 The `DefaultKafkaProducerFactory` can now be configured to create a producer per thread.
 See <<producer-factory>> for more information.
 
+You can also provide `Serializer` suppliers in the constructor as an alternative to either configured classes (which require no-arg constructors)
+or constructing with `Serializer` instances, which are then shared between all Producers.
+
+The same option is also available with `Deserializer` suppliers in `DefaultKafkaConsumerFactory` - see <<kafka-container>>
```
artembilan: Missed period in the end.

src/reference/asciidoc/kafka.adoc
```diff
@@ -2357,6 +2414,7 @@ For more complex or particular cases, the `KafkaConsumer` (and, therefore, `Kafk
 constructors to accept `Serializer` and `Deserializer` instances for `keys` and `values`, respectively.
 
 When you use this API, the `DefaultKafkaProducerFactory` and `DefaultKafkaConsumerFactory` also provide properties (through constructors or setter methods) to inject custom `Serializer` and `Deserializer` instances into the target `Producer` or `Consumer`.
+Also, you can pass in Supplier functions for custom `Serializer` and `Deserializer` instances through constructors - these functions are called on creation of each `Producer` or `Consumer`
```
artembilan: Missed period in the end.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
christopher-gilbert: Yes, that makes sense. I'll change it.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultKafkaConsumerFactoryTests.java
christopher-gilbert: Ah, good idea - the target code doesn't lend itself easily to accessing the config that gets passed to the KafkaConsumer and I did feel that my tests looked a bit hacky, so yes, assigning the value to a variable referenced outside the overridden method and asserting after sounds like a better way to go.

At one point I did refactor the DefaultKafkaConsumerFactory code to make it a bit more easy to test this behaviour, but I have not pushed those changes as I don't want to get into any extensive refactoring. I'll just update these tests.

Personally I'm not enjoying going back to JUnit as I've been using Spock for the last few years which is much better for clearly defined tests with explicit given when then clauses!

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultKafkaConsumerFactoryTests.java
christopher-gilbert: agreed

src/reference/asciidoc/kafka.adoc
christopher-gilbert: Thanks for spotting that

src/reference/asciidoc/kafka.adoc
christopher-gilbert: True, and happy to remove it. I was just struggling to find a good Spring reference for using spring-kafka with Spring Boot. I see lots of teams in my organisation either writing loads of code manually copying properties wired in from application.yml into a map even though they could just use KafkaProperties, or else going out and copy/pasting any old code they find on StackOverflow. Reading through this documentation I can see how it would easily lead to the former approach.

christopher-gilbert:
Sorry, just realised that marking comments as resolved implies they have been addressed - my mistake, they are not resolved until I commit changes

src/reference/asciidoc/kafka.adoc
christopher-gilbert: Ah good, I wasn't sure - will apply the highlighting

src/reference/asciidoc/kafka.adoc
christopher-gilbert: Thanks for spotting

christopher-gilbert:
Hi @artembilan I have made some changes - grateful if you would review again, thanks, Chris

src/reference/asciidoc/whats-new.adoc
artembilan: The period is still missed 😄

artembilan: Can you say like this  "`Supplier<Deserializer>` isntances" ?
It is really what we expect from end-user to provide, so why don't get a generics syntax benefit in the docs as well? 😉

src/reference/asciidoc/kafka.adoc
```diff
@@ -2357,6 +2392,7 @@ For more complex or particular cases, the `KafkaConsumer` (and, therefore, `Kafk
 constructors to accept `Serializer` and `Deserializer` instances for `keys` and `values`, respectively.
 
 When you use this API, the `DefaultKafkaProducerFactory` and `DefaultKafkaConsumerFactory` also provide properties (through constructors or setter methods) to inject custom `Serializer` and `Deserializer` instances into the target `Producer` or `Consumer`.
+Also, you can pass in `Supplier` functions for custom `Serializer` and `Deserializer` instances through constructors - these functions are called on creation of each `Producer` or `Consumer`.
```
artembilan: `Supplier` is not a `Function`. See my comment below about generics opportunity.

src/reference/asciidoc/kafka.adoc
```diff
@@ -810,7 +829,23 @@ containerProps.setMessageListener(new MessageListener<Integer, String>() {
...
+                        new DefaultKafkaConsumerFactory<>(consumerProps());
+KafkaMessageListenerContainer<Integer, String> container =
+                        new KafkaMessageListenerContainer<>(cf, containerProps);
+return container;
+----
+====
+
+Note that when creating a `DefaultKafkaConsumerFactory`, using the constructor that just takes in the properties as above means that key and value `Deserializer` classes are picked up from configuration.
+Alternatively, `Deserializer` instances may be passed to the `DefaultKafkaConsumerFactory` constructor for key and/or value, in which case all Consumers share the same instances.
+Another option is to provide Suppliers of Deserializers that will be used to obtain separate `Deserializer` instances for each `Consumer`:
```
artembilan: If you'd use `Supplier<Deserializer>` here, you wouldn't miss that it is not a code.

christopher-gilbert:
Excellent, thanks Artem. I'll get those changes in this evening or tomorrow UK time.

Best wishes, Chris

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/d85040a1851e2f11bded265b23d05484dcebd872 after some code style polishing and fixing formatting in docs.

@christopher-gilbert ,

Issue #1129, opened 2019-06-20
A MimeTypeSerializer to store MimeType as a string in JSON is needed.


hanishi:
**Affects Version(s):**  2.2.4

When a Spring Cloud Stream's `http` Source app writes a `Message<?>` whose `MessageHeaders` includes `List<MediaType>` using kafka-binder to a topic assigned as its `output` destination, it is serialized and stored in it as follows: 

```
accept=[{"type":"text","subtype":"html","parameters":{},"qualityValue":1.0,"wildcardType":false,"wildcardSubtype":false,"charset":null,"concrete":true},{"type":"application","subtype":"xhtml+xml","parameters":{},"qualityValue":1.0,"wildcardType":false,"wildcardSubtype":false,"charset":null,"concrete":true},{"type":"application","subtype":"xml","parameters":{"q":"0.9"},"qualityValue":0.9,"wildcardType":false,"wildcardSubtype":false,"charset":null,"concrete":true},{"type":"image","subtype":"webp","parameters":{},"qualityValue":1.0,"wildcardType":false,"wildcardSubtype":false,"charset":null,"concrete":true},{"type":"image","subtype":"apng","parameters":{},"qualityValue":1.0,"wildcardType":false,"wildcardSubtype":false,"charset":null,"concrete":true},{"type":"*","subtype":"*","parameters":{"q":"0.8"},"qualityValue":0.8,"wildcardType":true,"wildcardSubtype":true,"charset":null,"concrete":false},{"type":"application","subtype":"signed-exchange","parameters":{"v":"b3"},"qualityValue":1.0,"wildcardType":false,"wildcardSubtype":false,"charset":null,"concrete":true}]
```

This record is deserialized as a `java.util.LinkedHashMap` in `MessageHeaders`, which makes the `setHttpHeader` private method of `DefaultHttpHeaderMapper` fail with the following **java.lang.IllegalArgumentException**,  when a receiving app tries to build a `HttpEntity` from a `Message<?>`

```
Caused by: java.lang.IllegalArgumentException: Expected MediaType or String value for 'Accept' header value, but received: class java.util.LinkedHashMap
    at org.springframework.integration.http.support.DefaultHttpHeaderMapper.setHttpHeader(DefaultHttpHeaderMapper.java:593) ~[spring-integration-http-5.1.3.RELEASE.jar!/:5.1.3.RELEASE]
[...]

PR #1130, opened 2019-06-20
GH-1129: Add JacksonMimeTypeModule

artembilan:
Fixes https://github.com/spring-projects/spring-kafka/issues/1129

Introduce a `JacksonMimeTypeModule` with a simple `MimeTypeSerializer`
for the proper inter-platform `MimeType` objects serialization.
Essentially we call its `toString()` which is enough to carry mime-type
info over the network.
Register this module in the `JacksonUtils.enhancedObjectMapper()` which
is used from the `DefaultKafkaHeaderMapper`.
Such a module can be registered as a bean in the application context
and will be automatically discovered by Spring Boot auto-configuration
for Jackson
* Modify a `DefaultKafkaHeaderMapper` to not deal with `MimeType`
directly any more since it is covered by the `JacksonMimeTypeModule`
even if `MimeType` is a part of collection like it is in case of
mapped HTTP headers.
* Modify `DefaultKafkaHeaderMapperTests` to check that `MimeType` is
properly serialized into its string representation even if it in the
collection

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/JacksonMimeTypeModule.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/JacksonMimeTypeModule.java
@@ -32,7 +32,7 @@ import com.fasterxml.jackson.databind.module.SimpleModule;
  *
  * @since 2.3
  */
-public class JacksonMimeTypeModule extends SimpleModule {
+public final class JacksonMimeTypeModule extends SimpleModule {
 
 	private static final long serialVersionUID = 1L;
 
@@ -45,6 +45,10 @@ public class JacksonMimeTypeModule extends SimpleModule {
 	 * target JSON as a plain string.
 	 */
 	private static final class MimeTypeSerializer extends JsonSerializer<MimeType> {
+
+		MimeTypeSerializer() {
+			super();
+		}
 
 		@Override
 		public void serialize(MimeType value, JsonGenerator generator, SerializerProvider serializers)
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/JacksonMimeTypeModule.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/JacksonMimeTypeModule.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/JacksonMimeTypeModule.java
```diff
@@ -0,0 +1,58 @@
...
+	public JacksonMimeTypeModule() {
+		addSerializer(MimeType.class, new MimeTypeSerializer());
+	}
+
+	/**
+	 * Simple {@link JsonSerializer} extension to represent a {@link MimeType} object in the
+	 * target JSON as a plain string.
+	 */
+	private static final class MimeTypeSerializer extends JsonSerializer<MimeType> {
+
```
garyrussell: package-protected ctor needed to avoid synthesized ctor with extra arg.

garyrussell:
Travis (checkstyle).

garyrussell:
NPE in `testTrustedAndNot` (travis)

garyrussell:
I wonder if the deserializer should now (2.3) always decode to String?

artembilan:
> I wonder if the deserializer should now (2.3) always decode to String?

No, that's OK. See my latest commit.

The problem is that even if we serialize `MimeType` as a `String` we still store `MimeType` class in the `JSON_TYPES` header.

So, if another side is still `DefaultKafkaHeaderMapper` it is going to deserialize that into `MimeType` object. 
This problem has a place only if `MimeType` is top level header. If it is an item in the collection, there is choice unless represent it as a plain string.
That's why Jackson always has a `containerType`, `elementType` and also `keyType` for maps.
We store in `JSON_TYPES` only top-level classes.

Not perfect, but enough so far...
Not sure what is an ideal solution, but in Spring Integration we use `init(JsonTypeInfo.Id.CLASS, null).inclusion(JsonTypeInfo.As.PROPERTY);` for `ObjectMapper.DefaultTypeResolverBuilder` to include a type info into the JSON node for particular object.
It may be ignored on the other side, (e.g. by non-Java consumer) or can be used to reconstruct instances properly.
We use such a trick in the `GenericJackson2JsonRedisSerializer` for `RedisMessageStore`.

Just some thoughts to share...

garyrussell:
This broke backwards compatibility; a 3.0 app can't talk to a 2.1 app because the MimeType is no longer a String.

Issue #1134, opened 2019-06-21
KafkaListenerEndpointRegistry.stop(Runnable) does not call callback when there are no containers to stop


wilkinsona:
Boot's tests for its Kafka auto-configuration are taking 30 seconds each to run as we have a context with a `KafkaListenerEndpointRegistry` that has no `MessageListenerContainer`s. This causes it to fail to call the `callback` that's passed into `stop(Runnable)` so we have to wait for Framework's 30 second timeout to pop.

artembilan:
Good catch, Andy!

Will fix soon...

artembilan:
Do you need the fix until Spring Boot 1.5.x ?

wilkinsona:
Boot 1.5 doesn't appear to be affected. For example, [this test](https://build.spring.io/browse/BOOT-PUB7-JOB1-1417/test/case/217332935) takes less than 1 second. That may be due to the tests being different in 1.5, I'm not sure.

artembilan:
Yeah... I think so. We will back-port any way until the version which is pulled by Initializr (seems for me 1.3.x), because we haven't committed into `1.1.x` for a while already...

garyrussell:
Yeah I found and fixed this in my experimental reactive `@KafkaListener` because in that environment, the MLC list was empty.

https://github.com/spring-projects/spring-kafka/pull/1123/files#diff-f23a91d9c56afd55946555b6ec1a150fR331

wilkinsona:
Thanks, both. What's your schedule for the next Spring Kafka maintenance releases? I'd like to switch to snapshots if you have releases planned before the next round of Boot releases. We have 2.2 M5 scheduled for 24 July. Given the Framework releases planned for 17 July, we'll probably have 1.5 and 2.1 releases around then as well.

wilkinsona:
I've figured out why we're only seeing the 30 second delay in Boot 2.2.

It is due to https://github.com/spring-projects/spring-kafka/issues/1102. In Boot 2.1 (Spring Kafka 2.2), `isRunning()` only returns `true` if one of the MLCs is running. As there are no MLCs in the affected tests, it returns `false`. As a result, `stop(Runnable)` is never called and the 30 second delay doesn't occur.

garyrussell:
Thanks Andy, yes we'll have a new milestone (M4). I've scheduled it for July 23 (it's the day before Artem goes on extended PTO, so we may go a few days earlier - as long as SF stays where it is).

[...]

PR #1135, opened 2019-06-21
GH-1134: Fix `stop(Runnable)` usage

artembilan:
Fixes https://github.com/spring-projects/spring-kafka/issues/1134

We always have to call `callback` in the `stop(Runnable)` implementation
independently of the component state

**Cherry-pick until 1.3.x to support Spring Boot 1.5.x**


artembilan:
@garyrussell ,

Looks like the Spring Boot 1.5.x is not affected, but to avoid an issue in the end-user applications would be better to back-port at least until `1.3.x`.

Thanks

garyrussell:
New Boot 1.5.x apps from initializr gets 1.3.x. Versions before 1.3.x are no longer supported so the backport stops there.

garyrussell:
Cherry picked to all (back to 1.3.x).

Issue #1142, opened 2019-06-26
Support to accept more than one RecordInterceptors in AbstractKafkaListenerContainerFactory 


rajeshhereforyou:
Hi,

As of now, we are using spring-kafka 2.2.7.RELEASE and using 'RecordInterceptor' interface and 'setRecordInterceptor(RecordInterceptor<K, V> recordInterceptor) to intercept one consumer record. 

But this method (setRecordInterceptor(RecordInterceptor<K, V> recordInterceptor) of 'AbstractKafkaListenerContainerFactory ' can accept only one RecordInterceptor and we are looking for a method which can take list of interceptors. 

Can we add a method/ modify the existing method to a list of interceptors?

artembilan:
Why just don't have a `CompositeRecordInterceptor` and delegate to all your particular `RecordInterceptor` s ?
I mean there is just nothing to do in the Framework since you really can do that on your side.

Thanks for understanding

rajeshhereforyou:
Let me explain our use-case. We have two different custom libraries. One is for tracing the request id, parent request id etc where we have created a custom RecordInterceptor by implementing RecordInterceptor interface and added our custom tracing logic in it and setting this custom RecordInterceptor while creating the 'ConcurrentKafkaListenerContainerFactory' bean 

The other library is purely for tracing health and performance metrics of an app. 

Since these are two separate concerns and two different libraries, we are looking for more than one CustomRecordInterceptors.

I believe i've explained the scenario in detail for your understanding. Please let me know if that helps.

artembilan:
Right. So, what is wrong with the `CompositeRecordInterceptor`:
```
public class CompositeRecordInterceptor<K, V> implements RecordInterceptor<K, V> {
	
	private final Collection<RecordInterceptor<K, V>> delegates = new ArrayList<>();

	public CompositeRecordInterceptor(RecordInterceptor<K, V>... delegates) {
		this.delegates.addAll(Arrays.asList(delegates));
	}

	@Override
	public ConsumerRecord<K, V> intercept(ConsumerRecord<K, V> record) {
		ConsumerRecord<K, V> recordToIntercept = record;
		for (RecordInterceptor<K, V> delegate : delegates) {
			recordToIntercept = delegate.intercept(recordToIntercept);
		}
		return recordToIntercept;
	}

}
```
As you see I don't say anything about "CustomRecordInterceptor".

Hope that helps.

garyrussell:
[...]

PR #1143, opened 2019-06-26
GH-1142: Add CompositeRecordInterceptor

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1142

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeRecordInterceptor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeRecordInterceptor.java
@@ -22,11 +22,14 @@ import java.util.Collection;
 
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 
+import org.springframework.util.Assert;
+
 /**
  * A {@link RecordInterceptor} that delegates to one or more {@link RecordInterceptor} in
  * order.
  *
  * @author Artem Bilan
+ * @author Gary Russell
  * @since 2.3
  *
  */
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeRecordInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+
+/**
+ * A {@link RecordInterceptor} that delegates to one or more {@link RecordInterceptor} in
+ * order.
+ *
+ * @author Artem Bilan
```
artembilan: Your name here as well since you have added JavaDocs and other required annotation 😉

spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeRecordInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+ * @since 2.3
+ *
+ */
+public class CompositeRecordInterceptor<K, V> implements RecordInterceptor<K, V> {
+
+	private final Collection<RecordInterceptor<K, V>> delegates = new ArrayList<>();
+
+	@SafeVarargs
+	public CompositeRecordInterceptor(RecordInterceptor<K, V>... delegates) {
+		this.delegates.addAll(Arrays.asList(delegates));
```
artembilan: I think we need to `Assert` at least for `null`

spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeRecordInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+	@Override
+	public ConsumerRecord<K, V> intercept(ConsumerRecord<K, V> record) {
+		ConsumerRecord<K, V> recordToIntercept = record;
+		for (RecordInterceptor<K, V> delegate : delegates) {
+			recordToIntercept = delegate.intercept(recordToIntercept);
+		}
+		return recordToIntercept;
+	}
+
+}
```
artembilan: New like in the end of file

garyrussell: :( I have a new STS and didn't configure

artembilan:
See Travis failures.

Thanks

PR #1150, opened 2019-07-01
AckMode.MANUAL_IMMEDIATE - wake Consumer

garyrussell:
When processing manual acks on a "foreign" thread, the commits are
not made until the consumer wakes from his `poll()`.

For `AckMode.MANUAL_IMMEDIATE`, we should wake the consumer so the
offset(s) are committed as soon as possible.

**cherry-pick to 2.2.x**


artembilan:
... and cherry-picked to `2.2.x`.

PR #1154, opened 2019-07-04
testing.adoc example fix

jorgheymans:


artembilan:
Good catch! Thank you for the fix; looking forward for more!

PR #1155, opened 2019-07-08
Use StreamsBuilder#build(Properties) to allow topology optimization

denisw:
Since Kafka Streams 2.1, StreamsBuilder has a `build(Properties)` method whose use is required to benefit from [topology optimization](https://docs.confluent.io/current/streams/developer-guide/optimizing-streams.html) (i.e., the `topology.optimization=all`). 

However, currently `StreamsBuilderFactoryBean` calls the older `build()` without arguments, which, since 2.3, doesn't do any topology optimization at all. (In 2.1.x and 2.2.x, it at least did the KTable
source changelog reuse optimization if `topology.optimization` is set on the `KafkaStreams` level.)

With this PR, we always use `build(Properties)` and thus allow Spring Kafka to make full use of topology optimization by just setting `topology.optimization`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
@@ -49,6 +49,7 @@ import org.springframework.util.Assert;
  * @author Soby Chacko
  * @author Zach Olauson
  * @author Nurettin Yilmaz
+ * @author Denis Washington
  *
  * @since 1.1.4
  */
@@ -290,7 +291,12 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 			try {
 				Assert.state(this.streamsConfig != null || this.properties != null,
 						"'streamsConfig' or streams configuration properties must not be null");
-				Topology topology = getObject().build(); // NOSONAR
+				Properties topologyProps = this.properties;
+				if (topologyProps == null) {
+					topologyProps = new Properties();
+					topologyProps.putAll(this.streamsConfig.originals());
+				}
+				Topology topology = getObject().build(topologyProps); // NOSONAR: getObject() cannot return null
 				LOGGER.debug(() -> topology.describe().toString());
 				if (this.properties != null) {
 					this.kafkaStreams = new KafkaStreams(topology, this.properties, this.clientSupplier);
--- a/spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
@@ -51,6 +51,7 @@ import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
  * @author Pawel Szymczyk
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Denis Washington
  */
 @SpringJUnitConfig
 @DirtiesContext
--- a/spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
@@ -17,6 +17,8 @@
 package org.springframework.kafka.config;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
 
 import java.io.IOException;
 import java.nio.file.Files;
@@ -24,7 +26,9 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Properties;
 
+import org.apache.kafka.streams.StreamsBuilder;
 import org.apache.kafka.streams.StreamsConfig;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
@@ -66,6 +70,9 @@ public class StreamsBuilderFactoryBeanTests {
 	@Autowired
 	private StreamsBuilderFactoryBean streamsBuilderFactoryBean;
 
+	@Autowired
+	private KafkaStreamsConfiguration kafkaStreamsConfiguration;
+
 	@Test
 	public void testCleanupStreams() throws IOException {
 		Files.createDirectory(Paths.get(stateStoreDir.toString(), APPLICATION_ID));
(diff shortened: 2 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -290,7 +290,15 @@ public synchronized void start() {
 			try {
 				Assert.state(this.streamsConfig != null || this.properties != null,
 						"'streamsConfig' or streams configuration properties must not be null");
-				Topology topology = getObject().build(); // NOSONAR
```
artembilan: We need to keep such a `// NOSONAR` message here. Just because Sonar thinks that `getObject()` may return `null`.

I would change your code to this:
```
Properties topologyProps = this.properties;
if (topologyProps == null) {
    topologyProps = new Properties();
	topologyProps.putAll(this.streamsConfig.originals());
}
Topology topology = getObject().build(); // NOSONAR
```

Much shorter and identical solution.

Although I doubt that `streamsConfig.originals()` returns all the configured properties in the `StreamsConfig`.
Not sure, though, that there is some other way...
Also, please, add your name to the `@author` list and would be great to have some unit-test to be sure that we are covered.

Thanks

denisw: @artembilan  Thanks for your quick review. I changed the control flow as you suggested, and will think of a unit test that covers the new behavior.

Regarding `streamsConfig.originals()`, the source code of `AbstractConfig` strongly suggests that it returns all configuration values. The simply values are simply not "processed" yet (e.g., qualified class names converted to `Class` objects), but that's the exact raw format `StreamsBuilder#build` expects.

denisw: I tried adding a unit test, but found it impossible to test the actual effect of passing down `topology.optimization` with the current data `StreamsBuilderFactoryBean` exposes. I'd need to make the created `Topology` accessible to the test, perhaps through a package-private getter.

artembilan: M-m-m. Haven't you looking into a `TopologyTestDriver` from the `kafka-streams-test-utils`: https://kafka.apache.org/11/documentation/streams/developer-guide/testing.html?

denisw: `TopologyTestDriver` requires a `Topology` object, which I don't have in the test. :) Also, `TopologyTestDriver` operates in a black box fashion and can thus not be used to find out whether topology optimization has been applied (because the point of the optimizations is that they don't change the topology's semantics).

As mentioned before, the only workable solution I could think of is to add a package-private `getTopology()` method to `StreamsBuilderFactoryBean`, and then `describe()` the topology to assert e.g. the absence of a separate changelog topic for a `KTable` source (the easier of the two optimizations to test for).

artembilan: OK. I see what you mean.
So, how about a simple unit test with mocks:
```
StreamsBuilderFactoryBean factoryBean = new StreamsBuilderFactoryBean(config) {
    
        @Override
	protected StreamsBuilder createInstance() {
               return spy(super.createInstance());
        }
  
}
```

and verify in the end that `factoryBean.getObject().build(props)` is called and, of course, we can capture that arg and verify that expected properties are propagated to this call.

Optimized topology or not is not our project responsibility. I just don't care on this level what they do with properties I provide. Our goal to have coverage for our own code. That's all.

Thanks for understanding.

denisw: Ah, good idea with the anonymous subclass! I‘ll go with that approach.

I agree with topology optimization not being this library‘s responsibility, by the way. I just couldn‘t think of any other observable effect to test on. The spy solves that nicely, though.

denisw: I now added tests for both types of property sources (`this.properties` and `this.streamsConfig`).

denisw:
> Your name to the list of `@author` for all the affected classes and, please, don't squash commits for PRs.
> We do that on merge, but with history preserving it is much easier for review.

Ok, I added the `@author` tag. I'll remember not to squash next time.

artembilan:
Travis has failed because of:
```
[ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-kafka/spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java:29:8: Unused import - java.util.Properties. [UnusedImports]
```
You can run `./gradlew clean check --parallel` locally to be sure that your changes are clean.

Thanks

denisw:
> Travis has failed because of:
> 
> ```
> [ant:checkstyle] [ERROR] /home/travis/build/spring-projects/spring-kafka/spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java:29:8: Unused import - java.util.Properties. [UnusedImports]
> ```
> 
> You can run `./gradlew clean check --parallel` locally to be sure that your changes are clean.
> 
> Thanks

Unfortunately, `./gradlew check` fails for me (`Unable to create Root Module:`), but the unused import is fixed now. Apparently all checks pass now.

artembilan:
@denisw ,

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -290,7 +291,12 @@ public synchronized void start() {
...
+				Properties topologyProps = this.properties;
+				if (topologyProps == null) {
+					topologyProps = new Properties();
+					topologyProps.putAll(this.streamsConfig.originals());
```
garyrussell: We should probably just dump this deprecated stuff in 2.3; we're now getting deprecation warnings in `start()` with the 2.3.0 client.

artembilan: Well, we need to deprecate a public API, but deal just with the `Properties` internally. Fortunately this `this.streamsConfig.originals()` may help us in setters.

garyrussell: We already deprecated it in 2.2.

artembilan: Doh. Missed that. So, yes: we are good to remove `StreamsConfig` support in `2.3`.

garyrussell: https://github.com/spring-projects/spring-kafka/issues/1159

PR #1163, opened 2019-07-11
ConsumerSeekAware: Add seek to time stamp

garyrussell:


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1462,6 +1463,16 @@ else if (ackMode.equals(AckMode.COUNT_TIME) && elapsed) {
 		}
 
 		private void processSeeks() {
+			List<TopicPartitionOffset> timestampSeeks = this.seeks.stream()
+				.filter(tpo -> SeekPosition.TIMESTAMP.equals(tpo.getPosition()))
+				.collect(Collectors.toList());
```
artembilan: I understand that this is not critical path in our logic, but such a `processSeeks()` method is called on every poll.
So, can we consider do not use Java Streams? According your logic it looks like a simple `Iterator` can do the same.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1641,6 +1652,17 @@ else if (offset >= 0) {
...
 		}
 
+		@Override
+		public void seekToTimestamp(String topic, int partition, long timestamp) {
+			this.seeks.add(new TopicPartitionOffset(topic, partition, timestamp, SeekPosition.TIMESTAMP));
+		}
+
+		@Override
+		public void seekToTimestamp(Collection<TopicPartition> topicParts, long timestamp) {
+			this.seeks.addAll(topicParts.stream()
```
artembilan: And why can't we do a simple iteration with a plain delegation to the `seekToTimestamp(String topic, int partition, long timestamp)`? I'm really not a fun of Java Stream at runtime. It's OK to use them during initialization, but if the code is called several times at runtime, it would be better do not use them in the Framework.

PR #1164, opened 2019-07-11
Add seek helper abstract class

garyrussell:
- also handle late arriving time stamp seek requests while processSeeks() is already running.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
@@ -16,6 +16,7 @@
 
 package org.springframework.kafka.listener;
 
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -48,8 +49,16 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	@Override
 	public void onPartitionsAssigned(Map<TopicPartition, Long> assignments, ConsumerSeekCallback callback) {
-		this.callbacks.clear();
-		assignments.keySet().forEach(tp -> this.callbacks.put(tp, callbackForThread.get()));
+		ConsumerSeekCallback threadCallback = callbackForThread.get();
+		if (threadCallback != null) {
+			assignments.keySet().forEach(tp -> this.callbacks.put(tp, threadCallback));
+		}
+	}
+
+	@Override
+	public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
+		partitions.forEach(tp -> this.callbacks.remove(tp));
+		callbackForThread.remove();
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


garyrussell:
Do not merge - needs more work.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
```diff
@@ -0,0 +1,82 @@
...
+ * listener subclasses this class, it can easily seek arbitrary topics/partitions without
+ * having to keep track of the callbacks itself.
+ *
+ * @author Gary Russell
+ * @since 2.3
+ *
+ */
+public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
+
+	private static final ThreadLocal<ConsumerSeekCallback> callbackForThread = new ThreadLocal<>();
```
artembilan: `static`?
Shouldn't such a thread store be an instance based?

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
```diff
@@ -0,0 +1,82 @@
...
+
+	private final Map<TopicPartition, ConsumerSeekCallback> callbacks = new ConcurrentHashMap<>();
+
+	private final Map<TopicPartition, ConsumerSeekCallback> unmodifiable = Collections.unmodifiableMap(this.callbacks);
```
artembilan: I think `unmodifiable` is not so informative name. What is the point to have it as a property?
Seems for me we never do that in other similar places...

src/reference/asciidoc/kafka.adoc
```diff
@@ -2153,6 +2164,35 @@ class Listener implements ConsumerSeekAware {
...
+public class SeekToLastOnIdleListener extends AbstractConsumerSeekAware {
+
+    @KafkaListener(id = "seekOnIdle", topics = "seekOnIdle")
+    public void listen(String in) {
+        ...
+    }
+
+    @Override
+    public void onIdleContainer(Map<org.apache.kafka.common.TopicPartition, Long> assignments,
+            ConsumerSeekCallback callback) {
```
artembilan: Is this argument out of use now?
What is the point to perform a search if we have this argument here?
Or what is the relationship between this argument and whatever is `assignments`?

Thanks for understanding.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
garyrussell: Good catch.

src/reference/asciidoc/kafka.adoc
garyrussell: Doh; of course; the collection is really only needed when performing seeks from some other arbitrary environment (like in the example above this).

Will polish.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
garyrussell: OK

Issue #1158, opened 2019-07-10
Deprecate TopicPartitionInitialOffset


garyrussell:
Rename to `TopicPartitionOffset` to reflect its more general use and create a (deprecated) subclass with the old name.

PR #1167, opened 2019-07-16
GH-1158: Fix TopicPartitionInitialOff. Deprecation

garyrussell:
Revert "GH-1158: Deprecate TopicPartitionInitialOffset"

Previous commit caused a breaking API change.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
@@ -20,7 +20,6 @@ package org.springframework.kafka.config;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
 import org.apache.commons.logging.LogFactory;
 
@@ -389,9 +388,8 @@ public abstract class AbstractKafkaListenerContainerFactory<C extends AbstractMe
 	@Override
 	public C createContainer(final Collection<org.springframework.kafka.support.TopicPartitionInitialOffset> topicPartitions) {
 		return createContainer(topicPartitions.stream()
-				.map(TopicPartitionOffset::fromTPIO)
-				.collect(Collectors.toList())
-				.toArray(new TopicPartitionOffset[0]));
+				.map(org.springframework.kafka.support.TopicPartitionInitialOffset::toTPO)
+				.toArray(TopicPartitionOffset[]::new));
 	}
 
 	@Override
@@ -400,7 +398,7 @@ public abstract class AbstractKafkaListenerContainerFactory<C extends AbstractMe
 
 					@Override
 					public TopicPartitionOffset[] getTopicPartitionsToAssign() {
-						return topicsAndPartitions;
+						return Arrays.copyOf(topicsAndPartitions, topicsAndPartitions.length);
 					}
 
 				};
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
@@ -393,7 +393,7 @@ public abstract class AbstractKafkaListenerContainerFactory<C extends AbstractMe
 	}
 
 	@Override
-	public C createContainer(final TopicPartitionOffset[] topicsAndPartitions) {
+	public C createContainer(final TopicPartitionOffset... topicsAndPartitions) {
 		KafkaListenerEndpoint endpoint = new KafkaListenerEndpointAdapter() {
 
 					@Override
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
@@ -20,6 +20,7 @@ package org.springframework.kafka.config;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import org.apache.commons.logging.LogFactory;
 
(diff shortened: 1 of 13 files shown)


spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
```diff
@@ -379,13 +380,27 @@ else if (this.autoStartup != null) {
...
+	public C createContainer(final Collection<org.springframework.kafka.support.TopicPartitionInitialOffset> topicPartitions) {
+		return createContainer(topicPartitions.stream()
+				.map(TopicPartitionOffset::fromTPIO)
+				.collect(Collectors.toList())
+				.toArray(new TopicPartitionOffset[0]));
+	}
+
 	@Override
-	public C createContainer(final Collection<TopicPartitionOffset> topicPartitions) {
+	public C createContainer(final TopicPartitionOffset[] topicsAndPartitions) {
```
artembilan: Can this be a `varargs`?

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
```diff
@@ -379,13 +380,27 @@ else if (this.autoStartup != null) {
...
 		KafkaListenerEndpoint endpoint = new KafkaListenerEndpointAdapter() {
 
 					@Override
-					public Collection<TopicPartitionOffset> getTopicPartitions() {
-						return topicPartitions;
+					public TopicPartitionOffset[] getTopicPartitionsToAssign() {
+						return topicsAndPartitions;
```
artembilan: Don't we need to return a copy from here do not allow end-user to modify an internal state of this component?

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
```diff
@@ -379,13 +380,27 @@ else if (this.autoStartup != null) {
...
+	 * {@inheritDoc}
+	 * @deprecated in favor of {@link #createContainer(TopicPartitionOffset[])}
+	 */
+	@Deprecated
+	@SuppressWarnings("deprecation")
+	@Override
+	public C createContainer(final Collection<org.springframework.kafka.support.TopicPartitionInitialOffset> topicPartitions) {
+		return createContainer(topicPartitions.stream()
+				.map(TopicPartitionOffset::fromTPIO)
+				.collect(Collectors.toList())
```
artembilan: Doesn't `toArray(TopicPartitionOffset[]::new)` work here?

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
```diff
@@ -379,13 +380,27 @@ else if (this.autoStartup != null) {
...
+	/**
+	 * {@inheritDoc}
+	 * @deprecated in favor of {@link #createContainer(TopicPartitionOffset[])}
+	 */
+	@Deprecated
+	@SuppressWarnings("deprecation")
+	@Override
+	public C createContainer(final Collection<org.springframework.kafka.support.TopicPartitionInitialOffset> topicPartitions) {
+		return createContainer(topicPartitions.stream()
+				.map(TopicPartitionOffset::fromTPIO)
```
artembilan: Maybe such a mapper should be in an opposite place - `TopicPartitionInitialOffset.toTopicPartitionOffset()`?

Even if this is going to cause some class tangle, that mapper is going to disappear together with deleted deprecated class.

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
```diff
@@ -190,6 +191,26 @@ public void setTopics(String... topics) {
...
+	 * @deprecated in favor of {@link #setTopicPartitions(TopicPartitionOffset...)}.
+	 * @see #setTopics(String...)
+	 * @see #setTopicPattern(Pattern)
+	 */
+	@Deprecated
+	@SuppressWarnings("deprecation")
+	public void setTopicPartitions(org.springframework.kafka.support.TopicPartitionInitialOffset... topicPartitions) {
+		Assert.notNull(topicPartitions, "'topics' must not be null");
+		this.topicPartitions.clear();
+		this.topicPartitions.addAll(Arrays.asList(topicPartitions).stream()
```
artembilan: `Arrays.stream(topicPartitions)`?

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
```diff
@@ -190,6 +191,26 @@ public void setTopics(String... topics) {
...
+				.map(TopicPartitionOffset::fromTPIO)
+				.collect(Collectors.toList()));
```
artembilan: `forEach(this.topicPartitions::add)` ?

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
```diff
@@ -202,10 +223,25 @@ public void setTopicPartitions(TopicPartitionOffset... topicPartitions) {
...
 	 * Return the topicPartitions for this endpoint.
 	 * @return the topicPartitions for this endpoint.
+	 * @deprecated in favor of {@link #getTopicPartitionsToAssign()}
+	 */
+	@Deprecated
+	@SuppressWarnings("deprecation")
+	@Override
+	public Collection<org.springframework.kafka.support.TopicPartitionInitialOffset> getTopicPartitions() {
+		return Collections.unmodifiableCollection(this.topicPartitions.stream()
+				.map(TopicPartitionOffset::toTPIO)
```
artembilan: Same "ball of mud" isolation concern: `TopicPartitionInitialOffset.fromTopicPartitionOffset()`.

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
garyrussell: No; `@SafeVarArgs` can't be used on a non-final method.

garyrussell: My plan is to deprecate this in 2.4 and reintroduce `Collection<TPO>`.

artembilan: Why do we talk about `@SafeVarArgs` here if our type is just an explicit and there is no any generics?
😕

garyrussell: Not sure what happened, but I got a warning yesterday; now it's not complaining - fixing.

artembilan:
Gary, here is some polishing for your confirmation: https://github.com/artembilan/spring-kafka/commit/e7374d3107c16dbac965f1545bd10a2afc131025

Pay attention to the redundant suppresses when we already have a `@Deprecated` and also to the `default` method in the `KafkaListenerEndpoint`.
Plus some goodies from Java Streams.

Everything else is like inconsistency between our IDEs. 😢

garyrussell:
LGTM; thanks.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/2a05612410788ddaa1527da2ad9e4dd0ed63486e.

PR #1172, opened 2019-07-22
Bump to the latest BUILD-SNAPSHOTs

artembilan:
After introducing a `BeanDefinition.getResolvableType()` we have
several failed builds because of `NoSuchMethodError`.
Looks like interface and its implementation classes are loaded from
different versions.

* Exclude `org.springframework` from `spring-data-commons` to avoid
CLASSPATH conflicts

Related to
https://build.spring.io/browse/SK-SON-1242/
https://github.com/spring-projects/spring-framework/issues/23178


artembilan:
If this is OK, I'll do the same for Spring AMQP. That one doesn't use SF B-S and that causes problems in SI `master`.

Time to think about global `exclude` for `org.springframework` group...

PR #1174, opened 2019-07-22
Use `@Configuration(proxyBeanMethods=false)`

artembilan:
Related to https://github.com/spring-projects/spring-integration/issues/2988

For faster startup (and possibility to compile to native code with GraalVM)
use a `proxyBeanMethods = false` on all the `@configuration` classes
in the Framework
* Provide some other code style polishing


garyrussell:
Strange travis issue.

artembilan:
Pushed more upgrades.
Let's see how it works now!

Issue #1171, opened 2019-07-19
ReplyingKafkaTemplate supports message headers only from spring-kafka


mkamalov:
Hi,

This is a short description of a problem:
I have my own service that cunsuming messages from kafka topic, make a calculation and send a result to output topic. The sevice uses output topic name from kafka header and supports correlationId for requests.
I can't configure ReplyingKafkaTemplate to support my service if I use my own header names for reply topic and correlation id. ReplyingKafkaTemplate is written to support message headers only from spring-kafka and there is no way to map spring-kafka headers to custom ones.

It means that I have to use Spring Kafka header names if I want my service is invoked via ReplyingKafkaTemplate. I can't use CORRELATION_ID instead of kafka_correlationId for message header if I want to support spring kafka integration. Do you have a plan to support custom headers mapper for Kafka Headers and support it in ReplyingKafkaTemplate?

Thanks, Marat.

**Affects Version(s):** \<Spring for Apache Kafka version>

artembilan:
I don't understand what is the problem with built-in headers.
Let's look at this service!
```
@KafkaListener(id = A_REQUEST, topics = A_REQUEST)
@SendTo  // default REPLY_TOPIC header
public String handleA(String in) {
	return in.toUpperCase();
}
```
there is just no any headers you mention. Everything is hidden from you by the Framework.
The `KafkaHeaders.REPLY_TOPIC` and `KafkaHeaders.CORRELATION_ID` headers are picked up from the request message and stored internally until you return something for sending a reply.
On the producer side everything is transparent as well: you just use `ReplyingKafkaTemplate.sendAndReceive(ProducerRecord);`.

What is the point where one would need custom headers for such a correlation mechanism?

Please, elaborate more.

mkamalov:
Hi Artem.

You are right if I use ReplyingKafkaTemplate with @SendTo, but what if my service doesn't have dependency to spring-kafka and it is written on native kafka API or it is written on python?

Thanks.

artembilan:
Well, if you really can't modify a service according `ReplyingKafkaTemplate` contract, then I don't see any problems to let you to override a header name for reply topic, reply partition and correlation key.

Unfortunately this is going to involve much more, than just `ReplyingKafkaTemplate`. We rely exactly on these headers in many other framework components.
So, if you are going to change keys on request-reply side, you need to ensure that receive-reply side is configured respectively...

I see the point, but it is much involved, than it seems at a glance.

[...]

PR #1182, opened 2019-07-31
GH-1171: Custom headers for ReplyingKafkaTemplate

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1171

Allow the use of custom header names for request/reply semantics.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: java
 jdk: oraclejdk8
 install: true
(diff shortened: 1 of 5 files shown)


garyrussell:
I thought about that, but decided why should we over-complicate there. Since we are the "server" we can dictate what we expect.

This is really for the case where the server is "foreign" and has its own requirements.

Therefore, I'd prefer to wait for a valid end-user use case before making changes there.

artembilan:
Right, I have the same opinion. Sorry not being an explicit.

Merging...

PR #1183, opened 2019-08-01
Update assert messages to supplier messages

mpoqq:
Improve the code by using supplier in the assert messages instead of static messages.

We had a problem in our project, because of the string generation in AbstractKafkaListenerEndpoint Line [474](https://github.com/spring-projects/spring-kafka/blob/2a05612410788ddaa1527da2ad9e4dd0ed63486e/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java#L474). The method endpointDescription in MethodKafkaListenerEndpoint needs the EventListener Bean, which in our case was not yet correctly initialized.

By using the lazy message generation our problem was solved and in the same context I also updated the other assert calls.


garyrussell:
Thanks!

Don't be concerned about the unrelated travis failure; I've fixed that in another PR.

Issue #1196, opened 2019-08-16
Use close with timeout in KafkaTemplate


garyrussell:
See https://stackoverflow.com/questions/57517986/how-to-use-messagechannel-in-transactional-context-with-spring-cloud-stream/57526826#57526826

PR #1197, opened 2019-08-16
Use close(Duration) instead of close()

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1196

Add `closeTimeout` to `KafkaTemplate` and `KafkaTransactionManager` (default 5s).
Use a zero timeout if a transaction operation failed with a timeout.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -174,7 +174,7 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V> {
 	/**
 	 * Set the maximum time to wait when closing a producer; default 5 seconds.
 	 * @param closeTimeout the close timeout.
-	 * @since 1.3.11
+	 * @since 2.3
 	 */
 	public void setCloseTimeout(Duration closeTimeout) {
 		Assert.notNull(closeTimeout, "'closeTimeout' cannot be null");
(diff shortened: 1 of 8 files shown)


garyrussell:
I will do manual backports after review/merge because there will be conflicts.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -169,6 +171,16 @@ public void setTransactionIdPrefix(String transactionIdPrefix) {
 		this.transactionIdPrefix = transactionIdPrefix;
 	}
 
+	/**
+	 * Set the maximum time to wait when closing a producer; default 5 seconds.
+	 * @param closeTimeout the close timeout.
+	 * @since 1.3.11
```
artembilan: The `Duration` is a Java 8 feature. You may consider to provide a separate PR for Java 6 based branch

garyrussell: Right; they have `close(long, TimeUnit)` there - I should fix these `@since`.

garyrussell: Actually, even our 2.2.x only has that; so this is `@since 2.3`; I need think some more about how to support forward kafka-clients in those old versions.

artembilan: So, merging without back-porting?

garyrussell:
So let's treat this as master only and I will work on the backport.

Issue #1186, opened 2019-08-01
Add pollDelay container property


garyrussell:
Several users have requested the ability to add a fixed delay between polls, rather than polling immediately after the records from the previous poll have been processed.

It will be important to ensure that `pollDelay + timeToProcessPolledRecords` is less than `max.poll.interval.ms` to avoid a rebalance.

artembilan:
We have similar option in Kinesis Inbound Channel Adapter:
```
/**
 * The sleep interval in milliseconds used in the main loop between shards polling cycles.
 * Defaults to {@code 1000}l minimum {@code 250}.
 * @param idleBetweenPolls the interval to sleep between shards polling cycles.
 */
public void setIdleBetweenPolls(int idleBetweenPolls) {
```
But that one is a standard AWS requirement to avoid throttling and there is no rebalance...

Should we really care about that `pollDelay + timeToProcessPolledRecords` problem here and ensure `Max.min()` before spleeping?

garyrussell:
Good point - it still won't avoid a rebalance if `Math.min()` returns <0 but that's ok.

garyrussell:
We should use the same property name for consistency.

PR #1198, opened 2019-08-16
GH-1186: Add `idleBetweenPolls` container option

artembilan:
Fixes https://github.com/spring-projects/spring-kafka/issues/1186

A new `ContainerProperties.idleBetweenPolls` option allows to configure
a sleep interval between `poll` calls in the main container loop.
The actual sleep interval is selected as a minimum from provided
option and difference between `max.poll.interval.ms` and records
processing time

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -648,7 +648,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			boolean isAutoCommit;
 			String autoCommitOverride = consumerProperties.getProperty(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG);
 			if (!KafkaMessageListenerContainer.this.consumerFactory.getConfigurationProperties()
-					.containsKey(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG)
+							.containsKey(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG)
 					&& autoCommitOverride == null) {
 				consumerProperties.setProperty(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "false");
 				isAutoCommit = false;
@@ -689,9 +689,9 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 					if (timeout != null) {
 						Object timeoutToLog = timeout;
 						this.logger.warn(() -> "Unexpected type: " + timeoutToLog.getClass().getName()
-								+ " in property '"
-								+ ConsumerConfig.DEFAULT_API_TIMEOUT_MS_CONFIG
-								+ "'; defaulting to 60 seconds for sync commit timeouts");
+							+ " in property '"
+							+ ConsumerConfig.DEFAULT_API_TIMEOUT_MS_CONFIG
+							+ "'; defaulting to 60 seconds for sync commit timeouts");
 					}
 					return Duration.ofSeconds(SIXTY);
 				}
@@ -782,11 +782,10 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			}
 			Class<?> clazz = errHandler.getClass();
 			Assert.state(batch
-							? BatchErrorHandler.class.isAssignableFrom(clazz)
-							: ErrorHandler.class.isAssignableFrom(clazz),
+						? BatchErrorHandler.class.isAssignableFrom(clazz)
+						: ErrorHandler.class.isAssignableFrom(clazz),
 					() -> "Error handler is not compatible with the message listener, expecting an instance of "
-							+ (batch ? "BatchErrorHandler" : "ErrorHandler") + " not " + errHandler.getClass()
-							.getName());
+					+ (batch ? "BatchErrorHandler" : "ErrorHandler") + " not " + errHandler.getClass().getName());
 		}
 
 		@Override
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -883,6 +916,23 @@ private void checkIdle() {
 			}
 		}
 
+		private void idleBetweenPollIfNecessary() {
+			long idleBetweenPolls = this.containerProperties.getIdleBetweenPolls();
+			if (idleBetweenPolls > 0) {
+				idleBetweenPolls = Math.min(idleBetweenPolls,
+						this.maxPollInterval - (System.currentTimeMillis() - this.lastPoll));
```
garyrussell: `this.lastPoll` is when the last poll exited.

I think there might be a problem with a large `pollTimeout` because I think `max.poll.interval` is the time between polls, not the time between when the last poll exited and the next poll.

Since `lastPoll` is only used (otherwise) for idle consumer detection, I think it is safe to move it to before the poll.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -883,6 +916,23 @@ private void checkIdle() {
...
+				if (idleBetweenPolls > 0) {
+					try {
+						TimeUnit.MILLISECONDS.sleep(idleBetweenPolls);
+					}
+					catch (InterruptedException ex) {
+						Thread.currentThread().interrupt();
+						throw new IllegalStateException("Consumer Thread [" + this + "] has been interrupted", ex);
```
garyrussell: I don't think this is necessary and it will cause `wrapUp()` to be skipped because the call to this method is outside of the `handleConsumerException()` catch block.

Most likely an interrupt is because we're shutting down, but perhaps the safest thing to do is move the call to the try block so that `handleConsumerException()` will log this exception.

src/reference/asciidoc/kafka.adoc
```diff
@@ -934,6 +934,9 @@ Starting with version 1.3, the `MessageListenerContainer` provides access to the
 In the case of `ConcurrentMessageListenerContainer`, the `metrics()` method returns the metrics for all the target `KafkaMessageListenerContainer` instances.
 The metrics are grouped into the `Map<MetricName, ? extends Metric>` by the `client-id` provided for the underlying `KafkaConsumer`.
 
+Starting with version 2.3, the `ContainerProperties` provides an `idleBetweenPolls` option to let the main loop in the listener container to sleep between `KafkaConsumer.poll()` calls.
+An actual sleep interval is selected as a minimum from the provided option and difference between a `max.poll.interval.ms` consumer config and the current records batch processing time.
```
garyrussell: as **the** minimum... difference between **the** max.poll... and the ...previous record batch...

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: Resolved in the latest commit. See its description.

Thank you for feedback!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -883,6 +917,17 @@ private void checkIdle() {
 			}
 		}
 
+		private void idleBetweenPollIfNecessary() throws InterruptedException {
```
garyrussell: We still need to catch the interrupt and reset the interrupted flag because `handleConsumerException()` simply logs and eats the exception and it won't set the interrupted flag.

I meant that the choice was to just set the interrupted flag and exit (so something else deals with the interrupt), or throw the illegal state exception here and move the call to the try block.

artembilan: Pushed. Looks like I still need to learn how to handle `InterruptedException` properly 😄

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: After further though, I think we need some "headroom" here - otherwise if we wait for all of the remaining time in `maxPollInterval`, we'll be in a race with the broker regarding a rebalance.

Maybe we should consider the minimum to no more than several seconds (10?) less than the remaining time before a rebalance will occur.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -515,6 +518,7 @@ public void testRecordAck() throws Exception {
 			finally {
 				for (Entry<TopicPartition, OffsetAndMetadata> entry : map.entrySet()) {
 					if (entry.getValue().offset() == 2) {
+						recordTime.add(System.currentTimeMillis());
```
garyrussell: I think you need to stub `poll()` instead (and `callRealMethod()`); there is no guarantee that there will be 2 calls here - all 4 records might be returned by the first `poll()`.

artembilan: Well, that's why I added that `props.put(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, 1);` config above.

garyrussell: Doh; sorry.

xzhan:
@artembilan  @garyrussell  would you mind check this issue https://github.com/spring-projects/spring-kafka/issues/1601

Issue #1203, opened 2019-08-19
Extract ConsumerProperties super class from the ContainerProperties for common options for MessageSource


artembilan:
See https://github.com/spring-projects/spring-integration-kafka/issues/280 and its PR for more info.

PR #1206, opened 2019-08-19
GH-1203: Extract ContainerProperties Super Class

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1203

Common properties that can be used with any consumer, not just a consumer that's
in a container.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
@@ -125,22 +125,6 @@ public class ConsumerProperties {
 	 * Create properties for a container that will assign itself the provided topic
 	 * partitions.
 	 * @param topicPartitions the topic partitions.
-	 * @deprecated in favor of {@link #ContainerProperties(TopicPartitionOffset...)}.
-	 */
-	@Deprecated
-	ConsumerProperties(org.springframework.kafka.support.TopicPartitionInitialOffset... topicPartitions) {
-		this.topics = null;
-		this.topicPattern = null;
-		Assert.notEmpty(topicPartitions, "An array of topicPartitions must be provided");
-		this.topicPartitionsToAssign = Arrays.stream(topicPartitions)
-				.map(org.springframework.kafka.support.TopicPartitionInitialOffset::toTPO)
-				.toArray(TopicPartitionOffset[]::new);
-	}
-
-	/**
-	 * Create properties for a container that will assign itself the provided topic
-	 * partitions.
-	 * @param topicPartitions the topic partitions.
 	 */
 	public ConsumerProperties(TopicPartitionOffset... topicPartitions) {
 		this.topics = null;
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
```diff
@@ -0,0 +1,325 @@
...
+	public ConsumerProperties(TopicPartitionOffset... topicPartitions) {
+		this.topics = null;
+		this.topicPattern = null;
+		Assert.notEmpty(topicPartitions, "An array of topicPartitions must be provided");
+		this.topicPartitionsToAssign = Arrays.copyOf(topicPartitions, topicPartitions.length);
+	}
+
+	@Nullable
+	public String[] getTopics() {
+		return this.topics; // NOSONAR
```
artembilan: Maybe `Arrays.copyOf()` as well? For consistency: see `getTopicPartitionsToAssign()`

spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
```diff
@@ -0,0 +1,325 @@
...
+		this.topicPartitionsToAssign = null;
+	}
+
+	/**
+	 * Create properties for a container that will assign itself the provided topic
+	 * partitions.
+	 * @param topicPartitions the topic partitions.
+	 * @deprecated in favor of {@link #ContainerProperties(TopicPartitionOffset...)}.
+	 */
+	@Deprecated
```
artembilan: This class is new and I believe it really can be set free from any deprecated code.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
```diff
@@ -0,0 +1,325 @@
...
+	 * <ul>
+	 * <li>this property</li>
+	 * <li>{@code ConsumerConfig.DEFAULT_API_TIMEOUT_MS_CONFIG} in
+	 * {@link #setConsumerProperties(Properties)}</li>
+	 * <li>{@code ConsumerConfig.DEFAULT_API_TIMEOUT_MS_CONFIG} in the consumer factory
+	 * properties</li>
+	 * <li>60 seconds</li>
+	 * </ul>
+	 * @param syncCommitTimeout the timeout.
+	 * @since 2.3
```
artembilan: No reason in this `@since`, since the whole class is since `2.3`...

spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
garyrussell: Right; that's why I made it package protected - I can't do the assertion in the subclass before calling `super(...)`.

garyrussell: But there's another way...

PR #1209, opened 2019-08-21
Try/Catch around error recovery [Backport]

garyrussell:
- if a recoverer `BiConsumer` threw an exception, the seeks would not be performed
- also add try/catch around after rollback processor calls

**cherry pick to 2.1.x, 1.7.x**


artembilan:
Looks like `FailedRecordTracker` with its `BiConsumer<ConsumerRecord<?, ?>, Exception> recoverer` has been introduced since `2.2`.
So, seems for me there is nothing to back-port to `2.1.x`.
Also `1.7.x`  is Java 6 based: we can't talk about `BiConsumer` there yet...

Please, elaborate more about the back-port task.

Thanks

garyrussell:
My mistake - I forgot recovery was only added in 2.2.x; ignore 😄

PR #1213, opened 2019-08-23
TopicBuilder Improvements

garyrussell:
- add `assignReplicas()`


artembilan:
Any thoughts about failed build on Travis?
I wonder if we have everything correctly in Gradle config in regards to JUnit 5...

Otherwise LGTM

PR #1240, opened 2019-09-17
Pull Up Consumer Properties

garyrussell:
- move `consumerProperties` from `ContainerProperties` to `ConsumerProperties`
- rename to `kafkaConsumerProperties`
- deprecate old accessors

- allows overriding factory consumer properties in SIK message source

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
@@ -18,6 +18,7 @@ package org.springframework.kafka.listener;
 
 import java.time.Duration;
 import java.util.Arrays;
+import java.util.Properties;
 import java.util.regex.Pattern;
 
 import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
@@ -93,6 +94,8 @@ public class ConsumerProperties {
 	private boolean syncCommits = true;
 
 	private LogIfLevelEnabled.Level commitLogLevel = LogIfLevelEnabled.Level.DEBUG;
+
+	private Properties kafkaConsumerProperties = new Properties();
 
 	/**
 	 * Create properties for a container that will subscribe to the specified topics.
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
```diff
@@ -270,6 +273,35 @@ public void setCommitLogLevel(LogIfLevelEnabled.Level commitLogLevel) {
...
+	 * provided by the consumer factory; properties here will supersede any with the same
+	 * name(s) in the consumer factory.
+	 * {@code group.id} and {@code client.id} are ignored.
+	 * @param kafkaConsumerProperties the properties.
+	 * @see org.apache.kafka.clients.consumer.ConsumerConfig
+	 * @see #setGroupId(String)
+	 * @see #setClientId(String)
+	 */
+	public void setKafkaConsumerProperties(Properties kafkaConsumerProperties) {
+		Assert.notNull(kafkaConsumerProperties, "'consumerProperties' cannot be null");
```
artembilan: Copy/paste artefact. Must be ` "'kafkaConsumerProperties' cannot be null`

spring-kafka/src/main/java/org/springframework/kafka/core/LiKafkaProducerFactory.java
```diff
@@ -0,0 +1,34 @@
...
+package org.springframework.kafka.core;
+
+import org.springframework.context.ApplicationEvent;
+
+/**
+ * @author Gary Russell
+ * @since 5.2
+ *
+ */
+public class LiKafkaProducerFactory extends DefaultKafkaProducerFactory {
```
artembilan: Gary, I think this is side effect of your other work.
Please, take it away from the PR.

garyrussell: Oops; fixed.

PR #1241, opened 2019-09-18
Upgrade dependencies

artembilan:
Use a `dependencyManagement` for those projects which bring BOMs

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,10 +14,10 @@ plugins {
 	id 'base'
 	id 'project-report'
 	id 'idea'
-	id "org.sonarqube" version '2.7.1'
-	id "org.asciidoctor.convert" version '2.3.0'
+	id 'org.sonarqube' version '2.7.1'
+	id 'org.asciidoctor.convert' version '2.3.0'
 	id 'org.ajoberstar.grgit' version '3.1.1'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -15,10 +15,10 @@ plugins {
 	id 'project-report'
 	id 'idea'
 	id "org.sonarqube" version '2.7.1'
-	id 'org.asciidoctor.convert' version '1.6.1'
+	id "org.asciidoctor.convert" version '2.3.0'
```
garyrussell: why single to double here but double to single below?

artembilan: Copy/paste artefact: https://plugins.gradle.org/plugin/org.asciidoctor.convert

Although, I agree that I've changed the version template, but missed the name.

Shame on me.

Any other observation in this change?
Any objections at all?

I've just noticed that with BOM for JUnit we don't need to manage version for platform, as well as for Reactor Kafka.

If this is OK, I'll go ahead and fix similar way AMQP and SI
I don't see reason (yet) in managing all the dependency versiona in this section to avoid templating in the target sub-projects as it is done in SF: https://github.com/spring-projects/spring-framework/blob/master/build.gradle#L41

garyrussell: No objections from me.

PR #1246, opened 2019-09-24
Upgrade to Jackson 2.10

garyrussell:
- one deprecated method replaced by `JsonBuilder.configure(...)`.

GA is expected before our GA.


artembilan:
Let's hold of merging until really their release!

PR #1248, opened 2019-09-26
Fix new Sonar smells

artembilan:


artembilan:
Here is a report, though: https://sonar.spring.io/project/issues?id=org.springframework.kafka%3Aspring-kafka-dist%3Asonar&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL

garyrussell:
I haven't found a way to disable that rule on Sonar.

Issue #1247, opened 2019-09-26
WakeupException after calling KafkaListenerEndpointRegistry#stop


PawelAdamski:
My situation is following, I have a Kafka listener which performs long operation. While it is processing a message I'm calling KafkaListenerEndpointRegistry#stop. It correctly waits until the processing is finished and sends acknowledgment but after that WakeupException is thrown.

```
org.apache.kafka.common.errors.WakeupException: null
	at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.maybeTriggerWakeup(ConsumerNetworkClient.java:498)
	at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:284)
	at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:242)
	at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:218)
	at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.commitOffsetsSync(ConsumerCoordinator.java:692)
	at org.apache.kafka.clients.consumer.KafkaConsumer.commitSync(KafkaConsumer.java:1416)
	at org.apache.kafka.clients.consumer.KafkaConsumer.commitSync(KafkaConsumer.java:1377)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.ackCurrent(KafkaMessageListenerContainer.java:1356)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.invokeOnMessage(KafkaMessageListenerContainer.java:1257)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.doInvokeRecordListener(KafkaMessageListenerContainer.java:1217)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.doInvokeWithRecords(KafkaMessageListenerContainer.java:1198)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.invokeRecordListener(KafkaMessageListenerContainer.java:1118)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.invokeListener(KafkaMessageListenerContainer.java:933)
[...]

PR #1249, opened 2019-09-26
GH-1247: Only wake consumer while polling

garyrussell:
Fixes https://github.com/spring-projects/spring-kafka/issues/1247

Previously, stopping the container unconditionally called `consumer.wakeUp()`.
It should only be woken while actually polling, otherwise operations such
as `commitSync()` would fail (after committing the offset but before calling
any callbacks or interceptors).

Add logic to only wake the consumer while it is in the `poll()` method.

**cherry-pick to 2.2.x, 2.1.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -33,6 +33,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -887,7 +890,18 @@ protected void pollAndInvoke() {
 			}
 			checkPaused();
 			this.lastPoll = System.currentTimeMillis();
+			this.polling.set(true);
 			ConsumerRecords<K, V> records = this.consumer.poll(this.pollTimeout);
+			if (!this.polling.compareAndSet(true, false)) {
+				/*
+				 * There is a small race condition where wakeIfNecessary was called between
```
artembilan: Can't we achieve the same with a `Semaphore`?

Otherwise LGTM

garyrussell: A semaphore is a bit more expensive than an `AtomicBoolean`.

I just tested it and got 18 vs 13 seconds for a billion iterations.

artembilan:
Back-ported into `2.2.x` as https://github.com/spring-projects/spring-kafka/commit/7cfc482abd614320e03dff0d2ed4b196ac781a95 after fixing conflicts.
Into `2.1.x` as https://github.com/spring-projects/spring-kafka/commit/309d28e29960da6c9f97576f6a09b6b8f1a37e27 after more conflicts fixing.

PR #1250, opened 2019-09-26
DefaultKafkaHeaderMapper Improvement

garyrussell:
String-valued headers are now mapped as raw strings by default instead
of being encoded as JSON strings.

This can be disabled until all applications have been upgraded to use
2.3 or later.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
@@ -20,7 +20,6 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -79,6 +80,8 @@
 
 	private final Set<String> toStringClasses = new LinkedHashSet<>();
 
+	private boolean encodeStrings;
```
artembilan: I don't see a usage of new property...

spring-kafka/src/test/java/org/springframework/kafka/support/DefaultKafkaHeaderMapperTests.java
```diff
@@ -48,6 +48,7 @@
  * @since 1.3
  *
  */
+@SuppressWarnings("deprecation")
```
artembilan: What deprecated API do we use in this test class?
Can we get rid of it?

garyrussell: Oops - originally I was going to deprecate the raw strings properties, but they are also used by the SimpleKafkaHeaderMapper.

Issue #1252, opened 2019-09-27
no errors are reported when it is not possible to honor a @KafkaListener


ccronemberger:
In my project I have the following ThreadPoolTaskExecutor:

    @Bean(name = "threadPoolExecutor")
    public ThreadPoolTaskExecutor getAsyncExecutor() {
        ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
        executor.setCorePoolSize(2);
        executor.setMaxPoolSize(10);
        executor.setQueueCapacity(10);
        executor.setThreadNamePrefix("consumer-");
        executor.initialize();
        return executor;
    }

And in the application I have 3 methods annotated with @KafkaListener. When the application starts I can see on the Kafka server that only 2 listeners managed to start.
In this case I would expect an error to be reported.
By increasing the core pool size to 3 I managed to fix the issue. This is also strange because the max pool size is 10, so I think this is another problem.

garyrussell:
The framework has no way to detect an improperly configured executor - each container is independent.

How do you expect it to work?

ccronemberger:
It could have a timeout for example. Or at least log a warning message if it takes too long to start the listener.

    Em sexta-feira, 27 de setembro de 2019 11:45:09 BRT, Gary Russell <notifications@github.com> escreveu:  
 
 
The framework has no way to detect an improperly configured executor - each container is independent.

How do you expect it to work?

PR #1253, opened 2019-09-27
GH-1252: Add more consumer lifecycle events

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1252

- also log an error when a consumer fails to start


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -320,9 +326,20 @@ protected void doStart() {
...
+		this.startLatch = new CountDownLatch(1);
 		this.listenerConsumerFuture = containerProperties
 				.getConsumerTaskExecutor()
 				.submitListenable(this.listenerConsumer);
+		try {
+			if (!this.startLatch.await(containerProperties.getConsumerStartTimout().toMillis(), TimeUnit.MILLISECONDS)) {
+				this.logger.error("Consumer thread failed to start - does the configured task executor "
+						+ "have enough threads to support all containers and concurrency?");
+			}
+			publishConsumerFailedToStart();
```
artembilan: Why this call is not inside `if()`?
Doesn't looks like we really have to emit this event unconditionally...

garyrussell: Doh

Issue #1225, opened 2019-09-04
KafkaListener does not try to resolve type for ConsumerRecord


ST-DDT:
🐞 Bug report / 🎁 Enhancement

According to your [docs](https://docs.spring.io/spring-kafka/reference/html/#kafka-listener-annotation)

I can create a `@KafkaListener` with a `ConsumerRecord<K, V>` like this:

````java
@KafkaListener(id = "thing2")
public void listen(ConsumerRecord<String, MyType> record) {
    ...
}
````

However in that case, the [`MessagingMessageListenerAdapter#determineInferredType(Method)`](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java#L477) method does not search for/resolve the `MyType` type and thus the `MessagingMessageConverter` is invoked with the fallback type `Object`.

(AFAICT this is defined in [`MessagingMessageListenerAdapter#eligibleParameter(MethodParameter)`](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java#L594))

**Callstack**

* org.springframework.kafka.support.converter.MessagingMessageConverter.extractAndConvertValue(ConsumerRecord<?, ?>, Type)
* org.springframework.kafka.support.converter.MessagingMessageConverter.toMessage(ConsumerRecord<?, ?>, Acknowledgment, Consumer<?, ?>, Type)
* org.springframework.kafka.listener.adapter.MessagingMessageListenerAdapter.toMessagingMessage(ConsumerRecord<K, V>, Acknowledgment, Consumer<?, ?>)

**TLDR**

The given `type` is just `Object` in [`RecordMessageConverter#toMessage(ConsumerRecord, Acknowledgment, Consumer, Type)`](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/support/converter/RecordMessageConverter.java#L44) when the `@KafkaListener` uses a `ConsumerRecord<Key, Value>` parameter but it should be of type `Value`.

garyrussell:
There is no conversion performed on ConsumerRecord, only POJO or Message<MyType> parameters. For a typed consumer record value you need an appropriate value deserializer so that Kake creates the right type.

ST-DDT:
[...]

PR #1256, opened 2019-09-30
GH-1225: @KL - skip conversion when not needed

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1225

When no argument is sourced from a `Message<?>` (no `@Headers`, `@Payload`,
etc - only `ConsumerRecord`, `Acknowledgment` or `Consumer`) then don't
convert.


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordMessagingMessageListenerAdapter.java
```diff
@@ -71,8 +72,17 @@ public RecordMessagingMessageListenerAdapter(Object bean, Method method, KafkaLi
...
+		Message<?> message;
+		if (isConversionNeeded()) {
+			message = toMessagingMessage(record, acknowledgment, consumer);
+		}
+		else {
+			message = NULL_MESSAGE;
+		}
+		if (logger.isDebugEnabled()) {
+			final Message<?> toLog = message;
+			logger.debug(() -> "Processing [" + toLog + "]");
```
artembilan: Or we don't need `isDebugEnabled()` or we don't need `toLog` variable and shouldn't use `Supplier`-based `debug()`

Issue #1259, opened 2019-10-03
Exception in recoverer with DefaultAfterRollbackProcessor still commit the offset


Rajh:
Hi,
I'm using version 2.2.9.RELEASE.

Issue:
When an exception occur in a custom recoverer for DefaultAfterRollbackProcessor, the transaction seems to not be rolled back and thus the offset is commit.

Reproduction:
Use transactional kafka listener which throws an exception.
Use DefaultAfterRollbackProcessor with a customer recoverer as follow :
`((consumerRecord, e) -> {
            throw new RuntimeException("Error in recover");
        });`

AfterRollbackProcessor:
```
@Bean
    public AfterRollbackProcessor<?, ?> afterRollbackProcessor(KafkaTemplate<Object, Object> template, BiConsumer<ConsumerRecord<?, ?>, Exception> recoverer) {
        // Send to DLQ on multiple errors
        DefaultAfterRollbackProcessor defaultRollbackProcessor = new DefaultAfterRollbackProcessor(recoverer, 1);
        // Commit offset when error has been rollback and recovered to a DLQ
        defaultRollbackProcessor.setProcessInTransaction(true);
        defaultRollbackProcessor.setKafkaTemplate(template);
        return defaultRollbackProcessor;
    }
```
The record is not reprocessed even with a restart (without any new messages after the error)

Logs:
2019-10-03 16:39:42.683 ERROR 20993 --- [ntainer#3-0-C-1] essageListenerContainer$ListenerConsumer : Transaction rolled back
2019-10-03 16:39:42.686 ERROR 20993 --- [ntainer#3-0-C-1] o.s.k.l.DefaultAfterRollbackProcessor    : Recoverer threw exception

java.lang.RuntimeException: Error in recover

PR #1261, opened 2019-10-03
GH-1259: Handle Failed Record Recovery

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1259

Previously if the recoverer in a `SeekToCurrentErrorHandler` or
`DefaultAfterRollbackProcessor` failed to recover a record, the
record could be lost; the `FailedRecordTracker` simply logged
the exception.

Change the `SeekUtils` to detect a failure in the recoverer (actually
any failure when determining if the failed record should be recovered)
and include the failed record in the seeks.

In this way the recovery will be attempted once more on each delivery
attempt.

**cherry-pick to 2.2.x**


artembilan:
... and cherry-picked into `2.2.x` as https://github.com/spring-projects/spring-kafka/commit/b05b04cef8c3498d8dd12bf40d652634f4aff49f after fixing conflicts.

Issue #1260, opened 2019-10-03
Add BinaryExceptionClassifier to DefaultAfterRollbackProcessor


Rajh:
Hi,

BinaryExceptionClassifier has been added to SeekToCurrentErrorHandler but I think the same behavior could be required in DefaultAfterRollbackProcessor ?

PR #1263, opened 2019-10-04
GH-1260: Add exception classification to DARP

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1260

Refactor common code with `SeekToCurrentErrorHandler` into
`FailedRecordProcessor` super class.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
@@ -20,13 +20,11 @@ import java.util.Collections;
 import java.util.List;
 import java.util.function.BiConsumer;
 
-import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
 
-import org.springframework.core.log.LogAccessor;
 import org.springframework.kafka.core.KafkaTemplate;
 import org.springframework.kafka.support.SeekUtils;
 import org.springframework.lang.Nullable;
@@ -49,14 +47,7 @@ import org.springframework.util.backoff.FixedBackOff;
  * @since 1.3.5
  *
  */
-public class DefaultAfterRollbackProcessor<K, V> implements AfterRollbackProcessor<K, V> {
-
-	private static final LogAccessor LOGGER =
-			new LogAccessor(LogFactory.getLog(DefaultAfterRollbackProcessor.class));
-
-	private final FailedRecordTracker failureTracker;
-
-	private boolean commitRecovered;
+public class DefaultAfterRollbackProcessor<K, V> extends FailedRecordProcessor implements AfterRollbackProcessor<K, V> {
 
 	private KafkaTemplate<K, V> kafkaTemplate;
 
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
```diff
@@ -0,0 +1,222 @@
...
+		this.failureTracker = new FailedRecordTracker(recoverer, backOff, LOGGER);
+		this.classifier = configureDefaultClassifier();
+	}
+
+	/**
+	 * TODO: remove when the deprecated dependent CTORs are removed.
+	 * @param recoverer the recoverer.
+	 * @param maxFailures the max failures.
+	 */
+	FailedRecordProcessor(@Nullable BiConsumer<ConsumerRecord<?, ?>, Exception> recoverer, int maxFailures) {
```
artembilan: Why do we need to this ctor here if we can just do `FixedBackOff` wrapping in that deprecated ctor in the `DefaultAfterRollbackProcessor`?

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
```diff
@@ -0,0 +1,222 @@
...
+	 * All others will be retried.
+	 * The classifier's {@link BinaryExceptionClassifier#setTraverseCauses(boolean) traverseCauses}
+	 * will be set to true because the container always wraps exceptions in a
+	 * {@link ListenerExecutionFailedException}.
+	 * This replaces the default classifier.
+	 * @param classifier the classifier.
+	 */
+	public void setClassifier(BinaryExceptionClassifier classifier) {
+		Assert.notNull(classifier, "'classifier' + cannot be null");
+		classifier.setTraverseCauses(true);
```
artembilan: Can't we request `Map<Class<? extends Throwable>, Boolean>` instead here and create a `BinaryExceptionClassifier` internaly to avoid such a mutation of the provided one?

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
garyrussell: Because we'd have to duplicate the code in both subclasses.

artembilan: OK. We can live with that. Looks like the matter of personal preferences. 😄

spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
```diff
@@ -126,7 +117,8 @@ public DefaultAfterRollbackProcessor(BiConsumer<ConsumerRecord<?, ?>, Exception>
 	public DefaultAfterRollbackProcessor(@Nullable BiConsumer<ConsumerRecord<?, ?>, Exception> recoverer,
 			int maxFailures) {
 
-		this.failureTracker = new FailedRecordTracker(recoverer, new FixedBackOff(0L, maxFailures - 1), LOGGER);
+		// Remove super CTOR when this is removed.
```
artembilan: Should we add `TODO` here do not forget to address it when time comes for fixing ?

garyrussell: We could but we already have it flagged as deprecated, which is trigger enough, I think.

PR #1269, opened 2019-10-15
Fix typo around NRCEvent properties

garyrussell:
- was `pollInterval` instead of `pollTimeout`

**cherry-pick to 2.2.x**


artembilan:

Issue #1273, opened 2019-10-18
Auto ack a recovered record


garyrussell:
If an error handler "handles" an error (does not throw an exception), the offset for the record should be committed (if container acks are in use),

PR #1275, opened 2019-10-18
GH-1273: Fix offset commit after recovery

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1273

Previously, offsets were only committed if an error handler "handled" an
error (did not throw an exception), when transactions are being used.

Now, if an error handler successfully recovers, the offset will automatically
be committed.

Since this is a behavior change, a new boolean `ackAfterHandle()` is added to
the `GenericErrorHandler`, default `false` and the `SeekToCurrentErrorHandler`
returns `true`; it is the only framwork handler that can recover from failures.

In a future release, this boolean will be `true` by default.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/GenericErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/GenericErrorHandler.java
@@ -56,4 +56,15 @@ public interface GenericErrorHandler<T> {
 		// NOSONAR
 	}
 
+	/**
+	 * Return true if the offset should be committed for a handled error (no exception
+	 * thrown).
+	 * @return true to commit.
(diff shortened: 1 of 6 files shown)


artembilan:
Is your change related to the failure on Travis:
```
org.springframework.kafka.listener.KafkaMessageListenerContainerTests > testRecordAckAfterRecoveryMock() FAILED

    org.opentest4j.AssertionFailedError at KafkaMessageListenerContainerTests.java:673
```
?

Thanks

garyrussell:
Yes; I decided to make it configurable at the last minute and failed to rerun tests. Thank you travis!

garyrussell:
Pushed docs; not sure if you got a notification because I rebased as well.

Issue #1278, opened 2019-10-18
Batch Listener - Support Zombie Fencing


garyrussell:
Currently, zombie fencing is not supported for a batch listener because a batch can contain records from multiple partitions and zombie fencing requires a `transactional.id` based on the group/topic/partition.

Add an option to support zombie fencing by processing `ConsumerRecords` one partition at a time - i.e. pass a list of `ConsumerRecord` to the listener for each partition received, instead of a single list containing all records received.

A transaction will then span each partition and the proper `transactional.id` can be used for each sub batch.

`FilteringBatchMessageListenerAdapter` will need to create a new list if any records are filtered because we get an unmodifiable list.

PR #1280, opened 2019-10-21
GH-1278: Zombie Fencing with Batch Listener

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1278

Support transaction per partition with Batch Listeners so the `transactional.id`
is tied to the group/topic/partition.


spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -590,6 +592,23 @@ public void setConsumerStartTimout(Duration consumerStartTimout) {
...
 	}
 
+	public boolean isSubBatchPerPartition() {
+		return this.subBatchPerPartition;
+	}
+
+	/**
+	 * When using a batch message listener whether to dispatch records by partition (with
+	 * a transaction for each sub batch if transactions are in use) or the complete batch
+	 * received by the {@code poll(}. Useful when using transactions to enable zombie
```
artembilan: Typo: no closing brace for method.

PR #1281, opened 2019-10-24
fix for #1270

### LINKED ISSUE #1270: DefaultKafkaProducerFactory logging out of place ?
In 2.3.0 , `DefaultKafkaProducerFactory` logs this:

```
...
		String txId = (String) this.configs.get(ProducerConfig.TRANSACTIONAL_ID_CONFIG);
		if (StringUtils.hasText(txId)) {
			setTransactionIdPrefix(txId);
			LOGGER.info(() -> "If 'setTransactionIdPrefix()' is not going to be configured, "
					+ "the existing 'transactional.id' config with value: '" + txId
					+ "' will be suffixed for concurrent transactions support.");
		}
...
```

this seems a bit out of place, especially since i'm using `setTransactionIdPrefix()` (but at the time of constructor there is no way of knowing) and i am seeing this at INFO level.  Also the wording is strange.

jorgheymans:
fixes #1270


jorgheymans:
Note locally i get an exception during `gradlew build` about checkstyle, not sure if it's related to my changes ?

```
Caused by: java.io.FileNotFoundException: /home/heymjo/.gradle/daemon/5.6.2/src/checkstyle/checkstyle-suppressions.xml (No such file or directory)
        at java.xml/com.sun.org.apache.xerces.internal.impl.XMLEntityManager.setupCurrentEntity(XMLEntityManager.java:654)
        at java.xml/com.sun.org.apache.xerces.internal.impl.XMLVersionDetector.determineDocVersion(XMLVersionDetector.java:150)
        at java.xml/com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:860)
        at java.xml/com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:824)
```

jorgheymans:
@artembilan tests are failing now because in some test scenario's `DefaultKafkaProducerFactory` is not used as a Spring bean. Is this a valid usage scenario ? If yes then i should undo moving the logic to `afterPropertiesSet`

artembilan:
Well, It is OK to call that `afterPropertiesSet()` from tests, but we might need to consider to revise your logic and do only an `info` in the  `afterPropertiesSet()`. I would leave an initialization from ctor as is, but would have an extra `transactionIdPrefixExplicitlySet` boolean flag populated from that `setTransactionIdPrefix()`. This one would be an indication for us to `info` or not in the   `afterPropertiesSet()`. And the last one won't be required from the tests.

artembilan:
Not sure what is going on with Checkstyle on your side, but it works well for us on Mac, Windows locally and on Linux in Travis and Bamboo.

jorgheymans:
checkstyle error was due to missing newline, that is fixed now.

> I would leave an initialization from ctor as is, but would have an extra transactionIdPrefixExplicitlySet boolean flag populated from that setTransactionIdPrefix(). This one would be an indication for us to info or not in the afterPropertiesSet(). And the last one won't be required from the tests.

Not sure i understand, can this class be an InitializingBean or not, meaning is the use case where `DefaultKafkaProducerFactory` is _not_ a spring bean valid ?

artembilan:
No, that use-case is really not valid. Excluding our unit-tests, of course.
See `DefaultKafkaProducerFactory` contract:
```
public class DefaultKafkaProducerFactory<K, V> implements ProducerFactory<K, V>, ApplicationContextAware,
		ApplicationListener<ContextStoppedEvent>, DisposableBean {
```

So, it expects `ApplicationContext`, a `ContextStoppedEvent` even and it also does already something in the `destroy()` callback.
Therefore I didn't expect any problems with introducing an `afterPropertiesSet()`.
But if our tests doesn't initialize it properly, it doesn't mean that we need to go through the whole project and fix those tests to really present this `DefaultKafkaProducerFactory` as a bean.

Saying that I mean that let's try find a middle ground here!
Let's consider some logic with an additional internal `transactionIdPrefixExplicitlySet` flag to check in the mentioned `afterPropertiesSet()`! This way we won't need to change tests since that new callback is going to have only INFO logging logic.
And also we need to keep in mind a backward compatibility. We can't have guarantees that everybody uses it as a bean. Such a new `afterPropertiesSet()` with moved logic over there will break their apps.
So, while I agree with you about a robust solution, it would be better to do something what would fit your original requirements and won't break anyone else.

Does it make sense?

jorgheymans:
see #1286

PR #1286, opened 2019-10-29
Spring kafka 1270 alternative

jorgheymans:
@artembilan so i had a second thought about this and decided that the best course of action would be to remove the statement alltogether. Printing the logging statement in correct fashion would require refactoring of the way the property is set, because the constructor calls setTransactionIdPrefix as well meaning we can't just use a toggle variable for this. I guess you could set the transactionIdPrefix using a separate method, and have both the ctor and `setTransactionIdPrefix` call this method with each their own boolen "isSet" variable, and then in `afterPropertiesSet` work things out but it's just not worth it. So i'll close my other PR, feel free to apply or not this one.


artembilan:
OK. Thanks for the feedback and your effort.

I think we can't just remove that info message since @garyrussell had something in mind introducing it anyway.

My point was to have that `boolean` flag and toggle it in the mentioned `setTransactionIdPrefix()`, but don't call this method in the ctor in favor of its logic just in place. 
This way we won't affect the mentioned `boolean` flag and won't need drastic changes throughout the project.

Does it make sense to you?

garyrussell:
Actually, it was your PR @artembilan https://github.com/spring-projects/spring-kafka/pull/1065

Given that we've just done something similar for `client.id`; I don't mind removing this log.

artembilan:
OK, sorry for misleading.

So, yeah, if we all agree to remove let's rebase this PR and accept it after green light from Travis!

artembilan:
@jorgheymans ,

any chances that you rebase your branch against latest `master`, fix conflicts and repush into this PR?
So, we will have a clean merge situation.

Thank you!

jorgheymans:
ok i'll have a look.

garyrussell:
Rebased and merged as 4ad074077e2c7eff8bde0f72ea5b87d4c41b653d

Sorry for the delay; completely forgot about this.

Issue #1236, opened 2019-09-13
Kafka Streams create store support


nyilmaz:
Hi there!

I'm trying to achieve adding stores (in my case global one) via `StreamsBuilder`. I think adding state store and global store should be handled inside `StreamsBuilderFactoryBean`. Because those operations do indeed change underlying topology. My first question would be; how can we safely add store to `StreamsBuilder` before it is created here: https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java#L313

I have in mind for usage;

Provide `KafkaStreamsStoreCustomizer`(s) having `customize(StreamsBuilder)` and call this before we initialize `Topology`.

Any comments, suggestions?

artembilan:
I wonder if this is something what can help you from Kafka Streams Binder perspective: https://cloud.spring.io/spring-cloud-static/spring-cloud-stream-binder-kafka/3.0.0.M3/reference/html/spring-cloud-stream-binder-kafka.html#_state_store

Although I won't mind if you provide more context how that `KafkaStreamsStoreCustomizer` could help us.

Thanks

nyilmaz:
Well, it seems the kafka-streams setup occurs inside `KafkaStreamsStreamListenerSetupMethodOrchestrator`. Currently with only spring-kafka and kafka-streams dependencies, I got a custom store working like;

```
@Configuration
class KafkaStreamsConfiguration {

    @Bean("categoryStreamsBuilderFactoryBean")
    @Primary
    public StreamsBuilderFactoryBean streamsBuilderFactoryBean(@Qualifier(CONFIG_BEAN_NAME) StreamsConfig streamsConfig)
        throws Exception {
      StreamsBuilderFactoryBean factoryBean = new StreamsBuilderFactoryBean(streamsConfig);
      factoryBean.setStateRestoreListener(new CategoryStateRestoreListener());
      factoryBean.afterPropertiesSet();
      return factoryBean;
    }
}


@Configuration
class CategoryKafkaStreamsConfiguration {

    private final StreamsBuilder streamsBuilder;

    private final String categoryTopicName;

    private final String categoryStoreName;

    @PostConstruct
    public void init() {
        StoreBuilder<KeyValueStore<Long, Category>> categoryStoreBuilder =
            Stores.keyValueStoreBuilder(inMemoryKeyValueStore(this.categoryStoreName),
[...]

PR #1305, opened 2019-11-08
Proposal: Support Kafka Streams Topology customizer

renatomefi:
The streams Topology is current in a limited scope, there are use cases where the user might want to customize it, by adding processors that influence GlobalKTables which is my case.

I think it can possibly assist on use cases like https://github.com/spring-projects/spring-kafka/issues/1236

Please let me know if it's a valid idea so I can proceed with documentation and more tests, the current test is very simple and actually doesn't represent the power of the Topology customization, but it's the fastest functional example I can write while the idea is evaluated!

I'm glad to hear your thoughts, thanks you!

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/KafkaStreamsTopologyCustomizer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/KafkaStreamsTopologyCustomizer.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2018-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.config;
+
+import org.apache.kafka.streams.KafkaStreams;
+import org.apache.kafka.streams.Topology;
+
+/**
+ * Callback interface that can be used to configure {@link KafkaStreams} directly.
+ *
+ * @author Renato Mefi
+ *
+ * @since 2.4.0
+ *
+ * @see StreamsBuilderFactoryBean
+ */
+@FunctionalInterface
+public interface KafkaStreamsTopologyCustomizer {
+
+	void customize(Topology topology);
+
+}
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/CompositeKafkaStreamsTopologyCustomizer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/CompositeKafkaStreamsTopologyCustomizer.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2018-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


pivotal-issuemaster:
@renatomefi Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1305)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1305) to manually synchronize the status of this Pull Request.

garyrussell:
@renatomefi We'd be happy to take a look, but we're not allowed to even look at your code until you sign the CLA. Thanks.

pivotal-issuemaster:
@renatomefi Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1305)!

renatomefi:
@garyrussell could you please check then? thank you!

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaStreamsTopologyCustomizer.java
```diff
@@ -0,0 +1,36 @@
...
+ * Callback interface that can be used to configure {@link KafkaStreams} directly.
+ *
+ * @author Renato Mefi
+ *
+ * @since 2.4.0
+ *
+ * @see StreamsBuilderFactoryBean
+ */
+@FunctionalInterface
+public interface KafkaStreamsTopologyCustomizer {
```
artembilan: Why can't we leave just with the `java.util.function.Consumer`?
Even for composition we can use its `Consumer<T> andThen(Consumer<? super T> after)`.

garyrussell:
2.4 will be released soon; the 2.4.0 clients are not binary compatible with code compiled against the 2.3.0 clients, so 2.4. will be (mostly) functionally the same as 2.3.

The 2.4 clients are currently scheduled to be released 11/15; I have scheduled 2.4 RC1 for 11/22.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaStreamsTopologyCustomizer.java
renatomefi: Thanks for the info!

Given that, I followed the same interface as `KafkaStreamsCustomizer` for user experience, should I still try with the `java.util.function.Consumer`?

Given the added info by @garyrussell I think we have 2 paths: Follow with the current inferface or breaking all the customizers with your suggestion!

Which path should I take?

artembilan: I think the `KafkaStreamsCustomizer` abstraction has been chosen for Spring Boot auto-configuration  purposes.
The same might happened with this option as well.
But I think it won't hurt to stick with the particular generic for the `Consumer`.
I even don't think that this customization option should be considered as a bean in the target application configuration: a lambda for consumer could do the trick with access to any possible bean in the configuration.

If we don't go a `java.util.function.Consumer` approach, I don't think we need that `CompositeKafkaStreamsTopologyCustomizer`: the same `andThen()` implementation really will lead us to the same result. Even if we will end up with a bean, which could be useful in Spring Boot auto-configuration.

garyrussell: I don't see any problems adding this to 2.3; I generally concur with Artem but, since we already have a precedent of an explicit interface, I guess it would be cleaner to stick with it here too. I don't see any reason to break the API.

artembilan: Good. Let's stick with an explicit functional interface then!
But let's try to implement that `andThen()` instead of composite! 😉

altayeb1980:
@garyrussell based on my question
https://stackoverflow.com/questions/59689821/adds-a-state-store-to-the-underlying-topology-before-streamsbuilder-started
Is there a plan for this PR to continue?

garyrussell:
@renatomefi Thanks for the contribution!

Closed in favor of https://github.com/spring-projects/spring-kafka/pull/1350

Also see https://stackoverflow.com/questions/59689821/adds-a-state-store-to-the-underlying-topology-before-streamsbuilder-started/59690035#59690035

PR #1316, opened 2019-11-18
Upgrade to 2.4 clients

garyrussell:
- Remove references to ZkClient
- deprecation of `Consumer.committed(TopicPartition tp)`
- Addition of `onPartitionsLost` to `ConsumerRebalanceListener`
- since `onPartitionsLost` calls `onPartitionsRevoked` by default with a
  possible partial list, remove the revoked partitions from the assignments
- fix some reflection uses in tests (`groupId` field and `NewTopic`)
- fix `EKIT` `listen12` listener - we now always get the full batch
- `ConsumerRebalanceListener.onPartitionsRevoked` is no longer called with an
  empty collection
- deprecation of some `TopologyDriver` helper classes
- update what's new and change log appendix
- add `unregisterSeekCallback` to `ConsumerSeekAware`
  - `AbstractCSA` now removes TL there since we may have only a partial revoke

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ ext {
 	jaywayJsonPathVersion = '2.4.0'
 	junit4Version = '4.12'
 	junitJupiterVersion = '5.5.2'
-	kafkaVersion = '2.4.0-SNAPSHOT'
+	kafkaVersion = '2.4.0'
 	log4jVersion = '2.12.1'
 	micrometerVersion = '1.3.2'
 	mockitoVersion = '3.0.0'
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ ext {
 	jaywayJsonPathVersion = '2.4.0'
 	junit4Version = '4.12'
 	junitJupiterVersion = '5.5.2'
-	kafkaVersion = '2.3.1'
+	kafkaVersion = '2.4.0-SNAPSHOT'
 	log4jVersion = '2.12.1'
 	micrometerVersion = '1.3.2'
 	mockitoVersion = '3.0.0'
@@ -90,6 +90,7 @@ allprojects {
 		if (version.endsWith('BUILD-SNAPSHOT')) {
 			maven { url 'https://repo.spring.io/libs-snapshot' }
 		}
+		mavenLocal()
 	}
 
 }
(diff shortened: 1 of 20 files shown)


src/reference/asciidoc/whats-new.adoc
```diff
@@ -1,171 +1,29 @@
 === What's New in 2.3 Since 2.2
```
artembilan:
```suggestion
=== What's New in 2.4 Since 2.3
```

garyrussell:
Agreed (merge after release).

garyrussell:
2.4.0 clients were released yesterday; ready for final review and merge.

build.gradle
```diff
@@ -90,6 +90,7 @@ allprojects {
 		if (version.endsWith('BUILD-SNAPSHOT')) {
 			maven { url 'https://repo.spring.io/libs-snapshot' }
 		}
+		mavenLocal()
```
artembilan: I think this one should go now

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
```diff
@@ -146,10 +146,22 @@ public ReactiveKafkaConsumerTemplate(ReceiverOptions<K, V> receiverOptions) {
...
+	/**
+	 * Get the committed {@link OffsetAndMetadata} for the partition.
+	 * @param partition the partition.
+	 * @return the {@link OffsetAndMetadata}.
+	 * @deprecated in favor of {@link #committed(Set)}.
+	 */
+	@Deprecated
+	@SuppressWarnings("deprecation")
 	public Mono<OffsetAndMetadata> committed(TopicPartition partition) {
 		return doOnConsumer(consumer -> consumer.committed(partition));
```
artembilan: Can we just build a `Set` here and call that new method instead of `@SuppressWarnings` ?

garyrussell: That's not so easy because we'd have to convert a Mono of Map to a Mono of the single value.

I think this is easier.

PR #1317, opened 2019-11-18
Add parameter names to default object mapper

jehrhardt:
The parameter names module allows to use constructor parameter names during deserialization. This is one of the basic modules that should typically be added to an object mapper to make things work as expected.

The current workaround is to provide a custom object mapper instead.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/JacksonUtils.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/JacksonUtils.java
@@ -76,6 +76,15 @@ public final class JacksonUtils {
 		}
 
 		try {
+			Class<? extends Module> parameterNamesModule = (Class<? extends Module>)
+					ClassUtils.forName("com.fasterxml.jackson.module.paramnames.ParameterNamesModule", classLoader);
+			objectMapper.registerModule(BeanUtils.instantiateClass(parameterNamesModule));
+		}
+		catch (@SuppressWarnings(UNUSED) ClassNotFoundException ex) {
+			// jackson-parameter-names not available
+		}
+
+		try {
 			Class<? extends Module> javaTimeModule = (Class<? extends Module>)
 					ClassUtils.forName("com.fasterxml.jackson.datatype.jsr310.JavaTimeModule", classLoader);
 			objectMapper.registerModule(BeanUtils.instantiateClass(javaTimeModule));


spring-kafka/src/main/java/org/springframework/kafka/support/JacksonUtils.java
```diff
@@ -75,6 +75,15 @@ private static void registerWellKnownModulesIfAvailable(ObjectMapper objectMappe
 			// jackson-datatype-jdk8 not available
 		}
 
+		try {
+			Class<? extends Module> parameterNamesModule = (Class<? extends Module>)
+					ClassUtils.forName("com.fasterxml.jackson.module.paramnames.ParameterNamesModule", classLoader);
```
artembilan: Well, our goal was to be consistent with Spring Framework.
See its `Jackson2ObjectMapperBuilder`. It doesn't load this module by default.

I understand that it is a bit cumbersome to inject your own `ObjectMapper`, but I believe it is better to always inject whatever is auto-configured with Spring Boot instead of trying to pursue all the feature in the particular out-of-the-box instance here in this specific project.

Does it make sense to you?

garyrussell:
Perhaps we should add something like this to `JacksionUtils`...

```java
	private static ObjectMapperCustomizer customizer = om -> om;

	public static void setCustomizer(ObjectMapperCustomizer cusomizer) {
		JacksonUtils.customizer = customizer;
	}

	public interface ObjectMapperCustomizer {

		ObjectMapper configure(ObjectMapper mapper);

	}

```

?

It would make it easier to customize the OM, when Kafka instantiates the deserializer.

spring-kafka/src/main/java/org/springframework/kafka/support/JacksonUtils.java
jehrhardt: Yes, that makes sense.

Although I do not understand, why kafka-spring does not simply use the object mapper from Spring MVC. Then you do not need to configure two mappers.

artembilan: Because there is just no any MVC in Spring Kafka. Therefore we can't use that dependency in the project.
You are feel free to inject that MVC's instance to reuse its configuration.

artembilan:
Gary, `static` is done against the whole JVM: we can't assume that end-users are going to have only one application in their JVMs.
So, mutating in one application will definitely effect an another one.

I don't have a good answer right now for the problem, unless we may agree to parse some Jackson-specific options from the `public void configure(Map<String, ?> configs, boolean isKey)` impl...

spring-kafka/src/main/java/org/springframework/kafka/support/JacksonUtils.java
jehrhardt: That's not what I mean.

Spring MVC uses a bean of type `ObjectMapper` by default. If this bean is defined, it will be used. So why doesn't spring-kafka default to an existing bean in the application context?

artembilan: Where do you see that?
AFAIK, auto-configuration comes from Spring Boot. 
I'm fully not sure that Spring MVC does something conditional...

jehrhardt: Ok, I might confuse things here.

In a Spring Boot web application I have an object mapper bean by default. But I can just define my own object mapper bean and it is automatically used.

What I would love to see is that spring-kafka uses that bean automatically as well in my Spring Boot application. Currently I have to do more than just defining my own object mapper bean.

And of course I would expect spring-kafka to use the default object mapper bean in a Spring Boot web application as well 😉.

But that's not part of this pull request.

artembilan: Well, at least we are close to the same page. 😉 
As you have noticed that it is indeed a part of Spring Boot to auto-configure Spring MVC with an appropriate `ObjectMapper` bean. I wonder why do you think that it is this project responsibility to auto-configure that bean, but not the same Spring Boot one?

If you have some concerns that Spring Boot doesn't auto-inject an `ObjectMapper`  bean into Spring Kafka components, then such an issue has to be raised against Spring Boot, but definitely not here.

Now let's see what we still can do here, since we all agreed that Spring MVC doesn't inject an `ObjectMapper` bean into its components and its default one is exactly the same what we have here in Spring Kafka!

jehrhardt: I will request the same change in the Spring project. So maybe this pull request might not become obsolete.

And for the auto config issue I will reach out to the Spring Boot project.

Thanks for the explanations.

jehrhardt:
I have created a pull request for Spring as well: spring-projects/spring-framework#24019

spring-kafka/src/main/java/org/springframework/kafka/support/JacksonUtils.java
garyrussell: Bear in mind, when Kafka creates the deserializer instead of Spring (when you configure it entirely with boot kafka consumer properties), it can't use the Boot `ObjectMapper` (Kafka doesn't know anything about Spring).

If you want to use the Boot instance, you need to define your own `DefaultKafkaConsumerFactory`, overriding Boot's - and add custom deserializers in the constructor. That way, you can use Boot's auto-configured mapper.

PR #1318, opened 2019-11-19
ValidatingReplyingKafkaTemplate

trajano:
This is a`ReplyingKafkaTemplate` that adds a simple validation semantic so it can take multiple responses and choose the first one that matches the validation condition.

The use case for this would be providing a farm of topic responders which are decoupled from the calling service and the calling service does not know who would respond and when, but knows some property of the response to consider it *valid*.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ValidatingReplyingKafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ValidatingReplyingKafkaTemplate.java
@@ -81,14 +81,14 @@ public class ValidatingReplyingKafkaTemplate<K, V, R> extends ReplyingKafkaTempl
 	/**
 	 * Filter out records that do not pass the validation predicate.
 	 * <p>
-	 * This does an initial filter to make sure only the ones with a correlation ID defined is processed.  This does
-	 * <b>not</b> check whether the correlation ID is something that needs to be considered as {@code futures} is not
-	 * accessible and it is relying on the super class to perform the extra test.
+	 * This does an initial filter to make sure only the ones with a correlation ID defined is processed and
+     * ensure that the correlation ID is inside the {@code futures} collection.
 	 */
 	@Override
 	public void onMessage(List<ConsumerRecord<K, R>> data) {
 		super.onMessage(data.stream()
 			.filter(record -> record.headers().lastHeader(KafkaHeaders.CORRELATION_ID) != null)
+			.filter(record -> this.futures.containsKey(new CorrelationKey(record.headers().lastHeader(KafkaHeaders.CORRELATION_ID))))
 			.filter(record -> this.validationPredicate.test(record.key(), record.value()))
 			.collect(Collectors.toList()));
 	}
--- a/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ValidatingReplyingKafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ValidatingReplyingKafkaTemplate.java
@@ -24,7 +24,6 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 
 import org.springframework.kafka.core.ProducerFactory;
 import org.springframework.kafka.listener.GenericMessageListenerContainer;
-import org.springframework.kafka.support.KafkaHeaders;
 
 /**
  * This is a {@link ReplyingKafkaTemplate} that adds a simple validation semantic so it can take multiple responses and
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@trajano Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1318)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1318) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@trajano Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1318)!

trajano:
One part I don't like is the prefiltering. I wish there was a `onCorrelatedMessage()` which had done the initial filters already.

garyrussell:
Thanks.

See the travis report.

Use `./gradlew check` to validate (checkstyle rules).

Don't squash commits; just add more; we squash during merge after review.

garyrussell:
>One part I don't like is the prefiltering.

Feel free to refactor the super class to provide that facility.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ValidatingReplyingKafkaTemplate.java
```diff
@@ -0,0 +1,77 @@
...
+ * <li>0:05 Kid: okay I gathered a few answers, 
+ * <li>0:05 Kid: Filter out who can't answer 2+2
+ * <li>0:05 Kid: The proper answer is 12, 4
+ * <li>0:06 Uncle 3: 12, 4
+ * <li>0:06 Kid: Sorry uncle 3 you're too slow, so I am ignoring you
+ * </ul>
+ */
+public class ValidatingReplyingKafkaTemplate<K, V, R> extends ReplyingKafkaTemplate<K, V, R> {
+
+    private final Function<ConsumerRecord<K, R>, Boolean> validationMethod;
```
garyrussell: You could use a `BiPredicate` here.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ValidatingReplyingKafkaTemplate.java
```diff
@@ -0,0 +1,77 @@
...
+        super.onMessage(data.stream()
+                .filter(this::isCorrelationIdMatch)
+                .filter(record -> validationMethod.apply(record))
+                .collect(Collectors.toList()));
+    }
+
+    private boolean isCorrelationIdMatch(ConsumerRecord<K, R> record) {
+        Iterator<Header> iterator = record.headers().iterator();
+        CorrelationKey correlationId = null;
+        while (correlationId == null && iterator.hasNext()) {
```
garyrussell: We've changed this on master to use `Header correlationHeader = record.headers().lastHeader(...)`. We should probably do the same here.

trajano: Would that work on the older version of spring-kafka?

garyrussell: Yes (but not with the configurable header name, just the standard header name).

trajano:
Ugh this won't work.  Because all I checked was correlationID exists.  I need the `futures` to check if it is something I want.

garyrussell:
Right, but you can pull it up as a `protected final` method to the super class which has access to `futures`.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ValidatingReplyingKafkaTemplate.java
trajano: > 
> 
> Right, but you can pull it up as a `protected final` method to the super class which has access to `futures`.

That would mean I would have to change the existing code which is what I am trying to avoid.  But okay

garyrussell: Of course, but there's no other way to do it aside from keeping your own copy of `futures`. Which is fine, if you just want to do that locally, but in the framework, we need to do it properly.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ValidatingReplyingKafkaTemplate.java
trajano: Just checked it should workl

garyrussell: Yes; making `futures` `protected` will work too but we have to mark it with `// NOSONAR` so our SonarQube checks won't bark.

However, now I think about it, we don't really need a subclass; we can just add a setter on `ReplyingKafkaTemplate`

    setValidationPredicate(BiPredicate<K, R> validationPredicate) {...}

with the default being `(key, value) -> true`.

garyrussell: Actually, I think it needs to be `Predicate<ConsumerRecord<K, R>` because the validation might want to look at headers.

trajano:
Closing based on this comment https://github.com/spring-projects/spring-kafka/pull/1318#discussion_r348095633

Issue #1323, opened 2019-12-03
Is it possible to use `spring-kafka-test` without AssertJ?


levelast:
Is it possible to use `spring-kafka-test` without AssertJ?
I've project with kotlin and excluded assertj from `spring-boot-starter-test`, 'cause we use more kotlin-style assert library and code is really messy when you have two strict library at once. Right now after added dependency `spring-kafka-test` and enable it, I've `java.lang.NoClassDefFoundError: org/assertj/core/util/Arrays`. The problem's line of code:
https://github.com/spring-projects/spring-kafka/blob/cda70d4730c77ce8a994631e2fb446798a220e05/spring-kafka-test/src/main/java/org/springframework/kafka/test/condition/EmbeddedKafkaCondition.java#L110
It's strange to use assertj here, but maybe it's not the one problem. 
Is it better to include asstertj to test scope?

Sorry for my English =)

garyrussell:
Thanks; it's a bug (wrong import - probably an artifact of the IDE auto-selecting.

artembilan:
I think we need to revise all the classes in `spring-kafka-test` for an explicit `org.assertj` imports.
For example `KafkaTestUtils` really can be used in the same Kotlin-based tests, but really without an AssertJ import.

Also it looks like the current fix won't let us go too far because an `EmbeddedKafkaBroker` still has that `import static org.assertj.core.api.Assertions.assertThat;`.

I would even say that all those utils have to be without any assertions. This is really a target test responsibility to check data and state explicitly.

PR #1324, opened 2019-12-03
GH-1323: s-k-test remove hard AssertJ dependency

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1323

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
@@ -16,8 +16,6 @@
 
 package org.springframework.kafka.test;
 
-import static org.assertj.core.api.Assertions.assertThat;
-
 import java.io.File;
 import java.io.IOException;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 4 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
```diff
@@ -546,11 +544,12 @@ public void consumeFromAnEmbeddedTopic(Consumer<?, ?> consumer, String topic) {
...
-		HashSet<String> diff = new HashSet<>(Arrays.asList(topicsToConsume));
-		diff.removeAll(new HashSet<>(this.topics));
-		assertThat(this.topics)
-				.as("topic(s):'" + diff + "' are not in embedded topic list")
-				.containsAll(new HashSet<>(Arrays.asList(topicsToConsume)));
+		List<String> notEmbedded = Arrays.stream(topicsToConsume)
+				.filter(topic -> !this.topics.contains(topic))
+				.collect(Collectors.toList());
+		if (notEmbedded.size() > 0) {
+			throw new IllegalStateException("topic(s):'" + notEmbedded + "' are not in embedded topic list");
```
artembilan: Shouldn't we JavaDoc this behavior? It's not a big breaking change though, but would be great to have some note about this exception.

Thanks

Issue #1102, opened 2019-05-20
KafkaListenerEndpointRegistry should use its own lifecycle state


garyrussell:
See https://stackoverflow.com/questions/56153611

PR #1327, opened 2019-12-06
KafkaListenerEndpointRegistry isRunning state is incorrect after calling stop(Runnable)

asibross:


spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistry.java
```diff
@@ -268,9 +268,14 @@ public void stop() {
 
 	@Override
 	public void stop(Runnable callback) {
+		Runnable callbackWrap = () -> {
+			this.running = false;
```
artembilan: The bug is confirmed. Thank you!

I think we don't need this wrapper and simple `this.running = false;` in the end of this `stop()` method should be enough.

It looks like this property was introduced here: https://github.com/spring-projects/spring-kafka/issues/1102

Sounds like `RabbitListenerEndpointRegistry` and `JmsListenerEndpointRegistry` have to be fixed the same way.
/CC @garyrussell

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaListenerEndpointRegistryLifecycleTests.java
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.kafka.annotation.EnableKafka;
+import org.springframework.kafka.annotation.KafkaListener;
+import org.springframework.kafka.config.ConcurrentKafkaListenerContainerFactory;
+import org.springframework.kafka.config.KafkaListenerEndpointRegistry;
+import org.springframework.kafka.core.ConsumerFactory;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+
```
artembilan: Need `@author` and `@since`.
As well as your name to the `@author` list of another class.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistry.java
asibross: According the `SmartLifecycle` stop method documentation: 
"The callback <b>must</b> be executed after the {@code SmartLifecycle} component does indeed stop".
I figure this means the isRunning status should already be changed to false before the callback is called.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaListenerEndpointRegistryLifecycleTests.java
asibross: Done.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistry.java
artembilan: Hm. Good point!
So, how about to change the flag in the beginning of the method then instead?
Technically it doesn't matter how do we process `stop()` for those internal containers: this registry is stopped already because of its `stop()` method call.

asibross: It feels a bit off to change the state before the actual stopping of the underlying containers.
Also, you cannot rely on the registry stop() method to be called, in fact the LifecycleProcessor will not call the stop() method but rather the stop(Runnable) variant when the lifecycle object is SmartLifecycle

artembilan: Well, I think that's right. This way any other concurrent attempts to stop this component will exit earlier because the state is already switched even if it has not finished yet.
I mean we need to have `this.running = false;` in the beginning of the method.

@garyrussell , WDYT?

asibross: If this is the reasoning, then the other stop method method should also change the state before actually stopping the containers?

asibross: I removed the wrapper and changed the state before calling the callback in two places. Please let me know what you think.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistry.java
```diff
@@ -270,7 +271,10 @@ public void stop() {
 	public void stop(Runnable callback) {
 		Collection<MessageListenerContainer> listenerContainersToStop = getListenerContainers();
 		if (listenerContainersToStop.size() > 0) {
-			AggregatingCallback aggregatingCallback = new AggregatingCallback(listenerContainersToStop.size(), callback);
+			AggregatingCallback aggregatingCallback = new AggregatingCallback(listenerContainersToStop.size(), () -> {
+				this.running = false;
```
artembilan: Well, you still have this logic inside a callback call and that's exactly what you try to justify is wrong anyway.
My point is to have a logic like this:
```
public void stop(Runnable callback) {
this.running = false;
...
```

So, this way we are stopped and it the state is changes exactly before we call a callback.

asibross: Ok, will change.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistry.java
garyrussell: I would say we can do what @artembilan suggests and move the `isRunning=false` in the simple `stop()` to the top as well, to make it consistent.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/16a12fe33fc8dfafb06a14ce38bf905a9ffcdbee.

@asibross ,

PR #1337, opened 2019-12-17
GH-1336: Introduce authExceptionRetryInterval to retry on auth errors

### LINKED ISSUE #1336: Allow denied authorization decisions to be retried
**Affects Version(s):** 2.3.5 and earlier

🎁 Enhancement
To allow for custom, more dynamic authorization models in Kafka, it would be great to have an option to allow the client to retry a failed authorization decision after some configured time has passed.

lukasz-kaminski:
Resolves #1336

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
@@ -38,6 +38,7 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  * @author Artem Yakshin
  * @author Johnny Lim
+ * @author Lukasz Kaminski
  */
 public class ContainerProperties extends ConsumerProperties {
 
@@ -181,6 +182,8 @@ public class ContainerProperties extends ConsumerProperties {
 	private Duration consumerStartTimout = DEFAULT_CONSUMER_START_TIMEOUT;
 
 	private boolean subBatchPerPartition;
+
+	private Duration authorizationExceptionRetryInterval;
 
 	/**
 	 * Create properties for a container that will subscribe to the specified topics.
@@ -600,13 +603,31 @@ public class ContainerProperties extends ConsumerProperties {
 	 * When using a batch message listener whether to dispatch records by partition (with
 	 * a transaction for each sub batch if transactions are in use) or the complete batch
 	 * received by the {@code poll()}. Useful when using transactions to enable zombie
-	 * fencing, by using a {code transactional.id} that is unique for each
+	 * fencing, by using a {@code transactional.id} that is unique for each
 	 * group/topic/partition.
 	 * @param subBatchPerPartition true for a separate transaction for each partition.
 	 * @since 2.3.2
 	 */
 	public void setSubBatchPerPartition(boolean subBatchPerPartition) {
 		this.subBatchPerPartition = subBatchPerPartition;
+	}
+
+	public Duration getAuthorizationExceptionRetryInterval() {
+		return this.authorizationExceptionRetryInterval;
+	}
+
+	/**
+	 * Set the interval between retries after {@code AuthorizationException} is thrown
+	 * by {@code KafkaConsumer}. By default the field is null and retries are disabled.
+	 * In such case the container will be stopped.
+	 *
+	 * The interval must be less than {@code max.poll.interval.ms} consumer property.
+	 *
+	 * @param authorizationExceptionRetryInterval the duration between retries
+	 * @since 2.4.0
+	 */
+	public void setAuthorizationExceptionRetryInterval(Duration authorizationExceptionRetryInterval) {
+		this.authorizationExceptionRetryInterval = authorizationExceptionRetryInterval;
 	}
 
 	@Override
@@ -631,6 +652,7 @@ public class ContainerProperties extends ConsumerProperties {
 				+ (this.scheduler != null ? ", scheduler=" + this.scheduler : "")
 				+ ", noPollThreshold=" + this.noPollThreshold
 				+ ", subBatchPerPartition=" + this.subBatchPerPartition
+				+ ", authorizationExceptionRetryInterval=" + this.authorizationExceptionRetryInterval
 				+ "]";
 	}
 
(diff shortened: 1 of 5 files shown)


garyrussell:
Please see the checkstyle violations in the [travis report](https://travis-ci.org/spring-projects/spring-kafka/builds/626338995?utm_source=github_status&utm_medium=notification).

You can run checkstyle locally with `./gradlew clean check`.

lukasz-kaminski:
@garyrussell My bad, sorry. Should be fine now.

anderseknert:
Nice one @lukasz-kaminski. This should be the final piece of the puzzle for us to resume our adventures in Kafka authorization land. About authorization.. I mentioned before that I'm not too familiar with how Kafka works, but on the other hand I'm quite deeply rooted in the authentication/authorization space, having worked the last 5 years or so primarily with identity systems. If I may pick a nit based on that it would be the use of "auth" to mean authorization. Though authorization often follows authentication, they are different beasts and should be treated as such. Using "auth" for one of them (or both combined) has a tendency to create confusion, and it also needlessly reserves the "auth" prefix should you need to add other properties for authentication later. So in this case I'd pay the price of verbosity and a few extra characters, but that's of course up to you and @garyrussell. Looking forward to trying this out!

garyrussell:
@anderseknert I agree we should lengthen the property name; sorry about that.

@lukasz-kaminski 

After further review; I don't think we can use pause/resume the consumer; we just have to `Thread.sleep()` (hence my comment about limiting the time to `max.poll.interval.ms` - I had not noticed you used pause/resume in your earlier PR).

The problem is that, with a `GroupAuthorizationException`, we'll continue to `poll()` the consumer and the consumer coordinator will try to send another `JoinGroupRequest` (which will fail again, ad nauseam).

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -910,9 +914,16 @@ public void run() {
 					break;
 				}
 				catch (AuthorizationException ae) {
-					this.fatalError = true;
 					ListenerConsumer.this.logger.error(ae, "Authorization Exception");
-					break;
+					if (this.authExceptionRetryInterval == null) {
+						ListenerConsumer.this.logger.debug("No authExceptionRetryInterval provided");
```
garyrussell: You lost the `ae` in the log so we won't know what kind it is.

lukasz-kaminski: I'll move it from line 914 to both log entries.

lukasz-kaminski:
> After further review; I don't think we can use pause/resume the consumer; we just have to `Thread.sleep()` (hence my comment about limiting the time to `max.poll.interval.ms` - I had not noticed you used pause/resume in your earlier PR).
>
> The problem is that, with a GroupAuthorizationException, we'll continue to poll() the consumer and the consumer coordinator will try to send another JoinGroupRequest (which will fail again, ad nauseam).

@garyrussell Nice catch. I wanted to avoid using `Thread.sleep()` as we lose pausing and resuming events, which we are, in fact, doing. I'll look if we can also patch GroupAuthorizationException problem and if I don't think of anything clever, I'll just use `sleep()`. Is that ok?

@anderseknert thanks for pointing that out, it's an important distinction.

lukasz-kaminski:
Since `KafkaConsumer` doesn't seem to take pausing into account when committing, we would have to either use `sleep` or guard whole `pollAndInvoke()` with `if (isPaused())`, which would result in busy waiting. `sleep()` it is, then.

lukasz-kaminski:
@garyrussell Now it's using `sleep()`. I've checked both group and topic authorization exceptions and the interval works fine. Please note, I've removed one test case that checked the interval as it would just test invoking `Thread.sleep()` and I didn't want to bring PowerMockito just for that.

src/reference/asciidoc/kafka.adoc
```diff
@@ -757,6 +757,11 @@ It does not apply if the container is configured to listen to a topic pattern (r
 Previously, the container threads looped within the `consumer.poll()` method waiting for the topic to appear while logging many messages.
 Aside from the logs, there was no indication that there was a problem.
 
+As of version 2.4.0, a new container property called `authorizationExceptionRetryInterval` has been introduced. This causes the container to retry fetching messages after getting any `AuthorizationException` from `KafkaConsumer`. This can happen when, for example, given user is denied access to read certain topic.
+Defining `authorizationExceptionRetryInterval` should help the application to recover as soon as proper permissions are granted.
```
garyrussell: Our doc policy is one sentence per line - will fix on merge.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -609,6 +612,24 @@ public void setSubBatchPerPartition(boolean subBatchPerPartition) {
...
+
+	/**
+	 * Set the interval between retries after {@code AuthorizationException} is thrown
+	 * by {@code KafkaConsumer}. By default the field is null and retries are disabled.
+	 * In such case the container will be stopped.
+	 *
+	 * The interval must be less than {@code max.poll.interval.ms} consumer property.
+	 *
+	 * @param authorizationExceptionRetryInterval the duration between retries
+	 * @since 2.4.0
```
garyrussell: I will back-port this to the 2.3.x branch so I'll change these 2.4.0 references to 2.3.5.

garyrussell:
Thanks for the contribution! Looking forward to more; in future, please limit commit headlines to 50 chars; thanks.

Merged as d26dba2853d59b32b90e83023ef2516b48018ef6 with minor doc polishing; I also pulled the new property up to the `ConsumerProperties` super class  so we can make a similar change in spring-integration-kafka's polled `KafkaMessageSource`.

Backported to 2.3.x as 930e201b42771dd727e6b6c224a6ad896f57b4e8

anderseknert:
Thanks a lot @lukasz-kaminski and thanks a lot @garyrussell 👍 Will test this out ASAP and roll out at scale after. Happy holidays to the both of you!

garyrussell:
@anderseknert You too.

I am expecting to release the 2.4 release candidate (2.4.0.RC1) today; the 2.3.5 release is scheduled for Jan 15.

Issue #1319, opened 2019-11-20
Add validation function to aggregating template


garyrussell:
See #1318 

Deprecate `releaseStrategy` and replace with `Function<List<ConsumerRecord<K, V>>, List<ConsumerRecord<K, V>> validatorFunction;`.

Call the validator function each time a new record for the correlation id arrives and when the request times out. The function allows releasing (a possibly modifield) list at those times; returning `null` means don't release.

trajano:
I'm not keen on the `null` part.  Can `Optional` semantics be used instead?

garyrussell:
Yes, we could do that, or we could use `Predicate<List<ConsumerRecords<K, V>>>` and the predicate could manipulate the `List` argument directly, returning true when it's time to release the list.

PR #1338, opened 2019-12-17
GH-1319: Agg. Replying Template improvement

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1319

- change the release strategy to a `BiConsumer`
- call the strategy during normal message delivery as before
- now call the strategy when releasing a partial result after timeout
- allow the strategy to modify the list (change from `Collection` to `List`).

**cherry-pick to 2.3.x**


artembilan:
... and cherry-picked into `2.3.x` after fixing conflict in `whats-new.adoc`

PR #1350, opened 2020-01-13
GH-1236: Add Streams infrastructure customizer

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1236

Allow modification of the builder and/or topolgy before the stream is
created.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/CompositeKafkaStreamsInfrastructireCustomizer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/CompositeKafkaStreamsInfrastructireCustomizer.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2018-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.config;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.kafka.streams.KafkaStreams;
+import org.apache.kafka.streams.StreamsBuilder;
+import org.apache.kafka.streams.Topology;
+
+/**
+ * Composite {@link KafkaStreamsCustomizer} customizes {@link KafkaStreams} by delegating
+ * to a list of provided {@link KafkaStreamsCustomizer}.
+ *
+ * @author Gary Russell
+ *
+ * @since 2.4.1
+ */
+public class CompositeKafkaStreamsInfrastructireCustomizer implements KafkaStreamsInfrastructureCustomizer {
+
+	private final List<KafkaStreamsInfrastructureCustomizer> kafkaStreamsCustomizers = new ArrayList<>();
+
+	/**
+	 * Construct an instance with the provided customizers.
+	 * @param customizers the customizers;
+	 */
+	public CompositeKafkaStreamsInfrastructireCustomizer(KafkaStreamsInfrastructureCustomizer... customizers) {
+
+		this.kafkaStreamsCustomizers.addAll(Arrays.asList(customizers));
+	}
+
+	/**
+	 * Add customizers.
+	 * @param customizers the customizers.
+	 */
+	public void addKafkaStreamsCustomizers(KafkaStreamsInfrastructureCustomizer... customizers) {
+		this.kafkaStreamsCustomizers.addAll(Arrays.asList(customizers));
+	}
+
+	@Override
+	public void configureBuilder(StreamsBuilder builder) {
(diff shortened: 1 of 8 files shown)


spring-kafka/src/main/java/org/springframework/kafka/config/CompositeKafkaStreamsInfrastructireCustomizer.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Composite {@link KafkaStreamsCustomizer} customizes {@link KafkaStreams} by delegating
+ * to a list of provided {@link KafkaStreamsCustomizer}.
+ *
+ * @author Gary Russell
+ *
+ * @since 2.4.1
+ */
+public class CompositeKafkaStreamsInfrastructireCustomizer implements KafkaStreamsInfrastructureCustomizer {
```
artembilan: Typo: s/Infrastructire/Infrastructure

spring-kafka/src/main/java/org/springframework/kafka/config/CompositeKafkaStreamsInfrastructireCustomizer.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.kafka.streams.KafkaStreams;
+import org.apache.kafka.streams.StreamsBuilder;
+import org.apache.kafka.streams.Topology;
+
+/**
+ * Composite {@link KafkaStreamsCustomizer} customizes {@link KafkaStreams} by delegating
+ * to a list of provided {@link KafkaStreamsCustomizer}.
```
artembilan: Doesn't look like this JavaDoc reflects reality... 😄

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaStreamsInfrastructureCustomizer.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.config;
+
+import org.apache.kafka.streams.StreamsBuilder;
+import org.apache.kafka.streams.Topology;
+
+/**
+ * A customizer for infrastructure components such as the builder and topology.
```
artembilan: I think we need to be more specific: like talk about a `StreamsBuilder` class via link and therefore `Topology`.
Might be good to mention also a link to the `StreamsBuilderFactoryBean` where this one is used.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaStreamsInfrastructureCustomizer.java
```diff
@@ -0,0 +1,47 @@
...
+	 */
+	default void configureBuilder(StreamsBuilder builder) {
+		// no-op
+	}
+
+	/**
+	 * Configure the topology.
+	 * @param topology the topology
+	 */
+	default void configureTopology(Topology topology) {
```
artembilan: I wonder if with this new API we ever need a separate `KafkaStreamsCustomizer` strategy.
Couldn't that be covered in this one as a separate `configureKafkaStreams(KafkaStreams)` callback?

Might be a story for other version, although we could deprecate now and suggest to move to a new API...

spring-kafka/src/main/java/org/springframework/kafka/config/CompositeKafkaStreamsInfrastructireCustomizer.java
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright 2018-2019 the original author or authors.
```
artembilan: This is definitely a new class, so only the current year.

src/reference/asciidoc/streams.adoc
```diff
@@ -121,6 +122,25 @@ private StreamsBuilderFactoryBean myKStreamBuilderFactoryBean;
...
+
+	void configureTopology(Topology topology);
+
+}
+----
+====
+
+Default no-op implementations are provided to avoid having to implement both methods if one is not required.
+
+A `CompositeKafkaStreamsInfrastructureCustomizer` if you need multiple customizers.
```
artembilan: I know I'm not good in English, but feels like this sentence is missing a predicate.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaStreamsInfrastructureCustomizer.java
garyrussell: Yeah - I thought about deprecating that, but I couldn't come up with a good name for a new all-encompassing interface and I don't want to break the `@FunctionalInterface` there in a point release and I don't want to force people who only want to customize the infrastructure to have to implement a `configureStream` method.

We can think about it for a future release.

artembilan: Yeah... I know what you mean.
How about otherwise then: pursue fine-graining?
Since we already have a `KafkaStreamsCustomizer` and `StreamsBuilderFactoryBeanCustomizer` it won't hurt to have to more `@FunctionalInterface`s instead of this not consistent strategy?

garyrussell: Too many customizers, I think (we already have too many; I'd prefer just two, but can't do this in a point release). We can reconsider in 2.5 (or 3.0), where I would like to do a complete overhaul of the streams support.

artembilan: Sure! Let's postpone such a discussion!

Issue #1358, opened 2020-01-21
Improving KafkaTestUtil getRecords method functionality


psawmora:
**Affects Version(s):** 2.2.0.RELEASE

- Scenario : Unit testing a domain logic that writes result to couple of Kafka topics.

- Difficulty faced : As a part of the verification stage, I was trying to read records from couple of topics using KafkaTestUtils.getRecords(:consumer, :timeoutInMilliseconds). The consumer has subscribed to couple of topics. The difficulty I faced was that the getRecords method internally uses consumer.poll which returns immediately event if it finds a single record. 
The timeout is essentially there to wait for at least one record. Since the success flow of my test case expects N number of records, it won't always succeed. 

- Proposed solution : Introducing an overloaded version of the getRecords method which accepts the expected minimum number of records as a parameter. The signature would be as follows.

`public static <K, V> List<ConsumerRecord<K,V>> getRecords(Consumer<K, V> consumer, long timeout, int expectedMinRecordCount)`

- The logic would be as follows.  
   Before returning, collect records untill either the timeout is up or the expectedMinRecordCount is 
   reached.

garyrussell:
In the meantime, you can set `ConsumerConfig,FETCH_MIN_BYTES_CONFIG` and `ConsumerConfig.FETCH_MAX_WAIT_MS_CONFIG` to delay the return until more records are received. Unfortunately Kafka has no `fetch.min.records`.

PR #1360, opened 2020-01-21
GH-1358: KTU.getRecords() - add minRecords option

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1358


garyrussell:
Hi; no; the early exit is specifically to avoid populating the map and building another unnecessary `ConsumerRecords<>`.

PR #1362, opened 2020-01-22
GH-1362: Add the possibility of adding custom HandlerMethodArgumentResolver

filip-halemba:
@garyrussell as you wrote on Gitter it's a little bit tricky to add custom argument resolvers, because the order is important. 

But in my opinion we can add custom resolvers before payload resolver - what do you think about it?

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -78,10 +78,7 @@ import org.springframework.messaging.Message;
 import org.springframework.messaging.converter.GenericMessageConverter;
 import org.springframework.messaging.converter.MessageConverter;
 import org.springframework.messaging.handler.annotation.support.DefaultMessageHandlerMethodFactory;
-import org.springframework.messaging.handler.annotation.support.HeaderMethodArgumentResolver;
-import org.springframework.messaging.handler.annotation.support.HeadersMethodArgumentResolver;
 import org.springframework.messaging.handler.annotation.support.MessageHandlerMethodFactory;
-import org.springframework.messaging.handler.annotation.support.MessageMethodArgumentResolver;
 import org.springframework.messaging.handler.annotation.support.PayloadMethodArgumentResolver;
 import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
 import org.springframework.messaging.handler.invocation.InvocableHandlerMethod;
@@ -807,38 +804,19 @@ public class KafkaListenerAnnotationBeanPostProcessor<K, V>
 				defaultFactory.setValidator(validator);
 			}
 			defaultFactory.setBeanFactory(KafkaListenerAnnotationBeanPostProcessor.this.beanFactory);
-
-			ConfigurableBeanFactory cbf =
-					KafkaListenerAnnotationBeanPostProcessor.this.beanFactory instanceof ConfigurableBeanFactory ?
-							(ConfigurableBeanFactory) KafkaListenerAnnotationBeanPostProcessor.this.beanFactory :
-							null;
-
-
 			this.defaultFormattingConversionService.addConverter(
 					new BytesToStringConverter(KafkaListenerAnnotationBeanPostProcessor.this.charset));
-
 			defaultFactory.setConversionService(this.defaultFormattingConversionService);
+			GenericMessageConverter messageConverter = new GenericMessageConverter(this.defaultFormattingConversionService);
+			defaultFactory.setMessageConverter(messageConverter);
 
-			List<HandlerMethodArgumentResolver> argumentResolvers = new ArrayList<>();
-
-			// Annotation-based argument resolution
-			argumentResolvers.add(new HeaderMethodArgumentResolver(this.defaultFormattingConversionService, cbf));
-			argumentResolvers.add(new HeadersMethodArgumentResolver());
-
-			// Type-based argument resolution
-			final GenericMessageConverter messageConverter =
-					new GenericMessageConverter(this.defaultFormattingConversionService);
-			argumentResolvers.add(new MessageMethodArgumentResolver(messageConverter));
-
-			// Custom argument resolution
-			argumentResolvers.addAll(KafkaListenerAnnotationBeanPostProcessor.this.registrar.getCustomMethodArgumentResolvers());
-
-			// Has to be at the end - if not it will try to resolver objects from CustomMethodArgumentResolvers
-			argumentResolvers.add(new KafkaNullAwarePayloadArgumentResolver(messageConverter, validator));
-
-			defaultFactory.setArgumentResolvers(argumentResolvers);
+			List<HandlerMethodArgumentResolver> customArgumentsResolver =
+					new ArrayList<>(KafkaListenerAnnotationBeanPostProcessor.this.registrar.getCustomMethodArgumentResolvers());
+			customArgumentsResolver.add(new KafkaNullAwarePayloadArgumentResolver(messageConverter, validator));
+			defaultFactory.setCustomArgumentResolvers(customArgumentsResolver);
 
 			defaultFactory.afterPropertiesSet();
+
 			return defaultFactory;
 		}
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -833,7 +833,7 @@ public class KafkaListenerAnnotationBeanPostProcessor<K, V>
 			// Custom argument resolution
 			argumentResolvers.addAll(KafkaListenerAnnotationBeanPostProcessor.this.registrar.getCustomMethodArgumentResolvers());
 
-			// it has to be at the end
+			// Has to be at the end - if not it will try to resolver objects from CustomMethodArgumentResolvers
 			argumentResolvers.add(new KafkaNullAwarePayloadArgumentResolver(messageConverter, validator));
 
 			defaultFactory.setArgumentResolvers(argumentResolvers);
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 the original author or authors.
+ * Copyright 2014-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -116,6 +116,7 @@ import org.springframework.validation.Validator;
  * @author Dariusz Szablinski
  * @author Venil Noronha
  * @author Dimitri Penner
+ * @author Filip Halemba
  *
  * @see KafkaListener
  * @see KafkaListenerErrorHandler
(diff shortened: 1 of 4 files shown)


garyrussell:
@filip-halemba Thanks for the contribution!

As I mentioned in Gitter, I am concerned about the order in which the resolvers are applied.

I think we need a more general solution; something like...

```java
public void setResolverCustomizer(Function<List<HandlerMethodArgumentResolver>>, <List<HandlerMethodArgumentResolver>> customizerFunction);
```

...where the `Function` gets an unmodifiable list of the current (standard) resolvers and returns a new list of resolvers.

filip-halemba:
@garyrussell thanks for your opinion. I will try to to prepare more general solution.

filip-halemba:
@garyrussell to be clear - in your opinion the method proposed by you should be in `KafkaListenerEndpointRegistrar` or `KafkaListenerAnnotationBeanPostProcessor`?

garyrussell:
Sorry, I didn't look closely at your original PR; I believe it should be in the BPP, not the registrar.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -828,7 +829,13 @@ private MessageHandlerMethodFactory createDefaultMessageHandlerMethodFactory() {
...
 					new GenericMessageConverter(this.defaultFormattingConversionService);
 			argumentResolvers.add(new MessageMethodArgumentResolver(messageConverter));
+
+			// Custom argument resolution
+			argumentResolvers.addAll(KafkaListenerAnnotationBeanPostProcessor.this.registrar.getCustomMethodArgumentResolvers());
+
+			// Has to be at the end - if not it will try to resolver objects from CustomMethodArgumentResolvers
 			argumentResolvers.add(new KafkaNullAwarePayloadArgumentResolver(messageConverter, validator));
+
 			defaultFactory.setArgumentResolvers(argumentResolvers);
```
artembilan: I think this one has to be changed to `setCustomArgumentResolvers()`.
We have similar logic in Spring Integration and the proper order is applied:https://github.com/spring-projects/spring-integration/blob/master/spring-integration-core/src/main/java/org/springframework/integration/handler/support/MessagingMethodInvokerHelper.java#L541

filip-halemba: @artembilan thank you for your pro tip.

filip-halemba:
@garyrussell after applying pro tip from comment of @artembilan I think that method signature:
 ```java
public void setResolverCustomizer(Function<List<HandlerMethodArgumentResolver>>, <List<HandlerMethodArgumentResolver>> customizerFunction);
 ```
is unnecessary because order of method argument resolvers is now restricted in this [class](https://github.com/spring-projects/spring-framework/blob/77517d6cff2b45c4bf1b526b4aa57fabe8e95626/spring-messaging/src/main/java/org/springframework/messaging/handler/annotation/support/DefaultMessageHandlerMethodFactory.java#L159).

Also I have removed some redundant code from `KafkaHandlerMethodFactoryAdapter` which resides in this [method](https://github.com/spring-projects/spring-framework/blob/77517d6cff2b45c4bf1b526b4aa57fabe8e95626/spring-messaging/src/main/java/org/springframework/messaging/handler/annotation/support/DefaultMessageHandlerMethodFactory.java#L159)

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistrar.java
```diff
@@ -74,6 +79,25 @@ public KafkaListenerEndpointRegistry getEndpointRegistry() {
 		return this.endpointRegistry;
 	}
 
+	/**
+	 * Return the list of {@link HandlerMethodArgumentResolver}.
+	 * @return the list of {@link HandlerMethodArgumentResolver}.
+	 * @since 2.5
+	 */
+	public List<HandlerMethodArgumentResolver> getCustomMethodArgumentResolvers() {
+		return this.customMethodArgumentResolvers;
```
artembilan: The good practice is to remove a `Collections.unmodifiableList()`

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistrar.java
```diff
@@ -74,6 +79,25 @@ public KafkaListenerEndpointRegistry getEndpointRegistry() {
...
+	}
+
+	/**
+	 * Add custom methods arguments resolvers to {@link KafkaListenerAnnotationBeanPostProcessor}
+	 * Default empty list.
+	 * @param methodArgumentResolvers the list of {@link HandlerMethodArgumentResolver} to use.
+	 * @since 2.5
```
artembilan: I think your change is fully non-braking, so it's OK to have this feature in the current `2.4.2` version.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistrar.java
```diff
@@ -74,6 +79,25 @@ public KafkaListenerEndpointRegistry getEndpointRegistry() {
...
+	 */
+	public void setCustomMethodArgumentResolvers(List<HandlerMethodArgumentResolver> methodArgumentResolvers) {
```
artembilan: Let's consider varargs instead of list!
Much easier from end-user perspective.

spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
```diff
@@ -2192,5 +2232,17 @@ public void setBar(int bar) {
 
 	}
 
+	static class CustomMethodArgument {
+
+		final String body;
+		final String topic;
```
artembilan: Every member of the class has to be surrounded with blank lines.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistrar.java
filip-halemba: I assume that you would like to write - The good practice is to **add** a Collections.unmodifiableList()

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistrar.java
filip-halemba: @artembilan to be honest I am not sure if this method should add elelments to existing list (maybe then I should change name to addCustomMethodArgumentResolvers) or assign to this filed completly new list. What do you think?

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistrar.java
artembilan: Doh, of course! :shame:

Right, *wrap* an original `List` into a `Collections.unmodifiableList()` in getters to avoid unexpected external mutation.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistrar.java
artembilan: Well, you are right: I'm not talking about *adding* - it is definitely has to be full replacement.
So, you get a varargs argument, clean up an existing `customMethodArgumentResolvers` state and add those new into it.

garyrussell:
Sorry, I am a bit behind in my review here.

1. Why is the setter on the registrar rather than the factory bean?
2. @artembilan how does setting the custom resolvers solve the ordering problem? What if I want to add a resolver that runs before one of the default resolvers?

Thanks.

artembilan:
> Why is the setter on the registrar rather than the factory bean?

Because that's where we already set a `MessageHandlerMethodFactory`. See docs: https://docs.spring.io/spring-kafka/docs/2.4.1.RELEASE/reference/html/#conversionservice-customization. We have that `KafkaListenerConfigurer` callback to let to configure features like this.

> What if I want to add a resolver that runs before one of the default resolvers?

So, set the whole `MessageHandlerMethodFactory` instance otherwise.
We need to be consistent: if we talk about a default one provided by the framework, be aware of the order for your custom argument resolvers. Otherwise you are on your own with full customization via `MessageHandlerMethodFactory`. That's how it works in Spring Framework for STOMP, same way in Spring Integration and seems for me in Spring Cloud Stream.

Am I missing anything?

Thanks

filip-halemba:
> 1. Why is the setter on the registrar rather than the factory bean?

@garyrussell In my opinion it is really convenient for the end user to simply implement `KafkaListenerConfigurer` and then just use this new method - example below:

```
class CustomKafkaConfig implements KafkaListenerConfigurer {

    @Override
    public void configureKafkaListeners(KafkaListenerEndpointRegistrar registrar) {
        registrar.setCustomMethodArgumentResolvers();
    }

}
```

Also I am not sure but access to KafkaListenerAnnotationBeanPostProcessor methods for end user is not so simple - Am I right?

garyrussell:
OK, both; all is good with me (I'll crawl back under my rock 😄).

garyrussell:
The only problem I see is a redundant `PayloadMethodArgumentResolver` after the `KafkaNullAwarePayloadArgumentResolver` but since it will never be called, I guess that's ok.

artembilan:
> The only problem I see is a redundant PayloadMethodArgumentResolver

Right, we don't need it: it is configured in the `DefaultMessageHandlerMethodFactory` as default one.

filip-halemba:
@artembilan As I understand the `PayloadMethodArgumentResolver` can be redundant?

If that's true my change is ready for review :)

garyrussell:
Yes, it will just sit there unused; unfortunate, but by no means a show-stopper.

src/reference/asciidoc/kafka.adoc
```diff
@@ -3161,6 +3161,26 @@ This lets you further customize listener deserialization without changing the de
 
 IMPORTANT: Setting a custom `MessageHandlerMethodFactory` on the `KafkaListenerEndpointRegistrar` through a `KafkaListenerConfigurer` bean disables this feature.
 
+===== Adding custom `HandlerMethodArgumentResolver` to `@KafkaListener`
+
+Starting with version 2.4.2 you are able to add your own `org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver` and resolve custom method parameters.
+All you need is to implement `org.springframework.kafka.annotation.KafkaListenerConfigurer` and use method `setCustomMethodArgumentResolvers()` from class `org.springframework.kafka.configKafkaListenerEndpointRegistrar`.
```
artembilan: Please, remove packages from class name. Really redundant.
And you see: you have missed one dot in the `KafkaListenerEndpointRegistrar` FQCN 😉

src/reference/asciidoc/kafka.adoc
```diff
@@ -3161,6 +3161,26 @@ This lets you further customize listener deserialization without changing the de
...
+
+====
+[source, java]
+----
+@Configuration
+class CustomKafkaConfig implements KafkaListenerConfigurer {
+
+    @Override
+    public void configureKafkaListeners(KafkaListenerEndpointRegistrar registrar) {
+        registrar.setCustomMethodArgumentResolvers();
```
artembilan: Would be better to show here some custom argument resolver.

src/reference/asciidoc/kafka.adoc
filip-halemba: Done.

src/reference/asciidoc/kafka.adoc
filip-halemba: Done.

src/reference/asciidoc/kafka.adoc
```diff
@@ -3174,7 +3174,23 @@ class CustomKafkaConfig implements KafkaListenerConfigurer {
 
     @Override
     public void configureKafkaListeners(KafkaListenerEndpointRegistrar registrar) {
-        registrar.setCustomMethodArgumentResolvers();
+        registrar.setCustomMethodArgumentResolvers(
+        		new HandlerMethodArgumentResolver() {
+
+						@Override
+						public boolean supportsParameter(MethodParameter parameter) {
```
artembilan: Tell, me, please, if you are sure that all of these indents are whitespaces, not tab.

That's a bit confusing , I know: we use tabs in the real code, but whitespaces in docs.
So, 4 spaces for indents in docs please! 😄

filip-halemba: Now I am sure ;)

artembilan:
@filip-halemba ,

thank you very much for contribution (and your patience with our review); looking forward for more!

filip-halemba:
Thank you also ;)

Issue #1370, opened 2020-01-29
On Partitions Assignment need of transaction-ability 


tomasAlabes:
The origin of this issue is this [SO question](https://stackoverflow.com/questions/59958682/on-partitions-assignment-and-chainedkafkatransactionmanager-at-startup-with-jpa/59959119), @garyrussell suggested creating this issue to implement a couple of options regarding this case.

## Question

I have many transactional consumers with a `ChainedKafkaTransactionManager` based on a `JpaTransactionManager` and a `KafkaTransactionManager` (all `@KafkaListener`'s).

The JPA one needs a ThreadLocal variable to be set, to be able to know to which DB to connect to (is the tenant id).

When starting the application, in the `onPartitionsAssigned` [listener](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2223), spring-kafka is trying to create a chained txn, hence trying to create a JPA txn, but there's no tenant set, then it fails.

That tenant is set through a http filter and/or kafka interceptors (through event headers).

I tried using the auto-wired `KafkaListenerEndpointRegistry` with `setAutoStartup(false)`, but I see that the consumers don't receive any events, probably because they aren't initialized yet (I thought they were initialized on-demand).

If I set a mock tenant id and call `registry.start()` when the application is ready, the initializations seem to be done in other threads (probably because I'm using a `ConcurrentKafkaListenerContainerFactory`), so it doesn't work.

Is there a way to avoid the JPA transaction on that initial `onPartitionsAssigned` listener, that is part of the consumer initialization?

## Suggestion

> If your chained TM has the KafkaTM first, followed by JPA TM (which would be the normal case), you can achieve similar functionality by just injecting the Kafka TM into the container and using @Transactional (with just the JPA TM on the listener) to start the JPA transaction when the listener is called.
> The time between the transaction commits will be marginally increased but it would provide similar functionality.

In parallel, this issue is created to provide the option/s to either 

* disable the initial commit on assignment, 
* or do it without a transaction at all

Thanks @garyrussell!

garyrussell:
[...]

PR #1371, opened 2020-01-30
GH-1370: Opt. commit current offsets on assignment

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1370

Add an option to control current offset commit behavior on partition assignment.

**cherry-pick to 2.3.x**

Container changes might be easier to review with `?w=1` after initial review.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -610,6 +642,21 @@ public void setSubBatchPerPartition(boolean subBatchPerPartition) {
 		this.subBatchPerPartition = subBatchPerPartition;
 	}
 
+	public AssignmentCommitOption getAssignmentCommitOption() {
+		return this.assignmentCommitOption;
+	}
+
+	/**
+	 * Set the assignment commit option. Default {@link AssignmentCommitOption#ALWAYS}.
+	 * In a future release it will default to {@link AssignmentCommitOption#LATEST_ONLY}.
```
artembilan: I think we need an issue for this do not forget to really change it in the "future release"

garyrussell:
I thought about adding docs but this is a pretty obscure use case and I think, maybe, it's better not to add information to the docs that most people don't care about.

But, I don't mind adding if you prefer.

artembilan:
@garyrussell ,

Doesn't cherry-pick clean.

Would you mind to take a look?

Thanks

garyrussell:
Cherry-picked to 2.3.x as e16b433bfbb9125db16e6fec040ee7c6c7721dde

Issue #1354, opened 2020-01-15
AfterRollbackProcessor and default FixedBackOff


Rajh:
https://github.com/spring-projects/spring-kafka/blob/f4e9728a27f7ef4988d739765f1aa0a16f92a262/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java#L74

It seems like when using "-1" as maxFailure "new FixedBackOff()" is created which has 5 seconds of delay.
But when using a maxFailure != -1 "new FixedBackOff(0L, maxFailures == 0 ? 0 : maxFailures - 1);" is used which has 0 seconds of delay.

garyrussell:
Please explain further; the folllowing logic handles the negative `maxAttempts` case:

https://github.com/spring-projects/spring-kafka/blob/f4e9728a27f7ef4988d739765f1aa0a16f92a262/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java#L75-L77

this sets the backOff's max attempts to `Long.MAX_VALUE`.

This test passes fine:

```java
public class FailedRecordProcessorTests {

	@Test
	void testInfiniteRetrys() {
		FailedRecordProcessor processor = new FailedRecordProcessor(null, -1) {
		};
		assertThat(KafkaTestUtils.getPropertyValue(processor, "failureTracker.backOff", FixedBackOff.class)
				.getMaxAttempts()).isEqualTo(Long.MAX_VALUE);
	}

}
```

garyrussell:
Please explain further; the folllowing logic handles the negative `maxAttempts` case:

https://github.com/spring-projects/spring-kafka/blob/f4e9728a27f7ef4988d739765f1aa0a16f92a262/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java#L75-L77

this sets the backOff's max attempts to `Long.MAX_VALUE`.

This test passes fine:

```java
public class FailedRecordProcessorTests {

	@Test
	void testInfiniteRetrys() {
		FailedRecordProcessor processor = new FailedRecordProcessor(null, -1) {
		};
		assertThat(KafkaTestUtils.getPropertyValue(processor, "failureTracker.backOff", FixedBackOff.class)
				.getMaxAttempts()).isEqualTo(Long.MAX_VALUE);
	}

}
```

It was fixed here: https://github.com/spring-projects/spring-kafka/pull/1276 in 2.3.2 - what version are you using?

[...]

PR #1373, opened 2020-02-03
GH-1354: Fix default BackOff with deprecated CTOR

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1354

When a negative `maxAttempts` was used, the default backOff interval was
5s instead of 0.

**cherry-pick to 2.3.x**


artembilan:
... and cherry-picked to `2.3.x`

Issue #1369, opened 2020-01-29
Spring kafka should handle some fatal exceptions like InvalidPidMappingException, UnknownProducerIdException.


bartosz-stasikowski-projectdrgn:
**Affects Version(s):** 2.3.1.RELEASE

🎁 Enhancement
Spring kafka should handle some fatal exceptions like InvalidPidMappingException, UnknownProducerIdException.

Context: processing messages with transactions in non batch mode
Problem: Some exceptions like InvalidPidMappingException, UnknownProducerIdException are fatal and will cause KafkaMessageListenerContainer to fail every time polled record is processed

Usecase:
1. Records are polled
2. Transaction fails because of InvalidPidMappingException or UnknownProducerIdException when processing record
3. Rollback processing fails for InvalidPidMappingException or will send message to DLT for UnknownProducerIdException (if configured to do so)
4. Next poll will get another record which will have the same problem because exception is fatal 
 - in case of InvalidPidMappingException records will be processed after service is restarted or the is a rebalance
 - in case of UnknownProducerIdException records wont be processed again since offsets were commited in rollback processing

Currently InvalidPidMappingException exception can be handled in ErrorHandler but the only thing I can think of is to restart listener container.
For UnknownProducerIdException it is possible to not perform rollback processing and handle in the same way as InvalidPidMappingException in error handler.
This is strange and it should be handled by spring-kafka since it is a common problem when using transactions.

garyrussell:
Please show some logs and stack traces - any failures related to transactions are considered fatal (`txFailed` contains the exception) from the producer perspective and the producer is closed...

https://github.com/spring-projects/spring-kafka/blob/909af16384e372e3699198c275db3827328a0bff/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java#L637-L660

and

https://github.com/spring-projects/spring-kafka/blob/909af16384e372e3699198c275db3827328a0bff/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java#L681-L689

bartosz-stasikowski-projectdrgn:
Yes it works exactly as you want, but producer is not closed because its cache is null.

```
@Override
		public void close(@Nullable Duration timeout) {
			LOGGER.trace(() -> toString() + " close(" + (timeout == null ? "null" : timeout) + ")");
			if (this.cache != null) {
```

[...]

PR #1378, opened 2020-02-04
GH-1369: Fix Fenced Consumer-based Producers

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1369

1. `CloseSafeProducer.cache` was incorrectly null
   (direct access to `this.cache` instead of `getCache()`
   Prevented closure of a fenced `consumerProducer`

2. Don't call the `AfterRollbackProcessor` when a producer is fenced -
   the partitions will have already been revoked so seeking is inappropriate
   and starting a new transaction will fence the new (current) producer.

**cherry-pick to 2.3.x if clean**


artembilan:
... and cherry-picked to `2.3.x`

PR #1380, opened 2020-02-08
Make ZK client timeouts configurable

pawellozinski:
For build servers with high load, it may happen that the build hangs for couple of seconds. This causes the internal ZK client session of embedded Kafka broker to timeout and fail the build.

It would be good if these were configurable.

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
@@ -258,6 +258,7 @@ public class EmbeddedKafkaBroker implements InitializingBean, DisposableBean {
 	 * Set connection timeout for the client to the embedded Zookeeper.
 	 * @param zkConnectionTimeout the connection timeout,
 	 * @return the {@link EmbeddedKafkaBroker}.
+	 * @since 2.4
 	 */
 	public EmbeddedKafkaBroker zkConnectionTimeout(int zkConnectionTimeout) {
 		this.zkConnectionTimeout = zkConnectionTimeout;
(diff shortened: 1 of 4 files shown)


pawellozinski:
Hi @garyrussell , sure, happy to take this forward. Didn't you mean `@EmbeddedKafka`?

garyrussell:
Sorry. Yes @EmbeddedKafka.

spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
```diff
@@ -249,6 +253,18 @@ public void setZkPort(int zkPort) {
 		this.zkPort = zkPort;
 	}
 
+	/**
+	 * Set timeouts for the client to the embedded Zookeeper.
+	 * @param zkConnectionTimeout the connection timeout,
+	 * @param zkSessionTimeout the session timeout.
+	 * @return the {@link EmbeddedKafkaBroker}.
+	 */
+	public EmbeddedKafkaBroker setZkClientTimeouts(int zkConnectionTimeout, int zkSessionTimeout) {
```
artembilan: This is not good.
All the setters has to follow Java Beans convention.
So, please, make separate setters for these options.

Or if we talk about builder API, then it cannot be with a `set` prefix.
Anyway I'd prefer to have these properties to be configured separately.

garyrussell:
And cherry-picked as abf2bfe006b2d18a00199e6754a3d5a2ff98bc93 to 2.3.x with deprecated `ZkClient` fix.

Thanks again @pawellozinski

pawellozinski:
:thumbsup:

PR #1395, opened 2020-02-20
Add property resolver to Kafka listener post processor

qwe333ty:
Kafka listener bean post processor decomposition: relocate property resolving methods to a suitable class 'KafkaEndpointPropertyResolver'.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -42,7 +42,6 @@ import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.beans.factory.ObjectFactory;
 import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.beans.factory.config.BeanExpressionContext;
-import org.springframework.beans.factory.config.BeanExpressionResolver;
 import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.config.Scope;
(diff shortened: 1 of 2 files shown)


artembilan:
@qwe333ty ,

thank you for contribution, but what did drive you to make this change?
Semantically you don't provide any new functionality - just a simple methods extraction to a separate not fully useful top-level class.

Thanks

qwe333ty:
@artembilan , @garyrussell , thank you for reviewing!

qwe333ty:
@artembilan , The answer is simple. I want Spring Kafka to follow the Single Responsibility Principle that states that classes should have responsibility over a single part of the functionality and entirely encapsulate it. Also states that there is always must be one reason for changing the implementation. In our case with Kafka listener post-processor there are 2 possible reasons: change the way of message listener container creation & change the way of resolving property. I think it should be fixed by separating post-processor class into independent parts. That's why I have created a merge request that contains one new class with all visible and useful methods to resolve properties using only @KafkaListener annotation.

qwe333ty:
@garyrussell , Yeeees, this would be a great addition to post-processing! I think I should try to implement an extension point to add a custom resolver and give it to review.

artembilan:
OK. Thanks for the explanation!

Perhaps we should start from the use-case: why would one need a custom properties resolver against the framework annotation?
I'm not against your "Single Responsibility Principle", but this post-processor was really intended for parsing that annotation to some set of beans to let end-user to avoid hard codding for the parser. 
May the final custom solution has to be fully away from the fact of the annotation. That's exactly why we provide a high-level API with a `KafkaListenerContainerFactory` abstraction.

For me your point feels like an abusing of the framework opinion regarding that annotation...

garyrussell:
Agreed.

>change the way of message listener container creation 

That's already achievable with the `RabbitListenerContainerFactory` abstraction.

>change the way of resolving property

We need to understand the exact use case.

qwe333ty:
@garyrussell , I have been thinking for a couple of days over the possible use cases but I still can't find anything useful :( I think we should close this pull request if someone do not have any ideas.

garyrussell:
Closing due to inactivity.

Thanks again for the desire to contribute but I think we agreed this is not needed.

Please don't let it stop you from making other contributions.

PR #1404, opened 2020-03-03
Use Spring Gradle Enterprise Conventions Plugin

garyrussell:


garyrussell:
@wilkinsona - it looks like the `publishOnFailure()` is not being picked up - we seem to be publishing successful builds now...

https://ge.spring.io/scans?failures.failureClassification=non_verification&list.offset=0&list.size=50&list.sortColumn=startTime&list.sortOrder=desc&search.buildToolType=gradle&search.buildToolType=maven&search.rootProjectName=spring-kafka-dist&search.startTimeMax=1583297999999&search.startTimeMin=1577854800000&tests.sortField=FAILED&tests.unstableOnly&trends.section=overview&trends.timeResolution=day&viewer.tzOffset=-300

settings.gradle
```diff
@@ -1,18 +1,20 @@
...
-		buildScan {
-			captureTaskInputFiles = true
-			obfuscation {
-				ipAddresses { addresses -> addresses.collect { address -> '0.0.0.0'} }
-			}
-			publishOnFailure()
-			server = System.getenv('GRADLE_ENTERPRISE_URL')
-		}
+gradleEnterprise {
+	apply from: "$rootDir/gradle/build-scan-user-data.gradle"
```
wilkinsona: You can remove this as the conventions plugin will take care of it.

settings.gradle
```diff
@@ -1,18 +1,20 @@
...
+	buildScan {
+		captureTaskInputFiles = true
```
wilkinsona: You can remove this as the conventions plugin will take care of it.

garyrussell: Thanks; my intention was to get your review before it was merged; sorry, @artembilan I should have made that clear.

wilkinsona:
Gah. I’d hoped that your `publishOnFailure()` would take precedence over the conventions plugin configuring scans to be published always (success or failure). Maybe we can tweak things a bit so that works, or we might need a different approach. I’ll dig a bit more tomorrow.

garyrussell:
OK; thanks for your help; get some sleep 😄 !!

wilkinsona:
The Gradle Enterprise plugin uses separate state to track `publishAlways()` and `publishOnFailure()`. As soon as the former has been called, the latter is unable to switch things back to only publishing on failure.

I can see a few ways forward:

1. Remove `publishAlways()` from the conventions plugin and require its users to call it or an alternative method
2. Change the conventionsPlugin to support a property configured in `gradle.properties` that can switch off `publishAlways`
3. Raise the problem with the Gradle team and see if the Gradle Enterprise plugin can be changed
4. Always publish build scans for Spring Kafka

PR #1407, opened 2020-03-05
Fix Gradle Deprecations

garyrussell:


artembilan:
I will merge this when you check the failure on Travis. Also the CI plan on Bamboo needs to disable Artifactory plugin since we use it directly from our build now. See SI master plan for difference.

Issue #1382, opened 2020-02-11
SeekToCurrentBatchErrorHandler infinite retry


stager0909-ray:
I use spring-kafka-test:2.3.4.RELEASE and spring-kafka:2.3.4.RELEASE. 

SeekToCurrentBatchErrorHandler and FixedBackOff are used to retry 2 times when an exception occurs in the consumer.
However, 20 retry attempts occurred during a 10 second sleep. Why doesn't FixedBackOff work?

```java
import static org.assertj.core.api.Assertions.assertThat;

import java.util.Map;

import org.apache.kafka.clients.consumer.ConsumerConfig;
import org.apache.kafka.clients.producer.ProducerConfig;
import org.apache.kafka.common.serialization.IntegerDeserializer;
import org.junit.Test;
import org.junit.runner.RunWith;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;
import org.springframework.kafka.annotation.EnableKafka;
import org.springframework.kafka.annotation.KafkaListener;
import org.springframework.kafka.config.ConcurrentKafkaListenerContainerFactory;
import org.springframework.kafka.config.KafkaListenerContainerFactory;
import org.springframework.kafka.core.ConsumerFactory;
import org.springframework.kafka.core.DefaultKafkaConsumerFactory;
import org.springframework.kafka.core.DefaultKafkaProducerFactory;
import org.springframework.kafka.core.KafkaTemplate;
import org.springframework.kafka.core.ProducerFactory;
import org.springframework.kafka.listener.SeekToCurrentBatchErrorHandler;
import org.springframework.kafka.support.KafkaHeaders;
import org.springframework.kafka.support.serializer.JsonDeserializer;
import org.springframework.kafka.support.serializer.JsonSerializer;
import org.springframework.kafka.test.EmbeddedKafkaBroker;
import org.springframework.kafka.test.context.EmbeddedKafka;
import org.springframework.kafka.test.utils.KafkaTestUtils;
import org.springframework.messaging.handler.annotation.Header;
import org.springframework.test.annotation.DirtiesContext;
import org.springframework.test.context.junit4.SpringRunner;
import org.springframework.util.backoff.FixedBackOff;

import lombok.extern.slf4j.Slf4j;

@Slf4j
@RunWith(SpringRunner.class)
[...]

PR #1410, opened 2020-03-06
GH-1382: RetryingBatchErrorHandler

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1382

**cherry-pick to 2.4.x, 2.3.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchErrorHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 the original author or authors.
+ * Copyright 2015-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,6 +15,8 @@
  */
 
 package org.springframework.kafka.listener;
+
+import java.util.function.Supplier;
 
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
@@ -39,6 +41,22 @@ public interface BatchErrorHandler extends GenericErrorHandler<ConsumerRecords<?
 	 */
 	default void handle(Exception thrownException, ConsumerRecords<?, ?> data, Consumer<?, ?> consumer,
 			MessageListenerContainer container) {
+		handle(thrownException, data);
+	}
+
+	/**
+	 * Handle the exception.
+	 * @param thrownException the exception.
+	 * @param data the consumer records.
+	 * @param consumer the consumer.
+	 * @param container the container.
+	 * @param invokeListener a callback to re-invoke the listener.
+	 * @throws InterruptedException if the thread is interrupted.
+	 * @since 2.3.7
+	 */
+	default void handle(Exception thrownException, ConsumerRecords<?, ?> data,
+			Consumer<?, ?> consumer, MessageListenerContainer container, Supplier<Boolean> invokeListener) {
+
 		handle(thrownException, data);
 	}
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchErrorHandler.java
@@ -16,8 +16,6 @@
 
 package org.springframework.kafka.listener;
 
-import java.util.function.Supplier;
-
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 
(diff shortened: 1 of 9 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/BatchErrorHandler.java
```diff
@@ -42,4 +44,20 @@ default void handle(Exception thrownException, ConsumerRecords<?, ?> data, Consu
...
+	 * @param thrownException the exception.
+	 * @param data the consumer records.
+	 * @param consumer the consumer.
+	 * @param container the container.
+	 * @param invokeListener a callback to re-invoke the listener.
+	 * @throws InterruptedException if the thread is interrupted.
+	 * @since 2.3.7
+	 */
+	default void handle(Exception thrownException, ConsumerRecords<?, ?> data,
+			Consumer<?, ?> consumer, MessageListenerContainer container, Supplier<Boolean> invokeListener) {
```
artembilan: Shouldn't this one be without a `container`?
Why do we need then that `ContainerAwareBatchErrorHandler` otherwise?

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerInvokingBatchErrorHandler.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+
+/**
+ * A batch error handler that is capable of invoking the listener during error handling.
+ *
+ * @author Gary Russell
+ * @since 2.3.7
+ *
+ */
+public interface ListenerInvokingBatchErrorHandler extends ContainerAwareBatchErrorHandler {
```
artembilan: Why do we need this abstraction if we have the same callback on the `BatchErrorHandler` or `ContainerAwareBatchErrorHandler`?

spring-kafka/src/test/java/org/springframework/kafka/listener/RetryingBatchErrorHandlerIntegrationTests.java
```diff
@@ -0,0 +1,212 @@
+/*
+ * Copyright 2018-2019 the original author or authors.
```
artembilan: Wrong date

spring-kafka/src/test/java/org/springframework/kafka/listener/RetryingBatchErrorHandlerIntegrationTests.java
```diff
@@ -0,0 +1,212 @@
...
+import org.springframework.kafka.event.ConsumerStoppedEvent;
+import org.springframework.kafka.test.EmbeddedKafkaBroker;
+import org.springframework.kafka.test.condition.EmbeddedKafkaCondition;
+import org.springframework.kafka.test.context.EmbeddedKafka;
+import org.springframework.kafka.test.utils.KafkaTestUtils;
+import org.springframework.util.backoff.FixedBackOff;
+
+/**
+ * @author Gary Russell
+ * @since 2.2
```
artembilan: Wrong `@since`

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1516,10 +1523,16 @@ private void doInvokeBatchOnMessage(final ConsumerRecords<K, V> records,
...
+		private void invokeBatchErrorHandler(final ConsumerRecords<K, V> records,
+				@Nullable List<ConsumerRecord<K, V>> list, RuntimeException e) {
+
 			if (this.batchErrorHandler instanceof ContainerAwareBatchErrorHandler) {
 				this.batchErrorHandler.handle(decorateException(e), records, this.consumer,
-						KafkaMessageListenerContainer.this.thisOrParentContainer);
+						KafkaMessageListenerContainer.this.thisOrParentContainer,
+						() -> {
+							invokeBatchOnMessage(records, list);
+							return true;
```
artembilan: Why do we need this contract as a `Consumer<Boolean>` since we always return `true`?
Won't plain `Runnable` callback be enough?

spring-kafka/src/main/java/org/springframework/kafka/listener/BatchErrorHandler.java
garyrussell: Notice how each sub interface provides a default implementation of the super; this avoids implementations from having to implement other methods.

Adding the method here avoids `instanceof` checks in the container; in fact, I now realize I can get rid of the one `instanceof` check that is there now.

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerInvokingBatchErrorHandler.java
garyrussell: So that implementations only have to implement this method.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Correct; will change.

artembilan:
... and cherry-picked to `2.4.x` & `2.3.x` after fixing JavaDocs issue.

PR #1411, opened 2020-03-09
DeadLetterPublishingRecoverer improvements

garyrussell:
- strip exception header from record published to DLT
- remove deprecated `ErrorHandlingDeserializer`
- rename `ErrorHandlingDeserializer2` to `ErrorHandlingDeserializer`
- add deprecated `ErrorHandlingDeserializer2`

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/ErrorHandlingDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/ErrorHandlingDeserializer.java
@@ -16,10 +16,15 @@
 
 package org.springframework.kafka.support.serializer;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectOutputStream;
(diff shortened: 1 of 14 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/serializer/ErrorHandlingDeserializer.java
```diff
@@ -55,44 +84,103 @@
...
 	}
 
 	public ErrorHandlingDeserializer(Deserializer<T> delegate) {
 		this.delegate = setupDelegate(delegate);
 	}
 
+	/**
+	 * Provide an alternative supplying mechanism when deserialization fails.
+	 * @param failedDeserializationFunction the {@link BiFunction} to use.
+	 * @deprecated since 2.2.8 in favor of {@link #setFailedDeserializationFunction(Function)}.
```
artembilan: May be we can remove this method already?

Issue #1366, opened 2020-01-27
Support Micrometer metrics for Kafka Producers


sparty02:
**Affects Version(s):** 2.4.x

🎁 Enhancement

Spring Kafka should support producer metrics, complementing the existing listener metrics implemented by PR #1242.  Some useful metrics (not necessarily exhaustive):

- successful send count
- failed send count
- timer for message send
- distribution timer for message send (min/max/median/mean/95%/etc) --- not sure what support micrometer has for this

garyrussell:
>distribution timer for message send (min/max/median/mean/95%/etc) --- not sure what support micrometer has for this

Micrometer delegates such analysis to the reporting tools (Prometheus etc).

PR #1412, opened 2020-03-09
GH-1366: Add Micrometer timers to KafkaTemplate

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1366

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -17,6 +17,7 @@
 package org.springframework.kafka.core;
 
 import java.time.Duration;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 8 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -489,6 +549,22 @@ else if (this.allowNonTransactional) {
...
+		MicrometerHolder holder = null;
+		try {
+			if (KafkaUtils.MICROMETER_PRESENT) {
+				holder = new MicrometerHolder(this.applicationContext, this.beanName,
+						"spring.kafka.template", "KafkaTemplate Timer",
+						this.micrometerTags);
+			}
+		}
+		catch (@SuppressWarnings("unused") IllegalStateException ex) {
+			this.micrometerEnabled = false;
```
artembilan: Having this, it sounds like `micrometerEnabled` has to be `volatile`.
No?

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaUtils.java
```diff
@@ -33,6 +35,12 @@
  */
 public final class KafkaUtils {
 
+	/**
+	 * True if micrometer is on the class path.
+	 */
+	public static final boolean MICROMETER_PRESENT = ClassUtils.isPresent(
+			"io.micrometer.core.instrument.MeterRegistry", KafkaMessageListenerContainer.class.getClassLoader());
```
artembilan: I think this is copy/paste artifact.
Or `KafkaUtils` or `ClassUtils.getDefaultClassLoader()`

PR #1420, opened 2020-03-20
Support `flush()` with send(Message<?>)

garyrussell:
- add a header that will tell the template to flush after sending.


artembilan:
I wonder why do we need such a logic in the framework with headers overhead, if you still need to make a decision which message should have that header, while there is just enough to call that `template.flush()` in that target logic whenever it is necessary instead...

garyrussell:
God point, I originally was just going to put the code in SIK where the template is not exposed.

We will need similar code there - especially when using a gateway with a producer that has `linger.ms` - the send will be delayed without a flush.

Originally I was just going to put the code there but thought it made more sense to put the header in `KafkaHeaders`. But I will revert to that approach.

PR #1427, opened 2020-03-26
GH-1370: Container props: Polish, Defaults, Docs

garyrussell:
See GH-1370

- deprecate `ContainerProperties.getTopicPartitionsToAssign()`
- defaults for `ackCount`, `ackTime`
- `subBatchPerPartition` default `true` for transactions
- `assignmentCommitOption` default to `LATEST_ONLY_NO_TX`
- deprecate `ackOnError`

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
@@ -1159,6 +1159,7 @@ public class EnableKafkaIntegrationTests {
 			return factory;
 		}
 
+		@SuppressWarnings("deprecation")
 		@Bean
 		public KafkaListenerContainerFactory<ConcurrentMessageListenerContainer<Integer, String>>
 				recordAckListenerContainerFactory() {
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
@@ -612,7 +612,6 @@ public class ConcurrentMessageListenerContainerTests {
 		});
 		containerProps.setSyncCommits(true);
 		containerProps.setAckMode(ContainerProperties.AckMode.RECORD);
-		containerProps.setAckOnError(false);
 		ConcurrentMessageListenerContainer<Integer, String> container = new ConcurrentMessageListenerContainer<>(cf,
 				containerProps);
 		container.setConcurrency(2);
@@ -682,6 +681,7 @@ public class ConcurrentMessageListenerContainerTests {
 		testAckOnErrorWithManualImmediateGuts(topic11, false);
 	}
 
+	@SuppressWarnings("deprecation")
 	private void testAckOnErrorWithManualImmediateGuts(String topic, boolean ackOnError) throws Exception {
 		logger.info("Start ack on error with ManualImmediate ack mode");
 		Map<String, Object> props = KafkaTestUtils.consumerProps("testMan" + ackOnError, "false", embeddedKafka);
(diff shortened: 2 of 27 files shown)


spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
```diff
@@ -1159,6 +1159,7 @@ public RecordPassAllFilter manualFilter() {
 			return factory;
 		}
 
+		@SuppressWarnings("deprecation")
```
artembilan: Isn't there a way to avoid this with existing/new behavior?

I guess we will come back to the problem in this test when we are going to remove that `setAckOnError` eventually...

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
```diff
@@ -682,6 +681,7 @@ public void testAckOnErrorManualImmediate() throws  Exception {
 		testAckOnErrorWithManualImmediateGuts(topic11, false);
 	}
 
+	@SuppressWarnings("deprecation")
```
artembilan: Same here I think...

spring-kafka/src/test/java/org/springframework/kafka/listener/SubBatchPerPartitionTxTests.java
```diff
@@ -201,7 +201,8 @@ public ConcurrentKafkaListenerContainerFactory kafkaListenerContainerFactory() {
 			factory.getContainerProperties().setAckMode(AckMode.BATCH);
 			factory.getContainerProperties().setTransactionManager(tm());
 			factory.setBatchListener(true);
-			factory.getContainerProperties().setSubBatchPerPartition(true);
+			// true by default for a transactional container since 2.5
+//			factory.getContainerProperties().setSubBatchPerPartition(true);
```
artembilan: Probably we don't need this code neither comment about `2.5`

src/reference/asciidoc/kafka.adoc
```diff
@@ -1889,6 +1889,272 @@ However, see the note at the beginning of this section; you can avoid using the
...
+|ackTime
+|5000
+|The time in milliseconds after which pending offsets are committed when the `ackMode` is `TIME` or `COUNT_TIME`.
+
+|assignment
+CommitOption
+|LATEST_ONLY _NO_TX
+|Whether or not to commit the initial position on assignment; by default, the initial offset will only be committed if the `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG` is `latest` and it won't run in a transaction even if there is a transaction manager present.
+See the javadocs for `ContainerProperties.AssignmentCommitOption` for more information about the available options.
+In 2.3 and 2.4, the default was `ALWAYS` to retain the previous default before it was configurable.
```
artembilan: Why should we mention an old behavior in this new docs?

spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
garyrussell: Right - I didn't want to remove the test given that the feature is deprecated only; this test will be changed when we remove it.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
garyrussell: Same reply 😄

spring-kafka/src/test/java/org/springframework/kafka/listener/SubBatchPerPartitionTxTests.java
garyrussell: This was more for your benefit during review; will remove.

src/reference/asciidoc/kafka.adoc
garyrussell: I was trying to make up for the fact that I failed to document the property in 2.3 😄

PR #1432, opened 2020-03-30
Add RoutingKafkaTemplate

garyrussell:
- select the producer factory using the topic name

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/RoutingKafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/RoutingKafkaTemplate.java
@@ -86,8 +86,8 @@ public class RoutingKafkaTemplate extends KafkaTemplate<Object, Object> {
 					break;
 				}
 			}
+			Assert.state(producerFactory != null, "No producer factory found for topic: " + topic);
 		}
-		Assert.state(producerFactory != null, "No producer factory found for topic: " + topic);
 		return producerFactory;
 	}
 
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/RoutingKafkaTemplate.java
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	@Override
+	public ProducerFactory<Object, Object> getProducerFactory(String topic) {
+		ProducerFactory<Object, Object> producerFactory = this.factoryMap.get(topic);
+		if (producerFactory == null) {
+			for (Entry<Pattern, ProducerFactory<Object, Object>> entry : this.factoryMatchers.entrySet()) {
+				if (entry.getKey().matcher(topic).matches()) {
+					producerFactory = entry.getValue();
+					this.factoryMap.put(topic, producerFactory);
```
artembilan: Why can't we use a `computeIfAbsent()` instead?

spring-kafka/src/main/java/org/springframework/kafka/core/RoutingKafkaTemplate.java
```diff
@@ -81,8 +81,7 @@ public RoutingKafkaTemplate(Map<Pattern, ProducerFactory<Object, Object>> factor
 		if (producerFactory == null) {
 			for (Entry<Pattern, ProducerFactory<Object, Object>> entry : this.factoryMatchers.entrySet()) {
 				if (entry.getKey().matcher(topic).matches()) {
-					producerFactory = entry.getValue();
-					this.factoryMap.put(topic, producerFactory);
+					producerFactory = this.factoryMap.computeIfAbsent(topic, key -> entry.getValue());
```
artembilan: No, I mean the whole `for ()` logic and pattern matching should go to that lambda instead of all of these `if`, `put` and `break`. The lambda would have a return for matching value.

Issue #1435, opened 2020-04-03
Add BatchListenerFailedException


garyrussell:
A user-thrown exception to communicate to the error handler which record in the batch failed.

This will allow the error handler to commit the successful offsets, and re-seek/recover the failed record.

PR #1436, opened 2020-04-06
GH-1435: Add RecoveringBatchErrorHandler

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1435

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchListenerFailedException.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchListenerFailedException.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.listener;
+
+import org.springframework.kafka.KafkaException;
+import org.springframework.lang.Nullable;
+
+/**
+ * An exception thrown by user code to inform the framework which record in a batch has
+ * failed.
+ *
+ * @author Gary Russell
+ * @since 2.5
+ *
+ */
+public class BatchListenerFailedException extends KafkaException {
+
+	private static final long serialVersionUID = 1L;
+
+	private final int index;
+
+	/**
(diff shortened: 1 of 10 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/BatchListenerFailedException.java
```diff
@@ -0,0 +1,69 @@
...
+	private static final long serialVersionUID = 1L;
+
+	private final int index;
+
+	/**
+	 * Construct an instance with the provided properties.
+	 * @param message the message.
+	 * @param index the index in the batch of the failed record.
+	 */
+	public BatchListenerFailedException(String message, int index) {
```
artembilan, line 41: I wonder why it can't be a `ConsumerRecord` per se? 
Why should I worry about record selection in the batch by myself?

spring-kafka/src/main/java/org/springframework/kafka/listener/RecoveringBatchErrorHandler.java
```diff
@@ -0,0 +1,181 @@
...
+ * {@link BackOff}. If the record is recovered, its offset is committed.
+ *
+ * @author Gary Russell
+ * @since 2.5
+ *
+ */
+public class RecoveringBatchErrorHandler extends FailedRecordProcessor
+		implements ContainerAwareBatchErrorHandler {
+
+	private static final LogAccessor LOGGER = new LogAccessor(LogFactory.getLog(RecoveringBatchErrorHandler.class));
```
artembilan, line 59: Looks like there is already a `logger` property in the super class...

spring-kafka/src/main/java/org/springframework/kafka/support/SeekUtils.java
```diff
@@ -98,7 +123,61 @@ public static boolean doSeeks(List<ConsumerRecord<?, ?>> records, Consumer<?, ?>
...
+	 * @since 2.5
+	 */
+	public static void seekOrRecover(Exception thrownException, List<ConsumerRecord<?, ?>> records,
+			Consumer<?, ?> consumer, MessageListenerContainer container, boolean commitRecovered,
+			BiPredicate<ConsumerRecord<?, ?>, Exception> skipPredicate, LogAccessor logger) {
+
+		if (ObjectUtils.isEmpty(records)) {
+			if (thrownException instanceof SerializationException) {
+				throw new IllegalStateException("This error handler cannot process 'SerializationException's directly; "
+						+ "please consider configuring an 'ErrorHandlingDeserializer2' in the value and/or key "
```
artembilan, line 146: I think this `2` has to be fixed already: that class has been deprecated.

spring-kafka/src/main/java/org/springframework/kafka/listener/BatchListenerFailedException.java
garyrussell: Good idea

garyrussell: However, it's no good when using a POJO listener - I can come up with a hybrid.

Issue #1446, opened 2020-04-13
Support Static Membership


garyrussell:
https://kafka.apache.org/documentation/#static_membership

PR #1447, opened 2020-04-13
GH-1446: Support Static Group Membership

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1446

See https://kafka.apache.org/documentation/#static_membership

Add a qualifier to `group.instance.id` for each child container.

Catch `FencedInstanceIdException` and stop the container.

Fix delay on self-stopping containers.

Treat `FencedInstanceIdException` the same as `ProducerFencedException` when
using transactions (do not call `AfterRollbackProcessor`.

**cherry-pick to 2.4.x, 2.3.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -415,7 +415,6 @@ public abstract class AbstractMessageListenerContainer<K, V>
 					doStop(latch::countDown);
 					try {
 						latch.await(this.containerProperties.getShutdownTimeout(), TimeUnit.MILLISECONDS); // NOSONAR
-						publishContainerStoppedEvent();
 					}
 					catch (@SuppressWarnings("unused") InterruptedException e) {
 						Thread.currentThread().interrupt();
@@ -424,6 +423,7 @@ public abstract class AbstractMessageListenerContainer<K, V>
 				else {
 					doStop(() -> { });
 				}
+				publishContainerStoppedEvent();
 			}
 		}
 	}
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -415,10 +415,10 @@ public abstract class AbstractMessageListenerContainer<K, V>
 					doStop(latch::countDown);
 					try {
 						latch.await(this.containerProperties.getShutdownTimeout(), TimeUnit.MILLISECONDS); // NOSONAR
+						publishContainerStoppedEvent();
 					}
 					catch (@SuppressWarnings("unused") InterruptedException e) {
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -399,16 +399,30 @@ public void checkGroupId() {
...
+						publishContainerStoppedEvent();
+					}
+					catch (@SuppressWarnings("unused") InterruptedException e) {
+						Thread.currentThread().interrupt();
+					}
 				}
-				catch (@SuppressWarnings("unused") InterruptedException e) {
-					Thread.currentThread().interrupt();
+				else {
+					doStop(() -> { });
```
artembilan, line 425: When `publishContainerStoppedEvent();` is emitted in this case, when we don't wait for container to be stopped?

On the other hand what is the point of such an event if we still wait for container to be stopped?..

garyrussell: Sorry; I don't understand; the no-wait stop is when the container stops itself; previously it deadlocked for the shutdown timeout.

What does this have to do with event publishing?

artembilan: OK. So, previously we always waited for the stop and then emitted an event.
Right now we have a `wait` condition and emit an event only when it is `true`.
In case of `false` we have only this non-blocking `doStop()`.
I don't mind about "container stops itself" logic, but I try to understand why we don't emit an event when we don't wait for stop.

Hope this is much cleaner now.

garyrussell: Sorry; missed that; fixed.

artembilan: Docs are OK, but I can't merge without an explanation over here.

Thanks for understanding.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -424,6 +423,7 @@ public final void stop(boolean wait) {
 				else {
 					doStop(() -> { });
 				}
+				publishContainerStoppedEvent();
```
artembilan, line 426: Is it really stopped at this point?

Don't we really emit such an event from that `callback` we provide for the `doStop()` ?

artembilan:
@garyrussell ,

Does failure on the Travis say anything to you?

```
ConcurrentMessageListenerContainerTests > testAutoCommit() FAILED

    org.opentest4j.AssertionFailedError at ConcurrentMessageListenerContainerTests.java:193
```

artembilan:
@garyrussell ,

it doesn't cherry-pick properly.

Would you mind to back-port it yourself?

Thanks

garyrussell:
Sure

PR #1448, opened 2020-04-15
Fix assigned partitions with cooperative rebalance

garyrussell:
In 2.4 we added support for cooperative rebalancing by removing
specific revoked partitions in `onPartitionsRevoked`.
However, `onPartitionsAssigned` is also called (often with no
partitions), so we should not replace the exiting assignments,
but add the new assignments.

**cherry-pick to 2.4.x**


artembilan:
... and cherry-picked to `2.4.x`

PR #1454, opened 2020-04-17
Fix commits on partition revocation

garyrussell:
Complete the remainder of `onPartitionsRevoked()` if the commits fail
fatally.

It is possible that commits might fail when using manual acks; we must
complete the remainder of the method if the commits fail.

**I will backport as needed**


garyrussell:
Back ported to 2.4.x, 2.3.x, 2.2.x

PR #1460, opened 2020-04-21
Producer/Consumer Factory Listeners and Micrometer

garyrussell:
- Refactor `DKPF` so that `CloseSafeProducer` is not aware of whether it is cached
- Add producer and consumer factory listeners
- Add listener implementations for Micrometer

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,18 +18,28 @@ package org.springframework.kafka.core;
 
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Properties;
 import java.util.function.Supplier;
 
+import org.aopalliance.aop.Advice;
+import org.aopalliance.intercept.MethodInterceptor;
+import org.aopalliance.intercept.MethodInvocation;
 import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.serialization.Deserializer;
 
+import org.springframework.aop.framework.ProxyFactory;
+import org.springframework.aop.support.NameMatchMethodPointcutAdvisor;
+import org.springframework.beans.factory.BeanNameAware;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 /**
@@ -57,7 +67,7 @@ import org.springframework.util.StringUtils;
  * @author Artem Bilan
  * @author Chris Gilbert
  */
-public class DefaultKafkaConsumerFactory<K, V> implements ConsumerFactory<K, V> {
+public class DefaultKafkaConsumerFactory<K, V> implements ConsumerFactory<K, V>, BeanNameAware {
 
 	private static final LogAccessor LOGGER = new LogAccessor(LogFactory.getLog(DefaultKafkaConsumerFactory.class));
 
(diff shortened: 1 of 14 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
```diff
@@ -104,14 +119,39 @@ public DefaultKafkaConsumerFactory(Map<String, Object> configs,
...
 
+	@Override
+	public void setBeanName(String name) {
+		this.beanName = name;
+	}
+
+	/**
+	 * Set the key deserializer.
+	 * @param keyDeserializer the deserializer.
+	 * @since 2.5
```
artembilan, line 130: Probably this `since` is wrong for existing methods...

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
```diff
@@ -202,9 +242,49 @@ private void checkForUnsupportedProps(Properties properties) {
...
+
+			@Override
+			public Object invoke(MethodInvocation invocation) throws Throwable {
+				DefaultKafkaConsumerFactory.this.listener.consumerRemoved(id, kafkaConsumer);
+				return invocation.proceed();
+			}
+
+		};
+		NameMatchMethodPointcutAdvisor advisor = new NameMatchMethodPointcutAdvisor(advice);
+		advisor.addMethodName("close");
```
artembilan, line 285: I think creating a proxy just for a `close()` method is an overhead. 
It probably might be more performant is we would just return a wrapper.

Although probably similar. So, just my observation.

spring-kafka/src/main/java/org/springframework/kafka/core/MicrometerConsumerListener.java
```diff
@@ -0,0 +1,86 @@
...
+			List<Tag> consumerTags = new ArrayList<>(this.tags);
+			consumerTags.add(new ImmutableTag("spring.id", id));
+			this.metrics.put(id, new KafkaClientMetrics(consumer, consumerTags));
+			this.metrics.get(id).bindTo(this.meterRegistry);
+		}
+	}
+
+	@Override
+	public synchronized void consumerRemoved(String id, Consumer<K, V> consumer) {
+		KafkaClientMetrics removed = this.metrics.remove(id);
```
artembilan, line 80: Don't we need to call here a `meterRegistry.remove()` for bound meter?
Or in the `destroy()`?
Or is it done some other way? Or we don't need the removal in this case?

Thanks

spring-kafka/src/main/java/org/springframework/kafka/core/MicrometerProducerListener.java
```diff
@@ -0,0 +1,87 @@
...
+			List<Tag> producerTags = new ArrayList<>(this.tags);
+			producerTags.add(new ImmutableTag("spring.id", id));
+			this.metrics.put(id, new KafkaClientMetrics(producer, producerTags));
+			this.metrics.get(id).bindTo(this.meterRegistry);
+		}
+	}
+
+	@Override
+	public synchronized void producerRemoved(String id, Producer<K, V> producer) {
+		KafkaClientMetrics removed = this.metrics.remove(id);
```
artembilan, line 81: Same `Meter` removal concern is here, too.

src/reference/asciidoc/kafka.adoc
```diff
@@ -2732,6 +2785,57 @@ The timers are named `spring.kafka.template` and have the following tags:
 
 You can add additional tags using the template's `micrometerTags` property.
 
+[[micrometer-native]]
+===== Micrometer Native Metrics
+
+Starting with version 2.5, the framework provides <<factory-listeners>> to manage Micrometer `KafkaClientMetric` s whenever producers and consumers are created and closed.
```
artembilan, line 2791: Probably you really mean it as a plain text "Kafka client metrics".
Looks like there is no `KafkaClientMetric` type...

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
garyrussell: The problem with wrappers is you have to use reflection when the underlying API changes (new methods added etc).

I doubt the overhead is more than noise; but I'll run some tests.

spring-kafka/src/main/java/org/springframework/kafka/core/MicrometerConsumerListener.java
garyrussell: Good catch - I assumed the `close()` would unregister, but it doesn't.

src/reference/asciidoc/kafka.adoc
garyrussell: It's `KafkaClientMetrics`.

spring-kafka/src/main/java/org/springframework/kafka/core/MicrometerProducerListener.java
garyrussell: Hmmm - they don't provide any hooks to unregister the underlying meters.

I'll talk to Tommy tomorrow but, for now; they will stay registered and the re-register will be a no-op.

garyrussell: https://github.com/micrometer-metrics/micrometer/issues/2018

PR #1469, opened 2020-04-28
Upgrade Asciidoctor

Buzzardo:
Upgrade Asciidoctor to the latest version, to eliminate technical debt.

I also fixed an invalid reference.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,6 @@ buildscript {
 		maven { url 'https://repo.spring.io/plugins-release' }
 	}
 	dependencies {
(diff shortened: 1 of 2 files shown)


Buzzardo:
The `asciidoctor` task is creating a bunch of false positives for invalid references. Fortunately, the PDF generator generates correct invalid reference messages, so I was able to find and fix two invalid references (both in the same file).

I have filed an issue with the asciidoctor-gradle-plugin project: https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/550

Issue #1475, opened 2020-05-03
Add a specialized ListenableFutureCallback


garyrussell:
See https://stackoverflow.com/questions/61554088/get-objects-value-in-onfailure-method-call-of-kafka-send-method/61560114#61560114

Make it easier to access the failed ProducerRecord.

PR #1476, opened 2020-05-04
GH-1475: Add KafkaSendCallback

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1475

Add `onFailure` with a narrowed `Throwable` to make it easier to access
the failed producer record.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaFailureCallback.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaFailureCallback.java
@@ -26,8 +26,10 @@ import org.springframework.util.concurrent.FailureCallback;
  * @param <V> the value type.
  *
  * @author Gary Russell
+ * @since 2.5
  *
  */
+@FunctionalInterface
 public interface KafkaFailureCallback<K, V> extends FailureCallback {
 
 	@Override
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaFailureCallback.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaFailureCallback.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaFailureCallback.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * An enhanced {@link FailureCallback} for reporting
+ * {@link KafkaProducerException}s.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Gary Russell
+ *
```
artembilan, line 29: `@since 2.5` ?

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaFailureCallback.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public interface KafkaFailureCallback<K, V> extends FailureCallback {
```
artembilan, line 31: `@FunctionalInterface` ?

spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
```diff
@@ -339,7 +344,66 @@ public void onFailure(Throwable ex) {
...
+		KafkaTemplate<Integer, String> template = new KafkaTemplate<>(pf);
+		ListenableFuture<SendResult<Integer, String>> future = template.send("foo", 1, "bar");
+		final CountDownLatch latch = new CountDownLatch(1);
+		final AtomicReference<SendResult<Integer, String>> theResult = new AtomicReference<>();
+		AtomicReference<String> value = new AtomicReference<>();
+		future.addCallback(new KafkaSendCallback<Integer, String>() {
+
+			@Override
+			public void onSuccess(SendResult<Integer, String> result) {
+			}
```
artembilan, line 370: How about making this `onSuccess` as a `default` one in the `KafkaSendCallback` so we can mark the last with with a `@FunctionalInterface`, too, and use it here (or anywhere in the end-user code) as a lambda.

garyrussell: I don't think it makes sense; in most real applications you would want to always implement both methods.

Issue #1501, opened 2020-06-10
Add Micrometer for KafkaStreams


garyrussell:
Also see https://github.com/spring-projects/spring-boot/issues/21890

garyrussell:
Make consistent with Producer/Consumer factory.

PR #1502, opened 2020-06-11
GH-1501: Add Micrometer Support for KafkaStreams

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1501

### DIFF
--- a/src/reference/asciidoc/streams.adoc
+++ b/src/reference/asciidoc/streams.adoc
@@ -141,6 +141,19 @@ Default no-op implementations are provided to avoid having to implement both met
 
 A `CompositeKafkaStreamsInfrastructureCustomizer` is provided, for when you need to apply multiple customizers.
 
+[[streams-micrometer]]
+==== KafkaStreams Micrometer Support
+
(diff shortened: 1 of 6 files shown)


src/reference/asciidoc/streams.adoc
```diff
@@ -141,6 +141,19 @@ Default no-op implementations are provided to avoid having to implement both met
 
 A `CompositeKafkaStreamsInfrastructureCustomizer` is provided, for when you need to apply multiple customizers.
 
+[[streams-micrometer]]
+==== KafkaStreams Micrometer Support
```
artembilan, line 145: I think this should go to `Whats' New` chapter.
Do we have one?

snicoll:
@garyrussell this change is not consistent with what we for producer and consumers. There was a `addListener` and `getListeners()`. I agree that the `FactoryBean` contract makes it a little bit different but the listener being handled in the `FactoryBean` directly, I think it makes sense to adapt.

Thoughts?

garyrussell:
Good point; will change.

Issue #1509, opened 2020-06-17
Support nested BatchListenerFailedException in RecoveringBatchErrorHandler


mhyeon-lee:
RecoveringBatchErrorHandler does not recognize nested BatchListenerFailedException type.

In my case, wrapping and throwing with MessagingException in StreamListenerMessageHandler.

I hope to improve the RecoveringBatchErrorHandler to check for nested BatchListenerFailedException.

PR #1510, opened 2020-06-17
GH-1509 Support nested BatchListenerFailedException in RecoveringBatchErrorHandler

mhyeon-lee:
#1509

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/RecoveringBatchErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/RecoveringBatchErrorHandler.java
@@ -20,9 +20,11 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.BiConsumer;
 
 import org.apache.kafka.clients.consumer.Consumer;
@@ -47,6 +49,7 @@ import org.springframework.util.backoff.BackOff;
  * this handler's {@link BackOff}. If the record is recovered, its offset is committed.
  *
  * @author Gary Russell
+ * @author Myeonghyeon Lee
  * @since 2.5
  *
  */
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/RecoveringBatchErrorHandler.java
```diff
@@ -103,16 +104,16 @@ public RecoveringBatchErrorHandler(@Nullable BiConsumer<ConsumerRecord<?, ?>, Ex
 	public void handle(Exception thrownException, ConsumerRecords<?, ?> data, Consumer<?, ?> consumer,
 			MessageListenerContainer container) {
 
-		Throwable cause = thrownException.getCause();
-		if (!(cause instanceof BatchListenerFailedException)) {
-			this.logger.debug(cause, "Expected a BatchListenerFailedException; re-seeking batch");
+		BatchListenerFailedException batchListenerFailedException = this.getBatchListenerFailedException(thrownException);
```
artembilan, line 107: No `this.` prefix for method calls.

spring-kafka/src/main/java/org/springframework/kafka/listener/RecoveringBatchErrorHandler.java
```diff
@@ -191,4 +192,23 @@ private void commit(Consumer<?, ?> consumer, MessageListenerContainer container,
 		}
 	}
 
+	private BatchListenerFailedException getBatchListenerFailedException(Throwable exception) {
```
artembilan, line 195: This method looks like could be as a recursion:
```
private BatchListenerFailedException getBatchListenerFailedExceptionFromCause(Throwable exception) {
    if (exception == null || exception instanceof BatchListenerFailedException ) {
	    return (BatchListenerFailedException) exception;
    }
    else {
        return getBatchListenerFailedExceptionFromCause(exception.getCause());
    }
}
```

spring-kafka/src/main/java/org/springframework/kafka/listener/RecoveringBatchErrorHandler.java
mhyeon-lee: removed `this.` prefix

spring-kafka/src/main/java/org/springframework/kafka/listener/RecoveringBatchErrorHandler.java
mhyeon-lee: @artembilan 
I fixed it because a stack overflow could occur.
In some cases, Throwable cause() may have itself.
Added check condition to escape the loop when the cause chain is repeated.

```
BatchListenerFailedException target = null;

Set<Throwable> checked = new HashSet<>();
while (throwable.getCause() != null && !checked.contains(throwable.getCause())) {
    throwable = throwable.getCause();
    checked.add(throwable);

    if (throwable instanceof BatchListenerFailedException) {
        target = (BatchListenerFailedException) throwable;
        break;
    }
}
```

I think the while statement is more natural than recursion as the check condition is entered.
If you have a better way, please give us feedback.

I referenced ExceptionUtils in apache commons lang3.

https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java#L861
https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java#L415

garyrussell: It's an unusual situation but I guess it doesn't hurt to add the check.

spring-kafka/src/main/java/org/springframework/kafka/listener/RecoveringBatchErrorHandler.java
```diff
@@ -191,4 +194,24 @@ private void commit(Consumer<?, ?> consumer, MessageListenerContainer container,
 		}
 	}
 
+	private BatchListenerFailedException getBatchListenerFailedException(Throwable throwable) {
```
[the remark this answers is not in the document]
mhyeon-lee: @artembilan 
If Exception and Cause are the same, it can defend.
But, If it has a circular relationship, it cannot escape the recursion.

```
A.cause() -> B
B.cause() -> A
```

This is an unusual case.
Please comment if you don't need to defend.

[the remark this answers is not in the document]
artembilan: OK. Let it be. I wanted to avoid that local `Set` to store causes, but if you both are OK with this, let's merge!

garyrussell:
Thanks for the contribution!

PR #1513, opened 2020-06-17
JsonDeserializer Type Resolution Improvements - javadoc, minor changes

tfactor2:
#1507 
@pivotal-issuemaster This is an Obvious Fix

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
@@ -97,12 +97,14 @@ public class JsonDeserializer<T> implements Deserializer<T> {
 	public static final String USE_TYPE_INFO_HEADERS = "spring.json.use.type.headers";
 
 	/**
-	 * A method name to determine the {@link JavaType} to deserialize the key to.
+	 * A method name to determine the {@link JavaType} to deserialize the key to:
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -97,12 +97,14 @@
 	public static final String USE_TYPE_INFO_HEADERS = "spring.json.use.type.headers";
 
 	/**
-	 * A method name to determine the {@link JavaType} to deserialize the key to.
+	 * A method name to determine the {@link JavaType} to deserialize the key to:
+	 * 'com.Foo.deserialize'. See {@link JsonTypeResolver#resolveType} for the signature.
```
artembilan, line 101: Can we avoid using `foo/bar` in the production code and docs?
Thanks

tfactor2: Thought that it's used intentionally in the project - see `TYPE_MAPPINGS` above:
```java
	/**
	 * Kafka config property to add type mappings to the type mapper:
	 * 'foo=com.Foo,bar=com.Bar'.
	 */
```

artembilan (approved):
OK. We accept your fix.
Sorry for delay and thank you for contribution; looking forward for more!

tfactor2:
Not a problem, thanks.

Issue #1520, opened 2020-06-25
Add producer/consumer decorators


garyrussell:
See https://github.com/spring-cloud/spring-cloud-sleuth/issues/1670#issuecomment-649132257

PR #1523, opened 2020-06-25
GH-1520: Add Producer/Consumer Post Processors

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1520

To assist with Sleuth instrumentation.


garyrussell:
I don't think we need to clutter the documentation with this; it's really for frameworks.

cc/ @adriancole

adriancole:
I think this will help with the specific thing indeed. we have to do some hacks using a MessagePostProcessor in spring-rabbit, and something similar I might follow-up with over there.

PR #1539, opened 2020-07-20
FIX: KafkaMessageListenerContainer#processTimestampSeeks offsetsForTimes method can return map with null values

hosuaby:
Sometimes I have this kind of exception while I try to rewind Kafka partitions by timestamp:

```
Caused by: java.lang.NullPointerException: null
    at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.lambda$processTimestampSeeks$31(KafkaMessageListenerContainer.java:2048)
    at java.util.HashMap.forEach(HashMap.java:1289)
    at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.processTimestampSeeks(KafkaMessageListenerContainer.java:2048)
    at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.processSeeks(KafkaMessageListenerContainer.java:1988)
    at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.pollAndInvoke(KafkaMessageListenerContainer.java:1040)
    at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.run(KafkaMessageListenerContainer.java:970)
    ... 3 common frames omitted
```

Looks like that it's due to the fact that method `consumer.offsetsForTimes(timestampSeeks)` can return map where values can be null. Javadoc of `offsetsForTimes` states it clearly:

```
{@code null} will be returned for the partition if there is no such message.
```

So this PR is basically protection agains null values.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -2090,7 +2090,13 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			if (timestampSeeks != null) {
 				Map<TopicPartition, OffsetAndTimestamp> offsetsForTimes = this.consumer
 						.offsetsForTimes(timestampSeeks);
-				offsetsForTimes.forEach((tp, ot) -> this.consumer.seek(tp, ot.offset()));
+
+				for (TopicPartition tp : offsetsForTimes.keySet()) {
+					OffsetAndTimestamp ot = offsetsForTimes.get(tp);
+					if (ot != null) {
+						this.consumer.seek(tp, ot.offset());
+					}
+				}
 			}
 		}
 


src/api/overview.html
```diff
@@ -1,7 +1,7 @@
 <html>
 <body>
 This document is the API specification for Spring for Apache Kafka project
-<hr/>
+<hr>
```
hosuaby, line 4: API doc generator was complaining about self-closing tag

garyrussell: Which API doc generator? I see no errors with `$ ./gradlew clean api`.

(Thanks for the PR, BTW).

hosuaby: I had this exception:

```
> Task :api
src/api/overview.html:4: error: self-closing element not allowed
<hr/>
^

> Task :api FAILED
```

If you don't have this error, I can rollback this modification.

Thanks for review.

garyrussell: @artembilan Any ideas? We have this tag in all our projects.

artembilan: @hosuaby , tell us, please, what Java version do you use to build this project?
Probably there were some changes since Java 9 in the JavaDoc generator...

hosuaby: Yes, I think it's because it tried to copile it with Java 13 JDK. Now I am on Java 11 and looks like everithing build just fine. I rollbacked this small modification

garyrussell:
Thanks; forgot about this so it missed Wednesday's releases.

Cherry-picked to 2.5.x, 2.4.x, 2.3.x.

hosuaby:
Thank you a lot, @garyrussell

Issue #1538, opened 2020-07-18
Allow for checking the status of a topic just being created


henke77:
🎁 Enhancement
When you create topics inside the EmbeddedKafkaBroker using one of the addTopics() methods you may run into topic already exists exceptions. The problem is that you cannot circumvent this by checking with getTopics() method because it doesn't return topics that are just being created. Therefore, it would be useful if the addTopics() methods would return the CreateTopicsResult that they receive from the AdminClient call or something analog that allows for checking the status of a topic just being created.

garyrussell:
Can you explain the use case? Normally, topics are added to an empty broker at the beginning of a test. It would be unusual to add topics that already exist.

It would be a breaking change for us to change the behavior of the existing methods (we could add other flavors that just return the `CreateTopicsResult` so the user can check the result of individual adds.

Contributions are welcome!

As an alternative, you could call `doWithAdmin()` with your own callback.

henke77:
I tried https://docs.spring.io/spring-kafka/docs/current/reference/html/#using-the-same-brokers-for-multiple-test-classes
Here I ran into the the topic exists exceptions.

henke77:
In addition, you must be sure the topic really exists before assigning a consumer.

garyrussell:
We generally recommend using different topics in each test, to avoid cross-talk between tests and issues with test ordering.

PR #1540, opened 2020-07-20
GH-1538: Embedded broker get results for addTopics

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1538

**cherry-pick to 2.5.x**

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
@@ -22,6 +22,7 @@ import java.io.UncheckedIOException;
 import java.net.InetSocketAddress;
 import java.nio.file.Files;
 import java.time.Duration;
+import java.util.AbstractMap.SimpleEntry;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -33,8 +34,11 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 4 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
```diff
@@ -417,22 +421,100 @@ private void createTopics(AdminClient admin, List<NewTopic> newTopics) {
...
+					try {
+						entry.getValue().get(this.adminTimeout.getSeconds(), TimeUnit.SECONDS);
+						result = null;
+					}
+					catch (InterruptedException | ExecutionException | TimeoutException e) {
+						result = e;
+					}
+					return new SimpleEntry<>(entry.getKey(), result);
+				})
+				.forEach(entry -> results.put(entry.getKey(), entry.getValue()));
```
artembilan, line 491: `Collectors.toMap()`?

spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
```diff
@@ -417,22 +421,100 @@ private void createTopics(AdminClient admin, List<NewTopic> newTopics) {
...
+		Assert.notNull(this.zookeeper, "Broker must be started before this method can be called");
+		for (NewTopic topic : topicsToAdd) {
+			Assert.isTrue(this.topics.add(topic.name()), () -> "topic already exists: " + topic);
+			Assert.isTrue(topic.replicationFactor() <= this.count
+							&& (topic.replicasAssignments() == null
+							|| topic.replicasAssignments().size() <= this.count),
+					() -> "Embedded kafka does not support the requested replication factor: " + topic);
+		}
+
+		return doWithAdminFunction(admin -> createTopicsWithResults(admin, Arrays.asList(topicsToAdd)));
```
artembilan, line 455: Why there is no wrapping to `Set` over here?
And with that it looks like we could call the same `createKafkaTopicsWithResults(set)` without this extra lambda...

spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
garyrussell: Doesn't work (uses `HashMap.merge()`) which doesn't allow `null` values.

spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
garyrussell: ? These are `NewTopic` objects not Strings; `createKafkaTopicsWithResults(set)` takes a set of String (which it converts to a list of `NewTopic`.

I didn't think it was worth trying to weed out duplicates when they provide `NewTopic` objects - because it won't really catch them unless all the propertied are the same.

artembilan: Ah, OK! Missed all those points. 😄 
Merging...

artembilan:
... and cherry-picked to `2.5.x` as https://github.com/spring-projects/spring-kafka/commit/e653cc447260354899bf30a96ce53c0dc1b100bd after squashing and fixing conflicts.

Issue #1552, opened 2020-07-31
Consumer properties set on ConcurrentKafkaListenerContainerFactory are silently ignored


orange-buffalo:
**Affects Version(s):** 2.5.4

**Use Case**
Setup a custom `ConcurrentKafkaListenerContainerFactory` with custom `max.poll.inteval.ms`:
```java
@Bean
public ConcurrentKafkaListenerContainerFactory<?, ?> customListenerContainerFactory(
        ConsumerFactory<Object, Object> kafkaConsumerFactory,
        ConcurrentKafkaListenerContainerFactoryConfigurer configurer) {
   ConcurrentKafkaListenerContainerFactory<Object, Object> factory = new ConcurrentKafkaListenerContainerFactory<>();
   configurer.configure(factory, kafkaConsumerFactory);
   factory.getContainerProperties()
                .getKafkaConsumerProperties()
                .put(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG, 10_000);
   return factory;
}
```

**Expected Result**
`max.poll.inteval.ms` for the consumers created by this factory is set to the provided value (`10_000` in this example) or exception is thrown if not possible or warning is logged if not possible.

**Actual Result**
`max.poll.inteval.ms` configuration is ignored.

**Assessment**
`DefaultKafkaConsumerFactory#createConsumerWithAdjustedProperties` only iterates string values in the provided properties (btw, what is the reason here? Kafka consumer accepts object types). There is a guard check `DefaultKafkaConsumerFactory#checkForUnsupportedProps` that logs warnings when values are not strings, but it only iterates the explicitly set values and does not iterate over defaults, while `containerProperties.kafkaConsumerProperties` is provided as defaults for the validated `Properties` instance.

garyrussell:
Notice that that property (`kafkaConsumerProperties`) is a `Properties` object. It's an unfortunate side-effect of the poor design of `java.util.Properties` being a subclass of `HashTable`. Properties can have defaults:

```java
    /**
     * Creates an empty property list with the specified defaults.
     *
     * @param   defaults   the defaults.
     */
    public Properties(Properties defaults) {
        this.defaults = defaults;
    }
```

If you iterate over the hash table instead, defaults are not looked up; you have to use `getProperty()`.

```java
    /**
     * Searches for the property with the specified key in this property list.
     * If the key is not found in this property list, the default property list,
     * and its defaults, recursively, are then checked. The method returns
[...]

PR #1556, opened 2020-07-31
GH-1552: Support more consumer properties override

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1552

Previously, non-String-valued consumer property overrides were ignored (and logged).

This is due to the unfortunate design of `Properties` being a subclass of `HashTable`
in that only String-valued properties can be accessed in the nested `default`
properties `HashTable`.

- when creating the merged `Properties` to create the consumer, copy any non-String-valued
properties from the default hash table to the new properties object
- in the DKCF, iterate over the hash table to find any other properties not already found
- log an error if we find non-String-valued default properties

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
@@ -303,13 +303,20 @@ public class DefaultKafkaConsumerFactory<K, V> extends KafkaResourceFactory
 	}
 
 	private void checkInaccessible(Properties properties, Map<String, Object> modifiedConfigs) {
+		List<Object> inaccessible = null;
 		for (Enumeration<?> propertyNames = properties.propertyNames(); propertyNames.hasMoreElements(); ) {
 			Object nextElement = propertyNames.nextElement();
 			if (!modifiedConfigs.containsKey(nextElement)) {
-				LOGGER.error(() -> "Non-String-valued default properties are inaccessible; use a String value or "
-						+ "make it an explicit property instead of a default: "
-						+ nextElement);
+				if (inaccessible == null) {
+					inaccessible = new ArrayList<>();
+				}
+				inaccessible.add(nextElement);
 			}
+		}
+		if (inaccessible != null) {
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
```diff
@@ -283,23 +285,32 @@ public boolean removeListener(Listener<K, V> listener) {
...
-	private void checkForUnsupportedProps(Properties properties) {
-		properties.forEach((key, value) -> {
-			if (!(key instanceof String) || !(value instanceof String)) {
-				LOGGER.warn(() -> "Property override for '" + key.toString()
-					+ "' ignored, only <String, String> properties are supported; value is a(n) " + value.getClass());
+	private void checkInaccessible(Properties properties, Map<String, Object> modifiedConfigs) {
+		for (Enumeration<?> propertyNames = properties.propertyNames(); propertyNames.hasMoreElements(); ) {
+			Object nextElement = propertyNames.nextElement();
+			if (!modifiedConfigs.containsKey(nextElement)) {
+				LOGGER.error(() -> "Non-String-valued default properties are inaccessible; use a String value or "
```
artembilan, line 309: Wouldn't it be better to collect them all and log ERROR only once for all of them?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -674,6 +674,18 @@ else if (listener instanceof MessageListener) {
 			this.subBatchPerPartition = setupSubBatchPerPartition();
 		}
 
+		private Properties propertiesFromProperties() {
+			Properties props = new Properties(this.containerProperties.getKafkaConsumerProperties());
+			Set<String> stringPropertyNames = props.stringPropertyNames();
+			// Copy non-string-valued properties from the default hash table to the new properties object
+			this.containerProperties.getKafkaConsumerProperties().forEach((key, value) -> {
```
artembilan, line 681: Let's extract local variable for this `this.containerProperties.getKafkaConsumerProperties()` !

PR #1564, opened 2020-08-14
Fix idleBetweenPolls max value calculation

garyrussell:
- Wrong default used when `max.poll.interval.ms` is not specified - 30s Vs 300s
- Value set in `@KafkaListener.properties` ignored for this calculatio and default
  used instead
- Also use actual default for `ConsumerConfig.DEFAULT_API_TIMEOUT_MS_CONFIG`

**I will back port - expecting conflicts**


artembilan:
@garyrussell , 

have back-ported this yet?

I'm about to cherry-picking your #1570 

Thanks

PR #1575, opened 2020-08-24
Seek Improvements

garyrussell:
- add convenience methods to `AbstractConsumerSeekAware`

- compile tests with Java 11

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -133,6 +133,10 @@ subprojects { subproject ->
 		targetCompatibility = 1.8
 	}
 
+	compileTestJava {
+		sourceCompatibility = 11
+	}
+
 	compileTestKotlin {
 		kotlinOptions {
 			jvmTarget = '1.8'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
-dist: trusty
+dist: bionic
 language: java
-jdk: oraclejdk8
+jdk: openjdk11
 install: true
 before_cache:
   - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -133,6 +133,10 @@ subprojects { subproject ->
 		targetCompatibility = 1.8
 	}
 
+	compileTestJava {
+		sourceCompatibility = 11
```
artembilan, line 137: What made you to move to Java 11?
Thanks

garyrussell: The main driver was `Map.of(new TopicPartition("foo", 0), 0L, new TopicPartition("foo", 1), 0L)` in tests. I hate building maps in tests with `put()`; that's why I added `MapBuilder` in SI and SA (but we don't have one in SK).

But we also get new language features (`var`) - like we did with SI 4.3.x and using Java 8 in tests.

artembilan: Sure! Just curious what could be so useful in other projects.

We might need re-test the whole project to be sure that JavaDocs, for example, are built OK, too.

garyrussell: I actually went to 14 initially, but travis only goes up to 11 😦

artembilan: Good to know. Merging...

garyrussell: Ugh... https://stackoverflow.com/questions/51992347/missing-iframe-view-for-javadocs-jdk-11

artembilan:
```
24-Aug-2020 17:29:57 | > Task :api
-- | --
24-Aug-2020 17:29:57 | 1 error
24-Aug-2020 17:29:57 | src/api/overview.html:4: error: self-closing element not allowed
24-Aug-2020 17:29:57 | <hr/>
24-Aug-2020 17:29:57 | ^
24-Aug-2020 17:29:57 |  
24-Aug-2020 17:29:57 | > Task :api FAILED
```
Hm. I should've gave you a couple minutes to check it before merging upstream.

Anyway let try switch it to Java 14 on CI!

build.gradle
garyrussell: I guess we'll get used to it.

PR #1579, opened 2020-08-27
Topic Provisioning - default partitions/replicas

garyrussell:
The `TopicBuilder` now supports deferring to the broker defaults
for number of partitions and replicas.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/TopicBuilder.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/TopicBuilder.java
@@ -26,7 +26,7 @@ import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.common.config.TopicConfig;
 
 /**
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/config/TopicBuilder.java
```diff
@@ -20,12 +20,14 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.common.config.TopicConfig;
 
 /**
- * Builder for a {@link NewTopic}.
+ * Builder for a {@link NewTopic}. Since 2.3 partitions and replicas default to
```
artembilan, line 29: Probably `2.6`? You don’t mention back-porting. Plus looks like this class is since `2.3` per se...

PR #1580, opened 2020-08-29
Update samples to setup kafka simply

dbgsprw:
- Add docker-compose.yml
- Update README.adoc

Suggest an explanation about simple setting up kafka on the sample.

### DIFF
--- a/samples/docker-compose.yml
+++ b/samples/docker-compose.yml
@@ -1,15 +1,20 @@
 version: '2'
 services:
   zookeeper:
-    image: wurstmeister/zookeeper
-    ports:
-      - "2181:2181"
+    image: confluentinc/cp-zookeeper:latest
+    environment:
+      ZOOKEEPER_CLIENT_PORT: 2181
+      ZOOKEEPER_TICK_TIME: 2000
   kafka:
-    image: wurstmeister/kafka
+    image: confluentinc/cp-kafka:latest
+    depends_on:
+      - zookeeper
     ports:
       - "9092:9092"
     environment:
-        KAFKA_ADVERTISED_HOST_NAME: localhost
-        KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
-    volumes:
-      - /var/run/docker.sock:/var/run/docker.sock
+      KAFKA_BROKER_ID: 1
+      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka:29092,PLAINTEXT_HOST://localhost:9092
+      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,PLAINTEXT_HOST:PLAINTEXT
+      KAFKA_INTER_BROKER_LISTENER_NAME: PLAINTEXT
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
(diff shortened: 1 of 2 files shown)


samples/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: '2'
+services:
+  zookeeper:
+    image: wurstmeister/zookeeper
```
daehokimm, line 4: IMHO, [confluentinc/cp-zookeeper](https://hub.docker.com/r/confluentinc/cp-zookeeper) & [confluentinc/cp-kafka](https://hub.docker.com/r/confluentinc/cp-kafka) image seems to be more maintainable. and if the images are changed, the options will have to change too.

daehokimm (commented):
IMHO, [confluentinc/cp-zookeeper](https://hub.docker.com/r/confluentinc/cp-zookeeper) & [confluentinc/cp-kafka](https://hub.docker.com/r/confluentinc/cp-kafka) image seems to be more maintainable. and if the images are changed, the options will have to change too.

samples/docker-compose.yml
dbgsprw: Thanks for your opinion.

artembilan:
I don't think it is this project responsibility to bias to some specific Docker image.
It is really out of this project scope how your Kafka Broker is provided.

So, here is my 👎 for this suggestion.

Thanks

dbgsprw:
> I don't think it is this project responsibility to bias to some specific Docker image.
> It is really out of this project scope how your Kafka Broker is provided.
> 
> So, here is my 👎 for this suggestion.
> 
> Thanks

I fully agree with your opinion on project responsibility, even though I may wish there was an easy way to execute the sample.

Then, what do you think about adding the default Kafka Broker used by this sample to README.md?

Thanks.

artembilan:
Well, these samples are fully based on Spring Boot where a default `bootstrapServer` is definitely this one: `localhost:9092`.

So, it is fully up to end-user to ensure that Kafka broker is on that port.
I probably would start from regular Apache Kafka tutorial where they recommend to download Kafka and start it locally. 
And then I go to Spring solution to see what is there over the standard client. One more time: "client". There is nothing broker-specific in this project therefore we don't take responsibility what and how broker is provided.

With your suggestion I have to install Docker somehow, which doesn't work for me on Windows.
So, what is a simpler way to go then? Probably just leave a decision up to end-user without forcing them for some opinion which is not relevant to this project.

dbgsprw:
> Well, these samples are fully based on Spring Boot where a default `bootstrapServer` is definitely this one: `localhost:9092`.
> 
> So, it is fully up to end-user to ensure that Kafka broker is on that port.
> I probably would start from regular Apache Kafka tutorial where they recommend to download Kafka and start it locally.
> And then I go to Spring solution to see what is there over the standard client. One more time: "client". There is nothing broker-specific in this project therefore we don't take responsibility what and how broker is provided.
> 
> With your suggestion I have to install Docker somehow, which doesn't work for me on Windows.
> So, what is a simpler way to go then? Probably just leave a decision up to end-user without forcing them for some opinion which is not relevant to this project.

@artembilan 
Thank you for your time and good opinion. I think it's important whether `localhost:9092` is predictable to end-user.
In your opinion it is predictable. I'll close this PR soon.

Thank you very much.

garyrussell:
We should add test cases with the embedded kafka broker - we are always getting questions about testing with the embedded broker in Boot applications.

Issue #1184, opened 2019-08-01
Add a test case for ReactiveKafkaConsumerTemplate.receiveExactlyOnce


garyrussell:

rhamedy:
@garyrussell how is this for contribution?

garyrussell:
Yes, contributions are welcome.

PR #1586, opened 2020-09-07
2.5.x Add a test case for ReactiveKafkaConsumerTemplate.receiveExactlyOnce

wkennedy:
#1184 

Add a test case for ReactiveKafkaConsumerTemplate.receiveExactlyOnce

Added two unit tests for this issue:

- shouldSendOneRecordTransactionallyViaTemplateAsSenderRecordAndReceiveItExactlyOnce ::: this method tests that consumption of message A and produces message B such that B = F(A) and A is considered consumed if and only if B is successfully produced.
- shouldSendOneRecordTransactionallyViaTemplateAsSenderRecordAndReceiveItExactlyOnceWithException ::: this method tests that if publishing of message B fails, then the transaction is aborted and the message is able to be consumed again.


garyrussell:
Thanks!

Merged as fb975dd45ccacb12f35d10f16d1cd43fa5dcd46e and cherry-picked to master.

Issue #1587, opened 2020-09-14
Kafka Consumer lag not zero when input topic transactional


ekanthi:
**Affects Version(s): 2.5.5.RELEASE
Description:

When using ConcurrentMessageListenerContainer, with batch processing and "read.committed" consumers, reading off of a transacted topic, we see that the consumer lag is never 0. It is off by 1 offset for each partition(Even after processing all messages)

Based on a similar jira  on Kafka Streams: https://issues.apache.org/jira/browse/KAFKA-6607, we see that Matthias, recommends the following: 
"Note that all applications using a plain consumer may face the same issue if they use `KafkaConsumer#commitSync(Map<TopicPartition, OffsetAndMetadata> offsets)`: to address the issue, the correct pattern is to either commit "nextRecord.offset()" (if the next record is available already, ie, was returned by `poll()`, or use `consumer.position()` that takes the commit marker into account and would "step over it")."

However, in spring-kafka source: KafkaMessageListenerContainer->ackImmediate method,  uses : new OffsetAndMetadata(record.offset() + 1), when commiting offsets.

garyrussell:
The consumer has no knowledge as to whether the producer uses transactions or not; in fact, there is nothing to prevent having both types of producer on a topic. Furthermore, there is only an extra slot per transaction - not per record. If a transaction sends 10 records in a transaction 10 slots are used. The consumer has no idea how many records were published in the transaction so we can't always advance by 2.

You should just ignore the bogus "lag"; it's not really a lag per se, just an artifact of the slot used to indicate whether a transaction committed or rolled back.

ekanthi:
Hi Gary, I was not even suggesting updating the offset by 2. I was just referring to Matthias's suggestion of using the offset of the next record, if a next record exists in the batch or the consumer or or use `consumer.position()` that takes the commit marker into account and would "step over it"

garyrussell:
My concern with that approach is (a) we'd have to check if there is a next record for the current topic/partition in the batch (we currently iterate over the entire `ConsumerRecords` not by topic/partition). 

My concern with using `position()` is it might cause another round trip to the broker (not sure about that; I will have to check the code tomorrow).

I'll look to see what he did to fix KAFKA-6607.

My bigger issue is whether it's really necessary to add such complexity just for an ostensibly cosmetic issue.

garyrussell:
I have confirmed that the `position()` calls don't need a round trip to the broker, so I think I can come up with a solution.

[...]

PR #1589, opened 2020-09-15
GH-1587: Don't Correct TX Offsets After Seek(s)

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1587

We should not advance the consumer partition if Seek operations have been
performed; skip fixing the offsets if that condition is detected.

Capture the positions after the poll and check during the fix operation.

This is difficult to write a unit test for; tested with a Boot application,
observing the correct behavior via DEBUG logs.

**cherry-pick to 2.5.x**


artembilan:

Issue #1591, opened 2020-09-16
DefaultAfterRollbackProcessor is ignoring backoff policy when using a batch consumer


Rajh:
**Affects Version(s):** 2.5.5.RELEASE

Configuration :
new DefaultAfterRollbackProcessor<>(null, new FixedBackOff(1000L, FixedBackOff.UNLIMITED_ATTEMPTS));

Actual behavior:
On an exception, the message is retried without any delay.

Expected behavior :
Wait 1 sec between each retries.

Issue :
The DefaultAfterRollbackProcessor is ignoring backoff policy when using a batch consumer.

PR #1596, opened 2020-09-24
GH-1591: Apply BackOff in DARP with Batch Listener

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1591

`BackOff` was ignored for batch listeners (not recoverable).

- move code from `SeekToCurrentBatchErrorHandler` to `ListenerUtils`
- call from both STCEH and DARP
- also clear the thread state (in both) if a batch fails and subsequently succeeds

**I will do the backports - conflicts are expected**


garyrussell:
Back-ported to 2.5.x, 2.4.x, 2.3.x.

PR #1609, opened 2020-11-05
GH-1587: Fix NPE with Foreign TM and fixTxOffsets

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1587

Code was using the presence of a transaction template instead of the KTM.

**cherry-pick to 2.5.x**


artembilan:
Cherry-picked to `2.5.x` as https://github.com/spring-projects/spring-kafka/commit/fb1977c14cd72030828256ec7b7750d0fc85442a after fixing conflicts.

Issue #1610, opened 2020-11-06
Add an Option to Reset the FailedRecordTracker if the Exception Type Changes Between Failures


garyrussell:
See https://stackoverflow.com/questions/64708287/how-to-reset-the-retry-count-in-spring-kafka-consumer-when-the-exception-thrown/64716441#64716441

PR #1611, opened 2020-11-06
GH-1610: Reset Retry State when Exception Changes

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1610


spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
```diff
@@ -53,6 +53,8 @@
 
 	private boolean resetStateOnRecoveryFailure = true;
 
+	private boolean resetStateOnExceptionChange;
```
artembilan, line 56: I'm sorry, do I miss anything?
Looks like this property is out of use...

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
garyrussell: Oops

Issue #1612, opened 2020-11-09
Enforcing container to stop on transaction fencing


Rajh:
**Affects Version(s):** 2.5.6.RELEASE

Regarding this stackoverflow question :
https://stackoverflow.com/questions/64665725/why-i-lost-messages-when-using-kafka-with-eos-beta/64665928

When a ProducerFencedException | FencedInstanceIdException occurs, the framework assumes that the consumer will stop, and thus is just logging : 
https://github.com/spring-projects/spring-kafka/blob/fb1977c14cd72030828256ec7b7750d0fc85442a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1519-L1523

It would be usefull to have a safeguard , enforcing the consumer container to stop to prevent losing message in an unexpected behavior concerning this exception.

garyrussell:
The consumer cannot tell if the fenced exception is caused by a rebalance or timeout. This is a known limitation of Kafka itself, and will be addressed in a future release.

We can certainly add an option to stop the container on these exceptions, but the proper fix can only be done when the new exception is thrown for a timeout.

https://issues.apache.org/jira/browse/KAFKA-9803

garyrussell:
In the meantime, you should make sure the `transaction.timeout.ms` is long enough to process a batch of records without timing out.

Rajh:
Reading https://issues.apache.org/jira/browse/KAFKA-9803 I can see :

> Currently the coordinator does not distinguish these two cases. Both will end up as a `ProducerFencedException`, which means the producer needs to shut itself down.

Actually the producer is not shutting down, is it ? Or if it is, what happens to the container ? Since it is not rollbacking, does it try to re produce the message and then the ProducerFactory creates a new one ?

garyrussell:
The producer factory should recycle the producer (it does for me with your test application with the short tx timeout). Any error on `begin/abort/commitTransaction()` closes the producer and a new one is created.

Rajh:
Yes but since the exception is ignored the consumer skip the messages with the failed producer and continues with a new batch with a new producer.

garyrussell:
Right, so until they fix it, the only thing we can do is stop the container.

Here's another work-around:

Set the container's `idleEventInterval`. Add an event listener to catch `ListenerContainerIdleEvent`; then perform a dummy (no-op) transaction; e.g. once a day.

Rajh:
[...]

PR #1614, opened 2020-11-09
GH-1612: Option: Producer Fenced: Stop Container

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1612

**cherry-pick to 2.5.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
@@ -254,6 +254,8 @@ public class ContainerProperties extends ConsumerProperties {
 
 	private TransactionDefinition transactionDefinition;
 
+	private boolean stopContainerWhenFenced;
+
 	/**
 	 * Create properties for a container that will subscribe to the specified topics.
 	 * @param topics the topics.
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -745,6 +747,28 @@ public void setAdviceChain(Advice... adviceChain) {
...
+	}
+
+	/**
+	 * Set to true to stop the container when a
+	 * {@link org.apache.kafka.common.errors.ProducerFencedException} is thrown.
+	 * Currently, there is no way to determine if such an exception is thrown due to a
+	 * rebalance Vs. a timeout. We therefore cannot call the after rollback processor. The
+	 * best solution is to ensure that the {@code transaction.timeout.ms} is large enough
+	 * so that transactions don't time out.
+	 * @param stopContainerWhenFenced true to stop the container.
```
artembilan, line 766: `@since 2.5.8`?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2848,4 +2849,13 @@ public void onSuccess(Object result) {
 
 	}
 
+	@SuppressWarnings("serial")
+	private static class StopAfterFenceException extends KafkaException {
```
artembilan, line 2853: Don't we bubble this exception to end-user code?
So, they won't be able to handle it properly since it is `private`.
Do I miss anything?

Thanks

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: No; this is thrown on the task executor thread (`run()` dies). We should, however, retain the root cause.

artembilan:

Issue #1613, opened 2020-11-09
Fix Race in DKPF.reset()


garyrussell:
The cache is cleared which could cause a producer that was in-use during the reset to be not returned to the cache.

PR #1616, opened 2020-11-09
GH-1613: Fix Race in Producer Factory

garyrussell:
Resolves

Do not clear the cache in `destroy()` - just drain the producer queues instead.
Fixes a race condition where a producer could be returned to the cache between draining
the queues and clearing the map.
Just leave the queue(s) in place to accept new returns.

**cherry-pick to 2.5.x, 2.4.x, 2.3.x, 1.3.x**


artembilan:
Cherry-picked to `2.5.x`, `2.4.x` & `2.3.x`.

There is no `cache.clear()` in the `1.3.x` branch. So, nothing to back-port.

Issue #1618, opened 2020-11-10
Add a Reason to ConsumerStoppedEvent


garyrussell:

garyrussell:
Child container added to event twice.

PR #1619, opened 2020-11-10
GH-1618: Remove Duplicate Parameter

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1618

Incorrectly added the child container as a parameter for the `ConsumerStoppedEvent`.
The child container is already passed in as the `source`.

**cherry-pick to 2.5.x**


artembilan:

PR #1635, opened 2020-11-20
Manual Assignment: Support Partition Ranges, Lists

garyrussell:


spring-kafka/src/test/java/org/springframework/kafka/listener/ManualAssignmentInitialSeekTests.java
```diff
@@ -111,6 +125,12 @@ public void discardRemainingRecordsFromPollAndSeek() throws Exception {
 		public void foo(String in) {
 		}
 
+		@KafkaListener(id = "pp", autoStartup = "false",
+				topicPartitions = @org.springframework.kafka.annotation.TopicPartition(topic = "foo",
+						partitions = "#{#parsePartitions('0-5, 7, 10-15')}"))
```
artembilan, line 130: Why can't we split and parse ranges in the framework?
As long as we got a value as a string we can do with it whatever we need.

I think even that `#{'0,1,2'.split(',')}` is redundant for end-user experience...

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -787,6 +811,36 @@ private void addFormatters(FormatterRegistry registry) {
...
 
+	/**
+	 * Parse a list of partitions into an array. Example: "0-5,10-15".
+	 * @param partsString the list of partions.
+	 * @return the array.
+	 */
+	public static Integer[] parsePartitions(String partsString) {
+		String[] partsStrings = partsString.split(",");
+		if (partsStrings.length == 1) {
+			return new Integer[] { Integer.parseInt(partsStrings[0].trim()) };
```
artembilan, line 822: Why do you exit here early?
Doesn't look like you check for the dash in that single value.
How something like this `0-5` is going to work here?

artembilan (commented):
Some review for your consideration.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -787,6 +790,39 @@ private void addFormatters(FormatterRegistry registry) {
...
+				}
+			}
+			else {
+				parsePartitions(part).stream().forEach(p -> parts.add(p));
+			}
+		}
+		return parts.stream()
+				.sorted()
+				.distinct()
+				.collect(Collectors.toList());
```
artembilan, line 823: Looks like you do a double `collect()`.
This method is `private` and if you don't want to mix concern with that `map()` in here, I think it is fully safe just return a `Stream<Integer>`  for the further processing rather than re-stream the list one more time.

artembilan (changes requested):

Issue #1646, opened 2020-11-26
Infinite errors due to fixTxOffsets


Rajh:
**Affects Version(s):** 2.5.9.RELEASE

**Description**
The application is consuming 6 differents topics.
4 "normal" and 2 "compacted" topics.

**Configuration**
Normal consumer (assuming the normal is causing those errors)
```
            factory.setConsumerFactory(kafkaConsumerFactory);
            factory.setAutoStartup(false);
            factory.setRecordFilterStrategy(deserializationErrorRecordFilterStrategy);
            factory.setAfterRollbackProcessor(afterRollbackProcessor);
            factory.getContainerProperties().setTransactionManager(kafkaTransactionManager);
            factory.getContainerProperties().setEosMode(ContainerProperties.EOSMode.BETA);
            factory.getContainerProperties().setFixTxOffsets(true);
            factory.getContainerProperties().setStopContainerWhenFenced(true);
            factory.getContainerProperties().setSubBatchPerPartition(false);
            factory.setBatchListener(true);
            factory.getContainerProperties().setAckMode(ContainerProperties.AckMode.MANUAL_IMMEDIATE);
```

**Problem**
Errors during a rolling update of the application, each partition of a single topic (not all) fired this error
```
Failed to seek myTopic-4 to 8
java.lang.IllegalStateException: No current assignment for partition myTopic-4
	at org.apache.kafka.clients.consumer.internals.SubscriptionState.assignedState(SubscriptionState.java:354)
	at org.apache.kafka.clients.consumer.internals.SubscriptionState.seekUnvalidated(SubscriptionState.java:371)
	at org.apache.kafka.clients.consumer.KafkaConsumer.seek(KafkaConsumer.java:1608)
	at jdk.internal.reflect.GeneratedMethodAccessor169.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:344)
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:205)
	at com.sun.proxy.$Proxy98.seek(Unknown Source)
	at org.springframework.kafka.listener.SeekUtils.lambda$seekPartitions$5(SeekUtils.java:119)
[...]

PR #1648, opened 2020-11-30
GH-1646 Clear lastCommits after fixTxOffset

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1646

```
Failed to correct transactional offset(s)
java.lang.IllegalStateException: You can only check the position for partitions assigned to this consumer.
```

It is not clear how this happened since we remove revoked partitions from `lastCommits`.

However, `lastCommits` should be cleared, even if successful, to avoid unnecessary
processing if no records are received by the next poll.

- add the `lastCommits` to the error log
- clear `lastCommits` in a finally block

**cherry-pick to 2.5.x**


artembilan:

PR #1649, opened 2020-12-01
2.7.0 RC3 Kafka Clients

garyrussell:
- remove `retries` from tests (deprecated)
- deprecate reactive template `flush()`

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 the original author or authors.
+ * Copyright 2019-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -130,7 +130,7 @@ public class ReactiveKafkaProducerTemplate<K, V> implements AutoCloseable, Dispo
 	 * scheduled to avoid a deadlock; see
 	 * https://issues.apache.org/jira/browse/KAFKA-10790
 	 */
-	@Deprecated(since = "2.7")
+	@Deprecated(since = "2.7", forRemoval = true)
 	public Mono<?> flush() {
 		return doOnProducer(producer -> {
 			producer.flush();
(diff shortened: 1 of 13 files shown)


artembilan:
@garyrussell ,

Probably this one should go to the `upstream/2.7.0-WIP`, not `master`.

Please, confirm.

garyrussell:
No; I believe it can go to master, I have already branched of 2.6.x.

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplate.java
```diff
@@ -121,10 +121,20 @@ public ReactiveKafkaProducerTemplate(SenderOptions<K, V> senderOptions, RecordMe
...
+	/**
+	 * Flush the producer.
+	 * @return {@link Mono#empty()}.
+	 * @deprecated - flush does not make sense in the context of a reactive flow since,
+	 * the send completion signal is a send result, which implies that a flush is
+	 * redundant. If you use this method with reactor-kafka 1.3 or later, it must be
+	 * scheduled to avoid a deadlock; see
+	 * https://issues.apache.org/jira/browse/KAFKA-10790
+	 */
+	@Deprecated(since = "2.7")
```
artembilan, line 133: What is this `@Deprecated` with `since`?
My one in Java 8 is still without it.
Is it going to work with that Java version?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplate.java
garyrussell: It was added in Java 9 - since we now build this project with Java 11 (with Java 8 as a target), we can use it.

I have verified that a Boot app targeted at Java 8 is fine; I pushed another commit with `forRemoval = true`.

artembilan: OK. So, I need to install that Java 11 to be able to build this project locally 😄 .

Merging then...

garyrussell: That's been true since https://github.com/spring-projects/spring-kafka/commit/c8cf57c4951d249ff1c0fa8180bdbda35d2a4e23#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7

PR #1666, opened 2021-01-04
Add getEndOffsets() to KafkaTestUtils

garyrussell:
**cherry-pick to 2.6.x**

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
@@ -251,8 +251,9 @@ public final class KafkaTestUtils {
 	 * @param partitions the partitions, or null for all partitions.
 	 * @return the map of end offsets.
 	 * @since 2.6.5
+	 * @see Consumer#endOffsets(Collection, Duration)
 	 */
-	public static <K, V> Map<TopicPartition, Long> getEndOffsets(Consumer<K, V> consumer, String topic,
+	public static Map<TopicPartition, Long> getEndOffsets(Consumer<?, ?> consumer, String topic,
 			Integer... partitions) {
 
 		Collection<TopicPartition> tps;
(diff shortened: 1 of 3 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
```diff
@@ -239,6 +242,38 @@ public static OffsetAndMetadata getCurrentOffset(String brokerAddresses, String
...
+			Map<String, List<PartitionInfo>> parts = consumer.listTopics(Duration.ofSeconds(10));
+			tps = parts.entrySet()
+					.stream()
+					.filter(entry -> entry.getKey().equals(topic))
+					.flatMap(entry -> entry.getValue().stream())
+					.map(pi -> new TopicPartition(topic, pi.partition()))
+					.collect(Collectors.toList());
+		}
+		else {
+			Assert.noNullElements(partitions, "'partitions' cannot have nulll elements");
```
artembilan, line 269: Typo:

```suggestion
			Assert.noNullElements(partitions, "'partitions' cannot have null elements");
```

spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
```diff
@@ -239,6 +242,38 @@ public static OffsetAndMetadata getCurrentOffset(String brokerAddresses, String
...
 
+	/**
+	 * Return the end offsets of the requested topic/partitions
+	 * @param <K> the key type.
+	 * @param <V> the value type.
+	 * @param consumer the consumer.
+	 * @param topic the topic.
+	 * @param partitions the partitions, or null for all partitions.
+	 * @return the map of end offsets.
+	 * @since 2.6.5
```
artembilan, line 253: We probably need to add a `@see` for JavaDocs of that `consumer.endOffsets()` because that is exactly an info we get back.

spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
```diff
@@ -239,6 +242,38 @@ public static OffsetAndMetadata getCurrentOffset(String brokerAddresses, String
...
+	 */
+	public static <K, V> Map<TopicPartition, Long> getEndOffsets(Consumer<K, V> consumer, String topic,
```
artembilan, line 255: I think these generic arguments don't give any value or benefits. 
Probably `Consumer<?, ?>` is what would work for end-users. Even if we need to do some casting internally, it is not going to affect target applications.

artembilan (changes requested):
Any chances to document this new API of testing utils?
Or we just point to its JavaDocs from the reference?

Thanks

artembilan:

Issue #1668, opened 2021-01-05
ConsumerAwareRebalanceListener methods inherited from base interface are never called


mvh77:
**Affects Version(s):** 2.6.4

I implemented this:

```java
var cp = new ContainerProperties(MY_TOPIC);
cp.setConsumerRebalanceListener(new ConsumerAwareRebalanceListener() {

    @Override
    public void onPartitionsAssigned(Consumer<?, ?> consumer, Collection<TopicPartition> partitions) {
        // this will be called ok        
    }

    @Override
    public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
        // from base interface, this will never be called
    }
});
```

onPartitionsRevoked (from base interface) is never called. The problem is at _KafkaMessageListenerContainer_ (line 2583-2589). The workaround is obviously to implement _onPartitionsRevokedBeforeCommit_ instead but the default implementation of _onPartitionsRevokedBeforeCommit_ should probably call _onPartitionsRevoked_ and the default implementations of _onPartitionsAssigned_ & _onPartitionsLost_ should also call the base IF methods?

garyrussell:
While it is intended that you implement the consumer-aware methods, the default implementation should invoke the non-consumer-aware methods if not implemented.

garyrussell:
A simple work around is to implement the consumer-aware method and ignore the consumer.

PR #1669, opened 2021-01-05
GH-1688: ConsumerAwareRebalListener Improvements

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1668

Users might implement one of the not-consumer-aware methods so the default
methods should delegate to those.

Call in try/catch block, expecially `...RevokedBeforeCommit` to prevent the
commits to be skipped.

**cherry-pick to 2.6.x**


artembilan:

PR #1671, opened 2021-01-07
fixed typo in ContainerProperties

gyugyu90:
- fixed variables and methods which include typo `timout` to `timeout`

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
@@ -49,6 +49,7 @@ import org.springframework.util.CollectionUtils;
  * @author Artem Yakshin
  * @author Johnny Lim
  * @author Lukasz Kaminski
+ * @author Kyuhyeok Park
  */
 public class ContainerProperties extends ConsumerProperties {
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
@@ -563,6 +563,11 @@ public class ContainerProperties extends ConsumerProperties {
 		return this.consumerStartTimeout;
 	}
 
+	@Deprecated
+	public Duration getConsumerStartTimout() {
+		return this.consumerStartTimeout;
+	}
+
 	/**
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@gyugyu90 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1671)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1671) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@gyugyu90 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1671)!

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -559,18 +559,18 @@ public void setMicrometerTags(Map<String, String> tags) {
 		return Collections.unmodifiableMap(this.micrometerTags);
 	}
 
-	public Duration getConsumerStartTimout() {
-		return this.consumerStartTimout;
+	public Duration getConsumerStartTimeout() {
```
artembilan, line 562: We can't do this unconditionally: it is a breaking change.
Even if it is a bug, some projects may rely on calling this API, therefore changing names for these method will cause a compilation (and execution) problem in those target applications.

We may consider to accept the change as is for the current `2.7` version, but we definitely should consider to have a `@Deprecated` variant for the previous `2.6.x` version. Although I would stay with deprecated variants until the next minor (or major) version...

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
gyugyu90: You're right. I restored the methods and marked them deprecated

artembilan (changes requested):
Please, add your name to the `@author` list and we will merge it shortly.

artembilan:
@gyugyu90 ,

Thank you for the contribution; looking forward for more!

artembilan:

Issue #1675, opened 2021-01-11
Add support for Payload validation on @KafkaHandler methods


ITman1:
Hello,
opening this issue to address possible enhancement which we are currently missing. 

It would be very benefitial for us if we could use JSR 303 validation on top of `@KafkaHandler` methods just like we can already on `@KafkaListener` methods.

You can check also my question on [StackOverflow](https://stackoverflow.com/questions/65645415/valid-payload-on-kafkahandler-bug-or-missing-feature/65669515#65669515).

Thank you.

PR #1677, opened 2021-01-11
GH-1675: Validation on @KafkaHandler Methods

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1675

Add support for payload validation with `@KafkaHandler` methods.

**cherry-pick to 2.6.x, 2.5.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/DelegatingInvocableHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/DelegatingInvocableHandler.java
@@ -151,7 +151,7 @@ public class DelegatingInvocableHandler {
 	 * @param beanExpressionContext the context.
 	 * @param beanFactory the bean factory.
 	 * @param validator the validator.
-	 * @since 2.1.11
+	 * @since 2.5.11
 	 */
 	public DelegatingInvocableHandler(List<InvocableHandlerMethod> handlers,
 			@Nullable InvocableHandlerMethod defaultHandler,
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/DelegatingInvocableHandler.java
```diff
@@ -114,13 +129,35 @@ public DelegatingInvocableHandler(List<InvocableHandlerMethod> handlers,
...
+	/**
+	 * Construct an instance with the supplied handlers for the bean.
+	 * @param handlers the handlers.
+	 * @param defaultHandler the default handler.
+	 * @param bean the bean.
+	 * @param beanExpressionResolver the resolver.
+	 * @param beanExpressionContext the context.
+	 * @param beanFactory the bean factory.
+	 * @param validator the validator.
+	 * @since 2.1.11
```
artembilan, line 154: `2.5.11`?

Why then you say in the PR description that we need to back-port it back to `2.3.x`?

artembilan (changes requested):
OK. I can live with this solution - only need to agree with `@since` and all the back-port versions.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/DelegatingInvocableHandler.java
garyrussell: Fixed and updated PR comment.

artembilan:
... and cherry-picked to `2.6.x` & `2.5.x` after fixing conflicts in `What's New`.

Issue #1683, opened 2021-01-17
Payload value must not be empty for Acknowledgment parameter


selalerercapitolis:
**Affects Version(s):2.6.5

Copying from https://github.com/spring-projects/spring-framework/issues/26389 since it seems this is right place for it.

I have a @KafkaListener with this signature:

```
    @KafkaListener(topics = "${matcher.reply.channel:matcher_response${matcher.kafka.topic.suffix}}")
    public void processMessage(@Payload ConsumerRecord<String, String> record, Acknowledgment acknowledgment) {

```
When I receive a message I get an exception that Payload value must not be empty about the acknowledgment parameter:

```
[org.springframework.kafka.KafkaListenerEndpointContainer#0-0-C-1] SeekToCurrentErrorHandler - Backoff none exhausted for ConsumerRecord(topic = multilateral_matcher_results, partition = 0, leaderEpoch = 0, offset = 2, CreateTime = 1610624163545, serialized key size = -1, serialized value size = 76, headers = RecordHeaders(headers = [], isReadOnly = false), key = null, value = {"matchProcessId":"dc033ece-4e0e-4a60-86ce-7f8071678418","status":"SUCCESS"})
org.springframework.kafka.listener.ListenerExecutionFailedException: Listener method could not be invoked with the incoming message
Endpoint handler details:
Method [public void com.capitolis.matcher_client.kafka.consumer.MatcherReplyKafkaConsumer.processMessage(org.apache.kafka.clients.consumer.ConsumerRecord<java.lang.String, java.lang.String>,org.springframework.kafka.support.Acknowledgment)]
[...]

PR #1684, opened 2021-01-18
GH-1683: Assert the payload resolver does not resolve the Acknowledge…

selalerercapitolis:
…ment argument

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -74,6 +74,7 @@ import org.springframework.kafka.config.KafkaListenerEndpointRegistry;
 import org.springframework.kafka.config.MethodKafkaListenerEndpoint;
 import org.springframework.kafka.config.MultiMethodKafkaListenerEndpoint;
 import org.springframework.kafka.listener.KafkaListenerErrorHandler;
+import org.springframework.kafka.support.Acknowledgment;
 import org.springframework.kafka.support.KafkaNull;
 import org.springframework.kafka.support.TopicPartitionOffset;
 import org.springframework.lang.Nullable;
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@selalerercapitolis Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1684)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1684) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@selalerercapitolis Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1684)!

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -982,6 +989,21 @@ public Object resolveArgument(MethodParameter parameter, Message<?> message) thr
...
+			Class<?> parameterType = resolveTargetClass(parameter, message);
+			if (parameterType == Acknowledgment.class) {
+				String paramName = parameter.getParameterName();
+				paramName = (paramName != null ? paramName : "Arg " + parameter.getParameterIndex());
+				BindingResult bindingResult = new BeanPropertyBindingResult(message.getPayload(), paramName);
+				bindingResult.addError(new ObjectError(paramName,
+						"Parameter of type Acknowledgement should not be resolved by this resolver.\n" +
+						"Use Acknowledgement argument only if the listener is configured to use manual acknowledgment:\n" +
+						"\tspring.kafka.consumer.auto-offset-reset=earliest\n" +
+						"\tspring.kafka.listener.ack-mode=manual\n"));
```
artembilan, line 1002: First of all we don't say anything Spring Boot-specific in the code. This project is really can be used outside of any Spring Boot features.

On the other hand I don't see how this unconditional exception makes sense: you just don't check any provided `ackMode` in this context. Plus this `KafkaNullAwarePayloadArgumentResolver` is really used to replace a regular `PayloadArgumentResolver`, so still not clear why it must reject the `Acknowledgement` param and its argument?

What you are trying to achieve is really looks like a single configuration phase validation - not against every single consumer call... Probably something like `HandlerAdapter` or somewhere else in the `KafkaListenerEndpoint` hierarchy should be better place to check expected params and configured options before any real Kafka records processing.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
selalerercapitolis: Hi @artembilan , 

Thank you for taking the time to review this PR.
The details of the message may be inappropriate at this location but a validation that the payload resolver isn't trying to resolve an argument of type `Acknowledgement` does seem correct to me. If I understand the resolver purpose correctly, that shouldn't happen. Do you think I should completely remove this exception or just change the text?

I will try to look at the locations you suggested to add a validation that `Acknowledgement` argument is not used when `ackMode` is not manul and free up some time to write a proper validation there. 

Thank you, Sela

artembilan: Well, this resolver is not going to deal with `Acknowledgment` at all since it cannot be supplied in the `message.payload`. So, such a validation in this resolver is misleading. Plus we really don't talk here about actual message against particular Kafka record, but more about misconfiguration where non-manual ackMode cannot be used with `Acknowledgement` param. That's all: nothing more should be involved.

artembilan (changes requested):

Issue #1700, opened 2021-02-06
Default destination resolver in DeadLetterPublishingRecoverer issue when DLT partitions are less than original topic partitions


ecapoccia:
The default destination resolver in the DeadLetterPublishingRecoverer routes the records to a topic partition with the same name as the incoming consumer record + the suffix ".DLT", and the same partition as the incoming consumer record.

`DEFAULT_DESTINATION_RESOLVER = (cr, e) -> new TopicPartition(cr.topic() + ".DLT", cr.partition());`

This is a problem as this logic assumes that the number of partitions for the DLT is the same as the number of partitions of the original topic, which is not necessarily true.
In case the partitions for the DLT are less, this result in exceptions, as the destination partitions do not exist. Likewise in the (unlikely) case the DTL has more partitions than the original topic, some of them will be idle.

The solution I have successfully used is to set the destination partition to -1, which will delegate the choice of the destination partition within the DLT to Kafka.

`DEFAULT_DESTINATION_RESOLVER = (cr, e) -> new TopicPartition(cr.topic() + ".DLT", -1);`

Will submit a PR with a test to prove the case. I think this should be the default mechanism.

garyrussell:
As I said on the PR; we can't just change the default like that; we will make the documentation more visible,

garyrussell:
On reflection, I think we could inspect the DLT (if we have permissions) and log a WARN if the resolved partition doesn't exist.

That will satisfy your request, and prevent a breaking behavior change.

PR #1701, opened 2021-02-06
GH-1700 - changed partition number in default destination resolver to…

ecapoccia:
PR to fix issue described in GH-1700; default destination resolver for DL publisher should route to -1 (kafka to select the partition) rather than assuming the DLT has the same partitioning as the original topic.
In general can have less, as the DLT has far less traffic in normal conditions.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
@@ -55,7 +55,7 @@ public class DeadLetterPublishingRecoverer implements ConsumerRecordRecoverer {
 	protected final LogAccessor logger = new LogAccessor(LogFactory.getLog(getClass())); // NOSONAR
 
 	private static final BiFunction<ConsumerRecord<?, ?>, Exception, TopicPartition>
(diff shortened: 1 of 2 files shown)


ecapoccia:
@garyrussell @artembilan a really tiny PR, can you please have a look whenever is more convenient for you?
Many thanks, Emiliano

garyrussell:
The documentation clearly states, in the first paragraph...

>By default, the dead-letter record is sent to a topic named <originalTopic>.DLT (the original topic name suffixed with .DLT) and to the same partition as the original record. Therefore, when you use the default resolver, the dead-letter topic must have at least as many partitions as the original topic. I

We can't change the default because existing users might be relying on the current behavior.

We could, at most, highlight that documentation by tagging it with IMPORTANT.

garyrussell:
Thanks for the contribution, but we have to reject it for the reason stated.

ecapoccia:
Yes that makes sense @garyrussell , I appreciate changing a default is going to be an issue. Not a big deal. My intention was to save other people the headache we went through, of course we did not RTFM.
It is, if I can say, a bit against the general policy of spring's defaults. One expect the default in spring to "just work", but anyways, not a big deal.
One curiosity: the remote build failed, but building locally didn't, despite I tried multiple times. Is it something you have already seen?

garyrussell:
That one test is a bit flaky - doesn't fail often, but there's probably some race condition in the test itself.

PR #1706, opened 2021-02-10
GH-1700: Check Partition in DLPublishingRecoverer

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1700

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
@@ -20,14 +20,18 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.BiFunction;
 
 import org.apache.commons.logging.LogFactory;
+import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.header.internals.RecordHeader;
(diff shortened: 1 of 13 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -195,6 +232,30 @@ public void accept(ConsumerRecord<?, ?> record, Exception exception) {
...
+			boolean anyMatch = partitions.stream().anyMatch(pi -> pi.partition() == tp.partition());
+			if (!anyMatch) {
+				this.logger.warn(() -> "Destination resolver returned non-existent partition " + tp
+						+ ", KafkaProducer will determine partition to use for this topic");
+				return new TopicPartition(tp.topic(), -1);
+			}
+			return tp;
+		}
+		catch (Exception ex) {
+			this.logger.debug(ex, () -> "Could not obtain partition info for " + tp.topic());
```
artembilan, line 254: Just to double check: is it intentional to have it as `debug`  over here on this error and it is not a copy/paste artifact from the code above?
I mean returned wrong partition is a `warn`, but this exception is `debug` 🤷

spring-kafka/src/main/java/org/springframework/kafka/listener/RecoveryStrategy.java
```diff
@@ -35,10 +36,11 @@
...
 	 * @param ex the exception.
 	 * @param container the container (or parent if a child container).
+	 * @param consumer the consumer.
 	 * @return true to skip.
 	 * @throws InterruptedException if the thread is interrupted.
 	 */
-	boolean recovered(ConsumerRecord<?, ?> record, Exception ex, @Nullable MessageListenerContainer container)
-			throws InterruptedException;
+	boolean recovered(ConsumerRecord<?, ?> record, Exception ex, @Nullable MessageListenerContainer container,
+			@Nullable Consumer<?, ?> consumer) throws InterruptedException;
```
artembilan, line 44: Looks like a breaking change in the public API signature...

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
garyrussell: It's intentional - we might not have permissions and I didn't want to flood the logs.

spring-kafka/src/main/java/org/springframework/kafka/listener/RecoveryStrategy.java
garyrussell: It's not public; it's package-private.

garyrussell: Oops - sorry - no - this interface is new in 2.7.

artembilan (approved):

Issue #920, opened 2019-01-01
Topic-based retry support


henryjcee:
Retry support is (to my understanding) currently provided by Spring Retry. It would be nice to have built-in support for topic-based retries. See [here](https://eng.uber.com/reliable-reprocessing/) for more info on topic-based retries.

There is an [existing project](https://github.com/ZupIT/spring-kafka-retry) that has tried to do this but the approach could be simplified. Ideally, the `@KafkaListener` annotation could be configured parameterised to give full support with no other changes (assuming auto topic creation) e.g. 

`@KafkaListener(topics="test.topic", retries=5, backoff=Backoff.LINEAR, initialBackoff=60, dlq=true)`

Which would then autocreate the required topics (`test.topic.retry.60`, `test.topic.retry.120`, ..., `test.topic.dlq`) and create backoff-aware listeners for them (listeners that block until the retry period has expired, see the link above for more on this) that then call the annotated method with the backed-off message.

If there's support for this I can contribute.

garyrussell:
I am not sure what changed, but it works fine with version 2.2.2 (boot 2.1.1); I'll look into why it doesn't work with 2.1.x tomorrow.

henryjcee:
Sounds reasonable, I'll also take a look.

nromanetti:
Just a remark, this pattern covers only the cases where ordering is not critical as you may loose the ordering that kafka guarantees. Assume that a message N fails, it will go to the retry topic... if  message N+1 succeeds before message N is successfully retried your order is lost.

henryjcee:
Yes, it's essentially a way to achieve non-consumer group blocking retries at the expense of message ordering. It might seem a bit strange to tradeoff a key Kafka feature but I believe it's a widespread approach for people using Kafka for inter-service communication.

lucas-piske:
Any progress on this issue? I think this feature might be really useful and I would like to contribute.

henryjcee:
I started to have a look through the Spring retry source to see how this could be integrated but I quite quickly got out of my depth. Am still happy to contribute but I'll need to work alongside someone with Spring contribution experience.

garyrussell:
The framework has had partial support for this since version 2.2.

See [Publishing Dead-letter Records](https://docs.spring.io/spring-kafka/docs/2.2.7.RELEASE/reference/html/#dead-letters).

You would have to manually set up the consumer(s) for the retry queue(s), however.

```
[...]

PR #1707, opened 2021-02-16
GH-920 - Topic-based retry support - Improvements and new features

tomazfernandes:
- Added the ability of setting a global timeout for the retries
- Added the ability of choosing between suffixing the topics with it's delay or index
- Added isPartitionPaused to MessageListenerContainer and it's implementations
- Added option of not using a DLT
- Changed the Backoff Timestamp Header from String (formatted LocalDateTime) to long (millis since epoch UTC)
- Unwraps ListenerExecutionFailedException so that the cause can be classified
- Increased test coverage
- A few cleanups to the API

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
@@ -16,7 +16,7 @@
 
 package org.springframework.kafka.listener.adapter;
 
-import java.time.LocalDateTime;
+import java.math.BigInteger;
 import java.util.Optional;
 
 import org.apache.kafka.clients.consumer.Consumer;
@@ -69,19 +69,18 @@ public class KafkaBackoffAwareMessageListenerAdapter<K, V> extends AbstractDeleg
 	@Override
 	public void onMessage(ConsumerRecord<K, V> data, Acknowledgment ack, Consumer<?, ?> consumer) throws KafkaBackoffException {
 		maybeGetBackoffTimestamp(data)
-				.ifPresent(timestamp -> this.kafkaConsumerBackoffManager.maybeBackoff(createContext(data, timestamp)));
+				.ifPresent(nextExecutionTimestamp -> this.kafkaConsumerBackoffManager.maybeBackoff(createContext(data, nextExecutionTimestamp)));
 		super.getDelegate().onMessage(data, ack, consumer);
 		ack.acknowledge();
 	}
 
-	private KafkaConsumerBackoffManager.Context createContext(ConsumerRecord<K, V> data, LocalDateTime timestamp) {
-		return this.kafkaConsumerBackoffManager.createContext(timestamp, this.listenerId, new TopicPartition(data.topic(), data.partition()));
+	private KafkaConsumerBackoffManager.Context createContext(ConsumerRecord<K, V> data, long nextExecutionTimestamp) {
+		return this.kafkaConsumerBackoffManager.createContext(nextExecutionTimestamp, this.listenerId, new TopicPartition(data.topic(), data.partition()));
 	}
 
-	private <V, K> Optional<LocalDateTime> maybeGetBackoffTimestamp(ConsumerRecord<K, V> data) {
+	private <V, K> Optional<Long> maybeGetBackoffTimestamp(ConsumerRecord<K, V> data) {
 		return Optional
 				.ofNullable(data.headers().lastHeader(this.backoffTimestampHeader))
-				.map(timestampHeader -> new String(timestampHeader.value()))
-				.map(timestamp -> LocalDateTime.parse(timestamp, RetryTopicHeaders.DEFAULT_BACKOFF_TIMESTAMP_HEADER_FORMATTER));
+				.map(timestampHeader -> new BigInteger(timestampHeader.value()).longValue());
 	}
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
@@ -17,7 +17,7 @@
 package org.springframework.kafka.retrytopic;
 
 import java.math.BigInteger;
-import java.util.stream.StreamSupport;
+import java.util.function.Consumer;
 
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.producer.ProducerRecord;
@@ -45,11 +45,11 @@ import org.springframework.util.Assert;
  */
 public class DeadLetterPublishingRecovererFactory {
 
-	private static final CharSequence KAFKA_DLT_HEADERS_PREFIX = "kafka_dlt";
-
 	private static final String NO_OPS_RETRY_TOPIC = "internal-kafka-noOpsRetry";
 
 	private final DestinationTopicResolver destinationTopicResolver;
+
+	private Consumer<DeadLetterPublishingRecoverer> recovererCustomizer = recoverer -> { };
 
 	public DeadLetterPublishingRecovererFactory(DestinationTopicResolver destinationTopicResolver) {
 		this.destinationTopicResolver = destinationTopicResolver;
@@ -62,18 +62,25 @@ public class DeadLetterPublishingRecovererFactory {
 			@Override
 			protected void publish(ProducerRecord<Object, Object> outRecord, KafkaOperations<Object, Object> kafkaTemplate) {
 				if (NO_OPS_RETRY_TOPIC.equals(outRecord.topic())) {
-					this.logger.warn(() -> "Processing failed for dlt topic, giving up.");
+					this.logger.warn(() -> "Processing failed for last topic, giving up.");
 					return;
 				}
 
 				KafkaOperations<Object, Object> kafkaOperationsForTopic = (KafkaOperations<Object, Object>)
-						DeadLetterPublishingRecovererFactory.this.destinationTopicResolver.getKafkaOperationsFor(outRecord.topic());
+						DeadLetterPublishingRecovererFactory.this.destinationTopicResolver
+								.getCurrentTopic(outRecord.topic())
+								.getKafkaOperations();
 				super.publish(outRecord, kafkaOperationsForTopic);
 			}
 		};
 
 		recoverer.setHeadersFunction((consumerRecord, e) -> addHeaders(consumerRecord, e, getAttempts(consumerRecord)));
+		this.recovererCustomizer.accept(recoverer);
 		return recoverer;
+	}
+
+	public void setDeadLetterPublishingRecovererCustomizer(Consumer<DeadLetterPublishingRecoverer> customizer) {
+		this.recovererCustomizer = customizer;
 	}
 
 	private TopicPartition resolveDestination(ConsumerRecord<?, ?> cr, Exception e, Configuration configuration) {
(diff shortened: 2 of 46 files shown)


garyrussell:
@tomazfernandes Please rebase to master now that the main branch is merged; then I will try to review this afternoon (US EST).

tomazfernandes:
@garyrussell, I've rebased it, hope it's still in time. 

Also by doing a few more tests I found a bug where it wouldn't properly classify exceptions wrapped in ListenerExecutionFailedException, so if you decide not to merge this PR now maybe I could open a new PR only with this correction.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
```diff
@@ -104,24 +113,32 @@ private int getAttempts(ConsumerRecord<?, ?> consumerRecord) {
 	}
 
 	private Headers addHeaders(ConsumerRecord<?, ?> consumerRecord, Exception e, int attempts) {
-		Headers headers = filterPreviousDltHeaders(consumerRecord.headers());
+		// TODO: Integrate better with DeadLetterPublisherRecoverer so that headers end up better
+		// TODO: Maybe change all timestamps to long instead of String (String is better to read but worse for performance)
```
garyrussell, line 117: TODOs are useful, but creating a backlog issue for each is better.

This latter one would need to be done before 2.7 is released, or we'll have to handle both versions.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/destinationtopic/DestinationTopic.java
```diff
@@ -104,30 +104,35 @@ public int hashCode() {
 		return Objects.hash(this.destinationName, this.properties);
 	}
 
+	public long getDestinationTimeout() {
+		return this.properties.timeout;
+	}
+
 	public static class Properties {
 		private final long delayMs;
```
garyrussell, line 112: Space between fields - not critical for tomorrow.

garyrussell (commented):
A lot more good work; thanks!

Just a couple of comments - neither is critical for tomorrow.

There is one thing I missed in the last PR - our convention is to name test classes `...Tests` not the singular `...Test`, even when there is only one test method.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
tomazfernandes: Ok, I'll remove the todo's and write them on my notes so we can open the issues later. My main intention with those was letting you know of current shortcomings / decisions to be made.

That second TODO is actually outdated, I have changed all timestamps to long. I'll remove that.

tomazfernandes:
@garyrussell I'm really glad you liked the PR, thanks, means a lot to me. I can make these adjustments now if you think it's worth it.

garyrussell:
Yes, please; if you have time.

tomazfernandes:
@garyrussell I'll leave the TODOs here just to document them. I guess after it's out we can open the issues.

- Add support to MultiMethodEndpoint
- Enable class level annotation
- Maybe cache this result -> RetryTopicConfigurer
    -- this.method = Arrays.stream(ReflectionUtils.getDeclaredMethods(beanClass))...
- Integrate better with DeadLetterPublisherRecoverer so that headers end up better

Thanks a lot for your support and let me know if there's anything else needed.

garyrussell:
Oops - didn't notice this...

You should always give the first commit in a PR a headline (up to 50 chars).

Not a huge deal but it makes the commit log a bit messy...

```
* 777a9303288f8d4215359c30ca13f43412d583cc Skip Sonarqube for spring-kafka-docs
* 257bff50ada77933c56c0b94b62409d1de3b4151 GH-920: - Added the ability to set a global timeout for the retries - Added isPartitionPaused to MessageListenerContainer - Added new features to documentation - Added the ability to choose between suffixing the topics with the delay or it's index - Added the ability of not using a DLT - Changed the Backoff Timestamp Header from String (formatted LocalDateTime) to long (millis since epoch UTC) - Unwraps ListenerExecutionFailedException so the cause can be classified - A few cleanups in the API - Increased test coverage - Style adjustments
* 1a2f671c9c06587550ba5d25cd961506cf60413a GH 920 - Topic-based retry support (#1664)
```

tomazfernandes:
Hmmm, I see... Anything I can do about that now?

garyrussell:
No; we can't rewrite history once it is on master. I am sure it was a side effect of the rebase removing the previous commits. I should have noticed it before merging.

Never mind, I just wanted to give you a heads-up for any future PRs.

After I release M2 tomorrow, I am planning on writing a Blog post about the feature on spring.io; I'll give you a chance to review it before I publish it, if you have time.

Thanks again.

tomazfernandes:
Ok, noted. You may have noticed I don’t have that much experience with squashing / rebasing - I actually liked the result an plan on using more of it in my daily work.

About the blog post, sounds great! Sure I can review. Also feel free to reach out if there’s anything else I can help with.

garyrussell:
@tomazfernandes FYI, I had to do a bit of polishing: https://github.com/spring-projects/spring-kafka/commit/b2b22628d600bea3af198e2fd703c8432206e07d

You can see the SonarQube report here: https://sonar.spring.io/dashboard?id=org.springframework.kafka:spring-kafka-dist:master (until tonight when it will be overwritten).

We can't use port 9092 for the integration tests; it might not be available. We let the embedded kafka broker choose a random port instead.

tomazfernandes:
Looks great @garyrussell, thanks! Is there a way I could get this SonarQube report before we merged? Also, my personal machine's IntelliJ seems to not be warning me about unused variables, I'll check the configuration.

garyrussell:
@tomazfernandes We do have a test build plan for Sonar, which we usually run for large changes - I just forgot to do it this time.

For smaller changes, we generally wait for an email the next day with new code smells/bugs.

Here's the report against that plan with my changes: https://sonar.spring.io/dashboard?id=org.springframework.kafka%3Aspring-kafka-dist%3Asonar

tomazfernandes:
Looks a lot better :) 

I guess me committing lots of code the day before release didn't help much either... 

Seems it's alright now, thank you very much for reviewing the PR in such a short time!

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
```diff
@@ -69,19 +69,18 @@ public void onMessage(ConsumerRecord<K, V> data) {
 	@Override
 	public void onMessage(ConsumerRecord<K, V> data, Acknowledgment ack, Consumer<?, ?> consumer) throws KafkaBackoffException {
 		maybeGetBackoffTimestamp(data)
-				.ifPresent(timestamp -> this.kafkaConsumerBackoffManager.maybeBackoff(createContext(data, timestamp)));
+				.ifPresent(nextExecutionTimestamp -> this.kafkaConsumerBackoffManager.maybeBackoff(createContext(data, nextExecutionTimestamp)));
 		super.getDelegate().onMessage(data, ack, consumer);
 		ack.acknowledge();
```
garyrussell, line 74: @tomazfernandes I just noticed this - why are we acking here? Manual acks are the the responsibility of the user listener.

I didn't notice that you hard-wire the ack mode...

```
containerFactory.getContainerProperties().setAckMode(ContainerProperties.AckMode.MANUAL_IMMEDIATE);
```

In the configurer. What is the rationale here?

Thanks.

tomazfernandes: Hi @garyrussell, thanks for reaching out.

I needed to make sure the app committed the message's offset after a processing goes wrong and I publish it to another queue, so that there's no chance of processing the original message again. 

In my testing in the early days of this PR I remember having to set that AckMode.MANUAL_IMMEDIATE to be able to guarantee this happening properly, but maybe I missed out on something back then.

But now that you are calling that out, would that acking still work if the user sets a different AckMode and having only ackAfterHandle set to true in the error handler?

If so we might be able to remove this hardcoding and also the exception from the other onMessage method.

Thanks

tomazfernandes: Also, I'm remembering better now, I think back then I might have had a wrong understanding that the offset needed to be committed on the broker for the message not to be consumed again, when actually the consumer can handle it locally before the actual ack goes to the server.

tomazfernandes: Does that make sense? Also, is it something you’ll look into or do you prefer that I do? I’m fine with both.

garyrussell: I have to go to a funeral tomorrow (Friday), so I won't be here.

I was more concerned about a possible NPE, but your hard-coding avoids that (I was working on configuring `@Nullable` for Kotlin compatibility and noticed that the ack must not be null - see the latest commit on master).

I think we should allow the ackmode to be configured, but recommend the use of RECORD in this scenario.

Furthermore, I think we need to modify the DLPR to (at least optionally) wait on the send `Future` to make sure it was successful before acking the delivery (that's just a general comment, not related to this PR, but since we instantiate a DLPR in the DLPRF, we would need to set it somehow.

If you have time to explore this while I am away, that would be great.

tomazfernandes:
@garyrussell, I'm really sorry for your loss.

Ok, I'm actually working on this now. The ack there is really meaningless so I removed it.

With this I could change the KafkaBackoffAwareListenerAdapter to implement the MessageListener interface instead of the Acknowledging one, with the four MessageListener onMessage methods implemented. Gives more flexibility for any other use cases.

I'll let the ack mode for the user to set and update the docs to recommend the RECORD one, and also look into the DLPR change.

If there's anything else you'd like to add please do, I should have a PR open with these changes by tomorrow.

Thanks.

garyrussell:
>With this I could change the KafkaBackoffAwareListenerAdapter to implement the MessageListener interface instead of the Acknowledging one,

Actually, no; all adapters have to implement `AcknowledgingConsumerAwareMessageListener` and `AbstractDelegatingMessageListenerAdapter` so the listener container can determine the type of the ultimate listener, so it can decided which arguments to populate.

See the other adapters, e.g. the `FilteringMessageListenerAdapter`.

tomazfernandes:
Hmm, ok. There's definitely something I don't understand here, because the ListenerConsumer already checks the delegate's type for choosing which onMessage method to call. So from that logic I don't see why not using the adapter more like a proxy, calling the delegate in whichever method it has been called upon. Maybe that's so not to rely on the callers logic? If you could share the reasoning behind this it would be great.

Nevertheless I'll implement with the switch statement as the others adapters have.

Thanks.

tomazfernandes:
Also, one thing I’ve been exploring is creating a superclass for the DeadLetterPublishingRecoverer, a MessagePublishingRecoverer. It contains pretty much all the DLPR logic, but without the DLT references, which are provided by the subclass.

With this and a few small tweaks I can integrate the retry logic a lot better, and have the message headers make more sense to the user. Right now the DLT headers are overriden in every retry, so “original” headers lose their meaning, besides having “dlt” in message headers sent to the retry topics.

As a subclass, nothing changes in the DLPR api or it’s functionality. But I think it makes up for a more integrated and cohesive solution.

How does that sound?

tomazfernandes:
For the switch part I mean this logic here in KafkaMessageListenerContainer:

Captura de Tela 2021-02-21 às 04 24 01

garyrussell:
The container determines the type of the final delegate (user's listener) and populates the arguments appropriately (specifically the consumer and acknowledgment).

If the adapter only implements `MessageListener`, it will be impossible to pass those arguments to the ultimate listener if it is consumer aware and/or acknowledging.

I am not sure why you need to change the DLPR; you can simply wire in a custom destination resolver. I'll look at your PR later.

Issue #1708, opened 2021-02-17
Fix New Package Tangles


garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1708

<img width="382" alt="Screen Shot 2021-02-17 at 3 16 10 PM" src="https://user-images.githubusercontent.com/483832/108262573-184e7d80-7133-11eb-8793-e0f7630503c5.png">
<img width="167" alt="Screen Shot 2021-02-17 at 3 16 00 PM" src="https://user-images.githubusercontent.com/483832/108262576-18e71400-7133-11eb-8845-03ca29a830f2.png">

- add a new interface `ListenerContainerRegistry` in the `listener` package
- eliminate the `destinationtopic` sub-package
- move classes with references to the `annotation` package to `annotation`.
- add `package-info.java`

PR #1709, opened 2021-02-17
GH-1708: Fix New Package Tangles

garyrussell:
Screen Shot 2021-02-17 at 3 16 10 PM
Screen Shot 2021-02-17 at 3 16 00 PM

- add a new interface `ListenerContainerRegistry` in the `listener` package
- eliminate the `destinationtopic` sub-pavkage
- move classes with references to the `annotation` package to `annotation`.
- add `package-info.java`


garyrussell:
@tomazfernandes FYI, we use Structure101 to ensure we don't have bi-directional references between classes and packages.

tomazfernandes:
No problem @garyrussell, looks good. I had left a few package-private methods in the RetryTopicConfigurationBuilder class to clean up the API for the users, they are only necessary for annotation processing. But I guess that's not a problem.

Nice seeing everything getting into shape, thanks!

PR #1711, opened 2021-02-22
GH-920: Improvements in acknowledging and DeadLetterPublishingRecover…

tomazfernandes:
Removed the hardcoded AckMode and the compulsory ack in the onMessage method.

Extracted a method in the KafkaListenerAnottationBPP to decouple annotation processing from endpoint registration. With this I could refactor the RetryTopicConfigurer logic to be a lot more straightforward and easier to understand and maintain.

Made some changes to the DeadLetterPublishingRecoverer to better accommodate the retry topic logic. Now it:
- Can accept a template resolver so that the template can be resolved by some other criteria than the value class, for example the topic it will be sent to. Also now it has a getHeaderNames function that can be overridden by a subclass to use a different set of headers, for example without DLT references for the retry topic logic.
- Added the ability to not replace (it's not exactly replacing, but using the lastHeader method in the end it is) the ORIGINAL headers, so that when forwarding the message through the retry topics the ORIGINAL topic for example remains the first one, not the one from the previous retry.
- Added the option to throw or end processing if the destination resolver returns a null value. Previously it would throw a NPE.
- Added the option of throwing an exception if the message sending fails.
- No changes in functionality were made in the current API, everything works as before with these new features added.

Implemented tests for all these functions and changes. 
Updated the docs. Didn't add everything new in DLPR to the docs because it seemed it would get bloated.


tomazfernandes:
Hi @garyrussell, I made those changes we talked about and a few more that were in my mind's TODO list.

The entry point for the RetryTopicConfigurer logic was too heavy on indirections with all those consumers and customisers, and that made that logic much harder to understand and maintain. With a small change in the KLABPP I could make it a lot more straightforward.

Also, I made some changes to the DLPR to better accommodate the retry topic logic.

I think these changes added clarity and flexibility to those logics, but if you're not ok with any of these changes please let me know.

What I failed to do is to implement that switch statement in the Adapter. I really didn't understand the point and that makes it harder to know if it's doing what it's supposed to do or not. If you can explain better the reasoning for this (as we were talking in the other PR's discussion) I'd appreciate that. But I did change the interface to AcknowledgingConsumerAwareListenerAdapter.

Let me know if there's anything I should change and how the PR looks to you.

Thanks.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -335,39 +434,68 @@ protected void publish(ProducerRecord<Object, Object> outRecord, KafkaOperations
...
+
+	private void verifySendResult(ProducerRecord<Object, Object> outRecord,
+								@Nullable ListenableFuture<SendResult<Object, Object>> sendResult) {
+		if (sendResult == null) {
+			throw new KafkaException("Dead-letter publication failed for: " + outRecord);
+		}
+		try {
+			sendResult.get(this.waitForSendResultTimeout, TimeUnit.MILLISECONDS);
+		}
+		catch (InterruptedException | ExecutionException | TimeoutException e) {
```
garyrussell, line 450: Interrupt needs to be in a separate catch class with `Thread.currentThread.interrupt()` before throwing; otherwise we "eat" the interrupt and some downstream interruptible will block.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -378,4 +506,84 @@ private String getStackTraceAsString(Throwable cause) {
 		return stringWriter.getBuffer().toString();
 	}
 
+	/**
+	 * Override this if you want different header names to be used
+	 * in the sent record.
+	 * @return the default Dlt header names
+	 */
+	protected HeaderNames getHeaderNames() {
+		return DeadLetterPublishingHeaderNamesBuilder
```
garyrussell, line 515: We should make this a constant `DEFAULT_HEADERNAMES` to avoid building each time.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -378,4 +506,84 @@ private String getStackTraceAsString(Throwable cause) {
...
+					.exceptionMessage(KafkaHeaders.DLT_EXCEPTION_MESSAGE)
+					.keyExceptionStacktrace(KafkaHeaders.DLT_KEY_EXCEPTION_STACKTRACE)
+					.exceptionStacktrace(KafkaHeaders.DLT_EXCEPTION_STACKTRACE)
+				.build();
+	}
+
+	/**
+	 * Container class for the headers that will be added to the produced record.
+	 */
+	public static class HeaderNames {
```
garyrussell, line 535: This should be moved to the builder, or the builder moved here - otherwise we have a class tangle with each referencing each other.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -378,4 +506,84 @@ private String getStackTraceAsString(Throwable cause) {
...
+
+		private final HeaderNames.Original original;
+		private final HeaderNames.Exception exception;
+
+		HeaderNames(HeaderNames.Original original, HeaderNames.Exception exception) {
+			this.original = original;
+			this.exception = exception;
+		}
+
+		static class Original {
```
garyrussell, line 545: Why do we have these classes here and in the builder as well?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
```diff
@@ -64,15 +64,32 @@ public KafkaBackoffAwareMessageListenerAdapter(AcknowledgingConsumerAwareMessage
 	}
 
 	public KafkaBackoffAwareMessageListenerAdapter(AcknowledgingConsumerAwareMessageListener<K, V> adapter,
```
garyrussell, line 66: Ah! Now I see the disconnect about the switch - the other adapters filter, retry, etc can be used with non-POJO listeners (of any type), not just with `@KafkaListener` infrastructure.

In this case we are requiring the delegate to be an ACAML so we don't need the switch.

Do you think it would never be the case that someone would want to wrap their own `MessageListener` in this adapter? I suppose that's reasonable, but we should clearly document it.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
```diff
@@ -64,15 +64,32 @@ public KafkaBackoffAwareMessageListenerAdapter(AcknowledgingConsumerAwareMessage
...
+				.maybeBackoff(createContext(data, nextExecutionTimestamp)));
+		super.getDelegate().onMessage(data, acknowledgment);
+	}
+
+	@Override
+	public void onMessage(ConsumerRecord<K, V> data, Consumer<?, ?> consumer) throws KafkaBackoffException {
+		maybeGetBackoffTimestamp(data)
+				.ifPresent(nextExecutionTimestamp -> this.kafkaConsumerBackoffManager
+				.maybeBackoff(createContext(data, nextExecutionTimestamp)));
+		super.getDelegate().onMessage(data, consumer);
```
garyrussell, line 92: Given my comment above, these are not needed.

garyrussell (changes requested):
Wow; you sure can pump out a lot of code in a short amount of time 😄 

Please re-push the first commit with a headline <= 50 chars.

https://github.com/spring-projects/spring-kafka/blob/master/CONTRIBUTING.adoc#a-day-in-the-life-of-a-contributor

and

https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project

Then add any new changes as new commits.

Thanks.

tomazfernandes:
Hi @garyrussell, yeah, I guess I can get a little intense sometimes 😄. Specially doing this kind of work, which is the one I like the most. I'll answer some of your comments and work on them later today. Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingHeaderNamesBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+public class DeadLetterPublishingHeaderNamesBuilder {
+
+	private final Original original = new Original();
+	private final Exception exception = new Exception();
+
+	public static DeadLetterPublishingHeaderNamesBuilder.Original original() {
+		return new DeadLetterPublishingHeaderNamesBuilder().original;
+	}
+
+	public class Original {
```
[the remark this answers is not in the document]
tomazfernandes: Ok for the javadocs. As for the static class, I actually rely on this class not being static so that it can access the Builder's properties. This way I can provide a cleaner API without having to create too much of an infrastructure. So if I change this to static I'll probably have to change to a probably less clean API.

DYT it's a problem to leave it as it is?

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingHeaderNamesBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+			Assert.notNull(this.partitionHeader, "partitionHeader header cannot be null");
+			return new DeadLetterPublishingRecoverer.HeaderNames.Original(this.offsetHeader,
+					this.timestampHeader,
+					this.timestampTypeHeader,
+					this.topicHeader,
+					this.partitionHeader);
+		}
+	}
+
+	public class Exception {
```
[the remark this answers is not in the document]
tomazfernandes: Same as above

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
tomazfernandes: Ok. I think it's probably better to leave in the DLPR so I can access it's properties in a simpler way, if that's ok with you.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
tomazfernandes: It's basically to keep the properties more organised, instead of a single class with 11 properties, and also provide a more organised API. But maybe I can make it simpler if you prefer.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
tomazfernandes: Hmmm, I actually didn't give much thought to uses besides with the annotation - I myself have never used it in another way. I'll look into it. I think I still don't fully understand the question though, because the MessageListener interface also has all 4 onMessage methods, so I guess the user should be aware of its own MessageListener implementation and handle all 4 methods properly, according to the intended result. I'll give more thought on that later.

But do I think the user should have the flexibility to use this adapter just as any of the others.

tomazfernandes:
Ok, I commented the ones I had doubts, the ones I didn't comment I'm ok with and will make the adjustments later. Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
garyrussell: The hierarchy of interfaces is designed so that they are all `@FunctionalInterface`s, allowing them to be used in lambda expressions:

```java
properties.setMessageListner((MessageListener<String, String) record -> { ... });

...

properties.setMessageListner((ConsumerAwareMessageListener<String, String) (record, consumer) -> { ... });
```

etc.

So, no, users don't have to implement all 4 methods;

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingHeaderNamesBuilder.java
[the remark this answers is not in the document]
garyrussell: No; it's not a problem, but it is confusing to have these effectively duplicated in the DLPR.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
garyrussell: I understand original Vs. exception, my concern is duplication of similar classes in the builder.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingHeaderNamesBuilder.java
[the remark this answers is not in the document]
tomazfernandes: Yeah, I suppose it is. The main difference is that in the builder the fields are not final and it has the not null assertions in the build methods, while the HeaderNames class only has the final fields (no getters or setters or anything else).

I can try to come up with something less confusing, I'll look into that.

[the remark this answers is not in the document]
garyrussell: Don't spend too much time; I can live with it now that I know the rationale.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
tomazfernandes: Oh, I see, everything makes sense now 🙏🏼. Thanks a lot for your patience!

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
tomazfernandes: Actually, looking again, the Builder has a reference to the HeaderNames class, but the HeaderNames class doesn't know about the builder.

tomazfernandes: The DLPR knows about the Builder and the Builder knows about DLPR though. I've moved the Builder to the DLPR.

tomazfernandes:
Done as requested sir @garyrussell, hope it's to your liking! Please let me know if there's anything to be changed.

I'll owe you for now the docs for the KafkaBackoffAwareListenerAdapter and KafkaBackoffManager. I think they do have potentially interesting use cases outside the retry topic scope, and I plan to set up some examples to explore them, and with that I can document it properly. Hopefully we can live without that for now.

I'd like to know a bit more of the timeline for the releases, if you could please share with me. Considering you approve this PR, will it be in a M3 release? When would that be out? Also, is there a release date for 2.7.0-RELEASE?

Thanks a lot for our support!

garyrussell:
@tomazfernandes The next release is the release candidate March 17 - you can always check the milestones for scheduled releases.

https://github.com/spring-projects/spring-kafka/milestones

We generally don't add new functionality between the release candidate and GA (currently April 14) - just docs and bug fixes - but due to the magnitude of this feature, I am inclined to leave it marked 'experimental', even at GA, allowing us to make API changes if necessary.

tomazfernandes:
Sounds good @garyrussell, let me know if there’s anything else that needs to be done.

Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
garyrussell: Unresolved.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -79,6 +87,14 @@
 
 	private Duration partitionInfoTimeout = Duration.ofSeconds(FIVE);
 
+	private boolean replaceOriginalHeaders = true;
+
+	private boolean failIfSendResultIsError = false;
+
+	private boolean throwIfNoDestinationReturned = false;
+
+	private long waitForSendResultTimeout = 30000L;
```
garyrussell, line 96: Magic number  (see `partitionInfoTimeout` above) - also, we prefer `Duration` for these timeouts; see `setPartitionInfoTimeout()`.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -79,6 +87,14 @@
 
 	private Duration partitionInfoTimeout = Duration.ofSeconds(FIVE);
 
+	private boolean replaceOriginalHeaders = true;
+
+	private boolean failIfSendResultIsError = false;
+
+	private boolean throwIfNoDestinationReturned = false;
```
garyrussell, line 94: I think both of these should be true by default.

spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -5282,6 +5282,9 @@ A `LinkedHashMap` is recommended so that the keys are examined in order.
 
 When publishing `null` values, when there are multiple templates, the recoverer will look for a template for the `Void` class; if none is present, the first template from the `values().iterator()` will be used.
 
+Since 2.7 you can use the `setFailIfSendResultIsError` method so that an exception is thrown when message publishing fails.
+You can also set a timeout for the verification of the sender success with `setWaitForSendResultTimeout`.
```
garyrussell, line 5286: I think this should be true by default.

garyrussell (changes requested):
@tomazfernandes When I noticed the magic number, I ran the Sonar report against this branch; when I looked at this last time, I wondered if Sonar would complain about the classes named `Exception` 😄  Maybe change them to something like `ExceptionInfo`?

Full report here: https://sonar.spring.io/project/issues?id=org.springframework.kafka%3Aspring-kafka-dist%3Asonar&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
tomazfernandes: Yeah, me too, the get method from the Future interface takes a long. I'll do the conversion.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
tomazfernandes: Ok. By doing so, many tests break in the DLPR tests because the SettableListenableFuture times out on get, as it has no value. I'll assign some dummy value so it works again.

tomazfernandes:
@garyrussell, I think I fixed all the issues now. Sorry for the Builder, I had actually coded the change but it didn't get committed.

As I said in another comment, the DLPR tests failed when I set the default behavior to be checking the send result because they'd timeout as the Future is empty, so I added a new Object() to the SettableFutures. I could also have just turned the feature off for the tests, but since this is now the default behavior I think it was best to add it to them.

There's one test however (KafkaTemplateTransactionTests.testDeadLetterPublisherWhileTransactionActive) that I couldn't  do that because it uses a spy and I couldn't mock it's send method, so for that test I disabled the feature. If that raises a flag let me know, but it seemed like something related to the way the test works rather than the feature itself.

Please let me know if there's anything else to be changed.

Thanks!

garyrussell:
Suggested fixes for [remaining Sonar smells](https://sonar.spring.io/project/issues?id=org.springframework.kafka%3Aspring-kafka-dist%3Asonar&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL).

https://github.com/garyrussell/spring-kafka/commit/33deaab4f81e9fb6ef92d0d1179333ee38a3e71d

garyrussell (approved):

garyrussell:
Merged as 93a2fa1a3b1434ff90bff305be5cae02c633ced2

Issue #1715, opened 2021-02-25
@RetryableTopic Fails with "Another endpoint is already registered with id"


garyrussell:
```java
@SpringBootApplication
public class Kgh920Application {

	public static void main(String[] args) {
		SpringApplication.run(Kgh920Application.class, args);
	}

	@RetryableTopic(attempts = 5, backoff = @Backoff(delay = 1000, multiplier = 2.0), kafkaTemplate = "kafkaTemplate")
	@KafkaListener(id = "kgh920", topics = "kgh920", containerFactory = "kafkaListenerContainerFactory")
	public void listen(String in, @Header(KafkaHeaders.RECEIVED_TOPIC) String topic) {
		System.out.println(in + " from " + topic);
		throw new RuntimeException("test");
	}

	@DltHandler
	public void dlt(String in, @Header(KafkaHeaders.RECEIVED_TOPIC) String topic) {
		System.out.println(in + " from " + topic);
	}

}
```
```properties
spring.kafka.consumer.auto-offset-reset=earliest
```
```
2021-02-25 14:46:38.078 ERROR 66259 --- [           main] o.s.boot.SpringApplication               : Application run failed

java.lang.IllegalStateException: Another endpoint is already registered with id 'kgh920'
	at org.springframework.util.Assert.state(Assert.java:76) ~[spring-core-5.3.4.jar:5.3.4]
	at org.springframework.kafka.config.KafkaListenerEndpointRegistry.registerListenerContainer(KafkaListenerEndpointRegistry.java:177) ~[spring-kafka-2.7.0-SNAPSHOT.jar:2.7.0-SNAPSHOT]
	at org.springframework.kafka.config.KafkaListenerEndpointRegistry.registerListenerContainer(KafkaListenerEndpointRegistry.java:152) ~[spring-kafka-2.7.0-SNAPSHOT.jar:2.7.0-SNAPSHOT]
	at org.springframework.kafka.config.KafkaListenerEndpointRegistrar.registerAllEndpoints(KafkaListenerEndpointRegistrar.java:196) ~[spring-kafka-2.7.0-SNAPSHOT.jar:2.7.0-SNAPSHOT]
	at org.springframework.kafka.config.KafkaListenerEndpointRegistrar.afterPropertiesSet(KafkaListenerEndpointRegistrar.java:186) ~[spring-kafka-2.7.0-SNAPSHOT.jar:2.7.0-SNAPSHOT]
[...]

PR #1719, opened 2021-02-26
Fixes GH-1715.

tomazfernandes:
Fixed the bug and added the id parameter in two of the integration tests.


tomazfernandes:
Will have too many conflicts with the other issue, better to do it in one PR.

PR #1723, opened 2021-02-26
GH-920: Fix Class Tangle

garyrussell:
Screen Shot 2021-02-26 at 2 15 48 PM

- move `EndpointHandlerMethod` to its own file
- remove backward reference from `...Configuration` to its builder
- fix docs
- 
cc/ @tomazfernandes


tomazfernandes:
PR looks good. I don't remember if I had asked specifically about this before, sorry if I did, but do you know a way I can see those tangles locally? I tried using IntelliJ's analysis but it's nowhere as clear as this one.

Issue #1716, opened 2021-02-25
@RetryableTopic Delays Between Attempts Longer Than Expected


garyrussell:
```java
@SpringBootApplication
public class Kgh920Application {


	private static final Logger LOG = LoggerFactory.getLogger(Kgh920Application.class);


	public static void main(String[] args) {
		SpringApplication.run(Kgh920Application.class, args);
	}

	@RetryableTopic(attempts = 5, backoff = @Backoff(delay = 1000, multiplier = 2.0), kafkaTemplate = "kafkaTemplate")
	@KafkaListener(groupId = "kgh920", topics = "kgh920", containerFactory = "kafkaListenerContainerFactory")
	public void listen(String in, @Header(KafkaHeaders.RECEIVED_TOPIC) String topic) {
		LOG.info(in + " from " + topic);
		throw new RuntimeException("test");
	}

	@DltHandler
	public void dlt(String in, @Header(KafkaHeaders.RECEIVED_TOPIC) String topic) {
		LOG.info(in + " from " + topic);
	}

}
```
```log
$ grep 'foo from' ../tmp/retry.log
2021-02-25 15:11:15.639  INFO 69371 --- [ntainer#0-0-C-1] com.example.demo.Kgh920Application       : foo from kgh920
2021-02-25 15:11:26.196  INFO 69371 --- [etry-1000-0-C-1] com.example.demo.Kgh920Application       : foo from kgh920-retry-1000
2021-02-25 15:11:36.709  INFO 69371 --- [etry-2000-0-C-1] com.example.demo.Kgh920Application       : foo from kgh920-retry-2000
2021-02-25 15:11:47.221  INFO 69371 --- [etry-4000-0-C-1] com.example.demo.Kgh920Application       : foo from kgh920-retry-4000
2021-02-25 15:12:02.739  INFO 69371 --- [etry-8000-0-C-1] com.example.demo.Kgh920Application       : foo from kgh920-retry-8000
2021-02-25 15:12:03.245  INFO 69371 --- [ner#6-dlt-0-C-1] com.example.demo.Kgh920Application       : foo from kgh920-dlt
```

+10s, +10s, +10s, +15s

Perhaps something to do with the default poll timeout of 5 seconds, possibly affecting resume latency.

cc/ @tomazfernandes

garyrussell:
Yes, somewhat improved with

    spring.kafka.listener.poll-timeout=1s

garyrussell:
[...]

PR #1730, opened 2021-03-11
Fixes GH-1716. [NOT READY FOR MERGE]

tomazfernandes:
@garyrussell, this is the WIP of my work on this issue.

There were three main sources of imprecision:

- the time it takes from the moment the listener throws the exception to the moment the next attempt's timestamp was calculated (around 500ms in my machine);
- we have to wait for 2 polls after the Backoff manager sends the resume command, due to the order things happen in the listener consumer;
- for a consumer that handles only one partition we'll always wait in `pollTimeout` increments, so if the timing is off it would wait until after the timestamp to actually process the message;

For the timing correction I used your wakeup suggestion - I used the default boot's executor for that, mainly because since the RT's beans are not registered at application startup I can't use some Spring goodies such as a destroy method to shutdown the executor if I created one.

If you have a better suggestion please tell me, using the same executor that is offered to the user's application doesn't seem like the best solution. Maybe register RT beans on app startup, or maybe there's another way to register the executor shutdown hook?

Also, the BackOffException exception was still printing the stack trace - I wasn't sure of the best place to handle that - I did it in the ListenerConsumer.

Please let me know what you think and any suggestions you might have.

Thanks!

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
@@ -46,7 +46,7 @@ import org.springframework.kafka.support.KafkaHeaders;
  */
 public class DeadLetterPublishingRecovererFactory {
 
-	private static final LogAccessor logger = new LogAccessor(LogFactory.getLog(RetryTopicConfigurer.class));
+	private static final LogAccessor logger = new LogAccessor(LogFactory.getLog(DeadLetterPublishingRecovererFactory.class));
 
 	private final DestinationTopicResolver destinationTopicResolver;
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
@@ -19,6 +19,7 @@ package org.springframework.kafka.retrytopic;
 import java.math.BigInteger;
 import java.util.function.Consumer;
 
+import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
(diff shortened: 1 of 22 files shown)


spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
```diff
@@ -43,6 +46,8 @@
  */
 public class DeadLetterPublishingRecovererFactory {
 
+	private static final LogAccessor logger = new LogAccessor(LogFactory.getLog(RetryTopicConfigurer.class));
```
artembilan, line 49: Copy/paste artefact: wrong class is used for Log instance 😄

artembilan (changes requested):
Just my five fingers.
Mostly it is hard for me to review: first of all new stuff, secondly you did a lot of refactoring to existing code which is not related to the premise of this PR, I guess. But I promise: I'll catch up with you on this one day 😄

tomazfernandes:
> Just my five fingers.
> Mostly it is hard for me to review: first of all new stuff, secondly you did a lot of refactoring to existing code which is not related to the premise of this PR, I guess. But I promise: I'll catch up with you on this one day 😄

Hi @artembilan - welcome to the retry topic madness! Just kidding - it is indeed a lot of new code 😄. If there's anything I can help you with when you feel like catching up with this just let me know.

And you're right, I guess I could have made a clearer cut on the issue - I'd say 90%+ of the changes are for this issue, but there were other fixes that helped me debug the feature that could have gone into different commits / PR.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
tomazfernandes: Thanks for noticing - there should be a plugin that threw a CopyPasteException in these situations! :lay

garyrussell:
Replaced by https://github.com/spring-projects/spring-kafka/pull/1734

PR #1733, opened 2021-03-13
let pass topic to sendMethod of Message

woblak:
Allow to send Message<?> to chosen topic without setting topic in header or as defaultTopic of KafkaTemplate.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -392,7 +392,12 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V>, ApplicationCo
 	@SuppressWarnings("unchecked")
 	@Override
 	public ListenableFuture<SendResult<K, V>> send(Message<?> message) {
-		ProducerRecord<?, ?> producerRecord = this.messageConverter.fromMessage(message, this.defaultTopic);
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -392,7 +392,12 @@ public void onApplicationEvent(ContextStoppedEvent event) {
 	@SuppressWarnings("unchecked")
 	@Override
 	public ListenableFuture<SendResult<K, V>> send(Message<?> message) {
-		ProducerRecord<?, ?> producerRecord = this.messageConverter.fromMessage(message, this.defaultTopic);
+		return send(this.defaultTopic, message);
+	}
+
+	@SuppressWarnings("unchecked")
+	public ListenableFuture<SendResult<K, V>> send(String defaultTopic, Message<?> message) {
```
artembilan, line 399: Since the `Message` can have a `KafkaHeaders.TOPIC`, your argument is going to be ignored and you may be confused "why record has gone to a different topic". That was the main reason do not introduce a contract like this: to avoid end-user confusion and let them to make a decision only via single source of truth - that `KafkaHeaders.TOPIC`. 
Same applies for the `send(ProducerRecord<K, V>)` contract.

So, -1 from me for this fix.

artembilan (changes requested):

PR #1734, opened 2021-03-14
Gh 1716 - Refactoring and Test Coverage

tomazfernandes:
This PR contains everything the other GH-1716 PR does - I chose to open a separate PR because the mechanics of the solution should be easier to reason about in the first one. This one mainly adds some refactoring of the DestinationTopic interfaces, expanded test coverage, besides the documentation referenced in the issue.

Let me know if there's anything else I can do or should change.

Thanks!


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaConsumerBackoffManager.java
```diff
@@ -47,87 +53,162 @@
  */
 public class KafkaConsumerBackoffManager implements ApplicationListener<ListenerContainerPartitionIdleEvent> {
 
+	private static final LogAccessor logger = new LogAccessor(LogFactory.getLog(RetryTopicConfigurer.class));
```
artembilan, line 56: Still copy/paste flaw...

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaConsumerBackoffManager.java
```diff
@@ -47,87 +53,162 @@
...
+				context.topicPartition.topic(), backOffTime),
 				topicPartition, context.listenerId, context.dueTimestamp);
 	}
 
 	@Override
 	public void onApplicationEvent(ListenerContainerPartitionIdleEvent partitionIdleEvent) {
-		Context context = getBackoff(partitionIdleEvent.getTopicPartition());
-		if (context == null || isNotDue(context.dueTimestamp)) {
+		logger.debug(() -> String.format("partitionIdleEvent received in %s at %s. Partition: %s",
+				this.getClass().getSimpleName(), getCurrentMillisFromClock(), partitionIdleEvent.getTopicPartition()));
```
artembilan, line 113: Why do we need a class name in the logs? 
It could be derivatived from the log category, plus thread and sometime instance.
More over you use simple class name - no package. It is going confuse end-users.
I would just go without class at all...

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaConsumerBackoffManager.java
```diff
@@ -47,87 +53,162 @@
...
 
 	private final Clock clock;
 
+	private final TaskExecutor taskExecutor;
+
+	private final Sleeper sleeper;
+
 	public KafkaConsumerBackoffManager(ListenerContainerRegistry registry,
-									@Qualifier(INTERNAL_BACKOFF_CLOCK_BEAN_NAME) Clock clock) {
+									@Qualifier(INTERNAL_BACKOFF_CLOCK_BEAN_NAME) Clock clock,
```
artembilan, line 77: Since we develop here a library, not the target project, it is better to stay away from auto-wiring in this type of code.
Why do we need `Clock` at all? Why can't we just use a `System.nanoTime()`?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaConsumerBackoffManager.java
```diff
@@ -47,87 +53,162 @@
...
-		synchronized (this.backOffTimes) {
-			this.backOffTimes.remove(topicPartition);
+		synchronized (this.backOffContexts) {
+			this.backOffContexts.remove(topicPartition);
 		}
 	}
 
-	public Context createContext(long dueTimestamp, String listenerId, TopicPartition topicPartition) {
-		return new Context(dueTimestamp, listenerId, topicPartition);
+	public Context createContext(long dueTimestamp, String listenerId, TopicPartition topicPartition, Consumer<?, ?> consumer) {
```
artembilan, line 210: Can we name that `Consumer` parameter more explicit way?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1491,7 +1495,7 @@ private void resumePartitionsIfNecessary() {
 			if (partitionsToResume.size() > 0) {
 				this.consumer.resume(partitionsToResume);
 				this.pausedPartitions.removeAll(partitionsToResume);
-				this.logger.debug(() -> "Resumed consumption from " + partitionsToResume);
+				this.logger.info(() -> "Resumed consumption from " + partitionsToResume);
```
artembilan, line 1498: Why back to `INFO`?

artembilan (changes requested):
Just a little review. Although it raises more questions than just plain typos.
I think I'll stop for now because we need to fix other things for upcoming releases this week.
Will come back with my doubts to you later 😄

tomazfernandes:
The test that failed: EnableKafkaIntegrationTests > testSimple() FAILED -  org.opentest4j.AssertionFailedError at EnableKafkaIntegrationTests.java:276 - passes locally. I'll take a look into this later today.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaConsumerBackoffManager.java
tomazfernandes: This is a trickier one, and one I had my doubts too. The main purpose of injecting the clock is for testing purposes, so we can inject a fixed time clock - easier to hit than a moving target. Same reason to use a Sleeper instead of just Thread sleeping.

I had doubts about the qualifier - maybe we could just inject whatever clock that app uses, if any - and create an instance if none as . The only requirement of course is that it has to be the same clock throughout the backoff process so that it keeps in sync.

That being said, do you think it's better to just use System.millis (or nanos)? I can look into that if so.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaConsumerBackoffManager.java
tomazfernandes: I renamed it to consumerForTimingCorrection - not sure I like it better though 😄. Let me know if you have any suggestions.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaConsumerBackoffManager.java
tomazfernandes: I think the same kind of reasoning applies to the injected TaskExecutor, which I was looking for a better solution already. What I would do is create a separate TaskExecutor bean in the RetryTopicBootstrapper, qualify the injection here, and then shut it down on a ContextClosedEvent.

I could also create a new instance in the KafkaConsumerBackoffManager, but then it would spawn an actual thread in the unit tests - usually not what we'd want if it was an app.

tomazfernandes:
> Just a little review. Although it raises more questions than just plain typos.
> I think I'll stop for now because we need to fix other things for upcoming releases this week.
> Will come back with my doubts to you later 😄

Hi @artembilan, I appreciate you taking the time!

This is the first time I'm on the framework's side instead of the user's, so feel free to point out any issues you find - I'm sure there's a lot to learn.

Thanks!

garyrussell:
@tomazfernandes Should we close https://github.com/spring-projects/spring-kafka/pull/1730 ?

tomazfernandes:
Sure @garyrussell, you can close that one - I separated them only to make it easier to review without the refactorings.

Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaConsumerBackoffManager.java
garyrussell: Given the proximity of RC1, tomorrow, I think we can live with this for now; I can scheduled an RC2 if need be.

spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.core.NestedRuntimeException;
+
+/**
+ * A {@link NestedRuntimeException} that records the timestamp
+ * of when it was thrown.
+ *
+ * @author Tomaz Fernandes
+ * @since 2.7
+ */
+public class TimestampedException extends NestedRuntimeException {
```
garyrussell, line 31: Should extend `KafkaException` (which is a `NestedRuntimeException`).

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaConsumerBackoffManagerTest.java
```diff
@@ -18,23 +18,30 @@
...
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
+import org.springframework.core.task.TaskExecutor;
 import org.springframework.kafka.config.KafkaListenerEndpointRegistry;
 import org.springframework.kafka.event.ListenerContainerPartitionIdleEvent;
 import org.springframework.kafka.retrytopic.TestClockUtils;
+import org.springframework.retry.backoff.Sleeper;
```
garyrussell, line 44: Convention for tests is plural `...Tests.java`.

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapterTest.java
```diff
@@ -41,6 +45,7 @@
 
 import org.springframework.kafka.listener.AcknowledgingConsumerAwareMessageListener;
 import org.springframework.kafka.listener.KafkaConsumerBackoffManager;
+import org.springframework.kafka.listener.TimestampedException;
```
garyrussell, line 48: Convention for tests is plural `...Tests.java`.

garyrussell (changes requested):
I concur with @artembilan - it is very difficult to review such large PRs, especially so close to a release. Given that this feature is still "experimental", I am ok with merging this with a less than thorough review, especially if we release RC2 in a couple of weeks.

Hopefully, this is the last large refactoring and we can look forward to more compact PRs in future.

But, once again @tomazfernandes , thanks for the great work.

spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
tomazfernandes: > I concur with @artembilan - it is very difficult to review such large PRs, especially so close to a release. Given that this feature is still "experimental", I am ok with merging this with a less than thorough review, especially if we release RC2 in a couple of weeks.
> 
> Hopefully, this is the last large refactoring and we can look forward to more compact PRs in future.
> 
> But, once again @tomazfernandes , thanks for the great work.

Sounds good to me @garyrussell, I've made the adjustments you asked.

I'll sure break it down in smaller PRs in the future.

Let me know if there's anything else to be changed - for this PR or for RC2 if it's to be released.

Thanks!

PR #1735, opened 2021-03-16
Capture Test Results in PR Builds

garyrussell:


artembilan:
So, looks like if we don't add new commits, but just squash, the workflow doesn't start 😢

garyrussell:
It did start but I had a syntax error.

artembilan:
Probably ready for merging?

PR #1738, opened 2021-03-16
Enable Verbose Test Logging from Gradle

garyrussell:
With these options, we now see

```
KafkaAdminTests > testTopicConfigs() FAILED
    Expecting:
     <2>
    to be equal to:
     <3>
    but was not.
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:500)
        at org.springframework.kafka.core.KafkaAdminTests.testTopicConfigs(KafkaAdminTests.java:84)
```

on the console.

It is still useful to capture test results, in case we need to look at STDOUT.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -181,8 +181,14 @@ subprojects { subproject ->
 	[compileJava, compileTestJava]*.options*.compilerArgs = ['-Xlint:all,-options']
 
 	test {
-		// suppress all console output during testing unless running `gradle -i`
-		logging.captureStandardOutput(LogLevel.INFO)
+		testLogging {
+			events "skipped", "failed"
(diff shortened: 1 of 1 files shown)


artembilan (approved):
Cool!

garyrussell:
I found it in reactor-kafka - although they log all tests, including passed - which I think is a bit too much.

Issue #1741, opened 2021-03-19
Add Overloaded Future<Message<?>> sendAndReceive(Message<?>) to ReplyingKafkaTemplate


garyrussell:
The `KafkaTemplate` supports sending spring-messaging `Message<?>` via a `MessageConverter`, but there is no equivalent in the `ReplyingKafkaTemplate`.

PR #1743, opened 2021-03-22
GH-1741: Add Message<?> sendAndReceive(Message<?>)

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1741

`ReplyingKafkaTemplate` enable send and receive with `Message<?>`.

- also fix ToC level for new chapter
- also polishing some tests for left over state

### DIFF
--- a/spring-kafka-docs/src/main/kotlin/org/springframework/kafka/kdocs/requestreply/Application.kt
+++ b/spring-kafka-docs/src/main/kotlin/org/springframework/kafka/kdocs/requestreply/Application.kt
@@ -47,6 +47,9 @@ import java.util.function.Consumer
  */
 @SpringBootApplication
 class Application {
+
+    private val log = LoggerFactory.getLogger(Application::class.java)
+
     @Bean
     fun topics(): NewTopics {
         return NewTopics(
@@ -61,12 +64,8 @@ class Application {
         )
     }
 
-    /* This is ok in Intellij but fails in gradle
     @Bean
-    fun kafkaTemplate(pf: ProducerFactory<Any, Any>?): KafkaTemplate<Any, Any> {
-        return KafkaTemplate(pf)
-    }
-     */
+    fun kafkaTemplate(pf: ProducerFactory<Any, Any>?) = KafkaTemplate(pf)
 
 // tag::beans[]
     @Bean
@@ -123,14 +122,10 @@ class Application {
         }
     }
 
-    companion object {
-        private val log = LoggerFactory.getLogger(Application::class.java)
-        @JvmStatic
-        fun main(args: Array<String>) {
-            System.setProperty("spring.kafka.producer.value-serializer", ByteArraySerializer::class.java.name)
-            System.setProperty("spring.kafka.consumer.value-deserializer", ByteArrayDeserializer::class.java.name)
-            SpringApplication.run(Application::class.java, *args).close()
-        }
-    }
+}
 
+fun main(args: Array<String>) {
+    System.setProperty("spring.kafka.producer.value-serializer", ByteArraySerializer::class.java.name)
+    System.setProperty("spring.kafka.consumer.value-deserializer", ByteArrayDeserializer::class.java.name)
+    SpringApplication.run(Application::class.java, *args).close()
 }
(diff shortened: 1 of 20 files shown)


spring-kafka-docs/src/main/kotlin/org/springframework/kafka/kdocs/started/consumer/Application.kt
```diff
@@ -17,6 +17,9 @@ package org.springframework.kafka.kdocs.started.consumer
 
 import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.apache.kafka.clients.admin.NewTopic
+import org.apache.kafka.common.serialization.ByteArrayDeserializer
+import org.apache.kafka.common.serialization.ByteArraySerializer
+import org.slf4j.LoggerFactory
```
artembilan, line 22: Doesn't look like these imports are in use.

spring-kafka-docs/src/main/kotlin/org/springframework/kafka/kdocs/started/consumer/Application.kt
```diff
@@ -44,7 +47,10 @@ class Application {
         println(`in`)
     }
 
-    fun main(args: Array<String>) = runApplication<Application>(*args)
+    companion object {
+        @JvmStatic
```
artembilan, line 51: Why do we need this?
What happened with the previous version?

artembilan: OK. I found the problem. The `main()` method must be declared outside of the class. It has to be a top level function. Plus like this:
```
fun main(args: Array<String>) {
    runApplication<Application>(*args)
}
```

returning `Unit`.

So, or this, or let's leave like yours! 😄 

Sorry for errors: I should double checked before committing such a polishing.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaOperations.java
```diff
@@ -35,6 +37,57 @@
  */
 public interface ReplyingKafkaOperations<K, V, R> {
 
+	/**
+	 * 	 * Send a request message and receive a reply message with the default timeout.
```
artembilan, line 41:
```suggestion
	 * Send a request message and receive a reply message with the default timeout.
```

spring-kafka-docs/src/main/kotlin/org/springframework/kafka/kdocs/requestreply/Application.kt
```diff
@@ -0,0 +1,136 @@
...
+                .replicas(1)
+                .build()
+        )
+    }
+
+    /* This is ok in Intellij but fails in gradle
+    @Bean
+    fun kafkaTemplate(pf: ProducerFactory<Any, Any>?): KafkaTemplate<Any, Any> {
+        return KafkaTemplate(pf)
+    }
```
artembilan, line 68: Does it work if you do like this:
```
    @Bean
    fun kafkaTemplate(pf: ProducerFactory<Any, Any>?) = KafkaTemplate(pf)
```
?

spring-kafka-docs/src/main/kotlin/org/springframework/kafka/kdocs/requestreply/Application.kt
```diff
@@ -0,0 +1,136 @@
...
+                template.sendAndReceive(MessageBuilder.withPayload("getThings").build(),
+                    object : ParameterizedTypeReference<List<Thing?>?>() {})
+            log.info(future2?.sendFuture?.get(10, TimeUnit.SECONDS)?.recordMetadata.toString())
+            val things = future2?.get(10, TimeUnit.SECONDS)?.payload
+            things?.forEach(Consumer { thing1: Thing? -> log.info(thing1.toString()) })
+// end::sendReceive[]
+        }
+    }
+
+    @KafkaListener(id = "myId", topics = ["requests"], properties = ["auto.offset.reset:earliest"])
```
artembilan, line 107: So weird that Kotlin cannot determine a single value for annotation array attribute 😄 
Probably some interoperability with Java...

artembilan (changes requested):
Some review

Issue #1747, opened 2021-03-26
Add Consumer to RecordInterceptor


garyrussell:
e.g. so the interceptor can get metrics.

PR #1748, opened 2021-03-26
GH-1747: Add ConsumerAwareRecordInterceptor

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1747

Also add `Consumer` to the new 2.7 methods and batch interceptor.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerAwareRecordInterceptor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerAwareRecordInterceptor.java
@@ -31,8 +31,10 @@ import org.springframework.lang.Nullable;
  * @since 2.7
  *
  */
+@FunctionalInterface
 public interface ConsumerAwareRecordInterceptor<K, V> extends RecordInterceptor<K, V> {
 
+	@SuppressWarnings("deprecation")
 	@Override
 	@Nullable
 	default ConsumerRecord<K, V> intercept(ConsumerRecord<K, V> record) {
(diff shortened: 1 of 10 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerAwareRecordInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+ * A {@link RecordInterceptor} that has access to the {@link Consumer}.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Gary Russell
+ * @since 2.7
+ *
+ */
+public interface ConsumerAwareRecordInterceptor<K, V> extends RecordInterceptor<K, V> {
```
artembilan, line 34: `@FunctionalInterface` ?

I guess we can live without this intermediary since all the methods are present in the `RecordInterceptor`, but I believe you have introduced it for end-user convenience.

spring-kafka/src/main/java/org/springframework/kafka/listener/RecordInterceptor.java
```diff
@@ -43,21 +44,38 @@
 	@Nullable
 	ConsumerRecord<K, V> intercept(ConsumerRecord<K, V> record);
```
artembilan, line 45: I would say that we don't need this contract at all: it's up to end-user to do anything with that extra `consumer` or not.

More over I see that you have changed `BatchInterceptor` contract for that extra param.
Therefore for consistency this might be deprecated with removal in the future.

But yeah, I'll deffer the final decision to you...

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/RecordInterceptor.java
garyrussell: Yes, will deprecate - and then we can deprecate the sub-interface when we remove this.

PR #1756, opened 2021-04-05
GH-1744: Improve KafkaConsumerBackoffManager dependency management

### LINKED ISSUE #1744: Improve KafkaConsumerBackoffManager's dependency injection
Hi @garyrussell and @artembilan, bringing back this discussion from the last RetryTopic's PR.

As of now, we're injecting `KafkaConsumerBackoffManager`'s dependencies using this code:

```
public KafkaConsumerBackoffManager(ListenerContainerRegistry registry,
									@Qualifier(INTERNAL_BACKOFF_CLOCK_BEAN_NAME) Clock clock,
									TaskExecutor taskExecutor,
									Sleeper sleeper) {

		this.registry = registry;
		this.clock = clock;
		this.taskExecutor = taskExecutor;
		this.sleeper = sleeper;
		this.backOffContexts = new HashMap<>();
	}

```
(https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaConsumerBackoffManager.java)

There are a few considerations on this. The main one for me is the fact that we're using the user app's TaskExecutor to run the timing adjustment tasks, which seems like a no-no.

Also, @artembilan called out that from the framework's standpoint we shouldn't inject the Clock and should instead rely on methods such as System.currentTimeMillis(). What I don't really like with this approach is that I think it would make unit testing a lot harder, but I'm open to that if that's the best way to go. If you could share the reasoning behind this it'd be great. We should also check if the same reasoning applies to the injected Sleeper and TaskExecutor, which are there mainly for unit testing purposes too.

What's in my mind is having the `RetryTopicBootstrapper` class handle the KCBM's dependencies lookup / creation / injection, create the KCBM instance manually there, and then add it to the application context. This way we wouldn't have to qualify the injections in KCBM but we'd still have control over what we're specifically injecting - unless the user has already instantiated a KCBM with the same internal bean name, in which case we'll use theirs.

I'd also create a separate TaskExecutor for KCBM there and handle said Executor's graceful shutdown on a ContextClosedEvent. We're already registering / configuring all other RT's beans in the bootstrapper class.

(https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicBootstrapper.java)

Please let me know what you think of this and if any further clarification is needed.

Thanks!

tomazfernandes:
Resolves #1744
Introduces the KafkaConsumerTimingAdjuster interface
Introduces the KafkaBackOffManagerFactory interface

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/TimingAdjustingKafkaBackOffManagerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/TimingAdjustingKafkaBackOffManagerFactory.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright 2018-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.listener;
+
+import java.time.Clock;
+
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextClosedEvent;
+import org.springframework.core.task.TaskExecutor;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+import org.springframework.util.Assert;
+
+/**
+ *
+ * Creates a {@link KafkaConsumerBackoffManager} instance
+ * with or without a {@link KafkaConsumerTimingAdjuster}.
+ *
+ * @author Tomaz Fernandes
+ * @since 2.7
+ */
+public class TimingAdjustingKafkaBackOffManagerFactory extends AbstractKafkaBackOffManagerFactory {
+
+	private boolean isTimingAdjustmentEnabled = true;
+
+	private KafkaConsumerTimingAdjuster timingAdjustmentManager;
+
+	private TaskExecutor taskExecutor;
+
+	private final Clock clock;
+
+	/**
+	 * Constructs a factory instance that will create the {@link KafkaConsumerBackoffManager}
+	 * instances with the provided {@link KafkaConsumerTimingAdjuster}.
+	 *
+	 * @param timingAdjustmentManager the {@link KafkaConsumerTimingAdjuster} to be used.
+	 */
+	public TimingAdjustingKafkaBackOffManagerFactory(KafkaConsumerTimingAdjuster timingAdjustmentManager) {
+		this();
+		this.setTimingAdjustmentManager(timingAdjustmentManager);
+	}
+
+	/**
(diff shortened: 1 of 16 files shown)


tomazfernandes:
This ended up being a larger PR than I anticipated - although mostly new code. If it's too close to RC2 for you to review I don't mind if it goes into the next release.

Thanks!

spring-kafka/src/test/java/org/springframework/kafka/listener/WakingKafkaConsumerTimingAdjusterTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.common.TopicPartition;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+
+import org.springframework.core.task.TaskExecutor;
+import org.springframework.retry.backoff.Sleeper;
+
+class WakingKafkaConsumerTimingAdjusterTest {
```
garyrussell, line 33: Convention is `...Tests`

garyrussell (changes requested):
Just some minor polishing needed.

spring-kafka/src/main/java/org/springframework/kafka/listener/TimingAdjustingKafkaBackOffManagerFactory.java
```diff
@@ -0,0 +1,165 @@
...
+
+	/**
+	 * Constructs a factory instance that will create the {@link KafkaConsumerBackoffManager}
+	 * instances with the provided {@link KafkaConsumerTimingAdjuster}.
+	 *
+	 * @param timingAdjustmentManager the {@link KafkaConsumerTimingAdjuster} to be used.
+	 */
+	public TimingAdjustingKafkaBackOffManagerFactory(KafkaConsumerTimingAdjuster timingAdjustmentManager) {
+		this();
+		this.setTimingAdjustmentManager(timingAdjustmentManager);
```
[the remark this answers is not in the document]
tomazfernandes: Actually I needed to call this() to instantiate the default clock, but I guess it's better to do that explicitly.

tomazfernandes:
> Just some minor polishing needed.

@garyrussell changes made, please let me know if there's anything else. Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/TimingAdjustingKafkaBackOffManagerFactory.java
```diff
@@ -119,17 +119,17 @@ public void setIsTimingAdjustmentEnabled(boolean isTimingAdjustmentEnabled) {
 	 *
 	 * @param timingAdjustmentManager the adjustmentManager to be used.
 	 */
-	public void setTimingAdjustmentManager(KafkaConsumerTimingAdjuster timingAdjustmentManager) {
-		Assert.isTrue(this.isTimingAdjustmentEnabled, () -> "TimingAdjustment is disabled for this factory.");
+	final public void setTimingAdjustmentManager(KafkaConsumerTimingAdjuster timingAdjustmentManager) {
```
garyrussell, line 122: `public final` - will fix on merge.

garyrussell:
Merged as 290d11a319145a741303f6affcff23ff66ee57d2 with minor polishing.

tomazfernandes:
Thanks a lot @garyrussell! 🙏🏼

Issue #1757, opened 2021-04-07
Add Option to Add ConsumerRecord to Message Headers


garyrussell:
Useful for error handling in a listener error handler. See https://stackoverflow.com/questions/66982480/request-reply-and-retry-policy-for-kafka-listeners/66987003#66987003

PR #1758, opened 2021-04-07
GH-1757: Option for Raw ConsumerRecord Header

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1757


artembilan (changes requested):
LGTM, but shouldn't we mention such an improvement in the What's New?

Issue #1770, opened 2021-04-15
ContainerProperties from KafkaListenerContainerFactory are dropped


thomasmey:
Hi,

I think the overwritten method here
https://github.com/spring-projects/spring-kafka/blob/1c798665050936d19934811198732c09ad98282f/spring-kafka/src/main/java/org/springframework/kafka/config/ConcurrentKafkaListenerContainerFactory.java#L61
has a bug, as that it does *not* use the already existing container properties from super class org.springframework.kafka.config.AbstractKafkaListenerContainerFactory.containerProperties

So it basically drops any configuration like container clientId (!)

What do you think, bug or feature?

garyrussell:
If that were the case, nothing would work.

The properties are copied in later; see

https://github.com/spring-projects/spring-kafka/blob/1c798665050936d19934811198732c09ad98282f/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java#L402-L405

thomasmey:
Hi,

mhh, I think I see now where the confusion comes from:

when I do:
``
        @Bean
        public ConcurrentKafkaListenerContainerFactory<String, String> kafkaListenerContainerFactory() {
                ConcurrentKafkaListenerContainerFactory<String, String> factory = new ConcurrentKafkaListenerContainerFactory<>();
                factory.getContainerProperties().setClientId("clientId");
                return factory;
        }
``
this can not work with the current code. the question is: should this (set a container client id programatically)?

makes sense?

garyrussell:
Why do you say it "can not work"? It does work.

I tested it (again) when I answered this https://github.com/spring-projects/spring-kafka/issues/398#issuecomment-820414777

The only difference is I set it via a boot property `spring.kafka.consumer.client-id=foo`; before I added the clientIdPrefix to the annotations, I got the already exists exception.

garyrussell:
The client Id comes from the endpoint (populated from the `@KafkaListener`) and overwrites the container properties, even if not set (which is a bug).

If you want to change it programmatically, use a container customizer instead...

```java
factory.setContainerCustomizer(container -> {
	container.getContainerProperties().setClientId("client-for-"
			+ container.getContainerProperties().getTopics()[0]);
});
```

thomasmey:
> My apologies; I see it does NOT work in that case. Investigating...

No, worries. 
Thanks for looking into it 😃

PR #1771, opened 2021-04-15
GH-1770: Fix ClientId for @KafkaListener

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1770

The clientId property, if set on the container factory, is overwritten
by an empty String if not defined on the annotation.

Check for an empty String (groupId too).

**cherry-pick to 2.6.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
@@ -51,7 +51,6 @@ import org.springframework.kafka.support.converter.MessageConverter;
 import org.springframework.retry.RecoveryCallback;
 import org.springframework.retry.support.RetryTemplate;
 import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
 
 /**
  * Base {@link KafkaListenerContainerFactory} for Spring's base container implementation.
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
```diff
@@ -426,8 +427,10 @@ else if (this.autoStartup != null) {
 				.acceptIfNotNull(this.phase, instance::setPhase)
 				.acceptIfNotNull(this.applicationContext, instance::setApplicationContext)
 				.acceptIfNotNull(this.applicationEventPublisher, instance::setApplicationEventPublisher)
-				.acceptIfNotNull(endpoint.getGroupId(), instance.getContainerProperties()::setGroupId)
-				.acceptIfNotNull(endpoint.getClientIdPrefix(), instance.getContainerProperties()::setClientId)
+				.acceptIfCondition(StringUtils.hasText(endpoint.getGroupId()), endpoint.getGroupId(),
```
artembilan, line 430: I see there is a convenient `acceptIfHasText()` instead of this combination.
What am I missing?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
garyrussell: Oops... forgot about that.

artembilan:
... and cherry-picked.

PR #1773, opened 2021-04-19
GH-1772 Add RetryTopicNamesProvider

### LINKED ISSUE #1772: RetryableTopic: improve customization of retry/dl topic names
**PROBLEM**

In our company we have a naming convention to distinguish "public" topics, meant for communication between different applications, and "private" topics, like retry and deadletter topics, meant to be used internally by an application. This convention mandate different prefixes for the names of those topics, and this is not possible with the current customization possibilities of RetryableTopic.

**SOLUTION**
The first possibility is to add a new "prefix" configuration property, to be handled in a similar way to the current suffix property. Probably easier to do, but in addition to increase the number of properties, doesn't give complete control on the different names (not only topic name) to the user.

Another approach is to move all the naming logic (specifically the logic of org.springframework.kafka.retrytopic.RetryTopicConfigurer.EndpointCustomizerFactory#createEndpointCustomizer) to an injected strategy bean that could be overridden in the configuration.

apolci:
Resolves #1772 

Add provider to handle the names used on retry/dl topics (topic name, groupId, etc.). This can be extended to customize the behaviour in addition to the simple suffix parameter.

Currently customizing the default requires to register a bean  implementing RetryTopicNamesProviderFactory interface, with name "internalRetryTopicNamesProviderFactory".

### DIFF
--- a/spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
@@ -429,6 +429,50 @@ public RetryTopicConfiguration myRetryTopic(KafkaTemplate<String, MyPojo> templa
 
 NOTE: The default behavior is to suffix with the delay values, except for fixed delay configurations with multiple topics, in which case the topics are suffixed with the topic's index.
 
+===== Custom naming strategies
+
+More complex naming strategies can be accomplished by registering a bean that implements `RetryTopicNamesProviderFactory`. The default implementation is `SuffixingRetryTopicNamesProviderFactory` and a different implementation can be registered in the following way:
+
+====
+[source, java]
+----
+@Bean
+public RetryTopicNamesProviderFactory myRetryNamingProviderFactory() {
+    return new CustomRetryTopicNamesProviderFactory();
+}
+----
+====
+
+As an example the following implementation, in addition to the standard suffix, add a prefix to retry/dl topics names:
+
+====
+[source, java]
+----
+public class CustomRetryTopicNamesProviderFactory implements RetryTopicNamesProviderFactory {
+
+	@Override
+    public RetryTopicNamesProvider createRetryTopicNamesProvider(
+                DestinationTopic.Properties properties) {
+
+        if(properties.isMainEndpoint()) {
+            return new SuffixingRetryTopicNamesProvider(properties);
+        }
+        else {
+            return new SuffixingRetryTopicNamesProvider(properties) {
+
+                @Override
+                public String getTopicName(String topic) {
+                    return "my-prefix-" + super.getTopicName(topic);
+                }
+
+            };
+        }
+    }
+
+}
+----
+====
+
 ==== Dlt Strategies
 
 The framework provides a few strategies for working with DLTs. You can provide a method for DLT processing, use the default logging method, or have no DLT at all. Also you can choose what happens if DLT processing fails.
--- a/spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
@@ -443,7 +443,7 @@ public RetryTopicNamesProviderFactory myRetryNamingProviderFactory() {
 ----
 ====
 
-As an example the following implementation, in addition to the standard suffix, add a prefix to retry/dl topics names:
+As an example the following implementation, in addition to the standard suffix, adds a prefix to retry/dl topics names:
 
 ====
 [source, java]
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DefaultRetryTopicNamesProviderFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DefaultRetryTopicNamesProviderFactory.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.retrytopic;
+
+import org.springframework.kafka.config.MethodKafkaListenerEndpoint;
+import org.springframework.kafka.support.Suffixer;
+
+/**
+ * Default handling of retry and dead letter naming.
+ *
+ * @author Andrea Polci
+ */
+public class DefaultRetryTopicNamesProviderFactory implements RetryTopicNamesProviderFactory {
+	@Override
+	public RetryTopicNamesProvider createRetryTopicNamesProvider(DestinationTopic.Properties properties) {
+		return new DefaultRetryTopicNamesProvider(properties);
+	}
+
+	public static class DefaultRetryTopicNamesProvider implements RetryTopicNamesProvider {
+		private final Suffixer suffixer;
+
+		public DefaultRetryTopicNamesProvider(DestinationTopic.Properties properties) {
+			this.suffixer = new Suffixer(properties.suffix());
+		}
+
+		@Override
+		public String getEndpointId(MethodKafkaListenerEndpoint<?, ?> endpoint) {
+			return this.suffixer.maybeAddTo(endpoint.getId());
+		}
+
+		@Override
+		public String getGroupId(MethodKafkaListenerEndpoint<?, ?> endpoint) {
+			return this.suffixer.maybeAddTo(endpoint.getGroupId());
+		}
+
+		@Override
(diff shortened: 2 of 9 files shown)


spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurer.java
```diff
@@ -221,14 +221,18 @@
...
+								ListenerContainerFactoryResolver containerFactoryResolver,
+								ListenerContainerFactoryConfigurer listenerContainerFactoryConfigurer,
+								BeanFactory beanFactory,
+								RetryTopicNamesProviderFactory retryTopicNamesProviderFactory) {
 		this.destinationTopicProcessor = destinationTopicProcessor;
 		this.containerFactoryResolver = containerFactoryResolver;
 		this.listenerContainerFactoryConfigurer = listenerContainerFactoryConfigurer;
 		this.beanFactory = beanFactory;
+		this.retryTopicNamesProviderFactory = retryTopicNamesProviderFactory;
 	}
```
garyrussell, line 236: Since this is a public API and we have released 2.7.0 already, we need to keep the old constructor (and deprecate it), populating the new factory with the default.

It is not necessary to do this on the non-public APIs.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicNamesProviderFactory.java
```diff
@@ -0,0 +1,42 @@
...
+	interface RetryTopicNamesProvider {
+		String getEndpointId(MethodKafkaListenerEndpoint<?, ?> endpoint);
+
+		String getGroupId(MethodKafkaListenerEndpoint<?, ?> endpoint);
+
+		String getClientIdPrefix(MethodKafkaListenerEndpoint<?, ?> endpoint);
+
+		String getGroup(MethodKafkaListenerEndpoint<?, ?> endpoint);
+
+		String getTopicName(String topic);
```
garyrussell, line 40: These need javadocs; blank line before first method and between `}` below.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurer.java
```diff
@@ -221,14 +221,18 @@
...
+
 	public RetryTopicConfigurer(DestinationTopicProcessor destinationTopicProcessor,
-						ListenerContainerFactoryResolver containerFactoryResolver,
-						ListenerContainerFactoryConfigurer listenerContainerFactoryConfigurer,
-						BeanFactory beanFactory) {
+								ListenerContainerFactoryResolver containerFactoryResolver,
+								ListenerContainerFactoryConfigurer listenerContainerFactoryConfigurer,
+								BeanFactory beanFactory,
+								RetryTopicNamesProviderFactory retryTopicNamesProviderFactory) {
 		this.destinationTopicProcessor = destinationTopicProcessor;
```
garyrussell, line 231: Our convention for multi-line signatures is to add a blank line before the code.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DefaultRetryTopicNamesProviderFactory.java
```diff
@@ -0,0 +1,65 @@
...
+ * @author Andrea Polci
+ */
+public class DefaultRetryTopicNamesProviderFactory implements RetryTopicNamesProviderFactory {
+	@Override
+	public RetryTopicNamesProvider createRetryTopicNamesProvider(DestinationTopic.Properties properties) {
+		return new DefaultRetryTopicNamesProvider(properties);
+	}
+
+	public static class DefaultRetryTopicNamesProvider implements RetryTopicNamesProvider {
+		private final Suffixer suffixer;
```
garyrussell, line 34: blank line above; and line # 28 and between last `}` at the end of the file.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurer.java
```diff
@@ -494,4 +501,6 @@ public void logMessage(Object message) {
 			}
 		}
 	}
+
+
```
garyrussell, line 505: remove one of these blank lines.

garyrussell (changes requested):
@apolci Thanks for the contribution! Just some minor issues.

LGTM otherwise. @tomazfernandes any objections?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurer.java
apolci: Will apply requested changes.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicNamesProviderFactory.java
```diff
@@ -0,0 +1,42 @@
...
+ * Handles the naming related to the retry and dead letter topics.
+ *
+ * @author Andrea Polci
+ * @see DefaultRetryTopicNamesProviderFactory
+ */
+public interface RetryTopicNamesProviderFactory {
+
+	RetryTopicNamesProvider createRetryTopicNamesProvider(DestinationTopic.Properties properties);
+
+	interface RetryTopicNamesProvider {
```
tomazfernandes, line 31: I don't think I remember having ever seen an interface inside another. One more trick for the book, thanks!

tomazfernandes (changes requested):

tomazfernandes:
No objections whatsoever, looks good! I've just added a comment about the default implementation class name, but it's a matter of personal taste more than anything.

tomazfernandes:
Also @garyrussell do you think this is a feature worth documenting, or should we leave it for more advanced users looking for extension points? It seems like a pretty useful customisation, and a not so hard one to implement.

garyrussell:
@tomazfernandes Yes, we should add a few words to the docs; perhaps you can guide @apolci where it should go.

Thanks.

tomazfernandes:
@apolci it'd be great if you could add this feature to the documentation. We already have a section on Topic Naming: https://docs.spring.io/spring-kafka/reference/html/#topic-naming, 4.4.5, so I think it'd be a matter of adding it after the existing  text, before 4.4.6. 

What do you think?

It's in this file:
https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka-docs/src/main/asciidoc/retrytopic.adoc

apolci:
> @apolci it'd be great if you could add this feature to the documentation...

Yes sure, I'll try to do it tomorrow.

I'm not 100% sure of how it can be configured at the moment. I followed the handling of the other constructor parameters of RetryTopicConfigurer regarding the initialization in RetryTopicBootstrapper.
This means, if I'm not wrong, that a bean named "internalRetryTopicNamesProviderFactory" need to be present in the context in order to override the default handling. Do you think this is the correct way to do that?

tomazfernandes:
> > @apolci it'd be great if you could add this feature to the documentation...
> 
> Yes sure, I'll try to do it tomorrow.
> 
> I'm not 100% sure of how it can be configured at the moment. I followed the handling of the other constructor parameters of RetryTopicConfigurer regarding the initialization in RetryTopicBootstrapper.
> This means, if I'm not wrong, that a bean named "internalRetryTopicNamesProviderFactory" need to be present in the context in order to override the default handling. Do you think this is the correct way to do that?

Yep, that's what I had in mind. But maybe since this interface is really about the retry topic logic and thus there's little chance the user will have more than one bean implementing it for a different purpose, we might be a little more lenient and try to find the bean by type in the bootstrapper logic, and refraining to register the bean if there's already one there, so that the user doesn't have to know the exact bean name.

I'm ok with both approaches.

apolci:
> ...
> I'm ok with both approaches.

I implemented the second approach (search bean by class) and added the documentation. Sorry for the many commits, should I squash them?

garyrussell:
No; best to keep commits separate; we will squash during merge.

Thanks!

apolci:
Sorry, I still see 1 change request pending, but I think I resolved all of them. I just want to be sure I'm not missing something

garyrussell:
That's from the previous review; I just haven't gotten around to re-reviewing (and we have a day off tomorrow, so it will probably be next week now).

Thanks for your patience.

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -429,6 +429,50 @@ public RetryTopicConfiguration myRetryTopic(KafkaTemplate<String, MyPojo> templa
...
+[source, java]
+----
+@Bean
+public RetryTopicNamesProviderFactory myRetryNamingProviderFactory() {
+    return new CustomRetryTopicNamesProviderFactory();
+}
+----
+====
+
+As an example the following implementation, in addition to the standard suffix, add a prefix to retry/dl topics names:
```
tomazfernandes, line 446: I think it should be "adds a prefix", shouldn't it?

tomazfernandes:
Apart from a tiny typo I just commented on looks good to me!

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
apolci: Yes, thank you 👍

garyrussell (approved):

garyrussell:
@apolci Thanks for the contribution (and @tomazfernandes thanks for reviewing).

Issue #1774, opened 2021-04-20
Update EmbeddedKafkaBroker for 2.8.0 Compatibility


garyrussell:
`BrokerStates.scala` replaced by `BrokerState.java`.

https://github.com/apache/kafka/commit/772f2cfc828d9da56559968a8eed06f2c8c43062

PR #1775, opened 2021-04-20
GH-1774: kafka-clients 2.8.0 Compatibility

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1774

- changes to `BrokerState`
- suppress some deprecation warnings

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
@@ -119,12 +119,9 @@ public class EmbeddedKafkaBroker implements InitializingBean, DisposableBean {
 
 	public static final int DEFAULT_ZK_SESSION_TIMEOUT = 6000;
 
-	private static final String clientVersion;
-
 	private static final Method getBrokerState;
 
 	static {
-		clientVersion = AppInfoParser.getVersion();
 		try {
 			Method method = KafkaServer.class.getDeclaredMethod("brokerState");
 			if (method.getReturnType().equals(AtomicReference.class)) {
@@ -136,7 +133,7 @@ public class EmbeddedKafkaBroker implements InitializingBean, DisposableBean {
 		}
 		catch (NoSuchMethodException | SecurityException e) {
 			throw new RuntimeException("Failed to determine KafkaServer.brokerState() method; client version: "
-					+ clientVersion);
+					+ AppInfoParser.getVersion());
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
```diff
@@ -115,6 +119,27 @@
 
 	public static final int DEFAULT_ZK_SESSION_TIMEOUT = 6000;
 
+	private static final String clientVersion;
```
artembilan, line 122: Doesn't look like we need this as a property: you just use a version info in the exception message.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -188,6 +188,7 @@ public void setStateListener(KafkaStreams.StateListener stateListener) {
 		this.stateListener = stateListener; // NOSONAR (sync)
 	}
 
+	@SuppressWarnings("deprecation")
```
artembilan, line 191: I don't see that you have change `kafka-streams` version, but probably this is a missed artifact from the previous project state. I wonder what is a replacement for this deprecated API then...

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaStreamBrancher.java
```diff
@@ -86,6 +86,7 @@
 	 * @param stream {@code KStream} to split
 	 * @return the provided stream
 	 */
+	@SuppressWarnings("deprecation")
```
artembilan, line 89: DITTO

artembilan (changes requested):
Some review. Mostly nit-picking 😉

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
garyrussell: I haven't changed any versions; this is just for compatibility, to allow users to override the client version.

kafka-streams has to match kafka-clients version.

```java
     * @deprecated Since 2.8.0. Use {@link KafkaStreams#setUncaughtExceptionHandler(StreamsUncaughtExceptionHandler)} instead.
```

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaStreamBrancher.java
garyrussell:
```java
     * @deprecated since 2.8. Use {@link #split()} instead.
     */
    @Deprecated
    KStream<K, V>[] branch(final Predicate<? super K, ? super V>... predicates);
```

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
garyrussell: Oh; I see what you mean; we really don't need these suppressions because we're still compiling against the 2.7 clients. Will revert.

Issue #1781, opened 2021-04-22
JsonSerializer.writer is private


cnygardtw:
When trying to customize the date formatting of the JsonSerializer, the objectMapper is protected, but the ObjectWriter is private.  Since the writer is initialized only through constructor, and certain customizations to the ObjectMapper won't magically propagate, I had to provide much more code than is probably necessary for this use-case.

Instead of:
```java
public class CustomJsonSerializer<T> extends JsonSerializer<T> {

  private ObjectWriter myWriter;

  /**
   * constructor for custom json serializer. configure the ObjectMapper using the
   * KafkaSerdeObjectMapperConfig function.
   */
  public CustomJsonSerializer() {
    super();
    log.debug("Configuring Kafka object mapper [{}]", this.objectMapper);
    // Use our standard configuration bean to configure the Kafka serializer
    KafkaSerdeObjectMapperConfig configurer = new KafkaSerdeObjectMapperConfig();
    configurer.configureObjectMapper(objectMapper);
    myWriter = objectMapper.writerFor((JavaType) null);
  }

  @Override
  @Nullable
  public byte[] serialize(String topic, @Nullable T data) {
    if (data == null) {
      return null;
    }
    try {
      return this.myWriter.writeValueAsBytes(data);
    } catch (IOException ex) {
      throw new SerializationException(
          "Can't serialize data [" + data + "] for topic [" + topic + "]", ex);
    }
  }

}
```

I'd rather see something like:

```java
class JsonSerializer<T> implements Serializer<T> {
  // ...

  public JsonSerializer(JavaType targetType, ObjectMapper objectMapper) {
    Assert.notNull(objectMapper, "'objectMapper' must not be null.");
    this.objectMapper = customizeObjectMapper(objectMapper);
    this.writer = this.objectMapper.writerFor(targetType);
  }

  public ObjectMapper customizeObjectMapper(ObjectMapper mapper) {
    return mapper;
  }

  //..
}
```

which would allow me to:

```java
public class CustomJsonSerializer<T> extends JsonSerializer<T> {

  /**
   * constructor for custom json serializer. 
   */
  public CustomJsonSerializer() {
    super();
  }

  @Override
  ObjectMapper customizeObjectMapper(ObjectMapper mapper) {
    log.debug("Configuring Kafka object mapper [{}]", this.objectMapper);
    mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
    return mapper;
  }

}

```

[...]

PR #1782, opened 2021-04-22
GH-1781: Configure ObjectMapper in JsonSerializer/JsonDeserializer

cnygardtw:
Fixes GH-1781 (https://github.com/spring-projects/spring-kafka/issues/1781)

* Add ObjectMapperCustomizer interface for customizing local ObjectMapper
* Implement ObjectMapperCustomizer in JsonSerializer / JsonDeserializer
* call customizeObjectMapper() in ctor so ObjectWriter is properly constructed
* added tests / docs

### DIFF
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -3920,6 +3920,55 @@ They have no effect if you have provided `Serializer` and `Deserializer` instanc
 Starting with version 2.2, the type information headers (if added by the serializer) are removed by the deserializer.
 You can revert to the previous behavior by setting the `removeTypeHeaders` property to `false`, either directly on the deserializer or with the configuration property described earlier.
 
+[[serdes-customize-object-mapper]]
+====== Customize the ObjectMapper
+
+Occasionally, the only way to configure the way information is serialized is by customizing the ObjectMapper used in (de)serialization.
+Because Kafka uses its own ObjectMapper separate from the normal Spring-Boot ObjectMapper, customizing the ObjectMapper typically involved (re)implementing various Kafka factories.
+Since 2.7.1, a new interface `ObjectMapperCustomizer` has been introduced and implemented by the `JsonSerializer<T>` and `JsonDeserializer<T>` classes.
+This makes it simpler to configure the ObjectMapper without losing the convenience of the Spring property configuration capabilities.
+
+The following is an example of how to provide a simple customized serializer class (derived from `JsonSerializer<T>`) to do the configuration.
+
+====
+[source, java]
+----
+package com.example;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+
+import org.springframework.kafka.support.serializer.JsonSerializer;
+
+public class CustomJsonSerializer<T> extends JsonSerializer<T> {
+
+	/**
+	 * constructor for custom json serializer.
+	 */
+	public CustomJsonSerializer() {
+		super();
+	}
+
+	@Override
+	public ObjectMapper customizeObjectMapper(ObjectMapper mapper) {
+		mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+		return mapper;
+	}
+}
+----
+====
+
+Then in your properties file, configure your `CustomJsonSerializer<T>` class as the `value-serializer`:
+
+====
+[source, properties]
+----
+spring.kafka.producer.value-serializer=com.example.CustomJsonSerializer
+----
+====
+
+The same general pattern can be applied to create and configure a `CustomJsonDeserializer<T>` class to configure how Json data is deserialized.
+
 [[serdes-mapping-types]]
 ====== Mapping Types
 
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -3795,6 +3795,7 @@ Also since version 2.7 `ConsumerPartitionPausedEvent` and `ConsumerPartitionResu
 [[serdes]]
 ==== Serialization, Deserialization, and Message Conversion
 
+[[serdes-overview]]
 ===== Overview
 
 Apache Kafka provides a high-level API for serializing and deserializing record values as well as their keys.
@@ -3921,25 +3922,19 @@ Starting with version 2.2, the type information headers (if added by the seriali
 You can revert to the previous behavior by setting the `removeTypeHeaders` property to `false`, either directly on the deserializer or with the configuration property described earlier.
 
 [[serdes-customize-object-mapper]]
-====== Customize the ObjectMapper
+====== Customize the `ObjectMapper`
 
-Occasionally, the only way to configure the way information is serialized is by customizing the ObjectMapper used in (de)serialization.
-Because Kafka uses its own ObjectMapper separate from the normal Spring-Boot ObjectMapper, customizing the ObjectMapper typically involved (re)implementing various Kafka factories.
-Since 2.7.1, a new interface `ObjectMapperCustomizer` has been introduced and implemented by the `JsonSerializer<T>` and `JsonDeserializer<T>` classes.
-This makes it simpler to configure the ObjectMapper without losing the convenience of the Spring property configuration capabilities.
+Occasionally, the only way to configure the way information is serialized is by customizing the `ObjectMapper` used in (de)serialization.
+When instantiated by Apache Kafka, the Spring JSON Serializer and Deserializer use a locally declared `ObjectMapper`.
+The only way to customize the `ObjectMapper` was to inject instances into the producer and consumer factories as discussed <<serdes-overview,here>> rather than by using properties.
+Since version 2.5.13, a new interface `ObjectMapperCustomizer` has been introduced and implemented by the `JsonSerializer<T>` and `JsonDeserializer<T>` classes.
+This makes it simpler to configure the `ObjectMapper` without losing the convenience of the Spring property configuration capabilities.
 
 The following is an example of how to provide a simple customized serializer class (derived from `JsonSerializer<T>`) to do the configuration.
 
 ====
 [source, java]
 ----
-package com.example;
-
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
-
-import org.springframework.kafka.support.serializer.JsonSerializer;
-
 public class CustomJsonSerializer<T> extends JsonSerializer<T> {
 
 	/**
@@ -3958,7 +3953,7 @@ public class CustomJsonSerializer<T> extends JsonSerializer<T> {
 ----
 ====
 
-Then in your properties file, configure your `CustomJsonSerializer<T>` class as the `value-serializer`:
+Then in your Spring Boot properties file, for example, configure your `CustomJsonSerializer<T>` class as the `value-serializer`:
 
 ====
 [source, properties]
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
@@ -60,7 +60,7 @@ import com.fasterxml.jackson.databind.type.TypeFactory;
  * @author Torsten Schleede
  * @author Ivan Ponomarev
  */
-public class JsonDeserializer<T> implements Deserializer<T> {
+public class JsonDeserializer<T> implements Deserializer<T>, ObjectMapperCustomizer {
 
 	private static final String KEY_DEFAULT_TYPE_STRING = "spring.json.key.default.type";
 
@@ -114,7 +114,7 @@ public class JsonDeserializer<T> implements Deserializer<T> {
 
 	protected Jackson2JavaTypeMapper typeMapper = new DefaultJackson2JavaTypeMapper(); // NOSONAR
 
-	private ObjectReader reader;
+	protected CachedObjectReader cachedReader;
 
 	private boolean typeMapperExplicitlySet = false;
 
@@ -243,7 +243,8 @@ public class JsonDeserializer<T> implements Deserializer<T> {
 			boolean useHeadersIfPresent) {
 
 		Assert.notNull(objectMapper, "'objectMapper' must not be null.");
-		this.objectMapper = customizeObjectMapper(objectMapper);
+		this.objectMapper = objectMapper;
+		this.cachedReader = new CachedObjectReader();
 		JavaType javaType = null;
 		if (targetType == null) {
 			Class<?> genericType = ResolvableType.forClass(getClass()).getSuperType().resolveGeneric(0);
@@ -287,7 +288,8 @@ public class JsonDeserializer<T> implements Deserializer<T> {
 			boolean useHeadersIfPresent) {
 
 		Assert.notNull(objectMapper, "'objectMapper' must not be null.");
-		this.objectMapper = customizeObjectMapper(objectMapper);
+		this.objectMapper = objectMapper;
+		this.cachedReader = new CachedObjectReader();
 		initialize(targetType, useHeadersIfPresent);
 	}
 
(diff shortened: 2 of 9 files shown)


pivotal-issuemaster:
@cnygardtw Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1782)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1782) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@cnygardtw Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-kafka&pullRequestId=1782)!

spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -3920,6 +3920,55 @@ They have no effect if you have provided `Serializer` and `Deserializer` instanc
 Starting with version 2.2, the type information headers (if added by the serializer) are removed by the deserializer.
 You can revert to the previous behavior by setting the `removeTypeHeaders` property to `false`, either directly on the deserializer or with the configuration property described earlier.
 
+[[serdes-customize-object-mapper]]
+====== Customize the ObjectMapper
+
+Occasionally, the only way to configure the way information is serialized is by customizing the ObjectMapper used in (de)serialization.
+Because Kafka uses its own ObjectMapper separate from the normal Spring-Boot ObjectMapper, customizing the ObjectMapper typically involved (re)implementing various Kafka factories.
```
artembilan, line 3927: Please, call the trade mark as `Apache Kafka`.

spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -3920,6 +3920,55 @@ They have no effect if you have provided `Serializer` and `Deserializer` instanc
 Starting with version 2.2, the type information headers (if added by the serializer) are removed by the deserializer.
 You can revert to the previous behavior by setting the `removeTypeHeaders` property to `false`, either directly on the deserializer or with the configuration property described earlier.
 
+[[serdes-customize-object-mapper]]
+====== Customize the ObjectMapper
+
+Occasionally, the only way to configure the way information is serialized is by customizing the ObjectMapper used in (de)serialization.
```
artembilan, line 3926: The `ObjectMapper` should be wrapped into a code snippet using backtick symbol

spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -3920,6 +3920,55 @@ They have no effect if you have provided `Serializer` and `Deserializer` instanc
...
+
+====
+[source, java]
+----
+package com.example;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+
+import org.springframework.kafka.support.serializer.JsonSerializer;
```
artembilan, line 3941: I think we can live without imports and `package` declarations in the samples shown in the docs...

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -310,6 +310,17 @@ public void setTypeMapper(Jackson2JavaTypeMapper typeMapper) {
 		}
 	}
 
+	/**
+	 * Configure the ObjectMapper before it is used to construct the ObjectWriter.
+	 * @param objectMapper configurable (or replaceable) ObjectMapper instance
+	 * @return configured (or replaced) ObjectMapper instance
+	 * @since 2.7.1
+	 */
+	public ObjectMapper customizeObjectMapper(ObjectMapper objectMapper) {
```
artembilan, line 319: I don't think this is what @garyrussell has meant in the issue.
We can't call non-`final` methods from the ctors.
Probably the point was to inject an `ObjectMapperCustomizer` into such a serialize impl...

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerializer.java
```diff
@@ -134,6 +134,17 @@ public void setUseTypeMapperForKey(boolean isKey) {
 		}
 	}
 
+	/**
+	 * Configure the ObjectMapper before it is used to construct the ObjectWriter.
+	 * @param objectMapper configurable (or replaceable) ObjectMapper instance
+	 * @return configured (or replaced) ObjectMapper instance
+	 * @since 2.7.1
+	 */
+	public ObjectMapper customizeObjectMapper(ObjectMapper objectMapper) {
```
garyrussell, line 143: This can be a `default` method on the interface.

Please change all sinces to 2.5.13; we will be backporting this to 2.6, 2.5.

artembilan (changes requested):
Thank you for contribution!

Please, find some review on code lines: I really have some doubts about proposed solution...

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
garyrussell, line 319: See comment in serializer - pull up to interface.

spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -3920,6 +3920,55 @@ They have no effect if you have provided `Serializer` and `Deserializer` instanc
...
 You can revert to the previous behavior by setting the `removeTypeHeaders` property to `false`, either directly on the deserializer or with the configuration property described earlier.
 
+[[serdes-customize-object-mapper]]
+====== Customize the ObjectMapper
+
+Occasionally, the only way to configure the way information is serialized is by customizing the ObjectMapper used in (de)serialization.
+Because Kafka uses its own ObjectMapper separate from the normal Spring-Boot ObjectMapper, customizing the ObjectMapper typically involved (re)implementing various Kafka factories.
+Since 2.7.1, a new interface `ObjectMapperCustomizer` has been introduced and implemented by the `JsonSerializer<T>` and `JsonDeserializer<T>` classes.
+This makes it simpler to configure the ObjectMapper without losing the convenience of the Spring property configuration capabilities.
+
```
garyrussell, line 3930: Suggested rewording - the Apache Foundation require us to mention Apache whenever we reference `Kafka`.

```
$ git diff
diff --git a/spring-kafka-docs/src/main/asciidoc/kafka.adoc b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
index 2b1cd6a2..faa0a1e1 100644
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -3795,6 +3795,7 @@ Also since version 2.7 `ConsumerPartitionPausedEvent` and `ConsumerPartitionResu
 [[serdes]]
 ==== Serialization, Deserialization, and Message Conversion
 
+[[serdes-overview]]
 ===== Overview
 
 Apache Kafka provides a high-level API for serializing and deserializing record values as well as their keys.
@@ -3924,8 +3925,9 @@ You can revert to the previous behavior by setting the `removeTypeHeaders` prope
 ====== Customize the ObjectMapper
 
 Occasionally, the only way to configure the way information is serialized is by customizing the ObjectMapper used in (de)serialization.
-Because Kafka uses its own ObjectMapper separate from the normal Spring-Boot ObjectMapper, customizing the ObjectMapper typically involved (re)implementing various Kafka factories.
-Since 2.7.1, a new interface `ObjectMapperCustomizer` has been introduced and implemented by the `JsonSerializer<T>` and `JsonDeserializer<T>` classes.
+When instantiated by Apache Kafka, the Spring JSON Serializer and Deserializer use a locally declared `ObjectMapper`.
+The only way to customize the `ObjectMapper` was to inject instances into the producer and consumer factories as discussed <<serdes-overview,here>> rather than by using properties.
+Since version 2.5.13, a new interface `ObjectMapperCustomizer` has been introduced and implemented by the `JsonSerializer<T>` and `JsonDeserializer<T>` classes.
 This makes it simpler to configure the `ObjectMapper` without losing the convenience of the Spring property configuration capabilities.
 
 The following is an example of how to provide a simple customized serializer class (derived from `JsonSerializer<T>`) to do the configuration.
@@ -3958,7 +3960,7 @@ public class CustomJsonSerializer<T> extends JsonSerializer<T> {
 ----
 ====
 
-Then in your properties file, configure your `CustomJsonSerializer<T>` class as the `value-serializer`:
+Then in your Spring Boot properties file, for example, configure your `CustomJsonSerializer<T>` class as the `value-serializer`:
 
 ====
 [source, properties]
```

garyrussell (changes requested):
Thanks for the contribution!

Just some minor comments.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
garyrussell: Good catch @artembilan ; yes a `Function<ObjectMapper, ObjectMapper>` setter would be safer.

cnygardtw: The alternative is to stop creating the ObjectWriter in the ctor and instead construct in the serialize() function (which is what JsonDeserializer does currently).  If this were done, then there's probably a better way to just hook into the configure() function.  The root cause was that the ObjectWriter was private, and there seemed no other way that wasn't brittle and likely to break in future versions.  If that's the preferred change, I can re-look at the implementation.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerializer.java
cnygardtw: That was my original thought, but I took guidance from JsonTypeResolver.  If we provide default implementation, then @FunctionalInterface goes away.  Is that what you intend?

garyrussell: I don't think this needs to be a `FunctionalInterface` the type resolver can be used as lambda expression, or method reference:

https://github.com/spring-projects/spring-kafka/blob/bbc5003cf7e26a3166924f06a2a407b33cf591eb/spring-kafka/src/test/java/org/springframework/kafka/support/serializer/JsonSerializationTests.java#L360

This is just a super interface for the (de)serializer.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
cnygardtw: As an alternative, I added `Map<String, ?>` to the customize function, and call it first thing from the configure() function instead of calling in the constructor.  I'm assuming that everywhere a serializer is instantiated, the configure() function will get called appropriately.  The other bit is to create the writer in serialize().  updated patch shortly.

garyrussell: Thanks; we have a day off tomorrow so it will probably be next week before we continue.

Thanks again.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -718,4 +717,27 @@ private JsonTypeResolver buildTypeResolver(String methodProperty) {
 		};
 	}
 
+	protected class CachedObjectReader {
```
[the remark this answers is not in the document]
cnygardtw: The problems are: 
- can't create ObjectWriter in ctor, because there's no way to provide customization of ObjectMapper in ctor
- can't create ObjectWriter in configure() function, because unit tests indicate it is not guaranteed that configure() will get called
- construct ObjectWriter at point of need, and possibly suffer performance issues

Therefore making no assumptions about how costly it is to construct an ObjectWriter.  Considering that the current serializer constructs the ObjectWriter just once in the constructor, I took the conservative approach and tried to at least cache the object so that if conditions are right it would still only get constructed once.  On the other hand, hashcode does have a cost, so if you don't think it matters to performance, I can pull it out.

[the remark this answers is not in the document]
artembilan: OK. Probably you just didn't see through my suggestion:

```
private volatile ObjectReader objectReader;
...
public T deserialize(...) {
    if (this.objectReader == null) {
		this.objectReader = this.objectMapper.readerFor(this.targetType);
	}
}
```

Does it make sense?
Technically your caching is just the same, but without `volatile`.
And that's why I don't see a reason in that extra instance if we just can deal with `volatile` and plain `if()`.

Sorry for not being clear enough from the beginning.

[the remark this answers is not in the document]
cnygardtw: That is certainly simpler, but I'm not familiar enough with all the supported code flows to decide if it's sufficient or not. 
 Are there any flows where a serializer might get reconfigured?  With this code, once the ObjectReader is constructed, changes to the ObjectMapper will not have any effect on the serialization (which was the original reason for the PR).  If that's an acceptable constraint, I'll make the change.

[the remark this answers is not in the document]
artembilan: I don't think so: you call `customizeObjectMapper()` from the `configure(Map<String, ?> configs, boolean isKey)` which is a Kafka Client API, which is called once from Kafka Client instance, e.g.:
```
 if (valueSerializer == null) {
                this.valueSerializer = config.getConfiguredInstance(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG,
                                                                                           Serializer.class);
                this.valueSerializer.configure(config.originals(Collections.singletonMap(ProducerConfig.CLIENT_ID_CONFIG, clientId)), false);
            } else {
                config.ignore(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG);
                this.valueSerializer = valueSerializer;
            }
```

And you see: they also set `config.ignore(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG);`, which means do not look into that property any more. So, any changes to the `ObjectMapper` at runtime are not going to be visible anyway.

Sounds reasonable?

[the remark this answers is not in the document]
cnygardtw: On the flip side, changes to the ObjectMapper may or may not be reflected in the hashcode.  If not, then the CachedObjectReader won't work any better (or worse) than a volatile.

artembilan (changes requested):
Looking closely to an issue I would say that it is probably better just to reject all of these changes in favor of plain Java inheritance:

````
public class CustomJsonSerializer extends JsonSerializer<Object> {

	public CustomJsonSerializer() {
		super(customizedObjectMapper());
	}

	private static ObjectMapper customizedObjectMapper() {
		ObjectMapper mapper = JacksonUtils.enhancedObjectMapper();
		mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
		return mapper;
	}

}
````

and this is going to work as expected for both Java configuration and Apache Kafka properties setting.

What am I missing?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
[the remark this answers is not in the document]
cnygardtw: ok, will update.

cnygardtw:
not sure, I'll try that also.

cnygardtw:
That worked as far as my tests required it.  Reject the PR.  Do you perhaps want a different PR with just documentation updates?

artembilan:
>  Do you perhaps want a different PR with just documentation updates?

I wonder why, since we talk just about a regular Java inheritance. 
But that still probably worth it if you have a time to contribute: with that in docs we may not have similar discussions in the future...

Thanks for understanding!

artembilan:
Sure thing! 

This chapter should be a good one for such a simple example: https://docs.spring.io/spring-kafka/docs/current/reference/html/#tips-n-tricks.

Also we won't mind if you add something for custom (de)serialization use-case into this directory: https://github.com/spring-projects/spring-kafka/tree/main/samples

But yeah: different story - deserves different PR!

garyrussell:
@cnygardtw Closing this; I opened this issue https://github.com/spring-projects/spring-kafka/issues/1784

Thanks for your efforts anyway.

Issue #748, opened 2018-07-20
`onMessage` methods in *MessageListener should allow for checked exception throwing


fcristovao:
One can use the `@KafkaListener` annotation in methods that throw checked exceptions.
However, when implementing the `*MessageListener` interfaces, none of the `onMessage` methods allow for (checked) Exceptions to be thrown.

Changing the signature to include `throws Exception` would prevent catching exceptions just to wrap them in `RuntimeException`.

garyrussell:
That would be a big breaking change; something we could not consider until the next major release (3.0).

antonio-tomac:
I would like to mention that checked exceptions are possible to be thrown when having listener written in Kotlin (I believe it could be done in Java if one were to use lombok's `@SneakyThrows`).
Current behaviour is that when such exception is thrown, it is not caught inside `doInvokeBatchListener()`/`doInvokeRecordListener()` and consequently, records are not passed to errorHandler. Instead, what happens is that this `Exception` is caught up in call stack in `KafkaMessageListenerContainer.run()` and passed to `handleConsumerException()`. đ
This is problematic because now, errorHadler/batchErrorHandler does not get records, and if one uses `SeekToCurrentBatchErrorHandler` then there is no seeking happening and consumer continues and without replaying messages that failed leading to data loss.

garyrussell:
@antonio-tomac I was not aware of that, I must say that I am appalled that Kotlin violates a fundamental assumption about bytecode. As for `@SneakyThrows`; I don't believe that deserves a comment by me.

However, I am equally appalled by a comment in the `@SneakyThrows` javadoc:

>The JVM does not check for the consistency of the checked exception system; javac does, and this annotation lets you opt out of its mechanism.

This is only a problem if you implement `MessageListener` (or one of its variants) directly.

For example, for POJO listeners (`@KafkaListener`), the framework wraps any exception thrown in a `ListenerExceptionFailedException` before it is thrown to the container.

I will (reluctantly) make the change, but in the meantime, you should do the same (catch and wrap in a `LEFE` in your listener).

artembilan:
@garyrussell , are we good closing this issue since your PR has been merged?

garyrussell:
No; I made a note in the PR; this has to remain open.

antonio-tomac:
@garyrussell Nice, thanks for quick reaction.

> you should do the same (catch and wrap in a LEFE in your listener).

[...]

PR #1783, opened 2021-04-27
GH-748: Handle Kotlin/Lombok Message Listeners

garyrussell:
See https://github.com/spring-projects/spring-kafka/issues/748#issuecomment-827333267 (issue not resolved by this)

Kotlin and Lombok (`@SneakyThrows`) can throw checked exceptions even though
the method signatures do not allow it. Such exceptions were not passed to the
error handlers until the failed record is no longer in scope. They were called
as if the consumer itself threw an exception.

Always wrap listener exceptions in `ListenerExecutionFailedException`s.

Also remove Java9 constructs in `@Deprecation`s.

**cherry-pick to 2.6.x, 2.5.x**


artembilan:
Cherry-picked into `2.6.x` & `2.5.x` as https://github.com/spring-projects/spring-kafka/commit/51ea8e380ab7e8b08775afcb5e9850324c3dba6a after resolving conflicts.

Issue #1787, opened 2021-05-05
Avoid Double Validation


garyrussell:
A multi-method listener that has no default handler will validate the payload twice (if a validator is provided).

Discussion here: https://github.com/spring-projects/spring-amqp/pull/1333#issuecomment-832853109

PR #1788, opened 2021-05-05
GH-1787: Only Validate Once

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1787

The validator was invoked twice with a class-level `@KafkaListener` when
no default `@KafkaHandler` is present.

**cherry-pick to all supported**


artembilan:
Cherry-picked to `2.6.x` & `2.5.x` as is.


Doesn't look like `2.4.x` or `2.3.x` have validation functionality yet...

Issue #1789, opened 2021-05-06
Use MessageConverter from spring messaging for content negotiation


dsyer:
It would be good to have a Kafka record converter that uses the Kafka headers and payload (per global serdes) but then delegates to Spring Messaging for the conversion to a type that matches the listener. Spring already has a `CompositeMessageConverter` that can handle some of the logic. As a user I would like to have some reasonable defaults and then simply  add beans of type MessageConverter to express my preferences.

garyrussell:
Needs docs.

PR #1791, opened 2021-05-07
GH-1789: Content Negotiation for Conversion

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1789

Default message converter can now delegate to spring-messaging `SmartMessageConverter`
for payload conversion.

Add a spring-messaging implementation of the `JsonMessageConverter`.

**Needs doc before merge**


artembilan (approved):
LGTM.
Feel free to go with docs!
Or we just can merge as is, so Dave will be able to play with it on Monday.

garyrussell:
OK; let's do that (and re-open the issue for the docs).

PR #1797, opened 2021-05-11
GH-1789: Content Type Conversion (Outbound)

garyrussell:
See https://github.com/spring-projects/spring-kafka/issues/1789

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
@@ -223,6 +223,12 @@ public class MessagingMessageConverter implements RecordMessageConverter {
 			return null;
 		}
 		else {
+			if (this.messagingConverter != null) {
+				Message<?> message2 = this.messagingConverter.toMessage(payload, message.getHeaders());
+				if (message2 != null) {
+					return message2.getPayload();
+				}
+			}
 			return payload;
 		}
 	}
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -54,6 +54,7 @@ import org.springframework.kafka.support.converter.RecordMessageConverter;
 import org.springframework.kafka.support.micrometer.MicrometerHolder;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
+import org.springframework.messaging.converter.SmartMessageConverter;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.springframework.util.Assert;
 import org.springframework.util.concurrent.ListenableFuture;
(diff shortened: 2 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
```diff
@@ -223,6 +223,12 @@ protected Object convertPayload(Message<?> message) {
 			return null;
 		}
 		else {
+			if (this.messagingConverter != null) {
+				Message<?> message2 = this.messagingConverter.toMessage(payload, message.getHeaders());
```
artembilan, line 227: Why do we do `fromMessage()` not `toMessage()`?
I also see that in the `extractAndConvertValue()` we use `fromMessage()`, but not `to`...

What am I missing? Why exactly this way? It may also confuse end-users who are going to supply their own `SmartMessageConverter` impl...

Thanks

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
garyrussell: It looks bizarre; I know.

These converters assume the source (in fromMessage) is a `Message<?>` with a serialized payload and we want to convert it to an object; similarly toMessage is assuming want to create a `Message<?>` with a serialized payload.

Hence, I had to create a dummy message and use `fromMessage` in `extractAndConvertValue`

See the `MappingJackson2Converter`. 

I think it's because spring-messaging works that way - it converts from a serialized message payload to a POJO to call a method; similarly, it takes a POJO and creates a message with a serialized payload.

In our case, it's the other way around.

I will clearly have to document this (when I do the docs(.

artembilan: OK. Let's probably just document in in the JavaDocs!
Maybe we can rework it other way in the next version. Although Dave's idea is to reuse Spring Boot's auto-configured converter which is essentially that `SmartMessageConverter` anyway...

dsyer: Spring Kafka already has a converter for the last hop - to and from a Kafka record and a `Message` - and the value type of the record is fixed globally by the Kafka client configuration. I might be missing something, but in Spring Messaging `fromMessage()` is generally for inbound, and `toMessage()`is for outbound, so isn't that exactly what we need here?

garyrussell: That's correct; the reason for the "odd" behavior that @artembilan pointed out is that I wanted to avoid creating a message and then creating another one after converting the raw payload (we still need a message for the handler invoker).

Hence, I call the messaging converters "before" there is a real message on inbound, just to convert the payload.

But, since I have to create (and discard) a dummy message anyway (each way), I guess it won't hurt to move the code so that it's more logical.

We'll still have to explain in docs that `KafkaMessageConverter.toMessage` will call `fromMessage` on the messaging converter.

Issue #1808, opened 2021-05-21
Impossible to create Listener Containers with different Kafka consumer properties


sergeylemekhov:
**Affects Version(s):** 2.7.0

🎁 Enhancement

When manually creating a listener container via listener container factory `org.springframework.kafka.config.AbstractKafkaListenerContainerFactory#createContainer(java.lang.String...)` method the listener container is initialized with properties from the factory.
The `org.springframework.kafka.config.AbstractKafkaListenerContainerFactory#initializeContainer` method sets listener container's `kafkaConsumerProperties` as a reference to the factory `consumerProperties`.
When calling code then changes `kafkaConsumerProperties` of the created listener container the changes affect all other listener containers which is the problem.

Because of this it is impossible to create two listener containers with different offset reset policies for example:
```
ConcurrentMessageListenerContainer<String, String> container1 = listenerContainerFactory.createContainer("topic");
container1.getContainerProperties().setGroupId("group1");
container1.getContainerProperties().getKafkaConsumerProperties().setProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "lateset");

ConcurrentMessageListenerContainer<String, String> container2 = listenerContainerFactory.createContainer("topic");
container2.getContainerProperties().setGroupId("group2");
container2.getContainerProperties().getKafkaConsumerProperties().setProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");

```
Both listeners in the snippet end up with "earliest" offset reset policy.

A way to fix this would be creating a copy of factory `consumerProperties` before setting it for the listener container.

garyrussell:
Something is off here;  each container gets a new `ContainerProperties`:

https://github.com/spring-projects/spring-kafka/blob/bd154520a5cc7454748b4ca24e692d82594edd6a/spring-kafka/src/main/java/org/springframework/kafka/config/ConcurrentKafkaListenerContainerFactory.java#L69-L71

We then copy properties from the factory's properties into the instance properties:

https://github.com/spring-projects/spring-kafka/blob/bd154520a5cc7454748b4ca24e692d82594edd6a/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java#L403-L433

So it's not at all clear how we could get cross-talk between container properties.

Please provide a minimal project that exhibits the behavior you observe.

garyrussell:
[...]

PR #1809, opened 2021-05-21
GH-1808: Fix Container Properties Crosstalk

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1808

Each container got a reference to the same `Properties` object in
`kafkaConsumerProperties`, unless it was created via a `@KafkaListener`
with property overrides.

Do not copy the property from the factory's properties.

**Cherry-pick to all supported branches**


sergeylemekhov:
@garyrussell please also see this PR https://github.com/spring-projects/spring-kafka/pull/1810
there is an additional properties initialization and a separate test method

artembilan:
... and cherry-picked to all the supported versions.

PR #1810, opened 2021-05-21
1808: use a separate Properties instance for kafka consumer properties

sergeylemekhov:


artembilan:
Thank you for contribution, @sergeylemekhov !

Please, take a look what @garyrussell has just PR'ed this morning #1809.

Looks like you both did similar fix.

Please, review and let me know which PR I should close as duplicated.

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
```diff
@@ -402,7 +403,8 @@ private void configureEndpoint(AbstractKafkaListenerEndpoint<K, V> aklEndpoint)
 	protected void initializeContainer(C instance, KafkaListenerEndpoint endpoint) {
 		ContainerProperties properties = instance.getContainerProperties();
 		BeanUtils.copyProperties(this.containerProperties, properties, "topics", "topicPartitions", "topicPattern",
-				"messageListener", "ackCount", "ackTime", "subBatchPerPartition");
+				"messageListener", "ackCount", "ackTime", "subBatchPerPartition", "kafkaConsumerProperties");
+		properties.setKafkaConsumerProperties(new Properties(this.containerProperties.getKafkaConsumerProperties()));
```
artembilan, line 407: I comment out this line of code and no one test fails, even your new one.
What am I missing in the logic?
Thanks

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
garyrussell: This code is not necessary; the constructor initializes the field with a new `Properties`.

artembilan: OK. Merging your PR then, Gary.

artembilan:
Duplicates #1809.

@sergeylemekhov , 

thank you for effort any way; looking forward for more contribution!

Issue #1811, opened 2021-05-24
NPE createListenerContainer in Kotlin


lukas-krecan:
🐞 Bug report

When using `AbstractKafkaListenerContainerFactory.createListenerContainer` from Kotlin getting 

```
java.lang.NullPointerException: Parameter specified as non-null is null: method XXXX.setupListenerContainer, parameter messageConverter
```

Caused by missing Nullable annotation. Which forces Kotlin implementations to make the param non-nullable which in turn fails when called with null.

lukas-krecan:
See https://github.com/spring-projects/spring-kafka/pull/1812

PR #1812, opened 2021-05-24
#1811 Make messageConverter parameter nullable

lukas-krecan:
It fails in Kotlin when `messageConverter` is null


garyrussell (changes requested):
Just setting it here is not enough (although it might solve the NPE), we need to propagate it to the implementations and all downstream uses of that argument in other method calls.

lukas-krecan:
Sure, feel free to close it, it was just an easy way for me tho show what's wrong.

garyrussell:
Thanks for the contribution; replaced by https://github.com/spring-projects/spring-kafka/pull/1814

Issue #1817, opened 2021-06-01
Allow Users to Post Process @KafkaListener


garyrussell:
See https://stackoverflow.com/questions/67777974/kafkalistener-generate-clientid-and-groupid-by-incoming-parameters/67795364#67795364

PR #1818, opened 2021-06-01
GH-1817: Allow Annotation Attribute Modification

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1817

### DIFF
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -1863,6 +1863,36 @@ void listen(Object in, @Header(KafkaHeaders.RECORD_METADATA) ConsumerRecordMetad
 ----
 ====
 
+[[kafkalistener-attrs]]
+===== `@KafkaListener` Attribute Modification
+
+Starting with version 2.7.2, you can now programmatically modify annotation attributes before the container is created.
+To do so, override the framework's `KafkaListenerAnnotationBeanPostProcessor` bean and add an `AnnotationEnhancer` to it.
+`AnnotationEnhancer` is a `BiFunction<Annotation, AnnotatedElement, Annotation>` and must return an annotation.
+The attributes can contain SpEL and/or property placeholders; the enhancer is called before any resolution is performed.
+
+An example follows:
+
+====
+[source, java]
+----
+@Bean(KafkaListenerConfigUtils.KAFKA_LISTENER_ANNOTATION_PROCESSOR_BEAN_NAME)
+public static KafkaListenerAnnotationBeanPostProcessor<Object, Object> bpp() {
+    KafkaListenerAnnotationBeanPostProcessor<Object, Object> bpp =
+            new KafkaListenerAnnotationBeanPostProcessor<>();
+    bpp.setEnhancer((ann, element) -> {
+        Map<String, Object> attrs = AnnotationUtils.getAnnotationAttributes(ann);
+        attrs.put("groupId", attrs.get("id") + "." + (element instanceof Class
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -1070,4 +1089,14 @@ protected boolean isEmptyPayload(Object payload) {
 
 	}
 
+	/**
+	 * Post processes each annotation.
+	 *
+	 * @since 2.7.2
+	 *
+	 */
+	public interface AnnotationEnhancer extends BiFunction<KafkaListener, AnnotatedElement, KafkaListener> {
```
artembilan, line 1098: OK. You have introduced a more specific type, but you don't mention it in the docs I've commented above.
Thanks

artembilan (changes requested):
Here is some review, but I'm OK if you convince me that the fix must be really your way.

Thanks

spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -1863,6 +1863,36 @@ void listen(Object in, @Header(KafkaHeaders.RECORD_METADATA) ConsumerRecordMetad
...
+public static KafkaListenerAnnotationBeanPostProcessor<Object, Object> bpp() {
+    KafkaListenerAnnotationBeanPostProcessor<Object, Object> bpp =
+            new KafkaListenerAnnotationBeanPostProcessor<>();
+    bpp.setEnhancer((ann, element) -> {
+        Map<String, Object> attrs = AnnotationUtils.getAnnotationAttributes(ann);
+        attrs.put("groupId", attrs.get("id") + "." + (element instanceof Class
+                ? ((Class<?>) element).getSimpleName()
+                : ((Method) element).getDeclaringClass().getSimpleName()
+                        +  "." + ((Method) element).getName()));
+        return AnnotationUtils.synthesizeAnnotation(attrs, KafkaListener.class, null);
```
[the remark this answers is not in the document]
garyrussell: Good points - will rework.

spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -1867,32 +1867,32 @@ void listen(Object in, @Header(KafkaHeaders.RECORD_METADATA) ConsumerRecordMetad
...
-    });
-    return bpp;
+        return attrs;
+    };
 }
 ----
 ====
 
+IMPORTANT: `AnnotationEnhancer` bean definitions cannot be declared in the same class as a `@KafkaListener` method (or class).
+It will prevent those listeners from being found by the `KafkaListenerAnnotationBeanPostProcessor`.
```
artembilan, line 1894: Probably better to say that such a bean must be declared `static` since it is used from the `BeanPostProcessor`?
Or... The logic in the `KafkaListenerAnnotationBeanPostProcessor` must be reworked from that `InitializingBean`. 
We probably can collect `AnnotationEnhancer` beans in the beginning of the `afterSingletonsInstantiated()` impl...

artembilan (changes requested):

spring-kafka-docs/src/main/asciidoc/kafka.adoc
garyrussell: That is where I had it initially, but it is too late. The beans have all been preprocessed by then. I will look to see if static helps tomorrow.

Issue #1745, opened 2021-03-24
RetryTopicConfigurer / DestinationTopicProcessor refactor


tomazfernandes:
When these classes were first created one of the goals was to change existing Spring Kafka classes as little as possible in order to reduce the risk of breaking existing functionalities, and that led to lots of indirections and complexities in the `RetryTopicConfigurer` class. 

I've since then made one large refactor of this part which has mitigated this issue a lot, but I still think both this class and the `DestinationTopicProcessor` interface are a lot more complex than they should be.

This is the last major refactor I have in mind, and I think it should solve probably be the last maintenance pain point for this functionality as a whole.

This is not as important as the other issue I've just opened about KafkaConsumerBackoffManager, but it's something I'd like to do after that, if that's ok.

Please let me know what you think.

Thanks!

apolci:
I just had to look at the internals of those two classes in search for a possible extension point (see #1772) and I can confirm your statement.
I planned to try to make a pull request for my feature request (not sure that I'm able to do it), but I fear it would conflict with this refactoring.

tomazfernandes:
You're welcome to raise the PR! And since it looks like a really necessary feature, it's something I can work on as well if you prefer. Either way, I think we can hold this refactor for a while in the meantime and just work on this part. Let me know what you think, and thanks again for bringing this up!

PR #1821, opened 2021-06-04
GH-1745 - Improve RetryTopicConfigurer flexibility

tomazfernandes:
Just opening some doors for advanced users who want to have more control over the RetryTopicConfigurer. Since the PR from @apolci already covered many use cases through DI it didn't look like it was worth having another DI setup for this.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2021 the original author or authors.
+ * Copyright 2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,6 @@
 package org.springframework.kafka.retrytopic;
 
 import java.util.Collection;
-import java.util.function.Function;
 
 import org.springframework.kafka.config.MethodKafkaListenerEndpoint;
 
@@ -31,19 +30,15 @@ import org.springframework.kafka.config.MethodKafkaListenerEndpoint;
  * @see EndpointCustomizerFactory
  *
  */
-interface EndpointCustomizer extends Function<MethodKafkaListenerEndpoint<?, ?>,
-		Collection<EndpointCustomizer.TopicNamesHolder>> {
+@FunctionalInterface
+interface EndpointCustomizer {
 
 	/**
-	 *
-	 * Customizes the endpoint and returns the topic names generated for this endpoint.
-	 *
+	 * Customize the endpoint and return the topic names generated for this endpoint.
 	 * @param listenerEndpoint The main, retry or DLT endpoint to be customized.
 	 * @return A collection containing the topic names generated for this endpoint.
 	 */
-	default Collection<TopicNamesHolder> customizeEndpointAndCollectTopics(MethodKafkaListenerEndpoint<?, ?> listenerEndpoint) {
-		return apply(listenerEndpoint);
-	}
+	Collection<TopicNamesHolder> customizeEndpointAndCollectTopics(MethodKafkaListenerEndpoint<?, ?> listenerEndpoint);
 
 	class TopicNamesHolder {
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizer.java
@@ -31,7 +31,7 @@ import org.springframework.kafka.config.MethodKafkaListenerEndpoint;
  *
  */
 @FunctionalInterface
-interface EndpointCustomizer {
+public interface EndpointCustomizer {
 
 	/**
 	 * Customize the endpoint and return the topic names generated for this endpoint.
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizer.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2018-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizer.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2018-2021 the original author or authors.
```
artembilan, line 2: Well, the point of Copyright to identify a lifespan of the class.
I don't think its license can somehow affect `2018` since there was not that class yet.
So, our preference just have it in the current year and then change the last one in the future if we really do some changes to that class.

I may not understand the licensing properly, but for consistency I would like to see a Copyright for new classes like `Copyright 2021 the original...`

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizerFactory.java
```diff
@@ -0,0 +1,107 @@
...
+	EndpointCustomizerFactory(DestinationTopic.Properties destinationProperties, EndpointHandlerMethod beanMethod,
+							BeanFactory beanFactory, RetryTopicNamesProviderFactory retryTopicNamesProviderFactory) {
+
+		this.destinationProperties = destinationProperties;
+		this.beanMethod = beanMethod;
+		this.beanFactory = beanFactory;
+		this.retryTopicNamesProviderFactory = retryTopicNamesProviderFactory;
+	}
+
+	public EndpointCustomizer createEndpointCustomizer() {
```
artembilan, line 60: Maybe your idea was to have this method as `final` since you are exposed those `protected` methods for custom logic...

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizerFactory.java
```diff
@@ -0,0 +1,107 @@
...
+			TopicPartitionOffset[] topicPartitionsToAssign = endpoint.getTopicPartitionsToAssign();
+			if (topicPartitionsToAssign != null && topicPartitionsToAssign.length > 0) {
+				topics = Arrays.stream(topicPartitionsToAssign)
+						.map(TopicPartitionOffset::getTopic)
+						.collect(Collectors.toList());
+			}
+		}
+
+		if (topics.isEmpty()) {
+			throw new IllegalStateException("No topics where provided for RetryTopicConfiguration.");
```
artembilan, line 103: Maybe we can give end-user more clues what is going on here, e.g. print those `destinationProperties` for context. This way, with a comprehensive exception message in the logs, end-user would be able to identify what exactly configuration is wrong.

artembilan (changes requested):
I might not see the whole picture, but my review is more about common issues and general design decisions.

Thank you for understanding!

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizerFactory.java
garyrussell: Also `were` not `where`.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizer.java
```diff
@@ -0,0 +1,67 @@
...
+ * Customizes main, retry and DLT endpoints in the Retry Topic functionality
+ * and returns the resulting topic names.
+ *
+ * @author Tomaz Fernandes
+ * @since 2.7.2
+ *
+ * @see EndpointCustomizerFactory
+ *
+ */
+interface EndpointCustomizer extends Function<MethodKafkaListenerEndpoint<?, ?>,
```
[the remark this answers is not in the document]
tomazfernandes: You're totally right, sorry about that!

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizerFactory.java
tomazfernandes: Good call. I think for the end user the properties themselves won't help much, so I added the endpoint method and bean class for context, let me know if you prefer otherwise.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizer.java
```diff
@@ -31,19 +30,15 @@
  * @see EndpointCustomizerFactory
  *
  */
-interface EndpointCustomizer extends Function<MethodKafkaListenerEndpoint<?, ?>,
-		Collection<EndpointCustomizer.TopicNamesHolder>> {
+@FunctionalInterface
+interface EndpointCustomizer {
```
artembilan, line 34: Why this is still not `public`?

Can you move your tests to different package to see the problem?

artembilan (changes requested):

tomazfernandes:
Hi @artembilan and @garyrussell, thanks a lot for your input! I've made the changes as suggested, please let me know if there's anything else.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizer.java
tomazfernandes: Sorry again, got distracted by the other changes. Fixed it now.

tomazfernandes:
Is there any known issues with this test? Doesn't look like anything related to this PR.

org.springframework.kafka.core.reactive.ReactiveKafkaProducerTemplateTransactionIntegrationTests.shouldSendOneRecordTransactionallyViaTemplateAsSenderRecordAndReceiveItExactlyOnce(ReactiveKafkaProducerTemplateTransactionIntegrationTests.java:319)

garyrussell:
@tomazfernandes Yes, it's a known flaky test; the latest commit on main has a fix for it.

https://github.com/spring-projects/spring-kafka/commit/ad1ae60cd8fe98333b16696ba4c37aa5ad07a43a

Try rebasing your PR.

garyrussell:
Hmmm - looks like my "fix" just moved the flakiness someplace else 😦

PR #1823, opened 2021-06-04
Fix for Issue #1482 - Allow update of topic-config during topic modification via NewTopic 

### LINKED ISSUE #1482: Allow update of retention.ms during topic modification via NewTopic
**Affects Version(s):** 2.3.7.RELEASE

I'm trying to modify topic configuration after the creation of the topic (mainly retention.ms in my case, but why not other properties).  

When creating a NewTopic, configuration is set correctly at creation by KafkaAdmin.addTopics. When increasing the partitions in a NewTopic, after creation, they are updated automatically by KafkaAdmin.modifyTopics.

The idea would be to update KafkaAdmin.modifyTopics to also modify retention.ms (or possibly others in NewTopics.configs).

blianix:
Resolves #1482.
Needs refactoring, but wanted to share the code already to get some feedback.


garyrussell (commented):
Looks fine to me; just one minor comment about formatting - thanks!

Please proceed to make any more changes you are planning - we prefer to add more commit(s) to the PR rather than squashing, so we can easily see the incremental changes.

Thanks

PR #1829, opened 2021-06-14
GH-1828: Use consumer record's partition for @RetryableTopic

deepesh-verma:
Resolves https://github.com/spring-projects/spring-kafka/issues/1828

**Update DeadLetterPublishingRecovererFactory#resolveDestination**
- Extract out a protected method to `resolveTopicPartition` for destination DLT
- Set the partition to the same partition as the original record

**Test Changes**
- Add tests for `RetryableTopic` with `autoCreate=false`
- Remove unnecessary mockito stubbing from tests

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
@@ -118,6 +118,19 @@ public class DeadLetterPublishingRecovererFactory {
 					: resolveTopicPartition(cr, nextDestination);
 	}
 
+	/**
+	 * Creates and returns the {@link TopicPartition}, where the original record should be forwarded.
+	 * By default, it will use the partition same as original record's partition, in the next destination topic.
+	 *
+	 * <p>{@link DeadLetterPublishingRecoverer#checkPartition} has logic to check whether that partition exists,
+	 * and if it doesn't it sets -1, to allow the Producer itself to assign a partition to the record.</p>
+	 *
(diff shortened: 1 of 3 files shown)


tomazfernandes:
Looks good to me, thanks!

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
```diff
@@ -115,8 +115,11 @@ private TopicPartition resolveDestination(ConsumerRecord<?, ?> cr, Exception e)
 
 		return nextDestination.isNoOpsTopic()
 					? null
-					: new TopicPartition(nextDestination.getDestinationName(),
-				cr.partition() % nextDestination.getDestinationPartitions());
+					: resolveTopicPartition(cr, nextDestination);
+	}
+
+	protected TopicPartition resolveTopicPartition(final ConsumerRecord<?, ?> cr, final DestinationTopic nextDestination) {
```
garyrussell, line 121: Needs javadoc since it's a public API.

It should explain the default resolution algorithm.

garyrussell (changes requested):
Thanks for the contribution!

Although this is technically a breaking change (a user might have configured retry topics with only one partition) we are saved by the DLPR falling back to `null` for an invalid partition. (S)he'll get a bunch of warning logs, but I think the chances are small.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
deepesh-verma: Added javadoc. Thanks!

Issue #1827, opened 2021-06-11
There is no easy way to set up the order of listening to topics


BOOMeranGG:
I used spring-kafka in my work, and I had a need to configure the order of starting listening to topics. I mean, start listening to topic B only after all the messages from topic A have been processed.
At the moment, there is no easy way to do this (like setting up the order in @KafkaListener). Please correct me if I'm wrong.

If you think that this functionality is worth implementing, I can try to do it.

garyrussell:
You can add 2 `@KafkaListener` annotations; set the one for topic B `autoStartup = "false"`.

For the first container enable idle container events and add an event listener.

When the first container goes idle, in the event listener, stop the first container and start the second one.

BOOMeranGG:
Yes, I have implemented as you wrote. But it becomes not so convenient when there are several such dependencies. Topic A is waiting for B, C, D, topic E is waiting for F and G, e.t.c.
Recently I had to implement this for the second time, and I think it would be convenient to specify in @KafkaListener.
For example `initializationOrder=2`, and the topic will start only when other topics with `initializationOrder=1` finish receiving data.

garyrussell:
We already have a property `containerGroup`; I think a new component would be simpler, and easier to configure/understand. Something like this...

```java
/**
 * Sequence the starting of container groups when all containers in the previous group are
 * idle.
 *
 * @author Gary Russell
 * @since 2.8
 *
 */
public abstract class ContainerGroupSequencer {

	/**
	 * Set containers in each group to not auto start. Start the containers in the first
	 * group. Start containers in group[n] when all containers in group[n-1] are idle;
	 * stop the containers in group[n-1].
	 * @param registry the registry.
	 * @param defaultIdleEventInterval the idle event interval if not already set.
	 * @param containerGroups The list of container groups, in order.
	 */
	public ContainerGroupSequencer(KafkaListenerEndpointRegistry registry, long defaultIdleEventInterval,
			String... containerGroups) {

		// ...
	}

}
```

BOOMeranGG:
Спасибо! Это похоже на то, что мне нужно

garyrussell:
I created a PR; sorry, I didn't notice your comment about submitting a change.

It turned out to be a bit more complicated than I thought it would be due to some initialization timing issues.

PR #1832, opened 2021-06-14
GH-1827: Add ContainerGroupSequencer

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1827

Add a mechanism to automatically control a sequence of containers,
starting a group of containers when the current group goes idle.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -81,6 +81,7 @@ import org.springframework.kafka.config.KafkaListenerEndpointRegistrar;
 import org.springframework.kafka.config.KafkaListenerEndpointRegistry;
 import org.springframework.kafka.config.MethodKafkaListenerEndpoint;
 import org.springframework.kafka.config.MultiMethodKafkaListenerEndpoint;
+import org.springframework.kafka.listener.ContainerGroupSequencer;
 import org.springframework.kafka.listener.KafkaListenerErrorHandler;
 import org.springframework.kafka.retrytopic.RetryTopicBootstrapper;
 import org.springframework.kafka.retrytopic.RetryTopicConfiguration;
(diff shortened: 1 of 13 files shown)


spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -305,6 +306,9 @@ public void afterSingletonsInstantiated() {
 
 		// Actually register all listeners
 		this.registrar.afterPropertiesSet();
+		Map<String, ContainerGroupSequencer> sequencers =
+				this.applicationContext.getBeansOfType(ContainerGroupSequencer.class, false, false);
+		sequencers.values().forEach(seq -> seq.initialize());
```
artembilan, line 311: Doesn't look like you do anything else but just that `initialize`.
Why `ContainerGroupSequencer` can't implement a `SmartInitializingSingleton` then instead?

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerGroupSequencer.java
```diff
@@ -0,0 +1,190 @@
...
+import org.springframework.core.task.TaskExecutor;
+import org.springframework.kafka.config.KafkaListenerEndpointRegistry;
+import org.springframework.kafka.event.ListenerContainerIdleEvent;
+
+/**
+ * Sequence the starting of container groups when all containers in the previous group are
+ * idle.
+ *
+ * @author Gary Russell
+ * @since 2.8
```
artembilan, line 41: `2.8` ?

spring-kafka/src/test/resources/log4j2-test.xml
```diff
@@ -6,7 +6,7 @@
 		</Console>
 	</Appenders>
 	<Loggers>
-		<Logger name="org.springframework.kafka" level="warn"/>
+		<Logger name="org.springframework.kafka" level="debug"/>
```
artembilan, line 9: Do we still need this in `debug`?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
garyrussell: I tried that but there is no way to control the order, and the `ContainerGroupSequencer.aSI()` ran before this one - the `afterPropertiesSet()` creates the group beans so we found none there; so I had to move it here.

Issue #1804, opened 2021-05-17
Documentation: Clarify best practices and pitfalls for retrying consumption of messages


jacktwilliams:
**Affects Version(s):** 2.7

Looking at this documentation: https://docs.spring.io/spring-kafka/reference/html/#retrying-deliveries

## The desired use-case
* Retry some types of exceptions never.
* Retry some exceptions a fixed number of times
* Retry some exceptions indefinitely

## The current problem
The documentation focuses on configuring a RetryTemplate on the KafkaListenerContainerFactory. This seems to present numerous ways to achieve the use case. However, the default error handler, a SeekToCurrentErrorHandler, also has retry logic, and these two functionalities interact, in a way not described, that leads to unexpected behavior. The documentation doesn't adequately warn you of this, nor prescribe best practices for dealing with this. 

For example, I configured a RetryTemplate with an ExceptionClassifierRetryPolicy that said I should retry consuming the message 3 times. What actually happened, was I think it retries like thirteen times, because the SeekToCurrentErrorHandler does 10 retries by default. (This part may be wrong, but something like this.) Using the ExceptionClassifierRetryPolicy to specify that you want to indefinitely retry an error doesn't lead to infinite retries, because of the default error handler. 

I think Spring maintainers comment on this issue in the comments on this stack overflow answer
https://stackoverflow.com/a/58433991/9790600
and in this issue thread
https://github.com/spring-projects/spring-kafka/issues/968#issuecomment-472479560

jacktwilliams:
It seems that the simplest approach is to only configure the SeekToCurrentErrorHandler, however the interface isn't quite as nice as RetryTemplate. Outside the scope of this issue, but it does seem (from an outsider) that configuring the RetryTemplate only would be more ideal. 

Anyway, I would love to write a couple of improvement issues for SeekToCurrentErrorHandler, assuming the format of this issue was ok.

garyrussell:
The retry template mechanism has indeed been replaced by the STCEH as more and more functionality has been added there. I considered deprecating use of the retry template because it causes a rebalance with long aggregate delays, but decided not to because there may be valid use cases.

Feel free to make suggestions about improving configurability.

jacktwilliams:
[...]

PR #1833, opened 2021-06-15
GH-1804: Add ExponentialBackOffWithMaxRetries

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1804

Simply a convenience to automatically calculate the max elapsed time.


artembilan:
I have merged it, but sounds more like an improvement for existing `ExponentialBackOff`. WDYT?

garyrussell:
Agreed; I just thought this would be faster. It would also be a bit more complicated to modify the existing back off; it was easy to reject conflicting configuration this way.

garyrussell:
We can revert this if [this](https://github.com/spring-projects/spring-framework/pull/27071) is accepted before 2.7.3.

Issue #615, opened 2018-03-21
Collapse ErrorHandler and BatchErrorHandler interface hierarchies into a single interface for each


garyrussell:

garyrussell:
Closing as won't fix.

garyrussell:
Reopening: https://github.com/spring-projects/spring-kafka/pull/1837

PR #1837, opened 2021-06-21
Data in handle method is @Nullable

kkocel:
Motivation - Kotlin uses jsr305 annotations to deduct non-null / nullable types. This PR marks data as explicitly nullable (which happen in code).

Without this it's impossible to pass null value as `data` parameter in Kotlin.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchLoggingErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchLoggingErrorHandler.java
@@ -23,6 +23,8 @@ import org.apache.kafka.clients.consumer.ConsumerRecords;
 
 import org.springframework.core.log.LogAccessor;
 
+import org.springframework.lang.Nullable;
+
 /**
  * Simple handler that invokes a {@link LoggingErrorHandler} for each record.
  *
@@ -35,7 +37,7 @@ public class BatchLoggingErrorHandler implements BatchErrorHandler {
 			new LogAccessor(LogFactory.getLog(BatchLoggingErrorHandler.class));
 
 	@Override
-	public void handle(Exception thrownException, ConsumerRecords<?, ?> data) {
+	public void handle(Exception thrownException, @Nullable ConsumerRecords<?, ?> data) {
 		StringBuilder message = new StringBuilder("Error while processing:\n");
 		if (data == null) {
 			message.append("null ");
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchLoggingErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchLoggingErrorHandler.java
@@ -22,7 +22,6 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 
 import org.springframework.core.log.LogAccessor;
-
 import org.springframework.lang.Nullable;
 
 /**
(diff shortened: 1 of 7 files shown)


artembilan (changes requested):
Please, fix Checkstyle issue: 
```
Error: eckstyle] [ERROR] /home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchLoggingErrorHandler.java:26:1: Extra separation in import group before 'org.springframework.lang.Nullable' [ImportOrder]
```

You can check it locally with Gradle `check` task.

Thanks

kkocel:
@artembilan Hi, fixed right away. Sorry for inconvienience.

artembilan (approved):
I'm OK with this change: doesn't look like it breaks something.
Although I have doubts if we should propagate something reasonable in those default methods instead of `null`.

Let's see what @garyrussell thinks on the matter!

garyrussell:
@kkocel Under what circumstances are you seeing it being called with `null` values?

Even with consumer exceptions (before we receive any `ConsumerRecords`), the container calls it with empty objects.

https://github.com/spring-projects/spring-kafka/blob/e624891e424a28a53ea95081c742c1d566b94f52/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1597-L1605

This is misleading...

https://github.com/spring-projects/spring-kafka/blob/e624891e424a28a53ea95081c742c1d566b94f52/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandler.java#L39-L41

All error handlers provided by the framework implement this method so it is never called.

delor:
If you implement `ErrorHandler` in Kotlin like so
```kotlin
import org.apache.kafka.clients.consumer.ConsumerRecord
import org.springframework.kafka.listener.ErrorHandler
import java.lang.Exception

class MyErrorHandler : ErrorHandler {
    override fun handle(thrownException: Exception, data: ConsumerRecord<*, *>) {
        TODO("Not yet implemented")
    }
}
```
parameter `data` can be null b/c of `@NonNullApi`. Then `org.springframework.kafka.listener.ErrorHandler#handle` will pass null to it and you will get exception in runtime b/c `data` parameter in `MyErrorHandler` can't be null.

kkocel:
@garyrussell having proper annotations is crucial for having proper types in Kotlin, as you can run into a problem @delor stated. More info here: https://www.youtube.com/watch?v=DvqXdnzhlSc&t=622s

artembilan:
Thank you for the patience, @kkocel, and info!

May you share with us a stack trace for this issue to have a whole picture?
We definitely can live with your change since it is not a breaking change for Java at all.
At the same time it makes Kotlin happy.

Nevertheless I still think that we may revise those `default` method impls, but that already going to some breaking change in the API. Therefore next framework version.

garyrussell:
>crucial for having proper types in Kotlin,

I fully understand that.

I believe we should deprecate that method; users should always implement `ContainerAwareErrorHandler` or `ConsumerAwareBatchErrorHandler` since the container always calls the method with 3 parameters.

Making it `@Nullable` seems wrong to me since it will never be called with a `null` value by the container.

delor:
Sorry, can't provide a stacktrace as we have moved on with different approach and can't recreate it.

garyrussell:
I have tried several different ways to fix this in a non-breaking way without success, so I think I will go ahead and merge this; thanks for the contribution.

I have re-opened this issue https://github.com/spring-projects/spring-kafka/issues/615 so we can rework this error handler hierarchy in a breaking way in 3.0.

Issue #1842, opened 2021-06-24
Add ConditionalDelegatingErrorHandler


garyrussell:
Delegate to different error handlers based on the exception type.

See https://stackoverflow.com/questions/68118649/how-to-process-all-messages-between-two-timestamps-with-spring-kafka/68118744#68118744

garyrussell:
Use assignable instead of hard equals.

PR #1843, opened 2021-06-24
GH-1842: Add Conditional Delegating Error Handlers

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1842

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConditionalDelegatingBatchErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConditionalDelegatingBatchErrorHandler.java
@@ -23,6 +23,7 @@ import java.util.Map.Entry;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 
+import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ConditionalDelegatingBatchErrorHandler.java
```diff
@@ -0,0 +1,110 @@
...
+		if (cause instanceof ListenerExecutionFailedException) {
+			cause = thrownException.getCause();
+		}
+		if (cause != null) {
+			Class<? extends Throwable> causeClass = cause.getClass();
+			for (Entry<Class<? extends Throwable>, ContainerAwareBatchErrorHandler> entry : this.delegates.entrySet()) {
+				if (entry.getKey().equals(causeClass)) {
+					handled = true;
+					entry.getValue().handle(thrownException, records, consumer, container, invokeListener);
+					break;
```
artembilan, line 101: Why just `return;` from here is not enough instead of that `handled` state tracking?

spring-kafka/src/main/java/org/springframework/kafka/listener/ConditionalDelegatingErrorHandler.java
```diff
@@ -0,0 +1,97 @@
...
+		if (cause instanceof ListenerExecutionFailedException) {
+			cause = thrownException.getCause();
+		}
+		if (cause != null) {
+			Class<? extends Throwable> causeClass = cause.getClass();
+			for (Entry<Class<? extends Throwable>, ContainerAwareErrorHandler> entry : this.delegates.entrySet()) {
+				if (entry.getKey().equals(causeClass)) {
+					handled = true;
+					entry.getValue().handle(thrownException, records, consumer, container);
+					break;
```
artembilan, line 88: DITTO

artembilan (changes requested):

Issue #1858, opened 2021-07-07
Property spring.json.trusted.packages=* doesn't work when also using property spring.json.type.mapping=something:com.example.Something


idkw:
**Affects Version(s):** 2.7.3

When using the property `spring.json.trusted.packages=*` the code at this line clears the trusted packaged : 
https://github.com/spring-projects/spring-kafka/blob/3a31b3cbeab54d1620b487957dc5b3c79aaf7c76/spring-kafka/src/main/java/org/springframework/kafka/support/converter/DefaultJackson2JavaTypeMapper.java#L79-L92

An empty trusted packages list is later interpreted as a "trust all" at : 

https://github.com/spring-projects/spring-kafka/blob/3a31b3cbeab54d1620b487957dc5b3c79aaf7c76/spring-kafka/src/main/java/org/springframework/kafka/support/converter/DefaultJackson2JavaTypeMapper.java#L149-L160

However when also using the property `spring.json.type.mapping=something:com.example.Something` the code at this line is adding type mappings to the trusted packages.

https://github.com/spring-projects/spring-kafka/blob/3a31b3cbeab54d1620b487957dc5b3c79aaf7c76/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java#L390-L395

This means that the trustedPackages collection is no longer empty and the `isTrustedPackage` check will return false for any class that is not defined in the type mappings even though the property `spring.json.trusted.packages=*` is supposed to allow everything.

The bug was introduced in that commit 
https://github.com/spring-projects/spring-kafka/commit/bd49070c626ed572db1c850117bbb5e0e237eec0

The workaround I have for now is to declare type mappings for all classes even the ones I didn't need to, which is not very practical.

idkw:
Thanks @garyrussell for the fix
I have added a comment on the commit https://github.com/spring-projects/spring-kafka/commit/acc7fcfc8460668ae41921951170426d1ccf7826#commitcomment-53179546 to propose a cleaner solution to avoid future similar bug when working on this class

PR #1859, opened 2021-07-07
GH-1858: Fix Regression in JsonDeserializer

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1858

Do not add mapping class packages if trusted packages is already all (`*`).

**cherry-pick to 2.7.x**


artembilan:

Issue #1861, opened 2021-07-08
Add receive() Operation to KafkaTemplate


garyrussell:
Fetch a specific record.

saagar2000:
do we have plan to support receive for multiple topic/partition/offsets ? I  would be happy to contribute.

garyrussell:
Do you mean fetch several records in one operation?

We don't have plans to do so, but feel free to submit a PR (we can reopen this issue).

saagar2000:
yes - fetch several records across different topics - okay  I will try to work on it. Thank you

garyrussell:
https://github.com/spring-projects/spring-kafka/issues/1948

PR #1862, opened 2021-07-08
GH-1861: Add KafkaTemplate.receive()

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1861

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
@@ -16,10 +16,12 @@
 
 package org.springframework.kafka.core;
 
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java
```diff
@@ -267,6 +269,29 @@ default boolean inTransaction() {
 		throw new UnsupportedOperationException("This implementation does not support this operation");
 	}
 
+	/**
+	 * Receive a single record with the default poll timeout (5 seconds).
```
artembilan, line 273: I don't think it is legit to mention those `5 seconds` on the interface level. 
Better to say something like "with some default timeout" or leave it as is but without those 5 secs.

(Not so strong opinion: just was confused do not see those 5 seconds in this class)

artembilan (changes requested):

PR #1865, opened 2021-07-12
GH-1863 - Setup listener error handler and type converter for @RetryableTopics

### LINKED ISSUE #1863: How to configure errorHandler when using @RetryableTopic
When I use **@RetryableTopic**, the **errorHandler** configuration of **@KafkaListener** does not take effect and is replaced with **SeekToCurrentErrorHandler**, I want to print the error stack trace. How should I configure?

```
@RetryableTopic(
        attempts = "2",
        backoff = @Backoff(delay = 1000L, multiplier = 2.0),
        topicSuffixingStrategy = TopicSuffixingStrategy.SUFFIX_WITH_INDEX_VALUE,
        fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC,
        listenerContainerFactory = "retryKafkaListenerContainerFactory"
)
@KafkaListener(
        topics = TOPIC_ENTERPRISE,
        clientIdPrefix = "enterprise",
        containerFactory = "kafkaListenerContainerFactory",
        errorHandler = "customKafkaListenerErrorHandler"
)
public void enterpriseConsumer(ConsumerRecord<Long, String> record) {
    LOGGER.info("{} - topoc:{}, part: {}, offset: {}， key: {}, value: {}", Thread.currentThread().getName(), record.topic(), record.partition(), record.offset(), record.key(), record.value());
    int i = 1 / 0;
}
```

tomazfernandes:
Fixes #1863.

I remember having read somewhere that some endpoint properties should be set after registering the endpoint - this PR sets all properties before registration for the retry topic endpoints, but all tests pass nevertheless so it doesn't seem to be a problem. If there's any concerns about this please let me know, it'd be a reasonable bigger change if we need to take that into account.

I also noticed the type converter specified in the annotation wasn't being setup too, so I added it to this PR as well.

I added both features to the first integration test in RetryTopicIntegrationTests suite.

Hope it's ok, let me know if there's anything else to be done about this.

Thanks!

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -465,7 +465,7 @@ public class KafkaListenerAnnotationBeanPostProcessor<K, V>
 		}
 
 		RetryTopicConfigurer.EndpointProcessor endpointProcessor = endpointToProcess ->
-				this.doProcessKafkaListenerAnnotation(endpointToProcess, kafkaListener, bean);
+				this.processKafkaListenerAnnotationForRetryTopic(endpointToProcess, kafkaListener, bean);
 
 		String beanRef = kafkaListener.beanRef();
 		this.listenerScope.addListener(beanRef, bean);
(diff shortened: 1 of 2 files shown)


garyrussell (approved):

garyrussell:
Cherry-picked to 2.7.x

Issue #1866, opened 2021-07-13
Since version 2.7, paused consumer auto-resume without asking just after pausing.


theogoria:
**Affects Version(s):** 2.7.1

Hello,

I think I have found a bug with the pause method of  Listener Containers.
When I call `registry.getListenerContainers().forEach(MessageListenerContainer::pause);` I have something like this in the log :

```
2021-07-06T18:20:16,650- [DEBUG] [T-1] k.l.KafkaMessageListenerContainer$ListenerConsumer - [{}] - Paused consumption from: [partition-0]
2021-07-06T18:20:16,678- [DEBUG] [T-1] k.l.KafkaMessageListenerContainer$ListenerConsumer - [{}] - Received: 0 records
2021-07-06T18:20:16,678- [DEBUG] [T-1] k.l.KafkaMessageListenerContainer$ListenerConsumer - [{}] - Paused consumption from: [partition-1]
2021-07-06T18:20:16,718- [DEBUG] [T-1] k.l.KafkaMessageListenerContainer$ListenerConsumer - [{}] - Received: 0 records
2021-07-06T18:20:16,718- [DEBUG] [T-1] k.l.KafkaMessageListenerContainer$ListenerConsumer - [{}] - Paused consumption from: [partition-2]
2021-07-06T18:20:16,767- [DEBUG] [T-1] k.l.KafkaMessageListenerContainer$ListenerConsumer - [{}] - Received: 0 records
2021-07-06T18:20:16,767- [DEBUG] [T-1] k.l.KafkaMessageListenerContainer$ListenerConsumer - [{}] - Paused consumption from: [partition-3]
2021-07-06T18:20:21,651- [DEBUG] [T-1] k.l.KafkaMessageListenerContainer$ListenerConsumer - [{}] - Resumed consumption from [partition-0]
2021-07-06T18:20:21,652- [DEBUG] [T-1] k.l.KafkaMessageListenerContainer$ListenerConsumer - [{}] - Received: 0 records
2021-07-06T18:20:21,679- [DEBUG] [T-1] k.l.KafkaMessageListenerContainer$ListenerConsumer - [{}] - Resumed consumption from [partition-1]
2021-07-06T18:20:21,679- [DEBUG] [T-1] k.l.KafkaMessageListenerContainer$ListenerConsumer - [{}] - Received: 0 records
[...]

PR #1867, opened 2021-07-13
fix issue #1866 -> pause consumer.

theogoria:
fix issue #1866 : call `pausePartition` on all partitions when pausing the consumer so partitions are not restarted by `resumePartitionsIfNecessary.`

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -1524,6 +1524,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			if (this.consumerPaused && !isPaused() && !this.pausedForAsyncAcks) {
 				this.logger.debug(() -> "Resuming consumption from: " + this.consumer.paused());
 				Set<TopicPartition> paused = this.consumer.paused();
+				paused.forEach(part -> resumePartition(part));
 				this.consumer.resume(paused);
 				this.consumerPaused = false;
 				publishConsumerResumedEvent(paused);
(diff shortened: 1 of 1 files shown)


garyrussell:
@theogoria Thanks for the contribution.

However, I think this needs more sophistication - the two functions (pausing the consumer for all partitions)and pausing individual partitions for `@RetryableTopic`) need to be kept separate - a partition paused by one function should not be resumed by the other, and vice-versa.

It also needs test cases.

cc/ @tomazfernandes

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1270,7 +1270,9 @@ protected void pollAndInvoke() {
 			}
 			debugRecords(records);
 			resumeConsumerIfNeccessary();
-			resumePartitionsIfNecessary();
+			if (this.consumerPaused && !isPaused()) {
+				resumePartitionsIfNecessary();
+			}
```
garyrussell, line 1275: My comment on the issue was wrong, this should be `if (!this.consumerPaused)`.

theogoria: Yes, you're right, I didn't pay enough attention yesterday, it's fixed.

garyrussell: Thanks; I see some other issues while writing a test for this; I'll take it from here.

Thanks again for the contribution.

garyrussell:
Thanks again; replaced by https://github.com/spring-projects/spring-kafka/pull/1868

PR #1868, opened 2021-07-14
GH-1866: Fix Pause/Resume

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1866

The new retryable topic feature pauses/resumes individual partitions.

This broke normal container pause/resume by incorrectly resuming partitions
that were paused by the container pause operation.

Similarly, if individual partitions were paused and then the container was
paused and resumed, the container resumed all partitions.

Decouple the functionality to prevent this cross-talk.

Do not resume any individually paused partitions when the container is in a paused state.
Do not resume any individually paused partitions when the container is resumed.

Also

Use a `ConcurrentHashMap.newKeySet()` instead of synchronization on partition pause requests.
Use `getAssignedPartitions()` to allow the retry topic feature to work with manual assignments.

Add tests to verify no cross-talk between pausing individual partitions and the container.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -18,11 +18,11 @@ package org.springframework.kafka.listener;
 
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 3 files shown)


garyrussell:
cc/ @tomazfernandes @theogoria

tomazfernandes:
LGTM @garyrussell!

artembilan:
@garyrussell , let me know if we need to cherry-pick this to `2.7.x`

garyrussell:
Yes, please; sorry - I keep forgetting main is now 2.8.x.

garyrussell:
It might not cherry-pick cleanly; If so, I can back port.

artembilan:
Cherry-picked to `2.7.x` as https://github.com/spring-projects/spring-kafka/commit/726c48517d6fc5ccf3a6c1b7e51df65ee093c66c after including the second race condition fix.

PR #1869, opened 2021-07-14
GH-615: Error Handler Evolution Part 1

garyrussell:
Flatten `GenericErrorHandler` into a single interface.
Create an adapter for legacy error handlers.

See https://github.com/spring-projects/spring-kafka/issues/615

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
```diff
@@ -0,0 +1,143 @@
...
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+
+import org.springframework.kafka.support.TopicPartitionOffset;
+
+/**
+ * Replacement for {@link ErrorHandler} and {@link BatchErrorHandler} and their
+ * sub-interfaces.
+ *
+ * @author Gary Russell
+ * @since 2.7.4
```
artembilan, line 32: I'm confused a little bit: the issue says `3.0 Backlog`, but looks like you even would like to back-port it to `2.7.x`.
Although I may see your point: you want to introduce it now to deprecate and remove those gradually in the future versions.
What is you thought  on the matter, please?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
garyrussell: Oops - sorry; I started this work a few weeks ago before I decided to make main 2.8.x.

I also want to make a few more changes to ease migration of the old error handlers.

PR #1871, opened 2021-07-16
GH-615: Add CommonLoggingErrorHandler

garyrussell:
See https://github.com/spring-projects/spring-kafka/issues/615

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
@@ -18,6 +18,7 @@ package org.springframework.kafka.listener;
 
 import java.util.List;
 
+import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
@@ -35,14 +36,6 @@ import org.springframework.kafka.support.TopicPartitionOffset;
 public interface CommonErrorHandler extends DeliveryAttemptAware {
 
 	/**
-	 * Return true if this error handler is for a batch listener.
-	 * @return true for batch.
-	 */
-	default boolean isBatch() {
-		return false;
-	}
-
-	/**
 	 * Return false if this error handler should only receive the current failed record;
 	 * remaining records will be passed to the listener after the error handler returns.
 	 * When true (default), all remaining records including the failed record are passed
@@ -52,7 +45,7 @@ public interface CommonErrorHandler extends DeliveryAttemptAware {
 	 * @see #handleRemaining(Exception, List, Consumer, MessageListenerContainer)
 	 */
 	default boolean remainingRecords() {
-		return true;
+		return false;
 	}
 
 	/**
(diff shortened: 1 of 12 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
```diff
@@ -34,12 +34,50 @@
  */
 public interface CommonErrorHandler extends DeliveryAttemptAware {
 
+	/**
+	 * The type of error handler.
+	 */
+	enum Type {
```
artembilan, line 40: Why do we need this extra condition layer?
Why just don't make an assumption that target error handler should implement both strategies?
I said "should" not "must" because in many cases the error handler is used in the target application for record use-case or batch.
I really doubt that end-users are going to have both variants in one project. If they do, then they just need to implement everything in the contract.
Otherwise it is their choice.
The framework code would become much easier: just call respective contract from one or another branch of the logic in the container.
I mean the decision to handle records or batches or both is really made on design phase, but definitely not at runtime.

Does it make sense, Gary?

artembilan (changes requested):
I'll continue to review when you give me an answer to my concern about error handler type.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
garyrussell: When this work is complete, I want to deprecate (and then remove) the old interfaces.

See https://github.com/spring-projects/spring-kafka/issues/1872 - since I want a single container factory to support both types of listener, the standard error handlers provided by the framework would have to support both types.

It is inconvenient to have to define 2 container factories for just this one property (batch Vs. record). Furthermore, defining another factory disables auto configuration in boot, so the user has to define two container factory beans, usually with the only difference being the `batchListener` property. (This has been asked for at least once, to my recollection, maybe more times).

I am open to other suggestions to solve this problem, if you have them.

artembilan: Right. That's why I'm asking why to have this `type` choice when in most cases we need to implement both and custom one may decide to implement only one port if they don't use another in their logic.
I don't see a business purpose of this static type option if we definitely may use such an error handler only in one use-case making decision upfront on design phase...
I may miss something or don't see the whole picture, so will take a look closer locally what you do with this after your next comment . 😉

garyrussell: I would have to make the `ErrorHandlerAdapter` support both; or I could make it package private so it is only used internally (which is probably better).

I can remove the enum, but I believe I still need the `supportsBatch()`, `supportsRecord()` (both default false?) so we can do an early check for compatibility rather than waiting for the first error to occur.

WDYT?

artembilan: M-m-m. Can't it really be a part of of that `ErrorHandlerAdapter` then, which is going to be package protected and won't destruct target `CommonErrorHandler` implementations?
We may state clearly in the `CommonErrorHandler` JavaDocs what and when should be implemented.
I also believe that this `ErrorHandlerAdapter` is going away eventually, so in the end we will just call `handleRecord()` or `handleBatch()` from the listener container.
My point is since you are going to support both modes on the listener container level why don't support both in the error handler unconditionally?

garyrussell: Good point - and we should definitely make the adapter private.

Ok; convinced 😄

garyrussell: Squashed - much of the code in the first commit was removed.

PR #1874, opened 2021-07-19
Documentation: fixup example for sending a message

ttrei:


garyrussell:
This is invalid; the success method gets a `SendResult`. Why do you believe it should be changed?

https://github.com/spring-projects/spring-kafka/blob/cf89aeedf8994a5f74df800ab20c7d22e0f060f6/spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java#L353-L365

ttrei:
Because otherwise the code doesn't compile :)

Here is an example from the same test file:
https://github.com/spring-projects/spring-kafka/blob/cf89aeedf8994a5f74df800ab20c7d22e0f060f6/spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java#L387-L392

garyrussell:
Oops; yes; sorry about that; the first test doesn't use the specialized callback.

garyrussell:
Thanks.

PR #1881, opened 2021-07-22
Make `JacksonUtils` compatible with Native

artembilan:
* Extract inner classes for Jackson modules
and use them only when respective `static boolean` is true
according presence of required class in CP
* Add `optional` dependencies for Jackson modules for compilation

**Cherry-pick to `2.7.x`**


garyrussell:
...and cherry-picked.

Issue #1872, opened 2021-07-16
Support Listener Type Override on @KafkaListener


garyrussell:
Currently, if you have both record and batch listeners, you need 2 container factories; the listener type (batch or record) is defined at the factory level and not configurable on a per listener basis.

PR #1893, opened 2021-07-29
GH-1872: @KafkaListener Override Listener Type

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1872

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListener.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListener.java
@@ -283,4 +283,18 @@ public @interface KafkaListener {
 	 */
 	String contentTypeConverter() default "";
 
+	/**
+	 * Override the container factory's {@code batchListener} property. The listener
+	 * method signature should receive a {@code List<?>}; refer to the reference
+	 * documentation. This allows a single container factory to be used for both record
+	 * and batch listeners; previously separate container factories were required.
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListener.java
```diff
@@ -283,4 +283,18 @@
...
+	 * documentation. This allows a single container factory to be used for both record
+	 * and batch listeners; previously separate container factories were required.
+	 * @return "true" for the annotated method to be a batch listener or "false" for a
+	 * record listener. If not set, the container factory setting is used. SpEL and
+	 * property placeholders are not supported because the listener type cannot be
+	 * variable.
+	 * @since 2.8
+	 * @see Boolean#parseBoolean(String)
+	 */
+	String batch() default "";
```
artembilan, line 298: Why do we need to support 3 states?
Why can't `record` be a default? 
Maybe with this change container factory should not expose such an option any more?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListener.java
garyrussell: Because I can't tell if the user explicitly set it to `false` (and I need to know so I don't apply the container factory value).

I think it should remain on the factory - a user might want to create batch listeners by default and override for a particular listener only to record.

Issue #1894, opened 2021-07-29
DLPR addExceptionInfoHeaders Incorrectly Called Twice


garyrussell:
With a key `DeserializationException`, the exception information is incorrectly added to both the key and value exception headers.

PR #1895, opened 2021-07-30
GH-1894: DLPR: Fix Headers with Only Key Exception

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1894

When there was only a key deserialization exception, the exception info was added
twice, under the key keys and value keys.

**cherry-pick to 2.7.x**


artembilan:

Issue #1882, opened 2021-07-23
KafkaHeaders.DLT_EXCEPTION_FQCN is always a KafkaListenerExecutionException


jbotuck:
When using the dead letter recoverer, KafkaHeaders.DLT_EXCEPTION_FQCN is almost always a KafkaListenerExecutionException. This makes the header less helpful than it could be. Is there any way we can add a header for the cause of the exception?

garyrussell:
This is a reasonable request; you can, however, see the cause in the stack trace header.

PR #1896, opened 2021-07-30
GH-1882 DLPR: Add Exception Cause Header

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1882


artembilan (changes requested):
LGTM, but Checkstyle violations.

Issue #1900, opened 2021-08-05
Cannot use Kotlin property syntax on ContainerProperties.idleEventInterval since release 2.7.3


psychobaatezu:
Since release 2.7.3, it's not possible anymore to use the Kotlin property syntax on ContainerProperties.idleEventInterval field.

![image](https://user-images.githubusercontent.com/15527217/128357297-649bb74a-2ab1-4d50-baf6-3f9ffca92399.png)

This is caused by this commit : 0ee2d49ea525de038d315bc243b62f1004a26dc6

![image](https://user-images.githubusercontent.com/15527217/128357511-8b2a3581-0f08-4f62-ba77-3d2997e07c7f.png)

Granted, the field is actually nullable. But by adding the annotation on the getter instead of the actual field, it tells the Kotlin interpreter that the getter will return a nullable Long type (Long?) and the setter will apply the given value to a non-nullable Long Type (Long). This discrepancy blocks Kotlin property syntax because the underneath getter/setter are not for the same type.

So to use 2.7.3, existing code need to change to this to continue to work :

![image](https://user-images.githubusercontent.com/15527217/128358237-7530e1d7-4eb1-4f82-9100-028b9d6138e8.png)

The solution to solve that is to put `@Nullable` annotation on the field instead, which by the way is more accurate.

psychobaatezu:
Possible fix : https://github.com/spring-projects/spring-kafka/pull/1901

garyrussell:
Thanks; merged and cherry-picked to 2.7.x.

PR #1901, opened 2021-08-05
GH-1900: Add @nullable to setters for the corresponding nullable getters.

psychobaatezu:


garyrussell:
@psychobaatezu This does not pass muster with Sonar's null checking rules, which I just ran against this branch.

https://sonar.spring.io/project/issues?id=org.springframework.kafka%3Aspring-kafka-dist%3Asonar&resolved=false&sinceLeakPeriod=true&types=BUG

Can we not simply add `@Nullable` to the setter parameters to make the getter/setter back in synch?

psychobaatezu:
@garyrussell That's another possibility. I've amended my commit with that approach instead.

PR #1905, opened 2021-08-09
GH-1827: ContainerGroupSequencerTests Diagnostics

garyrussell:
Set `autoStartup` to `false` so we see the `Starting first group` message.

**cherry-pick to 2.7.x**


artembilan:

Issue #1879, opened 2021-07-21
Call `(De)Serializer.configure()` in the respective setters of `DefaultKafkaProducerFactory` and `DefaultKafkaConsumerFactory`


artembilan:
**Expected Behavior**

End-user should be free from finding the way to propagate configuration properties to their custom `(De)Serializier` when they configure it programmatically.
The general idea for the target solution is to call  `(De)Serializer.configure()` directly in the DefaultKafkaProducerFactory` or `DefaultKafkaConsumerFactory` when respective setter is used for external  `(De)Serializier` .
Such a behavior of those setters must be mentioned in their JavaDocs.
Out-of-the-box `(De)Serializier` (e.g. `JsonSerializer`) must be guarded with some state like `initialized` to not call a `configure()`  if that has been called externally, e.g. in projects which don't have this fix yet.
Another consistency matter is about to not call `configure()` if some setters of that `(De)Serializer` has been called: or setters configuration, or via properties, but not both.

**Current Behavior**

We have to call `configure()` explicitly in the factory customizer before setting (de)serializer into that factory back:
```
    public void customize(DefaultKafkaProducerFactory<?, ?> producerFactory) {
        JsonSerializer<Object> jsonSerializer = new JsonSerializer<>(objectMapper());
        jsonSerializer.configure(producerFactory.getConfigurationProperties(), false);
        producerFactory.setValueSerializer((JsonSerializer) jsonSerializer);
    }
```

**Context**

More info in this SO thread: https://stackoverflow.com/questions/68472258/spring-json-type-mapping-property-ignored-when-specifying-a-custom-objectmapper

PR #1907, opened 2021-08-11
GH-1879: Factories Now Configure (De)Serializers

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1879

It was not intuitive that configuration properties were not applied when
construcing (De)Serializers programmatically.

The producer and consumer factories now call the `configure()` method.
However, the JSON implementations cannot be configured with a mixture
of setters and configuration properties.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
@@ -49,6 +49,10 @@ import com.fasterxml.jackson.databind.type.TypeFactory;
 /**
  * Generic {@link org.apache.kafka.common.serialization.Deserializer Deserializer} for
  * receiving JSON from Kafka and return Java objects.
+ * <p>
+ * IMPORTANT: Configuration must be done completely with property setters or via
+ * {@link #configure(Map, boolean)}, not a mixture. If any setters have been called,
+ * {@link #configure(Map, boolean)} will be a no-op.
  *
  * @param <T> class of the entity, representing messages
  *
@@ -123,6 +127,8 @@ public class JsonDeserializer<T> implements Deserializer<T> {
 	private boolean useTypeHeaders = true;
 
 	private JsonTypeResolver typeResolver;
+
+	private boolean configured;
 
 	/**
 	 * Construct an instance with a default {@link ObjectMapper}.
@@ -308,6 +314,7 @@ public class JsonDeserializer<T> implements Deserializer<T> {
 		if (typeMapper instanceof AbstractJavaTypeMapper) {
 			addMappingsToTrusted(((AbstractJavaTypeMapper) typeMapper).getIdClassMapping());
 		}
+		this.configured = true;
 	}
 
 	/**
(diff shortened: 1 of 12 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
```diff
@@ -365,10 +376,14 @@ public void setTypeFunction(BiFunction<byte[], Headers, JavaType> typeFunction)
 	 */
 	public void setTypeResolver(JsonTypeResolver typeResolver) {
 		this.typeResolver = typeResolver;
+		this.configured = true;
 	}
 
 	@Override
 	public void configure(Map<String, ?> configs, boolean isKey) {
+		if (this.configured) {
+			return;
```
artembilan, line 385: May be better to throw an exception (or at least WARN) to let end-users to know that they are wrong?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerializer.java
```diff
@@ -132,10 +140,14 @@ public void setUseTypeMapperForKey(boolean isKey) {
 			((AbstractJavaTypeMapper) getTypeMapper())
 					.setUseForKey(isKey);
 		}
+		this.configured = true;
 	}
 
 	@Override
 	public void configure(Map<String, ?> configs, boolean isKey) {
+		if (this.configured) {
+			return;
```
artembilan, line 149: DITTO

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonDeserializer.java
garyrussell: We can't do that; because the supplier is called each time a new consumer is created. The default suppliers return the same instance each time.

I could add a DEBUG log; WDYT?

artembilan: Got it. Then WARN anyway: people have to know that they are doing something wrong. 
With debug it may slip out and then we will get some issue that they config doesn't work this or that way.

garyrussell: Hmmm - warn would be too much noise - maybe I should use 2 flags, one for setters (and throw exception) and one for `configAlreadyDone`?

garyrussell: Or, maybe just close this as won't fix?

artembilan: It will be, but it will say you that there is something wrong with the app. Then you go to docs and JavaDocs and you see that you can't combine setters configuration with properties one.
Isn't the the reason why we use that `Assert` all around?

garyrussell: My concern is a single shared version is used and if we call `configure()` multiple times, even though the complete operation is idempotent, the configuration will be in an indeterminate state while `configure()` is running, and another thread might be using it to deserialize.
I therefore only want to run `configure()` once.

Hence my suggestion to use 2 flags instead of 1.

garyrussell: Working on polishing it...

PR #1908, opened 2021-08-12
GH-1827: Fix Race

garyrussell:
- increase idle event interval to prevent sequencer stopping the container too soon.
- tighten up synchronization in the sequencer

**cherry-pick to 2.7.x**


artembilan:

Issue #1704, opened 2021-02-09
Trace information is missing when Exception is thrown from KafkaListener methods


jonatan-ivanov:
**Affects Version(s):** All currently supported versions (including `2.6.5`)

🐞 Bug report

The issue was originally reported for Spring Cloud Sleuth but it seems it is not a Sleuth issue (and I can't transfer issues across orgs) so I'm opening this one to track it at the right place.

The original issue is this: https://github.com/spring-cloud/spring-cloud-sleuth/issues/1659 opened by @m-grzesiak
It contains the details to understand what is going on, a sample project that reproduces the issue and some investigation details. The issue is very similar to https://github.com/spring-cloud/spring-cloud-sleuth/issues/1660

**Description**
When an exception is thrown from a method annotated with `@KafkaListener` all tracing information is lost when it reaches the error handler so that error logs do not contain this information.

**Sample**: https://github.com/jonatan-ivanov/sleuth-gh-1659
The sample project uses `spring-kafka` `2.6.4` but the issue must be present in the latest `spring-kafka` too (`2.6.5`).

Investigation details: https://github.com/spring-cloud/spring-cloud-sleuth/issues/1659
Possible fix (breaking change): should be similar to https://github.com/spring-projects/spring-amqp/pull/1287

MaksimMyshkin:
Also waiting for fix. It was unexpected that ErrorHandler is not getting original traceId context from Sleuth.

kdowbecki:
For `@KafkaListener` in batch mode Sleuth by default doesn't take the trace context from `b3` header of individual records and instead generates a single trace that represents an entire batch. I like this. However I'd like to start the trace earlier, just before the Kafka Consumer `poll()` is called and end the trace later, after `BatchErrorHandler` is invoked. That way I can get a single trace for all operations related to a batch of records.

However there is no mechanism to intercept `KafkaMessageListenerContainer.ListenerConsumer.run()` method. If there was an interceptor that allows to do something around `pollAndInvoke()` method I could add my own trace because everything happens in a single thread. 

Currently there are few scenarios that loose the trace context:
1. Exceptions throw by `@KafkaListener` which is annotated with `@NewSpan`. 
2. Exceptions throw by Key/Value Deserializer class if we specify `spring.deserializer.value.function` that re-throws the exception.
3. Logs directly from by `KafkaMessageListenerContainer.ListenerConsumer` e.g. "Error handler threw an exception".

[...]

PR #1912, opened 2021-08-18
GH-1704 Broader Batch/RecordInterceptor

kdowbecki:


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1233,6 +1233,9 @@ public void run() {
 				catch (Exception e) {
 					handleConsumerException(e);
 				}
+				finally {
+					finishInvoke();
```
kdowbecki, line 1237: I wanted it to run after `handleConsumerException()` so the trace context will be available for an exception outside of the listener processing e.g. when offsets fail to commit.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1441,6 +1476,37 @@ private void fixTxOffsetsIfNeeded() {
...
+			}
+		}
+
+		private void batchInterceptBeforePoll() {
+			if (this.commonBatchInterceptor != null) {
+				try {
+					this.commonBatchInterceptor.beforePoll(this.consumer);
+				}
+				catch (Exception e) {
+					this.logger.error(e, "BatchInterceptor threw an exception");
```
kdowbecki, line 1494: Do we need a more detailed log message to tell us that it happened for `beforePoll()`. Right now the same message is duplicated in few places for `beforePoll()`, `intercept()` and `finishInvoke()`. It feels like exceptions when invoking interceptor are a corner case, not sure if anyone really pays attention to this log.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Right, but as I said in my comment on the issue; it needs to be called conditionally in the finally block - only if an exception was thrown; otherwise it will be called twice.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: The stack trace is enough to tell us where it happened, but it wouldn't hurt to mention the method name; interceptors should not throw exceptions, but we need to protect against it.

garyrussell:
I am not working today - I will try to do a full review tomorrow.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
kdowbecki: You said:

> We need to handle the `poll()` throwing an exception, with a conditional call to `finishInvoke()` in a `finally` block in the `run()` method. e.g. a boolean `success` set to false just after the `while()` and true after the `pollAndInvoke()` and use it in the `finally` block to call the clean up if an exception occurred.

My understanding of the tracing use-case that started this PR is that `finishInvoke()` must be run if `beforePoll()` was run. Whether or not records were returned from `poll()`, or processed successfully by listener, or exception was thrown somewhere, is not important. The only goal is to run `finishInvoke()` after `beforePoll()`.

Can you please explain what do you mean by:

>  clean up if an exception occurred.

because it feels like the `finishInvoke()` cleanup is not dependent on an exception being thrown during processing. Unless you are suggesting that `finishInvoke()` should only run if `beforePoll()` run without throwing an exception? My apologies, I'm clearly missing something.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1270,6 +1273,7 @@ protected void pollAndInvoke() {
 				return;
 			}
 			this.polling.set(true);
+			beforePoll();
```
kdowbecki, line 1276: `pollAndInvoke()` starts with offset related operation e.g. it can commit offsets. Are these offsets of the previous record/batch processing? Perhaps it would be better to call `finishInvoke()` just before `idleBetweenPollIfNecessary()` few lines above if single batch of records is fully processed in two `pollAndInvoke()` invocations?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Well, I was thinking that we should call `finishInvoke()` after each record in a record listener but I guess you are right; these methods should be just scoped around the `poll()` and completion of processing all records from the poll.

In that case, we should rename `finishInvoke` to something like `recordsProcessed()` to decouple it from invoking the listener.

Most likely record interceptors won't use them, but we should provide them for consistency.

kdowbecki: I renamed `finishInvoke()` to `afterRecordsProcessed()` and improved the Javadoc. 

I still have to figure out how to test it because `beforePoll()` and `afterRecordsProcessed()` is called in a loop outside of the test. I plan to look into it over the weekend.

spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
```diff
@@ -178,9 +180,10 @@
 		"annotated18", "annotated19", "annotated20", "annotated21", "annotated21reply", "annotated22",
 		"annotated22reply", "annotated23", "annotated23reply", "annotated24", "annotated24reply",
 		"annotated25", "annotated25reply1", "annotated25reply2", "annotated26", "annotated27", "annotated28",
-		"annotated29", "annotated30", "annotated30reply", "annotated31", "annotated32", "annotated33",
```
kdowbecki, line 181: `annotated29` was not used as far as I can tell.

spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
```diff
@@ -600,7 +603,16 @@ public void testBatchRecords() throws Exception {
 		assertThat(this.config.listen12Message.getPayload()).isInstanceOf(List.class);
 		List<?> errorPayload = (List<?>) this.config.listen12Message.getPayload();
 		assertThat(errorPayload.size()).isGreaterThanOrEqualTo(1);
+	}
+
+	@Test
+	public void testBatchInterceptor() throws Exception {
+		template.send("annotated43", null, "foo");
+		assertThat(this.config.batchAfterRecordsProcessedLatch.await(30, TimeUnit.SECONDS)).isTrue();
```
kdowbecki, line 611: Because I can't wait on a latch inside `@KafkaListener`, it's too early and the test might run next assertion before the interceptor methods are called. Current approach doesn't work because `batchAfterRecordsProcessedLatch` is called too early, in the previous `pollAndInvoke()` iteration before the actual test sends the message. I need to improve it.

kdowbecki:
@garyrussell Apologies for delay, I fixed the tests. I haven't found a way to test it perfectly because `poll()` is done in a loop, I can't clearly distinguish one `poll()` invocation from another without writing a lot of test code. I didn't went this route and written a simple test. Can you please review?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: I concur that `beforePoll()` should be imediately before the poll. It's not clear why you have 2 calls to `beforePoll()` (one here and on in each branch of `doPoll()`.

However, we now have a new method `pollConsumer()` (common to both batch and record listeners - the `beforePoll()` should be done there.

I don't think that calling `afterRecordsProcessed()` before `processCommits()` will cause any problems; also, moving it from the `finally` block might cause us to miss the last call (when `isRunning()` goes false, when the container is stopped.

garyrussell (changes requested):
For the test scenario, we would normally use mocks an an `InOrder` to ensure the consumer.poll and interceptor methods are called in the right sequence.

There are several examples in other test cases; if you are not comfortable with doing that, I can add a test before the final merge.

Thanks again for the contribution!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
kdowbecki: Two calls to `beforePoll()` method was bug. I removed redundant call when writing `InOrder` tests.

garyrussell: Thanks - I will take another look tomorrow; thanks again.

kdowbecki:
I added new `InOrder` interceptor tests in `KafkaMessageListenerContainerTests` class. I wonder do we still need the integration tests that I added before in this PR in `EnableKafkaIntegrationTests` class. I'm leaning towards "no" since these old tests verify the same thing, I think I should revert `EnableKafkaIntegrationTests`. What's your opinion @garyrussell ?

garyrussell:
@kdowbecki Nice work on the mock tests; thanks; yes, I agree, we can revert the integration tests.

spring-kafka/src/main/java/org/springframework/kafka/listener/BatchInterceptor.java
```diff
@@ -61,4 +61,29 @@ default void success(ConsumerRecords<K, V> records, Consumer<K, V> consumer) {
...
+	 * Called after records were processed by listener and error handler.
+	 * <p>
+	 * It can be used to clear thread-bound resources which were set up in {@link #beforePoll(Consumer)}.
+	 * This is the last method called by the {@link MessageListenerContainer} before the next record
+	 * processing cycle starts.
+	 * </p>
+	 * @param consumer the consumer.
+	 * @since 2.8
+	 */
+	default void afterRecordsProcessed(Consumer<K, V> consumer) {
```
artembilan, line 86: Don't we need to have a `ConsumerRecords<K, V> records` context over here as well?
You maybe would like to know for what batch is this stateless interceptor.
See an `ExecutorChannelInterceptor` contract in Spring Messaging for example.

spring-kafka/src/main/java/org/springframework/kafka/listener/RecordInterceptor.java
```diff
@@ -81,4 +81,29 @@ default void success(ConsumerRecord<K, V> record, Consumer<K, V> consumer) {
...
+	 * Called after records were processed by listener and error handler.
+	 * <p>
+	 * It can be used to clear thread-bound resources which were set up in {@link #beforePoll(Consumer)}.
+	 * This is the last method called by the {@link MessageListenerContainer} before the next record
+	 * processing cycle starts.
+	 * </p>
+	 * @param consumer the consumer.
+	 * @since 2.8
+	 */
+	default void afterRecordsProcessed(Consumer<K, V> consumer) {
```
artembilan, line 106: This one is for a single record interceptor, so the contract must be called `afterRecordProcessed` and have a `ConsumerRecord<K, V> record` as a first argument.
I might miss something and probably this per-record interceptor is not designed for an "after processed" hook...
Otherwise it sounds awkward to implement a `RecordInterceptor` and figure out somehow a contract for the whole batch as it sounds right now.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1455,6 +1490,37 @@ private void fixTxOffsetsIfNeeded() {
 			}
 		}
 
+		private void beforePoll() {
+			if (this.isBatchListener) {
+				batchInterceptBeforePoll();
+			}
+			else {
+				recordInterceptBeforePoll();
+			}
```
artembilan, line 1499: Sounds more like there has to be a common Before-After contract, but not the same copy in record and batch interceptors.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1236,6 +1236,9 @@ public void run() {
 				catch (Exception e) {
 					handleConsumerException(e);
 				}
+				finally {
+					afterRecordsProcessed();
```
artembilan, line 1240: I see now why no records in the "after" contract. 
Probably that `pollAndInvoke()` may return that `ConsumerRecords`?
Yes, it might be a `null` since we may fail doing all of that hard logic, but still might be better to propagate such a context down to this "after" contract invocation.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -3497,6 +3497,280 @@ void stopImmediately() throws InterruptedException {
...
+		ConsumerFactory<Integer, String> cf = mock(ConsumerFactory.class);
+		Consumer<Integer, String> consumer = mock(Consumer.class);
+		given(cf.createConsumer(eq("grp"), eq("clientId"), isNull(), any())).willReturn(consumer);
+		ConsumerRecord<Integer, String> firstRecord = new ConsumerRecord<>("foo", 0, 0L, 1, "foo");
+		ConsumerRecord<Integer, String> secondRecord = new ConsumerRecord<>("foo", 0, 1L, 1, "bar");
+		Map<TopicPartition, List<ConsumerRecord<Integer, String>>> records = new HashMap<>();
+		records.put(new TopicPartition("foo", 0), List.of(firstRecord, secondRecord));
+		ConsumerRecords<Integer, String> consumerRecords = new ConsumerRecords<>(records);
+		given(consumer.poll(any(Duration.class))).willAnswer(i -> {
+			Thread.sleep(50);
```
artembilan, line 3511: Do we really need this sleep?
Not clear what further logic expects some delays with records delivery...

artembilan (changes requested):
I might be late for the party and may miss something you have discussed before, but I thought it won't hurt too much if I raise my concerns even if they won't be accepted in the end.

Thank you!

spring-kafka/src/main/java/org/springframework/kafka/listener/BatchInterceptor.java
garyrussell: That's a good suggestion.

spring-kafka/src/main/java/org/springframework/kafka/listener/RecordInterceptor.java
garyrussell: No; this would be redundant - we already call success or failure for each record.

This hook is to clean up state when all records have been processed - the `beforePoll()` is intended so that thread-bound resources can be registered and made available in components called by the `KafkaConsumer` within the `poll()` (e.g. deserializers, partition assignors, etc, etc).

This hook is to unbind those resources.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
garyrussell: Yes; otherwise the consumer thread will call it in a tight loop.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: I guess the same comment about redundancy applies here too - the batch interceptor gets the `ConsumerRecord` ini success or failure and the record interceptor gets each record individually.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
artembilan: OK. won't another `willAnswer(i -> null)` after this one help?
According Mockito docs only the last stub is used for the rest of call.
We probably need to have a idle on the container in addition...
But yeah, we can live it as is: just wanted to share with you a logic possible via `.willAnswer(...).willAnswer(...)` chain.

spring-kafka/src/main/java/org/springframework/kafka/listener/RecordInterceptor.java
artembilan: Right; so should not be there then some other contract not the one tied to the interceptors?
It feel right now for me that this is not an interception  logic: we just don't intercept anything before and after.
Feels more like a `BeforeAfterPollProcessor`.
No? What was a premise to put such a logic into the interceptor contract?

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
garyrussell: No; returning null will not prevent the tight loop; we could add a latch so the second poll blocks until the test completes but I don't feel it's that important - there are many tests with this logic; we can put in on the slate for improvement later.

spring-kafka/src/main/java/org/springframework/kafka/listener/RecordInterceptor.java
garyrussell: It was [my suggestion on the issue](https://github.com/spring-projects/spring-kafka/issues/1704#issuecomment-896997190).

However, I see that we could.should extract a super-interface with these two methods and have the interceptors inherit from it.

The challenge is naming it 😄 

Suggestions?

garyrussell: It was [my suggestion on the issue](https://github.com/spring-projects/spring-kafka/issues/1704#issuecomment-896997190).

However, I see that we could/should extract a super-interface with these two methods and have the interceptors inherit from it.

The challenge is naming it 😄 

`BeforeAfterPollProcessor` sounds ok to me.

artembilan: That will work for me, too.
So, the logic in the Listener Container would be much simpler.

kdowbecki: Works for me, I'll add `BeforeAfterPollProcessor`

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
kdowbecki: Perhaps the issue is naming of `afterRecordsProcessed()` method. Perhaps we should remove `records` from it and call it `afterPollProcessed()` or similar?

artembilan: If we go with a new `BeforeAfterPollProcessor` contract, then just `afterPoll()` would be enough. 
Something in parallel with that `beforePoll()`.

garyrussell: 👍 

I am planning on building the next milestone (2.8.0-M3) later today; if would be great if we can get this into that ready for the next Boot milestone next week.

It's not really due until Monday, but I will be on PTO next week.

garyrussell: I don't think `afterPoll()` sounds right; it implies it's done right after. `afterPollProcessed()` sounds good.

garyrussell: or `afterPollResultsProcessed()`.

artembilan: OK. Then I'm good with that `afterPollProcessed()`.

kdowbecki: `AfterRollbackProcessor` and few other classes have `clearThreadState()` method already. Maybe instead of `afterPoll()` it should be `clearThreadState()` hear as well?

garyrussell: I could live with `clearThreadState()`.

I am not sure `ConsumerPollProcessor` conveys the right meaning.

(I agree that naming is hard 😄 ).

Still thinking - let's stick with `BeforeAfterPollProcessor.clearThreadState()` for now.

It's only a milestone, so we are at liberty to change it if we can come up with something better.

...time passes...

How about something more general...

```java
public ThreadScopeProcessor {

    void setupThreadState();

   void clearThreadState();

}
```

??

kdowbecki: Unfortunately it seems that we won't be able to do `AfterRollbackProcessor extends ThreadScopeProcessor` because `AfterRollbackProcessor` lacks the setup side. I like the generic `ThreadScopeProcessor` as an idea but it would have to be applied in more place to be valid. If it's just this one place I like other names that mention "poll" better.

I finished refactor the code and used `BeforeAfterPollProcessor`. It's getting late here in London so I will leave the renaming of the classes to you. I'm happy with either option.

garyrussell: I think we could make the `setUpThreadState` a default method; the DARP and stateful error handlers set up thread state as a side-effect of their normal processing.

garyrussell: Thanks for the work - I can take it from here.

kdowbecki: Few minor fixed to compiler warnings, but now I'm done. Have fun, and thanks for helping me with the PR.

garyrussell:
Thanks again for this; I will submit a new PR with the polishing we discussed.

Issue #1916, opened 2021-08-24
RetryTopicConfigurationBuilder#includeTopic does not work with SPEL topics


celcius112:
When manually defining a bean `RetryTopicConfiguration` with an `includeTopic`, such as:
```
  @KafkaListener(topics = "${kafka.topic}", groupId = "id")
  public void listen(String message) {

  }

  @Bean
  public RetryTopicConfiguration retryTopicConfiguration(KafkaTemplate<String, String> template) {
    return RetryTopicConfigurationBuilder.newInstance()
        .includeTopic("topic")
        .create(template);
  }
```
where `kafka.topic` is a SPEL expression that resolves to `topic`, reply and dead letter topics are not automatically created as they are not matched in `AllowDenyCollectionManager`.

Diving a bit into the implementation, `AllowDenyCollectionManager#areAllowed` will try to match the `includeTopic` with the non-resolved SPEL expression `${kafka.topic}`.

Here's a sample application. By launching the tests or the application without `.includeTopic("topic")` we see in the logs that the retry and dead letter topics are created, but adding `.includeTopic("topic")` the logs are not shown.
[demo 2.zip](https://github.com/spring-projects/spring-kafka/files/7038340/demo.2.zip)

garyrussell:
`${...}` is not SpEL; it is a property placeholder; SpEL expressions use `#{...}`.

(Just for clarification - the issue is still valid).

PR #1920, opened 2021-08-24
GH-1916: Fix RetryableTopic for SpEL and Prop PH

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1916

Previously, the `topics` property was used during reply topic configuration
before it had been resolved for SpEL expressions and Property Placeholders.

**cherry-pick to 2.7.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -438,7 +438,11 @@ public class KafkaListenerAnnotationBeanPostProcessor<K, V>
 		for (KafkaListener classLevelListener : classLevelListeners) {
 			MultiMethodKafkaListenerEndpoint<K, V> endpoint =
 					new MultiMethodKafkaListenerEndpoint<>(checkedMethods, defaultMethod, bean);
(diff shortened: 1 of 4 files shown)


garyrussell:
cc/ @tomazfernandes

artembilan:
... and back-ported to `2.7.x` after fixing small conflict in the `KafkaListenerAnnotationBeanPostProcessor` and squashing commits.

hariraogotit:
@garyrussell - includeTopics turns off the retry and falls back to the default. Here is my config.  When I comment out includeTopics, it works okay i.e non-blocking retry logic kicks in. Kindly let me know what I missed here?

 ```
@Bean
  public RetryTopicConfiguration myRetryTopic(KafkaTemplate<String, byte[]> kafkaTemplate) {
    return RetryTopicConfigurationBuilder
            .newInstance()
            .includeTopics(List.of("request-topic"))
            .exponentialBackoff(3000, 2, 100000)
            .maxAttempts(5)
            .retryTopicSuffix("-asl-nm.retry")
            .dltSuffix("-asl-nm.dlt")
            .create(kafkaTemplate);
  }
```

garyrussell:
I am not sure why you commented here; this problem has no relation to the subject of this pull request.

`includeTopics` is used to restrict retryable topics to those in the provided list; when omitted, all topics become retryable. If that is not working as expected; start a discussion (see the tab above) and include much more information, including the listener itself.

Issue #1929, opened 2021-08-31
BatchInterceptor Result Ignored Unless the Listener Consumes `ConsumerRecords` Instead of `List<ConsumerRecord>`.


marosivanco:
**Affects Version(s):** 2.7.6

I am trying to filter consumer records **before** the deserialization.

### Use Cases
Here are some scenarios in which the deserialization skipping makes perfect sense:

1. Consume only records not older than few seconds. The record creation time is in the header.
An example is a notification system which sends push notifications to the consumer devices. Such notifications are often relevant only at the moment of an event. The records corresponding to the old events need to be skipped.
2. Process only records which hold entity of certain type. The entity type is in the header.
An example is a sale system which maintains list of sale items. The items are updated based on the records from the topic. Each item type is represented by its entity (e.g. AVRO record). The books sale system consumes only Book entities, the other entities (Electronics, Cloths,...) can be skipped safely.

### Related existing APIs
There are several APIs which are kind of similar to what I need - Record Filter Strategy, Batch/Record Interceptor. Unfortunately, the Record Filter Strategy is invoked after deserialization.

The Batch Interceptor is also invoked after deserialization. However, when the factory is configured with `ByteArrayDeserializer`, the interceptor could be used to transform `ConsumerRecord` to a kind of `LazyConsumerRecord`, which could perform the deserialization (String/Json/Avro) on-demand. However, the interceptor results are ignored, see: [org.springframework.kafka.listener.KafkaMessageListenerContainer.ListenerConsumer#doInvokeBatchOnMessage](https://github.com/spring-projects/spring-kafka/blob/d8cfee6c36e866a73317db83c667d55bb8a53920/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2104), the original `recordList` is used. Instead, the new list constructed from the modified `recordsArg` should be used.

### Workaround
Use `ByteArrayDeserializer` and defer deserialization to the message listener. This has obvious disadvantage of cluttering message listener with the deserialization and the deserialization error handling.

### Possible Solution
[...]

PR #1930, opened 2021-08-31
GH-1929: Fix BatchInterceptor

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1929

Need to rebuild the record list after interception (when `interceptBeforeTx` is false).

Otherwise changes made to the records in the `ConsumerRecords` is lost.

Always intercept early when transactions are not used; rebuild list if transactions in
use, `interceptBeforeTx` is false, and a batch interceptor is present.

**cherry-pick to 2.7.x**


artembilan:

Issue #1936, opened 2021-09-08
Improve Fatal Conversion Exceptions in the Batch Listener Adapter


garyrussell:
See https://stackoverflow.com/questions/69055510/can-i-use-a-spring-kafka-recoveringbatcherrorhandler-to-handle-deserialization-e/69106462#69106462

PR #1937, opened 2021-09-08
GH-1936: Batch Listener Handle ConversionException

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1936

Make it easier to detect and recover from conversion exceptions in batch
listeners.

Also add `DelegatingByTypeSerializer`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.support.serializer;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.kafka.common.errors.SerializationException;
+import org.apache.kafka.common.header.Headers;
+import org.apache.kafka.common.serialization.Serializer;
+
+/**
+ * Delegates to a serializer based on type.
+ *
+ * @author Gary Russell
+ * @since 2.8
+ *
+ */
+public class DelegatingByTypeSerializer implements Serializer<Object> {
+
+	@SuppressWarnings("rawtypes")
+	private final Map<Class<?>, Serializer> delegates = new HashMap<>();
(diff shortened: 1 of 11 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
```diff
@@ -0,0 +1,77 @@
...
+	public byte[] serialize(String topic, Headers headers, Object data) {
+		Serializer delegate = findDelegate(data);
+		return delegate.serialize(topic, headers, data);
+	}
+
+	@SuppressWarnings("rawtypes")
+	private Serializer findDelegate(Object data) {
+		Serializer delegate = this.delegates.get(data.getClass());
+		if (delegate == null) {
+			throw new SerializationException("No matching delegate for type");
```
artembilan, line 71: Can we extend this message to say what type is wrong and what types are supported to let end-user to know where he/she is mistaken?

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/BatchAdapterConversionErrorsTests.java
```diff
@@ -0,0 +1,174 @@
+/*
+ * Copyright 2019 the original author or authors.
```
artembilan, line 2: `2021`

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/BatchAdapterConversionErrorsTests.java
```diff
@@ -0,0 +1,174 @@
...
+import org.springframework.kafka.support.converter.JsonMessageConverter;
+import org.springframework.kafka.support.serializer.DeserializationException;
+import org.springframework.kafka.support.serializer.ErrorHandlingDeserializer;
+import org.springframework.messaging.handler.annotation.Header;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+/**
+ * @author Gary Russell
+ * @since 2.2.5
```
artembilan, line 53: `2.8`

artembilan (changes requested):
Just a couple nit-picks.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	@SuppressWarnings("rawtypes")
+	private Serializer findDelegate(Object data) {
+		Serializer delegate = this.delegates.get(data.getClass());
+		if (delegate == null) {
+			throw new SerializationException("No matching delegate for type: " + data.getClass().getName()
+					+ "; supported types: " + this.delegates.keySet().stream()
+							.map(clazz -> clazz.getName())
+							.collect(Collectors.toList()));
```
artembilan, line 79: > supported types order is random.

Yes, I think the `Stream` API you use here doesn't preserve an order of the original collection: https://stackoverflow.com/questions/29216588/how-to-ensure-order-of-processing-in-java8-streams.
It doesn't matter for end-user consideration, so I think your matching solution in the unit-test is OK.

Merging...

Issue #1940, opened 2021-09-13
Add ReactiveListenerAdapter


garyrussell:
The `asyncAcks` feature now allows a reactive work load to run on the listener container thread.

Add an adapter to avoid users having to implement boilerplate code.

sobychacko:
Closing this issue since there has been no recent activity. If this type of adapter is still needed and there are sufficient use cases, we can reopen and try to prioritize it, but we will close it for the time being. Thanks!

PR #1941, opened 2021-09-13
GH-1940: Add ReactiveListenerAdapter

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1940

### DIFF
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -1279,6 +1279,213 @@ The consumer will be paused (no new records delivered) until all the offsets for
 
 IMPORTANT: While this feature allows applications to process records asynchronously, it should be understood that it increases the possibility of duplicate deliveries after a failure.
 
+[[reactive-adapter]]
+===== Reactive Listener Adapter
+
+Starting with version 2.8, the `ReactiveListenerAdapter` is now provided.
+While it is generally recommended to not perform asynchronous operations on the listner container thread, it is sometimes desirable - for example when sending data to a reactive sink.
+The <<ooo-commits,new feature for committing offsets out of order>> has provided the capability for adapting a listener to a `Flux<Mono<ConsumerRecord<?, ?>>>`.
+The adapter provides a hook for configuring each `Mono` (including error handling).
+
+IMPORTANT: Each mono will acknowledge the delivery when terminated (successfully or not), therefore the `Mono` pipeline must handle all errors.
+
+NOTE: True reactor backpressure is not supported, but backpressure is implicitly applied until all records from a poll have been acknowledged.
+
+Here is an example where the records from the poll are processed in parallel.
+
+====
+[source, java]
+----
+ConcurrentMessageListenerContainer<String, String> container =
+        containerFactory.createContainer("someTopic");
+...
+ReactiveListenerAdapter<String, String> adapter = new ReactiveListenerAdapter<>() {
+
+    @Override
+    public Mono<ConsumerRecord<String, String>> assemble(Mono<ConsumerRecord<String, String>> mono) {
+        return mono
+                .doOnNext(rec -> {
+                    ...
+                });
+    }
+
+};
+adapter.parallel(10);
+container.setMessageListener(adapter);
+----
+====
+
+To use the adapter from a `@KafkaListener`, simply invoke it from the listener; the following Spring Boot application shows an example with random processing times for each record, together with the resulting log messages to see how the processing works.
+It sends 20 records and receives them in two batches of 10 - refer to the log to see how the commits are actioned.
+
+====
+[source, java]
+----
+@SpringBootApplication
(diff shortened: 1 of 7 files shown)


spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -1279,6 +1279,213 @@ The consumer will be paused (no new records delivered) until all the offsets for
...
+    @Override
+    public Mono<ConsumerRecord<String, String>> assemble(Mono<ConsumerRecord<String, String>> mono) {
+        return mono
+                .doOnNext(rec -> {
+                    ...
+                });
+    }
+
+};
+adapter.parallel(10);
```
artembilan, line 1313: It is a `setParallel()`

spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -1279,6 +1279,213 @@ The consumer will be paused (no new records delivered) until all the offsets for
...
+[parallel-5] Received: skReactor-0@624
+[parallel-9] Received: skReactor-0@628
+[parallel-6] Received: skReactor-0@625
+[parallel-7] Received: skReactor-0@626
+[parallel-3] Received: skReactor-0@622
+[parallel-8] acking:   skReactor-0@627
+[parallel-7] acking:   skReactor-0@626
+[parallel-3] acking:   skReactor-0@622
+[parallel-1] acking:   skReactor-0@620
+[skReactor-0-C-1] Received: 0 records
```
artembilan, line 1431: This is not related to the topic of this PR, but I see such a message in logs confusing: when we are paused, we definitely not going to receive any message. 
So, probably don't log it at all during pause period, or log something else indicating that we don't have data because it is paused.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ReactiveListenerAdapter.java
```diff
@@ -0,0 +1,159 @@
...
+	 * Set the SmartLifecyle phase - must start before the listener container.
+	 * @param phase the phase.
+	 */
+	public void setPhase(int phase) {
+		this.phase = phase;
+	}
+
+	@Override
+	public void propertiesSupplier(Supplier<ContainerProperties> containerProperties) {
+		ContainerProperties props = containerProperties.get();
```
artembilan, line 109: I wonder why it is a `Supplier` contract since you just call `get()` immediately, not later on demand or per some call?..

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ReactiveListenerAdapter.java
```diff
@@ -0,0 +1,159 @@
...
+
+	@Override
+	public void start() {
+		if (this.parallel > 0) {
+			this.disposable = this.sink.asFlux()
+					.parallel(this.parallel)
+					.runOn(this.scheduler)
+					.doOnNext(mono -> {
+						assemble(mono).subscribe();
+					})
```
artembilan, line 124: Probably this one should work instead:
```
.flatMap(this::assemble)
```
We kinda don't need to control interim subscriptions ourselves...

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ReactiveListenerAdapter.java
```diff
@@ -0,0 +1,159 @@
...
+ * ack mode and for async acks.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Gary Russell
+ * @since 2.8
+ *
+ */
+public abstract class ReactiveListenerAdapter<K, V> implements AcknowledgingMessageListener<K, V>, SmartLifecycle {
```
artembilan, line 50: I think this must not be a `SmartLifecycle`, but rather plain, dependent `Lifecycle`.
The container would control it and `@KafkaListener` delegation would require a `start()` call before using.
Or... We can get rid of this contract at all: and assemble that `disposable` on demand, when the first `onMessage()` is done.
Then we can dispose in `destroy()` or so.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ReactiveListenerAdapter.java
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+
+	@Override
+	public int getPhase() {
+		return this.phase;
+	}
+
+	/**
+	 * Set the SmartLifecyle phase - must start before the listener container.
```
artembilan, line 100: Typo, but if we remove `SmartLifecycle` contract, we are good.

artembilan (changes requested):
Not too much, but still something to think and/or redo.

Thanks

garyrussell:
Closing for now; deferring to a future release.

Issue #1938, opened 2021-09-13
Add Whole Batch Filtering


garyrussell:
Add the ability for a filter to iterate over the collection instead of one-at-a-time filtering.

See https://stackoverflow.com/questions/68934330/records-are-filtered-one-by-one-in-filteringbatchmessagelisteneradapter/68940594#comment122238384_68940594

PR #1942, opened 2021-09-14
GH-1938: Support Whole Batch Filtering

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1938

Previously, when adding a filter strategy to a batch listener, the batch
members were filtered one-at-a-time. When database lookups are involved
it is more efficient to act on the entire batch.

Add another method to the filter strategy to allow filtering the entire
batch in one call.


spring-kafka-docs/src/main/asciidoc/whats-new.adoc
```diff
@@ -27,6 +27,9 @@ Batch listeners can now handle conversion exceptions.
 
 See <<batch-listener-conv-errors>> for more information.
 
+`RecordFilterStrategy`, when used with batch listeners, can now filter the entire batch in one call.
+See <<>> for more information.
```
artembilan, line 31: Missed anchor.

artembilan (changes requested):
Just small doc concern.

PR #1946, opened 2021-09-17
GH-1704: Interceptor Improvements

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1704

- Extract `ThreadStateProcessor` and `ConsumerAwareThreadStateProcessor`
- Avoid all the if/else tests around calling the interceptor common methods
- Add `afterRecord` to the record interceptor so sleuth can defer cleaning up until after
  the error handler.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeBatchInterceptor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeBatchInterceptor.java
@@ -75,12 +75,12 @@ public class CompositeBatchInterceptor<K, V> implements BatchInterceptor<K, V> {
 	}
 
 	@Override
-	public void beforePoll(Consumer<K, V> consumer) {
-		this.delegates.forEach(del -> del.beforePoll(consumer));
+	public void setupThreadState(Consumer<?, ?> consumer) {
+		this.delegates.forEach(del -> del.setupThreadState(consumer));
 	}
 
 	@Override
-	public void clearThreadState(Consumer<K, V> consumer) {
+	public void clearThreadState(Consumer<?, ?> consumer) {
 		this.delegates.forEach(del -> del.clearThreadState(consumer));
 	}
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeRecordInterceptor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeRecordInterceptor.java
@@ -78,12 +78,13 @@ public class CompositeRecordInterceptor<K, V> implements ConsumerAwareRecordInte
 	}
 
 	@Override
-	public void beforePoll(Consumer<K, V> consumer) {
-		this.delegates.forEach(del -> del.beforePoll(consumer));
+	public void setupThreadState(Consumer<?, ?> consumer) {
+		this.delegates.forEach(del -> del.setupThreadState(consumer));
 	}
 
 	@Override
-	public void clearThreadState(Consumer<K, V> consumer) {
+	public void clearThreadState(Consumer<?, ?> consumer) {
 		this.delegates.forEach(del -> del.clearThreadState(consumer));
 	}
+
 }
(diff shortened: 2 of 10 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerAwareThreadStateProcessor.java
```diff
@@ -18,43 +18,34 @@
...
  *
- * @author Gary Russell
  * @author Karol Dowbecki
- * @author Artem Bilan
+ * @author Gary Russell
  * @since 2.8
  *
  */
-public interface BeforeAfterPollProcessor<K, V> {
+public interface ConsumerAwareThreadStateProcessor extends ThreadStateProcessor {
```
kdowbecki, line 31: From perspective of someone who uses the library and only infrequently digs into implementation it feels like common interfaces are getting quite complex to use. 

If I were to implement a new `RecordInterceptor` which methods exactly do I need to implement? I'd have to review how the `RecordInterceptor` interface is used to spot that only `ConsumerAwareThreadStateProcessor.setupThreadState(Consumer)` is called and `ConsumerAwareThreadStateProcessor.setupThreadState()` is not used,

Perhaps `ConsumerAwareThreadStateProcessor` should not extend `ThreadStateProcessor` if it's not overriding the parent methods? Or maybe it should override the methods and throw `UnsupportedOperationException` telling that the parent methods are not used?

spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeBatchInterceptor.java
```diff
@@ -75,12 +75,12 @@ public void failure(ConsumerRecords<K, V> records, Exception exception, Consumer
...
 	@Override
-	public void beforePoll(Consumer<K, V> consumer) {
-		this.delegates.forEach(del -> del.beforePoll(consumer));
+	public void setupThreadState(Consumer<?, ?> consumer) {
+		this.delegates.forEach(del -> del.setupThreadState(consumer));
 	}
 
 	@Override
-	public void clearThreadState(Consumer<K, V> consumer) {
+	public void clearThreadState(Consumer<?, ?> consumer) {
```
kdowbecki, line 83: Should it also delegate `setupThreadState()` and `clearThreadState()` inherited from `ThreadStateProcessor`?

spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeRecordInterceptor.java
```diff
@@ -78,12 +78,13 @@ public void failure(ConsumerRecord<K, V> record, Exception exception, Consumer<K
...
 	@Override
-	public void beforePoll(Consumer<K, V> consumer) {
-		this.delegates.forEach(del -> del.beforePoll(consumer));
+	public void setupThreadState(Consumer<?, ?> consumer) {
+		this.delegates.forEach(del -> del.setupThreadState(consumer));
 	}
 
 	@Override
-	public void clearThreadState(Consumer<K, V> consumer) {
+	public void clearThreadState(Consumer<?, ?> consumer) {
```
kdowbecki, line 86: Should it also delegate `setupThreadState()` and `clearThreadState()` inherited from `ThreadStateProcessor` and the new `afterRecord()` method?

spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerAwareThreadStateProcessor.java
garyrussell: Good point; I remove the inheritance.

garyrussell: I removed it altogether:

```
Remove ThreadStateProcessor (TSP), rename CATSP to TSP.
- confusing for interceptor implementors
- no components ever call setupThreadState; ARP and CEH build state during normal processing.
```

artembilan (approved):
LGTM.
If there is nothing else to add, I'll merge ASAP.

Thanks

garyrussell:
Thanks; all is good; I can then do the M3 release (unless you wouldn't mind doing it for me - I have updated the spring-build creds in the build plan for the main branch).

artembilan:
Merged. I can release it for you. No problem!
Enjoy your vacation!

kdowbecki:
@garyrussell  @artembilan shouldn't `CompositeRecordInterceptor` implement the new `afterRecord()` method as well?

Issue #1948, opened 2021-09-20
Add KafkaTemplate.receive(Collection<TopicPartitionOffset> records) Methods


garyrussell:
Extend the code for https://github.com/spring-projects/spring-kafka/issues/1861 to retrieve multiple records within the same consumer/call.

PR #1955, opened 2021-09-24
GH-1948: KafkaTemplate Receive Multiple Records

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1948

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -564,23 +564,12 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V>, ApplicationCo
 
 	@Override
 	@Nullable
-	public ConsumerRecord<K, V> receive(String topic, int partition, long offset) {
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -568,19 +571,49 @@ public void sendOffsetsToTransaction(Map<TopicPartition, OffsetAndMetadata> offs
...
+		Properties props = oneOnly();
+		try (Consumer<K, V> consumer = this.consumerFactory.createConsumer(null, null, null, props)) {
+			TopicPartition topicPartition = new TopicPartition(topic, partition);
+			return receiveOne(topicPartition, offset, pollTimeout, consumer);
+		}
+	}
+
+	@Override
+	public ConsumerRecords<K, V> receive(Collection<TopicPartitionOffset> requested) {
+		return receive(requested, DEFAULT_POLL_TIMEOUT);
```
artembilan, line 583: Looks like this can go as a `default` to the `KafkaOperations`.

artembilan (changes requested):

Issue #1964, opened 2021-09-28
Enhance DelegatingDeserializer to optionally select the delegate by topic


garyrussell:
https://github.com/spring-projects/spring-kafka/issues/430#issuecomment-929190319

PR #1965, opened 2021-09-28
GH-1964: Add DelegatingByTopic(De)Serializer

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1964

Select the `Serializer`/`Deserializer` based on the topic name.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
@@ -1,0 +1,229 @@
+/*
+ * Copyright 2019-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.support.serializer;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Pattern;
+
+import org.apache.kafka.common.header.Headers;
+import org.apache.kafka.common.serialization.Deserializer;
+
+import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+import org.springframework.util.StringUtils;
+
+/**
+ * A {@link Deserializer} that delegates to other deserializers based on the topic name.
+ *
+ * @author Gary Russell
+ * @since 2.8
+ *
+ */
+public class DelegatingByTopicDeserializer implements Deserializer<Object> {
+
+	private final Map<Pattern, Deserializer<? extends Object>> delegates = new ConcurrentHashMap<>();
+
+	private boolean forKeys;
+
+	private boolean cased = true;
+
+	/**
+	 * Construct an instance that will be configured in {@link #configure(Map, boolean)}
+	 * with consumer properties.
+	 */
+	public DelegatingByTopicDeserializer() {
+	}
+
+	/**
+	 * Construct an instance with the supplied mapping of topic name patterns to delegate
+	 * deserializers.
+	 * @param delegates the map of delegates.
+	 */
+	public DelegatingByTopicDeserializer(Map<Pattern, Deserializer<?>> delegates) {
+		this.delegates.putAll(delegates);
+	}
+
+	/**
+	 * Set to false to make topic name matching case insensitive.
+	 * @param caseSensitive false for case insensitive.
+	 */
+	public void setCaseSensitive(boolean caseSensitive) {
+		this.cased = caseSensitive;
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public void configure(Map<String, ?> configs, boolean isKey) {
+		this.forKeys = isKey;
+		Object insensitive = configs.get(DelegatingByTopicSerializer.CASE_SENSITIVE);
+		if (insensitive instanceof String) {
+			this.cased = Boolean.parseBoolean((String) insensitive);
+		}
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
```diff
@@ -0,0 +1,229 @@
...
+		}
+	}
+
+	public void addDelegate(Pattern pattern, Deserializer<?>  deserializer) {
+		this.delegates.put(pattern, deserializer);
+	}
+
+	@Nullable
+	public Deserializer<?> removeDelegate(String selector) {
+		return this.delegates.remove(selector);
```
artembilan, line 184: How does it work if we put by `Pattern`, but remove by `String`?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
```diff
@@ -0,0 +1,229 @@
...
+/**
+ * A {@link Deserializer} that delegates to other deserializers based on the topic name.
+ *
+ * @author Gary Russell
+ * @since 2.8
+ *
+ */
+public class DelegatingByTopicDeserializer implements Deserializer<Object> {
+
+	private final Map<Pattern, Deserializer<? extends Object>> delegates = new ConcurrentHashMap<>();
```
artembilan, line 42: Don't you consider something like `defaultDeserializer`?
Something like fallback option when no pattern matching, kinda for the rest of use-cases. Just because we can have only one (de)serializer might be good to have a flexibility.
WDYT?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
```diff
@@ -0,0 +1,229 @@
...
+	public void setCaseSensitive(boolean caseSensitive) {
+		this.cased = caseSensitive;
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public void configure(Map<String, ?> configs, boolean isKey) {
+		this.forKeys = isKey;
+		Object insensitive = configs.get(DelegatingByTopicSerializer.CASE_SENSITIVE);
+		if (insensitive instanceof String) {
```
artembilan, line 77: Can it be a plain `boolean`, too?
Shouldn't we fail if it is some other unexpected type?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicSerializer.java
```diff
@@ -0,0 +1,245 @@
...
+			this.delegates.putAll(createDelegates((String) value, configs, isKey));
+		}
+		else {
+			throw new IllegalStateException(
+					configKey + " must be a map or String, not " + value.getClass());
+		}
+	}
+
+	@SuppressWarnings("unchecked")
+	private void processMap(Map<String, ?> configs, boolean isKey, String configKey, Object value) {
```
artembilan, line 118: I think the `value` can directly be a `Map<Object, Object>`. You don't call this method unless `value instanceof Map`.
Am I missing anything specific?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicSerializer.java
```diff
@@ -0,0 +1,245 @@
...
+	/**
+	 * Name of the configuration property containing the serialization topic pattern map for
+	 * keys with format {@code pattern:class,...}.
+	 */
+	public static final String KEY_SERIALIZATION_TOPIC_CONFIG = "spring.kafka.key.serialization.bytopic.config";
+
+	/**
+	 * Set to false to make topic pattern matching case-insensitive.
+	 */
+	public static final String CASE_SENSITIVE = "spring.kafka.value.serialization.bytopic.case.insensitive";
```
artembilan, line 60: I wonder if we should have copies (references) for these constants from the deserializer.
Kinda thinking how end-users are going to search for similar constants when they configure deserializer...
(Might be an overhead on my side - don't mind 😄 )

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicSerializer.java
```diff
@@ -0,0 +1,245 @@
...
+						+ " map entries must be Serializers or class names, not " + value.getClass());
+			}
+		});
+	}
+
+	private String configKey() {
+		return this.forKeys ? KEY_SERIALIZATION_TOPIC_CONFIG : VALUE_SERIALIZATION_TOPIC_CONFIG;
+	}
+
+	protected Map<Pattern, Serializer<?>> createDelegates(String mappings, Map<String, ?> configs,
```
artembilan, line 142: Another crazy thought: how about to have a super class (or implementation of both `(De)Serializer`) with all of these common API?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
garyrussell: oops (copy/paste)

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicSerializer.java
garyrussell: Good suggestions, all.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	@Override
+	protected boolean isInstance(Object delegate) {
+		return delegate instanceof Deserializer;
+	}
+
+	@Override
+	public Object deserialize(String topic, byte[] data) {
+		throw new UnsupportedOperationException();
```
artembilan, line 70: Why not? Why can't we delegate same way as with headers?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
garyrussell: Kafka never calls this method (since 0.11).

PR #1969, opened 2021-10-01
doc: fix topic name in RequestReply-Example

j0xaf:
The Topic created in `kReplies()` is `"kReplies"`, so this should be reflected in the `repliesContainer`.


artembilan:
@pivotal-cla , this is Obvious Fix.
Merging.

@j0xaf ,
thank you very much for contribution; looking forward for more!

Issue #1974, opened 2021-10-05
Default SimpleAsyncTaskExecutor is Not Reused


garyrussell:
The child container attempts to save the executor in its properties, but each time the container is started, it gets a new set of properties.

artembilan:
See SO for reference: https://stackoverflow.com/questions/69448170/thread-name-convention-in-spring-kafka-concurrentmessagelistenercontainer-consum

PR #1975, opened 2021-10-05
GH-1974: Fix Default Task Executor

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1974

A new `SimpleAsyncTaskExecutor` was created for each child container when
the container was stopped/started.
Reuse the same container for each child; handle an increase to `concurrency` between
stop and start

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
@@ -33,6 +33,8 @@ import org.apache.kafka.common.TopicPartition;
 
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.core.task.AsyncListenableTaskExecutor;
+import org.springframework.core.task.SimpleAsyncTaskExecutor;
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -230,6 +234,19 @@ private void configureChildContainer(int index, KafkaMessageListenerContainer<K,
...
 		});
+		AsyncListenableTaskExecutor exec = container.getContainerProperties().getConsumerTaskExecutor();
+		if (exec == null) {
+			if ((this.executors.size() > index)) {
+				exec = this.executors.get(index);
+			}
+			else {
+				String containerBeanName = container.getBeanName();
+				exec = new SimpleAsyncTaskExecutor(
+						(containerBeanName == null ? "" : containerBeanName) + "-C-");
```
artembilan, line 245: It cannot be `null`. See logic in the beginning of this method: `container.setBeanName((beanName != null ? beanName : "consumer") + "-" + index);`

artembilan (changes requested):

artembilan:
@garyrussell , do GH Actions failures say anything to you?
Thanks

Issue #1981, opened 2021-10-19
Add an Option to Not Configure a DLT Consumer


garyrussell:
See https://stackoverflow.com/questions/69571401/dlt-message-consumption/69621464#comment123078112_69621464

Currently, if the user doesn't specify a DLT handler, a default (logging) handler is configured; add an option to not configure one at all.

PR #1985, opened 2021-10-21
GH-1981: Optionally Don't Start the DLT Container

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1981


tomazfernandes:
Hi @garyrussel, so it turns out there is a use case for not having a simple logging consumer after all! 😄 

I like the approach of creating the DLT consumer and not starting. My initial thought was to simply not create the consumer,
but this way the consumer is there if the user wants to start it, and this should also work if it's a custom DLT handler.

What I'm thinking is, what if instead of having an autoStartup flag, we let the users optionally choose a dedicated listener container factory for the DLT consumer in the annotation / builder? It might give the users even more configuration flexibility for the DLT and leverage the already existing autoStartup flag from the factory.

Of course that might be an overkill - I like the solution as is too!

PR #2002, opened 2021-11-08
Initial Commit for Spring Framework 6.0/JDK 17

garyrussell:


build.gradle
```diff
@@ -68,7 +73,7 @@ ext {
 	springBootVersion = '2.5.5' // docs module
 	springDataVersion = '2021.1.0-RC1'
```
artembilan, line 74: This one has to be `2022.1.0-SNAPSHOT`.
And Boot I guess has to move to `2.6.0`.
Jackson `2.13.0`.
Kotlin `1.5.31`

build.gradle
```diff
@@ -499,6 +512,11 @@ task api(type: Javadoc) {
 	destinationDir = new File(buildDir, 'api')
 }
 
+tasks.withType(Javadoc) {
+	options.addBooleanOption('Xdoclint:syntax', true) // only check syntax with doclint
+	options.addBooleanOption('Werror', true) // fail build on Javadoc warnings
+}
```
artembilan, line 518: This block has to be moved to the `subprojects` section.
Turns out we build javadoc jar from every single module.
And then we also build a single consolidated javadoc.
Don't ask me why we do that: probably SF does the same.
Or they just don't publish javadoc jars...

artembilan (changes requested):
That's it.
If build passes for you and we don't need to do any docs changes, we can merge this after the mentioned clean ups.

Thanks

Issue #1902, opened 2021-08-05
Consumer stuck at high cpu when reconnecting to broker fails by AuthenticationException


thimmwork:
**Affects Version(s):** 2.6.10 (also observed in 2.6.9, possibly more)
Stacktraces below reference 2.6.10

Setup and preface:
We're running Spring Boot 2.4.8, kafka-clients 2.6.2 in Kubernetes with 2 service instances (let's call them A and B) sharing the same consumer group.

Due to a scheduled maintenance, authentication to our broker failed for a short time.
When service instance A went down, a consumer rebalance was triggered. a new instance C was created instead, tried to connect to the broker and failed in the expected way.

The issue:
This occurs on (the surviving) instance B, which got stuck at max CPU.
The client logged the following errors:
> [Consumer clientId=my-client-5, groupId=my-group-id] Connection to node 14 (broker-14.<url>/<ip:port>) failed authentication due to: Authentication failed: Invalid username or password

and

> java.lang.IllegalStateException: This error handler cannot process 'org.apache.kafka.common.errors.SaslAuthenticationException's; no record information is available
> 	at org.springframework.kafka.listener.SeekUtils.seekOrRecover(SeekUtils.java:151)
> 	at org.springframework.kafka.listener.SeekToCurrentErrorHandler.handle(SeekToCurrentErrorHandler.java:113)
> 	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.handleConsumerException(KafkaMessageListenerContainer.java:1439)
> 	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.run(KafkaMessageListenerContainer.java:1136)
> 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
> 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
> 	at java.lang.Thread.run(Thread.java:748)
> Caused by: org.apache.kafka.common.errors.SaslAuthenticationException: Authentication failed: Invalid username or password

After that, CPU usage raised from moderate 5% to 700% and did not go down - even hours later.
When authentication against our broker worked again, instance C connected in a regular fashion.
Instance A remained stuck at max CPU and did not recover.

Instance A is dead now, but I managed to pull a thread dump, if that's of any help:

> java.lang.Thread.State: RUNNABLE
> 	at sun.nio.ch.EPollArrayWrapper.epollWait(Native Method)
[...]

PR #2004, opened 2021-11-09
GH-1902: AuthenticationException Fatal By Default

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/1902

Rename property and use for both exception types.


artembilan (changes requested):
> Error: eckstyle] [ERROR] /home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java:351: Javadoc has empty description section. [JavadocStyle]

PR #2011, opened 2021-11-17
Fix typo

chanhyeong:
There is a missing left quote in [doc](https://docs.spring.io/spring-kafka/docs/current/reference/html/#x28-kafka-client).


artembilan:
Thank you, @chanhyeong , for contribution; looking forward for more!

PR #2032, opened 2021-12-07
Change streamsBuilderFactoryBean setListener to addListener

michalkowol:


artembilan:
@michalkowol ,

PR #2058, opened 2021-12-24
Add MediaMessage annotation support.

scruel:
Supports the ability to define the content type of Message from the client side for resolving.

Usage example:
```
@Configuration
public class MediaMessageConfig implements KafkaListenerConfigurer {
  public CompositeMessageConverter createMessageConverter() {
    List<MessageConverter> converters = new ArrayList<>();
    converters.add(new StringMessageConverter());
    converters.add(new ByteArrayMessageConverter());
    converters.add(createJacksonConverter());
    return new CompositeMessageConverter(converters);
  }

  protected MappingJackson2MessageConverter createJacksonConverter() {
    MappingJackson2MessageConverter converter = new MappingJackson2MessageConverter();
    converter.setContentTypeResolver(new DefaultContentTypeResolver());
    return converter;
  }

  @Override
  public void configureKafkaListeners(@JsonDeserialize KafkaListenerEndpointRegistrar registrar) {
    registrar.setCustomMethodArgumentResolvers(new MediaMessageResolver(createMessageConverter()));
  }
}


public class ClientConsumer {

@KafkaListener(topics = <topics>, groupId = <groupId>)
  public void onMessage(@MediaMessage BusinessDTO info) {
    ...
  }
}
```


spring-kafka/src/main/java/org/springframework/kafka/support/resolver/MediaMessageResolver.java
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	@Override
+	public Object resolveArgument(MethodParameter parameter, Message<?> message) throws Exception {
+		MediaMessage ann = parameter.getParameterAnnotation(MediaMessage.class);
+		if (null == ann) {
+			throw new IllegalStateException("Annotation parsing failed.");
+		}
+		// If not present, parse content type from annotation
+		if (message.getHeaders().get(MessageHeaders.CONTENT_TYPE) == null) {
```
garyrussell, line 56: What about when `force` is true?

garyrussell (changes requested):
This looks interesting, but `@MediaMessage` doesn't seem to be correct. Perhaps something like `@ResolvableType`?

Also, this needs test coverage.

spring-kafka/src/main/java/org/springframework/kafka/support/resolver/MediaMessageResolver.java
scruel: > What about when `force` is true?

Oh, It seems that I forgot to complete the code of force logic part, let me finish it, thanks for the notice!

scruel:
> This looks interesting, but `@MediaMessage` doesn't seem to be correct. Perhaps something like `@ResolvableType`?
> 
> Also, this needs test coverage.

Done, for the name part, maybe `@MediaType` would be better?

spring-kafka/src/test/java/org/springframework/kafka/support/resolver/ResolvableTypeResolverTests.java
```diff
@@ -0,0 +1,165 @@
...
+				{
+					if (KafkaListenerConfigurer.class.equals(type)) {
+						put("bean", (T) new MediaMessageConfig());
+					}
+				}
+			};
+		}
+	}
+
+	static class MediaMessageConfig implements KafkaListenerConfigurer {
```
garyrussell, line 131: Still using `MediaMessage` here.

spring-kafka/src/main/java/org/springframework/kafka/annotation/ResolvableType.java
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.annotation;
+
+import org.springframework.util.MimeTypeUtils;
+
+import java.lang.annotation.*;
```
garyrussell, line 21: We don't allow wildcard imports; please run `./gradlew check` to check style.

garyrussell (changes requested):
Thanks; it needs some docs; probably in this section https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka-docs/src/main/asciidoc/kafka.adoc#kafkalistener-annotation

Also, maybe a "how to use" here https://docs.spring.io/spring-kafka/docs/current/reference/html/#tips-n-tricks

garyrussell:
@artembilan Any thoughts on the annotation name?

spring-kafka/src/main/java/org/springframework/kafka/annotation/ResolvableType.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Used to define the content type of {@code Message} with {@code MediaMessageResolver}.
+ *
+ * @author Scruel Tao
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.PARAMETER})
+public @interface ResolvableType {
```
artembilan, line 31: There is already a `org.springframework.core.ResolvableType`.
So, this one is going to confuse.
Plus it doesn't look like we really resolve a type, but talk more about a `content-type`.

I'm not fully sure that I follow with the goal of this feature, but probably better to have it as an `@ExpectedContentType` since there are many `ContentType` classes around 😸 .
I would say this is similar to `consumes` on the `@RequestMapping`, but you really have here a fallback to the expected if no one provided in the record.
Anyway I'd like to understand better what is going on.
Looks like you want to have a common composite converter and choose an appropriate one according the `content-type` header or fallback.
How about have an attribute on the `@KafkaListener` instead? Kinda `fallbackContentType`?
Populate it to the message (if needed) and still rely on that composite ability to choose an expected converter.

Just thinking aloud...

Thanks

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/annotation/ResolvableType.java
garyrussell: Thanks, Artem, for your perspective; I agree coercing the content type in the listener adapter via a `@KafkaListener` property would be better than polluting the method signature.

scruel: > Thanks, Artem, for your perspective; I agree coercing the content type in the listener adapter via a `@KafkaListener` property would be better than polluting the method signature.

In spring, annotation on parameters is very common, I think it can not treat as polluting the method signature, for example `@RequestBody`.

spring-kafka/src/test/java/org/springframework/kafka/support/resolver/ResolvableTypeResolverTests.java
```diff
@@ -0,0 +1,165 @@
...
+			List<MessageConverter> converters = new ArrayList<>();
+			converters.add(new StringMessageConverter());
+			converters.add(new ByteArrayMessageConverter());
+			converters.add(createJacksonConverter());
+			return new CompositeMessageConverter(converters);
+		}
+
+		protected MappingJackson2MessageConverter createJacksonConverter() {
+			MappingJackson2MessageConverter converter = new MappingJackson2MessageConverter();
+			converter.setContentTypeResolver(new DefaultContentTypeResolver());
```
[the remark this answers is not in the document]
scruel: It is just a custom usage, If needed, it can change.

spring-kafka/src/test/java/org/springframework/kafka/support/resolver/ResolvableTypeResolverTests.java
```diff
@@ -0,0 +1,165 @@
...
+			return converter;
+		}
+
+		@Override
+		public void configureKafkaListeners(@JsonDeserialize KafkaListenerEndpointRegistrar registrar) {
```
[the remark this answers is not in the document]
scruel: It was a mistake, I've removed this.

spring-kafka/src/main/java/org/springframework/kafka/annotation/ResolvableType.java
scruel: Done

spring-kafka/src/main/java/org/springframework/kafka/annotation/ResolvableType.java
artembilan: Yes, but we already have enough param annotations to pollute enough. 
And your idea makes me think that this kind of `fallbackContentType` should go to the existing `@Payload` annotation which we rely in Spring Kafka as well.
But again: your current goal could be achieved with an explicit converter for the particular `@KakfaListener`. The rest I believe is already covered by the composition if `contentType` header is present.

scruel: Hah, `@Payload` annotation is in `org.springframework.messaging` package but not this project, I treat the `@MediaMessage` as an enhanced annotation for `spring-kafka`.
I know it was covered if `contentType` header is present, but sometimes, you can not control the producers for which message they will send.

spring-kafka/src/test/java/org/springframework/kafka/support/resolver/ResolvableTypeResolverTests.java
```diff
@@ -0,0 +1,165 @@
...
+	private void beanPostProcessorConfig(KafkaListenerAnnotationBeanPostProcessor processor) {
+		processor.setBeanFactory(new KafkaBeanFactory());
+		try {
+			processor.afterSingletonsInstantiated();
+		} catch (NoSuchBeanDefinitionException ignore) {
+		}
+	}
+
+	@SuppressWarnings("unchecked")
+	static class KafkaBeanFactory extends DefaultListableBeanFactory {
```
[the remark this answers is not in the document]
scruel: It used to invoke `setCustomMethodArgumentResolvers`, because all relevant methods are private, so I choose this way to achieve the goal.

spring-kafka/src/test/java/org/springframework/kafka/support/resolver/ResolvableTypeResolverTests.java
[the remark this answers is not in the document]
scruel: By the way, This code snippet is just a copy from `org.springframework.messaging.simp.config.AbstractMessageBrokerConfiguration#createJacksonConverter`(Line 474)

spring-kafka/src/main/java/org/springframework/kafka/annotation/ResolvableType.java
scruel: > Thanks; it needs some docs; probably in this section https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka-docs/src/main/asciidoc/kafka.adoc#kafkalistener-annotation
> 
> Also, maybe a "how to use" here https://docs.spring.io/spring-kafka/docs/current/reference/html/#tips-n-tricks

I will fulfill the docs after you make your decision.

scruel: > Thanks, Artem, for your perspective; I agree coercing the content type in the listener adapter via a `@KafkaListener` property would be better than polluting the method signature.

Would you need me also to implement this idea?

spring-kafka/src/test/java/org/springframework/kafka/support/resolver/ResolvableTypeResolverTests.java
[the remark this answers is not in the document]
artembilan: This is test code. I believe the test code should be as simple as  possible.
It must answer questions, but not raise them.
More over your reference is misleading: you don't set a `defaultMimeType` like it is done in that `AbstractMessageBrokerConfiguration`:
```
protected MappingJackson2MessageConverter createJacksonConverter() {
		DefaultContentTypeResolver resolver = new DefaultContentTypeResolver();
		resolver.setDefaultMimeType(MimeTypeUtils.APPLICATION_JSON);
		MappingJackson2MessageConverter converter = new MappingJackson2MessageConverter();
		converter.setContentTypeResolver(resolver);
		return converter;
	}
```

spring-kafka/src/test/java/org/springframework/kafka/support/resolver/ResolvableTypeResolverTests.java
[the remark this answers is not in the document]
artembilan: OK. I see your point. It looks like you try mock some infrastructure for `spring-kafka`.
Let it be for now!
But I think it can be simplified anyway. See `HeaderMethodArgumentResolverTests` for example.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/annotation/ResolvableType.java
scruel: Thanks, I agree that the other projects have the rights to benefit from this.
> they still need to support an infrastructure for converter composition and so on.
I do want to add this in default configuration, but maybe we should consider how to implement this first, then consider should this support feature as a default feature.
I will wait for your decision, feel free to comment your ideas.

spring-kafka/src/test/java/org/springframework/kafka/support/resolver/ResolvableTypeResolverTests.java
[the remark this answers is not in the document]
scruel: Ok, thanks, I am a little busy right now, so will check this tomorrow.

spring-kafka/src/test/java/org/springframework/kafka/support/resolver/ResolvableTypeResolverTests.java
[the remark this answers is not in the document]
scruel: I agree that the test code should be as simple as possible, but it seems that this test case really need to verify some media type like `application/json` can be correctly process by the converter which support for converting it.
Set a `defaultMimeType` is not necessary, since the CONTENT_TYPE header will not be null at all.
(annotation type attribute can't be null, because null is not a constant value)
See `org.springframework.messaging.converter.DefaultContentTypeResolver#resolve`(Line 62)
and with the logic:
```
if (header.CONTENT_TYPE == null) {
  header.CONTENT_TYPE = ann.type();
}
```

spring-kafka/src/test/java/org/springframework/kafka/support/resolver/ResolvableTypeResolverTests.java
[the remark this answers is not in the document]
scruel: @artembilan Checked, and the code already based on `HeaderMethodArgumentResolverTests`, which is not contained the logic I need.

artembilan:
Closed as `Not planned`.

Issue #2061, opened 2022-01-08
EmbeddedKafkaBroker#consumeFromEmbeddedTopics ignores  latest offset reset strategy


coney:
According to the document https://docs.spring.io/spring-kafka/reference/html/#junit it says:

> Starting with version 2.5, the consumerProps method sets the ConsumerConfig.AUTO_OFFSET_RESET_CONFIG to earliest. This is because, in most cases, you want the consumer to consume any messages sent in a test case. The ConsumerConfig default is latest which means that messages already sent by a test, before the consumer starts, will not receive those records. To revert to the previous behavior, set the property to latest after calling the method.

I created my consumer and manually overrides the `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG` to `latest` as below:
``` java
    Map<String, Object> consumerProps = KafkaTestUtils.consumerProps(generateConsumerGroupName(), "true",
            embeddedKafkaBroker);
        consumerProps.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "latest");
        consumerProps.put(JsonDeserializer.TRUSTED_PACKAGES, "*");
        consumer = new DefaultKafkaConsumerFactory<>(consumerProps, keyDeserializer, valueDeserializer)
            .createConsumer();

        // start consuming before any records are written to kafka, make sure offset moves to latest
        embeddedKafkaBroker.consumeFromAnEmbeddedTopic(consumer, topic);
```

But implementation of `EmbeddedKafkaBroker#consumeFromEmbeddedTopics`, it tries to poll records and reset the offset to earilest if there are records present:

``` java
public void consumeFromEmbeddedTopics(Consumer<?, ?> consumer, String... topicsToConsume) {
               ...
		ConsumerRecords<?, ?> records = null;
		int n = 0;
		while (!assigned.get() && n++ < 600) { // NOSONAR magic #
			records = consumer.poll(Duration.ofMillis(100)); // force assignment NOSONAR magic #
		}
		if (records != null && records.count() > 0) {
			final ConsumerRecords<?, ?> theRecords = records;
			logger.debug(() -> "Records received on initial poll for assignment; re-seeking to beginning; "
					+ theRecords.partitions().stream()
					.flatMap(p -> theRecords.records(p).stream())
					// map to same format as send metadata toString()
					.map(r -> r.topic() + "-" + r.partition() + "@" + r.offset())
					.collect(Collectors.toList()));
[...]

PR #2064, opened 2022-01-11
GH-2061: EKB Seek to End Option for Embedded Topics

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2061

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2021 the original author or authors.
+ * Copyright 2018-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java
```diff
@@ -732,27 +765,30 @@ public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
...
-					.collect(Collectors.toList()));
-			consumer.seekToBeginning(records.partitions());
-		}
-		if (!assigned.get()) {
+			logger.debug(() -> "Partitions assigned "
+					+ assigned.get()
+					+ "; re-seeking to "
+					+ (seekToEnd ? "end; " : "beginning"));
+			if (seekToEnd) {
+				consumer.seekToEnd(assigned.get());
```
artembilan, line 785: > 		logger.debug(() -> "Records received on initial poll for assignment; re-seeking to beginning; "
					+ theRecords.partitions().stream()

Is it really correct to `seekToEnd` when we already got this records on a first poll?
Or that doesn't matter and we are not interested in this records any more?

garyrussell: >Or that doesn't matter and we are not interested in this records any more?

Correct; the seek was added because, sometimes, the poll that causes the assignment could fetch some records (generally not, but sometimes) and this caused some tests to fail occasionally because we never saw those records. Equally, I suspect, we might not get all records (from all partitions) on this initial poll so, to satisfy the reported problem, we just seek all the assignments (to beginning or end).

This way, the seeks are unconditional, regardless of whether we actually get any records here.

artembilan: Oh! Right. I see now that you have removed that `if (records != null && records.count() > 0) {`.
It is still there in `main` code, so I was confused.

LGTM now. 
Merging...

Issue #2070, opened 2022-01-13
Update the DefaultProducerFactory to support config overrides of transactional producers 


frosiere:
In some specific use cases, it may be required to update the configuration entries of a transactional consumer before creating it.

To allow this, a slight update of the DefaultKafkaProducer#doCreateTxProducer method may be done as follow

```
// private -> protected
// new rawConfigs parameters
protected CloseSafeProducer<K, V> doCreateTxProducer(String prefix, String suffix,
        BiPredicate<CloseSafeProducer<K, V>, Duration> remover, Map<String, Object> rawConfigs) {
    // same code as now
}
```

This update would allow a pre-processing of configuration entries as follow 

```
public class CustomDefaultKafkaProducerFactory<K, V> extends DefaultKafkaProducerFactory<K, V> {
	
	@Override
	protected CloseSafeProducer<K, V> doCreateTxProducer(String prefix, String suffix, BiPredicate<CloseSafeProducer<K, V>, Duration> remover, Map<String, Object> rawConfigs) {
		final Map<String, Object> newProducerConfigs = new HashMap<>(rawConfigs);
		// overrides
		return super.doCreateTxProducer(prefix, suffix, remover, newProducerConfigs);
	}
}
```

This issue follows a Stackoverflow discussion described [here](https://stackoverflow.com/questions/70697281/same-kafkatemplate-with-different-configuration-for-transactional-and-non-transa).

PR #2075, opened 2022-01-17
Updated the DefaultKafkaProducerFactory to support producer configs overrides

frosiere:
The DefaultKafkaProducerFactory has been updated to support producer config overrides as described in the issue #2070.

Two protected methods have been created to allow retrieving the configs of transactional and non-transactional producers.
These methods can be overridden in case of override.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 the original author or authors.
+ * Copyright 2016-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -908,6 +888,22 @@ public void closeThreadBoundProducer() {
 		}
 	}
 
+	protected Map<String, Object> getProducerConfigs() {
```
garyrussell, line 891: `protected` methods require javadocs.

garyrussell (changes requested):
Thanks for the contribution! LGTM; just one comment.

In future, please restrict the first line of the commit comment to 50 chars (see the contribution guidelines).

Add `Resolves https://github.com/spring-projects/spring-kafka/issues/2070` to the next commit so that the issue will automatically close when this is merged.

frosiere:
> Thanks for the contribution! LGTM; just one comment.
> 
> In future, please restrict the first line of the commit comment to 50 chars (see the contribution guidelines).
> 
> Add `Resolves https://github.com/spring-projects/spring-kafka/issues/2070` to the next commit so that the issue will automatically close when this is merged.

Thanks for the review.
Missing javadoc has been added.

garyrussell:
Thanks; merged as 0fc1548dd49920fc8867fa6a1e299c95df5cc659 cherry-picked as b669c12df7358ef35f542090b07d7bc9bed40149

Issue #2082, opened 2022-01-24
RecordInterceptor Logic Incorrect


garyrussell:
Early interceptor is now called when there is no tx manager, regardless of `isInterceptBeforeTx`.

However, early intercepting does not include the delivery attempt header https://stackoverflow.com/questions/70839849/problem-reading-kafka-headers-in-a-recordinterceptor-after-upgrading-to-spring-k

Regression: #1929

PR #2084, opened 2022-01-24
GH-2082: Fix RecordInterceptor

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2082

Now that the `earlyRecordInterceptor` is always used unless explicitly disabled,
move delivery attempt header processing earlier so that the header is available
in the interceptor.


artembilan:
... and cherry-picked to `2.8.x` & `2.7.x`.

PR #2087, opened 2022-01-26
JsonSerde Improvement

garyrussell:
Narrow return types of `serializer()`, `deserializer()` to avoid casting
if the user wants to programmatically customize the objects.

**cherry-pick to 2.8.x**


artembilan:
... and cherry-picked to `2.8.x`

Issue #2076, opened 2022-01-18
Consumer crash with StackOverflowError


Usernamelesss:
Hi,

a few days ago I experienced a reproducible StackOverflowError with a consumer. 
I have a high traffic topic with 100 partitions and I use KafkaMessageListenerContainer with asyncCommit enabled.

I was able to reproduce this error by writing 500.000 random messages on a topic before starting the consumer and I used docker-compose to deploy a kafka instance locally.
The consumer simply logs the received message. When the consumer starts, it consumes from the topic for a few minutes and then it crashes with a StackOverflowError.

Basically the consumer looks like this:
```    
public ConcurrentMessageListenerContainer<String, String> kafkaContainer() {
        ContainerProperties containerProperties = new ContainerProperties(groupId);
        containerProperties.setAckMode(ContainerProperties.AckMode.RECORD);
        containerProperties.setAckOnError(false);
        containerProperties.setSyncCommits(false);
        containerProperties.setLogContainerConfig(true);
        RetryingMessageListenerAdapter<String, String> adapter = new RetryingMessageListenerAdapter<>(
                messageHandler(),
                getRetryTemplate());
        containerProperties.setMessageListener(adapter);
        ConcurrentMessageListenerContainer<String, String> container = new ConcurrentMessageListenerContainer<>(
                consumerFactory(), containerProperties);
        container.setConcurrency(1);
        container.setErrorHandler(new SeekToCurrentErrorHandler(
                (record, exception) -> {
                    adapter.onMessage((ConsumerRecord<String, String>) record);
                }, new FixedBackOff(2000, FixedBackOff.UNLIMITED_ATTEMPTS)));
        return container;
    }

    private MessageListener<String, String> messageHandler() {
        return data -> System.out.println("Received " + data.value());
    }

    private RetryTemplate getRetryTemplate() {
        RetryTemplate retryTemplate = new RetryTemplate();
        ExponentialBackOffPolicy backOffPolicy = new ExponentialBackOffPolicy();
        backOffPolicy.setMaxInterval(10);
        backOffPolicy.setMultiplier(2);
        retryTemplate.setBackOffPolicy(backOffPolicy);
        SimpleRetryPolicy simpleRetryPolicy = new SimpleRetryPolicy();
        simpleRetryPolicy.setMaxAttempts(Integer.MAX_VALUE);
        retryTemplate.setRetryPolicy(simpleRetryPolicy);
        return retryTemplate;
    }
```

[...]

PR #2088, opened 2022-01-27
GH-2076: Fix Async Commit Retries

garyrussell:
https://github.com/spring-projects/spring-kafka/issues/2076

Do not attempt to retry asynchronous commits.
- a later commit for the same topic/partition may have already succeeded
- many consecutive retryable exceptions can cause a stack overflow

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -1452,7 +1452,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 								commitSync(toFix);
 							}
 							else {
-								commitAsync(toFix, 0);
+								commitAsync(toFix);
 							}
 						}
 						else {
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1937,12 +1937,8 @@ else if (this.syncCommits) {
 
 		private void commitAsync(Map<TopicPartition, OffsetAndMetadata> commits, int retries) {
```
artembilan, line 1938: Looks like `retries` param is out of use now...
Since we don't retry async commit any more, I guess it is time to let this `retries` param go...

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: Doh - yes - I meant to clean that up.

artembilan:
Back-ported to `2.8.x` as https://github.com/spring-projects/spring-kafka/commit/9722b8420b670eb256dc9000e53e5269a66253d4 after fixing conflicts.
And as https://github.com/spring-projects/spring-kafka/commit/45a6b06ba2a03fa442a204650adaad5545f96e96 to `2.7.x` after more conflicts.

PR #2093, opened 2022-02-02
add @Nullable to FailedRecordTracker.setBackOffFunction

### LINKED ISSUE #2092: Cannot assign null value into backOffFunction when using kotlin
**Affects Version(s):** over 2.6.x

image

Can't put null value to `FailedRecordProcessor#setBackOffFunction`
This is a similar issue with #1988 

It is a nullable field because there is a logic like below
https://github.com/spring-projects/spring-kafka/blob/676a5ad667839b1b96240e1a87e8876d300479e2/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java#L215-L221

#### Versions
- kotlin: 1.5.31
- spring boot: 2.5.x
- spring kafka: 2.7.9

chanhyeong:
Resolves #2092


garyrussell (commented):
Thanks; it is a trivial change, but it's not clear why you need to set it to `null`.

Issue #2069, opened 2022-01-13
Consider Allowing the Same Container Factory for Retryable and Normal @KafkaListener


garyrussell:
See https://stackoverflow.com/questions/70701185/infinite-loop-when-using-retryabletopic/70701460#70701460

garyrussell:
The `ListenerContainerFactoryConfigurer` unconditionally uses a customizer to `setupBackoffAwareMessageListenerAdapter()`.

cc/ @tomazfernandes Thoughts?

tomazfernandes:
Well, seems like an useful and interesting challenge - from the top of my head I can't really see a simple solution for this... The factory that will create the retryable topics' containers needs to have both the customizer and the error handler with the DLPR, and the factory that will create the non retryable topics' container can't have either.

Maybe we could add some conditional logic to the customizer so it'd skip certain containers. The error handler should be trickier though.

Did you think of something? I can look deeper into it next week if you'd like, or feel free to do so if you prefer.

Also, maybe we can see a way to at least throw an error if the user does this, so they won't get stuck in the loop not knowing what's happening.

garyrussell:
Thanks; I hadn't thought it through but, yes, I guess we could have some kind of marker on the container that can be tested in the customizer.

No rush; that's why I put it on the backlog; we should look to see if we can fail fast, though.

tomazfernandes:
Hi, I think I found a way to make this work. 

Instead of setting the properties directly to the factory, I created a wrapper for the factory that is returned by the `ListenerContainerFactoryConfigurer` instead of the provided instance. The wrapper sets the necessary properties directly to the container in the `createContainer` methods, so nothing is changed in the factory instance itself, and nothing is changed for non retryable topics.

The problem I see with this solution is that the `ListenerContainerFactoryConfigurer` public API (the `configure` and `configureWithoutBackOffValues` methods) relies specifically on the `ConcurrentKafkaListenerContainerFactory` class, instead of the `KafkaListenerContainerFactory` interface, as it should.

So instead of a clean wrapper implementing the interface, I had to subclass the `ConcurrentKafkaListenerContainerFactory ` class, and then it gets messier to guarantee all public method calls will be made to the delegate instead of the wrapper's instance superclass.

Maybe it's not that much of an issue - let me know what you think. I'll try to think of a way around this without breaking the contract. If you have any ideas please let me know.

[...]

PR #2096, opened 2022-02-04
GH-2069: Same factory for retryable and normal endpoints

tomazfernandes:
Resolves https://github.com/spring-projects/spring-kafka/issues/2069

I used @garyrussell's suggestion of creating an isRetryable method in the container, and also allowing composite containerCustomizers.

I wasn't sure whether to add the composite customizer method to the docs, seemed like an unnecessary complexity if we're keeping the regular setContainerCustomizer method. But I can add it if you think it's better.

Let me know if there's anything you'd like me to change.

Thanks!

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2021 the original author or authors.
+ * Copyright 2014-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -117,6 +117,8 @@ public abstract class AbstractKafkaListenerEndpoint<K, V>
 	private boolean splitIterables = true;
 
 	private BatchToRecordAdapter<K, V> batchToRecordAdapter;
+
+	private boolean retryable;
 
 	@Override
 	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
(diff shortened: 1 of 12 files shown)


tomazfernandes:
Hi, I see you're probably busy with the upcoming major release. I'd like to take advantage of the fact that this PR hasn't been yet reviewed and propose that we use #2108 instead. It changes a lot less existing code and should be easier to reason about and review.

Of course, we can reopen this one if you prefer this solution, no worries. And no hurry, if this doesn't get reviewed in time for 2.8.3 I can change the "since" tags and anything else needed.

Thanks!

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
```diff
@@ -341,6 +343,25 @@ public void setRetryTemplate(RetryTemplate retryTemplate) {
 		this.retryTemplate = retryTemplate;
 	}
 
+	/**
+	 * Sets whether this container is part of a non-blocking retries pattern.
```
garyrussell, line 347: Strictly speaking "...whether the container created for this endpoint..."

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
```diff
@@ -341,6 +343,25 @@ public void setRetryTemplate(RetryTemplate retryTemplate) {
...
+	public void setRetryable(boolean retryable) {
+		this.retryable = retryable;
+	}
+
+	/**
+	 * Whether this container is part of a non-blocking retries pattern.
+	 * @return true if it is part of a non-blocking retries pattern.
+	 * @since 2.8.3
+	 */
+	@Override
```
garyrussell, line 360: No need to duplicate the javadoc here.

garyrussell (commented):
LGTM; thanks! Just some minor javadoc polishing.

garyrussell:
Oops - sorry - reviewed the wrong one 😄

PR #2108, opened 2022-02-17
GH-2069: Same factory for retryable and normal endpoints

### LINKED ISSUE #2069: Consider Allowing the Same Container Factory for Retryable and Normal @KafkaListener

tomazfernandes:
Resolves #2069 

This solution uses a decorator to only interfere with container properties when we're in retryable topic logic.

As a result we only change two existing classes, instead of almost 10 from the other PR. I think this should be easier to review and has less chance of interfering with existing logic, while also keeping all retryable topic's factory configuration logic in one class.

I also deprecated two methods in the `ListenerContainerFactoryConfigurer` class, added the possibility of reverting back the behavior, and described this in the documentation. Since we're going to 3.0.0 and I see deprecations are being removed, I can remove these after and if we merge this code to 2.8.x.

Of course, we can still reopen the other PR if you prefer the other solution, no worries. And please let me know if there's anything to be changed.

Thanks!

### DIFF
--- a/spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
@@ -668,8 +668,6 @@ public RetryTopicConfiguration myRetryTopic(KafkaTemplate<Integer, MyPojo> templ
 
 By default the RetryTopic configuration will use the provided factory from the `@KafkaListener` annotation, but you can specify a different one to be used to create the retry topic and dlt listener containers.
 
-IMPORTANT: The provided factory will be configured for the retry topic functionality, so you should not use the same factory for both retrying and non-retrying topics. You can however share the same factory between many retry topic configurations.
-
 For the `@RetryableTopic` annotation you can provide the factory's bean name, and using the `RetryTopicConfiguration` bean you can either provide the bean name or the instance itself.
 
 ====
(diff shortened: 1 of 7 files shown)


spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -703,6 +701,27 @@ public RetryTopicConfiguration myOtherRetryTopic(KafkaTemplate<Integer, MyPojo>
...
+
+IMPORTANT: Since 2.8.3 you can use the same factory for retryable and non-retryable topics.
+
+If you need to revert the factory configuration behavior to prior 2.8.3, you can replace the standard `RetryTopicConfigurer` bean and set `useLegacyFactoryConfigurer` to `true`, such as:
+
+====
+[source, java]
+----
+
+@Bean(name = RetryTopicInternalBeanNames.RETRY_TOPIC_CONFIGURER)
```
tomazfernandes, line 713: A quick note about this: I have a few ideas on how to improve this kind of configuration for the feature and should propose something soon.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
```diff
@@ -95,20 +100,31 @@
 		this.clock = clock;
 	}
 
+	@Deprecated
```
garyrussell, line 103: We should add javadoc to describe why they are deprecated; we usually say `@deprecated in favor of #someOtherMethod`

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurer.java
```diff
@@ -395,6 +404,15 @@ public static EndpointHandlerMethod createHandlerMethodWith(Object bean, Method
 		return new EndpointHandlerMethod(bean, method);
 	}
 
+	/**
+	 * Set to true if you want the {@link ListenerContainerFactoryConfigurer} to behave as before 2.8.3.
+	 * @param useLegacyFactoryConfigurer Whether to use the legacy factory configuration.
+	 */
```
garyrussell, line 410: Add `@deprecated ...`

garyrussell (changes requested):
LGTM; thanks! Just some minor javadoc stuff.

tomazfernandes:
I've added the javadocs as requested, please LMK if there's anything else. Thanks!

garyrussell (approved):

Issue #2109, opened 2022-02-17
@EmbeddedKafka zkSessionTimeout default value


Fr0z3Nn:
Hi, spring-kafka team,

**Expected Behavior**

@EmbeddedKafka zkSessionTimeout should was 18000 by default.

**Current Behavior**

I see that your annotation @EmbeddedKafka has zkSessionTimeout 6000.

**Context**

I think change this default value will protect from unexpected fall. 
You can check it changes in Kafka Team project
https://cwiki.apache.org/confluence/display/KAFKA/KIP-537%3A+Increase+default+zookeeper+session+timeout
I got unexpected fall in my cloud integration test with 1% chance caused by
org.apache.zookeeper.ClientCnxn$SessionTimeoutException and after changes zkSessionTimeout to 18000 it worked properly

garyrussell:
I already patched it 😄

PR #2110, opened 2022-02-17
GH-2109: Increase EKB Zookeeper Timeouts

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2109

Increase default session timeout to 18 seconds to align with Kafka itself.
Also make connection timeout the same, similar to Kafka's default behavior.


artembilan:
... and cherry-picked to `2.8.x` & `2.7.x`

PR #2115, opened 2022-02-18
Use ArrayList to save time and memory in large size

xixingya:
Hello, becaue we know the size of list,so use ArrayList may save memory and time

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -2063,10 +2063,11 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 		}
 
 		private List<ConsumerRecord<K, V>> createRecordList(final ConsumerRecords<K, V> records) {
-			Iterator<ConsumerRecord<K, V>> iterator = records.iterator();
-			List<ConsumerRecord<K, V>> list = new LinkedList<>();
-			while (iterator.hasNext()) {
-				list.add(iterator.next());
+			//use ArrayList because we know it's size
+			List<ConsumerRecord<K, V>> list = new ArrayList<>(records.count());
+			Set<TopicPartition> partitions = records.partitions();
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2063,10 +2064,11 @@ private void batchAfterRollback(final ConsumerRecords<K, V> records,
 		}
 
 		private List<ConsumerRecord<K, V>> createRecordList(final ConsumerRecords<K, V> records) {
-			Iterator<ConsumerRecord<K, V>> iterator = records.iterator();
-			List<ConsumerRecord<K, V>> list = new LinkedList<>();
-			while (iterator.hasNext()) {
-				list.add(iterator.next());
+			//use ArrayList because we know it's size
+			List<ConsumerRecord<K, V>> list = new ArrayList<>(records.count());
```
artembilan, line 2068: I think there was a reason to use exactly `LinkedList` and populate its content exactly in the same order as values are present in the `ConsumerRecords`.
What you do with the `ArrayList` and then `records.records()` does not guarantee an order how entries were populated into the `ConsumerRecords`.

WDYT?

artembilan: Sorry, I'm not fully sure that I follow. 
The `keySet()` returns just keys. And it does that over here this way:
```
public Set<TopicPartition> partitions() {
        return Collections.unmodifiableSet(records.keySet());
    }
```

Then we do iteration over these keys and call `records(TopicPartition partition)` which returns:
```
return Collections.unmodifiableList(recs);
```

So, I'm not sure what memory are you talking about, but it feel like the `records.iterator()` (see their `ConcatenatedIterable`) with existing `LinkedList` is much more performant than jumping around extra wrapping objects.

Any reports confirming that your solution is better and it is really stable with an order preserving?

Thanks

artembilan (changes requested):

garyrussell:
We disagree with this change.

Issue #2116, opened 2022-02-18
Combine blocking and non-blocking error handling


mariecuriiee:
As before discussed [here on Github](https://github.com/spring-projects/spring-kafka/discussions/2101).

**Expected Behavior**

A combination of blocking and non-blocking retry should be possible to handle three different types of exceptions:

- non-blocking (but not fatal) exceptions should be retried.

- non-blocking (but fatal, e.g. DeserializationException) exceptions should be routed to the DLT directly.

- blocking exceptions (e.g. a database is currently not available) should not be routed to a retry topic since all following messages would be routed there as well. The consumer should remain at the current offset.

**Current Behavior**
It is possible to use @RetryableTopic annotation to satisfy the first two bullets. It is also possible to use DefaultErrorHandler to implement the last bullet. But a combination of both approaches cannot be configured at the moment.

**Context**
@garyrussell mentioned a workaround by stopping the container using the `stopImmediate` container property. Nevertheless, there are a few blog articles that address the different handling of error types, e.g., this [article](https://blog.devgenius.io/exceptions-and-retry-policy-in-kafka-6f0d3cf2c330). Therefore, having a proper way to configure this behavior would be a useful possibility.

tomazfernandes:
Thanks for the suggestion @mariecuriiee, that's a good article.

I'm currently working on a couple of issues that should pave the way for this feature to work, by leveraging the existing DefaultErrorHandler we already set for the RetryableTopics. So hopefully we'll have this implemented soon.

tomazfernandes:
Hi @mariecuriiee, I've added a PR with this feature, maybe you want to take a look:
https://github.com/spring-projects/spring-kafka/pull/2124

LMWYT, thanks

mariecuriiee:
Hey @tomazfernandes, thanks for implementing this so quickly! If I understand the implementation correctly, I only have to set the backoff using `setBlockingRetryBackoff` and add a classification behavior using `setErrorHandlerCustomizer`? And with `setErrorHandlerCustomizer` I could then define the behavior depending on the exception, e.g., for some exceptions the records go directly to the retry topics, for other exceptions, the records are handled using blocking retries or for other exceptions, the records are routed directly to the DLT?

tomazfernandes:
Hi @mariecuriiee, thanks for looking into the solution. We’re currently reviewing it to improve the classification for blocking retries and make it allowlistable instead of the current denylist only behavior.

[...]

PR #2124, opened 2022-02-20
GH-2116: Add blocking retries to RT

tomazfernandes:
Before we hardcoded a no-ops back off in the DefaultErrorHandler used in the Retryable Topics feature.
Adds a setter to let the user provide their own back off policy and configure blocking retries in conjunction with RT.

I also thought of making this configuration possible on a per-topic basis by adding it to the builder and annotation, but I think this is enough for this release since we won't have much time to review. We can open an issue for that if that's the case
and I can provide a PR soon.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ExceptionClassifier.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ExceptionClassifier.java
@@ -181,7 +181,8 @@ public abstract class ExceptionClassifier extends KafkaExceptionLogLevelAware {
 	 * </ul>
 	 * All others will be retried, unless {@link #defaultFalse()} has been called.
 	 * @param exceptionType the exception type.
-	 * @return true if the removal was successful.
+	 * @return the classification of the exception if removal was successful;
+	 * null otherwise.
 	 * @since 2.8.4
 	 * @see #addNotRetryableExceptions(Class...)
 	 * @see #setClassifications(Map, boolean)
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ExceptionClassifier.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ExceptionClassifier.java
@@ -22,6 +22,7 @@ import java.util.Map;
 import org.springframework.classify.BinaryExceptionClassifier;
 import org.springframework.kafka.support.converter.ConversionException;
 import org.springframework.kafka.support.serializer.DeserializationException;
+import org.springframework.lang.Nullable;
 import org.springframework.messaging.converter.MessageConversionException;
 import org.springframework.messaging.handler.invocation.MethodArgumentResolutionException;
 import org.springframework.util.Assert;
@@ -185,7 +186,8 @@ public abstract class ExceptionClassifier extends KafkaExceptionLogLevelAware {
 	 * @see #addNotRetryableExceptions(Class...)
 	 * @see #setClassifications(Map, boolean)
 	 */
-	public boolean removeClassification(Class<? extends Exception> exceptionType) {
+	@Nullable
+	public Boolean removeClassification(Class<? extends Exception> exceptionType) {
 		return this.classifier.getClassified().remove(exceptionType);
 	}
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
@@ -17,6 +17,7 @@
 package org.springframework.kafka.retrytopic;
 
 import java.time.Clock;
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
@@ -173,6 +174,10 @@ public class ListenerContainerFactoryConfigurer {
 	 */
 	public void setBlockingRetriesBackOff(BackOff blockingBackOff) {
 		Assert.notNull(blockingBackOff, "The provided BackOff cannot be null");
+		Assert.state(this.providedBlockingBackOff == null, () ->
+				"Blocking retries back off has already been set. Current: "
+						+ this.providedBlockingBackOff
+						+ " You provided: " + blockingBackOff);
 		this.providedBlockingBackOff = blockingBackOff;
 	}
 
@@ -187,6 +192,10 @@ public class ListenerContainerFactoryConfigurer {
 	public final void setBlockingRetryableExceptions(Class<? extends Exception>... exceptionTypes) {
 		Assert.notNull(exceptionTypes, "The exception types cannot be null");
 		Assert.noNullElements(exceptionTypes, "The exception types cannot have null elements");
+		Assert.state(this.blockingExceptionTypes == null,
+				() -> "Blocking retryable exceptions have already been set."
+						+  "Current ones: " + Arrays.toString(this.blockingExceptionTypes)
+						+ " You provided: " + Arrays.toString(exceptionTypes));
 		this.blockingExceptionTypes = exceptionTypes;
 	}
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
@@ -83,6 +83,8 @@ public class ListenerContainerFactoryConfigurer {
 
 	private BackOff providedBlockingBackOff = null;
 
+	private Class<? extends Exception>[] blockingExceptionTypes = null;
+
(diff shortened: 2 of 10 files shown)


garyrussell (changes requested):
I don't think I will merge this for 2.8.3; need to think it through some more. As I understand the logic, it will always go through blocking retries first and then through non-blocking (assuming default config). I think we need the reverse logic - i.e. don't retry (blocking) for any exceptions unless explicitly set by the user; I think that will need changes in the DLPR.

tomazfernandes:
> I don't think I will merge this for 2.8.3; need to think it through some more. As I understand the logic, it will always go through blocking retries first and then through non-blocking (assuming default config). I think we need the reverse logic - i.e. don't retry (blocking) for any exceptions unless explicitly set by the user; I think that will need changes in the DLPR.

Yes, I thought about this too. I think we could solve this by allowing the ExceptionClassifier to have retryOn (i.e. allowListable) and setting the default behavior to "false" for RT, if that's acceptable for the other subclasses.

tomazfernandes:
We'd then have something like:

```
	ListenerContainerFactoryConfigurer lcfc = new ListenerContainerFactoryConfigurer(kafkaConsumerBackoffManager, deadLetterPublishingRecovererFactory, clock);
	lcfc.setBlockingRetriesBackOff(new FixedBackOff(50, 3));
	lcfc.setErrorHandlerCustomizer(commonErrorHandler -> ((DefaultErrorHandler) commonErrorHandler)
			.addRetryableException(ShouldRetryViaBlockingException.class, ShouldRetryViaBothException.class));
	return lcfc;
}
```

tomazfernandes:
And just to be clear, with this PR as is the default behavior would stay the same, with the no ops BackOff policy. Only if the user provides their own BackOff policy the blocking retries would kick in.

garyrussell:
Right; I get that; I just have a problem with the default behavior if you just add a back off.

Thanks for all your work on these issues, by the way.

tomazfernandes:
Sure, no worries, I'm glad we were able to go through them in time for this release, thanks.

Let me know if there's anything else I can help with for this issue. We can close this PR and open another one when we have a clearer picture if you prefer.

If nothing more pressing comes up I should start working on improving the way RT's components are configured in general, if that's ok. So hopefully users will have a nice and clean API to use these features instead of all current non-intuitive verbosity.  I'll open an issue when I have a draft so we can all take a look into it before actual implementation starts.

garyrussell:
I am ok with this PR, I think we just need a modification to `ExceptionClassifier` so it can be configured with a classifier that doesn't retry by default as you suggested https://github.com/spring-projects/spring-kafka/pull/2124#issuecomment-1047062358

garyrussell:
Here's a proposal to support classifying exceptions as false by default...

https://github.com/garyrussell/spring-kafka/commit/a6bf67b15c8a7dbabd4d8ee1a4235434d9f689f2

garyrussell:
Correction: https://github.com/garyrussell/spring-kafka/commit/63c5a2f5dee2524963afd39a3d59ec026bc9a7dc

tomazfernandes:
With that improvement in the ExceptionClassifier I can resume working on this, thanks. Should have something later today for your appreciation.

tomazfernandes:
Now that we have the DHE set to default false, we no longer need the No Ops back off for the retryable topics - the user can just add Retryable exceptions and use the DHE default one.

Since the logic for when the back off policy has no retries and the logic for when the classification returns false are  a bit different, I had to make some minor adjustments to other classes so that the DLPR partition checking behavior stays the same.

As far as I could check this adjustment should have no impact besides enabling partition checking for users that use DLPR with exception classification that returns false, which I think should be a good thing for consistency and easy to disable if needed since there's a flag for that.

Of course, if that's a problem I can revert to using the NoOps back off.

Thanks

tomazfernandes:
I'll just add a small improvement to the API and the docs.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
```diff
@@ -158,6 +160,20 @@ public ListenerContainerFactoryConfigurer(KafkaConsumerBackoffManager kafkaConsu
 		return new RetryTopicListenerContainerFactoryDecorator(factory, configuration, false);
 	}
 
+	/**
+	 * Set a {@link BackOff} to be used with blocking retries.
+	 * You can specify the exceptions to be retried using the method
+	 * {@link org.springframework.kafka.listener.ExceptionClassifier#addRetryableExceptions(Class[])}
+	 * By default, no exceptions are retried via blocking.
```
garyrussell, line 167: We should add something like "If the backoff execution returns STOP (retries are exhausted), the record will then go to the next retry topic (if present)."

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
```diff
@@ -193,14 +209,20 @@ public void setErrorHandlerCustomizer(Consumer<CommonErrorHandler> errorHandlerC
...
 		errorHandler.setCommitRecovered(true);
 		errorHandler.setLogLevel(KafkaException.Level.DEBUG);
 		this.errorHandlerCustomizer.accept(errorHandler);
 		return errorHandler;
 	}
 
+	protected DefaultErrorHandler createDefaultErrorHandlerInstance(DeadLetterPublishingRecoverer deadLetterPublishingRecoverer) {
+		return this.providedBlockingBackOff != null
+				? new DefaultErrorHandler(deadLetterPublishingRecoverer, this.providedBlockingBackOff)
+				: new DefaultErrorHandler(deadLetterPublishingRecoverer);
```
garyrussell, line 223: This isn't right; the default fixed back off is (0, 9), we need (0, 0).

Why not just set the field to `new FixedBackOff(0, 0);`? We then don't need this test. Rename the field too.

garyrussell (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
tomazfernandes: Well, considering all exceptions will return false when classified by default, it doesn't really matter what's the backoff policy - we can just leave the default one and then the user would have only to add the retryable exceptions, which would then make the default or provided back off kick in.

Anyway, we can set the no ops back off too, no worries. Either way I think the other classes modifications to pass the consumer to DLPR are still needed, since we might go that route for exception classifications that return false.

garyrussell: Oh; Ok; makes sense to leave it.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
tomazfernandes: Yes, I think the aggregated behavior of blocking and non-blocking can be better documented in general. I'll see what I can come up with.

tomazfernandes:
Improved the documentation and made a slight improvement to the API.

spring-kafka/src/main/java/org/springframework/kafka/listener/ExceptionClassifier.java
```diff
@@ -185,7 +186,8 @@ public boolean removeNotRetryableException(Class<? extends Exception> exceptionT
 	 * @see #addNotRetryableExceptions(Class...)
 	 * @see #setClassifications(Map, boolean)
 	 */
-	public boolean removeClassification(Class<? extends Exception> exceptionType) {
+	@Nullable
+	public Boolean removeClassification(Class<? extends Exception> exceptionType) {
```
garyrussell, line 190: Why do we need this change?

garyrussell (commented):
Great work! Thanks. Just one question.

spring-kafka/src/main/java/org/springframework/kafka/listener/ExceptionClassifier.java
garyrussell: Oops; thanks!

garyrussell: We should change the javadoc to `@return the classification of the exception if the removal was successful; null otherwise.`

I can do it during the merge if you like; but I have to run out for a short while.

tomazfernandes: I can do that, no worries, thanks!

tomazfernandes: Changed the javadoc and added two assertions to the LCFC new methods

garyrussell (approved):

garyrussell:
...and cherry-picked as a7621a101ecf8382224358323878d5fde21cb8bf

Issue #2128, opened 2022-02-23
Calling `.nack()` suspends consumer group rebalance


tkornai:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

I tested the following versions, they all show the same behaviour:

2.3.13, 2.5.14, 2.6.3, 2.8.2

**Describe the bug**

After calling `KafkaMessageListenerContainer.nack()` if a consumer group rebalance is triggered then it won't finish until the `nack()` timeout has passed.

The extended CG rebalance prevents every member of the consumer group from progressing with message processing.

**To Reproduce**

Please follow the steps described in the sample repository: https://github.com/tkornai/nack-rebalance-demo

**Expected behavior**

Calling `nack()` must not delay subsequent consumer group rebalances.

**Sample**

A link to a GitHub repository with a [minimal, reproducible,  sample](https://github.com/tkornai/nack-rebalance-demo).

garyrussell:
Interesting problem; it's going to be tricky to solve it because we don't find out about the rebalance until we next poll the consumer.

I think we'll need to change the architecture to pause the consumer for the nack sleep time and continue to poll it; however, then, the accuracy of the sleep time will be affected by the poll timeout.

garyrussell:
Many thanks for the repro; big help!

Interesting problem; it's going to be tricky to solve it because we don't find out about the rebalance until we next poll the consumer.

I think we'll need to change the architecture to pause the consumer for the nack sleep time and continue to poll it; however, then, the accuracy of the sleep time will be affected by the poll timeout.

jffourmond:
Hi, following this fix, nack ignores the sleep time parameter. Messages are re-consumed 5 seconds later, no matter what the sleep time value is.

garyrussell:
The default poll timeout is 5 seconds, so that's a smoking gun.

I'll take a look, but the logic looks correct to me

When receive the nack, we set `nackWake`.

`this.nackWake = System.currentTimeMillis() + this.nackSleep;`

and resume only when

`if (System.currentTimeMillis() > this.nackWake) {`

garyrussell:
Works as expected for me:

```java
@SpringBootApplication
public class Kgh2128Application {

	private static final Logger log = LoggerFactory.getLogger(Kgh2128Application.class);

	public static void main(String[] args) {
		SpringApplication.run(Kgh2128Application.class, args);
	}
	@KafkaListener(id = "kgh2128", topics = "kgh2128")
[...]

PR #2131, opened 2022-02-23
GH-2128 Do Not Sleep Consumer Thread for Nack

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2128

Suspending polling delays rebalancing; instead pause the consumer and
continue polling. Check if partitions are already paused and only pause
the current active partitions and resume them after the sleep interval
has passed.

Re-pause as necessary after a rebalance.

**cherry-pick to 2.8.x**


artembilan (commented):
Just for clarification. 
Do we still need that `"nack() can only be called on the consumer thread"` assertion since we don't sleep any more?

Although it is not clear why would we need it before anyway...

Thanks

garyrussell:
Yes, it's still a limitation; we can't handle async processing in this mode because we ack all records before the failure and re-seek those after.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2458,17 +2472,29 @@ private void handleNack(final ConsumerRecords<K, V> records, final ConsumerRecor
...
+				try {
+					this.consumer.pause(this.pausedForNack);
+				}
+				catch (IllegalStateException ex) {
+					// this should never happen; defensive, just in case...
+					this.logger.warn(() -> "Could not pause for nack, possible rebalance in process: "
+							+ ex.getMessage());
+					Set<TopicPartition> nowPaused = new HashSet<>(this.consumer.paused());
+					nowPaused.removeAll(alreadyPaused);
+					this.consumer.resume(nowPaused);
```
tkornai, line 2495: Can we expect `consumer.resume()` to succeed if `consumer.pause()` threw an IllegalStateException?

garyrussell: `pause()` iterates over the collection and throws an exception if the partition is no longer assigned; therefore some partitions may have been paused; all this code is doing is resuming those that were successful. As the comment states, this code should never be executed. I added it before I removed the paused (for nack) partitions in `onPartitionsRevoked()`, but decided to leave it here (with the comment).

tkornai: I see, thanks for the blazing fast reaction on the bug report and the detailed explanation above!

artembilan:
... and cherry-picked to `2.8.x`

Issue #2132, opened 2022-02-23
Add Utility to Log ProducerRecord Metadata Only


garyrussell:
e.g.

https://github.com/spring-projects/spring-kafka/blob/38ed222a305e3b5f363ac51393da2d1c046fabd0/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java#L381

https://github.com/spring-projects/spring-kafka/blob/38ed222a305e3b5f363ac51393da2d1c046fabd0/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java#L399

Need an option to just log the metadata.

https://stackoverflow.com/questions/71244102/spring-kafka-replyingkafkatemplate-logs-full-details-instead-of-metadata-only-wh

garyrussell:
Closed by https://github.com/spring-projects/spring-kafka/commit/891664b63a647a4956f5be94cb39ebf79a6c0055 and back-ported to 2.8.x, 2.7.x

artembilan:
Closed according the previous comment.

PR #2133, opened 2022-02-23
GH-2132: Producer Record Log Formatting

garyrussell:
Partial solution for  https://github.com/spring-projects/spring-kafka/issues/2132

Allow the user to provide a function for formatting producer records.

Also move consumer record formatting to `KafkaUtils` will deprecate
`ListenerUtils` versions later.

There will be more polising in this area for 3.0 only (separate PR).


artembilan:
... and cherry-picked to `2.8.x` & `2.7.x`.

Issue #2148, opened 2022-03-04
Monitoring Listener Performance with multiple MeterRegistry beans


garyrussell:
### Discussed in https://github.com/spring-projects/spring-kafka/discussions/2105

<div type='discussions-op-text'>

<sup>Originally posted by **alexanderad** February 11, 2022</sup>
Hi!

Recently I've been looking for a way to monitor listener specific metrics and found a corresponding section in the docs ([Monitoring Listener Performance](https://docs.spring.io/spring-kafka/reference/html/#monitoring-listener-performance)) saying
> ... the listener container will automatically create and update Micrometer `Timer`s for the listener, **if Micrometer is detected** on the class path, **and a single MeterRegistry is present** in the application context ...

This way of collecting listener metrics is particulary useful given that it is able to account for multiple types of listeners (e.g. declared with `@KafkaListener`, or `@KafkaHandler`, or being batch listeners, or hidden behind custom factories for consumers, etc); as well as aware of concurrent listeners, and templates.

The corresponding implementation to be found in: https://github.com/spring-projects/spring-kafka/blob/7c2178ac4364989b84b9e644092d7dfb60882887/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/MicrometerHolder.java#L61-L79

The problem I faced is that in case when multiple registries are auto-configured for the application context, e.g. Prometheus + Graphite timers for metrics do not register themselves.This applies to any combination of monitoring systems providing more that one `MeterRegistry.class` bean visible via `context.getBeansOfType(MeterRegistry.class, ...)`. `KafkaMessageListenerContainer` silently rejects to collect these metrics, and it seems there is no approachable way to configure alternative behavior, e.g.:
* prefer a specific registry (say, provided via `containerProperties`)
* combine discovered registries into [`CompositeMeterRegistry`](https://www.javadoc.io/doc/io.micrometer/micrometer-core/1.3.1/io/micrometer/core/instrument/composite/CompositeMeterRegistry.html)

I'm looking for an advice here. Is there anything obvious I'm missing? Is the only one way reasonable in here would be to patch `spring-kafka` itself to achieve metrics collection for the case of multiple metric registries in use?

Thanks!</div>

PR #2151, opened 2022-03-07
GH-2148: Fix MeterRegistry Detection

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2148

- use Boot's `AutoConfiguredCompositeMeterRegistry`, if only one
  (tested with a Boot application)
- use `@Primary` registry bean

**cherry-pick to 2.8.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/MicrometerHolder.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/MicrometerHolder.java
@@ -20,7 +20,6 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionRegistry;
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/MicrometerHolder.java
```diff
@@ -69,12 +75,56 @@ public MicrometerHolder(@Nullable ApplicationContext context, String name,
...
+		if (registries.size() == 1) {
+			return registries;
+		}
+		Map<String, MeterRegistry> filtered = registries.entrySet()
+				.stream()
+				.filter(entry -> entry.getValue()
+						.getClass()
+						.getName()
+						.equals("org.springframework.boot.actuate.autoconfigure.metrics."
+								+ "AutoConfiguredCompositeMeterRegistry"))
```
artembilan, line 101: Why do we need this if it is `@Primary` in the auto-configuration:
```
	@Bean
	@Primary
	AutoConfiguredCompositeMeterRegistry compositeMeterRegistry(Clock clock, List<MeterRegistry> registries) {
		return new AutoConfiguredCompositeMeterRegistry(clock, registries);
	}
```
?

I mean it looks like the next logic covers this Spring Boot "pollution" as well.

Do I miss anything ?

Thanks

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/MicrometerHolder.java
garyrussell: Oh; thanks; I didn't look at that; I just saw 3 beans in my test app; will test and push.

PR #2152, opened 2022-03-07
GH-2132: Deprecate ListenerUtils.recordToString

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/pull/2133

Apply to 2.8.x, and 2.7.x.

In 3.0 they will be removed and `KafkaUtils` called directly instead.


artembilan:
This cannot be back-ported clearly.
Do, we really need to worry about those deprecations in version below `2.8.x`?

garyrussell:
Agreed; but then we need to revert part of the previous cherry-pick (just the consumer record parts).

I will do a new PR in the morning.

PR #2156, opened 2022-03-08
GH-2149: Expose API for providing BackOffPolicies

### LINKED ISSUE #2149: Expose public API for providing BackOffPolicies
As discussed [here](https://github.com/spring-projects/spring-boot/pull/29812#discussion_r819646339), we'd rather expose the `BackOff` generation logic we use for the `RetryableTopics` feature as an API  than duplicate it in Spring Boot for auto configuration.

I should have a PR for this soon if that's ok.

Thanks.

tomazfernandes:
Resolves #2149.

I chose not to change anything in the logic for creating the BackOffPolicy to make this PR cleaner, even though there are a few NPEs that need addressing there. I should submit another PR to fix that after and if this one is merged. But I can also make a new commit with the fixes to this PR if that's better.

I also chose to expose the API through the `RetryTopicConfigurationBuilder` because it allows users to externalize back off parameters, while it's also used by `RetryableTopicAnnotationProcessor` and can be used in Boot's auto configuration, so it becomes the only point of access to that logic.

Please LMK if there's anything to change, or any suggestions regarding these points.

Thanks

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2021 the original author or authors.
+ * Copyright 2018-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -39,11 +39,6 @@ import org.springframework.kafka.retrytopic.RetryTopicConstants;
 import org.springframework.kafka.retrytopic.RetryTopicInternalBeanNames;
 import org.springframework.kafka.support.EndpointHandlerMethod;
 import org.springframework.retry.annotation.Backoff;
-import org.springframework.retry.backoff.ExponentialBackOffPolicy;
-import org.springframework.retry.backoff.ExponentialRandomBackOffPolicy;
-import org.springframework.retry.backoff.FixedBackOffPolicy;
-import org.springframework.retry.backoff.SleepingBackOffPolicy;
-import org.springframework.retry.backoff.UniformRandomBackOffPolicy;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.ReflectionUtils;
@@ -129,9 +124,10 @@ public class RetryableTopicAnnotationProcessor {
 		if (StringUtils.hasText(annotation.autoStartDltHandler())) {
 			autoStartDlt = resolveExpressionAsBoolean(annotation.autoStartDltHandler(), "autoStartDltContainer");
 		}
+		BackOffValues boValues = createBackoffValues(annotation.backoff(), this.beanFactory);
 		return RetryTopicConfigurationBuilder.newInstance()
 				.maxAttempts(resolveExpressionAsInteger(annotation.attempts(), "attempts", true))
-				.customBackoff(createBackoffFromAnnotation(annotation.backoff(), this.beanFactory))
+				.backOffFor(boValues.min, boValues.max, boValues.multiplier, boValues.random)
 				.retryTopicSuffix(resolveExpressionAsString(annotation.retryTopicSuffix(), "retryTopicSuffix"))
 				.dltSuffix(resolveExpressionAsString(annotation.dltTopicSuffix(), "dltTopicSuffix"))
 				.dltHandlerMethod(getDltProcessor(method, bean))
@@ -151,11 +147,10 @@ public class RetryableTopicAnnotationProcessor {
 				.create(getKafkaTemplate(annotation.kafkaTemplate(), topics));
 	}
 
-	private SleepingBackOffPolicy<?> createBackoffFromAnnotation(Backoff backoff, BeanFactory beanFactory) { // NOSONAR
+	private BackOffValues createBackoffValues(Backoff backoff, BeanFactory beanFactory) { // NOSONAR
 		StandardEvaluationContext evaluationContext = new StandardEvaluationContext();
 		evaluationContext.setBeanResolver(new BeanFactoryResolver(beanFactory));
 
-		// Code from Spring Retry
 		Long min = backoff.delay() == 0 ? backoff.value() : backoff.delay();
 		if (StringUtils.hasText(backoff.delayExpression())) {
 			min = resolveExpressionAsLong(backoff.delayExpression(), "delayExpression", true);
@@ -168,27 +163,9 @@ public class RetryableTopicAnnotationProcessor {
 		if (StringUtils.hasText(backoff.multiplierExpression())) {
 			multiplier = resolveExpressionAsDouble(backoff.multiplierExpression(), "multiplierExpression", true);
 		}
-		if (multiplier != null && multiplier > 0) {
-			ExponentialBackOffPolicy policy = new ExponentialBackOffPolicy();
-			if (backoff.random()) {
-				policy = new ExponentialRandomBackOffPolicy();
-			}
-			policy.setInitialInterval(min);
-			policy.setMultiplier(multiplier);
-			policy.setMaxInterval(max > min ? max : ExponentialBackOffPolicy.DEFAULT_MAX_INTERVAL);
-			return policy;
-		}
-		if (max != null && min != null && max > min) {
-			UniformRandomBackOffPolicy policy = new UniformRandomBackOffPolicy();
-			policy.setMinBackOffPeriod(min);
-			policy.setMaxBackOffPeriod(max);
-			return policy;
-		}
-		FixedBackOffPolicy policy = new FixedBackOffPolicy();
-		if (min != null) {
-			policy.setBackOffPeriod(min);
-		}
-		return policy;
+		Boolean random = backoff.random();
+
+		return new BackOffValues(min, max, multiplier, random);
 	}
 
 	private EndpointHandlerMethod getDltProcessor(Method listenerMethod, Object bean) {
@@ -384,6 +361,20 @@ public class RetryableTopicAnnotationProcessor {
 			return ((ConfigurableBeanFactory) this.beanFactory).resolveEmbeddedValue(value);
 		}
 		return value;
+	}
+
+	private static final class BackOffValues {
+		final Long min;
+		final Long max;
+		final Double multiplier;
+		final Boolean random;
+
+		private BackOffValues(Long min, Long max, Double multiplier, Boolean random) {
+			this.min = min;
+			this.max = max;
+			this.multiplier = multiplier;
+			this.random = random;
+		}
 	}
 
 }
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
```diff
@@ -386,4 +363,18 @@ private String resolve(String value) {
 		return value;
 	}
 
+	private static final class BackOffValues {
```
artembilan, line 366: This looks like Java `record`: https://blogs.oracle.com/javamagazine/post/records-come-to-java.

Although I wonder why do we need this intermediate class at all?
Why plain `HashMap` doesn't work?
Because of type?
But this is an internal logic, so simple casting won't hurt and won't affect anything else.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/BackOffPolicyProvider.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.retry.backoff.UniformRandomBackOffPolicy;
+
+/**
+ * Provides a {@link BackOffPolicy} from the given parameters.
+ * Original logic from Spring Retry project.
+ *
+ * @author Tomaz Fernandes
+ * @since 2.8.4
+ */
+class BackOffPolicyProvider {
```
artembilan, line 32: Why do we need this extra class, when there is only one method which could migrate to that `RetryTopicConfigurationBuilder` ?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
```diff
@@ -242,6 +252,21 @@ public RetryTopicConfigurationBuilder customBackoff(SleepingBackOffPolicy<?> bac
...
 
+	/**
+	 * Sets a {@link BackOffPolicy} based on the given values.
+	 * @param min        the minimum delay value
+	 * @param max        the maximum delay value
+	 * @param multiplier the multiplier to be applied in the exponential backoff policy
+	 * @param isRandom   whether the exponential back off is random
+	 * @return the builder instance
+	 */
+	public RetryTopicConfigurationBuilder backOffFor(Long min, Long max, Double multiplier, Boolean isRandom) {
```
artembilan, line 263: This is new public API, so `@since 3.0`

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/BackOffPolicyProviderTest.java
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.retry.backoff.ExponentialBackOffPolicy;
+import org.springframework.retry.backoff.ExponentialRandomBackOffPolicy;
+import org.springframework.retry.backoff.FixedBackOffPolicy;
+import org.springframework.retry.backoff.UniformRandomBackOffPolicy;
+
+/**
+ * @author Tomaz Fernandes
+ * @since 2.8.4
+ */
+class BackOffPolicyProviderTest {
```
artembilan, line 33: We prefer to name test classes as `*Tests`, but I guess this one could go away and its tests can migrate to the `RetryTopicConfigurationBuilderTests`. From there you can take that `backOffPolicy` from the builder using `DirectFieldAccessor`.

artembilan (changes requested):
Just some my nit-picks.
If @garyrussell is good with the PR, I won't mind to take my suggestions back.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
tomazfernandes: Hi @artembilan, thanks for looking into this. I've been meaning to ask about java 17 vs java 8 - but I guess the real question is: what's the criteria for a feature to be cherry-picked to 2.8.x, or only 3.0.0? I've been sticking with Java 8 since 3.0.0 seems to be a while away.

As for `HashMap`, correct me if I'm wrong, but I think it's more a matter of personal preference, or maybe framework standards. I for one tend to think that having a small inner class, or record, is easier to reason about than a Map, since the compiler helps us in not messing anything up, and the properties can have nice descriptive names. Are there any other tradeoffs I might be missing? Of course, I can change it to a `Map` if that's better, no worries.

I also thought of having this values class as an inner class of `BackOffProvider` for client classes to use, but then the class would have to be public, which seems like an unnecessary exposure since it's not meant for end users.

artembilan: OK. You know it sounds more like a `BackOffPolicyBuilder` has to be there.
Perhaps for now it could be implemented here in `spring-kafka`, but more logical to have it in the `spring-retry` in that `org.springframework.retry.backoff` package.
I see there is some API on the matter in the `RetryTemplateBuilder`:
```
public RetryTemplateBuilder exponentialBackoff(long initialInterval, double multiplier, long maxInterval,
			boolean withRandom) {

public RetryTemplateBuilder fixedBackoff(long interval) {

public RetryTemplateBuilder uniformRandomBackoff(long minInterval, long maxInterval) {

public RetryTemplateBuilder noBackoff() {

public RetryTemplateBuilder customBackoff(BackOffPolicy backOffPolicy) {
```

Doesn't that look we are duplicating over here in SK?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
tomazfernandes: Sure, maybe we can leave it here for now, open a PR for Spring Retry for having a `BackOffPolicyBuilder`, and when we have the API exposed we can use it and remove from here. We don't use the actual RetryTemplate for the Retry Topics feature, only the `BackOffPolicy` and `@BackOff` annotation. You can see how the policy is used in the [BackOffValuesGenerator](https://github.com/spring-projects/spring-kafka/blob/bee97d9c8fcc1f9489b4244a653ef19b4f8abc83/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/BackOffValuesGenerator.java) class.

Also, instead of having the inner class, we could pass the builder as an argument to the `createBackoffValues` method and just set it there, but then we'd need to break the fluent API to get a reference to the builder - which doesn't look like a problem in an internal code, as you mentioned.

When this feature was developed the Spring Retry project didn't seem to be getting many releases, so it didn't seem really practical to wait for it instead of simply duplicating the logic, since it's not that much code. I see the Spring team made a task force to make a new release of the project, thanks and congrats for that.

artembilan: Well, you are right: we are committed to have a better support for Spring Retry.
So, if you would implement a `BackOffPolicyBuilder`, then it is better, please, to PR it against `spring-retry`.
Then the current `RetryTopicConfigurationBuilder` would not have any changes.
And then we would benefit in Spring Boot as well.
We won't have that `BackOffValues` and so on.

Does it make sense?

garyrussell: I haven't looked at this yet, but if it's not too complex, I don't mind putting it into 2.8.x because 3.0.0 is many months away.

However, if we go ahead and do the "retry topics all in the main container" enhancement, I would start a 2.9 soon (for Boot 2.7) and we can stop back-porting new features like this.

tomazfernandes: > Well, you are right: we are committed to have a better support for Spring Retry. So, if you would implement a `BackOffPolicyBuilder`, then it is better, please, to PR it against `spring-retry`. Then the current `RetryTopicConfigurationBuilder` would not have any changes. And then we would benefit in Spring Boot as well. We won't have that `BackOffValues` and so on.
> 
> Does it make sense?

Sure. Let me just make sure we're on the same page. There are currently two ways of getting a `BackOffPolicy`, both in Spring Kafka and Spring Retry.

The first is through a `Builder` where we explicitly say the policy we want - as in the methods you mentioned. This part doesn't seem like something we'd change - it's already part of the public APIs of both projects.

The second is inferring the policy through the values we pass, which is used in the `@BackOff` annotation processing, originally in the [AnnotationAwareRetryOperationsInterceptor](https://github.com/spring-projects/spring-retry/blob/302f7999932421d39800b81b5fa9f6b1c557b957/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java#L404) class in the `Spring Retry` project, and duplicated in the [RetryableTopicAnnotationProcessor.java](https://github.com/spring-projects/spring-kafka/blob/43cb78127b3851c03037418533bf393c2fc7db3a/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java#L171) class, which is the  logic I extracted for the `BackOffProvider` class for this PR, and the one we need for Boot's auto configuration. This sounds like the one we should open up and deduplicate.

So my question is, are you proposing a `BackOffPolicyBuilder` that has both approaches, e.g., a class with the `Builder` methods you mentioned, and also a generic `long, long, double, boolean` method for inferring the policy?

artembilan: I think we just expose simple props and have a complex logic in the `build()` to determine what impl to return.

I guess we can start from a simple builder approach to satisfy our current requirements.
It could evolve eventually.

tomazfernandes: Oh, now I see. You mean having something like:

```
BackOffPolicyBuilder
    .withValues()
    .minDelay(1000L)
    .maxDelay(5000L)
    .multiplier(3.0)
    .random()
    .build()
```

That sure looks much better than what I suggested, thanks. If that's about what you had in mind, maybe we can close this PR and I'll open a new one for Spring Retry soon.

artembilan: Something like that, yes.
But probably without `.withValues()` and `random()` with param. We got that `true/false` from the annotation or Spring Boot configuration properties.
I believe it would look better to have it like this: `random(randomValue)` rather than `if..else`.

tomazfernandes: Sure, makes sense, thanks. I'll open a PR there then, and we can improve the API as needed. I'll also take a look into how Spring Retry's builders are implemented.

tomazfernandes:
Closed in favor of [Spring Retry #277](https://github.com/spring-projects/spring-retry/pull/277).

Issue #2154, opened 2022-03-08
DelegatingByTypeSerializer does not work for Tombstones


dspruth:
**Expected Behavior**

DelegatingByTypeSerializer should **not** crash when receiving Tombstone (null value)
I should be able to configure a Serializer for Tombstones based on `KafkaNull`
```java
@Bean
public ProducerFactory<Object, Object> producerFactory(KafkaProperties config) {
    return new DefaultKafkaProducerFactory<>(config.buildProducerProperties(),
            null,
            new MyDelegatingByTypeSerializer(Map.of(
                    String.class, new StringSerializer(),
                    byte[].class, new ByteArraySerializer(),
                    KafkaNull.class, new ByteArraySerializer())));
}
```

**Current Behavior**

DelegatingByTypeSerializer throws NullpointerException on Tombstone message due to
```java
Serializer<?> delegate = delegates.get(data.getClass()); //data is null for null message
```

**Context**

I'm using tombstone messages to remove entries from compacted topics and trigger deletion in consumer services. When starting to use DelegatingByTypeSerializer producing the Tombstones lead to errors.
A possible fix could be:
```java
if (data == null) {
    delegate = this.delegates.get(KafkaNull.class);
} else {
    delegate = this.delegates.get(data.getClass());
}
```

garyrussell:
Thanks for reporting; I've changed this to a bug.

`KafkaNull` never appears in a `ProducerRecord`; it is an artifact for converting to/from spring-messaging `Message<?>`; null values should simply be returned as null.

PR #2160, opened 2022-03-08
GH-2154: Fix Tombstones with Delegating Seriali...

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2154

Check for null and return.

**cherry-pick to 2.8.x, 2.7.x (partial)**


artembilan:
... and cherry-picked to `2.8.x` & `2.7.x` after fixing conflicts.

Issue #2155, opened 2022-03-08
RetryableTopic - provide a mechanism to suppress non-essential headers


v-chernyshev:
**Expected Behavior**

I'd like to have a way to suppress the generation of the exception headers as they are quite verbose and drastically increase the size of the messages.

**Current Behavior**

Exception information headers are unconditionally added in this function: https://github.com/spring-projects/spring-kafka/blob/dc47e72a374dbfd5893d79e6dc841bc90d44362d/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java#L363-L364

**Context**

In our application the initial attempt is quite likely to fail as we have to wait for data to be available in an external system that does not provide native notifications.  The current retryable topics implementation adds a number of exception-related headers, including the full stacktrace information. When combined they increase the average message size from a couple hundred bytes to approximately 10 kilobytes. For us this overhead is hard to ignore.

Unfortunately, I do not see any way to hook into the header generation process. `setHeadersFunction` cannot be used as it is already customised by the recoverer factory: https://github.com/spring-projects/spring-kafka/blob/dc47e72a374dbfd5893d79e6dc841bc90d44362d/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java#L139

The only workaround I can think of is to add a custom producer interceptor that strips the headers out of the records, but then the CPU resources are wasted on rendering the information that is almost immediately thrown away. It would be nice to avoid generating these headers in the first place.

tomazfernandes:
Possibly related to [this SO question](https://stackoverflow.com/questions/71379877/how-to-add-custom-header-while-sending-message-to-dlt-using-spring-kafka-non-blo).

But I think it might be worth it to have a dedicated option not to include the stacktrace in `DeadLetterPublishingRecoverer`.

v-chernyshev:
Yeah, even if I get access to the original header function via reflection it won't help with suppressing the generation of the headers that are not strictly required. The reason for it is that `this.headersFunction.apply` is called after all the exception headers have already been set up.

garyrussell:
I suggest we add something like 

```java
public interface HeaderCreator {

    void create(Headers kafkaHeaders, Exception exception, boolean isKey); 

}
```

with the default being `this::addExceptionInfoHeaders`

WDYT?

v-chernyshev:
[...]

PR #2161, opened 2022-03-08
GH-2155: DeadLetterPublishingRecoverer Improvement

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2155

- Add a `BitSet` property to suppress individual standard headers
- Support multiple `headersFunction`
- Allow complete customization of exception headers
- Add `setHeadersFunction` to the DLPR factory for retryable topics

**cherry-pick to 2.8.x**

### DIFF
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -5641,7 +5641,7 @@ Key exceptions are only caused by `DeserializationException` s so there is no `D
 There are two mechanisms to add more headers.
 
 1. Subclass the recoverer and override `createProducerRecord()` - call `super.createProducerRecord()` and add more headers.
-2. Provide a `BiFunction` to receive the consumer record and exception, returning a `Headers` object; headers from there will be copied to the final producer record.
+2. Provide a `BiFunction` to receive the consumer record and exception, returning a `Headers` object; headers from there will be copied to the final producer record; also see <<dlpr-headers>>.
 Use `setHeadersFunction()` to set the `BiFunction`.
 
 The second is simpler to implement but the first has more information available, including the already assembled standard headers.
@@ -5735,6 +5735,35 @@ When repeatedly republishing a failed record, these headers can grow (and eventu
 The reason for the two properties is because, while you might want to retain only the last exception information, you might want to retain the history of which topic(s) the record passed through for each failure.
 
 `appendOriginalHeaders` is applied to all headers named `*ORIGINAL*` while `stripPreviousExceptionHeaders` is applied to all headers named `*EXCEPTION*`.
+
+Starting with version 2.8.4, you now can control which of the standard headers will be added to the output record.
+
+The there is a new property `whichHeaders`, which is a `BitSet`; for example, to suppress the addition of adding the stack trace header, use the following:
+
+====
+[source, java]
+----
+DeadLetterPublishingRecoverer recoverer = new DeadLetterPublishingRecoverer(template);
+recoverer.getWhichHeaders().clear(HeaderNames.HeadersToAdd.EX_STACKTRACE.getBit());
+----
+====
+
+See the `enum HeadersToAdd` for the bit names of the (currently) 10 standard headers that are added by default.
+
+In addition, you can completely customize the addition of exception headers by adding an `ExceptionHeadersCreator`; this also disables all standard exception headers.
+
+====
+[source, java]
+----
+DeadLetterPublishingRecoverer recoverer = new DeadLetterPublishingRecoverer(template);
+recoverer.setExceptionHeadersCreator((kafkaHeaders, exception, isKey, headerNames) -> {
+    kafkaHeaders.add(new RecordHeader(..., ...));
+});
+----
+====
+
+Also starting with version 2.8.4, you can now provide multiple headers functions, via the `addHeadersFunction` method.
+This allows additional functions to apply, even if another function  has already been registered, for example, when using <<retry-topic>>.
 
 Also see <<retry-headers>> with <<retry-topic>>.
 
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -5737,18 +5737,17 @@ The reason for the two properties is because, while you might want to retain onl
 `appendOriginalHeaders` is applied to all headers named `*ORIGINAL*` while `stripPreviousExceptionHeaders` is applied to all headers named `*EXCEPTION*`.
 
 Starting with version 2.8.4, you now can control which of the standard headers will be added to the output record.
+See the `enum HeadersToAdd` for the names of the (currently) 10 standard headers that are added by default.
 
-There is a new property `whichHeaders`, which is a `BitSet`; for example, to suppress the addition of adding the stack trace header, use the following:
+To exclude headers, use the `excludeHeaders()` method; for example, to suppress adding the exception stack trace in a header, use:
 
 ====
 [source, java]
 ----
 DeadLetterPublishingRecoverer recoverer = new DeadLetterPublishingRecoverer(template);
-recoverer.getWhichHeaders().clear(HeaderNames.HeadersToAdd.EX_STACKTRACE.getBit());
+recoverer.excludeHeaders(HeaderNames.HeadersToAdd.EX_STACKTRACE);
 ----
 ====
-
-See the `enum HeadersToAdd` for the bit names of the (currently) 10 standard headers that are added by default.
 
 In addition, you can completely customize the addition of exception headers by adding an `ExceptionHeadersCreator`; this also disables all standard exception headers.
 
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -5738,7 +5738,7 @@ The reason for the two properties is because, while you might want to retain onl
 
 Starting with version 2.8.4, you now can control which of the standard headers will be added to the output record.
 
-The there is a new property `whichHeaders`, which is a `BitSet`; for example, to suppress the addition of adding the stack trace header, use the following:
+There is a new property `whichHeaders`, which is a `BitSet`; for example, to suppress the addition of adding the stack trace header, use the following:
 
 ====
 [source, java]
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -5737,7 +5737,7 @@ The reason for the two properties is because, while you might want to retain onl
 `appendOriginalHeaders` is applied to all headers named `*ORIGINAL*` while `stripPreviousExceptionHeaders` is applied to all headers named `*EXCEPTION*`.
 
 Starting with version 2.8.4, you now can control which of the standard headers will be added to the output record.
-See the `enum HeadersToAdd` for the names of the (currently) 10 standard headers that are added by default.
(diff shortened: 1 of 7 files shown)


spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -5736,6 +5736,35 @@ The reason for the two properties is because, while you might want to retain onl
 
 `appendOriginalHeaders` is applied to all headers named `*ORIGINAL*` while `stripPreviousExceptionHeaders` is applied to all headers named `*EXCEPTION*`.
 
+Starting with version 2.8.4, you now can control which of the standard headers will be added to the output record.
+
+The there is a new property `whichHeaders`, which is a `BitSet`; for example, to suppress the addition of adding the stack trace header, use the following:
```
v-chernyshev, line 5741:
```suggestion
There is a new property `whichHeaders`, which is a `BitSet`; for example, to suppress the addition of adding the stack trace header, use the following:
```

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -572,17 +646,28 @@ private void enhanceHeaders(Headers kafkaHeaders, ConsumerRecord<?, ?> record, E
...
-		maybeAddHeader(kafkaHeaders, this.headerNames.original.partitionHeader,
-				ByteBuffer.allocate(Integer.BYTES).putInt(record.partition()).array());
-		maybeAddHeader(kafkaHeaders, this.headerNames.original.offsetHeader,
-				ByteBuffer.allocate(Long.BYTES).putLong(record.offset()).array());
-		maybeAddHeader(kafkaHeaders, this.headerNames.original.timestampHeader,
-				ByteBuffer.allocate(Long.BYTES).putLong(record.timestamp()).array());
-		maybeAddHeader(kafkaHeaders, this.headerNames.original.timestampTypeHeader,
-				record.timestampType().toString().getBytes(StandardCharsets.UTF_8));
-		if (ex instanceof ListenerExecutionFailedException) {
+		if (this.whichHeaders.get(HeaderNames.HeadersToAdd.TOPIC.getBit())) {
```
tomazfernandes, line 649: Wouldn't it maybe be clearer if this if statement was inside the `maybeAddHeaders` method, and we passed the enum as an argument?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
```diff
@@ -137,6 +150,9 @@ public DeadLetterPublishingRecoverer create() {
 		};
 
 		recoverer.setHeadersFunction((consumerRecord, e) -> addHeaders(consumerRecord, e, getAttempts(consumerRecord)));
+		if (this.headersFunction != null) {
```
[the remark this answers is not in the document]
garyrussell: Yes; I was going to do it here too, but I had already added to the code the DLPR and it would have been duplicated here - I also thought this could be useful in other scenarios, where a library has a "standard" corporate DLPR bean in a jar and a user wants to use that, plus add more headers.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -665,7 +758,80 @@ protected HeaderNames getHeaderNames() {
...
+		/**
+		 * Bits representing which headers to add.
+		 * @since 2.8.4
+		 */
+		public enum HeadersToAdd {
+
+			/**
+			 * The offset of the failed record.
+			 */
+			OFFSET(0),
```
artembilan, line 770: Looks like just an `ordinal()` is enough.
I mean all those explicit values for this `enum` is fully aligned with an `ordinal()` of the enum:
```

    /**
     * Returns the ordinal of this enumeration constant (its position
     * in its enum declaration, where the initial constant is assigned
     * an ordinal of zero).
     *
     * Most programmers will have no use for this method.  It is
     * designed for use by sophisticated enum-based data structures, such
     * as {@link java.util.EnumSet} and {@link java.util.EnumMap}.
     *
     * @return the ordinal of this enumeration constant
     */
    public final int ordinal() {
```

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
```diff
@@ -63,10 +64,22 @@
...
+
 	public DeadLetterPublishingRecovererFactory(DestinationTopicResolver destinationTopicResolver) {
 		this.destinationTopicResolver = destinationTopicResolver;
 	}
 
+	/**
+	 * Set a function that creates additional headers for the output record, in addition to the standard
+	 * retry headers added by this factory.
+	 * @param headersFunction the function.
+	 * @since
```
artembilan, line 77: `2.8.4` ?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -1005,4 +1171,24 @@ public ExceptionInfo exceptionStacktrace(String exceptionStacktrace) {
...
+	public interface ExceptionHeadersCreator {
+
+		/**
+		 * Create exception headers.
+		 * @param kafkaHeaders the {@link Headers} to add the header(s) to.
+		 * @param exception The exception.
+		 * @param isKey whether the exception is for a key or value.
+		 * @param headerNames the heaader names to use.
+		 */
+		void create(Headers kafkaHeaders, Exception exception, boolean isKey, HeaderNames headerNames);
```
v-chernyshev, line 1190: How would `headerNames` be accessed? Both `exceptionInfo` and `original` member variables are `private final`.

tomazfernandes: Yes, also the inner classes are package-private - can't really remember why they're not private. And properties have no getters. If we're letting end users handle this, guess we'll have to open up the classes and add the proper getters.

garyrussell: Good catch!

artembilan (approved):
I'm OK with this.

@tomazfernandes , @v-chernyshev ?

tomazfernandes:
> I'm OK with this.
> 
> @tomazfernandes , @v-chernyshev ?

Looks great to me. I didn’t know this BitSet class, works perfectly for this use case. Thanks.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -83,9 +88,11 @@
 
 	private final Function<ProducerRecord<?, ?>, KafkaOperations<?, ?>> templateResolver;
 
+	private final BitSet whichHeaders = new BitSet(10);
```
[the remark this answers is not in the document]
v-chernyshev: Hm. They say that a new array is allocated for every `values()` call, so I don't know whether it would make sense to also cache the length value...

One more question. What's the motivation for not using `EnumSet<HeaderNames.HeadersToAdd>` instead of `BitSet`? The [documentation](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/EnumSet.html) says that:

>  Enum sets are represented internally as bit vectors.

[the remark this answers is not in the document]
artembilan: After reading `EnumSet` JavaDocs, I concur that this might be a better way for our use-case.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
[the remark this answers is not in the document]
garyrussell: Agreed; I wasn't aware of it.

artembilan (approved):

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -45,6 +46,7 @@
 import org.springframework.kafka.KafkaException;
 import org.springframework.kafka.core.KafkaOperations;
 import org.springframework.kafka.core.ProducerFactory;
+import org.springframework.kafka.listener.DeadLetterPublishingRecoverer.HeaderNames.HeadersToAdd;
```
v-chernyshev, line 49: Just to confirm. There is no policy to avoid imports from the class that is declared in the same file, is there? There are two places that use `HeadersToAdd` rather than `HeaderNames.HeadersToAdd`, which causes this import statement to appear.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -326,6 +339,73 @@ public void setSkipSameTopicFatalExceptions(boolean skipSameTopicFatalExceptions
...
+			BiFunction<ConsumerRecord<?, ?>, Exception, Headers> toCompose = this.headersFunction;
+			this.headersFunction = (rec, ex) -> {
+				Headers headers1 = toCompose.apply(rec, ex);
+				if (headers1 == null) {
+					headers1 = new RecordHeaders();
+				}
+				Headers headers2 = headersFunction.apply(rec, ex);
+				if (headers2 != null) {
+					headers2.forEach(headers1::add);
+				}
```
v-chernyshev, line 403: There's one minor point of concern with this implementation. Imagine that the original `toCompose` function does something similar to this:

```java
Headers addHeaders(ConsumerRecord<...> record, Exception exception) {
    RecordHeaders extraHeaders = new RecordHeaders();
    extraHeaders.add("key", ...);
    extraHeaders.setReadOnly(); // This is the important line.

    return extraHeaders;
}
```

The returned value would be stored in `headers1`. `headers2.forEach(headers1::add);` would then fail with an `IllegalStateException` because `headers1` can no longer be modified. A safer, albeit slightly more expensive, alternative could be:

```java
// I believe the constructor is null-safe, but please double-check.
RecordHeaders headers1 = new RecordHeaders(toCompose.apply(rec, ex));
Headers headers2 = headersFunction.apply(rec, ex);
if (headers2 != null) {
    headers2.forEach(headers1::add);
}
return headers1;
```

garyrussell: Unusual; but thanks.

v-chernyshev: Yeah, I highly doubt anyone would actually do it, so I'm happy with the current implementation too.

tomazfernandes:
> I'm OK with this.
> Any other concerns before merging?

Not from me, thanks. And thanks @v-chernyshev for the thorough review and bringing in the EnumSet.

v-chernyshev:
> Not from me, thanks. And thanks @v-chernyshev for the thorough review and bringing in the EnumSet.

You are welcome, thanks for implementing this feature suggestion so quickly :) I do not have any further comments for this PR.

artembilan:
... and cherry-picked to `2.8.x` as https://github.com/spring-projects/spring-kafka/commit/f3f3eac955aa461abd06d631b23b0cf8d670bbd9 after fixing conflicts.

Issue #2068, opened 2022-01-13
Add a Lightweight MessageListener Adapter Supporting Message Conversion


garyrussell:
See https://stackoverflow.com/questions/70684560/concurrentkafkalistenercontainerfactory-message-converter-is-ignored-when-config/70685111#70685111

Similar to RabbitMQ, we could provide a thin adapter that performs message conversion.

breader124:
Hi, I'll try to start working on this issue

breader124:
Hi once again, it seems that after update introduced in 4e2f9fd project doesn't compile anymore due to failures in resolving micrometer related imports. What's more, one test fails even after I resolved problem with imports. I guess you'll be fixing it all soon (as the commit I mentioned is the last one on main branch and it was pushed last Friday), so I'll just wait for it and start working whenever it's ready, from my perspective it doesn't make sense to double the work

garyrussell:
@breader124 My sincere apologies - I was off on Friday and we have a holiday tomorrow, so I was trying to get a jump on Tuesday's releases (they were delayed from tomorrow because of the holidays). I should have checked the CI builds (I actually did, but on my 'phone and failed to notice the failure). Fixing now...

garyrussell:
Fixed. @breader124 once again sorry about that.

breader124:
@garyrussell, it's absolutely not a problem, thanks so much for quick reaction. Now I'm able to jump in

breader124:
Hi @garyrussell, no problem, I'll do it. I'd also like to ask you for a little guidance and validation of my initial thoughts. So basically as I understand this task, I need to provide concrete implementation of MessageListener interface, right? In order to do that, I'd propose to create another interface extending MessageListener and called ConvertingMessageListener. In this new interface, I'd provide default implementation for onMessage(ConsumerRecord<K, V> data) simply throwing UnsupportedOperationException and add unimplemented method void onMessage(T data, Consumer<?, ?> consumer). This way I'd be able to provide then concrete implementation of ConvertingMessageListener named let's say SimpleConvertingMessageListenerImpl. This class during creation could accept desired type and Consumer for already converted message. Example usage could look similarly to 

kafkaMessageListenerContainer.setupMessageListener(new ConvertingMessageListenerAdapter<>(Type desiredType, Consumer<?, ?> consumer));

What's your opinion about your solution? Please bear with me if I'm proposing something totally incorrect, it's the first time with this codebase from my perspective :)

[...]

PR #2165, opened 2022-03-15
GH-2068: Add lightweight converting adapter

breader124:
I've added lightweight message listener adapter converting consumer records to desired type specified by developer as mentioned in #2068

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapter.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright 2016-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.listener.adapter;
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.common.header.Header;
+
+import org.springframework.kafka.listener.AcknowledgingConsumerAwareMessageListener;
+import org.springframework.kafka.listener.AcknowledgingMessageListener;
+import org.springframework.kafka.listener.ConsumerAwareMessageListener;
+import org.springframework.kafka.listener.MessageListener;
+import org.springframework.kafka.support.Acknowledgment;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.converter.MessageConversionException;
+import org.springframework.messaging.converter.MessageConverter;
+import org.springframework.messaging.converter.SimpleMessageConverter;
+import org.springframework.messaging.support.GenericMessage;
+
+/**
+ * A {@link AcknowledgingConsumerAwareMessageListener} adapter that implements
+ * converting received {@link ConsumerRecord} using specified {@link MessageConverter}
+ * and then passes result to specified {@link MessageListener}.
+ *
+ * @param <T> the key type.
+ * @param <U> the value type.
+ * @param <V> the desired value type after conversion.
+ *
+ * @author Adrian Chlebosz
+ * @since 3.0
+ * @see AcknowledgingConsumerAwareMessageListener
+ */
+public class ConvertingAndDelegatingMessageListenerAdapter<T, U, V> implements AcknowledgingConsumerAwareMessageListener<T, U> {
+
+	private final Object delegate;
+	private final MessageConverter messageConverter;
+	private final Class<V> desiredValueType;
+
+	/**
+	 * Construct an instance with the provided {@link MessageListener} and {@link Class}
+	 * as a desired type of {@link ConsumerRecord}'s value after conversion. Default value of
+	 * {@link MessageConverter} is used, which is {@link SimpleMessageConverter}.
+	 *
+	 * @param delegate the {@link MessageListener} to use when passing converted {@link ConsumerRecord} further.
+	 * @param desiredValueType the {@link Class} setting desired type of {@link ConsumerRecord}'s value.
+	 */
+	public ConvertingAndDelegatingMessageListenerAdapter(Object delegate, Class<V> desiredValueType) {
+		validateMessageListener(delegate);
+		Objects.requireNonNull(desiredValueType);
+
+		this.delegate = delegate;
+		this.desiredValueType = desiredValueType;
+
+		this.messageConverter = new SimpleMessageConverter();
+	}
+
+	/**
+	 * Construct an instance with the provided {@link MessageListener}, {@link MessageConverter} and {@link Class}
+	 * as a desired type of {@link ConsumerRecord}'s value after conversion.
+	 *
+	 * @param delegate the {@link MessageListener} to use when passing converted {@link ConsumerRecord} further.
+	 * @param messageConverter the {@link MessageConverter} to use for conversion.
+	 * @param desiredValueType the {@link Class} setting desired type of {@link ConsumerRecord}'s value.
+	 */
+	public ConvertingAndDelegatingMessageListenerAdapter(Object delegate, MessageConverter messageConverter, Class<V> desiredValueType) {
+		validateMessageListener(delegate);
+		Objects.requireNonNull(messageConverter);
+		Objects.requireNonNull(desiredValueType);
+
+		this.delegate = delegate;
+		this.messageConverter = messageConverter;
+		this.desiredValueType = desiredValueType;
+	}
+
+	private void validateMessageListener(Object messageListener) {
+		Objects.requireNonNull(messageListener);
+		if (!(messageListener instanceof MessageListener)) {
+			throw new IllegalArgumentException("Passed message listener must be of MessageListener type");
+		}
+	}
+
+	@Override
+	public void onMessage(ConsumerRecord<T, U> data, Acknowledgment acknowledgment, Consumer<?, ?> consumer) { // NOSONAR
+		ConsumerRecord<T, V> convertedConsumerRecord = convertConsumerRecord(data);
+		if (this.delegate instanceof AcknowledgingConsumerAwareMessageListener) {
+			((AcknowledgingConsumerAwareMessageListener<T, V>) this.delegate).onMessage(convertedConsumerRecord, acknowledgment, consumer);
+		}
+		else if (this.delegate instanceof ConsumerAwareMessageListener) {
+			((ConsumerAwareMessageListener<T, V>) this.delegate).onMessage(convertedConsumerRecord, consumer);
+		}
+		else if (this.delegate instanceof AcknowledgingMessageListener) {
+			((AcknowledgingMessageListener<T, V>) this.delegate).onMessage(convertedConsumerRecord, acknowledgment);
+		}
+		else if (this.delegate instanceof MessageListener) {
+			((MessageListener<T, V>) this.delegate).onMessage(convertedConsumerRecord);
+		}
+	}
+
+	private ConsumerRecord<T, V> convertConsumerRecord(ConsumerRecord<T, U> data) { // NOSONAR
+		Header[] headerArray = data.headers().toArray();
+		Map<String, Object> headerMap = Arrays.stream(headerArray)
+				.collect(Collectors.toMap(Header::key, Header::value));
+
+		Message<U> message = new GenericMessage<>(data.value(), headerMap);
+		V converted = (V) this.messageConverter.fromMessage(message, this.desiredValueType);
+
+		if (converted == null) {
+			throw new MessageConversionException(message, "Message cannot be converted by used MessageConverter");
+		}
+
+		return rebuildConsumerRecord(data, converted);
+	}
+
+	private ConsumerRecord<T, V> rebuildConsumerRecord(ConsumerRecord<T, U> data, V converted) {
+		return new ConsumerRecord<>(
+				data.topic(),
+				data.partition(),
+				data.offset(),
+				data.key(),
+				converted
+		);
+	}
+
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapter.java
```diff
@@ -0,0 +1,143 @@
...
+	 * Construct an instance with the provided {@link MessageListener} and {@link Class}
+	 * as a desired type of {@link ConsumerRecord}'s value after conversion. Default value of
+	 * {@link MessageConverter} is used, which is {@link SimpleMessageConverter}.
+	 *
+	 * @param delegate the {@link MessageListener} to use when passing converted {@link ConsumerRecord} further.
+	 * @param desiredValueType the {@link Class} setting desired type of {@link ConsumerRecord}'s value.
+	 */
+	public ConvertingAndDelegatingMessageListenerAdapter(Object delegate, Class<V> desiredValueType) {
+		validateMessageListener(delegate);
+		Objects.requireNonNull(desiredValueType);
```
artembilan, line 68: We prefer to use `org.springframework.util.Assert` instead of blind `NullPointerException`

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapter.java
```diff
@@ -0,0 +1,143 @@
...
+
+		this.delegate = delegate;
+		this.messageConverter = messageConverter;
+		this.desiredValueType = desiredValueType;
+	}
+
+	private void validateMessageListener(Object messageListener) {
+		Objects.requireNonNull(messageListener);
+		if (!(messageListener instanceof MessageListener)) {
```
artembilan, line 96: Why then just don't require a `MessageListener delegate` in the constructor?

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapterTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * @author Adrian Chlebosz
+ * @since 3.0.0
+ *
+ */
+class ConvertingAndDelegatingMessageListenerAdapterTest {
```
artembilan, line 37: We prefer to name test classes as `*Tests`.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapter.java
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * @param <T> the key type.
+ * @param <U> the value type.
+ * @param <V> the desired value type after conversion.
+ *
+ * @author Adrian Chlebosz
+ * @since 3.0
+ * @see AcknowledgingConsumerAwareMessageListener
+ */
+public class ConvertingAndDelegatingMessageListenerAdapter<T, U, V> implements AcknowledgingConsumerAwareMessageListener<T, U> {
```
artembilan, line 52: I would call it just `ConvertingMessageListener`. Doesn't look like it adapts to something not compatible in API: https://www.baeldung.com/java-adapter-pattern

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapter.java
```diff
@@ -0,0 +1,143 @@
...
+		}
+		else if (this.delegate instanceof AcknowledgingMessageListener) {
+			((AcknowledgingMessageListener<T, V>) this.delegate).onMessage(convertedConsumerRecord, acknowledgment);
+		}
+		else if (this.delegate instanceof MessageListener) {
+			((MessageListener<T, V>) this.delegate).onMessage(convertedConsumerRecord);
+		}
+	}
+
+	private ConsumerRecord<T, V> convertConsumerRecord(ConsumerRecord<T, U> data) { // NOSONAR
```
artembilan, line 118: I don't think we care somehow about `T` & `U` in the logic of this class.
In the end what we only need is a `V` for the data type we would like to convert to.

artembilan (changes requested):
I'm not fully sure in the purpose behind this class, but here you are some my review at a glance.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapter.java
breader124: I tried to replace T and U with just ?, but then it turned out that I can't override onMessage method without compiler yelling at me with communicate "same erasure, yet neither overrides the other". Also from code readability perspective it's kinda useful in my opinion, because it's clearly visible, that key (T) stays the same and U is at the end converted to V. Anyway, if you don't feel convinced I'll try to put some additional effort to investigate how could I avoid using T and U keeping only V as a type of a value after conversion.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapter.java
breader124: Solved in [603202f](https://github.com/spring-projects/spring-kafka/pull/2165/commits/603202fa68d5a5d0115aceb844f80ad3829d4997)

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapterTest.java
breader124: Solved in [603202f](https://github.com/spring-projects/spring-kafka/pull/2165/commits/603202fa68d5a5d0115aceb844f80ad3829d4997)

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapter.java
breader124: Solved in [603202f](https://github.com/spring-projects/spring-kafka/pull/2165/commits/603202fa68d5a5d0115aceb844f80ad3829d4997)

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingAndDelegatingMessageListenerAdapter.java
breader124: Solved in [603202f](https://github.com/spring-projects/spring-kafka/pull/2165/commits/603202fa68d5a5d0115aceb844f80ad3829d4997)

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * @param <T> the key type.
+ * @param <U> the value type.
+ * @param <V> the desired value type after conversion.
+ *
+ * @author Adrian Chlebosz
+ * @since 3.0
+ * @see AcknowledgingConsumerAwareMessageListener
+ */
+public class ConvertingMessageListener<T, U, V> implements AcknowledgingConsumerAwareMessageListener<T, U> {
```
artembilan, line 52: I look at this from the end-users perspective where they have to think what exactly need to be specified for the `T` and `U` when they really don't care about them since `T` is always going to be an original one and `U` is always going to be converted to `V` independently of `U`.
Therefore I find them redundant and want to avoid questions about them and trying to explain that we need them merely to satisfy an internal code consistency for compiler.
I believe we still can find some hard casting behavior internally even if it could be handled with a `rawtypes` suppression...

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
```diff
@@ -0,0 +1,135 @@
...
+
+	/**
+	 * Construct an instance with the provided {@link MessageListener}, {@link MessageConverter} and {@link Class}
+	 * as a desired type of {@link ConsumerRecord}'s value after conversion.
+	 *
+	 * @param delegate the {@link MessageListener} to use when passing converted {@link ConsumerRecord} further.
+	 * @param messageConverter the {@link MessageConverter} to use for conversion.
+	 * @param desiredValueType the {@link Class} setting desired type of {@link ConsumerRecord}'s value.
+	 */
+	public ConvertingMessageListener(MessageListener<T, V> delegate, MessageConverter messageConverter, Class<V> desiredValueType) {
```
artembilan, line 84: If `messageConverter` is optional and we have a default one, it is better to extract it into a setter instead.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
```diff
@@ -0,0 +1,135 @@
...
+		Assert.notNull(messageConverter, "Message converter cannot be null");
+		Assert.notNull(desiredValueType, "Desired value type cannot be null");
+
+		this.delegate = delegate;
+		this.messageConverter = messageConverter;
+		this.desiredValueType = desiredValueType;
+	}
+
+	@Override
+	public void onMessage(ConsumerRecord<T, U> data, Acknowledgment acknowledgment, Consumer<?, ?> consumer) { // NOSONAR
```
artembilan, line 95: I'm not sure that we really need a `// NOSONAR` over here. The method doesn't look so complicated.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
```diff
@@ -0,0 +1,135 @@
...
+		V converted = (V) this.messageConverter.fromMessage(message, this.desiredValueType);
+
+		if (converted == null) {
+			throw new MessageConversionException(message, "Message cannot be converted by used MessageConverter");
+		}
+
+		return rebuildConsumerRecord(data, converted);
+	}
+
+	private ConsumerRecord<T, V> rebuildConsumerRecord(ConsumerRecord<T, U> data, V converted) {
```
artembilan, line 125: This method should be `static`.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
```diff
@@ -0,0 +1,135 @@
...
+	 * @param desiredValueType the {@link Class} setting desired type of {@link ConsumerRecord}'s value.
+	 */
+	public ConvertingMessageListener(MessageListener<T, V> delegate, Class<V> desiredValueType) {
+		Assert.notNull(delegate, "Delegate message listener cannot be null");
+		Assert.notNull(desiredValueType, "Desired value type cannot be null");
+
+		this.delegate = delegate;
+		this.desiredValueType = desiredValueType;
+
+		this.messageConverter = new SimpleMessageConverter();
```
artembilan, line 73: The `SimpleMessageConverter` does not make too much sense for a logic of this listener.
It just does this: `return (ClassUtils.isAssignableValue(targetClass, payload) ? payload : null);`. So, if we expect in this converter whatever is there has been deserialized for us, then we just don't need to use this converter at all.
Perhaps a `GenericMessageConverter` would be better...

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
breader124: Changed in #eeb9961c

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
breader124: Changed in eeb9961c

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
breader124: Changed in eeb9961c

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
breader124: Changed in eeb9961c

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
breader124: Changed in eeb9961c

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
artembilan: Please, stop making comments like this.
You just push commit to the PR and we are notified. 
We also will see your changes and how they are related to the review we have left before.
Otherwise it is spamming my mail box: first your new commit and this "duplication" comments in their individual emails 😄

breader124:
Hi @garyrussell, what's your opinion about this PR?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ConvertingMessageListener.java
```diff
@@ -0,0 +1,127 @@
...
+			this.delegate.onMessage(convertedConsumerRecord, acknowledgment);
+		}
+
+		this.delegate.onMessage(convertedConsumerRecord);
+	}
+
+	private ConsumerRecord convertConsumerRecord(ConsumerRecord data) {
+		Header[] headerArray = data.headers().toArray();
+		Map<String, Object> headerMap = Arrays.stream(headerArray)
+				.collect(Collectors.toMap(Header::key, Header::value));
```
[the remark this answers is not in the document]
breader124: I analysed the code one more time and figured out that creating GenericMessage with headers is pointless, because they don't play any role in conversion of message payload. Also in newly built ConsumerRecord they are just copied form received record. Based on that I deleted these lines related to extracting headers and take care only about payload in conversion. What's your opinion?

breader124:
Hi @garyrussell , will this PR be merged?

Issue #2164, opened 2022-03-11
DKPF default producerPerConsumerPartition to false and deprecate


garyrussell:
No longer needed now that `EOSMode.V1` is not supported.

PR #2171, opened 2022-03-15
GH-2164: Remove EOSMode.V1

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2164#

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
@@ -23,7 +23,6 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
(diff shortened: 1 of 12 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -756,7 +749,6 @@ public EOSMode getEosMode() {
 	@SuppressWarnings("deprecation")
```
artembilan, line 749: We don't need this any more.

At the same time why do we need this `EOSMode` at all since there is only one supported mode left?..

garyrussell: I left it, in case they come up with `EOSMode.V3` 😄

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -148,7 +148,7 @@
 
 	private String beanName = "not.managed.by.Spring";
 
-	private boolean producerPerConsumerPartition = true;
+	private boolean producerPerConsumerPartition = false;
```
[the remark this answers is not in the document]
artembilan: Any comments about this then, please?

Issue #2172, opened 2022-03-15
@RetryableTopic: support for re-processing of records after DLT handling


jgslima:
**Expected Behavior**
Support, or at least easiness, to allow designs where records may need to be "reprocessed" again after the DLT handling has been done and after the root cause of the processing error has been fixed.

**Current Behavior**
After retries have been exceeded, the application has to deal with the record in the `@DltHandler` method. But in some application cases, the record cannot be completely discarded.

**Context**
As for critical topics records that exceeded retries cannot be discarded, what we do here is to persist the record data somewhere else (tipically in a database).
The application has an Administration Console where the operations team can query and inspect the records that exceeded retries (and also the Exception data that caused the error, because in the `@DltHandler` we actually submit the record to the actual listener one more time to have the information of the Exception being thrown by the listener).

After the root cause of the error has been fixed (for instance, a database or an external service that was not responding gets back online, or a bug in the code has been fixed), through the console the operator has the means to select some of the records (or all from a specific original topic) and trigger a "reprocessing" of such records.

A good way of doing this reprocessing would be to restart the flow again, that is, to resend the record to topics chain again.

However, we cannot send the message to the _original_ topic, because we would be creating dirt in it. As it is my problem that I was not able to consume the record, I should send the record to _the first retry topic_ in my retriy topics chain.

- for fixed delay and multiple topics, this would be "`originalTopic-retry-0`".
- for fixed delay and single topic, this would be "`originalTopic-retry`"
- for non-fixed delay, this would be the first retry topic, like say "`originalTopic-retry-1000`"

What `spring-kafka` might provide would be easiness for the application to send a record to the **first** retry topic in the chain, for a given original topic. And it is fine that when sending a message to the first topic, if the consumer is currently stopped waiting for the delay of the messages that are already there.

This might be done in more than one way:

- just providing some utility to return to the application the name of the first retry topic in the same way the framework internally mount this name, and the application takes care of everything else.
- providing some mean/method that receives a `ProducerRecord` already instantiated by the application, and then populate in this record the needed headers and then send the record to the first retry topic.

In fact, we are already doing this by ourselves here, just sending the record to the first retry topic. But we are doing this in a risky way, because we duplicated the logic to mount the first topic name, and also, counting on luck because it turns out that posting a record without the internal control headers just work. But ideally the library should populate all the control headers properly.

tomazfernandes:
Hi @jgslima, thanks for bringing this up.

[...]

PR #2176, opened 2022-03-17
GH-2174: Register Retry Topic's components earlier

### LINKED ISSUE #2174: Retry Topic's components can't be autowired before feature bootstrapping
As discussed [here](https://github.com/spring-projects/spring-kafka/issues/2172#issuecomment-1068532530), currently `RetryTopic`'s components are lazily instantiated when a `RetryTopicConfiguration` is found for a given `@KafkaListener` bean.

Because of that, when a bean that autowires any of the components is instantiated before such bootstrapping happens, application startup fails with `NoSuchBeanException`.

Bootstrap early to avoid this problem.

tomazfernandes:
Fixes #2174

Please let me know if there's a more adequate way of solving this, or if there's anything to be changed. Thanks. 

---

Non-blocking delayed retries feature's beans were being registered lazily, when the first @KafkaListener annotated bean had a configuration associated with it in the `KafkaListenerAnnotationBeanPostProcessor`.

With that, if any bean autowired one of the feature's components before such bootstrapping it would throw a `NoSuchBeanDefinitionException`.

Now the components are registered in the `RetryTopicRegistryPostProcessor` before singletons are instantiated, so autowiring the feature's components in beans works as expected.

Adds a few adjustments to related classes and integration and unit tests.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -58,8 +58,6 @@ import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.config.Scope;
-import org.springframework.beans.factory.support.BeanDefinitionRegistry;
-import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ConfigurableApplicationContext;
@@ -86,7 +84,6 @@ import org.springframework.kafka.config.MultiMethodKafkaListenerEndpoint;
 import org.springframework.kafka.listener.ContainerGroupSequencer;
 import org.springframework.kafka.listener.KafkaListenerErrorHandler;
 import org.springframework.kafka.listener.adapter.RecordFilterStrategy;
-import org.springframework.kafka.retrytopic.RetryTopicBootstrapper;
 import org.springframework.kafka.retrytopic.RetryTopicConfiguration;
 import org.springframework.kafka.retrytopic.RetryTopicConfigurer;
 import org.springframework.kafka.retrytopic.RetryTopicInternalBeanNames;
@@ -503,33 +500,11 @@ public class KafkaListenerAnnotationBeanPostProcessor<K, V>
 		KafkaListenerContainerFactory<?> factory =
 				resolveContainerFactory(kafkaListener, resolve(kafkaListener.containerFactory()), beanName);
 
-		getRetryTopicConfigurer()
+		this.beanFactory
+				.getBean(RetryTopicInternalBeanNames.RETRY_TOPIC_CONFIGURER, RetryTopicConfigurer.class)
 				.processMainAndRetryListeners(endpointProcessor, endpoint, retryTopicConfiguration,
 						this.registrar, factory, this.defaultContainerFactoryBeanName);
 		return true;
-	}
-
-	private RetryTopicConfigurer getRetryTopicConfigurer() {
-		bootstrapRetryTopicIfNecessary();
-		return this.beanFactory.getBean(RetryTopicInternalBeanNames.RETRY_TOPIC_CONFIGURER, RetryTopicConfigurer.class);
-	}
-
-	private void bootstrapRetryTopicIfNecessary() {
-		if (!(this.beanFactory instanceof BeanDefinitionRegistry)) {
-			throw new IllegalStateException("BeanFactory must be an instance of "
-					+ BeanDefinitionRegistry.class.getSimpleName()
-					+ " to bootstrap the RetryTopic functionality. Provided beanFactory: "
-					+ this.beanFactory.getClass().getSimpleName());
-		}
-		BeanDefinitionRegistry registry = (BeanDefinitionRegistry) this.beanFactory;
-		if (!registry.containsBeanDefinition(RetryTopicInternalBeanNames
-				.RETRY_TOPIC_BOOTSTRAPPER)) {
-			registry.registerBeanDefinition(RetryTopicInternalBeanNames
-							.RETRY_TOPIC_BOOTSTRAPPER,
-					new RootBeanDefinition(RetryTopicBootstrapper.class));
-			this.beanFactory.getBean(RetryTopicInternalBeanNames
-					.RETRY_TOPIC_BOOTSTRAPPER, RetryTopicBootstrapper.class).bootstrapRetryTopic();
-		}
 	}
 
 	private Method checkProxy(Method methodArg, Object bean) {
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -58,6 +58,8 @@ import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.config.Scope;
+import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ConfigurableApplicationContext;
@@ -84,6 +86,7 @@ import org.springframework.kafka.config.MultiMethodKafkaListenerEndpoint;
 import org.springframework.kafka.listener.ContainerGroupSequencer;
 import org.springframework.kafka.listener.KafkaListenerErrorHandler;
 import org.springframework.kafka.listener.adapter.RecordFilterStrategy;
+import org.springframework.kafka.retrytopic.RetryTopicBootstrapper;
 import org.springframework.kafka.retrytopic.RetryTopicConfiguration;
 import org.springframework.kafka.retrytopic.RetryTopicConfigurer;
 import org.springframework.kafka.retrytopic.RetryTopicInternalBeanNames;
@@ -500,11 +503,34 @@ public class KafkaListenerAnnotationBeanPostProcessor<K, V>
 		KafkaListenerContainerFactory<?> factory =
 				resolveContainerFactory(kafkaListener, resolve(kafkaListener.containerFactory()), beanName);
 
-		this.beanFactory
-				.getBean(RetryTopicInternalBeanNames.RETRY_TOPIC_CONFIGURER, RetryTopicConfigurer.class)
+		getRetryTopicConfigurer()
 				.processMainAndRetryListeners(endpointProcessor, endpoint, retryTopicConfiguration,
 						this.registrar, factory, this.defaultContainerFactoryBeanName);
 		return true;
+	}
+
+	private RetryTopicConfigurer getRetryTopicConfigurer() {
+		bootstrapRetryTopicIfNecessary();
+		return this.beanFactory.getBean(RetryTopicInternalBeanNames.RETRY_TOPIC_CONFIGURER, RetryTopicConfigurer.class);
+	}
+
+	private void bootstrapRetryTopicIfNecessary() {
+		if (!(this.beanFactory instanceof BeanDefinitionRegistry)) {
+			throw new IllegalStateException("BeanFactory must be an instance of "
+					+ BeanDefinitionRegistry.class.getSimpleName()
+					+ " to bootstrap the RetryTopic functionality. Provided beanFactory: "
+					+ this.beanFactory.getClass().getSimpleName());
+		}
+		BeanDefinitionRegistry registry = (BeanDefinitionRegistry) this.beanFactory;
+		if (!registry.containsBeanDefinition(RetryTopicInternalBeanNames
+				.RETRY_TOPIC_BOOTSTRAPPER)) {
+			this.logger.warn(() -> "Could not bootstrap RetryTopic early. Bootstrapping lazily.");
+			registry.registerBeanDefinition(RetryTopicInternalBeanNames
+							.RETRY_TOPIC_BOOTSTRAPPER,
+					new RootBeanDefinition(RetryTopicBootstrapper.class));
+			this.beanFactory.getBean(RetryTopicInternalBeanNames
+					.RETRY_TOPIC_BOOTSTRAPPER, RetryTopicBootstrapper.class).bootstrapRetryTopic();
+		}
 	}
 
 	private Method checkProxy(Method methodArg, Object bean) {
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryTopicRegistryPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryTopicRegistryPostProcessor.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2022-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.annotation;
+
+import java.util.Arrays;
+
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProcessor;
+import org.springframework.beans.factory.support.RootBeanDefinition;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.core.Ordered;
+import org.springframework.core.type.AnnotationMetadata;
+import org.springframework.kafka.retrytopic.RetryTopicBootstrapper;
+import org.springframework.kafka.retrytopic.RetryTopicConfiguration;
+import org.springframework.kafka.retrytopic.RetryTopicInternalBeanNames;
+
+/**
+ * A {@link BeanDefinitionRegistryPostProcessor} implementation that registers
+ * the non-blocking delayed retries feature's components' beans if a
+ * {@link RetryableTopic} annotated method or a {@link RetryTopicConfiguration}
+ * bean is found.
+ *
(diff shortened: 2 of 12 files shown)


spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DefaultDestinationTopicResolver.java
```diff
@@ -58,18 +59,31 @@
...
+	 * @see RetryTopicBootstrapper
+	 */
+	public DefaultDestinationTopicResolver() {
 		this.sourceDestinationsHolderMap = new HashMap<>();
 		this.destinationsTopicMap = new HashMap<>();
 		this.contextRefreshed = false;
+		this.clock = Clock.systemUTC();
+	}
+
+	@SuppressWarnings("deprecated")
```
artembilan, line 81: Why do we need this suppression here?
Doesn't look like you use some other deprecated API...

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DefaultDestinationTopicResolver.java
```diff
@@ -197,6 +211,20 @@ public boolean isContextRefreshed() {
...
+		this.applicationContext = applicationContext;
+	}
+
+	/**
+	 * Sets the {@link Clock} instance to be used to determine
+	 * if the retry process is past timeout.
+	 * @param clock the clock instance
+	 */
+	public void setClock(Clock clock) {
+		this.clock = clock;
```
artembilan, line 225: `Assert.notNull()` ?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicBootstrapper.java
```diff
@@ -46,25 +47,17 @@
...
+public class RetryTopicBootstrapper implements ApplicationContextAware {
 
-	private final ApplicationContext applicationContext;
+	private ApplicationContext applicationContext;
 
-	private final BeanFactory beanFactory;
+	public RetryTopicBootstrapper() {
+	}
 
+	@SuppressWarnings("deprecated")
```
artembilan, line 57: DITTO

spring-kafka/src/test/java/org/springframework/kafka/annotation/RetryTopicRegistryPostProcessorIntegrationTests.java
```diff
@@ -0,0 +1,225 @@
...
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+/**
+ * @author Tomaz Fernandes
+ * @since 2.8.4
+ */
+@SpringJUnitConfig
+@DirtiesContext
+@EmbeddedKafka
```
artembilan, line 54: I guess we don't need these top-level annotation since all your tests are inside those nested classes.
However I wonder what is the reason of those nested classes at all...
Why those tests cannot be in top-level class (or classes) ?

garyrussell:
Thanks @tomazfernandes . I don't think I can get to review this before Monday's release.

It's a pretty big change - I wonder if it needs to be back-ported or whether the work around in the discussion is sufficient for the current releases?

artembilan (changes requested):
I agree with, @garyrussell , that this is too much.
We may only consider this for the current 3.0 version. Back-port is too risky.

Thanks

spring-kafka/src/test/java/org/springframework/kafka/annotation/RetryTopicRegistryPostProcessorIntegrationTests.java
tomazfernandes: I had to test whether it works with only one configuration in the registry or another, or none, so I had to have a separate context for each test.

Maybe there's a better way to solve this?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DefaultDestinationTopicResolver.java
tomazfernandes: I thought my IDE was complaining, I'll remove it, thanks.

garyrussell:
I am ok with releasing a 2.9, but would prefer if we can sync it with Boot 2.7.

I don't think we can do M1 on Monday for Boot's M3; the next window is Boot's RC1 in a month.

That means we would either have to go straight to RC1 on April 18, or we could do an interim M1 when this feature is ready.

I am not sure how receptive the Boot team will be to get an RC out of the blue so 2.9 may end up being a Bootless version that users have to manually override.

tomazfernandes:
I see, I hadn't noticed the Mx versions are also synced to Boot. Do you think having a bootless version would be a good solution?

If you're not worried with the codebases drifting apart, we can merge this only to 3.0.0, or break the PR as I mentioned.

I was planning on adding the `DestinationTopicResolver` solution as a feature in the docs with some API polishing, but I think it might be messy if it might depend on a workaround, so maybe it'd also be a 3.0.0 or 2.9.x exclusive feature, or at least hold it until this fix is released in those versions.

We might also hold this PR and consider merging it in a later moment if that's better, no worries, I'm ok with whatever works for you.

Thanks

garyrussell:
After further thought; I think a stand-alone 2.9 (if we decide to do one) would be preferable, rather than forcing Boot 2.7 users to go to 2.9, given that it will get little milestone exposure before an RC will be due.

You can see the schedule on the calendar; we generally release Monday before Boot's Thursday releases.

https://calendar.spring.io/

A decoupled 2.9 would give us a bit more flexibility to add more feature content, and we could release it early Summer instead of late Spring.

tomazfernandes:
> After further thought; I think a stand-alone 2.9 (if we decide to do one) would be preferable, rather than forcing Boot 2.7 users to go to 2.9, given that it will get little milestone exposure before an RC will be due.

Sure, we can do that if that's better.

>
> You can see the schedule on the calendar; we generally release Monday before Boot's Thursday releases.

Then I probably mixed up with Boot's 😄

>
> https://calendar.spring.io/

That's useful, thanks.

> 
> A decoupled 2.9 would give us a bit more flexibility to add more feature content, and we could release it early Summer instead of late Spring.

Ok, let's go this way then. I still have a couple of things to submit for this PR, I should commit it later today or tomorrow.

Thanks for looking into this.

tomazfernandes:
I've made a few adjustments and addressed the review comments.

I'm still not sure I've made the best choices for this PR - for example, I might have left the current bootstrapping as a fallback to the new one - it might help reducing risk and making the PR smaller. But it still wouldn't be risk free, so I'm not sure if it'd be any better considering we'd have more logic to deal with in the end.

Please let me know if you have any suggestions or any further concerns.

Thanks.

tomazfernandes:
I've made a few adjustments to this.

* Reestablished the legacy bootstrapping as a fallback
* Rolled back non-essential changes to existing components
* Added an IT to assure that users can disable the new bootstrapping method by adding a no-ops `RetryTopicRegistryPostProcessor` subclass bean
* Added a WARN log to let users know if early bootstrapping didn't work and the fallback kicked in - likely if that happens a lot someone will reach out and let us know

I think with this changes this PR will be less risky and easier to review.

Please let me know if you have any further concerns.

Thanks.

tomazfernandes:
Now consider this application:

```java
@SpringBootApplication
public class SO71705876 {

		SpringApplication.run(SO71705876.class, args);
	}

	@Configuration
	static class MyConfiguration {

		@Bean
		public ConsumerFactory<String, String> consumerFactory() {
			Map<String, Object> config = new HashMap<>();
			config.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092");
			config.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
			config.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
			config.put(ConsumerConfig.GROUP_ID_CONFIG, "grp_STRING");
			config.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
			return new DefaultKafkaConsumerFactory<>(config);
		}

		@Bean
		public ConcurrentKafkaListenerContainerFactory<String, String> kafkaListenerContainerFactory() {
			ConcurrentKafkaListenerContainerFactory<String, String> factory = new ConcurrentKafkaListenerContainerFactory<>();
			return factory;
		}

			ListenerContainerFactoryConfigurer lcfc = new ListenerContainerFactoryConfigurer(kafkaConsumerBackoffManager, deadLetterPublishingRecovererFactory, clock);
			lcfc.setBlockingRetryableExceptions(IllegalArgumentException.class, IllegalStateException.class);
			lcfc.setBlockingRetriesBackOff(new FixedBackOff(500, 5)); // Optional
			return lcfc;
		}
	}
}
```

With this listener:

```java
public class MyKafkaListener {

	Logger logger = LoggerFactory.getLogger(this.getClass());

	@RetryableTopic
	@KafkaListener(topics = "topic_string_data")
	public void consume(@Payload String message, @Header(KafkaHeaders.RECEIVED_TOPIC) String topic) {
		logger.info("Received message {} from topic {}", message, topic);
		throw new RuntimeException("My exception");
	}

}
```

We get this error at bootstrap:

```
***************************
APPLICATION FAILED TO START
***************************

Description:

Parameter 0 of method lcfc in org.spring.kafka.playground.MyConfiguration required a bean of type 'org.springframework.kafka.listener.KafkaConsumerBackoffManager' that could not be found.


Action:

Consider defining a bean of type 'org.springframework.kafka.listener.KafkaConsumerBackoffManager' in your configuration.
```

tomazfernandes:
I've deleted my last comment, it was actually a configuration error on my part, sorry about that 😄

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -524,6 +524,7 @@ private void bootstrapRetryTopicIfNecessary() {
 		BeanDefinitionRegistry registry = (BeanDefinitionRegistry) this.beanFactory;
 		if (!registry.containsBeanDefinition(RetryTopicInternalBeanNames
 				.RETRY_TOPIC_BOOTSTRAPPER)) {
+			this.logger.warn(() -> "Could not bootstrap RetryTopic early. Bootstrapping lazily.");
```
artembilan, line 527: Why do we need this warn?
Looks like this one is going to happen in 99% application in the field.

What is wrong with just silently registering that bean if it does not present yet?
We do something like this conditional registration all the time and no warnings.

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryTopicRegistryPostProcessor.java
```diff
@@ -0,0 +1,112 @@
...
+	private boolean hasRetryTopicConfiguration(BeanDefinition beanDef) {
+		return RetryTopicConfiguration.class.isAssignableFrom(beanDef.getResolvableType().toClass())
+				|| AnnotationMetadata
+					.introspect(beanDef.getResolvableType().toClass())
+					.hasAnnotatedMethods(RetryableTopic.class.getName());
+	}
+
+	private void bootstrapRetryTopic(BeanDefinitionRegistry registry) {
+		BeanDefinition beanDef = getOrRegisterBeanDefinition(registry);
+		this.applicationContext.getBean(RETRY_TOPIC_BOOTSTRAPPER_BEAN_NAME, RetryTopicBootstrapper.class)
```
artembilan, line 82: we cannot call `getBean()` in the `postProcessBeanDefinitionRegistry()`. This is the same "Inversion of Control impedance" I mentioned in other place.
That's too early to deal with beans.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DefaultDestinationTopicResolver.java
```diff
@@ -197,6 +203,12 @@ public boolean isContextRefreshed() {
 		return this.contextRefreshed;
 	}
 
+	@Override
+	public void setApplicationContext(ApplicationContext applicationContext) {
+		this.applicationContext = applicationContext;
+		this.clock = applicationContext.getBean(RetryTopicInternalBeanNames.INTERNAL_BACKOFF_CLOCK_BEAN_NAME, Clock.class);
```
artembilan, line 209: You cannot call `getBean()` from the `setApplicationContext()` - too early, may cause some bean initialization problem...

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryTopicRegistryPostProcessor.java
```diff
@@ -0,0 +1,112 @@
...
+				.bootstrapRetryTopic();
```
artembilan, line 83: This better to be done via `SmartInitializingSingleton`  impl on that `RetryTopicBootstrapper`. So, we don't need to have this `getBean()` over here and no early `BeanFactory` access.
Said that it feels like we don't need this `RetryTopicRegistryPostProcessor` at all...

artembilan (changes requested):

tomazfernandes:
Hi @artembilan, thanks again for looking into this.

Before I try to address your specific concerns, let me give you some more context - hopefully not as much as in the other PR 😄 - on why this came up. Or maybe you got the context and I didn't get your solution, but at least we can be sure we're in the same page.

Currently, at application bootstrapping, no `RetryTopic` components' beans are automatically registered, such as in the `KafkaBootstrapConfiguration` class, or boot's auto configuration. I did it that way so that, if the feature is not being used,  the components are simply not registered. You might have noticed this is a rather bean-heavy feature.

So these are registered by the `RetryTopicBootstrapper` only if `KafkaListenerAnnotationBeanPostProcessor` process a bean that has and topic associated with a `RetryTopicConfiguration` or `@RetryableTopic` annotation.

The problem with this is that when users try to autowire  any of the components, such as the `DestinationTopicContainer` as requested in #2172, depending on the order things happen, the component is not registered yet and we get a `NoSuchBeanDefinitionException`.

So, maybe my premise is wrong, and we should just register the BeanDefinitions whether or not the feature is being used - that would certainly be an easier path.

If we stick with the current premise, though, we need an earlier point in application bootstrapping to make this check and register the components. This is how I got to this solution using this `RegistryPostProcessor`, since it's executed before singletons are instantiated and its dependencies autowired.

Makes sense overall? I'll look into the specific concerns now.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
tomazfernandes: So, this is why I got the impression that you didn't get the context and wrote the comment below. Actually it's the opposite - I expect all of the applications to bootstrap the feature via the `RegistryPostProcessor`, and only if for some unexpected reason the `RPP`  is not triggered properly this logic should be executed.

Since that's an unexpected path, I'd rather know that this is happening so we can properly fix it, and when we get enough confidence we should be able to remove this logic.

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryTopicRegistryPostProcessor.java
tomazfernandes: I see. I think I can find a way around this, but it might require more changes than I'd like to.

tomazfernandes: I mean changes to existing components such as the `RetryTopicBootstrapper`.

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryTopicRegistryPostProcessor.java
tomazfernandes: Again, I think it's about the premise. If we want to keep the component's bootstrapping optional while also enabling their autowiring to the end user, I think the `SmartInitializingSingleton` hook is too late for that.

artembilan: Do we really need to have them to be "autowireable" in the end?
Are they not infrastructure beans to support something else which is already those `@RetryableTopic` or so?

artembilan:
OK. All sounds good.

Here is my take on this: let's make those beans registered unconditionally if there is a requirements to be able to inject them somewhere in the end-user code.
Even if these beans are going to be dangling for nothing, doesn't look like they are heavy, active or store some context to exhaust memory or CPU.

We may revise a conditional feature in the future.

Your scan for `@RetryableTopic` doesn't look bad. Only the problem here that we need opposite injection from these infrastructure beans into end-user stuff.
This is kinda anti-pattern and hard to implement and support eventually.

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryTopicRegistryPostProcessor.java
tomazfernandes: > OK. All sounds good.
> 
> Here is my take on this: let's make those beans registered unconditionally if there is a requirements to be able to inject them somewhere in the end-user code. Even if these beans are going to be dangling for nothing, doesn't look like they are heavy, active or store some context to exhaust memory or CPU.
> 
> We may revise a conditional feature in the future.

Now that I think of it, we have two kinds of components - infrastructure components used at assembly time, and runtime components that are used at, well, runtime.

I think the runtime components are the ones users might be interested in. For example in #2172, the `DestinationTopicContainer` is injected for the user to figure out what is the first retry-topic name for a given main topic and reprocess DLT messages.

I think it might also be useful for users to have access to the `KafkaConsumerBackOffManager` so they can know which topic partitions are currently backed off and their due time, for example. But that one might be trickier because it has a more complex instantiation, but that's something fixable, and probably we won't need it now.

So maybe we can register early only `DestinationTopicContainer` and maybe `KafkaBackOffManager` and let the infrastructure components be instantiated where they currently are, when `KLABPP` detects the feature.

How does this sound? Would `KafkaBootstrapConfiguration` be the right place for that?

> 
> Your scan for `@RetryableTopic` doesn't look bad. Only the problem here that we need opposite injection from these infrastructure beans into end-user stuff. This is kinda anti-pattern and hard to implement and support eventually.

I'm not sure I understand what opposite injection is. You mean restricting users' access to the infrastrucure beans via injection?

artembilan: Yes. Sounds like a plan.

I mean that you are going to register a bean dynamically, but at the same time it is asked for end-user autowiring. 
That's my concern where we may cause circular dependency.
That's where it might not be conditional. At least for now.

tomazfernandes: > Yes. Sounds like a plan.
> 

Great, I'll do that then. I'll also take a closer look if there are any other beans that might be interesting for users.

> I mean that you are going to register a bean dynamically, but at the same time it is asked for end-user autowiring. That's my concern where we may cause circular dependency. That's where it might not be conditional. At least for now.

Oh, I see, thanks for the explanation.

Thanks a lot for looking into this!

tomazfernandes:
I'm wondering, since we're adding quite a few improvements that we both want to reduce risk, but also get it in the hands of the users quickly since these are actual issues they're facing, how about we schedule a 2.10.0 after 2.9.0?

This way we could release 2.9.0 early, after the milestone exposure we think it's necessary, but don't have to wait until all features are done, or postpone everything else to 3.0.0.

garyrussell:
I'd rather not have 5 active branches; although 2.7.x officially goes out of OSS support in May, realistically, we will probably have a release or two after that.

I do not anticipate much of a delay for 2.9; we can release it in June; I just don't want to cram things into a new release in a couple of weeks by tying it to the Boot release train.

tomazfernandes:
I see, makes sense. I think June is a good timeline. After that, we can work with the 2.9.x versions for small changes and leave bigger ones to 3.0.0.

tomazfernandes:
Closed in favor of #2227

PR #2185, opened 2022-03-23
GH-2184 KafkaBackOffException logged as ERROR

### LINKED ISSUE #2184: RetryableTopic - breaking changes in the error handler

2.8.4.

**Describe the bug**

`ListenerContainerFactoryConfigurer` now creates its error handler here:

https://github.com/spring-projects/spring-kafka/blob/93efeb4f604b5c96810a57bc584c18c194b7c321/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java#L235-L246

There are two breaking changes/regressions compared to version 2.8.3:

- `createDefaultErrorHandlerInstance` no longer provides a "no retry" back off strategy by default. The outcome of this change is that a to-be-recovered record is retried 10 times instead of immediately being sent into the next recovery topic.
- `errorHandler.defaultFalse` completely disables the default classification of the fatal exceptions (such as `DeserializationException.class`). In fact, every classification request now returns `false`.

The second change causes every back off exception to be printed at error level due to the code here:

https://github.com/spring-projects/spring-kafka/blob/f0ad7b0c0b2005eb2fd58fd361eacbfe007c7284/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java#L143-L164

`getClassifier().classify(thrownException)` now always returns false and `this.failureTracker.getRecoverer().accept` may throw a back off exception, which does not reach the following special back off exception handler any longer:

https://github.com/spring-projects/spring-kafka/blob/93efeb4f604b5c96810a57bc584c18c194b7c321/spring-kafka/src/main/java/org/springframework/kafka/listener/SeekUtils.java#L109-L113

**To Reproduce**

Unfortunately, I do not have a sample right now, but I will try to prepare it this week. I believe that any application that uses the retryable topics recovery mechanism is impacted.

**Expected behavior**

I expect the default behaviour of the retryable topics recovery system to be identical to version 2.8.3.

tomazfernandes:
Fixes #2184

I also looked into leveraging `KafkaException`'s self logging features here, but in order for it to work we'd need to make a few adjustments to `KafkaMessageListenerContainer#decorateException`, and that looks like too much change for a bugfix to me.

I still think we could have some configurable feature where a `KafkaException` could look into its causes, and use the deepest KE log level to self log. This way we could for example throw the original `KafkaBackOffException` as `DEBUG` (or what the user configures), and leave all other exceptions as `ERROR` (or what the user configures), and no matter how many wrappings it goes through it'd still be logged as `DEBUG`.

Not sure if that's something that would be useful for scenarios other than KBE though.

Please let me know if there's anything to change.

Thanks

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
@@ -152,8 +152,14 @@ public abstract class FailedRecordProcessor extends ExceptionClassifier implemen
 			}
 			catch (Exception ex) {
 				if (records.size() > 0) {
-					this.logger.error(ex, () -> "Recovery of record ("
-							+ KafkaUtils.format(records.get(0)) + ") failed");
+					if (SeekUtils.isBackoffException(ex)) {
+						this.logger.debug("Recovery of record ("
+								+ KafkaUtils.format(records.get(0)) + ") backed off: " + ex.getMessage());
+					}
+					else {
+						this.logger.error(ex, () -> "Recovery of record ("
+								+ KafkaUtils.format(records.get(0)) + ") failed");
+					}
 					this.failureTracker.getRetryListeners().forEach(rl ->
 							rl.recoveryFailed(records.get(0), thrownException, ex));
 				}


v-chernyshev:
The change looks good to me.

Out of curiosity, why does `getRecoveryStrategy` behave differently depending on the result of `getClassifier().classify`? If the result is `true` then a functor that can perform the recovery is returned (so there are no immediate side effects). In case of a `false` outcome, however, the first record is recovered immediately, which is a bit unexpected given the method name.

tomazfernandes:
@v-chernyshev thanks for looking into this. I don't have a proper answer to your question, so I defer to @garyrussell on this one.

garyrussell:
It's historical; before 2.7, the method was called `getSkipPredicate()` which returned `FailedRecordTracker.skip()` (for true) and `ALWAYS_SKIP_PREDICATE` (or `NEVER...` if recovery failed) for false (but it also did the immediate recovery for not retryable exceptions).

On reflection, it could (and maybe should) be refactored to use the newer `setBackOffFunction` to return a `FixedBackOff(0L, 0L)` for exceptions that are classified as not retryable.

tomazfernandes:
I agree there's room for streamlining this code a bit, and it would help keeping consistency between the various scenarios.

garyrussell:
Right, but not in this PR; I'll add an issue to the backlog.

garyrussell (approved):

garyrussell:
I think we should consider not calling the retry listeners for these exceptions, but that would be a different issue/PR.

garyrussell:
...and back-ported to 2.8.x as da0b820d

tomazfernandes:
> Right, but not in this PR; I'll add an issue to the backlog.

Sure. I was thinking something along these lines:
https://github.com/tomazfernandes/spring-kafka/commit/cc1b692c83ef4e35ab35d7791621e247a7df08fd

Would still need to nitpick if there's any behavior change, but all tests are passing, and we make sure that all scenarios are using the same recovery logic.

A possible improvement would be having the exception classification inside the `FailedRecordTracker` class, but we'd have to either pass the classification as an argument (would require signature change) or having it implement `ExceptionClassifier` and maybe delegate all method calls from `FailedRecordProcessor`. Looks like to much for too little to me.

Of course, this is just a suggestion, if anyone else has a different idea that's ok.

Thanks.

tomazfernandes:
A version without the inner class - it has less changes and should be better to review.
https://github.com/tomazfernandes/spring-kafka/commit/9db49ee5174da2737e072c4d9ed26506bfdd975d

garyrussell:
I was thinking something like this...
```
% git diff
diff --git a/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java b/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
index fa2a6071..3c6198f0 100644
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
@@ -28,7 +28,9 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.kafka.support.TopicPartitionOffset;
 import org.springframework.lang.Nullable;
 import org.springframework.util.backoff.BackOff;
 
 /**
  * Common super class for classes that deal with failing to consume a consumer record.
@@ -43,14 +45,27 @@ public abstract class FailedRecordProcessor extends ExceptionClassifier implemen
 
        private static final BiPredicate<ConsumerRecord<?, ?>, Exception> NEVER_SKIP_PREDICATE = (r, e) -> false;
 
+       private static final BackOff NO_RETRIES_OR_DELAY_BACKOFF = new FixedBackOff(0L, 0L);
+
+       private final BiFunction<ConsumerRecord<?, ?>, Exception, BackOff> noRetriesForClassified =
+                       (rec, ex) -> {
+                               if (!getClassifier().classify(ex)) {
+                                       return NO_RETRIES_OR_DELAY_BACKOFF;
+                               }
+                               return this.userBackOffFunction.apply(rec, ex);
+                       };
+
 
        private final FailedRecordTracker failureTracker;
 
        private boolean commitRecovered;
 
+       private BiFunction<ConsumerRecord<?, ?>, Exception, BackOff> userBackOffFunction = (rec, ex) -> null;
+
        protected FailedRecordProcessor(@Nullable BiConsumer<ConsumerRecord<?, ?>, Exception> recoverer, BackOff backOff) {
                this.failureTracker = new FailedRecordTracker(recoverer, backOff, this.logger);
+               this.failureTracker.setBackOffFunction(this.noRetriesForClassified);
        }
 
        /**
@@ -77,7 +92,8 @@ public abstract class FailedRecordProcessor extends ExceptionClassifier implemen
         * @since 2.6
         */
        public void setBackOffFunction(BiFunction<ConsumerRecord<?, ?>, Exception, BackOff> backOffFunction) {
-               this.failureTracker.setBackOffFunction(backOffFunction);
+               Assert.notNull(backOffFunction, "'backOffFunction' cannot be null");
+               this.userBackOffFunction = backOffFunction;
        }
 
        /**
```

But your way works too.

tomazfernandes:
Sure, this looks good too, and has the benefit of not having to expose the recovery logic to the `FailedRecordProcessor`.

I'm ok with both approaches, as both will let us have a single recovery logic for all paths.

So go ahead with yours if you prefer, no worries, or let me know and I can wrap up mine and open a PR.

PR #2188, opened 2022-03-23
GH-2170: Add a way to create a custom OffsetAndMetadata

### LINKED ISSUE #2170: Provide a way to create an OffsetAndMetdata with metadata when using Spring-Kafka
### Discussed in https://github.com/spring-projects/spring-kafka/discussions/2169

<div type='discussions-op-text'>

<sup>Originally posted by **frosiere** March 15, 2022</sup>
In some specific cases, it may make sense to set/customize the metadata field of the _OffsetAndMetadata_ object with data coming from the consumer. When using Spring Kafka, it's impossible as this object is created internally in _KafkaMessageListenerContainer.ListenerConsumer_.

Due to this, a callback can be set on the container properties but the metadata will always be empty when calling the _onComplete_ method of the callback.

This may help having a mechanism to set these metadata.

So, the proposal would be to either review the way to set a callback or to simply have an _OffsetAndMetadataProvider_ as follow
```
public interface OffsetAndMetadataProvider {

   OffsetAndMetadata provide(Consumer<?, ?> consumer, long offset);
}
```
This provider would then be set on the container properties as show in the following portion of code
```
public class MyOffsetAndMetadataProvider  implements OffsetAndMetadataProvider {

   @Override   
   OffsetAndMetadata provide(Consumer<?, ?> consumer, long offset) {
      // in this case, metadata would be the consumer group id...
      return new OffsetAndMetadata(offset, consumer.groupMetadata().groupId());
   }
}

// at consumer factory level
...
getContainerProperties().setOffsetAndMetadataProvider(new MyOffsetAndMetadataProvider());
...
```
Internally, Spring Kafka would need to be slightly reworked to use the following kind of method instead of directly creating the _OffsetAndMetadata_ object.

```
// in KafkaMessageListenerContainer.ListenerConsumer
// would be called 6 times
private OffsetAndMetadata createOffsetAndMetadata(long offset) {
   final OffsetAndMetadataProvider offsetAndMetadataProvider = containerProperties.getOffsetAndMetadataProvider();
   return offsetAndMetadataProvider == null
         ? new OffsetAndMetadata(offset)
         : offsetAndMetadataProvider.provide(consumer, offset)
}
```
Any other way to set metadata is more than welcome.

If we agree on the need, I can contribute on the implementation. This one or another one.

Thanks for your reply.</div>

frosiere:
Resolves #2170

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -34,6 +34,7 @@ import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.clients.admin.DescribeTopicsResult;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
+import org.apache.kafka.clients.consumer.OffsetCommitCallback;
 import org.apache.kafka.common.TopicPartition;
 
 import org.springframework.beans.BeanUtils;
@@ -166,6 +167,14 @@ public abstract class AbstractMessageListenerContainer<K, V>
 		}
 		if (this.containerProperties.getConsumerRebalanceListener() == null) {
 			this.containerProperties.setConsumerRebalanceListener(createSimpleLoggingConsumerRebalanceListener());
+		}
+		final OffsetCommitCallback commitCallback = containerProperties.getCommitCallback();
+		if (commitCallback != null) {
+			this.containerProperties.setCommitCallback(commitCallback);
+		}
+		final OffsetAndMetadataProvider offsetAndMetadataProvider = containerProperties.getOffsetAndMetadataProvider();
+		if (offsetAndMetadataProvider != null) {
+			this.containerProperties.setOffsetAndMetadataProvider(containerProperties.getOffsetAndMetadataProvider());
 		}
 	}
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -34,7 +34,6 @@ import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.clients.admin.DescribeTopicsResult;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
-import org.apache.kafka.clients.consumer.OffsetCommitCallback;
 import org.apache.kafka.common.TopicPartition;
 
 import org.springframework.beans.BeanUtils;
@@ -167,14 +166,6 @@ public abstract class AbstractMessageListenerContainer<K, V>
 		}
 		if (this.containerProperties.getConsumerRebalanceListener() == null) {
 			this.containerProperties.setConsumerRebalanceListener(createSimpleLoggingConsumerRebalanceListener());
-		}
-		final OffsetCommitCallback commitCallback = containerProperties.getCommitCallback();
-		if (commitCallback != null) {
-			this.containerProperties.setCommitCallback(commitCallback);
-		}
-		final OffsetAndMetadataProvider offsetAndMetadataProvider = containerProperties.getOffsetAndMetadataProvider();
-		if (offsetAndMetadataProvider != null) {
-			this.containerProperties.setOffsetAndMetadataProvider(containerProperties.getOffsetAndMetadataProvider());
 		}
 	}
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -557,6 +557,12 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 				? this.containerProperties.getCommitCallback()
 				: new LoggingCommitCallback();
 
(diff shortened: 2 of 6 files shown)


garyrussell:
Thanks for the contribution - see the PR build; we'll review when it is green...

>> Task :spring-kafka:checkstyleTest
Error: eckstyle] [ERROR] /home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java:26: Please use BDDMockito instead of Mockito.(when|doThrow|doAnswer). [RegexpSinglelineJava]

frosiere:
Build is green. Code will be updated based on the review feedbacks.

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerMetadata.java
```diff
@@ -0,0 +1,39 @@
...
+	 * Return the listener id.
+	 * @return the listener id.
+	 */
+	String getListenerId();
+
+	/**
+	 * Return the group id.
+	 * @return the group id.
+	 */
+	String getGroupId();
```
garyrussell, line 38: I think we should add `byte[] getListenerInfo()` (new property on `@KafkaListener` since 2.8.4).

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -167,6 +168,10 @@ protected AbstractMessageListenerContainer(ConsumerFactory<? super K, ? super V>
 		if (this.containerProperties.getConsumerRebalanceListener() == null) {
 			this.containerProperties.setConsumerRebalanceListener(createSimpleLoggingConsumerRebalanceListener());
 		}
+		final OffsetCommitCallback commitCallback = containerProperties.getCommitCallback();
+		if (commitCallback != null) {
+			this.containerProperties.setCommitCallback(commitCallback, containerProperties.getOffsetAndMetadataProvider());
```
garyrussell, line 173: It's a bit on an anti-pattern to name the method `set...` with 2 parameters; in any case, it's not clear why you are re-setting the callback to the same value.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3079,6 +3078,26 @@ public String toString() {
 					+ "\n]";
 		}
 
+		private OffsetAndMetadata createOffsetAndMetadata(long offset) {
+			final OffsetAndMetadataProvider metadataProvider = this.containerProperties.getOffsetAndMetadataProvider();
+			return metadataProvider == null
+					? new OffsetAndMetadata(offset)
+					: metadataProvider.provide(new ConsumerAwareListenerMetadata(), offset);
```
garyrussell, line 3085: No need to create a new one each time; since it is immutable - it can be a final field.

Also the null check can be avoided given that you already have a default implementation and you could add an `Assert.notNull` on the setter.

You could also add the provider as a final field.

garyrussell (changes requested):
Thanks for the contribution; a few suggestions...

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerMetadata.java
frosiere: Good idea, field added.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
frosiere: Agree, I created a second setter, following the same approach than the callback.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
frosiere: Code as been reworked as suggested.
I tried to keep the same approach than the callback at properties level. If you really prefer, I can create a default provider at property and  consider it will never be null using the assert but this may be weird to always see the default when rendering the properties...

garyrussell:
Thanks; in future, just add another commit so it's easier to see what has changed; we will squash during the merge.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -167,6 +168,14 @@ protected AbstractMessageListenerContainer(ConsumerFactory<? super K, ? super V>
...
 			this.containerProperties.setConsumerRebalanceListener(createSimpleLoggingConsumerRebalanceListener());
 		}
+		final OffsetCommitCallback commitCallback = containerProperties.getCommitCallback();
+		if (commitCallback != null) {
+			this.containerProperties.setCommitCallback(commitCallback);
+		}
+		final OffsetAndMetadataProvider offsetAndMetadataProvider = containerProperties.getOffsetAndMetadataProvider();
+		if (offsetAndMetadataProvider != null) {
+			this.containerProperties.setOffsetAndMetadataProvider(containerProperties.getOffsetAndMetadataProvider());
+		}
```
garyrussell, line 178: This code is unnecessary; these properties were already copied by `BeanUtils.copyProperties()` above.

garyrussell (changes requested):
Thanks. Just one nitpick.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
frosiere: Useless code has bee removed for both OffsetCommitCallback and OffsetAndMetadataProvider.
Thanks for the review.

garyrussell (approved):

garyrussell:
Thanks; merged as 50498f985ab476c5260e0621d918f721ad831676 and cherry-picked to 2.8.x as 5246daa3 after resolving conflicts.

PR #2189, opened 2022-03-24
DLPR - generate header values lazily.

v-chernyshev:
The `this.whichHeaders.contains` check happens after the byte array (in
case of `maybeAddHeader`) or the `RecordHeader` (in `appendOrReplace`)
has already been created. It may be quite an expensive operation, so in
this PR I propose to pass a supplier into these two functions instead.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
@@ -31,6 +31,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.Consumer;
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -659,59 +660,69 @@ private void enhanceHeaders(Headers kafkaHeaders, ConsumerRecord<?, ?> record, E
...
-				getStackTraceAsString(exception).getBytes(StandardCharsets.UTF_8)),
+		appendOrReplace(kafkaHeaders,
+				() -> new RecordHeader(
+						isKey ? names.exceptionInfo.keyExceptionStacktrace : names.exceptionInfo.exceptionStacktrace,
+						getStackTraceAsString(exception).getBytes(StandardCharsets.UTF_8)),
 				HeaderNames.HeadersToAdd.EX_STACKTRACE);
 	}
 
-	private void appendOrReplace(Headers headers, RecordHeader header, HeaderNames.HeadersToAdd hta) {
+	private void appendOrReplace(Headers headers, Supplier<RecordHeader> headerSupplier, HeaderNames.HeadersToAdd hta) {
```
tomazfernandes, line 723: Just as a cosmetic suggestion that might clean up the code a little bit, I wonder if we could change this method's signature to use `String header, Supplier<byte[]> valueSupplier` instead of `Supplier<RecordHeader> header supplier`, and then just call the `headers.add()` method with those like we do at `maybeAddHeaders`.

Of course, it doesn't really change anything, but since we're adding a few elements it might help keeping the overall size down, and also would make those method's signatures more consistent.

tomazfernandes:
I like this idea and am ok with the solution. Just added a cosmetic suggestion that I'm fine if we don't include.

Thanks @v-chernyshev for bringing this!

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
v-chernyshev: Yes, it indeed looks neater with the proposed change.

tomazfernandes (approved):
LGTM, thanks

artembilan (approved):

garyrussell:
```
[ant:checkstyle] [ERROR] /Users/grussell/Development/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java:43:8: Unused import - org.apache.kafka.common.header.internals.RecordHeader. [UnusedImports]
```
FYI, running `./gradlew clean check` locally will find these rather than waiting for the GH action.

Fixed during merge; merged as d65a9f05985468ff315b9ffc872221010f03694b and cherrry-picked to 2.8.x as 5df194ef610cc3017cdb532a8e3ddfeeb9e1dac1

PR #2200, opened 2022-03-29
GH-2198: Switch from Timer to Observation

garyrussell:
https://github.com/spring-projects/spring-kafka/issues/2198 Phase 1


artembilan (changes requested):
I thought Observation is about tracing.
Does your change cover both metrics and tracing together?

Any pointers from where did you take an inspiration?

Thanks

garyrussell:
I looked a Spring Batch, per this document https://docs.google.com/document/d/1hF3djSMo45NcaivbFEI3E9ruITm-3DjOSZ6VxO4Kg8U/edit#

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2079,11 +2073,10 @@ private void batchAfterRollback(final ConsumerRecords<K, V> records,
 		private RuntimeException doInvokeBatchListener(final ConsumerRecords<K, V> records, // NOSONAR
 				List<ConsumerRecord<K, V>> recordList) {
 
-			Object sample = startMicrometerSample();
-			try {
+			Observation observation = startObserving();
+			try (Scope scope = observation.openScope()) {
```
artembilan, line 2077: Thanks for the doc, Gary!

So, this one was my question: looks like `lowCardinality` is about metrics, the `highCardinality` for tracing.
And according this migration docs they say: https://github.com/micrometer-metrics/micrometer/wiki/Migrating-to-new-2.0.x-Observation-API#what-is-this-observationscope-about

> This might not be relevant from only Metrics point of view, however it’s critical from Tracing / Logging perspective.

Therefore my concern if you are going to stay just with metrics or looking forward for tracing as well.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
garyrussell: As I said on Slack yesterday...

>Next up: tracing.

PR #2201, opened 2022-03-30
Removed unused configuredBy set

im-pratham:
The configuredBy set is updated in the lambda but never used. Hence deleted it to avoid confusion. Let me know if this set is intended to be used somewhere.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,7 +66,8 @@ ext {
 	junitJupiterVersion = '5.8.2'
 	kafkaVersion = '3.1.0'
 	log4jVersion = '2.17.2'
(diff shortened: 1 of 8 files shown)


artembilan:
This was missed after moving to version `3.0`: https://github.com/spring-projects/spring-kafka/commit/0ba3e168a57089b43a59c8113a1a33d8c7ec7361#diff-2bc21a21eee3db747c6d5274c727e999da34089b0aa433ad7dca5106a70caf10.

Thank you, @im-pratham ; looking forward for more!

im-pratham:
Thanks for the details, @artembilan.

Issue #2197, opened 2022-03-28
Consider a Fatal Consumer Exception as an abnormal termination for a container


Rajh:
**Expected Behavior**

When there is an exception in a consumer like ACLs :
`Authentication/Authorization Exception and no authExceptionRetryInterval set` the container should be terminated abnormally.
The Container should report `isInExpectedState()` as `false`

**Current Behavior**

Currently, there is a log `Fatal consumer exception; stopping container` and then the container is stopped **normally**.
[github code](https://github.com/spring-projects/spring-kafka/blob/31b78c4386b70418c4b7c4a58c7b5f8fe0f2e96d/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1737)

PR #2202, opened 2022-03-31
GH-2197: Fix Container State After Fatal Stop

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2197

When a container is stopped due to any fatal error, `isInExpectedState()`
should return false.


garyrussell:
Oops; right - cancel that.

PR #2204, opened 2022-04-01
Add default ctor to ListenerContainerFactoryConfigurer

### LINKED ISSUE #2203: Improve RT's Blocking Retries Configuration
See: https://stackoverflow.com/questions/71705876/combining-blocking-and-non-blocking-retries-in-spring-kafka/71706953#71706953

tomazfernandes:
Resolves #2203
https://github.com/spring-projects/spring-kafka/issues/2203

I wonder how you feel about these changes. If there's a better way to handle the dependency injection or any other concerns, please let me know.

Thanks.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
@@ -27,6 +27,9 @@ import java.util.regex.Pattern;
 
 import org.apache.commons.logging.LogFactory;
 
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.BeanFactory;
+import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.kafka.KafkaException;
@@ -63,7 +66,7 @@ import org.springframework.util.backoff.BackOff;
  * @since 2.7
  *
  */
-public class ListenerContainerFactoryConfigurer {
+public class ListenerContainerFactoryConfigurer implements BeanFactoryAware {
 
 	private static final Set<ConcurrentKafkaListenerContainerFactory<?, ?>> CONFIGURED_FACTORIES_CACHE;
 
@@ -92,12 +95,16 @@ public class ListenerContainerFactoryConfigurer {
 	private Consumer<CommonErrorHandler> errorHandlerCustomizer = errorHandler -> {
 	};
 
-	private final DeadLetterPublishingRecovererFactory deadLetterPublishingRecovererFactory;
+	private DeadLetterPublishingRecovererFactory deadLetterPublishingRecovererFactory;
 
-	private final KafkaConsumerBackoffManager kafkaConsumerBackoffManager;
+	private KafkaConsumerBackoffManager kafkaConsumerBackoffManager;
 
-	private final Clock clock;
+	private Clock clock;
 
+	public ListenerContainerFactoryConfigurer() {
+	}
+
+	@Deprecated
 	public ListenerContainerFactoryConfigurer(KafkaConsumerBackoffManager kafkaConsumerBackoffManager,
 									DeadLetterPublishingRecovererFactory deadLetterPublishingRecovererFactory,
 									@Qualifier(RetryTopicInternalBeanNames
@@ -316,6 +323,18 @@ public class ListenerContainerFactoryConfigurer {
 				() -> String.format("The provided class %s is not assignable from %s",
 						obj.getClass().getSimpleName(), clazz.getSimpleName()));
 		return (T) obj;
+	}
+
+	@Override
+	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
+		this.kafkaConsumerBackoffManager =
+				beanFactory.getBean(RetryTopicInternalBeanNames.KAFKA_CONSUMER_BACKOFF_MANAGER,
+						KafkaConsumerBackoffManager.class);
+		this.deadLetterPublishingRecovererFactory =
+				beanFactory.getBean(RetryTopicInternalBeanNames.DEAD_LETTER_PUBLISHING_RECOVERER_FACTORY_BEAN_NAME,
+					DeadLetterPublishingRecovererFactory.class);
+		this.clock = beanFactory.getBean(RetryTopicInternalBeanNames.INTERNAL_BACKOFF_CLOCK_BEAN_NAME,
+				Clock.class);
 	}
 
 	private class RetryTopicListenerContainerFactoryDecorator
--- a/spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
@@ -449,11 +449,8 @@ Optionally, you can provide your own back off policy.
 [source, java]
 ----
 @Bean(name = RetryTopicInternalBeanNames.LISTENER_CONTAINER_FACTORY_CONFIGURER_NAME)
-public ListenerContainerFactoryConfigurer lcfc(KafkaConsumerBackoffManager kafkaConsumerBackoffManager,
-                                               DeadLetterPublishingRecovererFactory deadLetterPublishingRecovererFactory,
-                                               @Qualifier(RetryTopicInternalBeanNames
-                                                       .INTERNAL_BACKOFF_CLOCK_BEAN_NAME) Clock clock) {
-    ListenerContainerFactoryConfigurer lcfc = new ListenerContainerFactoryConfigurer(kafkaConsumerBackoffManager, deadLetterPublishingRecovererFactory, clock);
+public ListenerContainerFactoryConfigurer lcfc() {
+    ListenerContainerFactoryConfigurer lcfc = new ListenerContainerFactoryConfigurer();
     lcfc.setBlockingRetryableExceptions(MyBlockingRetryException.class, MyOtherBlockingRetryException.class);
     lcfc.setBlockingRetriesBackOff(new FixedBackOff(500, 5)); // Optional
     return lcfc;
@@ -478,11 +475,8 @@ Here's an example with both configurations working together:
 [source, java]
 ----
 @Bean(name = RetryTopicInternalBeanNames.LISTENER_CONTAINER_FACTORY_CONFIGURER_NAME)
-public ListenerContainerFactoryConfigurer lcfc(KafkaConsumerBackoffManager kafkaConsumerBackoffManager,
-                                            DeadLetterPublishingRecovererFactory deadLetterPublishingRecovererFactory,
-                                            @Qualifier(RetryTopicInternalBeanNames
-                                                    .INTERNAL_BACKOFF_CLOCK_BEAN_NAME) Clock clock) {
-    ListenerContainerFactoryConfigurer lcfc = new ListenerContainerFactoryConfigurer(kafkaConsumerBackoffManager, deadLetterPublishingRecovererFactory, clock);
+public ListenerContainerFactoryConfigurer lcfc() {
+    ListenerContainerFactoryConfigurer lcfc = new ListenerContainerFactoryConfigurer();
     lcfc.setBlockingRetryableExceptions(ShouldRetryOnlyBlockingException.class, ShouldRetryViaBothException.class);
     return lcfc;
 }
--- a/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicExceptionRoutingIntegrationTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicExceptionRoutingIntegrationTests.java
@@ -52,7 +52,6 @@ import org.springframework.kafka.core.KafkaAdmin;
 import org.springframework.kafka.core.KafkaTemplate;
 import org.springframework.kafka.core.ProducerFactory;
 import org.springframework.kafka.listener.ContainerProperties;
-import org.springframework.kafka.listener.KafkaConsumerBackoffManager;
 import org.springframework.kafka.support.KafkaHeaders;
 import org.springframework.kafka.support.converter.ConversionException;
 import org.springframework.kafka.test.EmbeddedKafkaBroker;
@@ -383,11 +382,8 @@ public class RetryTopicExceptionRoutingIntegrationTests {
 		}
 
 		@Bean(name = RetryTopicInternalBeanNames.LISTENER_CONTAINER_FACTORY_CONFIGURER_NAME)
-		public ListenerContainerFactoryConfigurer lcfc(KafkaConsumerBackoffManager kafkaConsumerBackoffManager,
-													DeadLetterPublishingRecovererFactory deadLetterPublishingRecovererFactory,
-													@Qualifier(RetryTopicInternalBeanNames
-															.INTERNAL_BACKOFF_CLOCK_BEAN_NAME) Clock clock) {
-			ListenerContainerFactoryConfigurer lcfc = new ListenerContainerFactoryConfigurer(kafkaConsumerBackoffManager, deadLetterPublishingRecovererFactory, clock);
+		public ListenerContainerFactoryConfigurer lcfc() {
+			ListenerContainerFactoryConfigurer lcfc = new ListenerContainerFactoryConfigurer();
 
 			lcfc.setBlockingRetriesBackOff(new FixedBackOff(50, 3));
 			lcfc.setBlockingRetryableExceptions(ShouldRetryOnlyBlockingException.class, ShouldRetryViaBothException.class);


tomazfernandes:
As a note, I'm aware that the previous constructor solution didn't qualify the `KafkaConsumerBackoffManager` and `DeadLetterPublishingRecovererFactory` beans by name, and the proposed solution does. 

The `RetryTopicBootstrapper` component registers those components if there are none with those names, so if by any chance the user provided a bean with a different name, it would throw a `NoUniqueBeanDefinitionException` when trying to autowire the constructor.

So it seems to me this doesn't break backwards compatibility, and it makes the solution more robust overall.

Please let me know if that's not the case.

Thanks again.

tomazfernandes:
As another note, this would also make it easier for users to add LCFC's `errorHandlerCustomizer` and `containerCustomizer`.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
```diff
@@ -318,6 +325,18 @@ private long applyLimits(long pollTimeoutValue) {
...
 
+	@Override
+	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
+		this.kafkaConsumerBackoffManager =
+				beanFactory.getBean(RetryTopicInternalBeanNames.KAFKA_CONSUMER_BACKOFF_MANAGER,
+						KafkaConsumerBackoffManager.class);
+		this.deadLetterPublishingRecovererFactory =
+				beanFactory.getBean(RetryTopicInternalBeanNames.DEAD_LETTER_PUBLISHING_RECOVERER_FACTORY_BEAN_NAME,
+					DeadLetterPublishingRecovererFactory.class);
+		this.clock = beanFactory.getBean(RetryTopicInternalBeanNames.INTERNAL_BACKOFF_CLOCK_BEAN_NAME,
```
artembilan, line 336: You cannot do ` beanFactory.getBean()` in the `setBeanFactory()`. That's too early and may cause "Inversion of Control impedance" syndrome.
I know: my invention 😉 , but still: this is a common explanation for all the problems when try to access to `BeanFactory` too early.

You probably may look into the `SmartInitializingSingleton` instead. However I don't think it makes sense to have these beans in such an intermediary at all.
Moreover: such a `Configurer` abstraction give me a clue that these options should be present as setters if we really would like to have a customization for the target component.

I might miss something though...

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
tomazfernandes: Hi @artembilan, thanks a lot for looking into this.

> You cannot do ` beanFactory.getBean()` in the `setBeanFactory()`. That's too early and may cause "Inversion of Control impedance" syndrome. I know: my invention 😉 , but still: this is a common explanation for all the problems when try to access to `BeanFactory` too early.
> 
> You probably may look into the `SmartInitializingSingleton` instead. 

Oh, makes perfect sense, thanks. I felt I was missing something, just couldn't really figure out what 😄

> However I don't think it makes sense to have these beans in such an intermediary at all. Moreover: such a `Configurer` abstraction give me a clue that these options should be present as setters if we really would like to have a customization for the target component.
> 
> I might miss something though...

Well, to be honest, I didn't use `Configurer` for this name as the `Spring` standard does, which if I understand correctly is usually an interface or class with which end users can configure framework components.

Let me try to give you some more context on this. Hopefully it's not too much information, and let me know if you have any questions.

### [ListenerContainerFactoryConfigurer](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java)

What this class actually does is configure (hence the name 😄) the `ListenerContainerFactory` that was resolved for the endpoint for the `Retry Topic` feature. Actually, currently it returns a decorator wrapper to the factory that configures the container directly.

https://github.com/spring-projects/spring-kafka/blob/6022835e6c6b7b2d27e9f631fd77844360fbed0b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java#L341-L348

### [DeadLetterPublishingRecovererFactory](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java)

This dependency, as the names implies hopefully correctly this time, creates the `DLPR` instance that will be set by the `LCFC` in the container's `DefaultErrorHandler`:

https://github.com/spring-projects/spring-kafka/blob/6022835e6c6b7b2d27e9f631fd77844360fbed0b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java#L124-L128

### [DestinationTopic](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopic.java) and [DestinationTopicResolver](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopicResolver.java)

In order to resolve the `DestinationTopic` the `DLPR` will send the record to, it needs the `DestinationTopicResolver` bean, which is also a singleton. So while maybe `DLPRF` could not be a singleton, it'd still probably need de `DTR` bean.

https://github.com/spring-projects/spring-kafka/blob/6022835e6c6b7b2d27e9f631fd77844360fbed0b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java#L176-L191

### [KafkaConsumerBackOffManager](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaConsumerBackoffManager.java)

This is a singleton that manages backing off all of the application's partitions, and goes in the [KafkaBackoffAwareMessageListenerAdapter](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java) set in the container.

https://github.com/spring-projects/spring-kafka/blob/6022835e6c6b7b2d27e9f631fd77844360fbed0b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java#L254-L267

### Singletons

One advantage of using singletons for this functions as that we can provide access to the centralized information via autowiring of these components, such as requested in #2172 for the `DestinationTopicResolver` (which is also the reason for my other open PR), and we might provide the same kind of information for the `KafkaConsumerBackOffManager` as well, for example letting the user know all the currently paused partitions and their due times.

Another approach we could look into would be having specific `KafkaConsumerBackOffManager` and `DestinationTopic` instances for each endpoint, and registering them in some kind of `Registry` beans that we'd give users access to. I could look into that if you think it's worth it, probably for 3.0.0.

artembilan: Yes, I think the `Registry` abstraction is a good direction for goals you explain.

I see so much breaking changes that I don't understand why you still doubt that this can be somehow merged into a version earlier than `3.0`...

tomazfernandes: Sure, what I meant is we probably don't want to make such changes for 2.x, but of course we might try to do that in a backwards compatible manner. Not that I think it's a good idea to try that 😄 

I'll take note of this architectural change and begin thinking about how to implement it.

As far as this PR goes, I think we might consider just making the `SmartInitializingSingleton` change, what do you think? Do you have any other concerns?

artembilan: Not clear yet why would we need to initialize something on behalf of end-user if he/she already provide that config.
So, or we register one ourselves if no one already in the bean factory, or we fully back off just because there is one from end-user.
Or, as I said once you in other places: we can suggest something like `WebMvcConfigurationSupport` where we can suggest to override only those beans which are in end-user interest and do the rest of infrastructure because, well, we have to.

tomazfernandes: The `WebMvcConfigurationSupport` looks interesting, I'll look further into how we'd be able to create something similar, with `Helper` classes.

I'm not sure I understand what you mean by initializing something on behalf of end-user.

Currently, we do beans registration and some components initialization in the [RetryTopicBootstrapper](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicBootstrapper.java) class.

It includes logic where we check for existing user registered beans before registering our own:

https://github.com/spring-projects/spring-kafka/blob/6022835e6c6b7b2d27e9f631fd77844360fbed0b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicBootstrapper.java#L82-L83

So users can provide their own beans if they want to override ours, both the `ListenerContainerFactoryConfigurer` and its dependencies.

We'd still need the beans to be injected somehow though, and we probably don't want to burden users with that since it's for internal use - if they want to configure this class' dependencies they should do so by providing their own beans for those. 

So even with a `WebMvcConfigurationSupport` style configuration, we'd probably need to initialize them somehow, or am I missing something?

Thanks

tomazfernandes: I've drafted how I imagine this would look based on `WebMvcConfigurationSupport` and `WebMvcConfigurer` with basically the same feature set as there was in that builder. I didn't use the IDE so it's pseudo-code-like.

Here's how the `Configurer` looks:

```java
public interface RetryTopicFeaturesConfigurer {

    default void configureBlockingExceptions(BlockingExceptionsConfigurer configurer) {
    }

    default void configureNonBlockingExceptions(NonBlockingExceptionsConfigurer configurer) {
    }

    @Nullable
    default Consumer<CommonErrorHandler> customizeErrorHandler() {
        return null;
    }

    @Nullable
    default Consumer<MessageListenerContainer> customizeListenerContainer() {
        return null;
    }

    @Nullable
    default Consumer<DeadLetterPublihingRecoverer> customizeDLPR() {
        return null;
    }

    default KafkaException.LogLevel setKafkaBackOffExceptionLogLevel() {
        return KafkaException.LogLevel.DEBUG;
    }
}
```

Users might implement this interface in a `@Component` bean and we'd fetch and apply it in the `RetryTopicBootstrapper` class, or create a class specifically for this. Feature's infrastructure components would still be conditionally registered only if the `RetryTopic` feature is used.

I like this approach because I think it looks simple enough for users to focus on the features, and most users will never  need to know about the actual components or any of that complexity.

Also, looks like it has room to grow as we add new features.

tomazfernandes: Now here's the same feature set with the `WebMvcConfigurationSupport` approach. This doesn't have all of the feature components, only the ones necessary for these features.

On one hand, it's obviously much more code and more complex than the other approach. Users have to have an idea of what to look for and the component that's responsible for the feature they want to configure.

On the other hand, everything is right there - no mystery or behind the curtain action. Might make it easier for them to know of all components and its configuration possibilities, such as the internals of `KafkaBackOffManager` and the `TimingAdjuster`.

```java
public class RetryTopicConfigurationSupport {

    @Bean(name = RetryTopicInternalBeanNames.LISTENER_CONTAINER_FACTORY_CONFIGURER_BEAN_NAME)
        ListenerContainerFactoryConfigurer lcfc =
                new ListenerContainerFactoryConfigurer(kafkaConsumerBackoffManager, deadLetterPublishingRecovererFactory, clock);

        lcfc.setBlockingRetryableExceptions(getBlockingRetryableExceptions());
        lcfc.setBlockingRetriesBackOff(getBlockingRetriesBackOff());
        lcfc.setErrorHandlerCustomizer(customizeErrorHandler())
        lcfc.setContainerCustomizer(customizeContainer());
        customizeListenerContainerFactory().accept(lcfc);
        return lcfc;
    }

    protected Collection<Class<? extends Exception>> getBlockingRetryableExceptions() {
        return Collections.emptyList();
    }

    protected BackOff getBlockingRetriesBackOff() {
        return new FixedBackOff(3, 1000L);
    }

    protected Consumer<CommonErrorHandler> customizeErrorHandler() {
        return ceh -> {};
    }

    protected Consumer<MessageListenerContainer> customizeContainer() {
        return container -> {};
    }

    @Bean(name = RetryTopicInternalBeanNames.DEAD_LETTER_PUBLISHING_RECOVERER_FACTORY_BEAN_NAME)
    public DeadLetterPublishingRecovererFactory dlprf(DestinationTopicResolver destinationTopicResolver) {

        DeadLetterPublishingRecovererFactory dlprf =
                new DeadLetterPublishingRecovererFactory(destinationTopicResolver);
        dlprf.setRecovererCustomizer(customizeDeadLetterPublishingRecoverer())
        return dlprf;
    }

    protected Consumer<DeadLetterPublishingRecoverer> customizeDeadLetterPublishingRecoverer() {
        return recoverer -> {};
    }

    @Bean(name = RetryTopicInternalBeanNames.DESTINATION_TOPIC_RESOLVER_BEAN_NAME)
    public DestinationTopicResolver destinationTopicResolver(@Qualifier(RetryTopicInternalBeanNames
											.INTERNAL_BACKOFF_CLOCK_BEAN_NAME Clock clock, ApplicationContext context) {

        DestinationTopicResolver ddtr =
                new DefaultDestinationTopicResolver(clock, context);
        ddtr.addNotRetryableExceptions(addNonBlockingFatalExceptions());
        ddtr.removeNotRetryableException(removedNonBlockingFatalExceptions());
        return dlprf;
    }

    protected Collection<Class<? extends Exception>> addNonBlockingFatalExceptions() {
        return Collections.emptyList();
    }

    protected Collection<Class<? extends Exception>> removedNonBlockingFatalExceptions() {
        return Collections.emptyList();
    }

    @Bean(name = RetryTopicInternalBeanNames.INTERNAL_BACKOFF_CLOCK_BEAN_NAME)
    public Clock defaultClock() {
        return Clock.systemUTC();
    }

    @Bean(name = RetryTopicInternalBeanNames.KAFKA_BACKOFF_MANAGER_BEAN_NAME)
    public KafkaConsumerBackoffManager kafkaConsumerBackoffManager(@Qualifier(RetryTopicInternalBeanNames.INTERNAL_KAFKA_CONSUMER_BACKOFF_MANAGER_FACTORY
                    KafkaBackOffManagerFactory kafkaBackOffManagerFactory) {
		if (PartitionPausingBackOffManagerFactory.class.isAssignableFrom(factory.getClass())) {
			((PartitionPausingBackOffManagerFactory) factory).setTimingAdjustmentManager(timingAdjuster());
		}
		return factory.create();
	}

    @Bean(name = RetryTopicInternalBeanNames.INTERNAL_KAFKA_CONSUMER_BACKOFF_MANAGER_FACTORY)
    public KafkaConsumerBackoffManagerFactory kafkaConsumerBackoffManagerFactory(@Qualifier(RetryTopicInternalBeanNames.INTERNAL_BACKOFF_TIMING_ADJUSTMENT_MANAGER
                    KafkaConsumerTimingAdjuster timingAdjuster) {
 		return new PartitionPausingBackOffManagerFactory(timingAdjuster);
	}

    @Bean(name = RetryTopicInternalBeanNames.BACKOFF_TASK_EXECUTOR)
    public TaskExecutor getTaskExecutor() {
		ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
		executor.initialize();
		applicationContext.addApplicationListener((ApplicationListener<ContextClosedEvent>) event -> executor.shutdown());
        return executor;
    }

    @Bean(name = RetryTopicInternalBeanNames.INTERNAL_BACKOFF_TIMING_ADJUSTMENT_MANAGER)
    public KafkaConsumerTimingAdjuster timingAdjuster(@Qualifier(RetryTopicInternalBeanNames.BACKOFF_TASK_EXECUTOR) TaskExecutor taskExecutor) {
        return new WakingKafkaConsumerTimingAdjuster(taskExecutor);
    }

}

```

tomazfernandes: Another option would be having this `Support` approach, and also have a feature-based `Options` approach as you suggested in the other issue, such as:

```java
RetryTopicOptions
   .blockingRetries()
   .retryOn(MyException.class, MyOtherException.class)
   .backOffPolicy(new FixedBackOff(3, 1000L))
```

And:

```java
RetryTopicOptions
   .nonBlockingRetries()
   .removeFatalExceptions(ConversionException.class)
   .addFatalExceptions(MyFatalException.class)
```

This way most users might be able to use only the `Options` approach, but any user that needs to go further would be able to see the components instantiation and have more capable customization options via extension.

artembilan: @tomazfernandes ,

Do you wait for something else from us for this feature?

I see by your last comment that `RetryTopicConfigurationSupport` is getting appealing with you, so we really are heading to the same page 😉 .

I'm not sure if @garyrussell has a minute to look into this, but perhaps yo can go ahead and push a real commit for whatever you believe is possible.

Thanks.

No rush, of course.

tomazfernandes: Sure, I can do that. My main concern about this is that the `RetryTopicConfigurationSupport` seems a bit overwhelming for regular end-users - that's why I thought of also having this `RetryTopicOptions` as a feature-oriented configuration system to go along.

I'll make a commit with this and we can take a better look.

I'd really like to have an improvement for the RetryTopic configuration for `2.8.5`- I think we're not in a good place right now. So I'll try to have this ready today or by tomorrow at most.

Thanks.

artembilan: I don't think it good to consider this for `2.8.x`. I only see some drastic breaking changes.
Not sure what the problem you see with the `extends RetryTopicConfigurationSupport` for end-user `@Configuration`.
With good Javadocs for those `protected confitgure...` methods it should be very clear what we are expecting from end-users and what they may left relying on our default preferences.
Yes, options might be good, but I don't think they have to be addressed in this PR some how...

tomazfernandes: I was thinking we could leave current configuration as is and add this `Support` configuration as an optional add-on. This way it shouldn't be too risky since users could just not use it. Current bootstrapping should back off if that is used instead. Of course, that's in theory, we would need to see in practice if that works as expected. But maybe I'm missing something.

As a user, I think I'd like to have a very clear overview of the main features, with a very friendly API to configure them. If users extend the `Support` class, they'd have to look at all public `@Bean` methods - there will be a lot of them - and then figure out what `configure` method to override for what they want.

I think that's probably overwhelming, and in contrast with the simple `@RetryableTopic` and `RetryTopicConfigurationBuilder` APIs - but maybe I will have a clearer view when I have a working example.

Anyway, if this doesn't look like a good fit for `2.8.x`, I think we should come up with something simpler that would be back portable. Users currently have to do some really [bad workarounds](https://stackoverflow.com/questions/71705876/combining-blocking-and-non-blocking-retries-in-spring-kafka) to configure the features for now.

tomazfernandes: Of course, that simpler solution could be something we might replace for a more robust solution later.

artembilan: Right. And for that reason Spring MVC has `@EnableWebMvc` with the `@Import(DelegatingWebMvcConfiguration.class)` where that is like `DelegatingWebMvcConfiguration extends WebMvcConfigurationSupport`.

tomazfernandes: Yeah, I thought of having a `@EnableRetryTopic` annotation so we don't have to guess whether users want the feature or not. Not sure if that would mean having to add logic to `Boot` though or if we'd be ok handling that on Spring Kafka.

Either way, I'm not sure if maybe it's better not to bother users with that and solve this in a way that they don't have to take any extra steps to just use the feature.

artembilan: Well, that probably could be addressed on the `@EnableKafka` with that `@Import(DefaultRetryTopicConfigurationSupport.class)` even if such a feature won't be used at runtime: just a couple infrastructure beans!

tomazfernandes: Well, if you look into the sample I provided, you’ll notice that that’s more like about a dozen beans 😄 :

https://github.com/spring-projects/spring-Kafka/pull/2204#discussion_r842230445

And that’s not all the beans… I can try to trim it down, but that’s the current scenario.

That’s why I also looked into a ‘configurer’ abstraction:

https://github.com/spring-projects/spring-kafka/pull/2204#discussion_r842228656

Anyway, I can work on the ‘support’ plan, have a commit or perhaps a separate PR, and then we can see if that looks like the best option.

tomazfernandes: Well, I guess this is where we've been discussing this 😄

https://github.com/spring-projects/spring-kafka/pull/2213#discussion_r846372610

Anyway, I'll work on this.

Thanks!

tomazfernandes:
Closed in favor of #2226.

Issue #2195, opened 2022-03-28
Avoid unnecessary partition seeks on successful record recovery


v-chernyshev:
I've reordered the sections for this request because, in my opinion, it is easier to understand the desired behaviour when the limitations of the current approach are explained first.

**Current Behavior**

Let's consider a situation where `max.poll.records` (500 by default) are available for consumption in a non-transactional record listener container, but the very first record cannot be processed immediately due to e.g. an issue with an external data provider. Here is what happens in the code in case of, at the very least, non-blocking retries:

- `KafkaMessageListenerContainer.ListenerConsumer.doInvokeRecordListener` is eventually invoked with `record` pointing to the first record and `iterator` holding the remaining 499.
- `this.invokeOnMessage` fails with an exception.
- An error handler is installed, so `this.invokeErrorHandler` is called.
- `this.commonErrorHandler.remainingRecords()` returns `true`. Both the failed record and all the remaining records from the iterator are drained into a temporary array, `this.commonErrorHandler.handleRemaining` leads to `DefaultErrorHandler.handleRemaining` that performs `SeekUtils.seekOrRecover`.
- An attempt is made to recover the failed record.
- No matter whether the recovery operation succeeds or not, the rest of the batch is rewound in `seekPartitions`.

Unfortunately, this unconditional rewind operation leads to enormous spikes in network I/O as pretty much the same records are requested from Kafka over and over again. To be precise, with a local Kafka broker and a test application that always triggers a recovery I could see spikes exceeding 2000 Mbps in `iptraf`.

**Expected Behavior**

I believe that it is not necessary to always rewind the remaining 499 records if the recovery is successful. This is what may happen with the failed record when non-blocking retries are enabled:

- If the record cannot be processed due to a fatal exception (e.g. `ClassCastException`) then it is sent straight into the DLT topic.
- If the exception is not fatal, then the record is sent into the next retry destination.

Both these cases allow the next `iterator` record to be processed immediately as if nothing happened. This very same record will be at the front of the batch during the next `poll` invocation anyway! The only special case I can think of is `KafkaBackoffException`, which does absolutely require rewinding all the offsets so that the same batch is consumed again when the affected partitions are resumed.

**Context**

The production rollout of our service that uses the non-blocking retries feature provided by Spring Kafka triggered a number of network I/O alarms, so we decided to find the root cause. This feature request is the result of the investigation :)

[...]

PR #2207, opened 2022-04-04
GH-2195: DefaultErrorHandler Improvements

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2195

Add an option to avoid seeks after handling exceptions.

Instead, pause the consumer for one `poll()` and use the remaining records as the
result of that poll.

New methods on `CommonErrorHandler` - `handleOne` for record listeners, returning
a boolean to indicate whether the record was recovered and should not be redelivered.

`handlaBatchAndReturnRemaining` for batch listeners, returning either the complete
set or a subset, e.g. when the `DEH` receives a `BatchListenerExecutionFailedException`
and commits a partial batch.

Also includes the classifier refactoring discussed here
https://github.com/spring-projects/spring-kafka/pull/2185#issuecomment-1078259823

The new logic is disabled by default, we can consider enabling it in 3.0 and remove
the deprecations.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
@@ -51,7 +51,7 @@ public interface CommonErrorHandler extends DeliveryAttemptAware {
 	}
 
 	/**
-	 * Return true if this error handle performs seeks on the failed record and remaining
+	 * Return true if this error handler performs seeks on the failed record and remaining
 	 * records (or just the remaining records after a failed record is recovered).
 	 * @return true if the next poll should fetch records.
 	 */
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
@@ -111,6 +111,7 @@ public interface CommonErrorHandler extends DeliveryAttemptAware {
 	 * @param container the container.
 	 * @return true if the error was "handled" or false if not and the container will
 	 * re-submit the record to the listener.
+	 * @since 2.9
 	 * @see #remainingRecords()
 	 */
 	@SuppressWarnings("deprecation")
@@ -171,7 +172,7 @@ public interface CommonErrorHandler extends DeliveryAttemptAware {
 	 * @param <K> the key type.
 	 * @param <V> the value type.
 	 * @return the consumer records, or a subset.
-	 * @since 2.8.5
+	 * @since 2.9
 	 */
 	default <K, V> ConsumerRecords<K, V> handleBatchAndReturnRemaining(Exception thrownException,
 			ConsumerRecords<?, ?> data, Consumer<?, ?> consumer, MessageListenerContainer container,
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 the original author or authors.
+ * Copyright 2021-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -41,11 +41,23 @@ public interface CommonErrorHandler extends DeliveryAttemptAware {
 	 * When true (default), all remaining records including the failed record are passed
 	 * to the error handler.
 	 * @return false to receive only the failed record.
+	 * @deprecated in favor of {@link #seeksAfterHandling()}.
 	 * @see #handleRecord(Exception, ConsumerRecord, Consumer, MessageListenerContainer)
 	 * @see #handleRemaining(Exception, List, Consumer, MessageListenerContainer)
 	 */
+	@Deprecated
 	default boolean remainingRecords() {
 		return false;
+	}
+
+	/**
+	 * Return true if this error handle performs seeks on the failed record and remaining
+	 * records (or just the remaining records after a failed record is recovered).
+	 * @return true if the next poll should fetch records.
+	 */
+	@SuppressWarnings("deprecation")
+	default boolean seeksAfterHandling() {
+		return remainingRecords();
 	}
 
 	/**
@@ -79,12 +91,39 @@ public interface CommonErrorHandler extends DeliveryAttemptAware {
 	 * @param record the record.
 	 * @param consumer the consumer.
 	 * @param container the container.
+	 * @deprecated in favor of
+	 * {@link #handleOne(Exception, ConsumerRecord, Consumer, MessageListenerContainer)}.
 	 * @see #remainingRecords()
 	 */
+	@Deprecated
 	default void handleRecord(Exception thrownException, ConsumerRecord<?, ?> record, Consumer<?, ?> consumer,
 			MessageListenerContainer container) {
 
 		LogFactory.getLog(getClass()).error("'handleRecord' is not implemented by this handler", thrownException);
+	}
+
+	/**
+	 * Handle the exception for a record listener when {@link #remainingRecords()} returns
+	 * false. Use this to handle just the single failed record.
+	 * @param thrownException the exception.
+	 * @param record the record.
+	 * @param consumer the consumer.
+	 * @param container the container.
+	 * @return true if the error was "handled" or false if not and the container will
+	 * re-submit the record to the listener.
+	 * @see #remainingRecords()
+	 */
+	@SuppressWarnings("deprecation")
+	default boolean handleOne(Exception thrownException, ConsumerRecord<?, ?> record, Consumer<?, ?> consumer,
+			MessageListenerContainer container) {
+
+		try {
+			handleRecord(thrownException, record, consumer, container);
+			return true;
+		}
+		catch (Exception ex) {
+			return false;
+		}
 	}
 
 	/**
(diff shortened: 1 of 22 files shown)


artembilan (changes requested):
This looks for me too bug to be considered for back-porting.
Why just not aim the fix for `3.0` and feel free from possible breaking changes?

Thanks

garyrussell:
I agree; maybe this is the final driver to create a 2.9 branch; which we can release sooner (May).

There is a big performance problem when using the retryable topic - see the discussions on the  issue,

tomazfernandes:
@garyrussell, I began looking into this today and should continue tomorrow - there's a lot going on in parts of the code I'm not that familiar with. Overall I think the solution looks great, just want to get a better grasp at it.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
```diff
@@ -41,13 +41,25 @@
...
 	 * @see #handleRecord(Exception, ConsumerRecord, Consumer, MessageListenerContainer)
 	 * @see #handleRemaining(Exception, List, Consumer, MessageListenerContainer)
 	 */
+	@Deprecated
 	default boolean remainingRecords() {
 		return false;
 	}
 
+	/**
+	 * Return true if this error handle performs seeks on the failed record and remaining
```
artembilan, line 54: Typo:

```suggestion
	 * Return true if this error handler performs seeks on the failed record and remaining
```

and `@since 2.9`

spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
```diff
@@ -133,8 +133,9 @@ private void checkConfig() {
 	public void process(List<ConsumerRecord<K, V>> records, Consumer<K, V> consumer,
 			@Nullable MessageListenerContainer container, Exception exception, boolean recoverable, EOSMode eosMode) {
 
-		if (SeekUtils.doSeeks(((List) records), consumer, exception, recoverable,
-				getRecoveryStrategy((List) records, exception), container, this.logger)
+		List records2 = records;
```
artembilan, line 136: Why do we need this variable?
It is going to point to the same `List` object anyway, so mutation on it is going to affect all the variables pointing to it.

spring-kafka/src/test/java/org/springframework/kafka/listener/DefaultErrorHandlerNoSeeksBatchAckTests.java
```diff
@@ -0,0 +1,219 @@
+/*
+ * Copyright 2017-2021 the original author or authors.
```
artembilan, line 2: Current year, please:
```suggestion
 * Copyright 2022 the original author or authors.
```

artembilan (changes requested):
Do we need to add anything to the docs for this change?

Thanks

artembilan:
So, is this OK now for merging?

garyrussell:
Let @tomazfernandes have a few more days to review; maybe merge on Monday?

tomazfernandes:
That would work nicely for me, thanks!

garyrussell:
We still need to seek on the retry containers because the delay is enabled by pausing the partitions.

I guess you should only configure this on the main container and use a small `max.poll.records` on the retry containers.

garyrussell:
It would add a little more complexity (in the container), but maybe we could come up with a hybrid approach - perform the seeks if listener throws a `KafkaBackOffException` and use the new approach for all other exceptions.

But that can be another PR.

tomazfernandes:
I think the ideal scenario there would be, even for retry containers: if we process a record, let's say it fails and goes through recovery logic, which succeeds (message sent to next topic) . If the next record for that container does not trigger a backoff (it's already past its due time), we shouldn't need to seek again. Makes sense?

If it does backoff, recovery does not succeed, and the BackOffManager pauses the container before any of this logic is executed. I'm not sure where in the PR the partition is resumed, but if it does not resume when it didn't pause, and don't serve the failed record while paused, maybe we're good?

tomazfernandes:
I think the trick here is understanding the interaction of this new logic with partition pausing.

If we do seek in case of a `KafkaBackOffException`, we shouldn't receive any records from the paused partition until it's resumed by the `BackOffManager`, since resuming the whole container explicitly leaves paused partitions alone. So we're basically polling for records that have nothing to do with the KBE itself.

But we might consider filtering out any records that are from a currently paused partition before serving them again, this way we don't need to seek all partitions due to one backing off.

tomazfernandes:
So summing up, I think that maybe rather than "coupling" the container to the `KafkaBackOffException`, we should probably figure out how it should work when handling paused partitions, and just let `KafkaBackOffManager` do its work.

Sorry I kind of `thought out loud` - still figuring out how these changes work 😄

tomazfernandes:
Just to remember, the `BackOffManager` pauses the partition before throwing the `KafkaBackOffException`, and resumes it after receiving a `PartitionIdleEvent`, if time is due.

garyrussell:
This clearly needs some more thought; so definitely not in the scope of this PR.

tomazfernandes:
Sure! I was really just thinking out loud, sorry about that. I took a better look at these changes yesterday and they look good to me, sure will increase performance a lot when dealing with errors, and also are defensive enough that shouldn't be a problem when disabled.

I'm wondering, considering we'll be having milestone exposure to this, DYT maybe it's worth it to leave it on by default so we get more feedback? If there's any problem users can just disable it, and then we can release a new milestone with the fix. Maybe I'm missing something though - perhaps it's a too big behavior change for a minor release for example.

Thanks

garyrussell:
That's an interesting thought; but I think we need to resolve the issue with the retry containers first because, if it is used there, yes, we will still pause the relevant partition, but there may be additional records from that partition in the remaining records. We'd need to week those out somehow, and only seek the paused partition.

tomazfernandes:
Sure. Also, that might impact users pausing partitions themselves. But I'm sure we'll work this out soon enough. As far as retryable topics go, we could disable it in RT setup if necessary.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1495,6 +1501,10 @@ private void fixTxOffsetsIfNeeded() {
 			}
 			else {
 				records = pollConsumer();
+				if (this.pendingRecordsAfterError != null && !isPaused()) {
```
tomazfernandes, line 1504: I might be completely wrong about this - lots of moving parts here. But let's say that, after an error is handled and the `pendingRecordsAfterError` property is set with the remaining records, the container is set to `pause`. If I'm not mistaken, this is how it would go:

* the consumer is paused
* the next poll returns empty and the `pendingRecordsAfterError` property is not changed since `isPaused` is true
* let's say the container is resumed right after a poll
* the consumer is resumed
* since the consumer was resumed, `doPoll` returns the next batch of records
* since `isPaused` is false and `pendingRecordsAfterError` is not null, the records returned by the poll are replaced by the records in `pendingRecordsAfterError`
* now we're processing the batch we had before the container was paused, but the consumer position has advanced a batch further, and we have no reference to that batch of records

Makes any sense? Am I missing something?

If that's correct, it seems to me that the problem would be that we're relying on the consumer being paused while we poll and already have records, but the consumer might have been paused and resumed. Maybe we should not resume the consumer if we have pending records?

garyrussell: Good catch - I think it should behave as if there are no errors - in other words, don't check `isPaused()`.

Normally, `pause()` does not take effect until the records from the previous poll are all processed.

With my next commit we won't look at `isPaused()` either during the records replacement or in the pause logic itself. i.e. the user call to `pause()` will not take effect until the records from the previous poll are processed, regardless of any errors.

This is a behavior change, but it's a side effect of not doing the seeks on every error.

tomazfernandes: A couple more thoughts on this:

* Are we protected from the same issue in case the consumer resumes automatically after a rebalance?

* I wonder if, in order to simplify things a bit and make it more explicit, we might kill the `isPauseAfterError` flag, and just have different behavior whenever we have `pendingRecordsAfterError != null`. The only part where `pendingRecordsAfterError` is null is between poll and invokeListener, and we don't seem to be using this flag there.

Thanks.

garyrussell: >Are we protected from the same issue in case the consumer resumes automatically after a rebalance?

I am not sure what you mean by that; the paused state of any revoked partitions is reset in the `Consumer` so if they are re-assigned, they will not be paused; hence the existing logic to re-pause the entire container if it was paused based on `consumerPaused`. Since we re-pause the TPs within the `onPartitionsAssigned()`, that poll (which is calling the rebalance listener) won't return any records.

Yes, I agree, we can get rid of that boolean.

tomazfernandes: > > Are we protected from the same issue in case the consumer resumes automatically after a rebalance?
> 
> I am not sure what you mean by that; the paused state of any revoked partitions is reset in the `Consumer` so if they are re-assigned, they will not be paused; hence the existing logic to re-pause the entire container if it was paused based on `consumerPaused`. Since we re-pause the TPs within the `onPartitionsAssigned()`, that poll (which is calling the rebalance listener) won't return any records.
> 

Sure, you're right - I forgot we are setting `consumerPaused` to true, so in a rebalance the `RebalanceListener` will pause the consumer again.

I'm just thinking of any situations where the consumer might be resumed in such a way that we might get to `doPoll` with `pendingRecords` before it's paused again. Guess we're good.

I wonder if it might make sense to make an assertion that, if `pendingRecords` is not null and the consumer is not `paused` when we get to the poll, we'd throw an exception just to be on the safe side. But of course, that's mostly my ignorance on the ways the consumer might behave - we can leave that out.

Thanks.

garyrussell: I can't think of a scenario where that could happen, but I can add an emergency container stop if the poll returns any records.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1333,12 +1337,14 @@ protected void pollAndInvoke() {
 				return;
 			}
 			debugRecords(records);
-			resumeConsumerIfNeccessary();
-			if (!this.consumerPaused) {
-				resumePartitionsIfNecessary();
-			}
 
 			invokeIfHaveRecords(records);
+			if (this.pendingRecordsAfterError == null) {
```
tomazfernandes, line 1342: Just a thought. Since we're not committed to listening to `pause`/ `resume` calls while in this logic, I wonder if maybe we should also ignore it for partition pausing / resuming, and instead handle back offs relying on the `KafkaBackOffException` itself. The exception contains information such as the `TopicPartition` that was backed off. Of course, this or any other solution can be part of a different PR.

Thanks.

artembilan (approved):
LGTM.

@tomazfernandes , anything else in your mind @garyrussell has to fix, yet?

Thanks

tomazfernandes:
> LGTM.
> 
> @tomazfernandes , anything else in your mind @garyrussell has to fix, yet?
> 
> Thanks

Not really, at this point I'm more nit picking about implementation details 😄 I think this looks great and will surely be a noticeable gain in performance regarding error handling.

This was a really interesting journey into this logic, which is maybe the framework's heart, so thanks a lot @garyrussell for the opportunity.

Thanks.

artembilan:
Sorry, this does not back-port to `2.9.x` clean.

Thanks

Issue #2208, opened 2022-04-05
Fix Manual Nack with Mutating RecordInterceptor


garyrussell:
https://stackoverflow.com/questions/71755448/kafkalistener-skipping-messages-when-using-acknowledgment-nack-after-recordin/71755711#71755711

PR #2210, opened 2022-04-05
GH-2208: Fix Manual Nack with Mutating Interceptor

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2208

`equals()` test on `ConsumerRecord` fails (not implemented).

Compare topic, partition and offset instead.

**cherry-pick to 2.9.x, 2.8.x, 2.7.x**


artembilan:
... and cherry-picked to all the supported branches.

Issue #2212, opened 2022-04-05
Original exception never gets logged in Retry Topic flow


tomazfernandes:
This [SO](https://stackoverflow.com/questions/71719065/backoffexceptions-are-logged-at-error-level-when-using-retrytopicconfiguration) user pointed out that, when the record goes through the Retry Topic flow, the original exception never gets logged.

We had already discussed this a bit earlier this year. 

@garyrussell [proposed](https://github.com/spring-projects/spring-kafka/issues/2063#issuecomment-1010404269_):

> But I see your point; since we always recover these failures, they are never logged. Maybe the DLPR needs an option to log the exception after successfully recovering a record (but that doesn't need to be part of this fix).

I suggested as an [alternative](https://github.com/spring-projects/spring-kafka/issues/2063#issuecomment-1010482981_):

> We could log the error message just before we either send the record to the DLT or give up processing - that way nothing gets logged if any of the retrials work, and gets logged only once if none.

The main difference seems to be whether we want to log the exception at every retry attempt or only before it goes to the DLT.

We might look into adding both and making these configurable - users might be able to set a flag in `DLPR`  to log after every attempt, and `DLPRF` to log before going to the DLT.

We could also have this as an enum in `DLPRF`, with something like `OriginalExceptionLoggingStrategy.NEVER`, `EVERY_ATTEMPT`, `AFTER_RETRIES_EXHAUSTED` - `DLPRF` would then set the appropriate flag in `DLPR`.

WDYT?

_Originally posted by @tomazfernandes in https://github.com/spring-projects/spring-kafka/discussions/2211_

tomazfernandes:
Looks like `Reference this discussion in new issue` is not the same as `converting discussion into issue` as you do 😄

Guess it's alright though - let me know if otherwise. I should open a PR for this shortly if that's ok.

Thanks.

garyrussell:
It doesn't really make a difference; this is fine.

PR #2213, opened 2022-04-05
GH-2212: Log listener exception in retry topic flow

tomazfernandes:
Resolves https://github.com/spring-projects/spring-kafka/issues/2212

As in normal `Retry Topic` flow recovery is always successful, the exception thrown by the listener is never logged.

Now such exception is logged at `DEBUG` level for intermediate retries and `ERROR` level when retries are exhausted.

Produces messages such as:

For intermediate retries:

```log
2022-04-05 20:37:36,713 DEBUG [org.springframework.kafka.KafkaListenerEndpointContainer#1-retry-500-0-C-1] [org.springframework.kafka.retrytopic.DeadLetterPublishingRecovererFactory] - Error processing record: topic = myRetryTopic2-retry-500, partition = 0, offset = 0, main topic = myRetryTopic2 at topic myRetryTopic2-retry-500. Sending to retry topic myRetryTopic2-retry-1000.
```

For messages heading to a DLT:

```log
2022-04-05 20:37:37,738 ERROR [org.springframework.kafka.KafkaListenerEndpointContainer#2-retry-1000-0-C-1] [org.springframework.kafka.retrytopic.DeadLetterPublishingRecovererFactory] - Record: topic = myRetryTopic2-retry-1000, partition = 0, offset = 0, main topic = myRetryTopic2 threw an error at topic myRetryTopic2-retry-1000 and won't be retried. Sending to DLT with name myRetryTopic2-dlt.
```

For messages that won't be forwarded, including those thrown in a DLT handler:

```log
2022-04-05 20:37:38,250 ERROR [org.springframework.kafka.KafkaListenerEndpointContainer#3-dlt-0-C-1] [org.springframework.kafka.retrytopic.DeadLetterPublishingRecovererFactory] - Record: topic = myRetryTopic2-dlt, partition = 0, offset = 0, main topic = myRetryTopic2 threw an error at topic myRetryTopic2-dlt and won't be retried. No further action will be taken with this record.
```

Also did some polishing in `TimestampedException` as it was cluttering the stacktrace with duplicated information from cause.getMessage(). Now it logs the time when the exception occurred.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2021 the original author or authors.
+ * Copyright 2018-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -34,14 +34,36 @@ public class TimestampedException extends KafkaException {
 
 	private final long timestamp;
 
+	@Deprecated
 	public TimestampedException(Exception ex, Clock clock) {
 		super(ex.getMessage(), ex);
 		this.timestamp = Instant.now(clock).toEpochMilli();
 	}
 
+	@Deprecated
 	public TimestampedException(Exception ex) {
 		super(ex.getMessage(), ex);
-		this.timestamp = Instant.now(Clock.systemDefaultZone()).toEpochMilli();
+		this.timestamp = Instant.now().toEpochMilli();
+	}
+
+	/**
+	 * Creates an instance with the timestamp of when it was thrown and its cause.
+	 * @param ex the exception cause.
+	 * @param timestamp the timestamp of when the exception was thrown.
+	 */
+	public TimestampedException(Exception ex, long timestamp) {
+		super("Exception thrown at " + Instant.ofEpochMilli(timestamp), ex);
+		this.timestamp = timestamp;
+	}
+
+	/**
+	 * Creates an instance with the timestamp of when it was thrown and its cause.
+	 * @param ex the exception cause.
+	 * @param now the Instant of when the exception was thrown.
+	 */
+	public TimestampedException(Exception ex, Instant now) {
+		super("Exception thrown at " + now, ex);
+		this.timestamp = now.toEpochMilli();
 	}
 
 	public long getTimestamp() {
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
@@ -53,7 +53,7 @@ public class TimestampedException extends KafkaException {
 	 * Creates an instance with the timestamp of when it was thrown and its cause.
 	 * @param ex the exception cause.
 	 * @param timestamp the millis from epoch of when the exception was thrown.
-	 * @since 2.8.5
+	 * @since 2.7.13
 	 */
 	public TimestampedException(Exception ex, long timestamp) {
 		super("Exception thrown at " + Instant.ofEpochMilli(timestamp), ex);
@@ -64,7 +64,7 @@ public class TimestampedException extends KafkaException {
 	 * Creates an instance with the Instant of when it was thrown and its cause.
 	 * @param ex the exception cause.
 	 * @param now the Instant of when the exception was thrown.
-	 * @since 2.8.5
+	 * @since 2.7.13
 	 */
 	public TimestampedException(Exception ex, Instant now) {
 		super("Exception thrown at " + now, ex);
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
@@ -40,16 +40,20 @@ public class TimestampedException extends KafkaException {
 		this.timestamp = Instant.now(clock).toEpochMilli();
 	}
 
-	@Deprecated
+	/**
+	 * Constructs an instance with the provided cause
+	 * and the current time.
+	 * @param ex the exception cause.
+	 */
 	public TimestampedException(Exception ex) {
-		super(ex.getMessage(), ex);
-		this.timestamp = Instant.now().toEpochMilli();
+		this(ex, Instant.now());
 	}
 
 	/**
 	 * Creates an instance with the timestamp of when it was thrown and its cause.
 	 * @param ex the exception cause.
-	 * @param timestamp the timestamp of when the exception was thrown.
+	 * @param timestamp the millis from epoch of when the exception was thrown.
+	 * @since 2.8.5
 	 */
 	public TimestampedException(Exception ex, long timestamp) {
 		super("Exception thrown at " + Instant.ofEpochMilli(timestamp), ex);
@@ -57,9 +61,10 @@ public class TimestampedException extends KafkaException {
 	}
 
 	/**
-	 * Creates an instance with the timestamp of when it was thrown and its cause.
+	 * Creates an instance with the Instant of when it was thrown and its cause.
 	 * @param ex the exception cause.
 	 * @param now the Instant of when the exception was thrown.
+	 * @since 2.8.5
 	 */
 	public TimestampedException(Exception ex, Instant now) {
 		super("Exception thrown at " + now, ex);
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
@@ -54,7 +54,8 @@ import org.springframework.util.Assert;
  */
 public class DeadLetterPublishingRecovererFactory {
 
-	private static final LogAccessor LOGGER = new LogAccessor(LogFactory.getLog(DeadLetterPublishingRecovererFactory.class));
+	private static final LogAccessor LOGGER =
+			new LogAccessor(LogFactory.getLog(DeadLetterPublishingRecovererFactory.class));
 
 	private final DestinationTopicResolver destinationTopicResolver;
 
@@ -65,6 +66,8 @@ public class DeadLetterPublishingRecovererFactory {
 	private Consumer<DeadLetterPublishingRecoverer> recovererCustomizer = recoverer -> { };
 
 	private BiFunction<ConsumerRecord<?, ?>, Exception, Headers> headersFunction;
+
+	private ListenerExceptionLoggingStrategy loggingStrategy = ListenerExceptionLoggingStrategy.AFTER_RETRIES_EXHAUSTED;
 
 	public DeadLetterPublishingRecovererFactory(DestinationTopicResolver destinationTopicResolver) {
 		this.destinationTopicResolver = destinationTopicResolver;
@@ -119,6 +122,22 @@ public class DeadLetterPublishingRecovererFactory {
 	 */
 	public boolean removeNotRetryableException(Class<? extends Exception> exceptionType) {
 		return this.nonFatalExceptions.add(exceptionType);
+	}
+
+	/**
+	 * Never logs the listener exception.
+	 * The default is logging only after retries are exhausted.
+	 */
+	public void neverLogListenerException() {
+		this.loggingStrategy = ListenerExceptionLoggingStrategy.NEVER;
+	}
+
+	/**
+	 * Logs the listener exception at each attempt.
+	 * The default is logging only after retries are exhausted.
+	 */
+	public void alwaysLogListenerException() {
+		this.loggingStrategy = ListenerExceptionLoggingStrategy.EACH_ATTEMPT;
 	}
 
 	@SuppressWarnings("unchecked")
@@ -185,33 +204,39 @@ public class DeadLetterPublishingRecovererFactory {
 				? "none"
 				: nextDestination.getDestinationName()));
 
-		logException(e, cr, nextDestination);
+		maybeLogListenerException(e, cr, nextDestination);
 
 		return nextDestination.isNoOpsTopic()
 					? null
 					: resolveTopicPartition(cr, nextDestination);
 	}
 
-	private void logException(Exception e, ConsumerRecord<?, ?> cr, DestinationTopic nextDestination) {
-		if (nextDestination.isDltTopic()) {
-			LOGGER.error(e, () -> getErrorMessage(cr)
+	private void maybeLogListenerException(Exception e, ConsumerRecord<?, ?> cr, DestinationTopic nextDestination) {
+		if (nextDestination.isDltTopic()
+				&& !ListenerExceptionLoggingStrategy.NEVER.equals(this.loggingStrategy)) {
+			LOGGER.error(e, () -> getErrorMessage(cr) + " and won't be retried. "
 					+ "Sending to DLT with name " + nextDestination.getDestinationName() + ".");
 		}
-		else if (nextDestination.isNoOpsTopic()) {
-			LOGGER.error(e, () -> getErrorMessage(cr)
+		else if (nextDestination.isNoOpsTopic()
+				&& !ListenerExceptionLoggingStrategy.NEVER.equals(this.loggingStrategy)) {
+			LOGGER.error(e, () -> getErrorMessage(cr) + " and won't be retried. "
 					+ "No further action will be taken with this record.");
 		}
+		else if (ListenerExceptionLoggingStrategy.EACH_ATTEMPT.equals(this.loggingStrategy)) {
+			LOGGER.error(e, () -> getErrorMessage(cr) + ". "
+					+ "Sending to retry topic " + nextDestination.getDestinationName() + ".");
+		}
 		else {
-			LOGGER.debug(e, () -> "Error processing record: " + getRecordInfo(cr) + " at topic " + cr.topic() + ". "
+			LOGGER.debug(e, () -> getErrorMessage(cr) + ". "
 					+ "Sending to retry topic " + nextDestination.getDestinationName() + ".");
 		}
 	}
 
-	private String getErrorMessage(ConsumerRecord<?, ?> cr) {
-		return "Record: " + getRecordInfo(cr) + " threw an error at topic " + cr.topic() + " and won't be retried. ";
+	private static String getErrorMessage(ConsumerRecord<?, ?> cr) {
+		return "Record: " + getRecordInfo(cr) + " threw an error at topic " + cr.topic();
 	}
 
-	private String getRecordInfo(ConsumerRecord<?, ?> cr) {
+	private static String getRecordInfo(ConsumerRecord<?, ?> cr) {
 		Header originalTopicHeader = cr.headers().lastHeader(KafkaHeaders.ORIGINAL_TOPIC);
 		return String.format("topic = %s, partition = %s, offset = %s, main topic = %s",
 				cr.topic(), cr.partition(), cr.offset(),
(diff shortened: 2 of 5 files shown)


tomazfernandes:
How the `TimestampedException` looks in the stacktrace now:

```
Caused by: org.springframework.kafka.listener.TimestampedException: Exception thrown at 2022-04-05T23:37:38.249Z; nested exception is org.springframework.kafka.listener.ListenerExecutionFailedException: Listener method 'public void org.springframework.kafka.retrytopic.RetryTopicIntegrationTests$MyCustomDltProcessor.processDltMessage(java.lang.Object)' threw exception; nested exception is java.lang.RuntimeException: Dlt Error!
```

spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
```diff
@@ -34,14 +34,36 @@
 
 	private final long timestamp;
 
+	@Deprecated
 	public TimestampedException(Exception ex, Clock clock) {
 		super(ex.getMessage(), ex);
 		this.timestamp = Instant.now(clock).toEpochMilli();
 	}
 
+	@Deprecated
```
artembilan, line 43: And what is wrong with the convenient ctor taking the `now()`  for its timestamp?
Why do we need to deprecate this one?

spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
```diff
@@ -34,14 +34,36 @@
 
 	private final long timestamp;
 
+	@Deprecated
 	public TimestampedException(Exception ex, Clock clock) {
```
artembilan, line 38: I won't mind to have this deprecated, but I wonder what drove you to use `Clock` as a required from end-user API in the beginning?

I never saw this class usage in the wild, but I see it more and more in your code 😄 

So, what is really a benefit to of this class and how does it help being a required dependency in the API for end-users?

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
```diff
@@ -34,14 +34,36 @@
...
+		super("Exception thrown at " + Instant.ofEpochMilli(timestamp), ex);
+		this.timestamp = timestamp;
+	}
+
+	/**
+	 * Creates an instance with the timestamp of when it was thrown and its cause.
+	 * @param ex the exception cause.
+	 * @param now the Instant of when the exception was thrown.
+	 */
+	public TimestampedException(Exception ex, Instant now) {
```
artembilan, line 64: Why just plain `long timestamp` not enough? 
Why end-user cannot do their `Instant.toEpochMilli()` when they use our `TimestampedException(Exception ex, long timestamp)` ?

spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
```diff
@@ -34,14 +34,36 @@
...
 	public TimestampedException(Exception ex) {
 		super(ex.getMessage(), ex);
-		this.timestamp = Instant.now(Clock.systemDefaultZone()).toEpochMilli();
+		this.timestamp = Instant.now().toEpochMilli();
+	}
+
+	/**
+	 * Creates an instance with the timestamp of when it was thrown and its cause.
+	 * @param ex the exception cause.
+	 * @param timestamp the timestamp of when the exception was thrown.
```
artembilan, line 52: `@since ` ?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
```diff
@@ -185,11 +185,39 @@ private TopicPartition resolveDestination(ConsumerRecord<?, ?> cr, Exception e)
...
 					? null
 					: resolveTopicPartition(cr, nextDestination);
 	}
 
+	private void logException(Exception e, ConsumerRecord<?, ?> cr, DestinationTopic nextDestination) {
+		if (nextDestination.isDltTopic()) {
+			LOGGER.error(e, () -> getErrorMessage(cr)
+					+ "Sending to DLT with name " + nextDestination.getDestinationName() + ".");
+		}
+		else if (nextDestination.isNoOpsTopic()) {
```
artembilan, line 200: "no ops" ? In the public API?

Doesn't sound neutral and technical.

How about `NONE` instead?

Not related, though, but I don't see any JavaDocs on those public APIs of the `DestinationTopic`.
At least that `Type` enums should have some explanations for their reason. 
Maybe then I wouldn't complain about "no ops" 😉

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
```diff
@@ -185,11 +185,39 @@ private TopicPartition resolveDestination(ConsumerRecord<?, ?> cr, Exception e)
...
+			LOGGER.debug(e, () -> "Error processing record: " + getRecordInfo(cr) + " at topic " + cr.topic() + ". "
+					+ "Sending to retry topic " + nextDestination.getDestinationName() + ".");
+		}
+	}
+
+	private String getErrorMessage(ConsumerRecord<?, ?> cr) {
+		return "Record: " + getRecordInfo(cr) + " threw an error at topic " + cr.topic() + " and won't be retried. ";
+	}
+
+	private String getRecordInfo(ConsumerRecord<?, ?> cr) {
```
artembilan, line 214: Looks like all of these methods can be `static`

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
tomazfernandes: Well, having `Clock` passed as a constructor parameter enables us to unit test time-sensitive logic, by injecting mock clocks, so that's the main driver. AFAIK, it's considered a good practice over methods such as System.currentTimeMillis(), which you can't really precisely unit test. 

The other reason would be enabling users to setup any clock they want, maybe there's more than one for different timezones, and whatnot. I've worked with applications that defined their own clocks - never more than one though.

So I've seen and used `Clock` for this in application code quite a few times for this reason. You can see [this SO question](https://stackoverflow.com/questions/27067049/unit-testing-a-class-with-a-java-8-clock) to confirm I'm not alone 😄

Also, for example, in Apache Kafka you have the [Time](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/utils/Time.java) interface, which is used all around for the same reason - just to give a couple examples, there's one in [KafkaConsumer.java](https://github.com/apache/kafka/blob/206ad4d2b530b1c0c776629256352cd044e68197/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L829-L837), or the [TimestampedWindowStoreBuilder.java](https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/state/internals/TimestampedWindowStoreBuilder.java) class.

So that's to say I didn't really invent this 😄 But I also think I may have gone too far in adding this to an `Exception`, so that's partially why I backed out on this.

artembilan: Thanks, @tomazfernandes .

I will need to learn more about this `Clock` abstraction.

My old school dictates me that no any testing hooks should be leaking into a production code, so we ended up with many reflection in our tests to modify a timestamp for assertions and so on 
Perhaps it's time to revise my vision for testing approaches 😄

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
tomazfernandes: Well, I remember I had to figure out a way to mark `end of line`, and at the time that seemed the best approach. I don't really remember which feature needed this instead of for example returning `null` as the next destination for an input topic - I can probably figure it out by reviewing the code.

Also, TBH I never thought of this as `public API` back then, though now of course I realize it is.

So I probably should write the javadocs to these methods and enums, and then we can evaluate if we need to change anything. But probably not as part of this PR, right?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
tomazfernandes: Sure. But if you don't bother me asking: why?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
artembilan: Memory and performance: https://www.simplilearn.com/tutorials/java-tutorial/static-keyword-in-java.

spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
tomazfernandes: TBH it's been some time since I've seen a `Date` for anything in Java code 😄.

Those new APIs are a real improvement over the older ones, [this article](https://www.baeldung.com/java-8-date-time-intro) does a good enough overview on them. The first ones are intensely used in production code these days.

I'd keep the `long` based one because there are places where you get a `long from epoch` timestamp, for example Kafka headers. So if users already have a `long` they could just use it. But of course, this is all theoretical - I'm ok with keeping only either of them.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
tomazfernandes: Nice, thanks! I'll change them.

tomazfernandes:
@artembilan, some of our topics regarding the `TimestampedException` were left hanging - please let me know if something still bothers you after these changes.

Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
```diff
@@ -34,14 +34,41 @@
...
-		super(ex.getMessage(), ex);
-		this.timestamp = Instant.now(Clock.systemDefaultZone()).toEpochMilli();
+		this(ex, Instant.now());
+	}
+
+	/**
+	 * Creates an instance with the timestamp of when it was thrown and its cause.
+	 * @param ex the exception cause.
+	 * @param timestamp the millis from epoch of when the exception was thrown.
+	 * @since 2.8.5
```
artembilan, line 56: So, back-port to `2.7.x` label on the issue is wrong. Isn't it?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
```diff
@@ -121,6 +124,22 @@ public boolean removeNotRetryableException(Class<? extends Exception> exceptionT
...
+	 * Never logs the listener exception.
+	 * The default is logging only after retries are exhausted.
+	 */
+	public void neverLogListenerException() {
+		this.loggingStrategy = ListenerExceptionLoggingStrategy.NEVER;
+	}
+
+	/**
+	 * Logs the listener exception at each attempt.
+	 * The default is logging only after retries are exhausted.
```
artembilan, line 137: `@since` ?

artembilan (changes requested):
Do we need to add some docs on the matter?

spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
tomazfernandes: Well, not sure really, I hadn't noticed that. If it's meant to be back-ported, what should I change this to? Maybe 2.7.x would be confusing since 2.8.0 - 2.8.4 doesn't have that? Not that proficient with some of these criteria yet.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactoryTests.java
```diff
@@ -62,6 +73,10 @@
 @SuppressWarnings({"unchecked", "rawtypes"})
 class DeadLetterPublishingRecovererFactoryTests {
 
+	private static final boolean isDebugEnabled =
+			((LogAccessor) ReflectionTestUtils.getField(DeadLetterPublishingRecovererFactory.class, "LOGGER"))
+					.isDebugEnabled(); // NOSONAR
```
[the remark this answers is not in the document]
tomazfernandes: > Do we need to add some docs on the matter?

About that, I'm not sure we should add this to the docs until we fix the configuration system for RT's components - might create more confusion than the problem it solves. For most users the default setting might be good enough, and if we solve the configuration for `2.8.5` we can add that later.

But, of course, I can add it now too. WDYT?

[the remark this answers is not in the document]
tomazfernandes: > This is something wrong to rely on the external static value which we don't control from this unit test.
> How about to use a `@LogLevels` on this class to be sure that our logging configuration is explicit for the test logic?

Oh, I didn't know about that annotation, cool! I'll use that then, thanks.

spring-kafka/src/main/java/org/springframework/kafka/listener/TimestampedException.java
artembilan: I guess this one `@since 2.7.13` as the current point release for that branch.
There is no reason to mention since what `2.8.x` or `2.9.x` just because you won't see this API in the `2.8.0` anyway.
Therefore if you see it like `@since 2.7.13` in your code base that means it is included into the release of `2.8.x` or `2.9.x`.
It might even not give any gain for end-users, unless there is a bug and it will be much easier to report the bug for the API which is marked with respective `@since`.
We then on our side will figure out what branches the fix belongs too.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactoryTests.java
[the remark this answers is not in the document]
artembilan: OK. Just fix the test and `@since` and I'll merge.

As I said on other your PR, the RT configuration refactoring is too drastic to be back-ported even to `2.8.x`.
It works some way over there already, so that's enough. 
The new vision or hard refactoring is better to do in new version (I guess `2.9.0` is good enough). So, it is easy for you to implement the feature without a breaking changes fear and it is more comfortable for end-users to refactor their projects to new API from our new version in their own new versions.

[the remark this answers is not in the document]
tomazfernandes: Well, about the configuration, I thought we came to the conclusion that if we can come up with something that is an optional add-on that that users can just not use if there's a problem or they simply don't want or need to, we might merge it to 2.8.5. But I think we'll have a clearer picture when I have something more concrete we can look into.

TBH, I think the current configuration system is too bad, specially when we consider that the next version in sync with Boot might be 3.0.0 at end of year. So if we decide not to back port this new approach, I'll still try to come up with something better for 2.8.x, even if it's only for the main features - of course, if that's not a problem.

artembilan (approved):
@tomazfernandes ,

do you have anything else to do for this PR?
Do I need to pay more attention to something?
Or may be you need @garyrussell 's opinion as well?

Thanks

tomazfernandes:
> @tomazfernandes ,
> 
> do you have anything else to do for this PR? Do I need to pay more attention to something? Or may be you need @garyrussell 's opinion as well?
> 
> Thanks

I don't see anything else to be done - it seems to be on par with what was discussed here: https://github.com/spring-projects/spring-kafka/discussions/2211

If you don't see any problems code-wise, I think we can merge this, and if @garyrussell has something else to add we can always patch it early next week.

That being said, I'm also ok with waiting for @garyrussell's input if you prefer.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactoryTests.java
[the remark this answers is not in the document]
tomazfernandes: I seem to be much more worried about this than you 😄 Maybe I'm overreacting. Probably @garyrussell can help settling this since he's been involved with this feature from the beginning. And we sure can take a better look after I work on the `Support`-based configuration.

But to provide some more context, we didn't use to have many features that relied on this kind of configuration - most features would use the `@RetryableTopic` annotation and `RetryTopicConfigurationBuilder`, which are APIs thought from the start to be used by end-users. That's probably why we didn't have so much bad feedback on this.

In the past couple versions, we've introduced a few features - e.g. the ability to set [framework level](https://docs.spring.io/spring-kafka/reference/html/#retry-topic-ex-classifier) `FATAL` exceptions, and the ability to [combine blocking and non-blocking retries](https://docs.spring.io/spring-kafka/reference/html/#retry-topic-combine-blocking), that can only be used through this kind of configuration.

Those seem to be gathering user attention - I think it took a couple of days from Boot launching with this last feature and the issue popping on SO.

So, as you've probably already have seen, to configure that feature, currently users might have to do:

```java
@Configuration
    public static class SO71705876Configuration {

                DeadLetterPublishingRecovererFactory deadLetterPublishingRecovererFactory) {
            ListenerContainerFactoryConfigurer lcfc = new ListenerContainerFactoryConfigurer(kafkaConsumerBackoffManager, deadLetterPublishingRecovererFactory, Clock.systemUTC());
            lcfc.setBlockingRetryableExceptions(IllegalArgumentException.class, IllegalStateException.class);
            lcfc.setBlockingRetriesBackOff(new FixedBackOff(500, 5)); // Optional
            return lcfc;
        }

        @Bean(name = RetryTopicInternalBeanNames.KAFKA_CONSUMER_BACKOFF_MANAGER)
        public KafkaConsumerBackoffManager backOffManager(ApplicationContext context) {
            PartitionPausingBackOffManagerFactory managerFactory =
                    new PartitionPausingBackOffManagerFactory();
            managerFactory.setApplicationContext(context);
            return managerFactory.create();
        }

        @Bean(name = RetryTopicInternalBeanNames.DEAD_LETTER_PUBLISHING_RECOVERER_FACTORY_BEAN_NAME)
        public DeadLetterPublishingRecovererFactory dlprFactory(DestinationTopicResolver resolver) {
            var DLPRF = new DeadLetterPublishingRecovererFactory(resolver);
            dlprf.neverLogListenerException();
            return new DeadLetterPublishingRecovererFactory(resolver);
        }

        @Bean(name = RetryTopicInternalBeanNames.DESTINATION_TOPIC_CONTAINER_NAME)
        public DestinationTopicResolver destinationTopicResolver(ApplicationContext context) {
            return new DefaultDestinationTopicResolver(Clock.systemUTC(), context);
        }
```

I think it was bad enough when we thought users had do declare only one bean in such way, but I think having to do all this for any feature is really not acceptable in my book. And it tends to get worse as we release new features - albeit we could defer some new things to `2.9.0`. But the logging configuration from the other PR is an example of this.

Let alone users having to look into `SO` for this workaround after being frustrated by errors - and I'd feel very bad if we actually documented this ugliness as an "official" workaround 😄 

So that's to provide some more context. Again, maybe I'm overreacting. I should have something on the `Support` configuration by Monday so we can take a better look.

I really appreciate you looking into all this. Thanks!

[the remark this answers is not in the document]
artembilan: I think you are hijacking this PR for other matter 😄 
Perhaps would be great if you are linking to this comment from the proper place.

Thanks

garyrussell:
...and cherry-picked to 2.9.x, 2.8.x, and to 2.7.x after resolving conflicts.

Issue #2214, opened 2022-04-06
Negative acknowledgment not working anymore when value is set to 0


Chr3is:
Version: 2.8.4

**Describe the bug**

With following commit: https://github.com/spring-projects/spring-kafka/commit/a1a5d48fcaf3b984fdd4a708f32982ddfaf2f63f#diff-49bc63455f16e2f0b1fe17ed1ec885135d494883d5a21a3da53d031098d09a33R2485
the behavior how the nack in the listener container is handled was changed. Allowed values for nack are `>=0`. However when setting the value to `0` you will end in an endless partition reseeking because the new introduced method `pauseForNackSleep` checks the nack value against `> 0` and **NOT** against `>= 0.` Due to this behavior the nack value is never reset to -1.

**To Reproduce**
execute `acknowledgment.nack(0);`

**Expected behavior**
When setting the value to 0 an endless partition reseeking does not happen.

Chr3is:
@sta-szek this bug was introduced with https://github.com/spring-projects/spring-kafka/commit/be4b1bf9921c7c9d918f6b7984fe31f4c25fa9e3 which was part of the v2.8.4 release (https://github.com/spring-projects/spring-kafka/releases/tag/v2.8.4) - the last working version should be 2.8.3 therefore. You can also "fix this" problem in 2.8.4 by setting the nack value from 0 to 1.

artembilan:
This is going to be released as a fix in the next `2.8.5` on April 18th: https://calendar.spring.io/

sta-szek:
> @sta-szek this bug was introduced with [be4b1bf](https://github.com/spring-projects/spring-kafka/commit/be4b1bf9921c7c9d918f6b7984fe31f4c25fa9e3) which was part of the v2.8.4 release ([`v2.8.4` (release)](https://github.com/spring-projects/spring-kafka/releases/tag/v2.8.4)) - the last working version should be 2.8.3 therefore. You can also "fix this" problem in 2.8.4 by setting the nack value from 0 to 1.

hi, thanks! indeed.
i switched back to spring-kafka 2.8.3 and the issue gone

however, if i add some events to topic, to avoid `nack(0)` then in my case the issue still happens in 2.8.4, but its, how to say it, less invasive.

i mean, if i nack(0) then after 1h i have ~200k events repeated.
if i put any event and do nack(x) where x > 0, then i got ~15 events repeated after 1h.

[...]

PR #2215, opened 2022-04-06
GH-2214: Fix Manual Nack with Zero Sleep

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2214

Regression; manual nack with zero sleep went into a tight loop since
2.8.4.

Add a copy of the current test, with a zero sleep; also verify that
nacking the first and last records work.

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/ManualNackRecordZeroSleepTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/ManualNackRecordZeroSleepTests.java
@@ -61,7 +61,6 @@ import org.springframework.kafka.core.ConsumerFactory;
 import org.springframework.kafka.listener.ContainerProperties.AckMode;
 import org.springframework.kafka.support.Acknowledgment;
 import org.springframework.kafka.test.utils.KafkaTestUtils;
-import org.springframework.lang.Nullable;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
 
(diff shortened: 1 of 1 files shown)


spring-kafka/src/test/java/org/springframework/kafka/listener/ManualNackRecordZeroSleepTests.java
```diff
@@ -0,0 +1,265 @@
...
+		}
+
+		@SuppressWarnings({ "rawtypes", "unchecked" })
+		@Bean
+		public ConcurrentKafkaListenerContainerFactory kafkaListenerContainerFactory() {
+			ConcurrentKafkaListenerContainerFactory factory = new ConcurrentKafkaListenerContainerFactory();
+			factory.setConsumerFactory(consumerFactory());
+			factory.getContainerProperties().setAckMode(AckMode.MANUAL);
+			factory.getContainerProperties().setMissingTopicsFatal(false);
+			factory.setRecordInterceptor(new RecordInterceptor() {
```
artembilan, line 249: Do we really need this interceptor to be copied in this test scope?
I had a pain back-porting it yesterday for other test class: some API changes in Kafka client.

Thanks

garyrussell: No; just a copy/paste artifact; will remove.

sta-szek:
hi, i left question there: https://github.com/spring-projects/spring-kafka/issues/2214#issuecomment-1091407124 but i feel this is better place to ask

Issue #2218, opened 2022-04-07
KafkaTestUtils getCurrentOffset() could take a AdminClient as argument


gurpiarbassi:
**Expected Behavior**

I'd like to be able to pass AdminClient as a argument to this method. That way I can initialise the AdminClient with relevant SSL properties to connect to a SSL cluster.

**Current Behavior**

Not able to pass SSL connection properties to AdminClient. It only uses bootstrap.servers at the moment.

**Context**

I cannot use this method at the moment to connect to a SSL cluster in my tests

PR #2219, opened 2022-04-07
GH-2218 add ability to pass AdminClient

gurpiarbassi:
- Fixes [GH-2218](https://github.com/spring-projects/spring-kafka/issues/2218)
-  Useful when you want to pass an AdminClient that has already
been initialised with specific properties to connect to the cluster
e.g. SSL properties.


artembilan (approved):
I think this is not harmful and can be accepted even if `KafkaTestUtils` has been designed for the testing against embedded Kafka.

garyrussell:
Thanks for the contribution; merged and cherry-picked to 2.9.x, 2.8.x

Issue #2220, opened 2022-04-07
Listener receives multiple messages when using @RetryableTopic with topicPartitions to reset offset


ashishqcs:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

Between 2.7.0 and 2.8.2

**Description**

I am trying to use @RetryableTopic for unblocking retries and topicPartitions in order to read messages from beginning.

Below are my main and DLT listeners:
```
@RetryableTopic(
            attempts = "4",
            backoff = @Backoff(delay = 1000),
            fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC)
    @KafkaListener(topicPartitions = {@TopicPartition(topic = "products",
            partitionOffsets = @PartitionOffset(partition = "0", initialOffset = "0"))})
    public void listen(ConsumerRecord<String, String> message, @Header(KafkaHeaders.RECEIVED_TOPIC) String topic) {

        log.info("message consumed - \nkey: {} , \nvalue: {}, \ntopic: {}, \nat: {}",
                message.key(),
                message.value(),
                message.topic(),
                LocalDateTime.now());
    }

    @DltHandler
    public void dltListener(ConsumerRecord<String, String> message, @Header(KafkaHeaders.RECEIVED_TOPIC) String topic) {
        log.info("message consumed at DLT - \nkey: {} , \nvalue: {}, \ntopic: {}, \nat: {}",
                message.key(),
                message.value(),
                message.topic(),
                LocalDateTime.now());
    }
```
With this main topic listener as well as DLT topic listener receive same event (since there is one retry topic - main listener receives same event 2 times and 1 by dlt listener as well - total 3 copies of same event).

logs:
```
15:10:50.950 [org.springframework.kafka.KafkaListenerEndpointContainer#8-0-C-1] INFO  c.m.s.c.n.SingleTopicRetryConsumer 
- message consumed - 
key: product1 , 
value: This is Product1, 
topic: products, 
at: 2022-04-07T15:10:50.950810          

15:10:50.950 [org.springframework.kafka.KafkaListenerEndpointContainer#9-retry-0-C-1] INFO  c.m.s.c.n.SingleTopicRetryConsumer - 
message consumed - 
key: product1 , 
value: This is Product1, 
topic: products, 
at: 2022-04-07T15:10:50.950810      

[...]

Issue #1554, opened 2020-07-31
@PartitionOffset Improvement


garyrussell:
When using manual assignment using `@TopicPartition` make it easier to apply the initial offset to all partitions (e.g. seekToBeginning).

PR #2223, opened 2022-04-08
GH-2220: Fix TopicPartitionOffset for Retry Topics

tomazfernandes:
Resolves https://github.com/spring-projects/spring-kafka/issues/2220

* Fix TopicPartitionOffset for Retry Topics

* Register NewTopics for Retry Topics only once per topic.

* Add docs explaining configuration behavior for retry topics when more than one listener container is used for the same topic.

Thanks


spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopic.java
```diff
@@ -219,6 +219,15 @@ public long delay() {
 			return this.delayMs;
 		}
 
+		/**
+		 * Returns the number of partitions the
```
artembilan, line 223: The method javadoc has to be imperative: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting
And `@since` ?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizerFactory.java
```diff
@@ -84,7 +92,17 @@ protected EndpointCustomizer addSuffixesAndMethod(DestinationTopic.Properties pr
 		};
 	}
 
-	protected Collection<EndpointCustomizer.TopicNamesHolder> customizeAndRegisterTopics(
```
artembilan, line 87: You cannot remove `public`/` protected` API if you'd like your fix to be back-ported.
This is what called a breaking change.
The point of point release is drop-n-run for the target project, so anything in that project can fail with a new point version.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizerFactory.java
```diff
@@ -84,7 +92,17 @@ protected EndpointCustomizer addSuffixesAndMethod(DestinationTopic.Properties pr
...
+	private TopicPartitionOffset[] getTopicPartitions(DestinationTopic.Properties properties,
```
artembilan, line 95: `static` ?

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicConfigurerTests.java
```diff
@@ -362,14 +364,23 @@ void shouldLogConsumerRecordMessage() {
 		RetryTopicConfigurer.LoggingDltListenerHandlerMethod method =
 				new RetryTopicConfigurer.LoggingDltListenerHandlerMethod();
 		method.logMessage(consumerRecordMessage);
-		then(consumerRecordMessage).should(never()).topic();
+		LogAccessor logger = (LogAccessor) ReflectionTestUtils
+				.getField(RetryTopicConfigurer.class, "LOGGER");
+		Assert.notNull(logger, "No LOGGER found in class " + RetryTopicConfigurer.class.getSimpleName());
+		if (logger.isInfoEnabled()) {
```
artembilan, line 370: Same `@LogLevels` trick?

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicIntegrationTests.java
```diff
@@ -138,6 +143,17 @@ void shouldRetryFourthTopicWithNoDlt() {
 		assertThat(awaitLatch(latchContainer.countDownLatch4)).isTrue();
 	}
 
+	@Test
+	void shouldRetryFifthTopicWithTwoListenersAndManualAssignment() {
+		logger.debug("Sending two messages to topic " + FIFTH_TOPIC);
```
artembilan, line 148: Why do we need this debug message in the test?

artembilan (changes requested):
Here is my review, but I'm not fully sure what is going on from the logic perspective, so I'll leave that review part to @garyrussell .

Thanks

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -148,6 +148,10 @@ public KafkaTemplate<String, Object> kafkaTemplate() {
 ----
 ====
 
+IMPORTANT: You can use multiple `@KafkaListener` annotations for the same topic with or without manual partition assignment along with non-blocking retries, but mind that only one configuration will be used for a given topic.
+It's best to use a single `RetryTopicConfiguration` bean for configuration of such topics;
+if you're using multiple `@RetryableTopic` annotations for the same topic, make sure all of them have the same values, otherwise one of them will be applied to all of that topic's listeners and the other annotations' values will be ignored.
```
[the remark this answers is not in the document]
tomazfernandes: > It's might be not relevant or just my own opinion, but technical documentation has to be impersonal. We cannot use `you`, `we`, `they` pointers. The sentence must be as neutral as possible.
> 

Totally agree, I'll change it, thanks.

> Also: one sentence per line: https://asciidoctor.org/docs/asciidoc-recommended-practices/

Hmm, maybe that's poor grammar on my part, but what do you consider a sentence? I'm breaking at periods and semi colons, but not on commas. Or am I missing something?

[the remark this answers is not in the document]
artembilan: Semicolon is not the end of sentence.
With breaking on the semicolon you cannot fulfill these tow rules: `You can easily swap sentences.`, `You can comment out sentences or add commentary to them.`

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizerFactory.java
tomazfernandes: Sure, for some reason I hadn't noticed it was `protected`, thanks.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicIntegrationTests.java
tomazfernandes: Well, I find it helpful when debugging tests and set the log level to debug, since there are many moving parts.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicConfigurerTests.java
tomazfernandes: That's a really neat and useful annotation. Have you ever thought of proposing that for the JUnit project?

artembilan: 😄  no, but I agree. We have it all over our with Gary projects, so it indeed could be a candidate for `spring-test`common module.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicIntegrationTests.java
tomazfernandes: Is that ok? Please let me know if it isn't. Those tests have a lot going on in parallel with all the retry topics being consumed, so logging is really helpful.

tomazfernandes:
Thanks a lot for the review @artembilan! I think I've addressed all the points, please let me know if I missed out on something.

artembilan (approved):
OK with me.

Let's give @garyrussell  to review this as well on Monday!

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizerFactory.java
```diff
@@ -71,7 +74,12 @@ protected EndpointCustomizer addSuffixesAndMethod(DestinationTopic.Properties pr
 			Collection<EndpointCustomizer.TopicNamesHolder> topics = customizeAndRegisterTopics(namesProvider, endpoint);
 			endpoint.setId(namesProvider.getEndpointId(endpoint));
 			endpoint.setGroupId(namesProvider.getGroupId(endpoint));
-			endpoint.setTopics(topics.stream().map(EndpointCustomizer.TopicNamesHolder::getCustomizedTopic).toArray(String[]::new));
+			if (endpoint.getTopics().isEmpty() && endpoint.getTopicPartitionsToAssign() != null) {
+				endpoint.setTopicPartitions(getTopicPartitions(properties, namesProvider, endpoint.getTopicPartitionsToAssign()));
```
garyrussell, line 78: This breaks the initial offset on the main listener because it overwrites the existing ones with a `null` initial offset.

We shouldn't do this for the `mainEndpoint` in `processAndRegisterEndpoint`.

```java
@SpringBootApplication
public class Kgh2220Application {

		SpringApplication.run(Kgh2220Application.class, args);
	}

	@Bean
	public NewTopic topic() {
		return TopicBuilder.name("kgh2220").partitions(1).replicas(1).build();
	}

	@KafkaListener(id = "kgh2220xxx", topicPartitions = @TopicPartition(topic = "kgh2220",
			partitionOffsets = @PartitionOffset(partition = "0", initialOffset = "0")))
	@RetryableTopic
	void listen(String in, @Header(KafkaHeaders.RECEIVED_TOPIC) String topic,
			@Header(KafkaHeaders.OFFSET) long offset) {
		System.out.println(in + "@" + offset);
		System.out.println(topic);
		System.out.println(KafkaUtils.getConsumerGroupId());
		throw new RuntimeException();
	}

	@Bean
	ApplicationRunner runner(KafkaTemplate<String, String> template) {
		return args -> {
			template.send("kgh2220", "foo");
		};
	}

}
```
```
foo@15
kgh2220
kgh2220xxx
```

garyrussell (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/EndpointCustomizerFactory.java
tomazfernandes: Good catch, makes sense. I'll fix it shortly, thanks.

garyrussell:
For test inspiration - see this test (using mocks) that verifies the container performed the initial seeks when the listener was so configured.

https://github.com/spring-projects/spring-kafka/blob/feb35635c3066937cebfdfcb15dd2e27e17c2673/spring-kafka/src/test/java/org/springframework/kafka/listener/SeekToCurrentOnErrorRecordModeTests.java#L99-L103

garyrussell:
I guess you could just clone that test and add a `@RetryableTopic` to its listener.

tomazfernandes:
Looks great, thanks a lot @garyrussell! Will finish up at least basic unit test coverage and try this out.

tomazfernandes:
@garyrussell, I stumbled into something that might be a small bug in the `TopicPartitionOffset` code.

The hashCode method is: 

```java
@Override
public int hashCode() {
    return this.topicPartition.hashCode() + this.position.hashCode();
}
```

But `position` is a nullable field, so in this case it throws a NPE if used. If that's an actual bug, DYT I should fix this as part of this PR, or should I leave it alone? The obvious fix I see would be something like:

```java
@Override
public int hashCode() {
    return this.position != null
                      ? this.topicPartition.hashCode() + this.position.hashCode()
                      : this.topicPartition.hashCode();
}
```

But if you have a different idea LMK.

Thanks.

tomazfernandes:
Or maybe add the other fields to the calculation if they're also not null.

tomazfernandes:
Also, the `equals` method doesn't take the `offset` into account:

```java
	@Override
	public boolean equals(Object o) {
		if (this == o) {
			return true;
		}
		if (o == null || getClass() != o.getClass()) {
			return false;
		}
		TopicPartitionOffset that = (TopicPartitionOffset) o;
		return Objects.equals(this.topicPartition, that.topicPartition)
				&& Objects.equals(this.position, that.position);
	}
```

I'm catching these due to unexpected results in assertThat().equals().

tomazfernandes:
@garyrussell, I actually can't really simply not call the endpoint customizer for the main topic because currently it's tied to the registration of the topics in the `DestinationTopicProcessor`, so I made it return early.

Although, what happens there today is that we use a no-ops suffixer for the main endpoint, so we basically get and set the same thing. But since users can provide their own implementation, I'm not sure that's not a breaking change - WDYT?

If that's a problem I can rework that to only leave the topic partition part out.

Thanks.

artembilan:
How about this:
```
@Override
	public boolean equals(Object o) {
		if (this == o) return true;
		if (o == null || getClass() != o.getClass()) return false;
		TopicPartitionOffset that = (TopicPartitionOffset) o;
		return this.relativeToCurrent == that.relativeToCurrent &&
				Objects.equals(topicPartition, that.topicPartition)
				&& this.position == that.position
				&& Objects.equals(this.offset, that.offset);
	}

	@Override
	public int hashCode() {
		return Objects.hash(this.topicPartition, this.position, this.offset, this.relativeToCurrent);
	}
```

Just include everything into those methods.

Separate issue/ PR though...

I guess your tests must not rely on the `equals()` and `hashCode()` for now or at all and the verification logic must be slightly different.

tomazfernandes:
I think the logic you propose would incur in the same error, wouldn't it? There's nothing to prevent the main endpoint from getting into the first if clause.  Or am I missing something?

I think if we just make sure the main endpoint doesn't enter either the topics or the TPO parts we should be good and least breaking as possible. WDYT?

And yes, all operations there are (or should be by default) no-ops for the main endpoint, except the `getTopics` part in the beginning which creates the `TopicNamesHolder` instances that we need.

tomazfernandes:
Sorry - `customizeAndRegisterTopics` in the beginning.

tomazfernandes:
Or better:

```java
	if (endpoint.getTopics().isEmpty() && endpoint.getTopicPartitionsToAssign() != null && !properties.isMainEndpoint()) {
		endpoint.setTopicPartitions(getTopicPartitions(properties, namesProvider, endpoint.getTopicPartitionsToAssign()));
	}
	else {
		endpoint.setTopics(endpoint.getTopics().stream().map(namesProvider::getTopicName).toArray(String[]::new));
	}
```

This has the least changes and also lets users go through the topics logic with a custom namesProvider if they wish to do something even for the main endpoint topics.

Since we're on a schedule for the release, I'll submit this logic along with the test changes, but I can change any of it if you prefer.

garyrussell (approved):
OK; I ran this version with my Boot app and it now works as expected.

If @artembilan 's approval from last week still stands, we can merge this.

artembilan (approved):

tomazfernandes:
I've just made a small improvement where users can also use the names provider with the main topic for topic partitions. To my understanding that's a really odd and unlikely combination, so I'm ok with leaving that out if you prefer - or I can submit if you don't mind looking at this again.

garyrussell:
I am confused about the hashcode change, though (reverted); it looks like a valid NPE.

garyrussell:
Sure; feel free if you have another commit.

artembilan:
I would suggest to make `TopicPartitionOffset` `equals()` and `hashCode()` in the separate PR which could be back-ported to all supported versions.

garyrussell:
@artembilan 
```java
	@Override
	public int hashCode() {
		return Objects.hash(this.topicPartition, this.position, this.offset, this.relativeToCurrent);
	}
```
Offset and relativeToCurrent are mutable so are not included in the equals or hash.

https://github.com/spring-projects/spring-kafka/issues/1554

tomazfernandes:
Well, TBH, since I'm not that familiar with `TopicPartitionOffset`, I'm not so sure about the new logic - that's why left it out in the first place - didn't want to add complexity so close to the release. I thought of doing this:

```java
	private static TopicPartitionOffset getTPOForMainTopic(RetryTopicNamesProviderFactory.RetryTopicNamesProvider namesProvider, TopicPartitionOffset tpo) {
		return new TopicPartitionOffset(namesProvider.getTopicName(tpo.getTopic()),
				tpo.getPartition(), tpo.getOffset(), tpo.getPosition());
	}
```

But I'm not so sure that's an exact copy of the original just with the topic name potentially changed (if the user provides a names provider) - there seems to be a `relativeToCurrent` flag which is not considered in this logic.

Any pointers? As I said, I'm ok with leaving the logic as is too.

garyrussell:
Good point (re fixing hash for back porting)

tomazfernandes:
Well, I guess this should do the trick:

```java
	private static TopicPartitionOffset getTPOForMainTopic(RetryTopicNamesProviderFactory.RetryTopicNamesProvider namesProvider, TopicPartitionOffset tpo) {
		TopicPartitionOffset newTpo = new TopicPartitionOffset(namesProvider.getTopicName(tpo.getTopic()),
				tpo.getPartition(), tpo.getOffset(), tpo.getPosition());
		newTpo.setRelativeToCurrent(tpo.isRelativeToCurrent());
		return tpo;
	}
```

garyrussell:
I am not sure why you want to replace the existing TPOs with an exact copy.

garyrussell:
Oh - it's so they can change the name; OK; that makes sense.

garyrussell (approved):

garyrussell:
...and cherry picked to 2.9.x, 2.8.x, and to 2.7.x after resolving minor conflicts.

garyrussell:
Hmmm - test failures on the back ports...
```
org.mockito.exceptions.verification.WantedButNotInvoked: Wanted but not invoked:
consumerRecordMessage.topic();
-> at org.springframework.kafka.retrytopic.RetryTopicConfigurerTests.shouldLogConsumerRecordMessage(RetryTopicConfigurerTests.java:369)
Actually, there were zero interactions with this mock.

Wanted but not invoked:
consumerRecordMessage.topic();
-> at org.springframework.kafka.retrytopic.RetryTopicConfigurerTests.shouldLogConsumerRecordMessage(RetryTopicConfigurerTests.java:369)
(88 more lines...)
```

Investigating...

tomazfernandes:
I changed these tests because they failed depending on the log level set for the tests. This is probably due to the change to `KafkaUtils` for logging. Easiest fix should be finding out which method is called on the consumer record and changing to assert that.

garyrussell:
Thanks; fixed.

Issue #2163, opened 2022-03-09
Deprecate Producer Listener


garyrussell:
Now that `ProducerInterceptor` is available.

https://github.com/spring-projects/spring-kafka/pull/2112

viktorardelean:
@garyrussell, I can work on this. Could you please assign it to me?

garyrussell:
Thanks!

We usually use `@deprecated since 3.0 in favor of #setProducerInterceptor` in the javadocs.

Now that we are on JDK 17, we can also set the `since` and for `Removal` attributes on `@Deprecated`

viktorardelean:
@garyrussell PR raised. Please have a look: https://github.com/spring-projects/spring-kafka/pull/2225

garyrussell:
Closing as invalid - the `ProducerInterceptor` does not have access to the outbound record.

PR #2225, opened 2022-04-08
[2163] Deprecate ProducerListener in favor of ProducerInterceptor

viktorardelean:
Resolves https://github.com/spring-projects/spring-kafka/issues/2163

Added Deprecated annotation where ProducerListener was used and suggested using ProducerInterceptor instead

Thanks

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/CompositeProducerListener.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/CompositeProducerListener.java
@@ -35,7 +35,7 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  *
  * @since 2.1.6
- * @deprecated since 3.0 in favor of {@link org.apache.kafka.clients.producer.ProducerInterceptor}
+ * @deprecated since 3.0 in favor of {@link org.springframework.kafka.support.CompositeProducerInterceptor}
  */
 
 @Deprecated(since = "3.0", forRemoval = true)
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/CompositeProducerListener.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/CompositeProducerListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2020 the original author or authors.
+ * Copyright 2018-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/CompositeProducerListener.java
```diff
@@ -35,8 +35,10 @@
  * @author Artem Bilan
  *
  * @since 2.1.6
- *
+ * @deprecated since 3.0 in favor of {@link org.apache.kafka.clients.producer.ProducerInterceptor}
```
artembilan, line 38: Maybe "in favor of CompositeProducerInterceptor"  ?

spring-kafka/src/main/java/org/springframework/kafka/support/LoggingProducerListener.java
```diff
@@ -32,7 +32,10 @@
  *
  * @author Marius Bogoevici
  * @author Gary Russell
+ * @deprecated since 3.0 in favor of {@link org.apache.kafka.clients.producer.ProducerInterceptor}
```
artembilan, line 35: Well, we don't have a `LoggingProducerInterceptor` to replace with.
I wonder if we need to implement one...

spring-kafka/src/main/java/org/springframework/kafka/support/ProducerListener.java
```diff
@@ -36,7 +36,11 @@
  * @author Endika Gutiérrez
  *
  * @see org.apache.kafka.clients.producer.Callback
+ *
+ * @deprecated since 3.0 in favor of {@link org.apache.kafka.clients.producer.ProducerInterceptor}
  */
+
+@Deprecated(since = "3.0", forRemoval = true)
```
artembilan, line 43: If we deprecate it, we have to replace its usage with the mentioned favor.
Any thoughts what could be done in the that `KafkaTemplate` for example?

I wonder if you ran your fix locally with the `./gradlew clean check`. It must show you that you still use some deprecate API.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/CompositeProducerListener.java
viktorardelean: yes, I have modified

garyrussell:
Thanks for the contribution; sorry for the delay.

After further review, I have decided not to deprecate `ProducerListener` because, unlike the `ProducerInterceptor`, it does not have access to the failed record in `onAcknowledgment()`.

viktorardelean:
Thanks for the update @garyrussell !

PR #2227, opened 2022-04-11
GH-2226: Add RetryTopicConfigurationSupport

### LINKED ISSUE #2174: Retry Topic's components can't be autowired before feature bootstrapping

### LINKED ISSUE #2226: Introduce RetryTopicConfigurationSupport
Improve the retry topic feature's component configuration by introducing a `RetryTopicConfigurationSupport`, based on `WebMvcConfigurationSupport`, as suggested by @artembilan.

tomazfernandes:
Resolves #2226
Also resolves #2174

This is a first draft of a solution with a `RetryTopicConfigurationSupport`, so we can evaluate if it's going in the right direction before doing a final review, adding docs, etc.

Overall, I liked it better than I thought I would - the bean count has been reduced to 3 or4 beans, and it gave a lot of clarity on the feature's design. It should make it easier for everyone to understand how the feature works. So thanks a lot @artembilan for insisting on this.

To be honest, I still think this might be a bit overwhelming for the average end-user to configure simple features, but if that's a real problem we can look into providing an easier alternative for some features in the future. It's definitely much better than what we have now, and a solid foundation to build upon.

There are a few decisions I made that are not exactly  `WebMvcSupport`-like - I'm ok with rolling any or all of them back if that's better. For example, I:

* Created a `RetryTopicComponentsFactory` to reduce the number of methods to be overridden in the `Support` class and keep it overall smaller;
* Separated the `KafkaBackOffManagerConfigurationSupport` because the `KafkaBackOffManager` is useful outside of Retry Topic's scope;
* Created the `@EnableRetryTopic` annotation - while we're using this as an optional add-on, I think it's easier to document adding the annotation than the `@Import` alternative - afterwards we can move this annotation to `@EnableKafka`, or maybe kill it in 3.0.0. It also imports the `KafkaBackOffManagerConfigurationSupport` class;
* Created a new `RetryTopicBeanNames` class and deprecated the `RetryTopicInternalBeanNames` class - since we're using a lot less beans, there's no need for so many names. Also this enables setting the names to the framework standards. For now the old names should still work - I need to do a more thorough review to ensure that. If that seems to risky, I'm ok with keeping the old names until 3.0.
* There are a few constructors that might be improved - for example adding `ApplicationContextAware` to some beans instead of injecting an `ApplicationContext` through the constructor. But I think we can review this in the future.

Also, after more consideration, although I don't see any real risk since we're not actually changing any legacy logic besides some optional bean names which we can rollback, I think maybe it's indeed better for us to hold this for 2.9 - time is flying by and we should release it soon enough. Of course, I'm also ok with releasing it as 2.8.x if that's better.

I used this new configuration support for the `RetryTopicIntegrationTests` and `RetryTopicExceptionRoutingIntegrationTests`, so these are good places to take a look into how it's like. I also made sure the former works with legacy configuration.

Let me know what you think about this - hopefully this will be a solid step in the right direction for the RT feature.

Thanks!


spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableRetryTopic.java
```diff
@@ -0,0 +1,87 @@
...
+ * Enables the non-blocking topic-based delayed retries feature. To be used in
+ * {@link Configuration Configuration} classes as follows:
+ * <pre class="code">
+ *
+ * &#064;EnableRetryTopic
+ * &#064;Configuration
+ * public class AppConfig {
+ *
+ * 	&#064;Bean
+ * 	public RetryTopicConfiguration myRetryTopicConfiguration(KafkaTemplate kafkaTemplate) {
```
artembilan, line 40: I this optional?
I thought something like `@EnableRetryTopic` should be fair enough for some default values...

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaBackOffManagerConfigurationSupport.java
```diff
@@ -0,0 +1,108 @@
...
+	/**
+	 * Override this method to provide a different {@link TaskExecutor}
+	 * to be used with the {@link KafkaConsumerTimingAdjuster}.
+	 * @return the instance.
+	 */
+	protected TaskExecutor timingAdjusterTaskExecutor() {
+		ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+		executor.initialize();
+		this.applicationContext
+				.addApplicationListener((ApplicationListener<ContextClosedEvent>) event -> executor.shutdown());
```
artembilan, line 99: This is a bit awkward.
Consider to implement `DisposableBean` instead.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -509,11 +510,15 @@ private boolean processMainAndRetryListeners(KafkaListener kafkaListener, Object
 		return true;
 	}
 
+	@SuppressWarnings("deprecation")
```
artembilan, line 513: Are you sure that we really have something in this method to suppress?

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicConfigurationSupport.java
```diff
@@ -0,0 +1,339 @@
...
+	 * @return the instance.
+	 */
+	protected RetryTopicComponentFactory createComponentFactory() {
+		return new RetryTopicComponentFactory();
+	}
+
+	@Deprecated
+	@Bean(name = RetryTopicInternalBeanNames.RETRY_TOPIC_BOOTSTRAPPER)
+	RetryTopicBootstrapper retryTopicBootstrapper(ApplicationContext context) {
+		return new RetryTopicBootstrapper(context, context.getAutowireCapableBeanFactory());
```
artembilan, line 269: This is an overhead. 
Consider to use `ApplictionContextAware` and `BeanFactoryAware`.
The framework will inject them respectively when this bean is created.

Perhaps many other features you create via specific factory could be handled also as just beans and application context hooks...

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
```diff
@@ -98,6 +98,7 @@ public class ListenerContainerFactoryConfigurer {
 
 	private final Clock clock;
 
+	@SuppressWarnings("deprecation")
```
artembilan, line 101: I don't see what is deprecated for this ctor to use...
Do we really need this annotation over here?

artembilan (changes requested):
At a glance this is great!
👍

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableRetryTopic.java
tomazfernandes: Haven't thought of that, sounds interesting. I can think this through, but at first it seems to me that it might not be so simple - currently if we have a greedy configuration that accepts all topics, it might impede users from having specific configurations for specific topics.

We could either make a greedy, general configuration possible (that's something I'd been thinking of anyway) with users only providing the specific ones, or maybe there's a way we can add a bean conditionally on not having a user provided one? I see all `@ConditionalOn` goodies are part of Boot and we don't have it here.

WDYT?

artembilan: The `@EnableRetryTopic` may come with attributes to configure something global or for default values...

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicConfigurationSupport.java
tomazfernandes: Yes, I think I mentioned this somewhere. This specific bean is deprecated and I'm not sure it's worth change anything on it, but there are a couple other beans that should benefit from it.

My initial concern was about changing this for a patch version, but since we're releasing this in a milestone I think we can just go ahead and do it.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
tomazfernandes: I'll double check, but I think we had some bean names. Mind that I deprecated the entire `RetryTopicInternalBeanNames` class, so some suppressed deprecations are because of that.

tomazfernandes: Yep, bean names it is.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
tomazfernandes: Same, deprecated RetryTopicInternalBeanNames.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaBackOffManagerConfigurationSupport.java
tomazfernandes: I agree this looks awkward, just couldn't think of a better way 😄 You mean implementing `DisposableBean` in the `Support` class and keeping a reference to the `TaskExecutor` to `destroy()` it later?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
artembilan: OK. But why do you still use that deprecated class then?
Looks like those are just string constants, so why just don't use their values in place instead of referring to that deprecated class? That's for the case if we cannot get rid of these deprecated names yet...

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaBackOffManagerConfigurationSupport.java
artembilan: Right. Why not?

artembilan: See the same `WebMvcConfigurationSupport`. It does have a state via its properties:
```
	@Nullable
	private List<Object> interceptors;

	@Nullable
	private PathMatchConfigurer pathMatchConfigurer;

	@Nullable
	private ContentNegotiationManager contentNegotiationManager;
```

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
tomazfernandes: Well, it's a matter of how rigorous we need to be with backwards compatibility.

All current documentation points to users declaring beans with these names, and more curious users might have other creative solutions that rely on overriding some of these.

So my thinking is, for 2.9, anything that worked before, should still work just the same - that's why we should be able to introduce new names but still make sure old ones work. And for 3.0 we could get rid of the old ones. Is that thinking right?

But I'm ok with ripping the bandaid sooner if you think that's ok.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaBackOffManagerConfigurationSupport.java
tomazfernandes: Oh, sure, makes sense. I'll change this, thanks.

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableRetryTopic.java
tomazfernandes: Sure, that sounds interesting too. But I wonder, Andy Wilkinson from the Boot team pinged me earlier today to look into having auto configuration for the retryable topics in RC1 next week, not sure you saw that. Wouldn't those two approaches maybe conflict a bit?

Users shouldn't need `@EnableRetryTopic` to make the feature work with auto configuration, since boot won't have access to it until 3.0, but I think they still should be able to use it if, for example, they want to inject a feature bean such as the `DestinationTopicResolver`, as we've discussed.

WDYT?

artembilan: Correct. Independently of the solution here, the auto-configuration should give end-user whatever is needed to be done by default or from the configuration properties.

I think I see your pain and perhaps we need just to deffer this new vision until `3.0` and let you concentrate on the auto-configuration for upcoming Spring Boot `2.7`.
Note: Gary doesn't have intention to release `spring-kafka-2.9.0` for that Spring Boot. 
So, whatever we try to achieve here cannot benefit from the auto-configuration for a while.
I mean I still doubt this can make it smoothly to `2.8.x`...

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ListenerContainerFactoryConfigurer.java
garyrussell: BTW we can push RC1 out a week or so if necessary, it is not tied to the Boot release train.

garyrussell: Sorry M1.

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableRetryTopic.java
tomazfernandes: I'm ok with working with either 2.x or 3.0, or both. What I don't think I'm really comfortable with is adding features to 2.x with the current configuration system in place, since it'll only bring confusion and pain for our users. Does that make sense, or maybe I am missing something?

So, the way I see it, we can insist on making this backwards compatible with 2.x, or maybe come up with simple, less-than-perfect patches to keep the 2.x ball rolling until it's a bit closer to the 3.0 release. Or we can just leave everything new to 3.0 and be done with 2.x as far as retryable topics go, except of course for bug fixes. I'm ok with both approaches.

WDYT? If you have any other ideas I'm all ears too.

Thanks!

tomazfernandes: As a note, for now all auto-configuration will do is create a global `RetryTopicConfiguration` bean with a few parameters - no infrastructure involved, and I guess not too much work to be done there too. While it might help getting the feature in the hands of more users, it shouldn't really help with any of this infrastructure problem. Of course, if you have any different ideas please let me know.

tomazfernandes: So, maybe what we might be able to do is having `@EnableRetryTopic` provide a base configuration if it doesn't find any other `RetryTopicConfiguration` beans - this way if the user has provided any other configuration, be it through auto-configuration or via a `@Configuration` class, it will back off, otherwise, this default configuration kicks in, and we could add a couple configuration properties there just as we do in auto configuration.

How does that sound? I like the idea since it allows users to have a declarative way of using the feature without needing to mess around with yml files.

tomazfernandes: I think we may have had some kind of misunderstanding here 😄

> Wouldn't those two approaches maybe conflict a bit?

What I meant is that auto configuration might conflict with the idea of `@EnableRetryTopic` providing a default configuration - not with this `Support` approach as a whole. Is that what you understood? If so, sorry for not being clearer.

> I think I see your pain and perhaps we need just to deffer this new vision until 3.0 and let you concentrate on the auto-configuration for upcoming Spring Boot 2.7.

So I didn't really understand why we might defer this to 3.0. If that's only because of focusing on auto-configuration, I don't think there's any conflict between them and I can do both.

Please let me know if I understood any of this incorrectly. Either way, I should submit the changes we've discussed sometime today, and I'll also begin looking at auto-configuration.

Thanks again.

tomazfernandes:
I've made the suggested changes - please let me know if there's anything else. If code-wise this looks ok, I can start working on tests and reviewing the javadocs.

If we want to make `@EnableRetryTopic` provide a default configuration, I think we can address that in a separate PR.

I'm working under the assumption that we'll be releasing this in 2.9 - if that's not the case please let me know so I can update the `since` tags.

Thanks.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurer.java
```diff
@@ -230,17 +232,34 @@ public class RetryTopicConfigurer {
...
+	}
+
+	/**
+	 * Create an instance with the provided properties.
+	 * @param destinationTopicProcessor the destination topic processor.
+	 * @param containerFactoryResolver the container factory resolver.
+	 * @param listenerContainerFactoryConfigurer the container factory configurer.
+	 * @param retryTopicNamesProviderFactory the retry topic names factory.
+	 */
+	@Autowired
```
artembilan, line 254: Can we just don't do this?

No one is going to create a bean instance of this class without calling this ctor anyway.

From my experience all the auto-wiring features is for end-user projects: the framework stuff has to be as explicit as possible.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurer.java
tomazfernandes: We need this for the legacy bootstrapping - it's autowired there. Since we have more than one constructor, we need tell the framework which one is to be used.

An alternative would be removing the old constructor, but that would be a breaking change.

Any ideas?

tomazfernandes: We actually used to have at least one more constructor for this class which was deprecated - probably @garyrussell removed it for 3.0, but it's probably still there for 2.x

artembilan: Right. And how it is used `new RetryTopicConfigurer()`?
So, no any auto-wire magic. Isn't it?

tomazfernandes: We actually don't have a default constructor.

It's declared here:
https://github.com/spring-projects/spring-kafka/blob/feb35635c3066937cebfdfcb15dd2e27e17c2673/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicBootstrapper.java#L86

And used here:
https://github.com/spring-projects/spring-kafka/blob/feb35635c3066937cebfdfcb15dd2e27e17c2673/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java#L514

The framework autowires the dependencies.

tomazfernandes: I've just checked and the legacy constructor is still there for 2.9 and 2.8.x. So for 3.0, we can remove the now deprecated constructor and remove the `@Autowired`, but for 2.9 we'd need to remove both deprecated constructors for that.

artembilan: OK. I see:
```
registry.registerBeanDefinition(beanName, new RootBeanDefinition(beanClass))
```

So, yeah... Let's leave it for now!

Thanks

tomazfernandes: Sure, and thanks for the pointer on making everything in the framework as explicit as possible, that’s a good thing to keep in mind.

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicComponentFactory.java
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @author Tomaz Fernandes
+ * @since 2.9
+ */
+public class RetryTopicComponentFactory {
+
+	private final Clock internalRetryTopicClock = createInternalRetryTopicClock();
+
+	/**
+	 * Creates the {@link RetryTopicConfigurer} that will serve as an entry point
```
artembilan, line 53: The method Javadoc must be imperative: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-kafka/src/main/java/org/springframework/kafka/listener/PartitionPausingBackOffManagerFactory.java
```diff
@@ -161,7 +159,15 @@ private TaskExecutor getOrCreateTimingAdjustmentThreadExecutor() {
...
 		ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
 		executor.initialize();
-		super.addApplicationListener((ApplicationListener<ContextClosedEvent>) event -> executor.shutdown());
+		this.taskExecutor = executor;
 		return executor;
 	}
+
+	@Override
+	public void destroy() throws Exception {
+		if (this.taskExecutor != null
```
artembilan, line 168: I think this is not this factory responsibility. 
Consider to move logic for the default `executor` directly to that `WakingKafkaConsumerTimingAdjuster`.
Do we really need a thread pool over there for that `adjustTiming()`?
Isn't a single dedicated thread enough?
Kinda `Executors.newSingleThreadExecutor()` ?

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicConfigurationSupport.java
```diff
@@ -0,0 +1,369 @@
...
+	}
+
+	@Override
+	public void destroy() throws Exception {
+		if (this.disposableBackOffManagerFactory != null) {
+			this.disposableBackOffManagerFactory.destroy();
```
artembilan, line 298: Therefore when we revise an `executor` for the `WakingKafkaConsumerTimingAdjuster`, we won't need this logic over here at all.

artembilan (changes requested):
Some further review.

Thanks

garyrussell:
I am inclined to push out the 2.9 M1 for up to a week, since we are still in review mode here.

WDYT?

tomazfernandes:
> I am inclined to push out the 2.9 M1 for up to a week, since we are still in review mode here.
> 
> WDYT?

TBH, I wasn't aware we had a release date for M1 😄 Yeah, I think the 25th should be doable. Anything happens, I'll let you know and we can either push the date a bit further or leave this for M2, if that works for you.

garyrussell:
I scheduled it for the 18th because it's easier for me to do all the releases the same day. 

Moving to the 25th (which doesn't have to be firm).

AFAIKT, the only one we need for Monday is https://github.com/spring-projects/spring-kafka/pull/2223

spring-kafka/src/main/java/org/springframework/kafka/listener/PartitionPausingBackOffManagerFactory.java
tomazfernandes: > I think this is not this factory responsibility.
> Consider to move logic for the default `executor` directly to that `WakingKafkaConsumerTimingAdjuster`.

Sure, I'll take a look into that, thanks.

> Do we really need a thread pool over there for that `adjustTiming()`?
> Isn't a single dedicated thread enough?
> Kinda `Executors.newSingleThreadExecutor()` ?

Currently we adjust the timing for increasing the `back off`'s precision by spawning a thread that will `wakeup` the consumer at the appropriate time, if there's such relevant adjustment to be made. This was created this way to keep it simple - fire and forget, no thread management required.

Since normally these should be short lived tasks, it shouldn't be much of a performance issue having that many threads and the thread pool assures we reuse the threads. Also, the threads spend most of their time sleeping, so there's not that much CPU usage. If we have a single thread with this strategy, it might get blocked by a longer delay and make other delays less precise.

It might be worth looking at something more of a manager with a single thread that goes through values and wakes up the consumers as needed. It's more complex, and might be a busy thread, but probably would be more performant overall, specially if we kill it when not in use.

But for the solution as is, I think we still need the `ThreadPool`.

WDYT?

Thanks

tomazfernandes:
> I scheduled it for the 18th because it's easier for me to do all the releases the same day.
> 
> Moving to the 25th (which doesn't have to be firm).
> 

Sure, sounds good, thanks.

> AFAICT, the only one we need for Monday is #2223

Yes, I'm working on that right now. I noticed since the `EndpointCustomizerFactory` was extracted from the `RetryTopicConfigurer` class at some point in time, it doesn't have such great unit test coverage - actually the coverage it had from the `RetryTopicConfigurerTest` was removed, probably at the same time, so I'm working on that.

Also, I couldn't really figure out an worthwhile way to adjust the integration tests so that they'd fail with the current logic - I think I'd need a way of setting the partition's offset to 1 before the container is created, and make sure that when the container starts it resets back to 0. If you have any pointers on that please let me know, probably on that PR so we don't hijack this one.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/PartitionPausingBackOffManagerFactory.java
tomazfernandes: I changed the `TaskExecutor` to the `WakingKafkaConsumerTimingAdjuster` and it indeed looks a lot better, thanks!

I thought of extending the `SmartLifecycle` interface in both the `KafkaConsumerBackOffManager` and `KafkaConsumerTimingAdjuster` interfaces instead of `DisposableBean`. While the abstraction itself seemed more appropriate since these are long running beans, it was clumsy to implement that without breaking backwards compatibility, so for now it doesn't seem like it's worth it - please let me know if you think otherwise.

Thanks a lot for your input, and please let me know if there's anything else.

tomazfernandes:
Added unit tests, if you care to take a look. It gives a nice perspective on how configurations can be achieved.

Also, happy to say that on aggregate (UT + IT), the new classes have 100% test coverage, and the `retrytopic` package has 94% test coverage on class and lines. I know that test coverage doesn't mean it's infallible, but that's certainly a nice safety net to have.

Thanks a lot for the support and guidance from you both! Next up, docs and javadocs.

tomazfernandes:
As a note, we're already using this new `Support` approach in two main integration test suites, that's why I'm not adding any new ones.

spring-kafka/src/main/java/org/springframework/kafka/listener/WakingKafkaConsumerTimingAdjuster.java
```diff
@@ -138,4 +154,18 @@ private void doApplyTimingAdjustment(Consumer<?, ?> consumerForTimingAdjustment,
 					"for TopicPartition " + topicPartition);
 		}
 	}
+
+	private static TaskExecutor createTaskExecutor() {
+		ThreadPoolTaskExecutor taskExecutor = new ThreadPoolTaskExecutor();
```
artembilan, line 159: Again: why plain `Executors.newSingleThreadExecutor()` is not enough to have a full default control of this?
This way we don't need that `destroy()` overhead in this bean.

spring-kafka/src/main/java/org/springframework/kafka/listener/PartitionPausingBackoffManager.java
```diff
@@ -229,4 +219,11 @@ protected void removeBackoff(TopicPartition topicPartition) {
 			this.backOffContexts.remove(topicPartition);
 		}
 	}
+
+	@Override
+	public void destroy() throws Exception {
+		if (this.kafkaConsumerTimingAdjuster instanceof DisposableBean) {
+			((DisposableBean) this.kafkaConsumerTimingAdjuster).destroy();
```
artembilan, line 226: If we remove `DisposableBean` on the `WakingKafkaConsumerTimingAdjuster`, we don't need this delegation.
Nevertheless the question is still why it is here anyway?
Isn't that  `WakingKafkaConsumerTimingAdjuster` injected into this manager a bean, so its disposition is controlled from the outside anyway?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/WakingKafkaConsumerTimingAdjuster.java
tomazfernandes: Sorry, I had answered this [here](https://github.com/spring-projects/spring-kafka/pull/2227#discussion_r849766492), but I think I marked it as resolved before you had a chance to look at it.

In a shorter version, we currently rely on spawning threads to wake the consumers after a calculated delay to increase delay precision. While we could improve upon this design and have a single manager-like thread doing all the work, currently if we have a `newSingleThreadExecutor()` and a longer delay task is submitted before a shorter one, we'd block the later and lose delay precision.

I chose Spring's `ThreadPoolExecutor` as default because it has a more robust implementation than Java's default `ThreadPoolExecutor` - for instance, by default, on `shutdown()` it will actually call `shutdownNow()` under the hood and cancel all remaining tasks, which is perfect for our use-case since there's nothing for the threads to do as cleanup.

What I think is missing here is having a default limit on the thread pool size, and perhaps making it clearer for users that this is being used, while also letting them easily change the pool size limit. Also, I could have used Java's `Executor` interface instead of Spring's `TaskExecutor` for this - not sure if it's worth changing this though since in `Spring Boot` applications users usually use the Spring interface anyway.

Even if we switched to a plain JDK `Executor` such as the ones returned by the `Executors` class, unless I'm missing something, it'd [still be necessary](https://jenkov.com/tutorials/java-util-concurrent/executorservice.html#executorservice-shutdown) to call `shutdown()` somewhere to make sure the `Executor` doesn't accept new tasks and let it free up resources. That's why in `ThreadPoolTaskExecutor` `shutdown()`  is called in the `destroy()` method - if our thread pool was a bean, we'd be able to leverage that, and thus free ourselves from doing it manually.

spring-kafka/src/main/java/org/springframework/kafka/listener/PartitionPausingBackoffManager.java
tomazfernandes: Not really, in the current `Support` implementation the `KafkaBackOffManagerFactory` instantiates it manually, so the container is not aware of it. I think this goes well with the purpose of not registering  beans that are not strictly necessary to be as such, but I can revert it back to a bean approach if that's better.

I don't seem to be able to get a permalink from the PR, but the logic in the factory is:

```java
return this.taskExecutor != null
 		? new WakingKafkaConsumerTimingAdjuster(this.taskExecutor)
 		: new WakingKafkaConsumerTimingAdjuster();
```

spring-kafka/src/main/java/org/springframework/kafka/listener/WakingKafkaConsumerTimingAdjuster.java
tomazfernandes: I think I can come up with a `KafkaBackOffManagerConfigurer` in the `RetryTopicConfigurationSupport` class to make all these options more explicit to the user.

artembilan: The `Executors.newSingleThreadExecutor()` is a `FinalizableDelegatedExecutorService`. So, it frees its resources when the instance is destroyed by garbage collector.

But yeah... to much attention to it.
Let's leave it as is for now and we always can revise it later (if any)!

tomazfernandes: That's nice, I wasn't aware of that. But I think your concern with using a `ThreadPoolTaskExecutor` is valid, and I think it's worth making it clear in the `Support` class that such an expensive resource is being used, while also allowing users to easily customize that. Maybe even add a small entry to the documentation about it. I'll work on these changes.

And it's a good thing to keep in mind to revise this solution when we have the time so we can use a single-threaded approach instead.

Thanks a lot for your input!

tomazfernandes:
Considering what was discussed with @artembilan, I've changed the `TaskExecutor` to a `@Bean` - this way Spring manages its lifecycle so we don't need to. This is a minor breaking change - users of the `PartitionPausingBackOffManagerFactory` class might expect a `ThreadPoolTaskExecutor` to be created there. I've documented that in the class' javadocs. But we'd have a breaking change either way by moving the `TaskExecutor` creation and management to the `WakingKafkaConsumerTimingAdjuster`, and those seems like niche use cases, so it seems it's a lesser problem - please let me know if otherwise.

I also created a `KafkaBackOffManagerConfigurer` to let users easily manage this. If `timingAdjustment` is disabled by the user, the `ThreadPoolTaskExecutor` is not created.

@garyrussell, I've also added a `defaultFatalExceptionsList()` method to the `ExceptionClassifier` that returns a new list with the default classifications. With that, when configuring the exceptions for the RT feature, we can provide a populated list for the user to manage, and then set the classifications back to the classifier. Please let me know if that looks ok - I can work around that if you think that's an unwelcome change or has a better idea.

Thanks!

tomazfernandes:
I've added documentation, changed the code examples to the new approach and reviewed the javadocs. Of course, that's a whole lot of text 😄 - there are probably some mistakes around we can fix as we go through the milestone cycle. But I think overall we can see in the code example diffs the improvements achieved with this approach.

As far as I'm concerned this is good to go, so I'm removing the `Draft` status - it should be a matter of addressing any further concerns you might have.

I think we look good for releasing this next week, so thank you very much for the support and guidance so far.

As usual, let me know if there's anything else to be changed, or any further concerns.

Thanks!

tomazfernandes:
Changed the Support class to use JavaUtils instead of many if statements.

Also added an acceptIfInstanceOf method to JavaUtils that checks if the value is of the given type, and if so it casts the value and invokes the consumer. This was helpful to me in two places of the Support code.

Hope that’s ok - I can roll it back otherwise.

Thanks!

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -151,6 +156,42 @@ public KafkaTemplate<String, Object> kafkaTemplate() {
...
 It's best to use a single `RetryTopicConfiguration` bean for configuration of such topics; if multiple `@RetryableTopic` annotations are being used for the same topic, all of them should have the same values, otherwise one of them will be applied to all of that topic's listeners and the other annotations' values will be ignored.
 
+[[retry-topic-global-settings]]
+===== Configuring Global Settings and Features
+
+Since 2.9, the previous bean-based configuration approach has been deprecated.
+Instead, the `RetryTopicConfigurationSupport` class should be extended, the proper methods overridden, and imported in a `@Configuration` annotated class.
+An example follows:
+
+====
```
garyrussell, line 166: Why must it be `@Import`ed; surely the `@Configuration` class can extend `RetryTopicConfigurationSupport` itself.

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableRetryTopic.java
```diff
@@ -0,0 +1,91 @@
...
+ * </pre>
+ *
+ * @author Tomaz Fernandes
+ * @since 2.9
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Documented
+@Import(RetryTopicConfigurationSupport.class)
+public @interface EnableRetryTopic {
```
garyrussell, line 90: Perhaps `@EnableKafkaRetryTopic`? There are many `@Enable*` annotations across the Spring portfolio.

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicComponentFactory.java
```diff
@@ -0,0 +1,198 @@
...
+	 * with the {@link KafkaBackoffAwareMessageListenerAdapter}.
+	 * @param deadLetterPublishingRecovererFactory the factory that will provide the
+	 * {@link DeadLetterPublishingRecoverer} instance to be used.
+	 * @param clock the {@link Clock} instance to be used with the listener adapter.
+	 * @return the instance.
+	 */
+	public ListenerContainerFactoryConfigurer listenerContainerFactoryConfigurer(KafkaConsumerBackoffManager kafkaConsumerBackoffManager,
+																				DeadLetterPublishingRecovererFactory deadLetterPublishingRecovererFactory,
+																				Clock clock) {
+
```
garyrussell, line 147: Strange indentation.

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicConfigurationSupport.java
```diff
@@ -0,0 +1,485 @@
...
+	 * @param beanFactory the {@link BeanFactory}.
+	 * @return the instance.
+	 * @see KafkaListenerAnnotationBeanPostProcessor
+	 */
+	@Bean(name = RetryTopicBeanNames.RETRY_TOPIC_CONFIGURER_BEAN_NAME)
+	public RetryTopicConfigurer retryTopicConfigurer(@Qualifier(KafkaListenerConfigUtils.KAFKA_CONSUMER_BACK_OFF_MANAGER_BEAN_NAME)
+																KafkaConsumerBackoffManager kafkaConsumerBackoffManager,
+													@Qualifier(RetryTopicBeanNames.DESTINATION_TOPIC_RESOLVER_BEAN_NAME)
+															DestinationTopicResolver destinationTopicResolver,
+													BeanFactory beanFactory) {
```
garyrussell, line 100: We don't use this indentation style; the additional parameters (and the first in this case) should start at 2 tabs past the `public`.

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicConfigurationSupport.java
```diff
@@ -0,0 +1,485 @@
...
+
+	/**
+	 * Internal method for processing the {@link DeadLetterPublishingRecovererFactory}.
+	 * Consider overriding the {@link #configureDeadLetterPublishingContainerFactory()}
+	 * method if further customization is required.
+	 * @param deadLetterPublishingRecovererFactory the instance.
+	 */
+	private void processDeadLetterPublishingContainerFactory(
+			DeadLetterPublishingRecovererFactory deadLetterPublishingRecovererFactory) {
+		CustomizersConfigurer customizersConfigurer = new CustomizersConfigurer();
```
garyrussell, line 144: With multi-line signatures, we prefer a blank line before the start of the code.

spring-kafka/src/test/java/org/springframework/kafka/config/RetryTopicConfigurationSupportTest.java
```diff
@@ -0,0 +1,329 @@
...
+import org.springframework.kafka.retrytopic.RetryTopicNamesProviderFactory;
+import org.springframework.kafka.support.converter.ConversionException;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+import org.springframework.util.backoff.BackOff;
+
+/**
+ * @author Tomaz Fernandes
+ * @since 2.9
+ */
+class RetryTopicConfigurationSupportTest {
```
garyrussell, line 61: Convention is `...Tests`, even when only one.

garyrussell (changes requested):
This is a big step forward; thanks! Just some minor formatting nits.

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -151,6 +156,42 @@ public KafkaTemplate<String, Object> kafkaTemplate() {
...
+[source, java]
+----
+
+@EnableKafka
+@Import(MyRetryTopicConfiguration.class)
+@Configuration
+public class KafkaConfiguration {
+}
+
+public class MyRetryTopicConfiguration extends RetryTopicConfigurationSupport {
```
artembilan, line 176: I don't think this is natural pattern.
Typically people do just this:
```
@EnableKafka
@Configuration
public class KafkaConfiguration  extends RetryTopicConfigurationSupport {
   ...
}
```

No need in extra inner class.

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -151,6 +156,42 @@ public KafkaTemplate<String, Object> kafkaTemplate() {
...
+
+    @Override
+    protected void manageNonBlockingFatalExceptions(List<Class<? extends Throwable>> nonBlockingFatalExceptions) {
+        nonBlockingFatalExceptions.add(MyNonBlockingException.class);
+    }
+}
+----
+====
+
+NOTE: When using this approach, the `@EnableRetryTopic` annotation is not necessary.
```
artembilan, line 193: It might not be relevant to this PR, but we need to investigate "and what will happen if we use both?".
Separate issue or so.

artembilan (changes requested):

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
tomazfernandes: Oh... 😄 That makes everything so much better, thanks. I'll update the integration tests and docs to reflect this.

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableRetryTopic.java
tomazfernandes: Sure, makes sense. @artembilan and I discussed somewhere about adding this `Support` class or this annotation directly to `@EnableKafka` in 3.0, and thus having this bootstrapping as a framework default. So the purpose of this annotation would be an intermediate step for 2.9.x.

But I think we can have this single `@EnableKafkaRetryTopic ` annotation for now and import the `Support` class directly in the `@EnableKafka` in the future.

WDYT?

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -48,6 +48,11 @@ NOTE: You can have separate `ListenerContainerFactory` instances for the main an
 
 ==== Configuration
 
+IMPORTANT: Since 2.9, the `@EnableRetryTopic` annotation should be used in a `@Configuration` annotated class.
```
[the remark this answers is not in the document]
tomazfernandes: Sure, I can take that out. Thanks.

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
tomazfernandes: Sure, @garyrussell had commented on the same thing before. I'll change this. Thanks.

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
tomazfernandes: In my integration tests they work well together - TBH I don't know in details how this works internally and why the duplicated imports don't trigger an exception for already having the beans. Maybe it's something from the test runtime?

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicConfigurationSupport.java
```diff
@@ -0,0 +1,485 @@
...
+				.acceptIfNotNull(configurer.clock, factory::setClock);
+	}
+
+	/**
+	 * Create the {@link TaskExecutor} instance that will be used with the
+	 * {@link WakingKafkaConsumerTimingAdjuster}, if timing adjustment is enabled.
+	 * @return the instance.
+	 */
+	@Bean(name = BACK_OFF_MANAGER_THREAD_EXECUTOR_BEAN_NAME)
+	public TaskExecutor backoffManagerTaskExecutor() {
```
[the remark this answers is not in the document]
tomazfernandes: Sure, and since this would probably be the only actual breaking change for this PR, that'd be a benefit of rolling this back too. At least we gave it a shot 😄 Thanks.

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
artembilan: As I said: let's revise it in the separate issue!
Yes, it might be the fact that Spring allows to override beans by default.
Spring Boot does not starting with some version.
I have something like this in one test to verify that override cause the problem:
```
public class NoBeansOverrideAnnotationConfigContextLoader extends AnnotationConfigContextLoader {

	@Override
	protected void customizeBeanFactory(DefaultListableBeanFactory beanFactory) {
		beanFactory.setAllowBeanDefinitionOverriding(false);
	}

}
...
@ContextConfiguration(loader = NoBeansOverrideAnnotationConfigContextLoader.class)
```

Or like this to make it explicit in the specific test configuration:
```
@Bean
public static BeanFactoryPostProcessor beanFactoryPostProcessor() {
	return beanFactory -> ((DefaultListableBeanFactory) beanFactory).setAllowBeanDefinitionOverriding(false);
}
```

tomazfernandes: Sure, makes sense. Thanks.

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicComponentFactory.java
tomazfernandes: I think maybe GitHub counts tabs as 8 spaces instead of the 4 I'm used to. The goal was to align the additional parameters to the one in the method declaration - which is kind of how it looks in my IDE.

Either way, the pointer of using two tabs past `public` will save me a lot of moving things around 😄 Thanks.

spring-kafka/src/test/java/org/springframework/kafka/config/RetryTopicConfigurationSupportTest.java
tomazfernandes: Sorry I missed this one. Updated my IDE to suffix new test classes with `Tests` - shouldn't be a problem anymore 😄 Thanks.

tomazfernandes:
> This is a big step forward; thanks! Just some minor formatting nits and one comment.

I appreciate that @garyrussell, thanks. Merit goes to @artembilan for insisting on this approach, and then seeing it through - turned out a lot better than I expected! Also, thanks for your patience, some of this work is sometimes still a bit overwhelming for me. I appreciate the guidance and support from you both.

I've addressed the review comments - please let me know if there's anything else.

Thanks!

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
tomazfernandes: I've changed the text to advise not to use both at the same time - that looks like the safest bet. As you mentioned, we can review this in a separate issue. Thanks.

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableKafkaRetryTopic.java
```diff
@@ -87,5 +82,6 @@
 @Target(ElementType.TYPE)
 @Documented
 @Import(RetryTopicConfigurationSupport.class)
-public @interface EnableRetryTopic {
+@EnableKafka
```
artembilan, line 85: I don't think this has to be there. Or at least it has to be documented properly: 
"For convenience the `@EnableKafkaRetryTopic` is marked also with the `@EnableKafka` since former cannot work without secondary"
(Or some other proper English language 😄 )

spring-kafka/src/main/java/org/springframework/kafka/listener/PartitionPausingBackOffManagerFactory.java
```diff
@@ -161,7 +173,20 @@ private KafkaConsumerTimingAdjuster getOrCreateBackOffTimingAdjustmentManager()
...
+		return this.taskExecutor != null
+				? new WakingKafkaConsumerTimingAdjuster(this.taskExecutor)
+				: new WakingKafkaConsumerTimingAdjuster(createTaskExecutor());
+	}
+
+	private TaskExecutor createTaskExecutor() {
+		ThreadPoolTaskExecutor taskExecutor = new ThreadPoolTaskExecutor();
+		JavaUtils.INSTANCE.acceptIfNotNull(this.maxThreadPoolSize, taskExecutor::setMaxPoolSize);
+		taskExecutor.initialize();
+		super.addApplicationListener((ApplicationListener<ContextClosedEvent>) event -> taskExecutor.shutdown());
```
artembilan, line 188: I still in doubts for all of this stuff around internal executor...
Can you, please, bring the bean back and let's revise it in a separate issue?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/PartitionPausingBackOffManagerFactory.java
tomazfernandes: Sure, no worries, I'll bring it back that and open the issue. Thanks.

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableKafkaRetryTopic.java
artembilan: No, there is no harm: the `@EnableKafka` does a decent work to not duplicate beans. See its target `KafkaBootstrapConfiguration`.
I'm not sure that retry topics should be a default feature.
Therefore keep that `@EnableKafkaRetryTopic` is the way to go from now on.

tomazfernandes: Well, if we're introducing a new annotation then, and probably 99% of the users will be using it in a `Spring Boot` application anyway, I'd stick with the simpler `@EnableRetryTopic` annotation that should be enough for them, and just add a note in the documentation that if users are not in a `Spring Boot` app they should also add `@EnableKafka`.

But no worries, I'll keep it as is and we can revise this down the road if needed.

Thanks.

tomazfernandes: Just as a thought, of course we can look into this later, but what if for 3.0 we added this class to `@EnableKafka`, but also added a custom `@Condition` that would perform that annotation / bean lookup we had in the other PR, to only add this beans it if there's at least one `RetryTopicConfiguration` or `@RetryableTopic` present?

This way we'd keep the automatic feature enabling users currently have, while not encumbering other users with this if they don't use it.

Thanks.

artembilan: Well, that's not how other `@Enable...`  features work. So, my vote for "No": if you are going to use `@RetryableTopic`, then add respective `@Enable...`.
The condition could be considered in Spring Boot which indeed has some opinion.
The framework by itself has to be as straightforward as possible. 
In the end it is just a library with minimal opinion 😉

spring-kafka/src/main/java/org/springframework/kafka/listener/PartitionPausingBackOffManagerFactory.java
tomazfernandes: I've moved it back - how grateful I am for `git` in these moments 😄

Also opened the issue: https://github.com/spring-projects/spring-kafka/issues/2239

Please let me know if there's anything else.

Thanks.

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableKafkaRetryTopic.java
tomazfernandes: Sure, makes sense. As you said in another post, eventually it'll be a good thing to better integrate this with `Spring Boot`'s auto configuration in the future. But since this approach won't be available for `Spring Boot` prior to 3.0, we don't really need to worry about it now.

Thanks.

tomazfernandes:
@artembilan, just a friendly reminder, if nothing has changed, this is supposed to be on Monday's `2.9.0-M1` release, so I wonder if there's anything else to change here or we're good.

Of course, if you're not available to look at this soon enough, or if there's anything else, no worries.

The two points I'm not 100% confident about this are:
* The annotation name - I still think `@EnableRetryTopic` would be cleaner
* An `IMPORTANT:` I added to `PartitionPausingBackOffManagerFactory`'s javadoc regarding the `TaskExecutor` breaking change, since it does not provide one by default anymore. Not sure if it's appropriate to have `IMPORTANT` on a javadoc.

I'm ok with leaving both as is, if you're ok with that. Please let me know if there's anything to be changed.

Thanks.

artembilan:
It is holiday season here for us, but I definitely will look shortly again on Monday if Gary insists on that release.

Nevertheless I agree with Gary that annotation name has to be `@EnableKafkaRetryTopic`.

tomazfernandes:
> It is holiday season here for us, but I definitely will look shortly again on Monday if Gary insists on that release.
> 
> Nevertheless I agree with Gary that annotation name has to be `@EnableKafkaRetryTopic`.

Sure, sorry, I didn't know you were on holidays. Also, after some more consideration, I think I'd be ok with this being merged to 3.0 instead - if you think this is good enough - and we let go of 2.9.

I'll much unfortunately need to change some work arrangements around here and I don't think I'll be able to keep up the pace of involvement with the project I've had for the past few months, so maybe it'd be better to have a longer cycle for this - at least as far as I'm concerned. Of course, that's @garyrussell's call.

It's been a blast though, and thank you both so very much for everything! Of course, feel free to ping me if anything comes up that I can help with - I should be able to at least provide some input.

Thanks again, and happy holidays!

garyrussell:
We can be flexible when we release M1 since nothing depends on it; Monday is just a placeholder.

tomazfernandes:
FYI: https://stackoverflow.com/questions/71977113/when-retryabletopic-annotation-is-used-serializationexception-exception-is-not/71981820#71981820

tomazfernandes:
> We can be flexible when we release M1 since nothing depends on it; Monday is just a placeholder.

Sure. As I mentioned though, I'll really need to shift my focus elsewhere beginning next week - I was counting on friday's deadline for this. It's been more than I month since I've started looking into these configuration issues, so I hope you consider that enough effort on my part. If there's anything else to be done, it would be great if you could just add a `polish commit` to the PR like our friends from `Boot` do - hope that's ok with you.

TBH, I still think we could have come with some simple fix we could merge to `2.8.x`, so we could help out users such as this one on today's `SO` and everyone else on `Boot`'s train, and left this larger change for closer to `3.0` so we could do any breaking changes we wanted to. Anyway, I've learned a lot from it, and I think it'll make it easier for you and everyone else to reason about the feature.

As I said before, feel free to ping me if there's anything I can help with - I should be able to at least provide some input on retryable topic issues - and I will try to help out all around if and when possible.

Thanks a lot for everything, you're both really great software engineers and this has been both an honor and a blast!

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -48,6 +48,11 @@ NOTE: You can have separate `ListenerContainerFactory` instances for the main an
 
 ==== Configuration
 
+Since 2.9, the `@EnableKafkaRetryTopic` annotation should be used in a `@Configuration` annotated class instead of the regular `@EnableKafka` annotation, which it contains.
```
artembilan, line 51: That's not good say like this: brings an impression that we must not use an `@EnableKafka` anymore. 
And you even didn't introduce what is `@EnableKafkaRetryTopic`. The side effect with the `@EnableKafka` could be explained in the end of this paragraph. Probably in the simple "NOTE: When `@EnableKafkaRetryTopic` is used, the `@EnableKafka` can be omitted: the former is meta-annotated with the second."

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -444,32 +480,21 @@ Starting with version 2.8.4, if you wish to add custom headers (in addition to t
 Starting in 2.8.4 you can configure the framework to use both blocking and non-blocking retries in conjunction.
 For example, you can have a set of exceptions that would likely trigger errors on the next records as well, such as `DatabaseAccessException`, so you can retry the same record a few times before sending it to the retry topic, or straight to the DLT.
 
-To configure blocking retries you just need to add the exceptions you want to retry through the `addRetryableExceptions` method in the `ListenerContainerFactoryConfigurer` bean as follows.
-The default policy is `FixedBackOff`, with nine retries and no delay between them.
-Optionally, you can provide your own back off policy.
+To configure blocking retries, override the `configureBlockingRetries` method in a `@Configuration` class that extends `RetryTopicConfigurationSupport` and add the exceptions you want to retry, along with the `BackOff` to be used.
+The default `BackOff` is a `FixedBackOff` with no delay and 9 attempts.
```
artembilan, line 484: Why `9 attempts`? 
Typically it is `3`: `SimpleRetryPolicy.DEFAULT_MAX_ATTEMPTS`...

spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableKafkaRetryTopic.java
artembilan: I still want to see some mentioning of this meta-annotation in the Javadocs of this `EnableKafkaRetryTopic` class.

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicComponentFactory.java
```diff
@@ -0,0 +1,200 @@
...
+		return new PartitionPausingBackOffManagerFactory(registry);
+	}
+
+	/**
+	 * Create the {@link TaskExecutor} that will be used in the
+	 * {@link KafkaConsumerTimingAdjuster}.
+	 * @return the task executor.
+	 */
+	public TaskExecutor taskExecutor() {
+		return new ThreadPoolTaskExecutor();
```
artembilan, line 179: Don't we provide a bean now?
What is the point of this factory method if end-user will have to just provide a respective bean to override a default one?

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicConfigurationSupport.java
```diff
@@ -0,0 +1,486 @@
...
+
+/**
+ * This is the main class providing the configuration behind the non-blocking,
+ * topic-based delayed retries feature. It is typically imported by adding
+ * {@link EnableKafkaRetryTopic @EnableRetryTopic} to an application
+ * {@link Configuration @Configuration} class. An alternative more advanced option
+ * is to extend directly from this class and override methods as necessary, remembering
+ * to add {@link Configuration @Configuration} to the subclass and {@link Bean @Bean}
+ * to overridden {@link Bean @Bean} methods. For more details see the javadoc of
+ * {@link EnableKafkaRetryTopic @EnableRetryTopic}.
```
artembilan, line 73: Some typo. Or missed renaming...

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicConfigurationSupport.java
```diff
@@ -0,0 +1,486 @@
...
+	 * @return the instance.
+	 */
+	@Bean(name = BACK_OFF_MANAGER_THREAD_EXECUTOR_BEAN_NAME)
+	public TaskExecutor backoffManagerTaskExecutor() {
+		KafkaBackOffManagerConfigurer configurer = new KafkaBackOffManagerConfigurer();
+		configureKafkaBackOffManager(configurer);
+		return configurer.timingAdjustmentEnabled
+				? this.componentFactory.taskExecutor()
+				: task -> {
+				};
```
artembilan, line 329: You can return `null` here instead if we just don't need such a bean.

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicConfigurationSupport.java
```diff
@@ -0,0 +1,486 @@
...
+	}
+
+	/**
+	 * Configure customizers for components instantiated by the retry topics feature.
+	 */
+	public static class CustomizersConfigurer {
+
+		private Consumer<CommonErrorHandler> errorHandlerCustomizer;
+		private Consumer<ConcurrentMessageListenerContainer<?, ?>> listenerContainerCustomizer;
+		private Consumer<DeadLetterPublishingRecoverer> deadLetterPublishingRecovererCustomizer;
```
artembilan, line 449: Every member of the class has to be surrounded with blank lines (including the last one in the class).

artembilan (changes requested):

tomazfernandes:
Hi. Can anyone be so kind to explain to me again why we need the `@EnableKafkaRetryTopic` instead of `@EnableRetryTopic`, considering most users will be on `Boot` and will already have `@EnableKafka` by default? Aren't we optimizing for the few rather than for the many? I'm sure I'm missing something - would like to know what though.

Thanks.

artembilan:
Because this is more Apache Kafka feature and it has nothing to do with JMS or AMQP (or even Hazelcast) topics. So, such a generic name might lead to the confusion for someone who is new to Kafka, but would rather expect a feature for his/her non-Kafka topics.

tomazfernandes:
> Because this is more Apache Kafka feature and it has nothing to do with JMS or AMQP (or even Hazelcast) topics. So, such a generic name might lead to the confusion for someone who is new to Kafka, but would rather expect a feature for his/her non-Kafka topics.

Sure, that sounds reasonable, thanks.

garyrussell:
@tomazfernandes My apologies; I have not been closely involved with this PR; I thought it was generally to make it easier to customize the configuration of the feature. Which random, non-intelligible errors do you mean?

garyrussell:
To solve the lazy bean creation, we should be able to backport just the `@EnableKafkaRetryTopics` to force creation of the legacy `RetryTopicBootstrapper` earlier in the application lifecycle.

Am I missing something?

garyrussell:
You even suggested that [here](https://github.com/spring-projects/spring-kafka/issues/2172#issuecomment-1068532530).

tomazfernandes:
I’m not sure I can visualize exactly what’s on your mind. From RT’s perspective this should work. My concerns would be:

- Legacy bootstrapper uses a lot of factory.getBean; not sure if that’s a proper moment of the lifecycle
- If the bootstrapper bean creation is lazy, it might actually bootstrap at the same time as of now, so it should probably  be made eager
- It’d still need to be executed before other configuration classes, so maybe it should have a higher priority order

tomazfernandes:
The other options I see would be:

- Adding default constructors to configurable components, making them beanFactoryAware, and injecting the dependencies manually afterSingletonsInstantiated. This way users can simply declare the bean with their proper configuration without worrying with dependencies. We can deprecate this on 2.9 and remove on 3.0.

- Lightening up this PR a bit by removing new bean names and rolling back task executor related things, and merging it to 2.8.x. I think by doing this there wouldn’t be any changes to existing logic at all, except maybe for some new constructors we can also take out if needed, and since it’s opt in, there should be no risk for users

artembilan:
Right. That was one my biggest concerns for the original solution: end-user beans cannot contribute to the infrastructure if they also depend on some infrastructure. 
It is going to be much worse when we come to Spring Boot's auto-configuration which will report for us "circular dependency" or "no bean yet" errors.
And that's why I suggested to look into `@EnableWebMvc` from the beginning.
Anyway the cat is out of the bag...

I believe we stated somewhere that retriable topics is an experimental feature. So, it is expected that something is not going to work.

Therefore an `afterSingletonsInstantiated()` might really be a good middle ground for `2.8.x`.

I still against pushing this PR back to `2.8.x`...

garyrussell:
I agree; that's not up for debate; I was just looking to see if there was some middle ground for 2.8.x users; if not, so be it.

tomazfernandes:
Thanks a lot for your answers, I really appreciate that. 

I do think that, if something can be done to reduce user frustration on the `2.8.x` branch, which is `Boot`'s official branch for the year, it's worth giving it a shot, even if only out of empathy for them.

Also, I think that this became a bigger issue mostly because I showed up out of the blue a few months ago, started pushing code, and by the time we realized we'd need a new minor version it was too late to get into `Boot`'s train. Otherwise, we'd probably have merged this to `2.9.0-RC1`, and it'd all be good. Under different circumstances we'd surely have been able to plan all this better.

Another important thing for us to figure out is whether this is a production-ready, robust and complete feature, as the experimental disclaimer states, and that deserves proper support as such, or if it's a fundamentally flawed experimental feature that the only worthy thing to do is to focus on rebuilding it for `3.0`. I'm ok with both views, but being in the middle of them without proper discussion was really frustrating.

I think we're on a good track overall, we did some really nice work so far, and as we get to know each other better we should be able to reduce friction in complex issues such as this. I'll try to look into this PR sometime later this week, or early next. Or, if you prefer, you can add that polish commit to release `2.9-M1` sooner. 

As I've stated, I shouldn't be able to focus as much on this project as I had been lately. But I'll try to help out as much as I can. Either way, I'm deeply grateful for all we've been able to do so far, and please let me know if there's anything else I can help with.

Thanks!

garyrussell:
Thanks @tomazfernandes I will take care of @artembilan 's latest comments in a polishing commit; I plan on releasing 2.9.0-M1 tomorrow, now.

Thanks for all your hard work on this.

spring-kafka/src/main/java/org/springframework/kafka/config/RetryTopicConfigurationSupport.java
garyrussell: We can't do that because it is a required dependency in the `kafkaConsumerBackoffManager` bean.

garyrussell:
Polishing commit: https://github.com/garyrussell/spring-kafka/commit/db68ab79c3ef0f72886d1bc21d3739d927bdc03e

garyrussell:
Merged as b2beec310d6a6a6d685bec106f3815b03d72d617 and cherry-picked to 2.9.x

Issue #2240, opened 2022-04-21
KafkaTemplate receive(Collection<TopicPartitionOffset> requested, Duration pollTimeout) can have nulls


gurpiarbassi:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

This is on main branch

**Describe the bug**

KafkaTemplate has a method:
```
      @Override
  	@Nullable
	public ConsumerRecord<K, V> receive(String topic, int partition, long offset, Duration pollTimeout) {
		Properties props = oneOnly();
		try (Consumer<K, V> consumer = this.consumerFactory.createConsumer(null, null, null, props)) {
			TopicPartition topicPartition = new TopicPartition(topic, partition);
			return receiveOne(topicPartition, offset, pollTimeout, consumer);
		}
	}
```

This method correctly states that the return value could be null via @Nullable and the javadoc also states the same.

However there is another method:
```
@Override
	public ConsumerRecords<K, V> receive(Collection<TopicPartitionOffset> requested, Duration pollTimeout) {
		Properties props = oneOnly();
		Map<TopicPartition, List<ConsumerRecord<K, V>>> records = new LinkedHashMap<>();
		try (Consumer<K, V> consumer = this.consumerFactory.createConsumer(null, null, null, props)) {
			requested.forEach(tpo -> {
				ConsumerRecord<K, V> one = receiveOne(tpo.getTopicPartition(), tpo.getOffset(), pollTimeout, consumer);
				records.computeIfAbsent(tpo.getTopicPartition(), tp -> new ArrayList<>()).add(one);
			});
			return new ConsumerRecords<>(records);
		}
	}
```
This method returns a ConsumerRecords object however the ConsumerRecord itself could be null. Therefore you end up with a collection with null inside it. 

**To Reproduce**
I reproduced it by calling `receive(Collection<TopicPartitionOffset> requested, Duration pollTimeout)` on a brand new topic that was empty. The javadoc doesn't mention the elements could be null.

Could we do a null check before add the ConsumerRecord to the ConsumerRecord**s** object?

Furthermore `tpo.getOffset()` can also return null which leads to an auto unboxing issue.

artembilan:
I think this is correct observation.

Would you mind to contribute the fix: https://github.com/spring-projects/spring-kafka/blob/main/CONTRIBUTING.adoc ?

Thanks

gurpiarbassi:
> I think this is correct observation.
> 
[...]

PR #2242, opened 2022-04-21
GH-2240 Bug fix for KafkaTemplate.receive(..)

gurpiarbassi:
- Ensures that we dont get ConsumerRecords with nulls in the list
- Ensures that we do not allow null offsets


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -569,9 +571,17 @@ public ConsumerRecords<K, V> receive(Collection<TopicPartitionOffset> requested,
 		Properties props = oneOnly();
 		Map<TopicPartition, List<ConsumerRecord<K, V>>> records = new LinkedHashMap<>();
 		try (Consumer<K, V> consumer = this.consumerFactory.createConsumer(null, null, null, props)) {
+			AtomicInteger index = new AtomicInteger(0);
 			requested.forEach(tpo -> {
+				if (tpo.getOffset() == null) {
+					throw new KafkaException(String.format("Offset supplied in TopicPartitionOffset [%d] is null", index.get()));
```
artembilan, line 577: I think we just can use a `tpo` in the exception message to indicate exactly which one is wrong.
And therefore no need in that `index` variable to track.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
gurpiarbassi: done

gurpiarbassi: @artembilan @garyrussell  please could you review when you get a chance?
Thanks.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -570,8 +570,14 @@ public ConsumerRecords<K, V> receive(Collection<TopicPartitionOffset> requested,
 		Map<TopicPartition, List<ConsumerRecord<K, V>>> records = new LinkedHashMap<>();
 		try (Consumer<K, V> consumer = this.consumerFactory.createConsumer(null, null, null, props)) {
 			requested.forEach(tpo -> {
+				if (tpo.getOffset() == null) {
+					throw new KafkaException(String.format("Offset supplied in TopicPartitionOffset is null: %s", tpo));
```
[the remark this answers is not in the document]
garyrussell: I am not too concerned about performance here, since this is an error case; however, we should also reject negative values.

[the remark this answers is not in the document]
gurpiarbassi: both points actioned.

spring-kafka/src/test/java/org/springframework/kafka/core/KafkaTemplateTests.java
```diff
@@ -621,4 +630,22 @@ void testCompositeProducerInterceptor() {
 		inOrder.verify(producerInterceptor2).onAcknowledgement(any(RecordMetadata.class), Mockito.isNull());
 	}
 
+	@ParameterizedTest(name = "{0} is invalid")
+	@NullSource
+	@ValueSource(longs =  -1)
```
artembilan, line 635: That's cool! 👍

artembilan (approved):
Nothing to complain about from my side any more.

Thank you!

garyrussell:
...and cherry-picked after resolving conflicts to 2.9.x, 2.8.x.

@gurpiarbassi Thanks for the contribution. 

In future, please don't force push (unless you rebase), just add more commits to the PR so we can see what changed more easily (not really an issue with this small PR, but it is with larger PRs).

Also, please leave a blank line after the commit headline.

PR #2245, opened 2022-04-28
GH-2172: Expose Retry Topic Chain at Runtime

tomazfernandes:
Resolves https://github.com/spring-projects/spring-kafka/issues/2172

* Add methods to `DestinationTopicContainer` interface
* Reduce memory footprint of `DefaultDestinationTopicResolver`
* Add documentation

This has been ready for quite a while and I was only waiting for us to solve the dependency injection issue to push it. Pretty much only exposed logic that was already there and renamed a few methods accordingly.

Also, there was a `Map` in `DefaultDestinationTopicResolver` that wasn't necessary, so I removed it.

Thanks.


garyrussell:
...and cherry-picked to 2.9.x

PR #2246, opened 2022-04-28
Annotation TopicPartition - partition attribute was not resolved from…

s7474:
Processing of TopicPartition annotation was not resolving partition property if it was set to wildcard (*) via SpEL


artembilan (changes requested):
Any chances to confirm the issue with some unit test?

Thanks

Issue #2249, opened 2022-04-29
The listener info header is not added when using a batch listener


frosiere:
The issue #2134 introduced a way to set arbitrary information using the info property of the KafkaListener annotation. This info is then added as a header on the ConsumerRecord.

When using a batch listener, this header is not added on the records.

Based on the documentation (https://docs.spring.io/spring-kafka/docs/current/reference/html/#li-header), the header is only available in the record interceptor but it would certainly make sense to also have it in when using a batch interceptor.

The internal headers are currently added in KafkaMessageListenerContainer#internalHeaders but this method is only called for single record listeners (see KafkaMessageListenerContainer#checkEarlyIntercept).

Batches are handled in KafkaMessageListenerContainer#invokeBatchListener. So, the interceptor could be called after having iterated on the list of records to allow adding the info header.

frosiere:
Thanks a lot for the very quick fix.

PR #2257, opened 2022-05-04
GH-2249: Batch Listener LISTENER_INFO Headers

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2249

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -2239,6 +2239,9 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 
 			ConsumerRecords<K, V> records = recordsArg;
 			List<ConsumerRecord<K, V>> recordList = recordListArg;
+			if (this.listenerinfo != null) {
+				records.iterator().forEachRemaining(rec -> listenerInfo(rec));
+			}
 			if (this.batchInterceptor != null) {
 				records = this.batchInterceptor.intercept(recordsArg, this.consumer);
 				if (records == null) {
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2475,10 +2478,14 @@ private void internalHeaders(final ConsumerRecord<K, V> record) {
 				record.headers().add(new RecordHeader(KafkaHeaders.DELIVERY_ATTEMPT, buff));
 			}
 			if (this.listenerinfo != null) {
-				record.headers().add(this.infoHeader);
+				listenerInfo(record);
 			}
 		}
 
+		private void listenerInfo(final ConsumerRecord<K, V> record) {
+			record.headers().add(this.infoHeader);
```
artembilan, line 2486: Looks like we add the same info to all the records.
What is the reason of that `@Header(KafkaHeaders.LISTENER_INFO) List<String> infos`?
Cannot it be just one single header for the message based on a batch?
I don't mind having it in every single record of the batch we produce, but I doubt in a list of the same info...

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
```diff
@@ -205,8 +204,8 @@ public Message<?> toMessage(List<ConsumerRecord<?, ?>> records, @Nullable Acknow
 				raws.add(record);
 			}
 		}
-		if (infos.size() > 0) {
-			rawHeaders.put(KafkaHeaders.LISTENER_INFO, infos);
+		if (info != null) {
```
artembilan, line 207: May it happen that batch is empty, but we still want to see such an info?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
garyrussell: The container will never submit an empty batch (unless the user has an interceptor that returns one). It could also be the result of applying a filter.

In either case, there is no way to get the info if there are no records to convert. I added a note to the docs.

artembilan: OK. I see now:
```
		private void invokeBatchListener(final ConsumerRecords<K, V> recordsArg) {
			ConsumerRecords<K, V> records = checkEarlyIntercept(recordsArg);
			if (records == null || records.count() == 0) {
				return;
			}
```

Thanks

artembilan (approved):

artembilan:

Issue #2262, opened 2022-05-06
Add BackOffHandler to FailedRecordTracker


Marincek:
**Expected Behavior**

BackOffHanlder should encapsulate the handling for the next backoff. FailedRecordTracker will use this interface and call handleNextBackoff in recovered method.

**Current Behavior**

Currently FailedRecordTracker it is tightly coupled with thread sleep. There is no way to inject another behavior.

**Context**

The idea is to make FailedRecordTracker able to handle the backoff in different ways. In my case I need to pause the listener on XX minutes. 
The other solution for me is to copy several classes just to change one line (FailedRecordTracker,FailedRecordProcessor, DefaultAfterRollbackProcessor)

Marincek:
PR : https://github.com/spring-projects/spring-kafka/pull/2263

PR #2263, opened 2022-05-06
Add BackOffHandler to FailedRecordTracker

Marincek:
Added new BackOffHandler interface. The idea is to decouple the FailedRecordTracker with the behavior on next back off time. 
Currently the FailedRecordTracker is sleeping the thread, and there is now way to change that behaviour.

Related issue : https://github.com/spring-projects/spring-kafka/issues/2262

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/BackOffHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/BackOffHandler.java
@@ -9,18 +9,9 @@ import org.springframework.lang.Nullable;
  *
  *  @since 1.3.5
  */
+@FunctionalInterface
 public interface BackOffHandler {
 
-	default void onNextBackOff(@Nullable MessageListenerContainer container, Exception exception, long nextBackOff) {
-		try {
-			if (container == null) {
-				Thread.sleep(nextBackOff);
-			} else {
-				ListenerUtils.stoppableSleep(container, nextBackOff);
-			}
-		} catch (InterruptedException e) {
-			throw new RuntimeException(e);
-		}
-	}
+	void onNextBackOff(@Nullable MessageListenerContainer container, Exception exception, long nextBackOff);
 
 }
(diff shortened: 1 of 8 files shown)


selalerercapitolis (approved):

artembilan (changes requested):
I wonder if existing `KafkaConsumerBackoffManager` abstraction can work here as well, without introducing anything else...

@tomazfernandes, WDYT ?

garyrussell:
@artembilan That is for pausing a single partition, in this case we need to pause the entire container.

@Marincek Are you planning on providing the implementation that pauses and resumes the container?

tomazfernandes:
> I wonder if existing `KafkaConsumerBackoffManager` abstraction can work here as well, without introducing anything else...
> 
> @tomazfernandes, WDYT ?

Missed this notification, sorry @artembilan. As @garyrussell pointed out, the `KafkaConsumerBackOffManager` is meant to back off a single partition, but we probably could just retrieve all assignments for that container and loop through them. 

One important thing is we need to make sure a `idlePartitionEventInterval` is set for the container - maybe we could look into providing a default value for that.

Now that you brought this up, and considering this `pause until` pattern appears here and there, I think in the future we might look into leveraging the `BackOffManager` to handle all such logic - the `nack` feature comes to mind as well for example - so that we don't need to implement it over and over again and have a centralized place to handle all partitions. Maybe we could auto configure a `BackOffManager` bean for that and inject where needed.

But I defer to @garyrussell what would be the proper solution for this particular case.

garyrussell:
@Marincek use `./gradlew check` to build the project (including running checkstyle).

garyrussell:
Sorry for the delay; I have squashed and merged your commits to this branch https://github.com/garyrussell/spring-kafka/tree/GH-2262

with this polishing commit: https://github.com/garyrussell/spring-kafka/commit/17f8f85d8c532268fa17ba506bfd23be388f5389

I will merge both after my commit is reviewed by @artembilan

garyrussell:
Merged and cherry-picked; thanks for the contribution!

Issue #2268, opened 2022-05-11
DeadLetterPublishingRecoverer should respect onlyLogRecordMatadata listener property


kdowbecki:
**Expected Behavior**

Currently the `org.springframework.kafka.listener.KafkaMessageListenerContainer` class has `onlyLogRecordMatadata` property that governs how `ConsumerRecord` is logged on processing failure. Same `onlyLogRecordMatadata` property should be respected by the `org.springframework.kafka.listener.DeadLetterPublishingRecoverer` class for `ProducerRecord`.

**Current Behavior**

`DeadLetterPublishingRecoverer` currently logs the whole `ProducerRecord` when publication fails with no option to switch to just the metadata.

**Context**

We already have set `onlyLogRecordMatadata = true` because we don't want the message payload to appear in the logs when something fails. Since `DeadLetterPublishingRecoverer` copies the payload from the failed message into a `ProducerRecord` we still see the original message payload in the logs when publication fails.

garyrussell:
Which version are you using? And which log message specifically?

It looks like all logs in 2.8.x properly log the record; in 2.7.x, I see one debug and 2 exceptions that log the full record.

garyrussell:
That flag only controls logging of the consumer record.

The DLPR does log the producer record; in 2.8 we added a new feature to format the logs for producer record.

https://github.com/spring-projects/spring-kafka/blob/f94620aea7e7b43416b67cf9a46dce0acf46aa95/spring-kafka/src/main/java/org/springframework/kafka/support/KafkaUtils.java#L168-L176

The default formatter logs the full record.

garyrussell:
Although it was back-ported to 2.7.x, the DLPR doesn't use it.

kdowbecki:
Good point @garyrussell. I'm on 2.7.12, inherited from Spring Boot 2.5.13. I don't see `DeadLetterPublishingRecoverer` using `KafkaUtils` formatting in these lines:

https://github.com/spring-projects/spring-kafka/blob/85cbbed1a28aca28a2c4c9e5dfc698d25cec7d9b/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java#L501

https://github.com/spring-projects/spring-kafka/blob/85cbbed1a28aca28a2c4c9e5dfc698d25cec7d9b/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java#L525

[...]

PR #2271, opened 2022-05-11
GH-2268: Fix DLPR Record Logging

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2268

2.7.x Only

The DLPR was incorrectly logging the full `ProducerRecord`; later versions
log the original `ConsumerRecord` using `KafkaUtils.format()`.
Backport the logging changes, but use a `ThreadLocal` for the input record
to avoid API changes in a patch release.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
@@ -75,6 +75,8 @@ public class DeadLetterPublishingRecoverer implements ConsumerAwareRecordRecover
 
 	private static final long THIRTY = 30L;
 
+	private static final ThreadLocal<ConsumerRecord<?, ?>> IN_RECORD = new ThreadLocal<>();
+
 	private final HeaderNames headerNames = getHeaderNames();
 
 	private final boolean transactional;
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -75,6 +75,8 @@ public class DeadLetterPublishingRecoverer implements ConsumerAwareRecordRecover
 
 	private static final long THIRTY = 30L;
 
+	private static final ThreadLocal<ConsumerRecord<?, ?>> IN_RECORD = new ThreadLocal<>();
```
artembilan, line 78: May be not `static` for cases when several `DeadLetterPublishingRecoverer` can be used in the project?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
artembilan: Yeah... Sure. I realized that later on.
Thanks for confirmation.

Merging...

PR #2278, opened 2022-05-18
Precise unit for sleep duration and wake time

szpak:
Prior to, it was required to dig a few levels into to realize they are in millis.

I'm not satisfied with the new names of the field, but I wanted to express that `nackSleep` is a duration and `nackWake` is a point of time (the same as `currentTimeMillis()`). Nevertheless, as they are used internally, it might be sufficient to just change `sleep` to `sleepMillis` in the method argument (+ change the javadoc).

`java.time.Duriation` in the `Acknowledgment` class would be probably better, however, it is a non-backward compatible change.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
@@ -37,35 +37,28 @@ public interface Acknowledgment {
 	/**
 	 * Negatively acknowledge the current record - discard remaining records from the poll
 	 * and re-seek all partitions so that this record will be redelivered after the sleep
-	 * time. Must be called on the consumer thread.
+	 * time (in milliseconds). Must be called on the consumer thread.
 	 * <p>
-	 * <b>When using group management,
-	 * {@code sleep + time spent processing the previous messages from the poll} must be
-	 * less than the consumer {@code max.poll.interval.ms} property, to avoid a
-	 * rebalance.</b>
-	 * @param sleep the time to sleep; the actual sleep time will be larger of this value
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
```diff
@@ -43,11 +43,11 @@ public interface Acknowledgment {
 	 * {@code sleep + time spent processing the previous messages from the poll} must be
 	 * less than the consumer {@code max.poll.interval.ms} property, to avoid a
 	 * rebalance.</b>
```
szpak, line 45: Btw, is it still true after the `sleep()` was removed in https://github.com/spring-projects/spring-kafka/pull/2131?

garyrussell: It is no longer true; those sentences can be removed.

szpak: I removed the sentence and precised millis also in the generic JavaDoc descirption.

szpak:
> Given that the nack methods are `default` methods; I would not object to adding new overloaded methods taking a duration and deprecate the current ones.

Great. However, as it's a slightly more complicated change, so I propose to merge this one as is (with "millis" only) and I will create a new with `Duration` later on to keep it separate.

garyrussell:
Thanks; agreed; in future, please just add more commits rather than squashing and force-pushing; it's easier to see what changed (not really a problem for small PRs like this, but important for larger ones). We will squash during the merge.

garyrussell (approved):

PR #2281, opened 2022-05-19
Add Acknowledgment.nack() variants accepting Duration

szpak:
To prevent confusion what unit should be used for "long sleep" arguments.
As the old and new methods in the Acknowledgment interface have default
implementations, the change itself is backward compatible.

The old methods are marked as deprecated and intended to be removed in the future.

I wonder, if we should also test old, deprecated methods. On one hand, they should be to "ensure" they still work fine. On the other hand the logic there is rather obvious and covering them would probably duplicate 2 tests. WDYT?

An extension to #2278.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
@@ -39,21 +39,6 @@ public interface Acknowledgment {
 	/**
 	 * Negatively acknowledge the current record - discard remaining records from the poll
 	 * and re-seek all partitions so that this record will be redelivered after the sleep
-	 * time (in milliseconds). Must be called on the consumer thread.
-	 * <p>
-	 * @param sleepMillis the time to sleep in milliseconds; the actual sleep time will be larger
-	 * of this value and the container's {@code maxPollInterval}, which defaults to 5 seconds.
-	 * @since 2.3
-	 * @deprecated in favor of {@link #nack(Duration)}
-	 */
-	@Deprecated(since = "2.8.7", forRemoval = true)
-	default void nack(long sleepMillis) {
-		nack(Duration.ofMillis(sleepMillis));
-	}
-
-	/**
-	 * Negatively acknowledge the current record - discard remaining records from the poll
-	 * and re-seek all partitions so that this record will be redelivered after the sleep
 	 * duration. Must be called on the consumer thread.
 	 * <p>
 	 * @param sleep the duration to sleep; the actual sleep time will be larger of this value
@@ -62,23 +47,6 @@ public interface Acknowledgment {
 	 */
 	default void nack(Duration sleep) {
 		throw new UnsupportedOperationException("nack(sleep) is not supported by this Acknowledgment");
-	}
-
-	/**
-	 * Negatively acknowledge the record at an index in a batch - commit the offset(s) of
-	 * records before the index and re-seek the partitions so that the record at the index
-	 * and subsequent records will be redelivered after the sleep time (in milliseconds).
-	 * Must be called on the consumer thread.
-	 * <p>
-	 * @param index the index of the failed record in the batch.
-	 * @param sleepMillis the time to sleep in milliseconds; the actual sleep time will be larger
-	 * of this value and the container's {@code maxPollInterval}, which defaults to 5 seconds.
-	 * @since 2.3
-	 * @deprecated in favor of {@link #nack(int, Duration)}
-	 */
-	@Deprecated(since = "2.8.7", forRemoval = true)
-	default void nack(int index, long sleepMillis) {
-		nack(index, Duration.ofMillis(sleepMillis));
 	}
 
 	/**
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
@@ -16,6 +16,8 @@
 
 package org.springframework.kafka.support;
 
+import java.time.Duration;
+
 /**
  * Handle for acknowledging the processing of a
  * {@link org.apache.kafka.clients.consumer.ConsumerRecord}. Recipients can store the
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
```diff
@@ -57,8 +74,25 @@ default void nack(long sleepMillis) {
 	 * @param sleepMillis the time to sleep in milliseconds; the actual sleep time will be larger
 	 * of this value and the container's {@code maxPollInterval}, which defaults to 5 seconds.
 	 * @since 2.3
+	 * @deprecated in favor of {@link #nack(int, Duration)}
 	 */
+	@Deprecated(since = "2.8.7", forRemoval = true)
```
szpak, line 79: For backports, `since` and `forRemoval` will be a problem. On the other hand, IDE generated a warning without them in 3.x. They could be dropped during backporting or removed also in master. WDYT?

spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
```diff
@@ -42,8 +44,23 @@ public interface Acknowledgment {
...
+		nack(Duration.ofMillis(sleepMillis));
+	}
+
+	/**
+	 * Negatively acknowledge the current record - discard remaining records from the poll
+	 * and re-seek all partitions so that this record will be redelivered after the sleep
+	 * duration. Must be called on the consumer thread.
+	 * <p>
+	 * @param sleep the duration to sleep; the actual sleep time will be larger of this value
+	 * and the container's {@code maxPollInterval}, which defaults to 5 seconds.
```
szpak, line 60: It's unrelated to that PR, however, yesterday I was very curious why in my integration tests the `nack(3000)` still checks every 5 seconds, even though, I set `maxPollInterval` to 1 seconds (and now also `idleBetweenPolls` to 0 which should be default). Today, checking if it is worth to change also the internal fields from `long` to `Instant`, I have found:
https://github.com/spring-projects/spring-kafka/blob/93f5ae5c7cc65e9ebf1b2e310fdeb106bb04ed51/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1722-L1723

I wonder, if that could have some impact, or maybe you see any other reason why I cannot enforce polling every 1 or 2 seconds (as `maxPollInterval` is set)?

Here an example with `nack(1991)` (precisely with `PT1.991779718S` to millis) and `max.poll.interval.ms = 750`:


```
2022-05-20 00:08:13.992 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Pausing for nack sleep: [pusher_retry_1-0]
2022-05-20 00:08:13.992 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Commit list: {}
2022-05-20 00:08:13.992 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Still paused for nack sleep
2022-05-20 00:08:18.142 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Received: 0 records
2022-05-20 00:08:18.143 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Commit list: {}
2022-05-20 00:08:18.146 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Received: 0 records
2022-05-20 00:08:18.146 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Commit list: {}
2022-05-20 00:08:18.227 DEBUG foobar --- [ntainer#0-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Received: 0 records
2022-05-20 00:08:18.227 DEBUG foobar --- [ntainer#0-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Commit list: {}
2022-05-20 00:08:18.234 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Received: 0 records
2022-05-20 00:08:18.234 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Commit list: {}
2022-05-20 00:08:18.927 DEBUG foobar --- [ntainer#1-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Received: 0 records
2022-05-20 00:08:18.927 DEBUG foobar --- [ntainer#1-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Commit list: {}
2022-05-20 00:08:18.975 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Received: 0 records
2022-05-20 00:08:18.975 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Commit list: {}
2022-05-20 00:08:18.992 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Received: 0 records
2022-05-20 00:08:18.992 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Resumed after nack sleep: [pusher_retry_1-0]
2022-05-20 00:08:18.993 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Commit list: {}
2022-05-20 00:08:18.993 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Received: 0 records
2022-05-20 00:08:18.993 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Commit list: {}
2022-05-20 00:08:18.997 DEBUG foobar --- [Container-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Received: 1 records
2022-05-20 00:08:18.998 DEBUG foobar --- [Container-0-C-1] .a.RecordMessagingMessageListenerAdapter : Processing [GenericMessage [payload=MyMessage.....]
```

szpak: @garyrussell What do you think about that case (and the PR in general ;-) )?

spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
garyrussell: We should make a separate PR for 2.8.x and actually remove the method in 3.0.

spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
garyrussell: It looks like you are using non-blocking retries; the precision there is also affected by the `idleEventInterval`.

https://docs.spring.io/spring-kafka/docs/current/reference/html/#back-off-delay-precision

garyrussell: Hmmm - but that shouldn't be the case here; will investigate...

spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
szpak: I created #2298 which should be JDK 8+ compatible and remove the methods for 3.0.

spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
szpak: Interesting :confused: . I will try to re-check my code with the latest version of spring-kafka.

wilerson:
Should this be backported into 2.9.x ? I had a whiplash when upgrading from 2.8.8

garyrussell:
No; we can't make breaking changes in a patch release; if you want it back ported, we would have to reinstate the old variants (and deprecate them).

wilerson:
> we would have to reinstate the old variants (and deprecate them).

Wasn't that done in 2.8.x? https://github.com/spring-projects/spring-kafka/pull/2298

Issue #2295, opened 2022-06-06
The addition of a  parameter of type `ConsumerRecordMetaData` breaks a working setup


kai-moritz:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

2.6.8

**Describe the bug**

I have written a consumer, that consumes multiple types of messages.
The consumer is annotated with `@KafkaListener` at class-level.
The handler-methods for the known messate-types are annotated with `@KafkaHandler`.
The methods only a parameter of the type of the expected message.

_Everything works fine..._

After adding an additional parameter of type `ConsumerRecordMetaData`, the consumer is no longer able to receive messages of any type, because no appropriate resolvers could be found.
In fact, debugging shows, that not reolvers at all are configured, when the handler-method is called.

After adding a default-method, that handles an messages of unknown type, everything works as expected again.

**To Reproduce**

* [Without an additional parameter of type `ConsumerRecordMetaData` everything works as expexted](https://github.com/juplo/demos-kafka-kafkahandler/releases/tag/works-with-multiple-message-types)
* [After the addtion of the parameter, the messages are no longer resolved](https://github.com/juplo/demos-kafka-kafkahandler/releases/tag/fails-with-metadata)
* [Adding a default-handler "resolves" the issue](https://github.com/juplo/demos-kafka-kafkahandler/releases/tag/default-handler-as-fix)

**Expected behavior**

If receiving of messages of multiple types works without a default-handler for messages of unknown type, the same setup should work without a default-handler.
If the handler-methods also require an additional parameter of type `ConsumerRecordMetaData`, because they require knowledge about the partition, offset or any other technical detail, this should not force the implementation of an additional handler-method for unknown messages.

**Sample**

[Example with a working setup, the broken setup, if metadata are required, and the crooked workaround](https://github.com/juplo/demos-kafka-kafkahandler)

garyrussell:
Strange behavior indeed; thanks for the reproducer.

PR #2296, opened 2022-06-06
GH-2295: No Resolvers with ConsumerRecordMetadata

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2295

A subclass of `InvocableHandlerMethod` was used to signal that a method
needs a `ConsumerRecordMetadata` to be constructed.

However, the constructor that takes an existing IHM, does not fully clone
the argument passed into it.

Then, the payload resolver was not invoked to resolve that parameter.

Use a map of booleans instead, to signal the creation of the metadata.

**cherry-pick to 2.9.x, 2.8.x, 2.7.x**


artembilan:
... and cherry-picked to all mentioned supporting versions.

PR #2300, opened 2022-06-07
in fovour of ConverterFactory

wruiiwang:
ConverterFactory converters are not supported by default。


garyrussell:
Thanks, merged as a07ace92 and cherry-picked to 2.9.x, 2.8.x

Issue #2297, opened 2022-06-06
MessageListenerContainer is started even if autoStartup is false


tbarabanov:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**
2.8,2

**Describe the bug**
I have a class like this: 
`@Autowired`
`private KafkaListenerEndpointRegistry registry;`

`@KafkaListener(`
`            id = "#{__listener.id}",`
`            groupId = "#{__listener.groupId}",`
`            topics = "#{__listener.topics}",`
`            properties = {"#{__listener.properties}"},`
`           autoStartup = "false") // autostart is disabled, use listener.start()`

Instances of the class are created this way:
`@Bean`
`@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)`
`public KafkaMessageListener<?, ?> kafkaListener(String groupId,`
`                                  String[] topics,`
`                                  String[] properties) {`
`return new KafkaMessageListener<>(groupId, topics, properties);`
when I create an instance of the KafkaMessageListener it starts up despite the fact that autoStartup is "false". I also tried to set 

ConcurrentKafkaListenerContainerFactory.setAutoStartup(false) but without success.

I found that KafkaListenerEndpointRegistry starts my listener in this method:
`private void startIfNecessary(MessageListenerContainer listenerContainer) {`
`		if (this.contextRefreshed || listenerContainer.isAutoStartup()) {`
`			listenerContainer.start();`
`		}`
`}`

because KafkaListenerEndpointRegistrar.startImmediately is true when I call appCtx.getBean(...) 

I expected that when I create an instance of KafkaMessageListener it must not be started.

garyrussell:
Please learn how to use GitHub markdown to format code properly.

`autoStartup` only applies during context initialization; it is not relevant when the context has been refreshed (fully initialized), including when creating new prototype bean instances after the initialization is complete.

We could consider adding an option to the registry.

tbarabanov:
Thx for the advice ). Is there any need in an option? I mean what if just get rid of the contextRefreshed check in KafkaListenerEndpointRegistry.startIfNecessary?

garyrussell:
I presume you are using `context.getBean()` to get new instance(s) of your listener; why not defer that until you actually want the container to start?

tbarabanov:
Yep, can do it this way. But it puts restrictions on how the code is organised, create/start/stop/ is just a bit more natural.

PR #2301, opened 2022-06-07
GH-2297: Add KLERegistry.alwaysStartAfterRefresh

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2297

Add an option to apply `autoStartup` semantics to late registrations.

### DIFF
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -1953,6 +1953,10 @@ A collection of managed containers can be obtained by calling the registry's `ge
 Version 2.2.5 added a convenience method `getAllListenerContainers()`, which returns a collection of all containers, including those managed by the registry and those declared as beans.
 The collection returned will include any prototype beans that have been initialized, but it will not initialize any lazy bean declarations.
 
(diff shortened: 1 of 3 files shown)


artembilan:

youngkangrui:
contextRereshed in KafkaListenerEndPointRegistry is set to true only if received ContextRefreshEvent, but  in Spring SmartLifeCycle  is triggered  before ContextFreshEvent is sended out , so when smartLIfeCycle wants to start the containers, contextRefreshed will always be false, so my question is adding  alwaysStartAfterRefresh this way will not start the containers even if context refreshed. 

using 2.8.7 ,follow code's output is still false,while expect is true
'''
public class TestListener {
    @KafkaListener(id="id1",topics = {"tp"},groupId = "consumerGroupId",autoStartup = "false")
    public void Listen(String info){
        System.out.println(info);
    }
}

ConfigurableApplicationContext context = SpringApplication.run(DemoApplication.class, args);
KafkaListenerEndpointRegistry registry = (KafkaListenerEndpointRegistry)context.getBean("org.springframework.kafka.config.internalKafkaListenerEndpointRegistry");
MessageListenerContainer id1 = registry.getListenerContainer("id1");
System.out.println(id1.isRunning());
'''

PR #2308, opened 2022-06-15
GH-666: Single global embedded Kafka with JUnit 5

artembilan:
Fixes https://github.com/spring-projects/spring-kafka/issues/666

* Introduce a `GlobalEmbeddedKafkaTestExecutionListener` which is registered
by the service loader from JUnit Platform
* Make its activation conditional based on the `spring.kafka.global.embedded.enabled`
system property
* Expose some other configuration properties for the target global `EmbeddedKafkaBroker`
* Verify its functionality via manual `Launcher.execute()`
* Add more `@DirtiesContext` to some tests in the `spring-kafka-test` which don't close
their embedded brokers on the exit


artembilan:
If this is OK, I'll go ahead with docs and perhaps some sample how to enable and configure this via Gradle which might be really a top interest from end-users.

(Perhaps the number of issue is so bad that it took for me a couple days to figure out how to test this 😢 😄 )

garyrussell (changes requested):
I don't think I can review this adequately because I am not familiar with the concepts. Perhaps we can ask @sbrannen to take a look?

artembilan:
> I am not familiar with the concepts

No problem, Gary! 
I'm in a chat with Sam, so we can see together from JUnit perspective.

From your side I just need 👍 about a general idea to have such a global embedded Kafka for the whole test plan as it is requested by users and as they demonstrate it via their custom Gradle plugins.

Re. Git commit: I think I had several original commits before pushing and probably when I wanted to push to PR as a single commit I accidentally included one of yours then rebased.
That's how your name could slip into the final commit message.
So, the probem not with an IDE, but someone's blind eyes between screen and keyboard. 😄

spring-kafka-test/src/main/java/org/springframework/kafka/test/junit/GlobalEmbeddedKafkaTestExecutionListener.java
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	public static final String COUNT_PROPERTY_NAME = "spring.kafka.embedded.count";
+
+	/**
+	 * The topics to create on the embedded broker(s).
+	 */
+	public static final String TOPICS_PROPERTY_NAME = "spring.kafka.embedded.topics";
+
+	/**
+	 * The topics to create on the embedded broker(s).
```
sbrannen, line 69: Copied-and-pasted from `TOPICS_PROPERTY_NAME`

spring-kafka-test/src/main/java/org/springframework/kafka/test/junit/GlobalEmbeddedKafkaTestExecutionListener.java
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	public static final String PORTS_PROPERTY_NAME = "spring.kafka.embedded.ports";
+
+	/**
+	 * The number of partitions on topics to create on the embedded broker(s).
+	 */
+	public static final String PARTITIONS_PROPERTY_NAME = "spring.kafka.embedded.partitions";
+
+	/**
+	 * The number of partitions on topics to create on the embedded broker(s).
```
sbrannen, line 79: Copied-and-pasted from `PARTITIONS_PROPERTY_NAME`

spring-kafka-test/src/test/resources/junit-platform.properties
```diff
@@ -0,0 +1,3 @@
+spring.kafka.embedded.count=2
+spring.embedded.kafka.brokers.property=spring.kafka.bootstrap-servers
```
sbrannen, line 2: Should this instead be `spring.kafka.embedded.broker.properties.location`?

Or is that a property not used by your `TestExecutionListener`?

spring-kafka-test/src/test/java/org/springframework/kafka/test/junit/GlobalEmbeddedKafkaTestExecutionListenerTests.java
```diff
@@ -0,0 +1,143 @@
...
+			assertThat(summary.getTestsSucceededCount()).isEqualTo(1);
+			assertThat(summary.getTestsFailedCount()).isEqualTo(1);
+		}
+		catch (Exception ex) {
+			summary.printFailuresTo(new PrintWriter(System.out));
+			throw ex;
+		}
+
+		System.clearProperty(GlobalEmbeddedKafkaTestExecutionListener.LISTENER_ENABLED_PROPERTY_NAME);
+		System.clearProperty(GlobalEmbeddedKafkaTestExecutionListener.BROKER_PROPERTIES_LOCATION_PROPERTY_NAME);
```
sbrannen, line 94: Typically, whenever I set system properties in a test and clear them at the end of the test, I wrap the body of the test in a try-finally block and clear the system properties in the `finally` block -- just to be on the safe side in case a particular test fails and the system properties adversely affect other tests in the test suite.

sbrannen (commented):
I think the `GlobalEmbeddedKafkaTestExecutionListener` implementation looks pretty good.

I only noticed a few very minor issues that I commented on.

artembilan:
Thank you, @sbrannen , for consultation and review!
I have addressed all your concerns, unless I decided to not play with test suites since I still want it to pass properly in the IDE as well without extra configuration on the end-user side.
Perhaps eventually, when we come up without more similar sophisticated unit testing, we will come back to a suite approach and respective Developer Guide.

From here I guess we don't need to bother you any more and @garyrussell simple can continue to review the rest of the changes.

samples/sample-05/mvnw
```diff
@@ -0,0 +1,316 @@
+#!/bin/sh
+# ----------------------------------------------------------------------------
```
garyrussell, line 2: This needs to be executable `chmod u+x` on Linux/MacOS; I can do it during merge.

samples/sample-05/pom.xml
```diff
@@ -0,0 +1,122 @@
...
+		<groupId>org.springframework.boot</groupId>
+		<artifactId>spring-boot-starter-parent</artifactId>
+		<version>3.0.0-SNAPSHOT</version>
+		<relativePath/> <!-- lookup parent from repository -->
+	</parent>
+	<groupId>com.example</groupId>
+	<artifactId>sample-05</artifactId>
+	<version>0.0.1-SNAPSHOT</version>
+	<name>sample-05</name>
+	<description>sample-05</description>
```
garyrussell, line 15:
```suggestion
	<description>kafka-sample-05</description>
```

For consistency.

samples/sample-05/src/main/resources/application.properties
```diff
@@ -0,0 +1,4 @@
+spring.kafka.bootstrap-servers=${spring.embedded.kafka.brokers}
+spring.kafka.producer.properties[max.block.ms]=1000
+
+logging.level.root=error
```
garyrussell, line 4: No newline at end.

samples/sample-05/src/test/java/com/example/Sample05Application1Tests.java
```diff
@@ -0,0 +1,52 @@
...
+@SpringBootTest
+@DirtiesContext
+class Sample05Application1Tests {
+
+	@Autowired
+	KafkaTemplate<String, String> kafkaTemplate;
+
+	@Test
+	void testKafkaListener(CapturedOutput output) {
+		this.kafkaTemplate.send("topic1", "testData");
```
garyrussell, line 47: Need to wait for future.

samples/sample-05/src/test/resources/kafka-broker.properties
```diff
@@ -0,0 +1 @@
+auto.create.topics.enable=false
```
garyrussell, line 1: No newline at end.

samples/sample-05/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,26 @@
...
+	<logger name="kafka.server.ReplicaFetcherThread" level="error"/>
+	<logger name="kafka.server.LogDirFailureChannel" level="error"/>
+	<logger name="kafka.server.BrokerMetadataCheckpoint" level="error"/>
+	<logger name="kafka.utils.CoreUtils$" level="error"/>
+	<logger name="org.apache.zookeeper.server.ZooKeeperServer" level="error"/>
+
+	<root level="error">
+		<appender-ref ref="STDOUT" />
+	</root>
+</configuration>
```
garyrussell, line 26: No newline at end.

spring-kafka-docs/src/main/asciidoc/testing.adoc
```diff
@@ -214,6 +215,24 @@ IMPORTANT: The preceding example provides no mechanism for shutting down the bro
 This could be a problem if, say, you run your tests in a Gradle daemon.
 You should not use this technique in such a situation, or you should use something to call `destroy()` on the `EmbeddedKafkaBroker` when your tests are complete.
 
+Starting with version 3.0, the framework exposes a `GlobalEmbeddedKafkaTestExecutionListener` for JUnit Jupiter Platform, which is disabled by default.
+The purpose of this listener is to start one global `EmbeddedKafkaBroker` for the whole test plan and stop it in the end of plan.
+To enable this listener, and therefore have a single global embedded Kafka cluster for all the tests in the project, a `spring.kafka.global.embedded.enabled` property must be set to `true` via system properties or JUnit platform configuration.
```
garyrussell, line 220:
```suggestion
Starting with version 3.0, the framework exposes a `GlobalEmbeddedKafkaTestExecutionListener` for the JUnit Jupiter Platform; it is disabled by default.
The purpose of this listener is to start one global `EmbeddedKafkaBroker` for the whole test plan and stop it at the end of the plan.
To enable this listener, and therefore have a single global embedded Kafka cluster for all the tests in the project, the `spring.kafka.global.embedded.enabled` property must be set to `true` via system properties or JUnit platform configuration.
```

spring-kafka-docs/src/main/asciidoc/testing.adoc
```diff
@@ -214,6 +215,24 @@ IMPORTANT: The preceding example provides no mechanism for shutting down the bro
...
+In addition, these properties can be provided:
+
+- `spring.kafka.embedded.count` - the number of Kafka brokers to manage;
+- `spring.kafka.embedded.ports` - ports (comma-separated value) for every Kafka broker to start, `0` if random port is a preference; number of values must be equal to the `count` mentioned above;
```
garyrussell, line 224:
```sugestion
- `spring.kafka.embedded.ports` - ports (comma-separated value) for every Kafka broker to start, `0` if random port is a preferred; the number of values must be equal to the `count` mentioned above;
```

spring-kafka-docs/src/main/asciidoc/testing.adoc
```diff
@@ -214,6 +215,24 @@ IMPORTANT: The preceding example provides no mechanism for shutting down the bro
...
+- `spring.kafka.embedded.topics` - topics (comma-separated value) to create in the started Kafka cluster;
+- `spring.kafka.embedded.partitions` - number of partitions to provision to the topics to create;
```
garyrussell, line 226:
```suggestion
- `spring.kafka.embedded.partitions` - number of partitions to provision for the created topics;
```

spring-kafka-docs/src/main/asciidoc/testing.adoc
```diff
@@ -214,6 +215,24 @@ IMPORTANT: The preceding example provides no mechanism for shutting down the bro
...
+- `spring.kafka.embedded.broker.properties.location` - the location to the file for addition Kafka broker configuration properties; must follow Spring resource abstraction pattern.
```
garyrussell, line 227:
```suggestion
- `spring.kafka.embedded.broker.properties.location` - the location of the file for additional Kafka broker configuration properties; the value of this property must follow the Spring resource abstraction pattern.
```

spring-kafka-test/src/test/resources/junit-platform.properties
garyrussell: This is used by the broker itself (it tells the broker to update that system property - useful in Boot tests) - @artembilan I think we should add docs about this, especially for Boot users.

garyrussell (changes requested):
The sample-05 tests fail for me

```
[ERROR] Errors: 
[ERROR]   Sample05Application1Tests.testKafkaListener:49 » ConditionTimeout Condition with com.example.Sample05Application1Tests was not fulfilled within 10 seconds.
[ERROR]   Sample05Application2Tests.testKafkaTemplateSend:53 » Kafka Send failed
[INFO] 
[ERROR] Tests run: 2, Failures: 0, Errors: 2, Skipped: 0
```

More info for the second one (because it waits for the future)...

```
Caused by: org.apache.kafka.common.errors.TimeoutException: Topic nonExistingTopic not present in metadata after 1000 ms.
```

artembilan:
> The sample-05 tests fail for me

Sorry, I forgot to mention that you need to `./gradlew publishToMavenLocal` the core project first: the sample is based on the content of this branch.

garyrussell:
I already did that - it won't compile otherwise.

artembilan:
> This is used by the broker itself (it tells the broker to update that system property - useful in Boot tests) 

Sorry, it is not clear what is your concern over here. This indeed was my intention to demonstrate how that property is populated by the embedded property and can be used by Spring Boot auto-configuration.

> I think we should add docs about this, especially for Boot users.

We do have some mentioning about that setting in our docs: https://docs.spring.io/spring-kafka/docs/current/reference/html/#testing.
And we also have something in Spring Boot by itself: https://docs.spring.io/spring-boot/docs/current/reference/html/messaging.html#messaging.kafka.embedded
Nevertheless I will look shortly what else we can add into docs and explain.

Thanks

spring-kafka-docs/src/main/asciidoc/testing.adoc
sbrannen: It's actually for the "JUnit Platform" (a.k.a., JUnit 5).

It's not specific to JUnit Jupiter.

garyrussell:
https://github.com/spring-projects/spring-kafka/blob/596f7d940caa2f562775d3f232be4ce7b58bffe9/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java#L646-L658

I'm getting that `Send failed` exception with ... 

```
Caused by: org.apache.kafka.common.errors.TimeoutException: Topic nonExistingTopic not present in metadata after 1000 ms.
```

artembilan:
> I'm getting that Send failed exception with ...

That's correct. See JavaDocs of that `Sample05Application2Tests`:
```
/**
 * This test is going to fail from IDE since there is no exposed {@code spring.embedded.kafka.brokers} system property.
 * This test is deliberately failing to demonstrate that global embedded Kafka broker config for
 * {@code auto.create.topics.enable=false} is in an effect.
 * See {@code /resources/kafka-broker.properties} and Maven Surefire plugin configuration.
 */
```

artembilan:
> but it fails under maven as well; 

That's correct as well.
I say there in the JavaDocs ` This test is deliberately failing to demonstrate that global embedded Kafka broker config for {@code auto.create.topics.enable=false} is in an effect.`.

The sample is about showing a global embedded Kafka, not how to right tests.
The failing test and an output that global embedded Kafka is stopped in the end of plan is a good sign for us to be sure that we are on a right track.
The failing report is still a report.

I also say that explicitly in the `README.adoc`:

> One of the tests is deliberately failing to demonstrate that global embedded Kafka broker config for `auto.create.topics.enable=false` is in an effect.

I'm not sure why you still insist that it has to pass...

samples/sample-05/src/test/java/com/example/Sample05Application2Tests.java
```diff
@@ -0,0 +1,58 @@
...
+	@Autowired
+	KafkaTemplate<String, String> kafkaTemplate;
+
+	@Test
+	void testKafkaTemplateSend() throws ExecutionException, InterruptedException, TimeoutException {
+		SendResult<String, String> sendResult =
+				this.kafkaTemplate.send("nonExistingTopic", "fake data").get(10, TimeUnit.SECONDS);
+
+		assertThat(sendResult).isNotNull();
+	}
```
garyrussell, line 56: Because it's ugly seeing all those errors on the command line - I think this is better:

```suggestion
	@Test
	void testKafkaTemplateSend() {
		Throwable thrown = catchThrowable(() ->
						this.kafkaTemplate.send("nonExistingTopic", "fake data").get(10, TimeUnit.SECONDS));

		assertThatExceptionOfType(TimeoutException.class).isThrownBy(() -> {
					throw thrown.getCause();
				})
				.withMessageContaining("Topic nonExistingTopic")
				.withMessageContaining("metadata");
	}
```

garyrussell (changes requested):

spring-kafka-test/src/test/resources/junit-platform.properties
artembilan: Oh! Sorry, you just answered to Sam's question :shame: 

But yeah, I have added more in docs anyway 😄

garyrussell (changes requested):

samples/sample-05/src/test/java/com/example/Sample05Application2Tests.java
artembilan: 😸 I still think my idea to demonstrate a robustness of our solution even if some tests are failing in between is the way to go for this sample.
On the other hand we just can remove this sample altogether: for me clean finish for the plan is OK even if the report is about failure.
The successful tests was really not a goal for me in this sample, but rather how global embedded Kafka work.
See also `GlobalEmbeddedKafkaTestExecutionListenerTests`: one if its tests is also deliberately failing.

PR #2310, opened 2022-06-19
GH-2288: Add option to deeply traverse exc chain

### LINKED ISSUE #2288: CommonDelegatingErrorHandler: How to traverse exceptions in order to find a correct delegate
Add an option to traverse the cause chain to find a delegate.

### Discussed in https://github.com/spring-projects/spring-kafka/discussions/2283

<div type='discussions-op-text'>

<sup>Originally posted by **amseager** May 25, 2022</sup>
Currently, everything in this handler uses this private method:
```
@Nullable
	private CommonErrorHandler findDelegate(Throwable thrownException) {
		Throwable cause = thrownException;
		if (cause instanceof ListenerExecutionFailedException) {
			cause = thrownException.getCause();
		}
		if (cause != null) {
			Class<? extends Throwable> causeClass = cause.getClass();
			for (Entry<Class<? extends Throwable>, CommonErrorHandler> entry : this.delegates.entrySet()) {
				if (entry.getKey().isAssignableFrom(causeClass)) {
					return entry.getValue();
				}
			}
		}
		return null;
	}
``` 
As you can see, if we have `ListenerExecutionFailedException`, we'll get its cause and find an appropriate delegate.

However, this wouldn't work in a lot of cases when we have our set of exceptions wrapped in something else, e.g. in `MessageConversionException` (I think it's the most popular case). And we cannot use `MessageConversionException` as a `delegates` map key since we want to have a different logic for each underlying exception (let's say we want custom `DefaultErrorHandler` for `ConstraintViolationException` with a fixed backoff and its own recoverer and the another one for any other exception with `ExponentialBackOff` without a recoverer at all - but both are wrapped in the same `MessageConversionException`).

How to solve this problem? I don't want to write a custom DelegatingErrorHandler because I'll need to copy almost all the code from the `CommonDelegatingErrorHandler`.
Previously, we had `ConditionalDelegatingErrorHandler` which had `handle` method being public so it could've been ok to override it but now it's deprecated in favor of `CommonDelegatingErrorHandler`.</div>

breader124:
Fixes #2288

Add possibility to deeply traverse cause chain in order to find proper delegate
for handling thrown exception. Keep old way of cause chain traversing as default
one.

Cover new code with unit test.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonDelegatingErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonDelegatingErrorHandler.java
@@ -16,15 +16,18 @@
 
 package org.springframework.kafka.listener;
 
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.stream.Collectors;
 
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 
+import org.springframework.classify.BinaryExceptionClassifier;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
@@ -44,7 +47,8 @@ public class CommonDelegatingErrorHandler implements CommonErrorHandler {
 
 	private final Map<Class<? extends Throwable>, CommonErrorHandler> delegates = new LinkedHashMap<>();
 
-	private boolean deepCauseChainTraversing = false;
+	private boolean causeChainTraversing = false;
+	private NoTraversingBinaryExceptionClassifier classifier = new NoTraversingBinaryExceptionClassifier(new HashMap<>());
 
 	/**
 	 * Construct an instance with a default error handler that will be invoked if the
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/CommonDelegatingErrorHandler.java
```diff
@@ -173,4 +182,29 @@ private CommonErrorHandler findDelegate(Throwable thrownException) {
...
+		return cause;
+	}
+
+	@Nullable
+	private Throwable deepTraverseCauseChain(Throwable thrownException) {
+		while (thrownException != null && thrownException.getCause() != null) {
+			thrownException = thrownException.getCause();
+		}
+		return thrownException;
+	}
```
amseager, line 208: Hello,
Looks like only the "root" exception will be checked for matching after that. In my initial request, I'd been thinking about more complex solution - to check every exception from the chain. This approach is used e.g. in `spring-retry` library (see `RetryTemplateBuilder.traversingCauses()`).
Will it be possible to improve the current logic with my suggestion?

breader124: Hi, I just pushed a commit with improvements, tried to follow idea you mentioned. Looking forward to hear whether solution satisfies your needs now

amseager: Actually, don't know much about all of these classifiers, but for me it's ok if it works :) 
Also, you'd probably need to actualize the `setCauseChainTraversing` javadoc

breader124:
Hi @garyrussell, @artembilan. Do you think this PR can be merged after last amendments?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/CommonDelegatingErrorHandler.java
```diff
@@ -59,11 +67,28 @@ public CommonDelegatingErrorHandler(CommonErrorHandler defaultErrorHandler) {
...
+		Map<Class<? extends Throwable>, Boolean> classifications = delegates.keySet().stream()
+			.map(commonErrorHandler -> Map.entry(commonErrorHandler, true))
+			.collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+		this.classifier = new BinaryExceptionClassifier(classifications);
 	}
 
+	/**
+	 * Set the flag enabling deep exception's cause chain traversing. If true, delegate
+	 * for first exception classified by {@link BinaryExceptionClassifier} will be retrieved.
+	 * @param causeChainTraversing the causeChainTraversing flag.
```
garyrussell, line 87: Needs `@since 2.8.8` will add during merge.

garyrussell (approved):

garyrussell:
Thanks for the contribution - looking forward to more!

Merged as 6a24685bf366bd966b9efa52c9d60cc8f111d9ac with minor Javadoc polishing.

PR #2312, opened 2022-06-21
Use KafkaListenerEndpoint instead of MethodKafkaListenerEndpoint in the RetryTopicNamesProvider

### LINKED ISSUE #2294: RetryTopicNamesProvider depends on an impl class instead of KafkaListenerEndpoint

2.8.6, but I think it has been like that since the creation of the `RetryTopicNamesProvider`

**Describe the bug**

We are using Spring Kafka to programatically register custom `KafkaListenerEndpoint`(s), using a custom `MessageListener`.
We are in the process of providing a retry mechanism, backed by Spring Kafka. However, when trying that out we realized that not all interfaces can be used. e.g. the `RetryTopicNamesProvider` depends on the `MethodKafkaListenerEndpoint`, when it can depend on the `KafkaListenerEndpoint`.

**Expected behavior**

I would expect to be able to use the `RetryTopicNamesProvider` with any `KafkaListenerEndpoint`

-----

The issue is raised after a question in the [gitter room](https://gitter.im/spring-projects/spring-kafka?at=6299cd0fd126691416cb47ed)

filiphr:
Fixes #2294

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicNamesProviderFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicNamesProviderFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 the original author or authors.
+ * Copyright 2021-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


artembilan (approved):
This is breaking change just on the byte code level.
From the API you won't see a difference calling this methods.

However this is just an improvement, so it goes only to `2.9.x` and `3.0`. not back-porting to `2.8.x` - not critical.

I'll defer merging to @garyrussell .

Thank you all for the contribution into this issue!

garyrussell:
Cherry-picked to 2.9.x as 78fa932f

@filiphr In future, please limit commit headlines to 50 chars (see contribution guidelines).

filiphr:
>@filiphr In future, please limit commit headlines to 50 chars (see contribution guidelines).

Sorry for that. Next time I'll limit the commit to 50 chars. I thought you might also do a squash and merge and adapt the headline to what you think makes most sense to you 😄

PR #2313, opened 2022-06-21
Setting up the backoff aware message listener adapter should work for any message listener

filiphr:
We had a problem here while trying to implement our programatic retry mechanism using Kafka.

This is one of the places that we had to adapt our `MessageListener`.

I am not sure whether the cast you require is mandatory or not, but the `KafkaBackoffAwareMessageListenerAdapter` does support different message listeners.


tomazfernandes:
Yep, I think you're right. It's the delegate that might not have access to the consumer if it's a different interface. So, no concerns from me regarding this. Thanks!

artembilan (approved):
We do this over there even right now:
```
new KafkaBackoffAwareMessageListenerAdapter<>(listener,
```
Which leads to this:
```
public KafkaBackoffAwareMessageListenerAdapter(MessageListener<K, V> adapter,
```

So, that is really strange to have it forced just to the `AcknowledgingConsumerAwareMessageListener`...

LGTM, though!

garyrussell:
Merged as 596f7d940caa2f562775d3f232be4ce7b58bffe9 and cherry-picked to 2.9.x, 2.8.x.

PR #2317, opened 2022-06-23
GH-2284: Publish Auth Error Events

### LINKED ISSUE #2284: Publish Event In Container when Authorization/Authentication Exception occurs and retry takes place
Hi @garyrussell  and @artembilan 

I am using spring kafka and i need to do something when kafka connection error happens due to authorization/authentication exception been thrown in the following class - https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java

No , i do not want the container to stop , hence i have the authExceptionRetryInterval set. But what i would also like is an application event to be published when this kind of error occurs so i can listen on it and then do something custom. 

Also  , i would also like an event to be published when the connection is successfully restored in the pollAndInvoke() method in the same class. Please let me know your thoughts on the same and if you guys are open to such a change.

danielgentes:
* Added Events for auth failed
and successful retry of auth
* publish of these events
* unit testing

Closes #2284

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerRetryAuthEvent.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerRetryAuthEvent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2022 the original author or authors.
+ * Copyright 2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,8 @@
 package org.springframework.kafka.event;
 
 /**
- * An event published when authentication or authorization of a consumer fails and is being retried. Contains the triggering throwable.
+ * An event published when authentication or authorization of a consumer fails and
+ * is being retried. Contains the triggering throwable.
  *
  * @author Daniel Gentes
  * @since 3.0
@@ -51,7 +52,8 @@ public class ConsumerRetryAuthEvent extends KafkaEvent {
 	/**
 	 * Construct an instance with the provided source and container.
 	 * @param source the container instance that generated the event.
-	 * @param container the container or the parent container if the container is a child.
+	 * @param container the container or the parent container
+	 *                     if the container is a child.
 	 * @param reason the reason.
 	 * @since 3.0
 	 */
@@ -61,9 +63,9 @@ public class ConsumerRetryAuthEvent extends KafkaEvent {
 	}
 
 	/**
-	 * Return the reason why the consumer was stopped.
+	 * Return the reason for the auth failure.
 	 * @return the reason.
-	 * @since 2.5.8
+	 * @since 2.8.8
 	 */
 	public Reason getReason() {
 		return this.reason;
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerRetryAuthEvent.java
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright 2018-2022 the original author or authors.
```
garyrussell, line 2: New files only need 2022.

spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerRetryAuthEvent.java
```diff
@@ -0,0 +1,77 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.event;
+
+/**
+ * An event published when authentication or authorization of a consumer fails and is being retried. Contains the triggering throwable.
```
garyrussell, line 20: Please wrap javadocs at 90 chars (code at 120).

spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerRetryAuthEvent.java
```diff
@@ -0,0 +1,77 @@
...
+	 * @param reason the reason.
+	 * @since 3.0
+	 */
+	public ConsumerRetryAuthEvent(Object source, Object container, Reason reason) {
+		super(source, container);
+		this.reason = reason;
+	}
+
+	/**
+	 * Return the reason why the consumer was stopped.
```
garyrussell, line 64: Looks like copy/paste.

Return the reason for the auth failure.

spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerRetryAuthEvent.java
```diff
@@ -0,0 +1,77 @@
...
+	 * @return the reason.
+	 * @since 2.5.8
```
garyrussell, line 66:
```suggestion
	 * @since 2.8.8
```

spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerRetryAuthSuccessfulEvent.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright 2018-2022 the original author or authors.
```
garyrussell, line 2: 2022

garyrussell (changes requested):
Thanks!

Just a few minor nits.

danielgentes:
Hi @garyrussell,
thank you for the review! I addressed the mentioned changed in my latest commit.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -529,6 +531,30 @@ private void publishConsumerFailedToStart() {
 		}
 	}
 
+	private void publishRetryAuthEvent(Throwable throwable) {
```
artembilan, line 534: Please, add your name to the `@author` list: you have made some changes to the class, so you deserve some credit.

danielgentes:
Hi @artembilan ,
thanks for the review! I updated the documentation and also made the other changes you suggested.

Issue #2318, opened 2022-06-23
Add A Mechanism To Wait For Manual Assignment to ReplyingKafkaTemplate


garyrussell:
See https://stackoverflow.com/questions/72730978

PR #2319, opened 2022-06-23
GH-2318: ReplyingKT - Wait for Assignment

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2318

Add an option to wait for assignment to the `ReplyingKafkaTemplate`; useful
when using `auto.offset.reset=latest` to avoid sending a request with the reply
sent before the container is initialized.

### DIFF
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -555,6 +555,11 @@ You can use this future to determine the result of the send operation.
 
 If the first method is used, or the `replyTimeout` argument is `null`, the template's `defaultReplyTimeout` property is used (5 seconds by default).
 
+Starting with version 2.8.8, the template has a new method `waitForAssignment`.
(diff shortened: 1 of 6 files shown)


artembilan (commented):
How all of this is going to work when rebalance happens and new partitions are assigned to that replying contrainer, and we interact with this template at the moment?

Thanks

garyrussell:
With multiple application instances, you would generally not use group management, and manually assign partitions. Or, as mentioned in the documentation, each instance must have a different `group.id`. In either case, there will be no rebalances.

>When you configure with a single reply `TopicPartitionOffset`, you can use the same reply topic for multiple templates, as long as each instance listens on a different partition. When configuring with a single reply topic, each instance must use a different group.id. In this case, all instances receive each reply, but only the instance that sent the request finds the correlation ID. This may be useful for auto-scaling, but with the overhead of additional network traffic and the small cost of discarding each unwanted reply. When you use this setting, we recommend that you set the template’s `sharedReplyTopic` to true, which reduces the logging level of unexpected replies to `DEBUG` instead of the default `ERROR`.

PR #2327, opened 2022-07-05
GH-2195: Polishing

garyrussell:
Related to https://github.com/spring-projects/spring-kafka/issues/2195

- check `isPartitionPauseRequested()` instead of `isPartitionPaused()`
- add diagnostic debug logging
- prevent re-pausing the entire consumer each time a poll exits but the first
  partition is still paused.

**cherry-pick to 2.9.x**


artembilan:

Issue #2239, opened 2022-04-20
Revise RetryTopicConfigurationSupport TaskExecutor Strategy


tomazfernandes:
Currently we add a `ThreadPoolTaskExecutor` `@Bean` in `RetryTopicConfigurationSupport`. That might have undesirable side-effects such as making `Spring Boot`'s auto configuration back off from providing a default `TaskExecutor` for users.

We should investigate this further and if necessary look for alternatives.

tomazfernandes:
I started looking into this, and TBH I think it's a slightly worse scenario than what I'd have thought.

Turns out when there is a `TaskExecutor` bean in the context, Boot's auto configured task executor does back off, as we supposed it would. That might already be bad enough, since auto configuration enables many task executor configurations that users may be relying on.

But what I think is the worst part is that there's logic in the `WebMvcAutoConfiguration#237` class that tries to fetch the auto configured task executor by name - since it's not there when we provide our own, it falls back to a `SimpleAsyncTaskExecutor` for the `RequestMappingHandlerAdapter`, which should be quite less performant than the default `ThreadPoolTaskExecutor` since it doesn't reuse threads.

I think this is really unexpected and weird behavior, but at least in my test application with Boot 2.6.6 that's how it goes.

I'm not sure what'd be a good move here - the less impactful thing I can think of would be if we'd go back to letting the `WakingTimingAdjuster` manage the `TaskExecutor` lifecycle, and maybe even having it instantiate it lazily on the first timing adjustment request.

Any thoughts?

Thanks.

artembilan:
You know I think the `SimpleAsyncTaskExecutor` would be a good compromise for a default settings.
I see we use it in many-many place in all the Spring projects.
Using a `SimpleAsyncTaskExecutor` we won't need to worry about its lifecycle.
The end-user would have a choice for other `TaskExecutor` impls to inject otherwise.

tomazfernandes:
Oh, that sounds great really. TBH, I have no clue of how frequently this timing adjuster kicks in in real life scenarios - it depends on too many factors. But as an overall default dependency for the feature, a full fledged `ThreadPoolTaskExecutor` indeed seems like too much.

I'll do some testing to see if it has any impact on actual precision and submit a PR with this change.

Please let me know if you think of anything else.

Thanks!

garyrussell:
Actually, a SATE is probably not good here; especially with long delays; we could end up with many threads sleeping.

[...]

PR #2328, opened 2022-07-05
GH-2239: RetryableTopic Refactoring

garyrussell:
Use a Task Scheduler to resume partitions, avoiding the need for timing adjustment.

**cherry-pick to 2.9.x**

I will make a What's New commit after the review/merge because it will certainly conflict.

I have left the commits separate to avoid you having to re-review what you have already looked at (if you so desire).

cc/ @tomazfernandes


spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerPausingBackOffHandler.java
```diff
@@ -51,4 +53,9 @@ public void onNextBackOff(@Nullable MessageListenerContainer container, Exceptio
 		}
 	}
 
+	@Override
+	public void onNextBackOff(MessageListenerContainer container, TopicPartition partition, long nextBackOff) {
```
tomazfernandes, line 57: Given the interface states `container` is `@Nullable`, maybe a null check is in order here?

tomazfernandes (approved):
Just a couple nits, LGTM as is, thanks!

tomazfernandes:
Thanks @garyrussell, this looks great!

There are a few other deprecations that I'm not sure if you missed or just decided not to deal with in this PR, such as:

- This part of the [RetryableTopicAnnotationProcessor](https://github.com/spring-projects/spring-kafka/blob/36665e9d16aec21164688eefdfffc971a447d5cc/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java#L216-L222) class can be removed, and with that we have no further use for `RetryTopicInternalBeanNames` and `RetryTopicInternalBeanNamesTests`.

- We can also remove the `configure` methods from `ListenerContainerFactoryConfigurer`, and kill some other logic there, such as `isSetContainerProperties` and related logic.

- Also, a couple of lines with `String` bean names in `ListenerContainerFactoryResolver`, and a couple `@SupressWarnings` in `RetryTopicConfigurer`.

Of course, these shouldn't change much overall, and can be part of a different `PR` or left for 3.0, so no worries if you prefer keeping them.

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
```diff
@@ -214,25 +214,17 @@ private EndpointHandlerMethod getDltProcessor(Method listenerMethod, Object bean
...
-					org.springframework.kafka.retrytopic.RetryTopicInternalBeanNames.DEFAULT_KAFKA_TEMPLATE_BEAN_NAME,
+			return this.beanFactory.getBean(RetryTopicBeanNames.DEFAULT_KAFKA_TEMPLATE_BEAN_NAME,
 					KafkaOperations.class);
 		}
-		catch (NoSuchBeanDefinitionException ex) {
+		catch (NoSuchBeanDefinitionException ex2) {
 			try {
-				return this.beanFactory.getBean(RetryTopicBeanNames.DEFAULT_KAFKA_TEMPLATE_BEAN_NAME,
-						KafkaOperations.class);
+				return this.beanFactory.getBean(DEFAULT_SPRING_BOOT_KAFKA_TEMPLATE_NAME, KafkaOperations.class);
```
artembilan, line 222: This is semantic cycle with Spring Boot.
Why just plain `getBean(KafkaOperations.class)` is not enough?

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerContainerRegistry.java
```diff
@@ -41,6 +41,17 @@ public interface ListenerContainerRegistry {
...
+	/**
+	 * Return the {@link MessageListenerContainer} with the specified id or {@code null}
+	 * if no such container exists. Returns containers that are not registered with the
+	 * registry, but exist in the application context.
+	 * @param id the id of the container
+	 * @return the container or {@code null} if no container with that id exists
+	 * @see #getListenerContainerIds()
+	 */
+	@Nullable
+	MessageListenerContainer getUnregisteredListenerContainer(String id);
```
artembilan, line 53: Why do we need this API, please?

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerPartitionPausingBackOffManager.java
```diff
@@ -0,0 +1,91 @@
...
+	/**
+	 * Construct an instance with the provided registry and back off handler.
+	 * @param listenerContainerRegistry the registry.
+	 * @param backOffHandler the handler.
+	 */
+	public ContainerPartitionPausingBackOffManager(ListenerContainerRegistry listenerContainerRegistry,
+			BackOffHandler backOffHandler) {
+
+		this.listenerContainerRegistry = listenerContainerRegistry;
+		this.backOffHandler = backOffHandler;
```
artembilan, line 55: `Assert.notNull()`? For both?
I believe `@NonNullApi` makes it working only for Kotlin and perhaps in the IDEA.
In other Java cases we simply can pass the `null`...

I won't insist, but let's think about our goal and consistency here and other similar places!

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerPartitionPausingBackOffManager.java
```diff
@@ -0,0 +1,91 @@
...
+		LOGGER.debug(() -> "Back off time: " + backoffTime + " Context: " + context);
+		if (backoffTime > 0) {
+			pauseConsumptionAndThrow(context, backoffTime);
+		}
+	}
+
+	private void pauseConsumptionAndThrow(Context context, Long backOffTime) throws KafkaBackoffException {
+		TopicPartition topicPartition = context.getTopicPartition();
+		getListenerContainerFromContext(context).pausePartition(topicPartition);
+		this.backOffHandler.onNextBackOff(getListenerContainerFromContext(context), topicPartition, backOffTime);
```
artembilan, line 75: I guess this `getListenerContainerFromContext(context)` can be extracted to the variable.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerPartitionPausingBackOffManager.java
```diff
@@ -0,0 +1,91 @@
...
+		throw new KafkaBackoffException(String.format("Partition %s from topic %s is not ready for consumption, " +
+				"backing off for approx. %s millis.", context.getTopicPartition().partition(),
+				context.getTopicPartition().topic(), backOffTime),
+				topicPartition, context.getListenerId(), context.getDueTimestamp());
+	}
+
+	private MessageListenerContainer getListenerContainerFromContext(Context context) {
+		MessageListenerContainer container = this.listenerContainerRegistry.getListenerContainer(context.getListenerId()); // NOSONAR
+		if (container == null) {
+			container = this.listenerContainerRegistry.getUnregisteredListenerContainer(context.getListenerId());
```
artembilan, line 85: Why this cannot be hidden in the `ListenerContainerRegistry` implementation details?

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerPartitionPausingBackOffManager.java
```diff
@@ -0,0 +1,91 @@
...
+			pauseConsumptionAndThrow(context, backoffTime);
+		}
+	}
+
+	private void pauseConsumptionAndThrow(Context context, Long backOffTime) throws KafkaBackoffException {
+		TopicPartition topicPartition = context.getTopicPartition();
+		getListenerContainerFromContext(context).pausePartition(topicPartition);
+		this.backOffHandler.onNextBackOff(getListenerContainerFromContext(context), topicPartition, backOffTime);
+		throw new KafkaBackoffException(String.format("Partition %s from topic %s is not ready for consumption, " +
+				"backing off for approx. %s millis.", context.getTopicPartition().partition(),
```
artembilan, line 77: Looks like you have already extracted a `topicPartition` variable...

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerPartitionPausingBackOffManagerFactory.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * A factory for {@link ContainerPartitionPausingBackoffManager}.
+ *
+ * @author Gary Russell
+ * @since 2.9
+ *
+ */
+public class ContainerPartitionPausingBackOffManagerFactory extends AbstractKafkaBackOffManagerFactory {
```
artembilan, line 28: I won't insist, but all of these looks like a `FactoryBean` abstraction if such an intermediary is needed at all...

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerContainerPauseService.java
```diff
@@ -81,13 +82,37 @@ public void pause(MessageListenerContainer messageListenerContainer, Duration pa
...
+			Duration pauseDuration) {
+
+		Instant resumeAt = Instant.now().plusMillis(pauseDuration.toMillis());
+		LOGGER.debug(() -> "Pausing container: " + messageListenerContainer + " partition: " + partition
+				+ ", resume scheduled for "
+				+ resumeAt.atZone(ZoneId.systemDefault()).toLocalDateTime());
+		messageListenerContainer.pausePartition(partition);
+		this.scheduler.schedule(() -> {
+			LOGGER.debug(() -> "Resuming container: " + messageListenerContainer + " partition: " + partition);
+			messageListenerContainer.resumePartition(partition);
```
artembilan, line 111: In Spring Integration I had a bug in the past that the framework kept the memory consumption having references to heavy objects from scheduled tasks like this.
WDYT if we should avoid that over here as well in favor of simple listener container id?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationSupport.java
```diff
@@ -64,13 +63,20 @@
...
 
+	private static final String BACK_OFF_MANAGER_THREAD_EXECUTOR_BEAN_NAME = "backOffManagerThreadExecutor";
+
+	private static final AtomicBoolean ONLY_ONE_ALLOWED = new AtomicBoolean(true);
+
 	private final RetryTopicComponentFactory componentFactory = createComponentFactory();
 
-	private static final String BACK_OFF_MANAGER_THREAD_EXECUTOR_BEAN_NAME = "backOffManagerThreadExecutor";
+	protected RetryTopicConfigurationSupport() {
+		Assert.state(ONLY_ONE_ALLOWED.getAndSet(false), "Only one 'RetryTopicConfigurationSupport' is allowed");
```
artembilan, line 78: I wonder if there still exist multi-application solutions in a single JVM out there.
Like Tomcat, JBOSS AS etc..
Kinda this `static` flag would be a show stopper, if several spring-kafka application are ran in the same JVM. 😸 .

Not sure though that I have some other solution. Perhaps check for some bean in the application context...

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicSchedulerWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+package org.springframework.kafka.retrytopic;
+
+import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.scheduling.TaskScheduler;
+
+/**
+ * A wrapper class for a {@link TaskScheduler} to use for scheduling container resumption
+ * when a partition has been paused for a retry topic. Using this class prevents breaking
+ * Spring Boot's auto configuration for other frameworks. Use this if you are using Spring
```
artembilan, line 26: This probably has to say that provided `TaskScheduler` must not be a bean.
Otherwise It is not clear why do we do the manual lifecycle control and how this would let us to avoid Spring Boot auto-configuration breakage.

artembilan (changes requested):
I left some review, but I guess all of that is optional.
And may be revised in other issues/PR.

Thank you for a great overhaul! 😄

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerContainerRegistry.java
garyrussell: The old back off manager only handled containers created via `@KafkaListener`. The new one `ContainerPartitionPausingBackOffManager` will call this when it doesn't find a registered one. This seemed a reasonable place to do it (we already have `getAllListenerContainers()`, rather than restructuring the code there to handle the application context directly.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerPartitionPausingBackOffManagerFactory.java
garyrussell: Maybe we can look at that in 3.0; I'd like to get RC1 out tomorrow.

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerContainerPauseService.java
garyrussell: It's only a reference; the container is in memory anyway (in the container registry or application context). I think in SI, it was the persisted MG that was kept in memory unnecessarily.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationSupport.java
garyrussell: I believe they would be in different class loaders there. But we can re-think this after the RC.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -508,8 +510,16 @@ private boolean processMainAndRetryListeners(KafkaListener kafkaListener, Object
 	}
 
 	private RetryTopicConfigurer getRetryTopicConfigurer() {
-		return this.beanFactory.getBean(RetryTopicBeanNames.RETRY_TOPIC_CONFIGURER_BEAN_NAME,
-				RetryTopicConfigurer.class);
+		try {
```
tomazfernandes, line 513: Isn't this missing something like `if (this.topicConfigurer == null)`?

artembilan:
Cherry-picked to `2.9.x` as https://github.com/spring-projects/spring-kafka/commit/b88cbf900905d96082564604b546c9322214cb2d after some minor conflict resolutions.

Issue #2329, opened 2022-07-06
Fix interceptBeforeTx


garyrussell:
Currently, this flag only works (when set to false) with a `KafkaAwareTransactionManager`; it should apply to any `PlatformTransactionManager`.

PR #2330, opened 2022-07-06
GH-2329: Fix interceptBeforeTx for Non-Kafka TM

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2329

The flag should apply, regardless of PTM type.

**cherry-pick to 2.9.x,**


artembilan (changes requested):
LGTM, but I wonder if it is legit to back-port it down to `2.8.x`.
Isn't it possible that someone relies on the current behavior?
I'm OK cherry-picking this to `2.9.x`, but I have some backward compatibility concern for `2.8.x`.

Thanks

garyrussell:
OK; I am fine with that given that there is a work around (use an advice instead). I will push doc changes (and what's new) after merge.

artembilan:

PR #2331, opened 2022-07-06
GH-2195: Fix Fix No Seek Retries with Pause

garyrussell:
Related to https://github.com/spring-projects/spring-kafka/issues/2195

Similar to the recent fix when the first partition is paused, we should
not continue to process the remaining records if the container itself
was paused during the poll after an error occurred with `seekAfterError=false`.

**cherry-pick to 2.9.x**


garyrussell:
Do not merge; found another bug.

artembilan:

PR #2333, opened 2022-07-06
GH-2332: Fix Container.pause() with Manual Assign.

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2332

Use manual assignments, if present for pause.
Also, in the rebalance listener, use `isPaused()` rather than `consumerPaused`
to determine whether the partitions should be paused.

Add logs and events for pauses in the rebal listener.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -1691,11 +1691,14 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			if (!this.consumerPaused && (isPaused() || this.pausedForAsyncAcks)
 					|| this.pauseForPending) {
 
-				this.consumer.pause(this.consumer.assignment());
-				this.consumerPaused = true;
-				this.pauseForPending = false;
-				this.logger.debug(() -> "Paused consumption from: " + this.consumer.paused());
-				publishConsumerPausedEvent(this.consumer.assignment());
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1691,11 +1691,14 @@ private void doPauseConsumerIfNecessary() {
 			if (!this.consumerPaused && (isPaused() || this.pausedForAsyncAcks)
 					|| this.pauseForPending) {
 
-				this.consumer.pause(this.consumer.assignment());
-				this.consumerPaused = true;
-				this.pauseForPending = false;
-				this.logger.debug(() -> "Paused consumption from: " + this.consumer.paused());
-				publishConsumerPausedEvent(this.consumer.assignment());
+				Collection<TopicPartition> assigned = getAssignedPartitions();
+				if (!CollectionUtils.isEmpty(assigned)) {
```
artembilan, line 1695: I guess this one has to come together with an `import org.springframework.util.CollectionUtils;`

artembilan (changes requested):

artembilan:
Cherry-picked to `2.9.x`, but it doesn't back-port clear to `2.8.x`.
No `pauseForPending` in the `KafkaMessageListenerContainer` and no `DefaultErrorHandler.setSeekAfterError()` used in the `PauseContainerManualAssignmentTests`.
Thanks

garyrussell:
Back ported to 2.8.x; found a problem during the back port - fixed there, as well as on main and 2.9.x a63359c1

Issue #2340, opened 2022-07-09
Losing messages in RetryingBatchErrorHandler/ErrorHandlingUtils during rebalancing


vooft:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

For example:

2.3.7+

**Describe the bug**

If a listener doing a retry and rabalancing kicks in, then the consumer will be unpaused and every consumer.poll() invocation will start returning records and the error handler will throw them away. 

It seems that the rebalancing listener KafkaMessageListenerContainer can re-pause the consumer again, but it is not aware that the error handler paused it.

**To Reproduce**

Create a listener with infinite retry RetryingBatchErrorHandler that will be constantly failing and force rebalancing. After rebalancing finished RetryingBatchErrorHandler will drain all the messages from assigned partitions without any processing.

**Expected behavior**

RetryingBatchErrorHandler should re-pause consumer after rebalancing.

**Sample**

Test is using RetryingBatchErrorHandler, but logic in ErrorHandlingUtils is essentially the same
https://github.com/vooft/kafka-retry-issue/blob/master/src/test/java/com/example/kafkaissue/KafkaIssueTest.java

vooft:
Thank you for your answer @garyrussell 
I'm looking at `DefaultErrorHander` and it seems that if you just throw an arbitrary exception from the listener it will still go all the way to `ErrorHandlingUtils` and will have the same issue as in `RetryingBatchErrorHandler`, or am I missing something?

garyrussell:
Yes; I agree; I was just pointing out that the recovering mechanism is preferred to the (older) retrying mechanism that suffers from this bug. I will have a fix soon.

garyrussell:
@vooft 2.8.8 with the fix is now in Maven Central.

PR #2341, opened 2022-07-11
GH-2340: Fix Retrying Batch Error Handling

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2340

The `RetryingBatchErrorHandler` - now called the `FallbackBatchErrorHandler`
pauses and resumes the consumer during retries, to allow it to poll the
consumer to avoid a forced rebalance.

However, if a normal rebalance occurs, for example if a new member joins,
the error handler does not re-pause the consumer and silently consumes
new records.

Add a mechanism to always re-pause the consume when in this retry mode.

As well as the unit test, tested proper functionality by running two copies of a Boot app.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/FallbackBatchErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/FallbackBatchErrorHandler.java
@@ -16,11 +16,13 @@
 
 package org.springframework.kafka.listener;
 
+import java.util.Collection;
 import java.util.function.BiConsumer;
 
 import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.common.TopicPartition;
 
 import org.springframework.core.log.LogAccessor;
 import org.springframework.lang.Nullable;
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/FallbackBatchErrorHandler.java
```diff
@@ -54,6 +56,8 @@ class FallbackBatchErrorHandler extends KafkaExceptionLogLevelAware
 
 	private boolean ackAfterHandle = true;
 
+	private boolean retrying;
```
artembilan, line 59: `volatile`?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/FallbackBatchErrorHandler.java
garyrussell: No need; always used from the consumer thread.

artembilan: So, you mean that `onPartitionsAssigned()` is called withing that `ErrorHandlingUtils.retryBatch()` process?
Otherwise I don't see how this variable is involved...

Thanks

garyrussell: `ErrorHandlingUtils.retryBatch` pauses the consumer and polls in a while loop until retries are exhausted, and then resumes.

We need to repause the consumer if `onPartitionsAssigned()` is called while we are in that loop (it is called from `poll()` in the kafka-clients).

So the thread that tests this field is the same thread that is looping within `retryBatch`.

artembilan:
... and cherry-picked to `2.9.x` .

It does not apply clearly to `2.8.x`. 
Would you mind taking a look, @garyrussell ?

Thanks

garyrussell:
Back-ported to 2.8.x

PR #2343, opened 2022-07-11
GH-2239: Fix Boot AutoConfiguration

garyrussell:
See https://github.com/spring-projects/spring-kafka/issues/2239

The previous commit removed the implicit bootstrap in favor of enforcing the
user to use the `@EnableKafkaRretryTopic` or explicitly extend
`RetryTopicConfigurationSupport`.

Unfortunately, this breaks Spring Boot because it can auto configure a
`RetryTopicConfiguration` bean, which means the infrastructure beans are required.

Fallback to late binding of the infrastructure beans if a `RetryTopicConfiguration`
bean is found in the application context.

Tested with a Boot app.

**cherry-pick to main**


artembilan:
... and cherry-picked to `main`

Issue #2357, opened 2022-07-19
Change APIs to use CompletableFuture Instead of ListenableFuture


garyrussell:
Spring is planning on deprecating ListenableFuture in 6.0.

https://github.com/spring-projects/spring-framework/issues/27780

Add methods that use `CompletableFuture` to the templates and deprecate those that return `ListenableFuture` so we can remove them and have a clean API in 3.0.

PR #2360, opened 2022-07-20
GH-2357: Switch to CompletableFuture

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2357

Spring Framework is planning to deprecate `ListenableFuture` in 6.0.

Add methods to the `KafkaOperations` (`KafkaTemplate`) that return
`CompletableFuture` instead; the `ListenableFuture` methods have now
been removed in 3.0.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations2.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations2.java
@@ -33,11 +33,12 @@ import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
 
+import org.springframework.kafka.core.KafkaOperations.OperationsCallback;
+import org.springframework.kafka.core.KafkaOperations.ProducerCallback;
 import org.springframework.kafka.support.SendResult;
 import org.springframework.kafka.support.TopicPartitionOffset;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
-import org.springframework.util.concurrent.ListenableFuture;
 
 /**
  * The basic Kafka operations contract returning {@link CompletableFuture}s.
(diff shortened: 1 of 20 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations2.java
```diff
@@ -0,0 +1,281 @@
...
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ *
+ * @author Gary Russell
+ * @since 2.9
+ * @deprecated no longer needed; use {@code KafkaOperations}.
+ */
+@Deprecated
+public interface KafkaOperations2<K, V> {
```
artembilan, line 54: Why cannot we remove it altogether?
I believe there is going to a migration from `2.8.x` to `3.0.0` mostly through Spring Boot, so that replacement in the API of the `KafkaOperations` is going to be a breaking change anyway.
Therefore this one is just an extra burden for us to bear for the next version.

What do I miss?

Thanks

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations2.java
garyrussell: I was thinking of people that had taken advantage of the migration path provided by 2.9.x...

	KafkaOperations2<Integer, String> templateTx = new KafkaTemplate<>(pfTx).usingCompletableFuture();

This code would still work with 3.0 (with a deprecation warning only).

But, I suppose with a major release, this breaking change might be acceptable.

WDYT?

artembilan: Yes, that's my idea as well: bite the bullet and remove. 
One more breaking change won't hurt as it is already with those replaced return types.

Issue #2359, opened 2022-07-20
Add Support for Spring AOT


garyrussell:
Migrate hints from `spring-native`.

PR #2364, opened 2022-07-21
GH-2359: Initial AOT Support for Native

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2359

This is sufficient for the basic Kafka smoke test.

I will issue another PR if changes are needed for the Streams and Avro versions.
I have a `KafkaAvroBeanRegistrationAotProcessor` but I have omitted it from
this commit because it hasn't been tested.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaAvroBeanRegistrationAotProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaAvroBeanRegistrationAotProcessor.java
@@ -48,9 +48,6 @@ import org.springframework.util.ReflectionUtils;
  */
 public class KafkaAvroBeanRegistrationAotProcessor implements BeanRegistrationAotProcessor {
 
-	private static final boolean AVRO_PRESENT = ClassUtils
-			.isPresent("io.confluent.kafka.serializers.KafkaAvroSerializer", null);
-
 	private static final String KAFKA_LISTENER_CLASS_NAME = KafkaListener.class.getName();
 
 	private static final String KAFKA_HANDLER_CLASS_NAME = KafkaHandler.class.getName();
@@ -60,6 +57,8 @@ public class KafkaAvroBeanRegistrationAotProcessor implements BeanRegistrationAo
 	private static final String CONSUMER_RECORDS_CLASS_NAME = ConsumerRecord.class.getName();
 
 	private static final String AVRO_GENERATED_CLASS_NAME = "org.apache.avro.specific.AvroGenerated";
+
+	private static final boolean AVRO_PRESENT = ClassUtils.isPresent(AVRO_GENERATED_CLASS_NAME, null);
 
 	@Override
 	@Nullable
--- a/spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaAvroBeanRegistrationAotProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaAvroBeanRegistrationAotProcessor.java
@@ -98,14 +98,14 @@ public class KafkaAvroBeanRegistrationAotProcessor implements BeanRegistrationAo
 
 	private static boolean isListener(Class<?> beanType) {
 		return GenericMessageListener.class.isAssignableFrom(beanType)
-				|| MergedAnnotations.from(beanType).get(KAFKA_LISTENER_CLASS_NAME).isPresent();
+				|| MergedAnnotations.from(beanType).isPresent(KAFKA_LISTENER_CLASS_NAME);
 	}
 
 	private static boolean hasListenerMethods(Class<?> beanType) {
 		for (Method method : beanType.getDeclaredMethods()) {
 			MergedAnnotations methodAnnotations = MergedAnnotations.from(method);
-			if (methodAnnotations.get(KAFKA_LISTENER_CLASS_NAME).isPresent()
-					|| methodAnnotations.get(KAFKA_HANDLER_CLASS_NAME).isPresent()) {
+			if (methodAnnotations.isPresent(KAFKA_LISTENER_CLASS_NAME)
+					|| methodAnnotations.isPresent(KAFKA_HANDLER_CLASS_NAME)) {
 				return true;
 			}
 		}
--- a/spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaAvroBeanRegistrationAotProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaAvroBeanRegistrationAotProcessor.java
@@ -16,7 +16,6 @@
 
 package org.springframework.kafka.aot;
 
-import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.HashSet;
@@ -32,8 +31,6 @@ import org.springframework.beans.factory.aot.BeanRegistrationAotProcessor;
 import org.springframework.beans.factory.support.RegisteredBean;
 import org.springframework.core.ResolvableType;
 import org.springframework.core.annotation.MergedAnnotations;
-import org.springframework.kafka.annotation.KafkaHandler;
-import org.springframework.kafka.annotation.KafkaListener;
 import org.springframework.kafka.listener.GenericMessageListener;
 import org.springframework.lang.Nullable;
 import org.springframework.util.ClassUtils;
@@ -47,10 +44,6 @@ import org.springframework.util.ReflectionUtils;
  *
  */
 public class KafkaAvroBeanRegistrationAotProcessor implements BeanRegistrationAotProcessor {
-
-	private static final String KAFKA_LISTENER_CLASS_NAME = KafkaListener.class.getName();
-
-	private static final String KAFKA_HANDLER_CLASS_NAME = KafkaHandler.class.getName();
 
 	private static final String CONSUMER_RECORD_CLASS_NAME = ConsumerRecord.class.getName();
 
@@ -67,7 +60,7 @@ public class KafkaAvroBeanRegistrationAotProcessor implements BeanRegistrationAo
 			return null;
 		}
 		Class<?> beanType = registeredBean.getBeanClass();
-		if (!isListener(beanType) && !hasListenerMethods(beanType)) {
+		if (!isListener(beanType)) {
 			return null;
 		}
 		Set<Class<?>> avroTypes = new HashSet<>();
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
```diff
@@ -0,0 +1,253 @@
...
+				.forEach(type -> reflectionHints.registerType(type, builder ->
+						builder.withMembers(MemberCategory.INVOKE_PUBLIC_CONSTRUCTORS)));
+
+		reflectionHints.registerType(TypeReference.of("java.util.zip.CRC32C"), builder ->
+				builder.withMembers(MemberCategory.INVOKE_PUBLIC_CONSTRUCTORS));
+
+		hints.proxies().registerJdkProxy(Consumer.class, SpringProxy.class, Advised.class, DecoratingProxy.class);
+		hints.proxies().registerJdkProxy(Producer.class, SpringProxy.class, Advised.class, DecoratingProxy.class);
+
+		if (ClassUtils.isPresent("org.apache.kafka.streams.StreamsBuilder", null)) {
```
artembilan, line 225: I guess it is preferable to use a dedicated for this purposes API on the hints: `registerTypeIfPresent()`

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
```diff
@@ -0,0 +1,253 @@
...
+	@Override
+	public void registerHints(RuntimeHints hints, @Nullable ClassLoader classLoader) {
+		ReflectionHints reflectionHints = hints.reflection();
+		Stream.of(
+					PartitionOffset.class,
+					TopicPartition.class,
+					ConsumerProperties.class,
+					ContainerProperties.class,
+					ProducerListener.class,
+					KafkaListener.class,
```
artembilan, line 125: I believe annotations must be registers with their special API like you did in Spring AMQP.

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
```diff
@@ -0,0 +1,253 @@
...
+					EnableKafka.class)
```
artembilan, line 126: I don't think we need this annotation to expose...

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
```diff
@@ -0,0 +1,253 @@
...
+					KafkaOperations.class,
+					ConsumerFactory.class,
+					LoggingProducerListener.class)
+				.forEach(type -> reflectionHints.registerType(type,
+						builder -> builder.withMembers(MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
+								MemberCategory.INVOKE_DECLARED_METHODS,
+								MemberCategory.INTROSPECT_PUBLIC_METHODS)));
+
+		Stream.of(
+					Message.class,
```
artembilan, line 157: Looks like you have duplicated this type.
I also wonder if we really use it via reflection at all...
In SI I registered it only for serialization.

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
```diff
@@ -0,0 +1,253 @@
...
+					KafkaListenerAnnotationBeanPostProcessor.class)
+				.forEach(type -> reflectionHints.registerType(type,
+						builder -> builder.withMembers(MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
+								MemberCategory.INVOKE_DECLARED_METHODS,
+								MemberCategory.INTROSPECT_PUBLIC_METHODS)));
+
+		Stream.of(
+					KafkaBootstrapConfiguration.class,
+					CreatableTopic.class,
+					KafkaListenerEndpointRegistry.class)
```
artembilan, line 171: I wonder if you tried to go baby steps with all of these types registration.
Looks like you just ported whatever there was in Spring Native...

artembilan (changes requested):
I wonder if Avro stuff has to be somewhere else: doesn't look like it is just Kafka-specific...

garyrussell:
The avro processor is specific to spring-kafka because it discovers listeners that use Avro types and registers them for reflection.

I left it out of this PR to avoid any confusion (and because I haven't tested it yet).

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
garyrussell: The problem with that is I have to use FQ class names instead of the classes themselves.

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
garyrussell: That is exactly correct; I would rather get it working quickly and then we can prune it later.

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
garyrussell: Again, I just copied what was in spring-native, for now.

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
```diff
@@ -115,24 +116,23 @@ public class KafkaRuntimeHintsRegistrar implements RuntimeHintsRegistrar {
 
 	@Override
 	public void registerHints(RuntimeHints hints, @Nullable ClassLoader classLoader) {
+		RuntimeHintsUtils.registerAnnotation(hints, KafkaListener.class);
+		RuntimeHintsUtils.registerAnnotation(hints, KafkaListeners.class);
+		RuntimeHintsUtils.registerAnnotation(hints, EnableKafka.class);
+		RuntimeHintsUtils.registerAnnotation(hints, PartitionOffset.class);
+		RuntimeHintsUtils.registerAnnotation(hints, TopicPartition.class);
+		RuntimeHintsUtils.registerAnnotation(hints, MessageMapping.class);
```
artembilan, line 124: This annotation is registered by Spring Framework by itself. See `MessagingAnnotationsRuntimeHints`

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
```diff
@@ -115,24 +116,23 @@ public class KafkaRuntimeHintsRegistrar implements RuntimeHintsRegistrar {
 
 	@Override
 	public void registerHints(RuntimeHints hints, @Nullable ClassLoader classLoader) {
+		RuntimeHintsUtils.registerAnnotation(hints, KafkaListener.class);
+		RuntimeHintsUtils.registerAnnotation(hints, KafkaListeners.class);
+		RuntimeHintsUtils.registerAnnotation(hints, EnableKafka.class);
```
artembilan, line 121: I still don't think we need this. The `@Import` processing is initiated by AOT. So, that our `KafkaBootstrapConfiguration` is already registered at runtime.
Therefore I don't see a reason in this registration.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
```diff
@@ -0,0 +1,239 @@
...
+					DelegatingSerializer.class,
+					JsonDeserializer.class,
+					JsonSerializer.class,
+					ParseStringDeserializer.class,
+					StringOrBytesSerializer.class,
+					ToStringSerializer.class)
+				.forEach(type -> reflectionHints.registerType(type, builder ->
+						builder.withMembers(MemberCategory.INVOKE_PUBLIC_CONSTRUCTORS)));
+
+		reflectionHints.registerType(TypeReference.of("java.util.zip.CRC32C"), builder ->
```
artembilan, line 205: Why FQCN? Look like this class is a part of Java by itself

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
artembilan: Are you going to fix this via `onReachableType(TypeReference.of("org.apache.kafka.streams.StreamsBuilder")` ?

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
```diff
@@ -0,0 +1,239 @@
...
+				builder.withMembers(MemberCategory.INVOKE_PUBLIC_CONSTRUCTORS));
+
+		hints.proxies().registerJdkProxy(Consumer.class, SpringProxy.class, Advised.class, DecoratingProxy.class);
+		hints.proxies().registerJdkProxy(Producer.class, SpringProxy.class, Advised.class, DecoratingProxy.class);
```
artembilan, line 209: `AopProxyUtils.completeJdkProxyInterfaces()` ?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
garyrussell: Again; copied from spring-native.

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHintsRegistrar.java
garyrussell: >That is correct: there is a variant for fqcn and classLoader

Right converting all those classes to FQCN is a pain.

Where is `onReachableType`?

artembilan: See how I used it in my SI PR: https://github.com/spring-projects/spring-integration/pull/3832/files#diff-b471878878aa8838786d4a93b19f8014d497db6ace582572f834da814720e01bR97

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaAvroBeanRegistrationAotProcessor.java
```diff
@@ -0,0 +1,175 @@
...
+				avroTypes.forEach(type -> reflectionHints.registerType(type,
+						builder -> builder.withMembers(MemberCategory.INVOKE_DECLARED_CONSTRUCTORS)));
+			};
+		}
+		return null;
+	}
+
+	private static boolean isListener(Class<?> beanType) {
+		return GenericMessageListener.class.isAssignableFrom(beanType)
+				|| MergedAnnotations.from(beanType).get(KAFKA_LISTENER_CLASS_NAME).isPresent();
```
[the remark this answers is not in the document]
garyrussell: Actually, it turns out we don't need to process the annotations, only `GenericMessageListener`s that use Avro types - the ones that are method parameters are already discovered.

artembilan (approved):
LGTM

Let me know if anything else you'd like to add to this PR or it is good to go for merging!

garyrussell:
Thanks; all smoke tests are green, but I will add a couple more before merging this.

garyrussell:
Ok to merge after this final commit; thanks!

PR #2384, opened 2022-08-30
Fix typo in whats-new.adoc

chanhyeong:
Can't move to table row "interceptBeforeTX" because there  is a wrong letter case.
I changed the capital letter "X" to lowercase in `whats-new.adoc`.
(https://docs.spring.io/spring-kafka/reference/html/#x29-lc-changes)


chanhyeong:
I also raised PR #2385 because the `changes-since-1.0.adoc` file in main branch has a same bug.

Issue #2400, opened 2022-09-20
ReplyingKT - Improve Readability of Correlation Id in Logs


garyrussell:
Currently, the `CorrelationKey` uses a `BigInteger` in `toString()`.

Represent the value as a hex string instead.

Related to #2399

PR #2401, opened 2022-09-20
GH-2400: ReplyingKT Improve CorrelationId Logging

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2400

Use a hex string instead of a `BigInteger` in `toString`. Also, if the bytes are the length of a UUID, use the standard representation thereof.

**cherry-pick to 2.9.x**


artembilan:

Issue #2399, opened 2022-09-19
ReplyingKT - Add Simpler Text-Based Correlation Mechanism


saurav28:
We are using AggregatingKafkaReplyingTemplate to achieve Request Reply pattern with the interacting microservices

For the testing purpose i have a spring boot producer application which produces the message on the request topic and expects a message back from the consumer on the response topic. 

I am using a Local UI as a consumer which will publish the response message manually as soon as the request message is sent.

For this i would need to send the Kafka internal CorrelationId back to the broker.

Is there any way to log/know the generated string value (not the bytes) CorrelationId by Kafka template in the Spring boot app itself (in logs maybe ) ? Can i take that value and put it in the local Kafka UI to send a message back in the receiving topic ?

i tried getting the UUID generated from Kafka template through debugging in the template class and sending it through my local Kafka UI. But then the correlation validation fails and i get this message.

`No pending reply: <CorrelationId> perhaps timed out, or using a shared reply topic`

I checked the actual UUID is not logged but the hash key of the byte value and bytes itself of the UUID is logged.

saurav28:
@garyrussell thanks for the reply.

One part of my question was to put this is as a requirement to log the CorrelationId header through ReplyingKafkaTemplate. Currently it only logs the BigInteger value.

I tried plugging in one Producer Interceptor and printed the Kafka CorrelationId like below

` if("CorrelationId".equals(header.key()) )
            {

                ByteBuffer bb = ByteBuffer.wrap(header.value());
                long firstLong = bb.getLong();
                long secondLong = bb.getLong();
                UUID uuid = new UUID(firstLong, secondLong);
            System.out.println( " header key " + header.key() + "header value " + uuid);
            }`

It is printing the same CorrelationId generated through the DefaultCorrelationStrategy.

But as i mentioned above if i use this CorrelationId and publish the message manually through an UI, then this fails and isPending check  in ReplyingKafkaTemplate returns false.

garyrussell:
The correlationId is currently a binary value derived from a UUID, rather than a simple string representation of the UUID, so it can't be "entered on a UI" (unless you control the UI).

We could add a simpler correlation option, but here is a work around:

```java
public class Interceptor<K, V, R> implements ProducerInterceptor<K, V>, ConsumerInterceptor<K, R> {

	private static final ConcurrentHashMap<String, byte[]> map = new ConcurrentHashMap<>();

	@Override
	public void configure(Map<String, ?> configs) {
	}

[...]

PR #2402, opened 2022-09-20
GH-2399: ReplyingKT Human Readable Correlation

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2399

By default, the correlationId header contains a binary representation of a UUID; add an option to use a String representation instead.

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
@@ -493,7 +493,7 @@ public class ReplyingKafkaTemplateTests {
 	}
 
 	@Test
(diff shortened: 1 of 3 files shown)


spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
```diff
@@ -472,6 +492,35 @@ public void testAggregateNormal() throws Exception {
 		}
 	}
 
+	@Test
+	public void testAggregateNormaStringCorrelationl() throws Exception {
```
artembilan, line 496: Nit-pick, but typo in the test method:
```suggestion
	public void testAggregateNormalStringCorrelation() throws Exception {
```
😄

artembilan (changes requested):

PR #2406, opened 2022-09-21
GH-2404: Refactor FallbackBatchErrorHandler

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2404

- implement `CommonErrorHandler` instead of using an adapter
- populate its retry listeners and log level if needed
- remove listeners thread local in `ErrorHandlingUtils`, needed for previous implementation
- remove final use of `SeekToCurrentBatchErrorHandler` - used in the fallback if recovery fails or the thread is interrupted while sleeping
- remove `SeekToCurrentBatchErrorHandler` and its tests


spring-kafka/src/main/java/org/springframework/kafka/listener/FallbackBatchErrorHandler.java
```diff
@@ -122,4 +141,23 @@ public void onPartitionsAssigned(Consumer<?, ?> consumer, Collection<TopicPartit
 		}
 	}
 
+	private final class SeekAfterRecoverFailsOrInterrupted implements CommonErrorHandler {
```
artembilan, line 144: Don't we need to have a package protected ctor to avoid extra generated code around constructor ?
Or new Java does that optimally already?

artembilan (changes requested):
Just one nit-pick.

spring-kafka/src/main/java/org/springframework/kafka/listener/FallbackBatchErrorHandler.java
garyrussell: Good catch; thanks.

Issue #2419, opened 2022-09-29
Return a reasonable default instead of throwing a UnsupportedOperationException


aadamovi:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

For example:

2.8.5

Between 2.7.0 and 2.8.2

**Describe the bug**
This request is based on the discussion on Stackoverflow: https://stackoverflow.com/questions/73894535/unsupportedoperationexception-when-using-retryabletopic

In short, the producer factory wrapper should implement getConfigurationProperties() and call the delegate factory to get its properties. In a case when the producer factory doesn't honor this contract instead of failing with a java.lang.UnsupportedOperationException perhaps the code could fall back to some default value.

**To Reproduce**

**Expected behavior**

I would expect for the value to be set to some reasonable default instead of throwing a UnsupportedOperationException

**Sample**
The URL contains a full description and a small sample of how the @RetryableTopic was used.

https://stackoverflow.com/questions/73894535/unsupportedoperationexception-when-using-retryabletopic

aadamovi:
thank you for taking care of this. 

one last thing, what 2.8.X version is this going to get into?

artembilan:
See release schedule here: https://github.com/spring-projects/spring-kafka/milestones

PR #2421, opened 2022-09-29
GH-2419: DLPR: Protect Against Non-Compliant PF

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2419

We try to obtain the configured request timeout using the producer factory's configuration properties. Some wrappers (e.g. open tracing) do not implement this method, so fallback to the default instead of throwing an USOE.


artembilan:
... and cherry-picked to 2.9.x & 2.8.x after fixing conflicts in tests.

Issue #2432, opened 2022-10-12
Retry-topics auto creation also creates the main topic


jgslima:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**
2.9.1
Actually I already saw this behaviour since 2.8, not sure the oldest version that has this behaviour.

**Describe the bug**
When the `autoCreateTopics` feature of the Non-blocking retries main feature is activated, the framework also creates the main topic with the configurations (partitions and replicas) of the retry topics, even if the context has as `NewTopic` bean for the main topic with different configurations.

**To Reproduce**

Run the application below:

```
@SpringBootApplication
@EnableScheduling
@EnableKafkaRetryTopic
@Slf4j
public class SpringKafkaStudies {

	public static void main(String[] args) {
		SpringApplication.run(SpringKafkaStudies.class, args);
	}

	@Bean
	public NewTopic mainTopic() {
		return TopicBuilder
				.name("mytopic")
				.partitions(10)
				.replicas(2)
				.build();
	}

	@RetryableTopic(attempts = "3",
			backoff = @Backoff(delayExpression = "5000"),
			fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC,
			autoCreateTopics = "true",
			numPartitions = "5",
			replicationFactor = "1")
	@KafkaListener(topics = "mytopic", concurrency = "3")
	public void nonBlockingRetriesListen(ConsumerRecord<String, String> record) {
		log.info("Received record: {}", record.value());
	}
	
}
```

**Expected behavior**
The main topic `"mytopic"` to be created with 10 partitions and 2 replicas, and the retry and DLT topic to be created with 5 partitions and 1 replicas.
But actually, the main topic `"mytopic"` is also created with 5 partitions and 1 replica.

**Sample**
See the code above.

jgslima:
Indeed, as the default behaviour of spring-kafka, when sending the record to the retry or DLT topics, is to assume that the partition in those topics should be the same as the original topic, this means that the retry topics should have the same amount of partitions (although I think in this case that the default behaviour of `DeadLetterPublishingRecovererFactory.resolveTopicPartition()` should be to not specify/resolve the partition, leaving this to the producer).

However the claim to have a distinct configuration of _replication factor_ should be enough to expect that the framework might have at least some configuration to indicate that, the topics auto-creation feature of the retry mechanism should be done solely for the retry and DLT topics.

garyrussell:
[...]

PR #2437, opened 2022-10-12
GH-2432: Fix Retryable Topic Provisioning

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2432

Don't provision an individual retry topic bean, if there is already a `NewTopic` bean with the same topic name.


garyrussell:
Problems resolved.

artembilan:
... and cherry-picked to `2.9.x` & `2.8.x` after some minor conflict resolution.

Issue #2282, opened 2022-05-23
Multiple @RetryableTopic on the same topic


loonis:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?** : 2.8.6

I have 2 listeners (2 groupIds) : service-1 & service-2 listening  the same topics : `myTopic`

When service-1 fails, I expect the message to be pushed in the retry queue : `myTopic-service_1_retry-0`
Similarly, when service-2 fails, I expect the message to be pushed in the retry queue : `myTopic-service_2_retry-0`

Here is an example of my listeners.

```java
@Service
@Slf4j
@RequiredArgsConstructor
public class MyListeners {

  @RetryableTopic(
      attempts = "2",
      numPartitions = "9",
      kafkaTemplate = "kafkaTemplate1",
      backoff = @Backoff(delay = 30000),
      topicSuffixingStrategy = TopicSuffixingStrategy.SUFFIX_WITH_INDEX_VALUE,
      retryTopicSuffix = "-service_1_retry",
      listenerContainerFactory = "kafkaListenerContainerFactory1",
      dltTopicSuffix = "-service_1_dlt")
  @KafkaListener(
      containerFactory = "kafkaListenerContainerFactor1",
      groupId = "service-1",
      concurrency = "9",
      topics = "myTopic",
      errorHandler = "logExceptionErrorHandler")
  public void processMessageService1(ConsumerRecord<String, Event> message) {
      throw new IllegalStateException("test");
  }

  @RetryableTopic(
      attempts = "2",
      kafkaTemplate = "kafkaTemplate2",
      numPartitions = "9",
      backoff = @Backoff(delay= 30000),
      topicSuffixingStrategy = TopicSuffixingStrategy.SUFFIX_WITH_INDEX_VALUE,
      retryTopicSuffix = "-service_2_retry",
      listenerContainerFactory = "kafkaListenerContainerFactory2",
      dltTopicSuffix = "-service_2_dlt")
  @KafkaListener(
      containerFactory = "kafkaListenerContainerFactory2",
      groupId = "service-2",
      concurrency = "9",
      topics = "myTopic",
      errorHandler = "logExceptionErrorHandler")
  public void processMessageService2(ConsumerRecord<String, Event> message) {
    return;
  }

}
```

When I push a message in the `myTopic` topic and it fails for service-**1**, the message is being pushed in the myTopic-service_**2**_retry-0, which is not what I expect.

For more details, here is the KafkaConfig file I use
```java

[...]

PR #2442, opened 2022-10-18
GH-2282: Fix Multi RetryableTopic Same Topic

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2282

### DIFF
--- a/spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
@@ -775,6 +775,37 @@ public class CustomRetryTopicNamesProviderFactory implements RetryTopicNamesProv
 ----
 ====
 
+[[multi-retry]]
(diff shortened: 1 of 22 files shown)


garyrussell:
cc/ @tomazfernandes FYI - we'd appreciate a review if you have time; no rush; RC2 is scheduled in 2 weeks' time.

tomazfernandes:
Sure @garyrussell!

The solution looks great, just one detail in the docs.

I wonder though if it'd make sense to create a `RetryTopicEndpoint` subclass so we can add the `mainListenerId` property, and whatever properties we need for retryable topics in the future, without adding complexity to the existing `Endpoint` classes and to the classes that deal with them.

WDYT?

tomazfernandes:
Sure, looks good to me then!

tomazfernandes (approved):
LGTM!

artembilan (changes requested):

Issue #2443, opened 2022-10-18
Distinct concurrency for retry and DLT topic containers


garyrussell:
### Discussed in https://github.com/spring-projects/spring-kafka/discussions/2434

<div type='discussions-op-text'>

<sup>Originally posted by **jgslima** October 11, 2022</sup>
I am struggling to understand how to achieve a configuration where I can have the listener container for the retry and DLT topics with a concurrency different from the container of the main topic (the retry and DLT topics do not need the same amount of threads/consumers).

I have two beans of `ConcurrentKafkaListenerContainerFactory<?, ?>`:
- one with the name `kafkaListenerContainerFactory`, to be used as the main factory for the main containers.
- another one, dedicated to the retry/DLT containers,  configured with the desired `concurrency` for them.

Then, I already tried to force the use of the second factory for the retry/DTL containers with these 2 ways:
- specifying the factory name in the `listenerContainerFactory` attribute of the `@RetryableTopic` annotation.
- forcing the name of this factory to be `RetryTopicBeanNames.DEFAULT_LISTENER_CONTAINER_FACTORY_BEAN_NAME`.

But in the end, I always end up having, for the retry and DLT topic, an amount of containers/threads equal to the concurrency of the main listener (that is, the concurrency declared in the `@KafkaListener` annotation).

I tried to understand and debug `ListenerContainerFactoryResolver`, at some point it seems to resolve to the factory with the customized concurrency, but in the end for some reason the amount of containers/threads are egual to the main listener.
</div>

PR #2444, opened 2022-10-18
GH-2443: Distinct Concurrency for Retry Containers

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2443

Optionally configure a different concurrency for retry containers.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
@@ -80,6 +80,8 @@ public class RetryTopicConfigurationBuilder {
 
 	private Boolean autoStartDltHandler;
 
+	private Integer concurrency;
+
 	/* ---------------- DLT Behavior -------------- */
 	/**
 	 * Configure a DLT handler method.
@@ -90,6 +92,11 @@ public class RetryTopicConfigurationBuilder {
 	 */
 	public RetryTopicConfigurationBuilder dltHandlerMethod(String beanName, String methodName) {
 		this.dltHandlerMethod = RetryTopicConfigurer.createHandlerMethodWith(beanName, methodName);
+		return this;
+	}
+
+	public RetryTopicConfigurationBuilder concurrency(Integer concurrency) {
+		this.concurrency = concurrency;
 		return this;
 	}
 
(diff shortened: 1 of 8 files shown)


garyrussell:
cc/ @tomazfernandes

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
```diff
@@ -93,6 +95,11 @@ public RetryTopicConfigurationBuilder dltHandlerMethod(String beanName, String m
 		return this;
 	}
 
+	public RetryTopicConfigurationBuilder concurrency(Integer concurrency) {
```
tomazfernandes, line 98: I know most of the methods in this class don't have proper javadocs, but perhaps we should at least include in the new ones?

tomazfernandes (commented):
LGTM, just one minor comment

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
tomazfernandes: BTW, do you think it'd be worth it adding javadocs for the other methods as well? I can open a PR with that if you'd like.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
```diff
@@ -190,15 +254,37 @@ public RetryTopicConfigurationBuilder maxAttempts(int maxAttempts) {
 		return this;
 	}
 
+	/**
```
tomazfernandes, line 257: Maybe we should add that this is in milliseconds? Had to go through the code to figure out 😄

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
```diff
@@ -138,50 +157,95 @@ public RetryTopicConfigurationBuilder autoStartDltHandler(@Nullable Boolean auto
...
+	 */
 	public RetryTopicConfigurationBuilder setTopicSuffixingStrategy(TopicSuffixingStrategy topicSuffixingStrategy) {
 		this.topicSuffixingStrategy = topicSuffixingStrategy;
 		return this;
 	}
 
 	/* ---------------- Configure BackOff -------------- */
 
+	/**
+	 * Configure the maximum delivery attempts.
```
tomazfernandes, line 245: Maybe add that this includes the first attempt? Or perhaps it's not necessary?

tomazfernandes (commented):
Awesome, thanks!

Just a couple comments

tomazfernandes (approved):
Looks great, thanks!

Issue #2451, opened 2022-10-20
Unable to create multiple listeners using same class with KafkaListener annotation at class level


bdbogjoe:
There is issue in this line, we have to exclude the current bean if has no annotation in methods **or** in class level, should be something like : 

```
if (annotatedMethods.isEmpty() && ! hasClassLevelListeners) {
          this.nonAnnotatedClasses.add(bean.getClass());
          this.logger.trace(() -> "No @KafkaListener annotations found on bean type: " + bean.getClass());
}
```

https://github.com/spring-projects/spring-kafka/blob/c6bbba8b55406ff6ba605f350959ab070bc0cfea/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java#L380

garyrussell:
Thanks for reporting.

Curious: How did you create an issue with no labels? The [New Issue]https://github.com/spring-projects/spring-kafka/issues/new/choose page does not have an option for that.

PR #2453, opened 2022-10-20
GH-2451: Fix Class Level Listener Multi Instances

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2451

Classes with class level `@KafkaListener` were incorrectly added to the `nonAnnotatedClasses` set, preventing multiple instances of the same class to be registered as listeners.


artembilan:

Issue #2459, opened 2022-10-24
FallbackBatchErrorHandler is not an ExceptionClassifier


garyrussell:
Need to populate not retryable exceptions from the main error handler to the fallback.

garyrussell:
https://stackoverflow.com/questions/74151506/spring-boot-kafka-batch-defaulterrorhandler-addnotretryableexceptions/74182263#74182263

PR #2461, opened 2022-10-25
GH-2459: FallbackBatchErrorHandler Retryable Ex

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2459

The `FallbackBatchErrorHandler` was not an `ExceptionClassifier`. The default error handler should propagate exception classifications.

**3.0.x - I will back port**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ExceptionClassifier.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ExceptionClassifier.java
@@ -21,6 +21,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.springframework.classify.BinaryExceptionClassifier;
 import org.springframework.kafka.support.converter.ConversionException;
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ExceptionClassifier.java
```diff
@@ -149,6 +150,16 @@ public void setClassifications(Map<Class<? extends Throwable>, Boolean> classifi
...
+		notRetryable(Arrays.stream(exceptionTypes));
+	}
+
+	/**
+	 * Subclasses can override this to receive notification of configuration of not
+	 * retryable exceptions.
+	 * @param notRetryable the not retryable exceptions.
+	 * @since 2.9.3
+	 */
+	protected void notRetryable(Stream<Class<? extends Exception>> notRetryable) {
```
artembilan, line 162: Why `Stream`?
What is wrong with a plain `Class<? extends Exception>...` propagation?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/ExceptionClassifier.java
garyrussell: Because `@SafeVarags` requires a `final` modifier.

Issue #2464, opened 2022-10-26
kafka_deliveryAttempt header with initial value 2 in non-blocking retries


jgslima:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**
2.9.1

**Describe the bug**
It is about the value of the Header "kafka_deliveryAttempt" (KafkaHeaders.DELIVERY_ATTEMPT).
When used in combination with non-blocking retries, from the 2nd non-blocking attempt, it starts with the value 2 instead of 1.

Example: in a configuration with 2 blocking retries (Backoff with 2 maxAttempts) and a RetryTopicConfiguration with .maxAttempts(3), one would expect the value of the Headers in each listener invocation to be:

kafka_deliveryAttempt=1, retry_topic-attempts=null
kafka_deliveryAttempt=2, retry_topic-attempts=null
kafka_deliveryAttempt=3, retry_topic-attempts=null

kafka_deliveryAttempt=1, retry_topic-attempts=2
kafka_deliveryAttempt=2, retry_topic-attempts=2
kafka_deliveryAttempt=3, retry_topic-attempts=2

kafka_deliveryAttempt=1, retry_topic-attempts=3
kafka_deliveryAttempt=2, retry_topic-attempts=3
kafka_deliveryAttempt=3, retry_topic-attempts=3

But instead we have:

kafka_deliveryAttempt=1, retry_topic-attempts=null
kafka_deliveryAttempt=2, retry_topic-attempts=null
kafka_deliveryAttempt=3, retry_topic-attempts=null

kafka_deliveryAttempt=**2**, retry_topic-attempts=2
kafka_deliveryAttempt=2, retry_topic-attempts=2
kafka_deliveryAttempt=3, retry_topic-attempts=2

kafka_deliveryAttempt=**2**, retry_topic-attempts=3
kafka_deliveryAttempt=2, retry_topic-attempts=3
kafka_deliveryAttempt=3, retry_topic-attempts=3

**To Reproduce**
Please see above.

**Expected behavior**
Please see above.

jgslima:
Please do not mind in trying to come up with a workaround solution. This is a small bug which we can live with until there is the actual fix.

garyrussell:
This is weird; it works fine in a JUnit integration test, but I see this behavior in a Boot app (unless I run it in a debugger, in which case it works there too). I thought maybe a race condition, but this state is maintained in a `ThreadLocal`.

Investigating...

garyrussell:
[...]

PR #2466, opened 2022-10-27
GH-2464: Fix Delivery Attempt Header (Rare)

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2464

`KafkaBackoffException` should never be classified as retryable; the design is such that those exceptions go straight to the recoverer, which will re-throw.

If, however, a user configures a super class as retryable (e.g. `RuntimeException`) then its classification wins.

This causes a new retry state (`FailedRecord`) to be started, causing an incorrect value of 2 instead of 1 on the first delivery from a retry topic.

**I will backport - conflicts expected**


garyrussell:
Back ported to 2.9.x only after resolving conflicts; not applicable to 2.8.x.

PR #2467, opened 2022-11-01
Add KafkaTestUtils methods using Duration for timeout

szpak:
To adjust codebase to modern JVM (and Kafka API itself - e.g. `poll(Duration)`). I've just bumped into that writing some tests recently. Old methods with "long timeout" are marked as deprecated and removed in separate commit.

It seems to be still a good moment before 3.0-final.

Btw, a PR for 2.9.x will be created with just method deprecation without the method removal.


artembilan (approved):
I'm OK with this.
Perhaps I'm too old being so comforted with plain `long` for timeouts after all those years working without `Duration`. I mean technically it doesn't matter for me what type for timeout we use, unless `Duration` is an object taking some memory.
But if modern world expects from us more obvious API, I'm OK to merge this.

PR #2468, opened 2022-11-01
Add KafkaTestUtils methods using Duration for timeout (2.9)

szpak:
To adjust codebase to modern JVM (and Kafka API itself - e.g. poll(Duration)). I've just bumped into that writing some tests recently. Old methods with "long timeout" are marked as deprecated and removed in separate commit.

A variant for 2.9. Methods are removed in PR for 3.x - https://github.com/spring-projects/spring-kafka/pull/2467.

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
@@ -136,10 +136,10 @@ public final class KafkaTestUtils {
 	 * @param <V> the value type.
 	 * @return the record.
 	 * @throws IllegalStateException if exactly one record is not received.
-	 * @see #getSingleRecord(Consumer, String, long)
+	 * @see #getSingleRecord(Consumer, String, Duration)
 	 */
 	public static <K, V> ConsumerRecord<K, V> getSingleRecord(Consumer<K, V> consumer, String topic) {
-		return getSingleRecord(consumer, topic, 60000); // NOSONAR magic #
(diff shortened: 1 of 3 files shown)


szpak:
Hmm, why CI build was not triggered for this PR?

artembilan:
The CI on PRs is triggered only against `main` branch.

why do we need this change at all ? What is “old” with plain `long`? More over we talk about just tests.

szpak:
> why do we need this change at all ? What is “old” with plain long?

Writing some tests recently, I wanted to use: 
> getOneRecord(String brokerAddresses, String group, String topic, int partition, boolean seekToLast, boolean commit, long timeout)

what unit is the timeout? milliseconds? Javadoc:

>  * @param timeout the timeout.

Cool. Some digging in source code - yes, in millis. Sure, in some other places Javadoc clearly says "timeout in millis" which is nice, but still - you have to download sources (in IDE) to check/verify it.

> More over we talk about just tests.

Tests should be maintainable as well. In many other places in our test suite, we use duration (for Awaitility, our utils, [spring-kafka](https://github.com/spring-projects/spring-kafka/blob/59eb2dccab277bf3a52bfb2fb56630d45e805379/spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java#L48) or even in Kafka Consumer itself [KafkaConsumer#poll(Duration)](https://kafka.apache.org/26/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#poll-java.time.Duration-) - please note that even there `poll(long)` is deprecated). Using the same input data format (which in addition is self explanatory) is - in my opinion - a good idea. And having 3.0 on the horizon, it could be a good time to perform that backwad incompatible change - see #2467 (for 3.0).

Issue #2481, opened 2022-11-16
Document prerequisites for spring-kafka-test and how to configure them


wilkinsona:
With thanks to @DennisBecker for [reporting this on Gitter](https://gitter.im/spring-projects/spring-boot?at=6374aacdb1a75964367434a5), `GlobalEmbeddedKafkaTestExecutionListener` requires `junit-platform-launcher` 1.8 or later. Surefire's default configuration doesn't meet this requirement which results in a failure when running `./mvnw test` on an app that depends on `spring-kafka-test`:

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.749 s
[INFO] Finished at: 2022-11-16T13:03:13Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.2:test (default-test) on project springtest: There are test failures.
[ERROR] 
[ERROR] Please refer to /Users/awilkinson/Downloads/springtest/target/surefire-reports for the individual test results.
[ERROR] Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
[ERROR] There was an error in the forked process
[ERROR] 'org.junit.platform.engine.ConfigurationParameters org.junit.platform.launcher.TestPlan.getConfigurationParameters()'
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
[ERROR] 'org.junit.platform.engine.ConfigurationParameters org.junit.platform.launcher.TestPlan.getConfigurationParameters()'
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:656)
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:282)
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:245)
[ERROR] 	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeProvider(AbstractSurefireMojo.java:1183)
[ERROR] 	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeAfterPreconditionsChecked(AbstractSurefireMojo.java:1011)
[...]

PR #2482, opened 2022-11-16
GH-2481: Global Embedded Broker and JUnit Platform

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2481

The Maven Surefire plugin uses an older version of JUnit platform which is incompatible with the Global embedded broker.

Check that the platform version is compatible before attempting to determine if a global embedded broker should be created.

Tested with a Spring Boot application.


spring-kafka-test/src/main/java/org/springframework/kafka/test/junit/GlobalEmbeddedKafkaTestExecutionListener.java
```diff
@@ -84,10 +84,27 @@ public class GlobalEmbeddedKafkaTestExecutionListener implements TestExecutionLi
 	public static final String BROKER_PROPERTIES_LOCATION_PROPERTY_NAME =
 			"spring.kafka.embedded.broker.properties.location";
 
+	private static final boolean JUNIT_PLATFORM_COMPATIBLE;
```
artembilan, line 87: Add you name to the `@author` list, please, 😄

artembilan (changes requested):

PR #2494, opened 2022-11-30
Mehrdad.karami enhancement

metao1:
- Some cosmetic code cleanup
- Some typo fixed both in doc and method names

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -3789,7 +3789,9 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 				KafkaMessageListenerContainer.this.logger
 						.debug(() -> KafkaMessageListenerContainer.this + " stopped normally");
 			}
-			this.callback.run();
+			if (this.callback != null) {
+				this.callback.run();
+			}
 		}
 
 	}
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1688,7 +1686,7 @@ private synchronized void captureOffsets(ConsumerRecords<K, V> records) {
 			}
 		}
 
-		private void checkRebalanceCommits() {
+		private void checkRebalancedCommits() {
```
garyrussell, line 1689: I prefer the existing name here.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2331,19 +2332,17 @@ private void commitOffsetsIfNeeded(final ConsumerRecords<K, V> records) {
...
 		private void batchInterceptAfter(ConsumerRecords<K, V> records, @Nullable Exception exception) {
-			if (this.commonBatchInterceptor != null) {
-				try {
-					if (exception == null) {
-						this.commonBatchInterceptor.success(records, this.consumer);
-					}
-					else {
-						this.commonBatchInterceptor.failure(records, exception, this.consumer);
-					}
+			try {
```
garyrussell, line 2335: This is incorrect; the interceptor can be null; it probably should be marked `@Nullable`.

garyrussell (changes requested):
Thanks.

I will look at this some more tomorrow, but a couple of comments so far.

artembilan (changes requested):
Thanks for contribution but why did you name the PR after yourself? 
Why just don't provide the the head line from commit message which GH suggests by default?
Also see this article for best practice around commit messages: https://cbea.ms/git-commit/

Plus: did you run your fix locally - `gradlew clean check`?
Looks like there is some unreachable method still in use:
```
 /home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java:3475: error: cannot find symbol
					checkRebalanceCommits();
```

garyrussell:
That reference to `checkRebalanceCommits()` was in new code merged Friday.

garyrussell:
That reference to `checkRebalanceCommits()` was in new code from a PR submitted Friday and merged today.

Issue #2496, opened 2022-12-05
Reuse retry topic for the maxInterval delay of exponential backoff


jgslima:
Related to the discussion below:

https://github.com/spring-projects/spring-kafka/discussions/2484

**Expected Behavior**
In order to allow the configuration of an exponential backoff policy that keeps trying for a long time (say some hours), to have some way to use a single retry topic for the attempts corresponding to the `maxInterval` delay. This makes it more viable to configure the policy to keep trying for a long time because a lot of topics are not needed.

**Current Behavior**
When using exponential backoff, the number of needed retry topics is always the configured `maxAttempts` minus 1.

**Context**
For instance, when configuring the exponential backoff with `initialInterval=1000`, `multiplier=2`, and `maxInterval=16000`, in order to keep trying for one hour, one would need to configure `maxAttempts` as 229, and by default the needed retry topics would be:

* -retry-1000
* -retry-2000
* -retry-4000
* -retry-8000
* -retry-16000-0
* -retry-16000-1
* -retry-16000-2
* ...
* -retry-16000-224

When using the strategy that reuses the retry topic for the same intervals, in the same configuration above the needed retry topics would be only 5:

* -retry-1000
* -retry-2000
* -retry-4000
* -retry-8000
* -retry-16000

jgslima:
> @jgslima what's the reason you don't want to just use a single retry topic for all intervals (like the single topic option for fixed delays)? That's what we'd like to do.
> 
> @garyrussell, anyone: why is the single retry topic option only supported for fixed backoff policies and not exponential or other sleeping backoff policies? Looking at the code I see it tracks the next sleep interval in a kafka header so doesn't require that it be part of the topic name.

It is not possible to use the single retry option for backoffs that may have distinct sleep times for distinct records. The general algorithm only works with the assumption that, in a partition (actually in the topic as a whole), records with higher offsets have should not be processed before records whith lower offsets. Because the Consumer is paused based on the sleep time of the next record.

The motivation to have this new kind of policy is to allow a configuration where you can have the first retries done quickly (with a low interval), keep retrying for a large amount of time (say 4 or 6 hours), but without having to have an excessive amount of retry topics. Currently, when using exponential backoff, you need a large amount of retries topics for this. See how I explained in the issue description.

ctoomey:
[...]

PR #2497, opened 2022-12-05
GH-2496: Reuse retry topic for maxInterval delay

jgslima:
Resolves spring-projects#2496

Created SameIntervalTopicReuseStrategy, used to keep retrying in the same retry topic when the delays of the last retries are the same.


spring-kafka/src/main/java/org/springframework/kafka/retrytopic/SameIntervalTopicReuseStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ *
+ * Strategy for topic reuse when multiple, sequential retries have the same backoff interval.
+ *
+ * @author João Lima
+ * @since 3.1
+ *
+ */
+public enum SameIntervalTopicReuseStrategy {
```
jgslima, line 27: Initially I had modeled something like `ExponentialBackoffMaxDelayTopicReuseStrategy`, but it would be too specific for exponential backoff. So I went with a more generic concept, with the logic in `DestinationTopicPropertiesFactory` also being more generic.

It turns out that, with this `SameIntervalTopicReuseStrategy` idea, the `FixedDelayStrategy` could be removed, because the logic implemented for `SameIntervalTopicReuseStrategy` itself would be enough to conclude that a single retry topic should be used also for fixed backoffs. Unfortunately, the feature went GA.

Maybe we may mark `FixedDelayStrategy` as deprecated, with it being removed and replaced by `SameIntervalTopicReuseStrategy` in some future major release.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DefaultDestinationTopicResolver.java
```diff
@@ -192,13 +193,26 @@ public void addDestinationTopics(String mainListenerId, List<DestinationTopic> d
...
 	}
 
+	private void validateDestinations(List<DestinationTopic> destinationsToAdd) {
+		for (int i = 0; i < destinationsToAdd.size(); i++) {
+			DestinationTopic destination = destinationsToAdd.get(i);
+			if (destination.isReusableRetryTopic()) {
+				Assert.isTrue((i == (destinationsToAdd.size() - 1) ||
+						((i == (destinationsToAdd.size() - 2)) && (destinationsToAdd.get(i + 1).isDltTopic()))),
+						String.format("In the destination topic chain, the type %s can only be "
+								+ "specified as the last retry topic.", Type.REUSABLE_RETRY_TOPIC));
```
jgslima, line 211: This invariant was implemented because right now, the idea is to have topic reuse only for exponential backoff, therefore for the `maxDelay` retries, which stay in the end of the chain.
In order to allow reuse of attempts with the same interval in the middle of the chain, more sophistication would be needed, maybe to have `DestinationTopic.Properties` to carry the `maxAttempts` of the specific topic.

But anyway, no existing backoff policy should have attempts sequences with a same interval in the middle of the chain.

garyrussell:
@jgslima Thanks for the PR; it might take a bit of time to get it reviewed; with the upcoming holidays (and after recently getting out major releases) we are beginning to wind down a little. After today, I will not be working until the 19th and then, only for a few days before breaking again until the New Year.

I just wanted to let you know, in case you thought we were ignoring you; your work is definitely appreciated.

Thanks.

jgslima:
> @jgslima Thanks for the PR; it might take a bit of time to get it reviewed; with the upcoming holidays (and after recently getting out major releases) we are beginning to wind down a little. After today, I will not be working until the 19th and then, only for a few days before breaking again until the New Year.
> 
> I just wanted to let you know, in case you thought we were ignoring you; your work is definitely appreciated.
> 
> Thanks.

Ok, good to know, have a nice vacation.

garyrussell:
@jgslima Sorry for the delay - I hope to get to this next week.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/SameIntervalTopicReuseStrategy.java
tomazfernandes: I think this makes total sense - fixed delays are only a specific case of the broader `topic resuse` concept. So IMHO having both concepts at the same time tend to be a bit confusing. I'm thinking we could deprecate the `FixedDelayStrategy` at once in this PR and use the new option. WDYT?

But for that to work, I think we'd need to adapt the `FixedDelayStrategy` part to respond to both configurations.

The part that called my attention is this one:
https://github.com/spring-projects/spring-kafka/blob/3f645b11a1a2f3afa6a8629d18cf83506174f0d1/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopicPropertiesFactory.java#L114-L120

Maybe it should also look for the new configuration, and if it's `SINGLE_TOPIC` respect that regardless of the `FIXED` strategy?

This way we can deprecate the old enum and methods and users can rely only on the new one.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopic.java
```diff
@@ -68,8 +68,8 @@ public boolean isNoOpsTopic() {
 		return Type.NO_OPS.equals(this.properties.type);
 	}
```
tomazfernandes, line 70: I think this is a breaking change - we can keep both methods and deprecate the old one.

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -748,6 +721,98 @@ public RetryTopicConfiguration myRetryTopic(KafkaTemplate<String, MyPojo> templa
...
+}
+----
+====
+
+NOTE: The default behavior is creating separate retry topics for each attempt, appended with their index value: retry-0, retry-1, ...
+
+
+[[single-topic-maxinterval-delay]]
+===== Single Topic for maxInterval Exponential Delay 
+
```
tomazfernandes, line 761: I think you know this, but there's a [one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line) rule for docs, this line has 2.

tomazfernandes (commented):
Cool feature @jgslima, thanks!!

Overall this looks great. I've left a few comments, and let's see what @garyrussell has to say.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopic.java
```diff
@@ -68,8 +68,8 @@ public boolean isNoOpsTopic() {
...
 
-	public boolean isSingleTopicRetry() {
-		return Type.SINGLE_TOPIC_RETRY.equals(this.properties.type);
+	public boolean isReusableRetryTopic() {
```
tomazfernandes, line 71: I think this is a breaking change - we can keep both methods and deprecate the old one.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/SameIntervalTopicReuseStrategy.java
jgslima: > I'm thinking we could deprecate the FixedDelayStrategy at once in this PR and use the new option. WDYT?

@tomazfernandes, you are the bosses here. Feel free to direct me.

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
jgslima: > I think you know this, but there's a [one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line) rule for docs, this line has 2.

I did not know that (first time with asciidoc). Maybe due to the file having been changed since then, the highlighted section does not show a multi-sentence line, but I will search here.

tomazfernandes: > I did not know that (first time with asciidoc).

I thought you did because everything else was ok :)

>Maybe due to the file having been changed since then, the highlighted section does not show a multi-sentence line, but I will search here.

I probably highlighted the wrong line, sorry. It's the line right next to that one ("If you're using exponential backoff policy")

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/SameIntervalTopicReuseStrategy.java
tomazfernandes: Now that I read again, I think I mostly said what you'd already said 😄

Only adding that in order to deprecate the `FixedDelayStrategy`, which I believe we should do, we need to make the `SameIntervalTopicReuseStrategy` also work for fixed delays.

jgslima: I will wait for feedbacks from @garyrussell before actually changing things.

garyrussell: I plan to look at this tomorrow; but since @tomazfernandes is the original author of this feature, I trust his words.

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -683,7 +650,7 @@ IMPORTANT: Note that the blocking retries behavior is allowlist - you add the ex
 
 IMPORTANT: The non-blocking exception classification behavior also depends on the specific topic's configuration.
 
-==== Topic Naming
+==== Topic Amount and Naming
```
garyrussell, line 653: I don't think we need to change the section title.

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -693,6 +660,11 @@ Examples:
 
 "my-other-topic" -> "my-topic-myRetrySuffix-1000", "my-topic-myRetrySuffix-2000", ..., "my-topic-myDltSuffix".
 
+NOTE: The default behavior is creating separate retry topics for each attempt, appended with their index value: retry-0, retry-1, ..., retry-n. Therefore, by default the amount of retry topics is the configured `maxAttempts` minus 1.
```
garyrussell, line 663:
```suggestion
NOTE: The default behavior is to create separate retry topics for each attempt, appended with an index value: retry-0, retry-1, ..., retry-n. 
Therefore, by default the number of retry topics is the configured `maxAttempts` minus 1.
```

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
garyrussell: We use one sentence per line to isolate future changes to a single sentence.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopicPropertiesFactory.java
```diff
@@ -119,23 +125,48 @@ private boolean isSingleTopicStrategy() {
...
+		int retryTopicsAmount = retryTopicsAmount();
+
 		return IntStream.rangeClosed(0, isNoDltStrategy()
-							? this.maxAttempts - 1
-							: this.maxAttempts)
-				.mapToObj(this::createRetryOrDltTopicSuffixes)
+				? retryTopicsAmount
+				: retryTopicsAmount + 1)
+				.mapToObj(this::createTopicProperties)
 				.collect(Collectors.toList());
```
garyrussell, line 140: Please restore the indentation.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/SameIntervalTopicReuseStrategy.java
garyrussell: I agree with deprecating the FDS, with the aim to remove it in 3.1.

garyrussell (changes requested):
Good stuff; thanks!!

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/SameIntervalTopicReuseStrategy.java
tomazfernandes: We should probably also add a disclaimer in the javadocs / documentation that `SameIntervalTopicReuseStrategy` doesn't work (at least yet) when the same interval is in the middle of the backoff values. Though the OOTB BackOff policies doesn't allow that, we've seen users create custom policies that would allow such cases.

jgslima:
Thank you for the reviews, I will be able to work on the changes on the weekend.

jgslima:
Besides marking `FixedDelayStrategy` and its related methods as deprecated, I tried to clean some classes, to make them work with just `SameIntervalTopicReuseStrategy` as much as possible.

So, I tried to change `RetryTopicConfigurationBuilder` and `DestinationTopicPropertiesFactory`, removing their `FixedDelayStrategy` attributes.

But after struggling for a while, it seems to me that it is best to leave the code as it is, and to clean it only when the actual remove is done in 3.1.

One of the issues is with `RetryTopicConfigurationBuilder`. We might make the methods `useSingleTopicForFixedDelays()` to define the value of the new attribute `SameIntervalTopicReuseStrategy`. However, the builder would have 2 pairs of methods that would act on the same attribute, and therefore, the `SameIntervalTopicReuseStrategy` attribute would end up having the value of the method called last (which would override the previous call), which might have misleading results.

Since today nobody uses the new methods dedicated to `SameIntervalTopicReuseStrategy` (like the new one `useSingleTopicForSameIntervals()`), we could make the builder to throw an error if both kinds of methods were called, requiring the user to call just one of them.

But the problem is with `RetryableTopicAnnotationProcessor`. Since both annotation attributes have default values, and are never `null`, `RetryableTopicAnnotationProcessor` needs to call both kinds of methods in the builder. I thought about having `RetryableTopicAnnotationProcessor` to check if both annotation attributes values are "compatible", throwing an error if not, so that only the method `builder.sameIntervalTopicReuseStrategy()` would be called. But this would break existing code that just declares `fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC`, since this would be incompatible with `SameIntervalTopicReuseStrategy.MULTIPLE_TOPICS`, which must be the default value.

It might be possible to work around this, by making the attribute `sameIntervalTopicReuseStrategy()` in the annotation with default as `null`, but I think this is not good for the annotation itself.

So for now, only the class `DefaultDestinationTopicResolver` would already be "cleaned".

jgslima:
In order to deprecate `FixedDelayStrategy`, I struggled a lot and was not able to find a proper approach for the `@RetryableTopic` annotation.

Its new `SameIntervalTopicReuseStrategy` attribute must have a default value (not just to simplify the annotation use, but also, to avoid breaking all existing declarations). 

For the 2 cases below, I think we can assume SINGLE, because, we can tell that the user actually declared something for `sameIntervalTopicReuseStrategy` (since it is different from its default value of `MULTIPLE`):

```
@RetryableTopic(
     sameIntervalTopicReuseStrategy = SameIntervalTopicReuseStrategy.SINGLE_TOPIC
)
```

```
@RetryableTopic(
     sameIntervalTopicReuseStrategy = SameIntervalTopicReuseStrategy.SINGLE_TOPIC,
     fixedDelayTopicStrategy = FixedDelayStrategy.MULTIPLE_TOPICS
)
```

But the problem occurs **in these 2 cases**:

```
// below, SameIntervalTopicReuseStrategy.MULTIPLE_TOPICS is assumed since it is the default value.
@RetryableTopic(
     fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC
)
```

```
@RetryableTopic(
     fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC
     sameIntervalTopicReuseStrategy = SameIntervalTopicReuseStrategy.MULTIPLE_TOPICS
)
```
For them,  we cannot actually give some form of _precedence_ for the configuration of `SameIntervalTopicReuseStrategy`, because we cannot tell whether the user actually configured it (I do not know means in the Java reflection API to tell whether an annotation attribute was actually declared).
If we could tell whether the user actually declared the attribute, we could assume SINGLE in the first case, and assume MULTIPLE in the second case (or even throw an error, requiring the user to configure just one of them).

We cannot assume MULTIPLE, because this would change the behaviour of existing codes like the first case.

We cannot also give precedence for the configuration of `fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC` because in the second case, it would not make sense to give precedence to a deprecated property.

I am really calling it... need help here.

jgslima:
In order to deprecate `FixedDelayStrategy`, new code that calls only the new methods in `RetryTopicConfigurationBuilder` should work fine. So the new methods like `sameIntervalTopicReuseStrategy()` should work also when using a single fixed delay topic.

However, by keeping the existing methods (like `useSingleTopicForFixedDelays()`) the builder would have 2 pairs of methods that would act on the same configuration, and therefore, the configuration would end up having the value of the method called last (which would override the previous call), which might have misleading results.

Since today nobody uses the new methods dedicated to `SameIntervalTopicReuseStrategy` (like the new one `useSingleTopicForSameIntervals()`), we could make the builder to throw an error if both kinds of methods were called, requiring the user to call just one of them.


But the problem is with `RetryableTopicAnnotationProcessor`. The new `SameIntervalTopicReuseStrategy` attribute of `@RetryableTopic` must have a default value (not just to simplify the annotation use, but also, to avoid breaking all existing declarations). Since both annotation attributes have default values, and are never `null`, for `RetryableTopicAnnotationProcessor` it is hard to know how to interact with the builder.

For the 2 cases below, I think we could give precedence for the `sameIntervalTopicReuseStrategy` and assume SINGLE, because, we can tell that the user actually declared something for `sameIntervalTopicReuseStrategy` (since it is different from its default value of `MULTIPLE`):

```
@RetryableTopic(
     sameIntervalTopicReuseStrategy = SameIntervalTopicReuseStrategy.SINGLE_TOPIC
)
```

```
@RetryableTopic(
     sameIntervalTopicReuseStrategy = SameIntervalTopicReuseStrategy.SINGLE_TOPIC,
     fixedDelayTopicStrategy = FixedDelayStrategy.MULTIPLE_TOPICS
)
```

But the problem occurs **in these 2 cases**:

```
// below, SameIntervalTopicReuseStrategy.MULTIPLE_TOPICS is assumed since it is the default value.
@RetryableTopic(
     fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC
)
```

```
@RetryableTopic(
     fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC
     sameIntervalTopicReuseStrategy = SameIntervalTopicReuseStrategy.MULTIPLE_TOPICS
)
```

For them,  we cannot actually give some form of _precedence_ for the configuration of `SameIntervalTopicReuseStrategy`, because we cannot tell whether the user actually configured it (I do not know means in the Java reflection API to tell whether an annotation attribute was actually declared).
If we could tell whether the user actually declared the attribute, we could assume SINGLE in the first case, and assume MULTIPLE in the second case (or even throw an error, requiring the user to configure just one of them).

We cannot assume MULTIPLE, because this would change the behaviour of existing codes like the first case.

We cannot also give precedence for the configuration of `fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC` because in the second case, it would not make sense to give precedence to a deprecated property.

So I am not being capable of thinking about a way to make `RetryableTopicAnnotationProcessor` to work fine with the new behaviour of `RetryTopicConfigurationBuilder`.  :(

jgslima:
I thought about having `RetryableTopicAnnotationProcessor` to check if both annotation attributes values are "compatible", throwing an error if not, so that only the method ` builder.sameIntervalTopicReuseStrategy()` would be called. But this would break existing code that just declares `fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC`, since this would be incompatible with `SameIntervalTopicReuseStrategy.MULTIPLE_TOPICS`, which must be the default value.

tomazfernandes:
Hi @jgslima, thanks for looking into this.

I might be missing something, but what's on my mind seems a bit simpler:

- Both enums default to `MULTIPLE_TOPICS`
- For the new logic, only `SameIntervalTopicReuseStrategy` is looked up, so no problems there.
- For existing logic, we can look at both enum values - if any of them is `SINGLE_TOPIC`, single topic it is.

My reasoning for the last one is:
- Existing logic with `SINGLE_TOPIC` for `fixedDelayTopicStrategy` would work as expected;
- Existing logic with default values / `MULTIPLE_TOPICS` would work as expected;
- New user logic using `SameIntervalTopicReuseStrategy#SINGLE_TOPIC` for fixed delays would work as expected
- New user logic using `SameIntervalTopicReuseStrategy#MULTIPLE_TOPICS` for fixed delays would work as expected.

What am I missing?

Thanks.

jgslima:
Tonight I may be able to take a deep dive into your thoughts. At first, one case I can think of is existing code like where the developer declared `FixedDelayStrategy.SINGLE_TOPIC`, but is in fact using an exponential backoff:

```
@RetryableTopic(
  backoff = xxx,  // here, a backoff that actually results in exponential.
  fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC
)
```

For instance, `@Backoff` may be configured with a `delayExpression` that sometimes may result in exponential, other times in fixed. Or it is always exponential, and `fixedDelayTopicStrategy` is there without need.

These cases, currently, result in multiple topics for the maxDelay. By making `RetryableTopicAnnotationProcessor` to assume single, would change the behaviour.

Later I will try to think about all combinations to see if we can make `RetryableTopicAnnotationProcessor` to assume the strategy based also on the actual backoff type (it can inspect the actual instance created for the backoff policy).

tomazfernandes:
> These cases, currently, result in multiple topics for the maxDelay. By making RetryableTopicAnnotationProcessor to assume single, would change the behaviour.

I don't think `RetryableTopicAnnotationProcessor` should really assume anything - it's role is to simply create the `RetryTopicConfiguration` with what's on the annotation. So, if it's `FixedDelayStrategy.SINGLE_TOPIC`, we just forward it. Same with the new configuration. Both configurations will need to cohexist in the `RetryTopicConfiguration` class, one being deprecated.

When we get to the point where we check for the strategy, we just check if all backoff values are the same, so if it's an exponential backoff we should simply ignore that configuration.

The `fixedDelayTopicStrategy` shouldn't have any interference in how the exponential backoff behaves, with old or new logic.

Hope this helps :)

Thanks.

jgslima:
Ok then.

I thought that, if a method is deprecated, ideally the framework should provide a recommended, non-deprecated alternative.

That's why I thought that the new methods in the builder should work to configure for fixed delay as well. Which in turn means a single configuration inside.

Anyway, I will proceed with the simple approach for the deprecation. Existing code that refers to FDS will have to be changed in 3.1.

jgslima:
I believe I have made all the requested changes. Please review. Thanks.

jgslima:
Hello guys. Any news on this? Thank you.

garyrussell:
Sorry for the delay; it's on my TODO list, will try to get to it soon.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/FixedDelayStrategy.java
```diff
@@ -22,8 +22,10 @@
  *
  * @author Tomaz Fernandes
  * @since 2.7
+ * @deprecated in a future release, will be replaced by {@link SameIntervalTopicReuseStrategy}.
  *
  */
+@Deprecated(forRemoval = true) // in 3.1
```
tomazfernandes, line 28: Hi João! I'm not sure I understand - as far as I am aware we won't remove anything in 3.1, as we need to keep backwards compatibility until the next major or at least a few minors. Care to elaborate on the comment? Thanks!

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/FixedDelayStrategy.java
```diff
@@ -22,8 +22,10 @@
  *
  * @author Tomaz Fernandes
  * @since 2.7
+ * @deprecated in a future release, will be replaced by {@link SameIntervalTopicReuseStrategy}.
```
tomazfernandes, line 25: We can simply say: `deprecated in favor of {@link SameIntervalTopicReuseStrategy}`

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/FixedDelayStrategy.java
jgslima: I followed this:
https://github.com/spring-projects/spring-kafka/pull/2497#discussion_r1093411586

Anyway, I can remove the removal mentions, just want to hear from @garyrussell again.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/FixedDelayStrategy.java
jgslima: Ok I will proceed with the changes

jgslima: > We can simply say: `deprecated in favor of {@link SameIntervalTopicReuseStrategy}`

Ok.

I do not think we can say that.
Previously, I had said that I was having hard time trying to condense (internally at `RetryTopicConfigurationBuilder` and `DestinationTopicPropertiesFactory`) both configurations, in order for these components to work only with the `SameIntervalTopicReuseStrategy` configuration. So that, when a method like `RetryTopicConfigurationBuilder.useSingleTopicForFixedDelays()` was called, internally this would actually configure the `SameIntervalTopicReuseStrategy`.

But then, we aggreed on keeping both configurations. Just marking the old configurations as deprecated, aiming to remove them in the future (when `DestinationTopicPropertiesFactory` should be changed).

So we cannot say to make use of `SameIntervalTopicReuseStrategy`, because right now, for Fixed delay topics, `FixedDelayStrategy` still must be used.

jgslima: And to recap... when trying to condense the configs, thus allowing new code to configure single fixed delay topic using the new methods like `useSingleTopicForSameIntervals()`, the real problem is in `RetryableTopicAnnotationProcessor` and how it would interact correctly with the builder.

Like I explained [here](https://github.com/spring-projects/spring-kafka/pull/2497#issuecomment-1418201556).

tomazfernandes: Sorry, I've read your comment before, I read it again now, and I'm having a hard time understanding why this is so complicated.

In my head, we can just leave both methods / parameters in all classes you mentioned, deprecating the old ones, and solve any ambiguity in the `DestinationTopicPropertiesFactory` class, by the rules I mentioned in the other comment I made earlier:

> Both enums default to MULTIPLE_TOPICS
For the new logic, only SameIntervalTopicReuseStrategy is looked up, so no problems there.
For existing logic, we can look at both enum values - if any of them is SINGLE_TOPIC, single topic it is.

Can you please help me understand why wouldn't this work? Thanks.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/FixedDelayStrategy.java
tomazfernandes: Oh, I didn't remember this comment. Perhaps @garyrussell can enlighten us on this?

garyrussell: My current plan is to look at this tomorrow, unless something more urgent comes in.

I am not opposed to removing stuff in the next minor release, it really depends on how critical or invasive it is.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/FixedDelayStrategy.java
tomazfernandes: Or being more specific:

- For fixed delays, we can look at both the old and new enums - whichever has `SINGLE_TOPIC` has precedence.
- For exponential (non-fixed) delays we only look at the new enum.

In your example:
> At first, one case I can think of is existing code like where the developer declared FixedDelayStrategy.SINGLE_TOPIC, but is in fact using an exponential backoff

For this case, we'd assume `SINGLE_TOPIC` for fixed delays, just as it is now. For exponential delays the new enum would be `MULTIPLE_TOPICS` so wouldn't change functionality. If the user sets the new enum for `SINGLE_TOPIC`, the new enum should take precedence and impact both the fixed delay and exponential delay variants.

Also, we don't really need to take care of all combinations. E.g. if the user explicitly sets the old enum to `MULTIPLE` and the new one to `SINGLE`, that's a misconfiguration (using both a deprecated and the new configuration at the same time) and we can just assume `SINGLE`. At most we'd need to document this in the javadocs and / or documentation.

Does this make sense? Thanks.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopic.java
```diff
@@ -280,6 +293,6 @@ public boolean isMainEndpoint() {
 	}
 
 	enum Type {
-		MAIN, RETRY, SINGLE_TOPIC_RETRY, DLT, NO_OPS
+		MAIN, RETRY, REUSABLE_RETRY_TOPIC, DLT, NO_OPS
```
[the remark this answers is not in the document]
jgslima: > We can't do this in a patch release; need to keep (and deprecate) the old value and make it an alias of the new one.

I did not push this now. This was in the first commit at Dec 4, 2022.

It is one of the basis of the logic reuse for both kinds of backoff. `DefaultDestinationTopicResolver' will not have a single concept anymore.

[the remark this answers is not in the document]
garyrussell: ?? I only see 5 commits total, and none of the subsequent patches reverted it - this is still the state of the code in the PR.

jgslima:
Pushed requested changes

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/DestinationTopicPropertiesFactoryTests.java
```diff
@@ -238,6 +238,48 @@ void shouldCreateOneRetryPropertyForFixedBackoffWithSingleTopicStrategy() {
...
+		assertThat(propertiesList.size() == 3).isTrue();
+
+		DestinationTopic mainDestinationTopic = destinationTopicList.get(0);
+		assertThat(mainDestinationTopic.isMainTopic()).isTrue();
+
+		DestinationTopic.Properties firstRetryProperties = propertiesList.get(1);
+		assertThat(firstRetryProperties.suffix()).isEqualTo(retryTopicSuffix);
+		DestinationTopic retryDestinationTopic = destinationTopicList.get(1);
+		assertThat(retryDestinationTopic.isSingleTopicRetry()).isTrue();
+		assertThat(retryDestinationTopic.isReusableRetryTopic()).isFalse();
```
jgslima, line 281: Here, it is a matter of convention. We might consider every `singleTopicRetry` to also be a `reusableRetryTopic`. Right now, these methods are just based on the internal `Type` attribute.

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -683,7 +651,7 @@ IMPORTANT: Note that the blocking retries behavior is allowlist - you add the ex
 
 IMPORTANT: The non-blocking exception classification behavior also depends on the specific topic's configuration.
 
-==== Topic Naming
+==== Topic Amount
```
garyrussell, line 654:
```suggestion
==== Topic Naming
```

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -693,6 +661,12 @@ Examples:
 
 "my-other-topic" -> "my-topic-myRetrySuffix-1000", "my-topic-myRetrySuffix-2000", ..., "my-topic-myDltSuffix".
 
+NOTE: The default behavior is to create separate retry topics for each attempt, appended with an index value: retry-0, retry-1, ..., retry-n.
+Therefore, by default the amount of retry topics is the configured `maxAttempts` minus 1.
```
garyrussell, line 665:
```suggestion
Therefore, by default the number of retry topics is the configured `maxAttempts` minus 1.
```

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -754,9 +729,106 @@ public RetryTopicConfiguration myRetryTopic(KafkaTemplate<String, MyPojo> templa
...
+[[single-topic-maxinterval-delay]]
+===== Single Topic for maxInterval Exponential Delay 
+
+If you're using exponential backoff policy (`ExponentialBackOffPolicy`), you can use a single retry topic to accomplish the non-blocking retries of the attempts whose delays are the configured `maxInterval`.
+
+This "final" retry topic will be suffixed with the provided or default suffix, and will have either the index or the `maxInterval` value appended.
+
+NOTE: By opting to use a single topic for the retries with the `maxInterval` delay, it may become more viable to configure an exponential retry policy that keeps retrying for a long time, because in this approach you do not need a large amount of topics.
+
+The default behavior is to work with an amount of retry topics equal to the configured `maxAttempts` minus 1, and when using exponential backoff, the retry topics are suffixed with the delay values, with the last retry topics (corresponding to the `maxInterval` delay) being suffixed with an additional index.
```
garyrussell, line 778:
```suggestion
The default behavior is to work with the number of retry topics equal to the configured `maxAttempts` minus 1 and, when using exponential backoff, the retry topics are suffixed with the delay values, with the last retry topic (corresponding to the `maxInterval` delay) being suffixed with an additional index.
```

spring-kafka-docs/src/main/asciidoc/retrytopic.adoc
```diff
@@ -754,9 +729,106 @@ public RetryTopicConfiguration myRetryTopic(KafkaTemplate<String, MyPojo> templa
...
+* -retry-16000-224
+
+When using the strategy that reuses the retry topic for the same intervals, in the same configuration above the needed retry topics would be:
+
+* -retry-1000
+* -retry-2000
+* -retry-4000
+* -retry-8000
+* -retry-16000
+
```
garyrussell, line 799:
```suggestion

This will be the default in a future release.
```

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
```diff
@@ -71,6 +71,7 @@ public class RetryTopicConfigurationBuilder {
 	@Nullable
 	private BinaryExceptionClassifierBuilder classifierBuilder;
 
+	@SuppressWarnings("deprecation")
 	private FixedDelayStrategy fixedDelayStrategy = FixedDelayStrategy.MULTIPLE_TOPICS;
```
garyrussell, line 75: This needs to be fully qualified (and remove the `import`). Otherwise the compiler will emit a deprecation warning for the import.

garyrussell (changes requested):
LGTM; just some minor polishing needed.

One other thing I'd like to verify is that the enum changes are binary compatible - can users that explicitly use the enum drop in the new jar and use it without recompiling. We always avoid API changes at the compiler level, but we strive to maintain binary compatibility too.

Most likely, we'll have to move the new value to the end of the enum but I am not sure if even that will work.

garyrussell:
Correction `DestinationTopic.Type` is not a public API (not visible).

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/SameIntervalTopicReuseStrategy.java
```diff
@@ -0,0 +1,45 @@
...
+ * Strategy for topic reuse when multiple, sequential retries have the same backoff
+ * interval.
+ *
+ * <p>
+ * It can be used only when the retries that have the same interval are located
+ * in the end of the retry chain (it cannot be used for retries with the same
+ * interval in the middle of the retry chain).
+ *
+ * @author João Lima
+ * @since 3.1
```
garyrussell, line 30:
```suggestion
 * @since 3.0.4
```

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopic.java
```diff
@@ -177,6 +178,17 @@
...
 	TopicSuffixingStrategy topicSuffixingStrategy() default TopicSuffixingStrategy.SUFFIX_WITH_DELAY_VALUE;
 
+
+	/**
+	 * Topic reuse strategy for sequential attempts made with a same backoff interval.
+	 *
+	 * <p>Note: for fixed backoffs, when this is configured as
+	 * {@link SameIntervalTopicReuseStrategy#SINGLE_TOPIC}, it has precedence over
+	 * the configuration in {@link #fixedDelayTopicStrategy()}.
+	 * @return the strategy.
```
garyrussell, line 188:
```suggestion
	 * @return the strategy.
	 * @since 3.0.4
```

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopic.java
```diff
@@ -284,6 +297,14 @@ public boolean isMainEndpoint() {
...
-		MAIN, RETRY, SINGLE_TOPIC_RETRY, DLT, NO_OPS
+		MAIN, RETRY,
+		/**
+		 * A single retry topic for all retries.
+		 *
+		 * @deprecated Use {@code REUSABLE_RETRY_TOPIC} instead.
+		 */
+		@Deprecated
+		SINGLE_TOPIC_RETRY,
+		REUSABLE_RETRY_TOPIC, DLT, NO_OPS
```
garyrussell, line 308: Please make them all one per line and add `@since 3.0.4` to the new one.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
```diff
@@ -241,6 +244,39 @@ public RetryTopicConfigurationBuilder setTopicSuffixingStrategy(TopicSuffixingSt
...
 
+	/**
+	 * Configure the {@link SameIntervalTopicReuseStrategy}.
+	 *
+	 * <p>Note: for fixed backoffs, when this is configured as
+	 * {@link SameIntervalTopicReuseStrategy#SINGLE_TOPIC}, it has precedence over
+	 * the configuration done through
+	 * {@link #useSingleTopicForFixedDelays(FixedDelayStrategy)}.
+	 * @param sameIntervalTopicReuseStrategy the strategy.
+	 * @return the builder.
```
garyrussell, line 255:
```suggestion
	 * @return the builder.
	 * since 3.0.4
```

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
```diff
@@ -241,6 +244,39 @@ public RetryTopicConfigurationBuilder setTopicSuffixingStrategy(TopicSuffixingSt
...
+	/**
+	 * Configure the use of a single retry topic
+	 * for the attempts that have the same backoff interval
+	 * (as long as these attempts are in the middle of the chain).
+	 *
+	 * Currently used only for the last retries of exponential backoff,
+	 * and in a future release this will dictate whether to use
+	 * a single retry topic for fixed backoff.
+	 *
+	 * @return the builder.
```
garyrussell, line 271:
```suggestion
	 * @return the builder.
	 * @since 3.0.4
```

garyrussell (changes requested):
I think we can get this into tomorrow's 3.0.4 release (late day US EST).

I tested it with a Boot app I wrote when you submitted the original discussion #2484 and it works nicely.

Thanks for all this work; sorry it took so long to get through the review process.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
jgslima: Well, there is no import in the top of the class. It is in the same package.

garyrussell: Doh - sorry.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
```diff
@@ -241,6 +244,41 @@ public RetryTopicConfigurationBuilder setTopicSuffixingStrategy(TopicSuffixingSt
...
+	}
+
+	/**
+	 * Configure the use of a single retry topic
+	 * for the attempts that have the same backoff interval
+	 * (as long as these attempts are in the end of the chain).
+	 *
+	 * Currently used only for the last retries of exponential backoff,
+	 * and in a future release this will dictate whether to use
+	 * a single retry topic for fixed backoff.
```
garyrussell, line 270: @jgslima Looks like we missed cleaning this up - I just tested it and the `SameIntervalTopicReuseStrategy` works fine for a fixed delay too.

It's true that the other strategy will be removed next release, but they can switch to the new one today (or am I missing something?).

jgslima: You are right. :(
One of the last changes was indeed to allow the new strategy to be used for fixed delays.
My mistake.
Do you want me to open a PR?

garyrussell: Thanks for confirming; no; I'll just push a fix.

Thanks again for implementing this.

Issue #2515, opened 2022-12-21
Confusing Documentation w.r.t. EOS Mode V2 and Zombie Fencing via `transactional.id` in Spring Kafka 3.0.0


LukasHeimann:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.0.0

**Describe the bug**

The documentation for Transactions, in it's Overview subsection, mentions how the `transactional.id` is derived from the `transactionIdPrefix`: https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka-docs/src/main/asciidoc/kafka.adoc#overview-1

Specifically, it describes why the naming scheme for transactions started by a listener container is different from the default of `transactionIdPrefix<n>`:

> This is to properly support fencing zombies, [as described here](https://www.confluent.io/blog/transactions-apache-kafka/).

This _used_ to be valid information, as with EOS Mode V1, there really were different semantics at play. Especially when having multiple instances of an application consuming messages, this information was in fact important to know about, as including the instance number in the `transactionIdPrefix` needed to be done for the normal case, but not for the transactions started by a listener container. Compare https://github.com/spring-projects/spring-kafka/issues/1503 and https://stackoverflow.com/questions/56971389/what-should-be-the-transactionidprefix-for-multiple-spring-boot-consumer-produce

With V2, however, this is not necessary anymore, as zombie fencing is handled differently. The _old_ documentation of Spring Kafka 2.8.x actually mentions this, because it supports both modes: https://github.com/spring-projects/spring-kafka/blob/2.8.x/spring-kafka-docs/src/main/asciidoc/kafka.adoc#transaction-id-prefix

> This problem (different rules for transactional.id) has been eliminated when EOSMode.BETA is being used (with broker versions >= 2.5); see [Exactly Once Semantics](https://docs.spring.io/spring-kafka/docs/2.8.11/reference/html/#exactly-once).

The current documentation does _not_ mention this, creating lots of confusion on what should be done.

**To Reproduce**

n/a (read the documentation and compare with previous versions)

**Expected behavior**

I think, the documentation could be clearer here. In the Overview section, it could mention that `transactionIdPrefix` is supposed to be unique per instance of the application, because of the way `transactional.id` is built (and it makes sense to write that pattern down). The section about fencing zombies should be removed and perhaps be replaced by a sentence mentioning that for transactions started by a listener container, EOS Mode V2 does not rely on the `transactional.id` to achieve transactional handling.

[...]

PR #2524, opened 2023-01-02
Remove outdated information for transactional.id

LukasHeimann:
Compare https://github.com/spring-projects/spring-kafka/issues/2515

With only EOSMode v2 supported, the alternative way of `transactional.id` creation is not used anymore and not needed anymore. This PR removes the outdated (and confusing) extra documentation in the overview section for transactions.

### DIFF
--- a/spring-kafka-docs/src/main/asciidoc/kafka.adoc
+++ b/spring-kafka-docs/src/main/asciidoc/kafka.adoc
@@ -3444,11 +3444,7 @@ Spring for Apache Kafka adds support in the following ways:
 Transactions are enabled by providing the `DefaultKafkaProducerFactory` with a `transactionIdPrefix`.
 In that case, instead of managing a single shared `Producer`, the factory maintains a cache of transactional producers.
 When the user calls `close()` on a producer, it is returned to the cache for reuse instead of actually being closed.
-The `transactional.id` property of each producer is `transactionIdPrefix` + `n`, where `n` starts with `0` and is incremented for each new producer, unless the transaction is started by a listener container with a record-based listener.
-In that case, the `transactional.id` is `<transactionIdPrefix>.<group.id>.<topic>.<partition>`.
-This is to properly support fencing zombies, https://www.confluent.io/blog/transactions-apache-kafka/[as described here].
-This new behavior was added in versions 1.3.7, 2.0.6, 2.1.10, and 2.2.0.
-If you wish to revert to the previous behavior, you can set the `producerPerConsumerPartition` property on the `DefaultKafkaProducerFactory` to `false`.
+The `transactional.id` property of each producer is `transactionIdPrefix` + `n`, where `n` starts with `0` and is incremented for each new producer.
 
 Also see <<exactly-once>>.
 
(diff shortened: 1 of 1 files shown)


spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -3444,11 +3444,7 @@ Spring for Apache Kafka adds support in the following ways:
 Transactions are enabled by providing the `DefaultKafkaProducerFactory` with a `transactionIdPrefix`.
 In that case, instead of managing a single shared `Producer`, the factory maintains a cache of transactional producers.
 When the user calls `close()` on a producer, it is returned to the cache for reuse instead of actually being closed.
-The `transactional.id` property of each producer is `transactionIdPrefix` + `n`, where `n` starts with `0` and is incremented for each new producer, unless the transaction is started by a listener container with a record-based listener.
-In that case, the `transactional.id` is `<transactionIdPrefix>.<group.id>.<topic>.<partition>`.
-This is to properly support fencing zombies, https://www.confluent.io/blog/transactions-apache-kafka/[as described here].
-This new behavior was added in versions 1.3.7, 2.0.6, 2.1.10, and 2.2.0.
-If you wish to revert to the previous behavior, you can set the `producerPerConsumerPartition` property on the `DefaultKafkaProducerFactory` to `false`.
+The `transactional.id` property of each producer is `transactionIdPrefix` + `n`, where `n` starts with `0` and is incremented for each new producer.
```
LukasHeimann, line 3447:
```suggestion
The `transactional.id` property of each producer is `transactionIdPrefix` + `n`, where `n` starts with `0` and is incremented for each new producer.
In previous versions of Spring Kafka, the `transactional.id` was generated differently for transaction started by a listener container with a record-based listener to support fencing zombies, which is not necessary anymore with EOSMode V2 being the only option starting with 3.0.0.
For applications running with multiple instances, the `transactionIdPrefix` should be unique per instance.
```

LukasHeimann (commented):
The PR currently only includes the minimum necessary change to remove the outdated documentation.

However, I propose to also add two additional sentences of clarification that might be helpful for readers (at least I would have been happy to have them) -- if you feel like they should be included, please let me know or just accept the suggested change.

1. Information on how the transactional.id was generated differently in the past. This is useful, as there are many old posts around that discuss this old behavior, so just giving a heads up that you don't have to care is probably nice
2. Information on `transactionIdPrefix` having to be unique per instance of an application. This is mixed up in the discussion of the old way of generation as well: for consumer-initiated transactions it used to be that the `transactionIdPrefix` should be the same for multiple instances, for producer-initiated it must be different. Hence, a small sentence of explanation is probably helpful as well.

spring-kafka-docs/src/main/asciidoc/kafka.adoc
```diff
@@ -3444,11 +3444,9 @@ Spring for Apache Kafka adds support in the following ways:
...
+In previous versions of Spring Kafka, the `transactional.id` was generated differently for transaction started by a listener container with a record-based listener to support fencing zombies, which is not necessary anymore with `EOSMode.V2` being the only option starting with 3.0.0.
+For applications running with multiple instances, the `transactionIdPrefix` should be unique per instance.
```
garyrussell, line 3449:
```suggestion
In previous versions of Spring for Apache Kafka, the `transactional.id` was generated differently for transactions started by a listener container with a record-based listener, to support fencing zombies, which is not necessary any more, with `EOSMode.V2` being the only option starting with 3.0.
For applications running with multiple instances, the `transactionIdPrefix` must be unique per instance.
```

garyrussell (changes requested):
Thanks! Just a couple of small comments

garyrussell (approved):

Issue #2525, opened 2023-01-03
Topic partition is not resumed after re-assignment for RetryTopic


dmyronenko:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

2.8.11

**Describe the bug**

From time to time we get a bug when doing rolling deployment with k8. What happens is that consumption from one of partitions is not being resumed after this partition is assigned to different consumer and then within retry-delay interval reassigned back to the initial consumer.

**To Reproduce**

Topic-A-retry-0 - the retry delay on this topic is 1 minute. In our case it has 5 partitions.

1. Topic-A-retry-0 partition 0 assigned to C1
2. Send message to this partition so that C1 pauses consumption for partition 0

[within 1 minute after sending message]

3. Simulate rolling update by running one more consumer C2. Topic-A-retry-0 partition 0 has to be assigned to C2
4. Bring down C2. Topic-A-retry-0 partition 0 assigned back to C1
5. After this the partition 0 on C1 will be put on pause but will never resume consumption.

[within 1 minute after sending message]

**Expected behavior**

Consuming should be resumed for a given partition after retry delay

**Notes**

I checked code a bit and what happens internally is that:
1. when partition is put on pause in PartitionPausingBackoffManager it calls 	 
```
AbstractMessageListenerContainer
	@Override
	public void pausePartition(TopicPartition topicPartition) {
		this.pauseRequestedPartitions.add(topicPartition);
	}
```
and creates a backoff Context that will resume partition(remove this item from list) based on processing of ListenerContainerPartitionIdleEvent.

One thing to note here is that during processing of the onPartitionsRevoked/Assigned the pauseRequestedPartitions are not cleaned up. But I guess it was a conscious choice because in onPartitionsAssigned these pauseRequestedPartitions are used to pause consumers.

2. What happens on Step 4-5 when partition 0 is assigned back to C1 is that this partition remains in pauseRequestedPartitions. And in onPartitionsAssigned it is being paused
```
KafkaMessageListenerContainer
				Collection<TopicPartition> toRepause = new LinkedList<>();
				partitions.forEach(tp -> {
					if (isPartitionPauseRequested(tp)) {
						toRepause.add(tp);
					}
				});
				if (!ListenerConsumer.this.consumerPaused && toRepause.size() > 0) {
					ListenerConsumer.this.consumer.pause(toRepause);
					ListenerConsumer.this.logger.debug(() -> "Paused consumption from: " + toRepause);
[...]

PR #2531, opened 2023-01-04
GH-2525: Fix Paused Partition Resume (Rebalance)

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2525

If a paused partition is revoked and re-assigned during a retry topic pause delay, the partition is not resumed because it no longer exists in the `pausedPartitions` field.

When re-pausing a paused partition, it must be re-added to the paused partitions so that it will not be filtered out of the resume logic.

**cherry-pick to 2.9.x**


artembilan:
... and back-ported to `2.9.x` as https://github.com/spring-projects/spring-kafka/commit/edf2b37f6b02165da7de01df198a30d7ffdf1660

PR #2555, opened 2023-01-23
GH-2554: Fix DEFAULT_SCHEDULER_WRAPPER_BEAN_NAME value

### LINKED ISSUE #2554: An application using default spring boot configuration fails at the startup when there are multiple @KafkaListener with @RetryableTopic

Since 2.9.0

**Describe the bug**
An application using default spring boot auto configuration beans fails at the startup when there are two kafka listeners with retry topic configuration. It throws the following exception:

```
Caused by: org.springframework.beans.factory.BeanNotOfRequiredTypeException: Bean named 'defaultRetryTopicKafkaTemplate' is expected to be of type 'org.springframework.kafka.core.KafkaOperations' but was actually of type 'org.springframework.kafka.retrytopic.RetryTopicSchedulerWrapper'
	at org.springframework.beans.factory.support.AbstractBeanFactory.adaptBeanInstance(AbstractBeanFactory.java:408)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:389)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:205)
	at org.springframework.kafka.annotation.RetryableTopicAnnotationProcessor.getKafkaTemplate(RetryableTopicAnnotationProcessor.java:216)
	at org.springframework.kafka.annotation.RetryableTopicAnnotationProcessor.processAnnotation(RetryableTopicAnnotationProcessor.java:150)
	at org.springframework.kafka.annotation.RetryTopicConfigurationProvider.findRetryConfigurationFor(RetryTopicConfigurationProvider.java:101)
	at org.springframework.kafka.annotation.KafkaListenerAnnotationBeanPostProcessor.processMainAndRetryListeners(KafkaListenerAnnotationBeanPostProcessor.java:511)
	at org.springframework.kafka.annotation.KafkaListenerAnnotationBeanPostProcessor.processKafkaListener(KafkaListenerAnnotationBeanPostProcessor.java:490)
	at org.springframework.kafka.annotation.KafkaListenerAnnotationBeanPostProcessor.postProcessAfterInitialization(KafkaListenerAnnotationBeanPostProcessor.java:391)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.applyBeanPostProcessorsAfterInitialization(AbstractAutowireCapableBeanFactory.java:435)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1754)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:599)
	... 92 more
```	

After some investigation, if I don't miss anything I found that the problem stems from this simple mistake: 

https://github.com/spring-projects/spring-kafka/blob/0d1727da5bbd10cfff78845dfb0a265fa8f5d5e1/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicBeanNames.java#L58-L65

`DEFAULT_KAFKA_TEMPLATE_BEAN_NAME` and `DEFAULT_SCHEDULER_WRAPPER_BEAN_NAME` are the same. That leads to replacing `KafkaTemplate` bean with `RetryTopicSchedulerWrapper` here in the first kafka listener initialization: 

https://github.com/spring-projects/spring-kafka/blob/0d1727da5bbd10cfff78845dfb0a265fa8f5d5e1/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java#L560-L562

This line throws an error on the second `RetryableTopic` annotation process since it retrieves `RetryTopicSchedulerWrapper` rather than `KafkaTemplate`:

https://github.com/spring-projects/spring-kafka/blob/0d1727da5bbd10cfff78845dfb0a265fa8f5d5e1/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java#L216-L217

We just need to change the value of `DEFAULT_SCHEDULER_WRAPPER_BEAN_NAME` variable for the fix.

As a workaround for the current version, an application can inject its TaskScheduler bean or specify `kafkaTemplate` in `@RetryableTopic`. 

**To Reproduce**

- Add multiple `@KafkaListener` with `@RetryableTopic`
- Use minimal configuration
-- Don't inject your own `TaskScheduler`
-- Don't specify `kafkaTemplate` in RetryableTopic

**Expected behavior**

Do not fail at the startup and configure the beans successfully.

**Sample**

A sample test case with spring boot and testcontainers:
https://github.com/cenkakin/spring-kafka-bug-demo/blob/main/src/test/java/com/github/cenk/springkafkabugdemo/SpringKafkaBugDemoTests.java

Also added another test case to my spring-kafka fork. 
https://github.com/cenkakin/spring-kafka/commit/56640edf4724b02ff5632c0a1c8397453308729d

**Fix**
The test passes in this branch:
https://github.com/spring-projects/spring-kafka/compare/main...cenkakin:spring-kafka:fix-multiple-listener-with-retry-bug

I can tidy up the code and create a PR if you like.

cenkakin:
Fixes #2554.


spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicMultipleListenerIntegrationTests.java
```diff
@@ -0,0 +1,249 @@
...
+
+/**
+ * @author Cenk Akin
+ * @since 3.0.3
+ */
+@SpringJUnitConfig
+@DirtiesContext
+@EmbeddedKafka(topics = {RetryTopicMultipleListenerIntegrationTests.FIRST_TOPIC,
+		RetryTopicMultipleListenerIntegrationTests.SECOND_TOPIC}, partitions = 1)
+public class RetryTopicMultipleListenerIntegrationTests {
```
cenkakin, line 65: I got mainly inspired by [this test](https://github.com/spring-projects/spring-kafka/blob/0d1727da5bbd10cfff78845dfb0a265fa8f5d5e1/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicSameContainerFactoryIntegrationTests.java#LL72C32-L72C32). Instead of adding a new test class, I can modify that class to make it cover this case or keep it as a separate class like in the PR. Or I can delete the test if you don't think it is worthwhile. It's up to you.

artembilan: Hi @cenkakin !

Not sure why do we need such a complex integration test just to verify that there is no conflict around `defaultRetryTopicKafkaTemplate` bean...
Right, I see that we need a second `@RetryableTopic` to make it fail after `gac.registerBean(RetryTopicBeanNames.DEFAULT_SCHEDULER_WRAPPER_BEAN_NAME,`, but probably we indeed can cover this use-case just adding extra config into that `RetryTopicSameContainerFactoryIntegrationTests` you mention before.

WDYT?

Thank you for contribution!

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicMultipleListenerIntegrationTests.java
cenkakin: Yup, it is what I can do. But we need to modify `Config` class in `RetryTopicSameContainerFactoryIntegrationTests` more to trigger the line that fails. 

As far as I understand from the test case intended to cover, it should be ok. Let me push another commit.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicSameContainerFactoryIntegrationTests.java
```diff
@@ -121,41 +118,52 @@ static class RetryableKafkaListener {
...
+
+		CountDownLatch countDownLatchDlt1 = new CountDownLatch(1);
+		CountDownLatch countDownLatchDlt2 = new CountDownLatch(1);
+		CountDownLatch customizerLatch = new CountDownLatch(9);
 	}
 
 	@EnableKafka
 	@Configuration
-	static class Config extends RetryTopicConfigurationSupport {
+	static class Config {
```
cenkakin, line 166: We shouldn't extend `RetryTopicConfigurationSupport` and shouldn't inject our`ThreadPoolTaskScheduler` own `https://github.com/spring-projects/spring-kafka/pull/2555/files#diff-80d8fe2567e6640c6d1edada33ec0bf783012606c2264cf4fc6b3aed4cca9742L242`

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicSameContainerFactoryIntegrationTests.java
```diff
@@ -186,10 +194,6 @@ public ConcurrentKafkaListenerContainerFactory<String, String> kafkaListenerCont
 			props.setPollTimeout(50L);
 			props.setIdlePartitionEventInterval(100L);
 			factory.setConsumerFactory(consumerFactory);
-			DefaultErrorHandler errorHandler = new DefaultErrorHandler(
-					(cr, ex) -> latchContainer.countDownLatch2.countDown(),
-					new FixedBackOff(0, 2));
-			factory.setCommonErrorHandler(errorHandler);
```
cenkakin, line 192: This is the part we no longer cover anymore.

I just pushed another commit that creates a new bean and keeps `BasicKafkaListener` instead of modifying it: 

https://github.com/spring-projects/spring-kafka/pull/2555/commits/01f0d6d7903ab2294cff0846c4bee86c6d93fe31

cenkakin:
Hey @artembilan! Thank you for the feedback! I pushed commits and left comments. Basically, the following three commits present different approaches:

- https://github.com/spring-projects/spring-kafka/pull/2555/commits/e4827ec64088f6e539cda8979ab63f8ff65e3c0e: Completely separate integration test
- https://github.com/spring-projects/spring-kafka/pull/2555/commits/e88fb4939f539eaa10a5e5097397634dad650b34: Modify `RetryTopicSameContainerFactoryIntegrationTests`, replace `basicKafkaListener` with another `retryableListener` 
- https://github.com/spring-projects/spring-kafka/pull/2555/commits/e88fb4939f539eaa10a5e5097397634dad650b34: Modify `RetryTopicSameContainerFactoryIntegrationTests`, add another `retryableListener` and keep `basicKafkaListener`

I would personally vote for the current status of the PR. Please let me know what you think, and I will update the PR accordingly.

artembilan:
Sorry, @garyrussell , some glitch has happened over here on GH, so I pressed `Rebase & merge` accidentally without squashing.
Next time will pull locally for more control 😄 

@cenkakin , thank you for contribution; looking forward for more!

Issue #2558, opened 2023-01-24
NPE in FailedBatchProcessor if Reported Index is Out of Bounds


garyrussell:
Can't log the record here because there is none...

https://github.com/spring-projects/spring-kafka/blob/9c2286875ccbd58aa644782c26e52384bb33ba08/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedBatchProcessor.java#L157-L160

PR #2559, opened 2023-01-25
GH-2558: Fix Possible NPE in FailedBatchProcessor

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2558

If a user throws a `BatchListenerExcecutionFailedException` with an invalid `index`, an NPE is logged instead of the true error.

**cherry-pick to 2.9.x**


artembilan:
... and cherry-picked as https://github.com/spring-projects/spring-kafka/commit/397da7d48328a5ab2c5861f9c879af7446ad2e2c

Issue #2566, opened 2023-02-02
Allow to use KafkaListenerContainerFactory with custom message converter for batch and record listeners


schmidti159:
**Expected Behavior**

The [spring-kafka documentation](https://docs.spring.io/spring-kafka/reference/html/#batch-listeners) mentions:

> Starting with version 2.8, you can override the factory’s batchListener property using the batch property on the @KafkaListener annotation. This, together with the changes to Container Error Handlers allows the same factory to be used for both record and batch listeners.

This should also work when a custom `MessageConverter` is defined on the factory. 

**Current Behavior**

For batch listeners, the message converter must be wrapped in a `BatchMessagingMessageConverter`. Otherwise, the message converter will not be used correctly and the wrong type will be supplied to the batch listener.

**Context**

We encapsulate spring-kafka into a library that preconfigures the factory as a bean in our setup. This library is then used by many services.

Another option would be to create a separate factory for batch listeners. With this, I am afraid that someone might use `@KafkaListener(batch="true")` as described in the official documentation but without supplying the correct library, which only works partly.

My current workaround looks like this, but I do not like it as it depends on how spring-kafka internally sets up its configuration, so it might break in future updates:
```java
factory.setContainerCustomizer(container -> {
    var messageListener = container.getContainerProperties().getMessageListener();
    if (messageListener instanceof FilteringBatchMessageListenerAdapter) {
        var meessageListenerDelegate =
                ((FilteringBatchMessageListenerAdapter<?, ?>) messageListener).getDelegate();
        if (meessageListenerDelegate instanceof BatchMessagingMessageListenerAdapter) {
            ((BatchMessagingMessageListenerAdapter<?, ?>) meessageListenerDelegate).setBatchMessageConverter(
                    new BatchMessagingMessageConverter(messageConverter));
        }
    }
});
```

See also [this StackOverflow question](https://stackoverflow.com/questions/75282411/how-to-reuse-kafkalistenercontainerfactory-with-a-custom-message-converter-for-b).

schmidti159:
Thanks! That was fast :rocket:

PR #2568, opened 2023-02-02
GH-2566: Both Converters in Container Factory

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2566

Now that both batch and record listners can be created by a single factory, it should be possible to configure both a `RecordMessageConverter` and `BatchMessageConverter`.

Previously, only a single `MessageConverter` could be configured.

**cherry-pick to 2.9.x**


artembilan (changes requested):
Don't we need some docs updates on the matter as well?
Thanks

artembilan:
Back-ported as https://github.com/spring-projects/spring-kafka/commit/b902b63b5530fec33a909968b38dbf35c8412f11 after some minor conflict fixes.

Issue #2576, opened 2023-02-10
Non-blocking retries: partition rebalance causes emergency consumer stop


v-chernyshev:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

2.9.4, 3.0.2.

**Describe the bug**

When the "seek after error" recovery mode is disabled, `KafkaMessageListenerContainer` stashes the records returned from the most recent `poll` call into a member variable:

https://github.com/spring-projects/spring-kafka/blob/addb888088fcaa8acd375309e26a00cd46e60384/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2929-L2932

If a partition rebalance happens before they are processed, the affected consumer instance terminates with an emergency stop error:

https://github.com/spring-projects/spring-kafka/blob/addb888088fcaa8acd375309e26a00cd46e60384/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1672-L1676

It seems that the only possible reason for it is that the newly assigned partitions are not properly paused.

**To Reproduce**

I've uploaded a minimal example into [this](https://github.com/v-chernyshev/spring-kafka-rebalance-issue) GitHub repository. Unfortunately, it lacks the Docker Compose setup as we are not allowed to use Docker Hub at work and, therefore, have a custom Kafka container that is not compatible with the open source one. Anyone wishing to reproduce this bug should create the following Kafka topics, with 4 partitions each:

- input
- input-retry-0
- input-dlt
- output
 
You may also have to adjust the bootstrap servers configuration in `application.yaml`. Assuming that this is done, the steps to trigger the problem are:

- Send like 10 messages to the input topic via e.g. `for i in {1..10}; do echo "test_$i" | kcat -b localhost:9192 -t input -P; done`.
- Clone the above example repository.
- Start one instance of the service via `mvn spring-boot:run`. Wait for the partitions to be assigned and for the first `Processing message=` to appear in the console log.
- Start another instance of the service and wait for the rebalance.

The first instance then fails with the emergency stop error, sometimes for multiple topics at once:

> 2023-02-10T17:48:49.149Z  INFO 40483 --- [ntainer#0-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : rebalance-issue-consumer: partitions revoked: [input-0, input-1, input-2, input-3]
[...]

PR #2579, opened 2023-02-14
GH-2576: Fix No Seeks After Error Handling

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2576

When the error handler `seeksAfterError` is false, we keep the unprocessed records in memory (`retainedRecords`), pause the consumer, and attempt redelivery.

If a rebalance occurs during this state, the behavior depends on whether a legacy or cooperative partition assignor is used.

With a legacy assignor, all partitions are unassigned and usually a subset is reassigned. This causes their positions to be reset to the last committed offset. In this case, all retained records must be cleared, otherwise an emergency stop occurs because records are polled when we don't expect them.

With a cooperative assignor, only a subset of the partitions are revoked. Records for these partitions must be pruned from the `retainedRecords` and not redelivered to the listener.

- add code to the rebalance listener to prune the revoked paritions from the retained records
- re-pause any assigned partitions if retained records are present

Add tests for both types of assignor.

Also tested with the reporter's reproducer.

**cherry-pick to 2.9.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -2643,7 +2643,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 
 		private boolean checkImmediatePause(Iterator<ConsumerRecord<K, V>> iterator) {
 			if (isPaused() && this.pauseImmediate) {
-				Map<TopicPartition, List<ConsumerRecord<K, V>>> remaining = new HashMap<>();
+				Map<TopicPartition, List<ConsumerRecord<K, V>>> remaining = new LinkedHashMap<>();
 				while (iterator.hasNext()) {
 					ConsumerRecord<K, V> next = iterator.next();
 					remaining.computeIfAbsent(new TopicPartition(next.topic(), next.partition()),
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3537,6 +3538,23 @@ public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
 				}
 			}
 
+			private void removeRevokationsFromPending(Collection<TopicPartition> partitions) {
```
artembilan, line 3541:
```suggestion
			private void removeRevocationsFromPending(Collection<TopicPartition> partitions) {
```
?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3498,6 +3498,7 @@ private class ListenerConsumerRebalanceListener implements ConsumerRebalanceList
 			@Override
 			public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
 				this.revoked.addAll(partitions);
+				removeRevokationsFromPending(partitions);
```
artembilan, line 3501:
```suggestion
				removeRevocationsFromPending(partitions);
```
?

artembilan (changes requested):
Just possible typo.

artembilan:
Cherry-picked to `2.9.x` as https://github.com/spring-projects/spring-kafka/commit/dc026c724a6b4dde8b9674f4154fbf51a8d51a49 after fixing conflicts.

Issue #2581, opened 2023-02-14
Container Does Not Pause for Missing Acks After a Rebalance


garyrussell:
### Discussed in https://github.com/spring-projects/spring-kafka/discussions/2575

Reproducer provided in the discussion exhibits the behavior; the container should again pause after the second set of 500 records are received.

AminArria:
Tested with 3.0.3-SNAPSHOT and everything is working as expected :smile: 

Thank you so much for the help with this and the quick turn around!

garyrussell:
Thanks for testing; the 3.0.3 release is scheduled for next Tuesday (and 2.9.6).

PR #2586, opened 2023-02-15
GH-2581: Fix OOO Commits with Rebalance

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2581

If a rebalance occurred while pending out of order commits were present, the pending commits were correctly purged, however, the `consumerPaused` boolean remained true.

This prevented the consumer from being paused again while the next batch of records is being processed and subsequent polls returned more records.

Reset the `consumerPaused` boolean when all pending commits are purged.

Also, with a cooperative assignor, ensure that any newly assigned partitions are paused if pending commits are present.

Add tests to verify the correct behavior with a legacy and cooperative assignor.

Also tested with the reporter's reproducer.

**cherry-pick to 2.9.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -3594,7 +3594,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 				boolean pending = false;
 				synchronized (ListenerConsumer.this) {
 					Map<TopicPartition, List<Long>> pendingOffsets = ListenerConsumer.this.offsetsInThisBatch;
-					if (!ObjectUtils.isEmpty(pendingOffsets) {
+					if (!ObjectUtils.isEmpty(pendingOffsets)) {
 						pending = true;
 					}
 				}
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -130,6 +130,7 @@ import org.springframework.transaction.support.TransactionTemplate;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.CollectionUtils;
+import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 
 import io.micrometer.observation.Observation;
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3586,7 +3591,16 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
 			}
 
 			private void repauseIfNeeded(Collection<TopicPartition> partitions) {
-				if (isPaused() || ListenerConsumer.this.remainingRecords != null && !partitions.isEmpty()) {
+				boolean pending = false;
+				synchronized (ListenerConsumer.this) {
+					Map<TopicPartition, List<Long>> pendingOffsets = ListenerConsumer.this.offsetsInThisBatch;
+					if (pendingOffsets != null && !pendingOffsets.isEmpty()) {
```
artembilan, line 3597:
```suggestion
					if (!ObjectUtils.isEmpty(pendingOffsets) {
```
?

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerMockTests.java
```diff
@@ -1116,6 +1118,178 @@ public boolean handleOne(Exception thrownException, ConsumerRecord<?, ?> record,
 		assertThat(recordsDelivered.get(2)).isEqualTo(record1);
 	}
 
+	@SuppressWarnings({ "unchecked", "rawtypes" })
+	@Test
+	void pruneRevokedPartitionsFromPendingOutOfOrderCommitsLegacyAssignor() throws InterruptedException {
```
artembilan, line 1123: Gary, I don't see any `Assignor` settings in the test code.
How does it work, please?

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerMockTests.java
garyrussell: It's simulated.

With a legacy assignor, all current assignments are revoked and all (or a subset) are re-assigned. With a coop assignor, only a subset are revoked and (possibly) previously unassigned partitions might be assigned.

See the `case 1:` poll phases in both tests - that is where the simulation is. In the coop case, I test that `consumer.pause()` is called twice (once for the original assignments and once for the newly assigned partition).

artembilan: That's cool. Thank you!

artembilan:
Cherry-picked as https://github.com/spring-projects/spring-kafka/commit/c8088ef66ebab4c13b36802365cd9a130949ba3e after some conflict fixes.

Issue #2377, opened 2022-08-09
Feature Request: Provide `DestinationTopic.Properties#isRetryTopic()` method


vivekragunathan:
Based on @tomazfernandes's [**request**](https://stackoverflow.com/a/73285101/495612), filing this feature request for a `DestinationTopic.Properties#isRetryTopic()` method. There are currently `isMainTopic` and `isDltTopic` methods but not a `isRetryTopic` method.

Workaround today is:

```java
Boolean isRetryTopic = !properties.isMainTopic() && !properties.isDltTopic();
```

tomazfernandes:
Thanks @vivekragunathan - also, feel free to provide a `PR` if you'd like.

vivekragunathan:
Yes sure ... that PR for some bed time stretching tonight

PR #2589, opened 2023-02-17
GH-2377: Add DestinationTopic.Properties#isRetryTopic()

echo304:
Resolve https://github.com/spring-projects/spring-kafka/issues/2377

`DestinationTopic.Properties` has `isDltTopic()` and `isMainEndpoint()` but not for retry topic. It adds `isRetryTopic()` method and adds tests to verify the correct behavior.

* Add `isRetryTopic()` instance method to `DestinationTopic.Properties`

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopic.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopic.java
@@ -212,7 +212,9 @@ public class DestinationTopic {
 			return Type.DLT.equals(this.type);
 		}
 
-		public boolean isRetryTopic() { return Type.RETRY.equals(this.type) || Type.SINGLE_TOPIC_RETRY.equals(this.type); }
+		public boolean isRetryTopic() {
+			return Type.RETRY.equals(this.type) || Type.SINGLE_TOPIC_RETRY.equals(this.type);
+		}
 
 		public String suffix() {
 			return this.suffix;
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopic.java
```diff
@@ -212,6 +212,8 @@ public boolean isDltTopic() {
 			return Type.DLT.equals(this.type);
 		}
 
+		public boolean isRetryTopic() { return Type.RETRY.equals(this.type) || Type.SINGLE_TOPIC_RETRY.equals(this.type); }
```
garyrussell, line 215:
```suggestion
		public boolean isRetryTopic() {
			return Type.RETRY.equals(this.type) || Type.SINGLE_TOPIC_RETRY.equals(this.type);
		}
```

garyrussell (changes requested):
Thanks for the contribution; please run `./gradlew clean check` to run checkstyle rules.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopic.java
echo304: Thank you for pointing it out!

echo304: Fixed! 👍

Issue #2596, opened 2023-02-22
 KafkaMessageListenerContainer pauses for a long time after error handling


antonin-arquey:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

`>= 2.9.0`

**Describe the bug**

It is stated in the documentation that

> Starting with version 2.9, the DefaultErrorHandler can be configured to provide the same semantics as seeking the unprocessed record offsets as discussed below, but without actually seeking. Instead, the records are retained by the listener container and resubmitted to the listener after the error handler exits (and after performing a single paused poll(), to keep the consumer alive; 

Listeners that receives a large amount of data will start to lag behind quickly if they encounter unexpected errors during processing because they are constantly paused and polling for a long time (by default the poll timeout is 5 seconds but could be larger if set otherwise by the user).

**To Reproduce**

Really easy to test, just throw any exception in a listener configured with a [CommonLoggingErrorHandler](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonLoggingErrorHandler.java) and a poll timeout set to any large value.

The listener will be stuck for the duration of the poll timeout before processing the remaining records.

**Expected behavior**

Do not poll with the consumer paused before processing the remaining records (or at least poll with a shorter poll timeout) as it is not expecting to receive a record anyway.

This should probably be mentionned in the [release notes](https://docs.spring.io/spring-kafka/reference/html/#spring-kafka-intro-new) as it seems small but is quite a big behavoural change and can lead to issues for some people upgrading to SpringBoot 3.

garyrussell:
>a big behavoural change

?? 

The option is turned off by default for exactly that reason.

>Do not poll with the consumer paused ...

We have to poll, to avoid a possible rebalance when there are many errors in the records returned by the previous poll.

But your point is well taken. We should certainly reduce the poll time to zero (or near it) when we know that all partitions are paused.

In a future release, we should consider reducing the default poll timeout.

antonin-arquey:
> ??
> 
> The option is turned off by default for exactly that reason.

Sorry you are right, the `DefaultErrorHandler` does not behave this way so in fact there is no behavioural change out of the box.

[...]

PR #2597, opened 2023-02-22
GH-2596: Fix No Seek Error Handling

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2596

- Do not pause and poll when an error handler "handles" an error
  - this was a bug in that the remaining records were incorrectly built after the error handler indicated that it had handled the error
- Reduce the poll timeout while the consumer is paused because we won't receive any records anyway

**cherry-pick to 2.9.x**

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/LoggingErrorHandlerNoPauseAfterHandle.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/LoggingErrorHandlerNoPauseAfterHandle.java
@@ -18,40 +18,34 @@ package org.springframework.kafka.listener;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyLong;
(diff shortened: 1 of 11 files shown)


spring-kafka/src/test/java/org/springframework/kafka/listener/LoggingErrorHandlerNoPauseAfterHandle.java
```diff
@@ -0,0 +1,216 @@
...
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+/**
+ * @author Gary Russell
+ * @since 2.9
+ *
+ */
+@SpringJUnitConfig
+@DirtiesContext
+public class LoggingErrorHandlerNoPauseAfterHandle {
```
artembilan, line 70: Is this a new style not using a `Test`suffix?

artembilan (changes requested):

artembilan:
Back-ported to `2.9.x` as https://github.com/spring-projects/spring-kafka/commit/28518485d951e4a485939e45a6d0e6084346893d

Issue #2600, opened 2023-02-27
Spring Messaging batch payload incompatible with Kotlin


samyempcc:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**
3.0.2

**Describe the bug**
Cross posting from Spring Messaging as suggested in https://github.com/spring-projects/spring-framework/issues/29963. Kotlin based batch listener does not receive the expected Message payload.

In: https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java#L682

with Kotlin the paramType resolves to:
` ? extends org.springframework.messaging.Message<?>`

while for java, it is:
`org.springframework.messaging.Message<?>`

The rest of the code flow in that method ends up assigning the payload content based on the test for paramType and ends up behaving differently for kotlin.

**To Reproduce**

Running the sample project, it is noticed that in the java.util.List based listener, a GenericMessage is passed to the listener with both header and payload. In Kotlin's List based listener, only the message payload is passed as Message.

**Expected behavior**

Kotlin listener is expected to be identical to Java.

**Sample**
I have a working code to reproduce the issue given a local kafka running on port 9092:
https://github.com/samyem/spring-messaging-issue29963/blob/master/src/main/kotlin/com/example/issue29963/service/BatchTest.kt#L35

garyrussell:
Confirmed; same thing happens with Java and

```java
@KafkaListener(id = "kgh2600", topics = "kgh2600", batch = "true")
void listen(List<? extends Message<?>> in) {
	System.out.println(in);
}
```

Thanks for reporting.

PR #2602, opened 2023-02-28
GH-2600: Fix Kotlin Listener List<Message<*>>

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2600

Also reproduced in Java with

	@KafkaListener(id = "kgh2600", topics = "kgh2600", batch = "true")
	void listen(List<? extends Message<?>> in) {
		System.out.println(in);
	}

MMLA did not handle a `WildcardType` as the list element type.

Also tested with reporter's Kotlin reproducer.

**cherry-pick to 2.9.x** (pattern matching instanceof will need to be changed)


artembilan:
Cherry-picked to `2.9.x` as https://github.com/spring-projects/spring-kafka/commit/ec7fe08f38d9e691f48fb2dd8bdf1d9f1aba02b2

Issue #2612, opened 2023-03-09
Spring Kafka Listener uses stale oauthbearer token, fails to connect after initial token expired.


mustafacantekir:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

2.8.11

Also tried with version 2.9.6 didn't work.

**Describe the bug**

I am connecting to a secure kafka that uses OAUTHBEARER SASL Mechnanism. The application initially starts as expected, gets the token and connects to the Kafka. The mentioned token is valid for 30 minutes. After ~25 minutes Kafka client executes reLogin using [ExpiringCredentialRefreshingLogin](https://github.com/a0x8o/kafka/blob/master/clients/src/main/java/org/apache/kafka/common/security/oauthbearer/internal/expiring/ExpiringCredentialRefreshingLogin.java). Relogin completed successfully and loginContext updated with the new token. Although after 30 minutes, KafkaMessageListenerContainer stops due to TopicAuthorizationException. Because the Kafka Consumers still using the stale token and doesn't fetch the new one. I am sharing the related logs below. 

```
2023-03-09 17:25:57.510  INFO 62343 --- [           main] .o.i.e.ExpiringCredentialRefreshingLogin : Successfully logged in.
2023-03-09 17:25:57.512  INFO 62343 --- [71-3b21d55ff7c8] .o.i.e.ExpiringCredentialRefreshingLogin : [Principal=:xxx]: Expiring credential re-login thread started.
2023-03-09 17:25:57.514  INFO 62343 --- [71-3b21d55ff7c8] .o.i.e.ExpiringCredentialRefreshingLogin : [Principal=xxx]: Expiring credential valid from Thu Mar 09 17:25:57 CET 2023 to Thu Mar 09 17:55:57 CET 2023
2023-03-09 17:25:57.514  INFO 62343 --- [71-3b21d55ff7c8] .o.i.e.ExpiringCredentialRefreshingLogin : [Principal=xxx]: Proposed refresh time of Thu Mar 09 17:51:23 CET 2023 extends into the desired buffer time of 300 seconds before expiration, so refresh it at the desired buffer begin point, at Thu Mar 09 17:50:57 CET 2023
2023-03-09 17:25:57.515  INFO 62343 --- [71-3b21d55ff7c8] .o.i.e.ExpiringCredentialRefreshingLogin : [Principal=:xxx]: Expiring credential re-login sleeping until: Thu Mar 09 17:50:57 CET 2023
[...]

PR #2613, opened 2023-03-09
GH-2612: Add Option To Restart Container After Auth

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2612

Add a container property to enable restarting the container after an emergency stop due to authentication/authorization exceptions, perhaps due to credentials expiring.

**cherry-pick to 2.9.x**


artembilan (changes requested):
Something is off or not committed.
See GH Action failure.

PR #2619, opened 2023-03-14
Making sure startup waits until consumer thread is completed

gossie:
Hi,

I ran into an issue when writing a test that uses the combination of the @EmbeddedKafka and the @MockBean annotation. It turned out to be a race condition. At first I opened an issue in the spring-boot issue tracker (https://github.com/spring-projects/spring-boot/issues/34595), but after an analysis by @wilkinsona he pointed out that the problem can be isolated to spring-kafka.

I had the problem using Spring Boot 2.7.9 with spring-kafka 2.8.11. That is why I first created this PR: https://github.com/spring-projects/spring-kafka/pull/2618
I didn't find any information, if PRs are only allowed on the main branch.


garyrussell:
2.8.x is no longer supported; use 2.9.6 with Boot 2.7.

As I said on the other PR, this won't help (the future won't complete until the consumer stops); use one of the techniques I described on the Boot issue.

gossie:
Thank you!

Issue #2620, opened 2023-03-14
Reclassify Exception in FallbackBatchErrorHandler When Exception Changes


garyrussell:
https://stackoverflow.com/questions/75734108/notretryableexception-after-a-retryableexception-considered-as-retryableexceptio/75734579#75734579

Not retryable exceptions are not processed if they follow a retryable.

PR #2621, opened 2023-03-14
GH-2620: FallbackBatchErrorHandler Reclassify Ex

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2620

Previously, if a not-retryable exception is thrown during batch retry, after a retryable exception, it was not reclassified and retries continued.

Also use the last thrown exception to the recoverer.

**cherry-pick to 2.9.x - will need work on instanceof pattern matching**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
@@ -63,11 +63,44 @@ public final class ErrorHandlingUtils {
 	 * @param retryListeners the retry listeners.
 	 * @param classifier the exception classifier.
 	 * @since 2.8.11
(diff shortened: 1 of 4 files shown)


artembilan:
@garyrussell ,

there are more conflicts than just `instanceof pattern`.
For example, `FallbackBatchErrorHandler`:
```
	/**
	 * Set one or more {@link RetryListener} to receive notifications of retries and
	 * recovery.
	 * @param listeners the listeners.
	 * @since 3.0
	 */
	public void setRetryListeners(RetryListener... listeners) {
```
In the `2.9.x` it is a `ErrorHandlingUtils.retryListeners`.

Any chances that you can look into that and fix respectively?
Thanks

garyrussell:
...cherry-picked as 6c01c6774 after resolving conflicts.

PR #2624, opened 2023-03-16
Update kafka to 3.4.0

maxbeckers:
Currently is included the kafka version 3.3.2 with a medium vulnerability. For more details see https://cwe.mitre.org/data/definitions/502 and https://kafka.apache.org/cve-list. With the version 3.4.0 this is fixed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -68,7 +68,7 @@ ext {
 	jaywayJsonPathVersion = '2.7.0'
 	junit4Version = '4.13.2'
 	junitJupiterVersion = '5.9.1'
-	kafkaVersion = '3.3.2'
+	kafkaVersion = '3.4.0'
 	log4jVersion = '2.19.0'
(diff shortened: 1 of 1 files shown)


garyrussell:
Spring projects cannot upgrade to a new major or minor version of dependencies in a patch release; we can only upgrade to a new patch release of the dependency; e.g. `3.3.2->3.3.3`.

Supported versions https://spring.io/projects/spring-kafka#support are fully compatible with the 3.4.0 clients; you just have to override the dependency in your application; see https://docs.spring.io/spring-kafka/docs/current/reference/html/#update-deps for how to do that in a Spring Boot app; if not using Boot, just override the dependency version.

garyrussell:
The CVE you cited is in Kafka Connect, which is not used by this project; it is only an issue if you use this project and connect in the same application.

PR #2631, opened 2023-03-18
GH-2273: introduce extension point for creation of DeadLetterPublishingRecoverer instance

### LINKED ISSUE #2273: Edit producerRecord before sending to dlt
Hello, I was reading the documentation and could not identify a solution to the following problem:

I want to add new attributes in the message before sending to the DLT topic by injecting this at the configuration level, I want to add attributes in the message body coming from the exception itself.

I believe what I'm looking for is a type of interceptor that allows me to change data.

And that it is possible to combine this solution with the RetryTopicConfiguration (non-blocking attempts).

_Originally posted by @Watlas at https://github.com/spring-projects/spring-kafka/discussions/2265_

breader124:
Fixes #2273.

I've introduced extension point for creation of `DeadLetterPublishingRecoverer` instance inside the `DeadLetterPublishingRecovererFactory` as suggested in this [comment](https://github.com/spring-projects/spring-kafka/discussions/2265#discussioncomment-2715773). Using `spring-kafka` version 2.9 and higher, the original issue can be solved following the procedure described in the [Javadoc](https://docs.spring.io/spring-kafka/api/org/springframework/kafka/retrytopic/RetryTopicComponentFactory.html) of `RetryTopicComponentFactory` class. In the subclassed `RetryTopicComponentFactory` there's need to override `deadLetterPublishingRecovererFactory` method and make sure that instance overriding `createDeadLetterPublishingRecovererInstance` is returned. In this method it's possible to actually create customised instance of subclassed `DeadLetterPublishingRecoverer` and there override `createProducerRecord` as suggested during the [discussion](https://github.com/spring-projects/spring-kafka/discussions/2265#discussioncomment-2701740).

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DeadLetterPublishingRecovererFactory.java
@@ -52,6 +52,7 @@ import org.springframework.util.Assert;
  *
  * @author Tomaz Fernandes
  * @author Gary Russell
+ * @author Adrian Chlebosz
  * @since 2.7
  *
  */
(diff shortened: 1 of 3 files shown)


garyrussell:
2.8.x is no longer supported as OSS: https://spring.io/projects/spring-kafka#support

2.9.x is fully compatible. with Boot 2.7.x and, given that there is already a solution for 2.9.x and later, and this won't be back ported, there is no point in merging it.

If you have a commercial support contract with VMware, and can't upgrade to 2.9.x for some reason, we can reconsider.

Issue #2641, opened 2023-03-27
"resetStateOnExceptionChange" has no effect, because the cause of "ListenerExecutionFailedException" is always a "TimestampedException"


sebastianloefflath:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

spring-kafka 3.0.5

**Describe the bug**

same bug as in [https://github.com/spring-projects/spring-kafka/issues/1653](https://github.com/spring-projects/spring-kafka/issues/1653).

spring-kafka does not recognize an exception change from `RetryException` to `FatalException`, because it is wrapped multiple times. (`ListenerExecutionFailedException` -> `TimestampedException` -> `ListenerExecutionFailedException` -> `FatalException`)

The `FailedRecordTracker` unwraps only one layer (that was the fix for [https://github.com/spring-projects/spring-kafka/issues/1653](https://github.com/spring-projects/spring-kafka/issues/1653))so it seems like the Exception does not change and is always `TimestampedException`

So this happens from a records view:
A record gets retried (blocking) -> the blocking error resolves -> fatal exception occurs on the same record -> record is being retried until configured number of blocking retry attempts.

**To Reproduce**

https://github.com/sebastianloefflath/spring-retry-bug
Run the tests of this repo to see the bug.

**Expected behavior**

Put the record to the DeadLetterTopic as soon as the `FatalException` occurs

**Sample**

https://github.com/sebastianloefflath/spring-retry-bug

garyrussell:
Thanks for the reproducer; it is much appreciated; I wish more folks would take the time to do that.

PR #2645, opened 2023-03-28
GH-2641: Fix Exception Change with RetryableTopic

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2641

Need to traverse the cause tree to find the actual user exception.

**cherry-pick to 2.9.x**


artembilan:

PR #2661, opened 2023-04-19
Fix Race in Micrometer Test

garyrussell:
When timers have dynamic tags, creation of the timers is deferred until the first completion of a timer with a set of tag values.

Use awaitility to wait for the timer to be created before checking the count.

**cherry-pick to 2.9.x**


artembilan:

Issue #2677, opened 2023-05-08
CommonErrorHandler for asyncAcks and AckMode.MANUAL


garyrussell:
### Discussed in https://github.com/spring-projects/spring-kafka/discussions/2668

<div type='discussions-op-text'>

<sup>Originally posted by **linus-learns** April 27, 2023</sup>
Hi,

I'm running a setup with `@KafkaListener` annotated consumers using a `ConcurrentKafkaListenerContainerFactory` setup with `asyncAcks` and `AckMode.MANUAL`. I have tried to implement a `CommonErrorHandler` for my setup with `isAckAfterHandle` set to `true` and `seeksAfterHandling` set to `false`, but the bookkeeping of the deferred acks does not work, as the the ack after the error handler has been invoked and the broken record handled is not done through the `ackInOrder` method. This causes the consumer to hang if a poison pill appears in the middle of the batch and higher offset commits are waiting in the `ListenerConsumer`.
It appears that the ack for the poison pill is never commited until the consumer get's the shutdown signal, as a rebalance causes the broken record to be fetched by the new partition owner but a restart resolves the issue.
Is this a bug, or should I use a different way of handling my errors given my setup?</div>

PR #2678, opened 2023-05-08
GH-2677: Fix EH AckAfterHandle With Async Acks

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2677

When an error handler handles an error (and `ackAfterHandle` is true), the ack bypassed the out of order commit logic, causing the consumer to be paused indefinitely, due to the missing ack.

**cherry-pick to 2.9.x**


artembilan:
... and cherry-picked to `2.9.x` as https://github.com/spring-projects/spring-kafka/commit/8725b654e43f79704def2d073f6385e89297075a after minor conflict fixing.

PR #2680, opened 2023-05-10
replace `.methodName()` with `methodName()` in documentation

appchemist:
There are two forms of method representation in the documentation.
- `methodName()`
- `.methodName()`

I think that `methodName()` is the commonly used form in this documentation.
How about replacing `.methodName()` to `methodName()` in this documentation.


garyrussell:
Thanks!

PR #2683, opened 2023-05-15
GH-2190: add ctors to reactive producer and consumer

### LINKED ISSUE #2190: Consider providing custom KafkaSender to ReactiveKafkaProducerTemplate
**Context**

My application uses `spring-cloud-sleuth` for tracing so I want it to be propagated through kafka. I am using `ReactiveKafkaProducerTemplate` to send events but it does not support tracing propagation. Although, `spring-cloud-sleuth-instrumentation` contains `TracingKafkaProducerFactory` that creates required `KafkaSender`. I tried to copy-paste `ReactiveKafkaProducerTemplate` implementation with setting this `KafkaSender` and it works fine as expected.

I guess `ReactiveKafkaConsumerTemplate` needs the same mechanism

**Expected Behavior**

I think constructor with `KafkaSender` would be fine

**Current Behavior**

There is no way to provide custom `KafkaSender` for now

breader124:
Fixes #2190

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2022 the original author or authors.
+ * Copyright 2019-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -51,6 +51,7 @@ import reactor.util.function.Tuples;
  * @param <V> the value type.
  *
  * @author Mark Norkin
+ * @author Adrian Chlebosz
  *
  * @since 2.3.0
  */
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplate.java
```diff
@@ -71,6 +72,13 @@ public ReactiveKafkaProducerTemplate(SenderOptions<K, V> senderOptions, RecordMe
 		this.messageConverter = messageConverter;
 	}
 
+	public ReactiveKafkaProducerTemplate(KafkaSender<K, V> sender, RecordMessageConverter messageConverter) {
```
garyrussell, line 75: We should probably also add a variant that does not take a converter (to be consistent with the other ctors).

garyrussell (changes requested):
Thanks! Just one small suggestion.

garyrussell:
@breader124 Thanks - in future, please just add more commit(s) to the PR rather than force-pushing a single commit; that way we can easily see the incremental change(s).

garyrussell (approved):

Issue #2590, opened 2023-02-17
Consumer committed offsets not tracking LSO with FixTxOffsets and rollbacked transactions


bgK:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**
2.8.11

**Describe the bug**
With a `MessageListenerContainer` configured with `FixTxOffsets` set to `true`, when all the available messages have been read, the committed offsets do not always match the Last Stable Offsets.

`MessageListenerContainer` commits the position of the consumer when records are returned by `poll()` for a given partition. If a transaction is then rollbacked on that partition and no other records are available, `MessageListenerContainer` does not commit the offsets consumed by the rollbacked transaction, resulting in observable "lag".

**To Reproduce**
Create and start a `MessageListenerContainer` with `FixTxOffsets` set to true subscribed to a test topic with a single partition. Create a transactional producer, send a message, wait for the ack and rollback the transaction. The commited offset will be 0, the last stable offset will be 2.

**Expected behavior**
When all the messages have been consumed, with a `MessageListenerContainer` configured with `FixTxOffsets` set to `true` the committed offsets should track the LSO even when some transactions are rollbacked.

garyrussell:
There is nothing we can do about this; the consumer has no knowledge that there is a rolled back record in the log; the container can only "fix" the offset if the consumer reports the position properly and, obviously, in this case, it does not.

As I have stated many times in many different forums, this bogus lag problem needs to be fixed in Kafka itself. There are other corner cases where `position()` does not include the marker.

See https://issues.apache.org/jira/browse/KAFKA-10683

But the Kafka folks seem to have no interest in fixing it; I suggest you comment there and vote it up.

bgK:
Thanks for your answer.

It seems to me that in the case of this issue, the position is correctly reported by the Kafka consumer. When I patch Spring-Kafka to always commit the consumer position even when no records are returned by `poll()`, the committed offsets account for the rolled back transactions.

Hacky patch for v2.8.11:
```diff
diff --git a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
index a787e0f5..480ff2cc 100644
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
[...]

PR #2685, opened 2023-05-17
Fix for issue #2590

thperchi:
Hello,

I hope this message finds you well. I am writing to submit a pull request for the spring-kafka project to address a fix for the issue #2590.

Here are the details of the pull request:

Repository: https://github.com/INTM-Group/spring-kafka
Issue: Consumer committed offsets not tracking LSO with FixTxOffsets and rollbacked transactions #2590

It has successfully resolved the issue. The changes I made adhere to the existing codebase's style and best practices.

I kindly request you to review and consider merging this pull request into the main branch. I am open to any feedback or suggestions you may have, and I am committed to addressing them promptly.

Thank you for your time and consideration. I appreciate the opportunity to contribute to the spring-kafka project.

Thibault Perchicot

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafkaContextCustomizerFactory.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafkaContextCustomizerFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 the original author or authors.
+ * Copyright 2017-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,10 +18,10 @@ package org.springframework.kafka.test.context;
 
 import java.util.List;
 
-import org.springframework.core.annotation.AnnotatedElementUtils;
 import org.springframework.test.context.ContextConfigurationAttributes;
 import org.springframework.test.context.ContextCustomizer;
 import org.springframework.test.context.ContextCustomizerFactory;
+import org.springframework.test.context.TestContextAnnotationUtils;
 
 /**
  * The {@link ContextCustomizerFactory} implementation to produce a
@@ -38,7 +38,7 @@ class EmbeddedKafkaContextCustomizerFactory implements ContextCustomizerFactory 
 	public ContextCustomizer createContextCustomizer(Class<?> testClass,
 			List<ContextConfigurationAttributes> configAttributes) {
 		EmbeddedKafka embeddedKafka =
-				AnnotatedElementUtils.findMergedAnnotation(testClass, EmbeddedKafka.class);
+				TestContextAnnotationUtils.findMergedAnnotation(testClass, EmbeddedKafka.class);
 		return embeddedKafka != null ? new EmbeddedKafkaContextCustomizer(embeddedKafka) : null;
 	}
 
--- a/spring-kafka-test/src/test/java/org/springframework/kafka/test/condition/WithNestedClassContextTests.java
+++ b/spring-kafka-test/src/test/java/org/springframework/kafka/test/condition/WithNestedClassContextTests.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2021-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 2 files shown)


garyrussell:
Thanks; but we can't look at your PR until you "sign" the CLA.

Do you think it would be possible to create a test that exhibits the behavior without the patch and passes with it?

thperchi:
Hi !
It should be good for the CLA.
Here is a test that fail without the pach and passes with it : [spring-kafka-2590-repro-v2.zip](https://github.com/spring-projects/spring-kafka/files/11514960/spring-kafka-2590-repro-v2.zip)
Thanks for your time.

thperchi:
Thank you for the review, I'm gonna check out how to solve this !

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1633,17 +1635,17 @@ private void fixTxOffsetsIfNeeded() {
 			if (this.fixTxOffsets) {
 				try {
 					Map<TopicPartition, OffsetAndMetadata> toFix = new HashMap<>();
-					this.lastCommits.forEach((tp, oamd) -> {
+					this.savedPositions.forEach((tp, saved) -> {
 						long position = this.consumer.position(tp);
-						Long saved = this.savedPositions.get(tp);
+						OffsetAndMetadata comitted = this.lastCommits.get(tp);
```
bgK, line 1640: `this.lastCommits` only contains offsets for `TopicPartition`s for which records were obtained during the previous `poll()` call. `this.lastCommits` is cleared between `poll`s.
This means:
* `poll`s returning no records will always trigger a commit for the complete assignment.
* `poll`s returning records for some partitions only will trigger commits for the other partitions even when the LSO did not change since the last commit.

Should the fix try to avoid unnecessary commits or is it good enough as is?

garyrussell: @bgK I guess there is a situation where a producer might keep rolling back transactions and the lag will increase until a successful transaction appears in the log. 

I must admit that I am not comfortable doing a commit for each empty poll but if the above is true, then there is probably no choice.

I really wish the kafka folks would fix the underlying problem of reporting this bogus lag.

bgK: > I must admit that I am not comfortable doing a commit for each empty poll but if the above is true, then there is probably no choice.

Me neither. What I was hinting at with this comment is whether Spring-Kafka should remember the last committed offsets across polls to avoid redundant commits.

garyrussell: Right, but as I described, I believe there are cases where the lag can still increase unless we do.

I believe the proper fix is for Kafka to not report a lag if it's a bogus lag due to uncommitted, or rolled-back, records.

bgK: > Right, but as I described, I believe there are cases where the lag can still increase unless we do.

I agree Spring-Kafka should commit the offsets when the consumer position changed after empty polls. What's unclear to me is what committing the same offsets multiple times when the position does not change after empty polls would accomplish. Do you have any insight?

> I believe the proper fix is for Kafka to not report a lag if it's a bogus lag due to uncommitted, or rolled-back, records.

As far as I know, at this point, Kafka does not provide any other mechanism to report the consumer lag accurately when using transactional producers. Shouldn't Spring-Kafka make sure the situations where Kafka reports the lag properly work when using the Spring abstractions?

alograg:
Hi @garyrussell 
I will participate in the correction.
I understand clearly that the "lag" does not exist but Kafka sends it as such, because it is a bugus lag.
And based on your experience it is better that be fixed in Kafka.
Look for such a report in KAfka JIRA and i only find the report [KAFKA-10683 - Consumer.position() Ignores Transaction Marker with read_uncommitted](https://issues.apache.org/jira/browse/KAFKA-10683).
Do you have the report number for this behavior?
I plan to correct this problem from both sides.
I will start  generating the tests that show this error/behavior.

garyrussell:
@alograg I am not sure I can add any value to such a discussion, but feel free to use the Discussions tab above.

alograg:
[Dicsution](https://github.com/spring-projects/spring-kafka/discussions/2749)

Issue #2699, opened 2023-06-07
Non-Blocking Retries don't work with BatchListener


wapkch:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.0.7

**Describe the bug**

When using a batch listener together with retry topic, an IllegalArgumentException encounted

**To Reproduce**
 1. Define a batch listener together with retry topic:
```java
@SpringBootApplication
public class SpringKafkaDemoApplication {

    @RetryableTopic(attempts = "5", backoff = @Backoff(delay = 2_000, maxDelay = 10_000, multiplier = 2))
    @KafkaListener(id = "fooGroup", topics = "topic1", batch = "true")
    public void listen(List<ConsumerRecord<String, String>> data, Acknowledgment acknowledgment) throws Exception {
        // ...
    }

}
```

2. Start the SpringBoot application, the IllegalArgumentException encounted:
```css
java.lang.IllegalArgumentException: The provided class BatchMessagingMessageListenerAdapter is not assignable from MessageListener: class org.springframework.kafka.listener.adapter.BatchMessagingMessageListenerAdapter
	at org.springframework.util.Assert.assignableCheckFailed(Assert.java:731) ~[spring-core-6.0.9.jar:6.0.9]
	at org.springframework.util.Assert.isAssignable(Assert.java:681) ~[spring-core-6.0.9.jar:6.0.9]
	at org.springframework.kafka.retrytopic.ListenerContainerFactoryConfigurer.checkAndCast(ListenerContainerFactoryConfigurer.java:201) ~[spring-kafka-3.0.7.jar:3.0.7]
	at org.springframework.kafka.retrytopic.ListenerContainerFactoryConfigurer.setupBackoffAwareMessageListenerAdapter(ListenerContainerFactoryConfigurer.java:190) ~[spring-kafka-3.0.7.jar:3.0.7]
	at org.springframework.kafka.retrytopic.ListenerContainerFactoryConfigurer$RetryTopicListenerContainerFactoryDecorator.decorate(ListenerContainerFactoryConfigurer.java:241) ~[spring-kafka-3.0.7.jar:3.0.7]
	at org.springframework.kafka.retrytopic.ListenerContainerFactoryConfigurer$RetryTopicListenerContainerFactoryDecorator.createListenerContainer(ListenerContainerFactoryConfigurer.java:226) ~[spring-kafka-3.0.7.jar:3.0.7]
[...]

PR #2704, opened 2023-06-13
GH-2699: Document Retry Limitation

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2699

`@RetryableTopic` is not supported with batch listeners.


garyrussell:
No it's something internal in the broker; I have seen it occasionally before.

Issue #2702, opened 2023-06-10
Auto Config Non-Blocking Retries with asyncAcks


wapkch:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.0.7

**Describe the bug**

when using asyncAcks together with RetryableTopic, the retried message will not be invoked on KafkaListener.

**To Reproduce**
```java
@RestController
@SpringBootApplication
public class SpringkafkaAsyncacksDemoApplication {

    public static void main(String[] args) {
        SpringApplication.run(SpringkafkaAsyncacksDemoApplication.class, args);
    }

    @Autowired
    private KafkaTemplate<Object, Object> template;

    @PostMapping(path = "/send/{what}")
    public void sendFoo(@PathVariable String what) {
        this.template.send("topic1", what);
    }

    @KafkaListener(id = "fooGroup", topics = "topic1")
    @RetryableTopic(attempts = "5", backoff = @Backoff(delay = 2_000, maxDelay = 10_000, multiplier = 2))
    public void onMessage(ConsumerRecord<String, ?> record, Acknowledgment acknowledgment, Consumer<?, ?> consumer) {
        if (record.value().equals("fail")) {
            throw new RuntimeException("failed");
        }
        acknowledgment.acknowledge();
    }

}
```

application.properties
```properties
spring.kafka.listener.async-acks=true
spring.kafka.retry.topic.enabled=true
spring.kafka.listener.ack-mode=manual
```

1. Run local kafka server
2.  Start the application 
3. Access `http://localhost:8080/send/fail`, and the retried message will not be invoked on KafkaListener.

**Expected behavior**

I expect Non-Blocking Retries can work with asyncAcks.

I konw using asyncAcks in above case is meaningless, but sometimes the async processing in @KafkaListener is useful. For example:
```java
@Component
public class FooConsumer {

    @KafkaListener(id = "fooGroup", topics = "topic1")
    @RetryableTopic(attempts = "5", backoff = @Backoff(delay = 2_000, maxDelay = 10_000, multiplier = 2))
    public void onMessage(ConsumerRecord<String, ?> record, Acknowledgment acknowledgment, Consumer<?, ?> consumer) {
       threadPoolTaskExecutor.submit(() -> {
           try {
                 doConsume(record);
                 acknowledgment.acknowledge();
           } catch (Exception ex) {
                 container.getCommonErrorHandler().handleOne(new RuntimeException(), record, consumer, container);
           }
        });
    }

}
```

[...]

PR #2706, opened 2023-06-13
GH-2702: RetryableTopic with asyncAcks

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2702

When using `asyncAcks` with manual ack modes, the `DefaultErrorHandler` must have `seekAfterError` set to `false`; this required user configuration.

The framework now unconditionally sets the property when it configures a container using a manual ack mode.

In addition, the default DLT handler was not compatible with any manual ack mode, regardless of the `asyncAcks` setting.

Add `Acknowledgment` to the `LoggingDltListenerHandlerMethod`.

**cherry-pick to 2.9.x (will require instanceof polishing for Java 8)**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
@@ -50,6 +50,8 @@ public class KafkaBackoffAwareMessageListenerAdapter<K, V>
 		extends AbstractDelegatingMessageListenerAdapter<MessageListener<K, V>>
 		implements AcknowledgingConsumerAwareMessageListener<K, V> {
 
+	private static final Acknowledgment NO_OP_ACK = new NoOpAck();
+
 	private final String listenerId;
 
 	private final String backoffTimestampHeader;
@@ -94,7 +96,12 @@ public class KafkaBackoffAwareMessageListenerAdapter<K, V>
 				.ifPresent(nextExecutionTimestamp -> this.kafkaConsumerBackoffManager
 						.backOffIfNecessary(createContext(consumerRecord, nextExecutionTimestamp, consumer)));
 		try {
-			invokeDelegateOnMessage(consumerRecord, acknowledgment, consumer);
+			Acknowledgment ack = acknowledgment;
+			if (ack == null) {
+				// The default DLT handler now requires an Acknowledgment.
+				ack = NO_OP_ACK;
+			}
+			invokeDelegateOnMessage(consumerRecord, ack, consumer);
 		}
 		catch (Exception ex) {
 			throw new TimestampedException(ex, Instant.now(this.clock));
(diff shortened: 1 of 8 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
```diff
@@ -94,7 +96,12 @@ public void onMessage(ConsumerRecord<K, V> consumerRecord, @Nullable Acknowledgm
 				.ifPresent(nextExecutionTimestamp -> this.kafkaConsumerBackoffManager
 						.backOffIfNecessary(createContext(consumerRecord, nextExecutionTimestamp, consumer)));
 		try {
-			invokeDelegateOnMessage(consumerRecord, acknowledgment, consumer);
+			Acknowledgment ack = acknowledgment;
+			if (ack == null) {
+				// The default DLT handler now requires an Acknowledgment.
```
artembilan, line 101: but I see it is `@Nullable` there. So, what is wrong to continue to use `acknowledgment` as it is?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java
garyrussell: Because of this

https://github.com/spring-projects/spring-kafka/blob/8efd15fe8e4791ab69d0dfc3893cfb4902058868/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java#L370-L372

and this

https://github.com/spring-projects/spring-kafka/blob/8efd15fe8e4791ab69d0dfc3893cfb4902058868/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java#L386-L392

I changed the reporter's reproducer to use default `AckMode` and hit this problem.

I didn't think it was worth it to remove the null check in the method, but I can do so if you insist,

garyrussell: Hmmm - maybe I could see if the parameter is annotated `@Nullable` and skip that check - will look at it tomorrow.

garyrussell: No; that won't work; the invoke already failed.

garyrussell: I guess I would have to re-invoke the method in that case, since the invoke already failed.

garyrussell: Correction: I can check the nullability before the first invocation.

artembilan:
It doesn't cherry-pick clearly: too many conflicts around deprecated utils over there in `2.9.x`
Would you mind taking a look yourself, please?
Thanks

garyrussell:
Back ported as 4401f9689

Issue #2711, opened 2023-06-21
Add DLPR Creator Function to DeadLetterPublishingRecovererFactory


garyrussell:
### Discussed in https://github.com/spring-projects/spring-kafka/discussions/2265

<div type='discussions-op-text'>

<sup>Originally posted by **Watlas** May  6, 2022</sup>
Hello, I was reading the documentation and could not identify a solution to the following problem:

I want to add new attributes to the message before sending to the DLT topic, injecting this at the configuration level, I want to add attributes to the message body coming from the exception itself.

I believe what I'm looking for is a type of interceptor that allows me to change data.</div>

Also https://stackoverflow.com/questions/76518611/how-to-edit-message-data-before-sending-to-dlt-in-spring-kafka

PR #2712, opened 2023-06-21
GH-2711: Non-Blocking Retries and Custom Publisher

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2711

Previously, it was not easy to provide custom DLPRs, for example to override `createProducerRecord`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
@@ -863,7 +863,8 @@ public class DeadLetterPublishingRecoverer extends ExceptionClassifier implement
 	 * in the sent record.
 	 * @return the header names.
 	 * @since 2.7
-	 * @deprecated since 3.0.9 - provide a supplier in the constructor instead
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -825,25 +863,17 @@ private String getStackTraceAsString(Throwable cause) {
 	 * in the sent record.
 	 * @return the header names.
 	 * @since 2.7
+	 * @deprecated since 3.0.9 - provide a supplier in the constructor instead
```
artembilan, line 866: But I don't see that ctor arg - only a setter.
Should this be fixed respectively then?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
garyrussell: Sorry, that was from an earlier iteration, but I couldn't make backwards compatibility work that way. Will fix.

Issue #2723, opened 2023-07-05
Fix Race Condition in DelegatingInvocableHandler


garyrussell:
See https://stackoverflow.com/questions/76610495/single-nullpointerexception-at-spring-kafka-startup

PR #2724, opened 2023-07-05
GH-2723: Fix NPE in DelegatingInvocableHandler

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2723

Previously, it was possible for `handlerReturnsMessage` to return `null` when a `cachedHandlers` was present, causing an NPE when unboxing the `Boolean` value.

Ensure that the `handlerReturnsMessage` map is populated before adding the handler to `cachedHandlers`.

**cherry-pick to 2.9.x**


artembilan:
... and cherry-picked to `2.9.x`.

Issue #2722, opened 2023-07-05
Behavior Change When Interceptor Throws an Exception


garyrussell:
See https://stackoverflow.com/questions/76591684/exception-handling-after-migration-from-spring-kafka-2-6-to-2-8/76605214#comment135081001_76605214

Previously, throwing an exception from an interceptor caused the error handler to be invoked.

Back port to 2.9.10

PR #2727, opened 2023-07-05
GH-2722: Only Consider interceptBeforeTx with TX

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2722

Prior to 2.8.x, `interceptBeforeTx` was default `false`; it is now `true`. However, it should only be considered if there is a `TransactionManager` present.

This allows an interceptor (when no transactions) to throw an exception, causing error handling to be invoked.

Also tested with user's reproducer.

**cherry-pick to 2.9.x**


artembilan:

PR #2739, opened 2023-07-12
GH-2738: Pausing container results in WakeupException without retrying failed listener invocation

### LINKED ISSUE #2738: Pausing container results in WakeupException without retrying failed listener invocation

 - `3.0.7` - happened in production
 -  confirmed locally with test on recent `main`-branch commit `f079ed37ce307118387b752b43530a4ef87fa70b`

**Describe the bug**

Bug has pretty low chance of happening and we were hunting it for the last couple of months.

Scenario goes like this:
 - We initially have normally working situation.
 - At some point in time application-specific issues starts to happen (some downstream remote API calls failing).
 - We have mechanism which monitors some application-specific circuit breakers and at some point decide to `pause()` `MessageListenerContainer` to suspend processing.
 - If the timing of invocation to `MessageListenerContainer.pause()` is right after `KafkaMessageListenerContainer` [completing next poll](https://github.com/spring-projects/spring-kafka/blob/a067eaa288df66601ba69fa2acf12235d025c0bb/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1726) on `Consumer` but before `KafkaMessageListenerContainer.polling` being set [back to `false`](https://github.com/spring-projects/spring-kafka/blob/a067eaa288df66601ba69fa2acf12235d025c0bb/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1503), and also if next invocation of listener throws an exception, then we are in situation when this bug will manifest
 - Due to `KafkaMessageListenerContainer.polling` was `true` during `pause()` invocation it resulted in [waking-up the consumer](https://github.com/spring-projects/spring-kafka/blob/a067eaa288df66601ba69fa2acf12235d025c0bb/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1780-L1782)
- Now, since listener container successfully polled some records it proceeds with normal invocation of listener (batch listener)
- Next, listener throws any exception (in our case because some circuit breaker was open - reason why we paused container)
- Listener container now passes failed batch to `CommonErrorHandler` (we use `DefaultErrorHandler` with MAX_LONG retries) and when `ErrorHandlingUtils.retryBatch` for very first time calls [`consumer.poll(Duration.ZERO)`](https://github.com/spring-projects/spring-kafka/blob/a067eaa288df66601ba69fa2acf12235d025c0bb/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java#L123) it throws `WakeupException` (remember `wakeup()` was called during `pause()` which was meant to wake up the consumer from normal polling)
- This `WakeupException` now propagates [all the way up](https://github.com/spring-projects/spring-kafka/blob/a067eaa288df66601ba69fa2acf12235d025c0bb/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2350) without re-seeking Consumer nor doing any retries
- Later, when listener container is resumed, Consumer will continue from offsets after/newer than our perviously failed messages and new messages will arrive, be successfully processed and finally committed.

**To Reproduce**

It is somewhat hard to reproduce without particular setup because issue depends on "low probability" race condition.
I wrote test which reproduces it every time, I'll submit PR with such test.

**Expected behavior**

At the end of a day, we aim for "at-least-once" semantics, so that we do not skip some unprocessed messages.
I would expect that arbitrary sequence/timing of `pause()`/`resume()` calls wouldn't ever cause "loss" of messages. 

My expectation would be that `WakeupException` on [`consumer.poll(Duration.ZERO)`](https://github.com/spring-projects/spring-kafka/blob/a067eaa288df66601ba69fa2acf12235d025c0bb/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java#L123) should re-seek the consumer.

Another approach would be to fix race-condition between `pause()...if(polling.get())` and `doPoll()...polling.compareAndSet(true, false)`, but even then ,it seems that any other `KafkaException` raised from `consumer.poll(Duration.ZERO)` could have similar scenario.

**Sample**

I'll link fixing PR

antonio-tomac:
Pausing container results in WakeupException without retrying failed listener invocation

GH-2738
Fixes GH-2738 (https://github.com/spring-projects/spring-kafka/issues/2738)
Closes #2738 

I wasn't sure what to put in `@since <version>` on test class, since it seems that this fix should be easily back-ported to older versions (if worthy to do so).


garyrussell:
Thanks for the PR; FYI, you can run checkstyle (and tests) locally, with `./gradlew clean check`.

antonio-tomac:
> you can run checkstyle (and tests) locally, with `./gradlew clean check`.

I've run all tests and all are passing, but I'm having trouble with code style checking. 
For example:
```
Execution failed for task ':checkstyleNohttp'.
> Could not resolve all files for configuration ':checkstyle'.
   > Could not resolve com.puppycrawl.tools:checkstyle:8.37.
     Required by:
         project :
      > Could not resolve com.puppycrawl.tools:checkstyle:8.37.
         > Could not get resource 'https://repo.maven.apache.org/maven2/com/puppycrawl/tools/checkstyle/8.37/checkstyle-8.37.pom'.
            > Could not GET 'https://repo.maven.apache.org/maven2/com/puppycrawl/tools/checkstyle/8.37/checkstyle-8.37.pom'.
               > The server may not support the client's requested TLS protocol versions: (TLSv1.2, TLSv1.3). You may need to configure the client to allow other protocols to be used. See: https://docs.gradle.org/7.4.1/userguide/build_environment.html#gradle_system_properties
                  > PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
 ```
I can see something problematic with download of resources, but then I gave up on troubleshooting TLS certificates. And just proceeded to submitting PR, sorry.

**Comment by garyrussell** [comment]
OK; thanks; I will fix the checkstyle and merge in the morning; releases are due on Monday.

Again, thanks for the analysis of this issue, and the PR.

**Comment by garyrussell** [comment]
Please don't force push in future; just add more commits (it makes it easier for us to see what changed).

File: spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java [hunk 1]
```diff
@@ -120,7 +122,13 @@ public static void retryBatch(Exception thrownException, ConsumerRecords<?, ?> r
 			Exception lastException = unwrapIfNeeded(thrownException);
 			Boolean retryable = classifier.classify(lastException);
 			while (Boolean.TRUE.equals(retryable) && nextBackOff != BackOffExecution.STOP) {
-				consumer.poll(Duration.ZERO);
+				try {
+					consumer.poll(Duration.ZERO);
+				}
+				catch (WakeupException we) {
+					seeker.handleBatch(thrownException, records, consumer, container, () -> { });
+					throw new KafkaException("Woken up during retry", logLevel, we);
```
garyrussell, line 130: I don't think we need to re-seek and throw an exception here, we can just ignore the wakeup (perhaps try another poll to keep the broker from triggering a rebalance because this poll failed.

spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
```diff
@@ -132,7 +140,13 @@ public static void retryBatch(Exception thrownException, ConsumerRecords<?, ?> r
 				if (!container.isRunning()) {
 					throw new KafkaException("Container stopped during retries");
 				}
-				consumer.poll(Duration.ZERO);
+				try {
+					consumer.poll(Duration.ZERO);
+				}
+				catch (WakeupException we) {
+					seeker.handleBatch(thrownException, records, consumer, container, () -> { });
+					throw new KafkaException("Woken up during retry", logLevel, we);
```
garyrussell, line 148: ditto

garyrussell (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
garyrussell: On the other hand, since this is such a rare race condition, I guess doing a seek under these circumstances will not significantly affect performance.

antonio-tomac: Yes, I've just tried it and the test is passing, but shouldn't we bail-out from retrying because my idea of pausng container is to suspend any new traffic and also suspend re-attempts.

garyrussell: Good point; ok; will merge and back-port tomorrow.

antonio-tomac: Now when I think more about what I just wrote, I realized that this retry loop doesn't check about container's _pause_ state at all. It does check `isRunning()` but not `isPaused()/isPauseRequsted()`, and retries will continue.

 Is that how it's supposed to behave or it should be addressed as separate issue? (seems like it's a bit out-of-scope of this PR)

garyrussell: Right - I just had the same thought and was going to comment here - we need to add some kind of hook to exit early if the container is paused; checking the paused state should work. 

But I agree, it probably should be a different issue/PR.

We might want to consider it also for when individual partitions are paused (and the retrying batch contains records from those partitions) - but maybe that's a stretch too far.

garyrussell:
@antonio-tomac I assume that you are aware that the preferred method of error handling in batch listeners is for the listener to throw a `BatchListenerFailedException` to tell the framework which record in the batch had the failure?

The `FallbackBatchErrorHandler` is called for all other exceptions; it is required because there is no guarantee that the batch will be the same after a seek and re-poll.

Of course, you might not have a choice if, for example, you are doing a batch insert into a DB from the listener.

antonio-tomac:
Thank you for fast response, I'm glad that this fix got accepted like last minute to 3.0.9 release, as you said due Monday.

spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
garyrussell: @antonio-tomac I created an issue for this: https://github.com/spring-projects/spring-kafka/issues/2742 if you want to take a stab at it, and it is relatively simple, we might be able to squeeze it into Monday's releases.

PR #2743, opened 2023-07-14
GH-2742: Exit batch retries when container is paused

### LINKED ISSUE #2742: FallbackBatchErrorHandler - Consider Seeking and Exiting Retries When Container is Paused
Currently, the FBEH retries internally (pausing and polling the consumer) until retries are exhausted.

There is a use case where the user might want to pause deliveries during the retry attempts, by pausing the container.

It should be possible to detect that a container pause has been requested, seek the batch records and exit the retry loop under this condition.

See the discussion here: https://github.com/spring-projects/spring-kafka/pull/2739#discussion_r1261772928

antonio-tomac:
Closes #2742 

Implementation is aligned with idea exit retrying as soon as possible if any of is true:
 - container is not running anymore
 - container is requested to pause
 - retrying batch which has record from partition that container is requested to pause

Why I extracted ?
```java
static Runnable NO_OP = () -> {};
```
 - To be able to verify arguments
 - Named field stating intention

Why did I introduced new method?
```java
public static void stoppableSleep(Supplier<Boolean> shouldSleepCondition, long interval) throws InterruptedException
```
 - At first it was tempting just to add `!container.isRunning() || container.isPauseRequested()` but not having access to retrying `ConsumerRecords` to check paused partitions
 - Changing behaviour of public api seems like bad idea, and I believe banned for patch (or even minor) versions

Regarding execution duration  assertions
 - Initially, I wrote something like 
   ```java
   assertThat(duration).isBetween(100L, 150L);
    ```
 - But, such assertions could become flaky tests when asking sleep for "precise" timing

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
@@ -132,7 +132,7 @@ public final class ErrorHandlingUtils {
 					throw new KafkaException("Woken up during retry", logLevel, we);
 				}
 				try {
-					ListenerUtils.stoppableSleep(
+					ListenerUtils.conditionalSleep(
 							() -> container.isRunning() &&
 									!container.isPauseRequested() &&
 									records.partitions().stream().noneMatch(container::isPartitionPauseRequested),
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
@@ -45,7 +45,7 @@ import org.springframework.util.backoff.BackOffExecution;
  */
 public final class ErrorHandlingUtils {
 
-	static Runnable NO_OP = () -> {};
+	static Runnable NO_OP = () -> { };
 
 	private ErrorHandlingUtils() {
 	}
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
@@ -45,6 +45,8 @@ import org.springframework.util.backoff.BackOffExecution;
  */
 public final class ErrorHandlingUtils {
 
+	static Runnable NO_OP = () -> {};
+
 	private ErrorHandlingUtils() {
 	}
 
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
```diff
@@ -126,25 +128,35 @@ public static void retryBatch(Exception thrownException, ConsumerRecords<?, ?> r
...
 					Thread.currentThread().interrupt();
-					seeker.handleBatch(thrownException, records, consumer, container, () -> { });
+					seeker.handleBatch(thrownException, records, consumer, container, NO_OP);
 					throw new KafkaException("Interrupted during retry", logLevel, e1);
 				}
 				if (!container.isRunning()) {
 					throw new KafkaException("Container stopped during retries");
 				}
+				if (container.isPauseRequested() ||
+						records.partitions().stream().anyMatch(container::isPartitionPauseRequested)) {
```
garyrussell, line 151: Perhaps traversing the records twice per iteration is a bit much (especially with large `max.poll.records`), given that user pausing individual partitions is a corner case (I am only aware of one user that uses this feature - it was added for non-blocking retries, which is not supported for batch listeners).

Maybe we could add `isAnyPartitionPauseRequested()` to the container and just use that here instead?

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerUtils.java
```diff
@@ -185,11 +187,22 @@ public static void unrecoverableBackOff(BackOff backOff, ThreadLocal<BackOffExec
...
 	public static void stoppableSleep(MessageListenerContainer container, long interval) throws InterruptedException {
+		stoppableSleep(container::isRunning, interval);
+	}
+
+	/**
+	 * Sleep for the desired timeout, as long as shouldSleepCondition supplies true.
+	 * @param shouldSleepCondition to.
+	 * @param interval the timeout.
+	 * @throws InterruptedException if the thread is interrupted.
+	 * @since 3.0.9
```
garyrussell, line 198: @antonio-tomac Are you using 3.0.x? If so, we won't need to back port this feature. If you are still using 2.9.x, we should make the since 2.9.10.

garyrussell (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
garyrussell: Oops - ignore this; we are not traversing the records, just the partitions.

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerUtils.java
antonio-tomac: using 3.0.7 but plan to upgrade to 3.0.9 ASAP

antonio-tomac: If you want backport, I can put since 2.9.10, but I don't need backport myself

spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
antonio-tomac: I wouldn't expect that set of TopicPartitions will be that gigantic. Even though max.poll.records could be 500k, all of them could be from dozen partitions.

>  Maybe we could add `isAnyPartitionPauseRequested()` to the container 

Hmmm, feels wrong to rely on it. Because, it could be that some other partition is paused, which none of records ate from that partition.

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerUtils.java
garyrussell: No; if you don't need it; I'd rather not back port.

garyrussell (approved):

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerUtils.java
```diff
@@ -185,11 +187,22 @@ public static void unrecoverableBackOff(BackOff backOff, ThreadLocal<BackOffExec
...
+	 */
+	public static void stoppableSleep(Supplier<Boolean> shouldSleepCondition, long interval) throws InterruptedException {
```
garyrussell, line 200: Perhaps rename to `conditionalSleep()` ?

antonio-tomac: yeah, agree

PR #2756, opened 2023-07-26
Refactor DeserializationException Detection Code

garyrussell:
- move to `ErrorHandlingUtils` for reuse.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
@@ -18,19 +18,27 @@ package org.springframework.kafka.listener;
 
 import java.time.Duration;
 import java.util.List;
+import java.util.Map;
+import java.util.Properties;
 import java.util.Set;
 import java.util.function.BiConsumer;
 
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
```diff
@@ -256,4 +264,66 @@ public static Exception findRootCause(Exception exception) {
...
+	 * @param consumerFactory the consumer factory.
+	 * @param consumerOverrides consumer factory property overrides.
+	 * @param isValue true to find the value deserializer.
+	 * @param applicationContext the application context.
+	 * @return true if the deserializer is an instance of
+	 * {@link ErrorHandlingDeserializer}.
+	 * @since 3.0.10
+	 */
+	public static <K, V> boolean checkDeserializer(ConsumerFactory<K, V> consumerFactory,
+			Properties consumerOverrides, boolean isValue, @Nullable ApplicationContext applicationContext) {
```
artembilan, line 281: I don't have an application context in my hands, but I do configure a `ClassLoader`.
How about to change it to that one instead, please?

artembilan (changes requested):

Issue #2762, opened 2023-08-02
AdminClient doesn't use configured properties while starting an container listener


zinnchen:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**
2.9.7
**Describe the bug**
AdminClient in class AbstractMessageListenerContainer#checkTopics doesn't use the **container** properties for connection to Kafka. It uses the properties from consumerFactory instead which are not in sync.

**To Reproduce**
- configure and create an instance of ConcurrentKafkaListenerContainerFactory with container-property "MissingTopicsFatal" set to true
- create a container using the container factory from above
- configure container properties with some SSL settings needed for connecting to Kafka
- start container
- "SslAuthenticationException: SSL handshake failed" Exception is thrown (no matter if topic exists or not)

If I set the "MissingTopicsFatal" set to false, all is working fine.

After debugging I noticed, that the checkTopics method in AbstractMessageListenerContainer class uses the AdminClient for checking for topic existence. This admin client uses the connection properties from consumer factory. These properties are not in sync with the container properties in container.

**Expected behavior**
starting a listener container with "MissingTopicsFatal" property set to true in container will check for topic existence and fail if topic doesn't exist USING configured connection parameters from container.

throwing following exception if topic doesn't exist:
`java.lang.IllegalStateException: Topic(s) [my-topic] is/are not present and missingTopicsFatal is true`

**Workaround**
Set same properties for kafka connection on consumer factory AND container.

**Sample**
```
  private void test(KafkaProperties kafkaProperties) {

    Map<String, Object> configs = kafkaProperties.buildConsumerProperties();
    var consumerFactory = new DefaultKafkaConsumerFactory<>(configs);

    var listenerContainerFactory = new ConcurrentKafkaListenerContainerFactory<>();
    listenerContainerFactory.setConsumerFactory(consumerFactory);
    listenerContainerFactory.setAutoStartup(Boolean.FALSE);
    listenerContainerFactory.setMissingTopicsFatal(Boolean.TRUE);
    Map<String, Object> connectionConfig = new HashMap<>();
    connectionConfig.put(SslConfigs.SSL_KEYSTORE_TYPE_CONFIG, "PEM");
    connectionConfig.put(SslConfigs.SSL_TRUSTSTORE_TYPE_CONFIG, "PEM");
    connectionConfig.put(SslConfigs.SSL_KEYSTORE_CERTIFICATE_CHAIN_CONFIG, "cert-chain");
[...]

PR #2763, opened 2023-08-02
GH-2762: Fix Topic Check with Consumer Overrides

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2762

When checking for missing topics, the admin was created with the raw consumer factory properties and did not apply any overriding consumer properties in the `ContainerProperties`.

Apply the overrides before creating the admin.

Pull up a method that merges default properties (if any) because we have to iterate over the hash table because the user might have used `put()` instead of `setProperty()` to set properties.

**cherry-pick to 2.9.x**

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/MissingTopicCheckOverrideAdminConfig.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/MissingTopicCheckOverrideAdminConfig.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.listener;
+
+import static org.assertj.core.api.Assertions.assertThatNoException;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.never;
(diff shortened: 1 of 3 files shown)


spring-kafka/src/test/java/org/springframework/kafka/listener/MissingTopicCheckOverrideAdminConfig.java
```diff
@@ -0,0 +1,101 @@
...
+import org.springframework.kafka.test.context.EmbeddedKafka;
+import org.springframework.kafka.test.utils.KafkaTestUtils;
+
+/**
+ * @author Gary Russell
+ * @since 3.0
+ *
+ */
+@EmbeddedKafka(topics = "mtccac")
+public class MissingTopicCheckOverrideAdminConfig {
```
artembilan, line 46: Is it intentional do not use a `*Tests` suffix?

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/listener/MissingTopicCheckOverrideAdminConfigTests.java
```diff
@@ -0,0 +1,101 @@
...
+import org.springframework.kafka.test.context.EmbeddedKafka;
+import org.springframework.kafka.test.utils.KafkaTestUtils;
+
+/**
+ * @author Gary Russell
+ * @since 3.0
+ *
+ */
+@EmbeddedKafka(topics = "mtccac")
+public class MissingTopicCheckOverrideAdminConfig {
```
artembilan, line 46:
```
 /home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/test/java/org/springframework/kafka/listener/MissingTopicCheckOverrideAdminConfigTests.java:46: error: class MissingTopicCheckOverrideAdminConfig is public, should be declared in a file named MissingTopicCheckOverrideAdminConfig.java
public class MissingTopicCheckOverrideAdminConfig {
```

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/listener/MissingTopicCheckOverrideAdminConfigTests.java
garyrussell: That's weird; it is correct locally.

artembilan:

PR #2770, opened 2023-08-07
Private Header Type for DeserializationExceptions

garyrussell:
Use a package-private header for deserialization exceptions.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DeserializationExceptionHeader.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DeserializationExceptionHeader.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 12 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DeserializationExceptionHeader.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.apache.kafka.common.header.internals.RecordHeader;
+
+/**
+ * A package-protected header used to contain serialized
+ * {@link DeserializationException}s. Only headers of this type will be examined for
+ * deserialization.
+ *
+ * @author Gary Russell
+ * @since 2.9.11
```
artembilan, line 27: I will back-port here, but I won't back to `2.8.x` until your confirmation.

garyrussell: Yeah; let's just leave it at 2.9.x for now; I thought about 2.8.x because it is still in commercial support (2.7.x goes out soon) and it might make it easier if we are asked later to release it. But I don't think it will make much difference doing it now or then.

artembilan:
... and cherry-picked to `2.9.x` after resolving conflicts.

Issue #2779, opened 2023-08-22
ParseStringDeserializer throws NullPointerException when message is null


lagweasel:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.0.7

**Describe the bug**

When deserialize() method is called with a `null` (bytes) message a `NullPointerException` is thrown. This is thrown from the line that attempts to create a new String from the bytes:
```
	@Override
	public T deserialize(String topic, Headers headers, byte[] data) {
		return this.parser.apply(new String(data, this.charset), headers);
	}
```

**To Reproduce**

Publish a `null` value to a kafka topic where the consumer uses ParseStringDeserializer for deserialization (e.g. via ToFromStringSerde).

**Expected behavior**

Underlying 'parser' function expected to be invoked with `null` String value. This expectation is supported by the javadoc on Deserializer interface (that ParseStringDeserializer implements).

```
    /**
     * Deserialize a record value from a byte array into a value or object.
     * @param topic topic associated with the data
     * @param headers headers associated with the record; may be empty.
     * @param data serialized bytes; may be null; implementations are recommended to handle null by returning a value or null rather than throwing an exception.
     * @return deserialized typed data; may be null
     */
    default T deserialize(String topic, Headers headers, byte[] data) {
        return deserialize(topic, data);
    }
```

garyrussell:
Thanks; a simple work around is to subclass the deserializer and call the super method if the value is not `null`.

lagweasel:
I notice that the fix has:
```
	public T deserialize(String topic, Headers headers, byte[] data) {
		if (data == null) {
			return null;
		}
		return this.parser.apply(new String(data, this.charset), headers);
	}
```

But wondering whether it should instead invoke the parser passing a null value instead. (There may be some reason for the parser to handle null and return something different?) e.g.
```
	public T deserialize(String topic, Headers headers, byte[] data) {
                String stringData = (data == null ? null : new String(data, this.charset));
		return this.parser.apply(stringData, headers);
	}
```

artembilan:
... and cherry-picked to `2.9.x`

PR #2781, opened 2023-08-23
GH-2779: ParseStringDeserializer: Add Null Check

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2779

Handle tombstone records.

**cherry-pick to 2.9.x**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/ParseStringDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/ParseStringDeserializer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 the original author or authors.
+ * Copyright 2016-2023 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


lagweasel:
I notice that the fix has:
```
	public T deserialize(String topic, Headers headers, byte[] data) {
		if (data == null) {
			return null;
		}
		return this.parser.apply(new String(data, this.charset), headers);
	}
```

But wondering whether it should instead invoke the parser passing a null value instead. (There may be some reason for the parser to handle null and return something different?) e.g.
```
	public T deserialize(String topic, Headers headers, byte[] data) {
                String stringData = (data == null ? null : new String(data, this.charset));
		return this.parser.apply(stringData, headers);
	}
```

garyrussell:
Agreed, will update soon.

garyrussell:
Agreed, will update soon, but the “something different” must still be a T.

PR #2786, opened 2023-08-26
add the check to reduce sync block

wizard4wu:
add the check to reduce sync block, just like double check in Singleton 

Can you help me to review the code? Thanks. @garyrussell

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
@@ -782,6 +782,10 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 		if (this.producerPerThread) {
 			return getOrCreateThreadBoundProducer();
 		}
+		//add the check to reduce sync block, just like double check in Singleton
+		if(this.producer != null && !this.producer.closed && !expireByTime(this.producer)){
+			return producer;
+		}
 		synchronized (this) {
 			if (this.producer != null && this.producer.closed) {
 				this.producer.closeDelegate(this.physicalCloseTimeout, this.listeners);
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -782,6 +782,10 @@ private Producer<K, V> doCreateProducer(@Nullable String txIdPrefix) {
 		if (this.producerPerThread) {
 			return getOrCreateThreadBoundProducer();
 		}
+		//add the check to reduce sync block, just like double check in Singleton
+		if(this.producer != null && !this.producer.closed && !expireByTime(this.producer)){
+			return producer;
+		}
```
garyrussell, line 788: No; this is not valid; there is a race condition between the null check and the returned value.

garyrussell (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
wizard4wu: i think if the produce

garyrussell: Please don't consider using ChatGPT for writing code.

>`this.producer != null `

Another thread could set the producer to `null` between this test and the `return`. Or, worse, it could be set to null between that test and `this.producer.closed` which would cause an NPE around this code.

Hence the need for `synchronization`.

wizard4wu: @garyrussell
Thank you very much.
Understand. 
So this is different from the singleton mode of the double-check lock. The singleton mode only judges whether the object is empty. Here, it not only judges whether the object is empty but also judges the properties of the object.

garyrussell: Exactly.

artembilan:
Closed as `Invalid`.
See discussion on related PR: https://github.com/spring-projects/spring-kafka/pull/2786

PR #2793, opened 2023-08-31
Deprecate DestinationTopicPropertiesFactory CTORs

garyrussell:
So they can be removed in 3.1, along with `FixedDelayStrategy`.


artembilan (changes requested):
I believe something is wrong with this PR and probably its branch has been made against `main`, but not `3.0.x`.

Please, look if that can be fixed or just open a new one against expected `upstream`.

Thanks

garyrussell:
Oops - done.

PR #2814, opened 2023-09-20
Trust record class for deserializer

quaff:
It's very common to use record as message, and record class is safe enough to trust, there is no known record-based gadget.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/mapping/DefaultJackson2JavaTypeMapper.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/mapping/DefaultJackson2JavaTypeMapper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2021 the original author or authors.
+ * Copyright 2017-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -39,6 +39,7 @@ import com.fasterxml.jackson.databind.type.TypeFactory;
  * @author Andreas Asplund
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Yanming Zhou
  *
  * @since 2.1
  */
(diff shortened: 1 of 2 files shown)


artembilan:
Thank for looking into this, but:

> there is no known record-based gadget.

I don't think that we really talk here about some structure type. 
We really talk about classes we can trust or not.
Even `record` can be implement the way that it is malicious.
Do you have some official report to prove otherwise?

Meanwhile I'd suggest to stick with the current approach where even `record` class name should be included into trusted packages: kinda maximum security.

quaff:
Configuring trusted packages is very annoying and not refactor (renaming package) friendly.
All known gadgets from [Jackson](https://github.com/FasterXML/jackson-databind/blob/5207dedf047c972aae8b6e487766834ce5e2219a/src/main/java/com/fasterxml/jackson/databind/jsontype/impl/SubTypeValidator.java) are vulnerable by by accident, record is data carrier and Jackson will invoke it's constructor only, It's safe enough, unless someone write malicious code in constructor intentionally, then it should be guarded at another level not here.
/CC @cowtowncoder, Sorry to bother you, could you clarify that record class is safe for deserialization or not?

cowtowncoder:
@quaff I am not aware of any Record types used as Gadgets; not seen vulnerabilities (CVEs) reported either.
For what that is worth.

Wrt Jackson big question is that of using (or, ideally, not :) ) of polymorphic deserialization so unless that is enabled (and user configures lax PolymorphicTypeValidator) attackers cannot really trigger gadget behavior using known types.

quaff:
> @quaff I am not aware of any Record types used as Gadgets; not seen vulnerabilities (CVEs) reported either. For what that is worth.
> 
> Wrt Jackson big question is that of using (or, ideally, not :) ) of polymorphic deserialization so unless that is enabled (and user configures lax PolymorphicTypeValidator) attackers cannot really trigger gadget behavior using known types.

According to the most authoritative clarification, I think mandatory trusted packages is excessive restriction, It should be `*` by default, let users configure their own if them think it's not safe, or step back all record classes should be trusted.

artembilan:
>  It's safe enough, unless someone write malicious code in constructor intentionally, then it should be guarded at another level not here.

I don't see a difference in `record` and regular Java class instantiations, Jackson or any other serialization mechanism.
Even if there was not published any CVE specific for `record` type, it might be because from Java perspective it is still a class with all the possible deserialization gadgets.

Spring policy is to secure by default and trust no one.
Therefore your `*` is not valid suggestion.
I even believe if Jackson would behave same way (trust no one), we wouldn't have our own solution like this one.
And we all simply would rely on a configuration from Jackson.

But that's fully different story and I'd prefer to stick with a solution where I would not deal with headache about fixing new vulnerability I've just introduced.

And apparently it is not so hard to configure `*` in your project if you really don't use polymorphic deserialization in Jackson.

artembilan:
After some team discussion we decided to stay conservative and still trust no one.
So, I'm sorry but we have to close this as not acceptable change.

Thanks for understanding.

Issue #2391, opened 2022-09-10
Support KRaft mode for EmbeddedKafkaBroker


vladimirfx:
**Expected Behavior**

The embedded Kafka broker should be able to run in KRaft mode.

**Current Behavior**

There is no option to run Kafka in KRaft mode nor to completely disable Zookeeper.

**Context**

KRaft mode becomes production ready from Kafka 3.3.0 and only available mode from Kafka 4+. So support for KRaft is becoming a requirement in 2023. But it is relevant just now for EmbeddedKafkaBroker because it is relative simplicity to run and minimize footprint by removing ZK.

artembilan:
Feel free to contribute this feature.
Or let's wait until it really becomes and official feature (3.3.0 GA), so we would be good to drop any ZK support we have in the embedded broker so far.

garyrussell:
We are now on Kafka 3.3.1 so this can be implemented; contributions are welcome.

govi20:
If no one is working on it, I would like to give it a shot.

artembilan:
@govi20 ,

no one as far as I am aware.

Thank you for help proposal!

eddumelendez:
Hi 👋🏽, sorry for hijacking the thread but I think this can be useful information. Testcontainers Kafka module version 1.18.0 offers support for Kraft mode. Look at this test

https://github.com/testcontainers/testcontainers-java/blob/main/modules/kafka/src/test/java/org/testcontainers/containers/KafkaContainerTest.java#L136-L145

garyrussell:
I also found this https://github.com/apache/kafka/pull/13375 which could be used as a basis for the changes needed here.

garyrussell:
I made some progress, but this is blocked by https://issues.apache.org/jira/projects/KAFKA/issues/KAFKA-15092

garyrussell:
Unblocked; the necessary class is in server-common...-test.jar.

PR #2820, opened 2023-09-21
GH-2391: Add Kraft Version of EmbeddedKafkaBroker

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2391

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
@@ -3,6 +3,16 @@
 
 The `spring-kafka-test` jar contains some useful utilities to assist with testing your applications.
 
+[[ekb]]
+== EmbeddedKafkaBroker
+
+Two implementations are provided:
+
+* `EmbeddedKafkaZKBroker` - legacy implementation which starts an embedded `Zookeeper` instance.
+* `EmbeddedKafkaKraftBroker` - uses Kraft instead of `Zookeeper` in combinder controller and broker modes (since 3.6)
+
+There are several techniques to configure the broker as discussed in the following sections.
+
 [[ktu]]
 == KafkaTestUtils
 
@@ -47,7 +57,7 @@ If this is not possible for some reason, note that the `consumeFromEmbeddedTopic
 Since it does not have access to the consumer properties, you must use the overloaded method that takes a `seekToEnd` boolean parameter to seek to the end instead of the beginning.
 ====
 
-A JUnit 4 `@Rule` wrapper for the `EmbeddedKafkaBroker` is provided to create an embedded Kafka and an embedded Zookeeper server.
+A JUnit 4 `@Rule` wrapper for the `EmbeddedKafkaZKBroker` is provided to create an embedded Kafka and an embedded Zookeeper server.
 (See xref:testing.adoc#embedded-kafka-annotation[@EmbeddedKafka Annotation] for information about using `@EmbeddedKafka` with JUnit 5).
 The following listing shows the signatures of those methods:
 
@@ -71,6 +81,8 @@ public EmbeddedKafkaRule(int count, boolean controlledShutdown, String... topics
  */
 public EmbeddedKafkaRule(int count, boolean controlledShutdown, int partitions, String... topics) { ... }
 ----
+
+NOTE: The `EmbeddedKafkaKraftBroker` is not supported with JUnit4.
 
 The `EmbeddedKafkaBroker` class has a utility method that lets you consume for all the topics it created.
 The following example shows how to use it:
@@ -162,7 +174,7 @@ You can use the same broker for multiple test classes with something similar to 
 ----
 public final class EmbeddedKafkaHolder {
 
-    private static EmbeddedKafkaBroker embeddedKafka = new EmbeddedKafkaBroker(1, false)
+    private static EmbeddedKafkaBroker embeddedKafka = new EmbeddedKafkaZKBroker(1, false)
             .brokerListProperty("spring.kafka.bootstrap-servers");
 
     private static boolean started;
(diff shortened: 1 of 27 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -12,3 +12,8 @@ For changes in earlier version, see xref:appendix/change-history.adoc[Change His
 
 This version requires the 3.5.1 `kafka-clients`.
 
+[[x30-ekb]]
```
artembilan, line 15: `x31`?

spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
```diff
@@ -0,0 +1,593 @@
...
+			}
+		}
+		else {
+			throw new IllegalStateException("Failed to be assigned partitions from the embedded topics");
+		}
+		LOGGER.debug("Subscription Initiated");
+	}
+
+	@Override
+	public void setZkPort(int zkPort) {
```
artembilan, line 576: I don't think this method belongs to this class.
So, probably just bite a bullet and remove it from the interface as well?

spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaZKBroker.java
```diff
@@ -0,0 +1,861 @@
...
+ * @author Gary Russell
+ * @author Kamill Sokol
+ * @author Elliot Kennedy
+ * @author Nakul Mishra
+ * @author Pawel Lozinski
+ * @author Adrian Chlebosz
+ *
+ * @since 2.2
+ */
+public class EmbeddedKafkaZKBroker implements EmbeddedKafkaBroker, InitializingBean, DisposableBean {
```
artembilan, line 94: Looks like those `InitializingBean` & `DisposableBean` are on the `EmbeddedKafkaBroker` interface...

spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
```diff
@@ -0,0 +1,593 @@
...
+	 */
+	public void setAdminTimeout(int adminTimeout) {
+		this.adminTimeout = Duration.ofSeconds(adminTimeout);
+	}
+
+	@Override
+	public void afterPropertiesSet() {
+		overrideExitMethods();
+		addDefaultBrokerPropsIfAbsent(this.brokerProperties, this.count);
+		start();
```
artembilan, line 200: If you start it from the `afterPropertiesSet()` why do we need a `SmartLifecycle` contract then?
Why cannot we stop everything in the `destroy()` as it is for ZK variant?

spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
```diff
@@ -0,0 +1,593 @@
...
+		try {
+			KafkaClusterTestKit.Builder clusterBuilder = new KafkaClusterTestKit.Builder(
+					new TestKitNodes.Builder()
+							.setCombined(true)
+							.setNumBrokerNodes(this.count)
+							.setNumControllerNodes(this.count)
+							.build());
+			this.brokerProperties.forEach((k, v) -> clusterBuilder.setConfigProp((String) k, (String) v));
+			this.cluster = clusterBuilder.build();
+//			cluster.nonFatalFaultHandler().setIgnore(true);
```
artembilan, line 218: Do you have anything else in mind for this line of code?

spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
```diff
@@ -181,5 +182,12 @@
 	 */
 	int adminTimeout() default EmbeddedKafkaBroker.DEFAULT_ADMIN_TIMEOUT;
 
+	/**
+	 * Use KRaft instead of Zookeeper.
+	 * @return whether to use KRaft.
+	 * @since 3.6
+	 */
+	boolean kraft() default false;
```
artembilan, line 190: Why cannot we use Kraft by default?
Isn't it faster?

spring-kafka-test/src/test/java/org/springframework/kafka/test/EmbeddedKafkaKraftBrokerTests.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * @author Gary Russell
+ * @since 3.1
+ *
+ */
+public class EmbeddedKafkaKraftBrokerTests {
+
+	@Test
+	void testUpDown() {
+		LogFactory.getLog(getClass()).info("foo");
```
artembilan, line 31: Not sure in this log message...

spring-kafka-test/src/test/java/org/springframework/kafka/test/EmbeddedKafkaKraftBrokerTests.java
```diff
@@ -0,0 +1,38 @@
...
+		EmbeddedKafkaKraftBroker kafka = new EmbeddedKafkaKraftBroker(1, 1, "topic1");
+		kafka.afterPropertiesSet();
+		kafka.start();
+		kafka.stop();
```
artembilan, line 35: Doesn't look like this test verifies anything.
Or you just haven't finished here yet.

artembilan (changes requested):
Looks great!
But there are some questions.
Thanks

spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
```diff
@@ -9,7 +9,7 @@ The `spring-kafka-test` jar contains some useful utilities to assist with testin
 Two implementations are provided:
 
 * `EmbeddedKafkaZKBroker` - legacy implementation which starts an embedded `Zookeeper` instance.
-* `EmbeddedKafkaKraftBroker` - uses Kraft instead of `Zookeeper` in combinder controller and broker modes (since 3.6)
+* `EmbeddedKafkaKraftBroker` - (default) uses Kraft instead of `Zookeeper` in combind controller and broker modes (since 3.1)
```
artembilan, line 12: I understand that `combinder` was a typo before, but this `combind` still looks like a typo.
May you mean this:
```suggestion
* `EmbeddedKafkaKraftBroker` - (default) uses Kraft instead of `Zookeeper` in combined controller and broker modes (since 3.1)
```
?

artembilan (changes requested):
OK. Just one time.

I guess even if you use that `kraft = false` in the modified tests, many others are in default KRaft mode anyway.

vladimirfx:
Thanks

PR #2821, opened 2023-09-21
Add Search in all Spring Docs

garyrussell:


garyrussell:
@rwinch is this ok? Thanks.

garyrussell:
@rwinch Sorry to bug you, but our RC is due next week; thanks.

rwinch:
@garyrussell It looks good to me!

PS: Keep in mind that in Antora we keep building the docs (even releases) so even if you released RC before merging this, you would still be able to get this into the site for the RC. This is because docs-build is versionless (pulls in the antora ui, extensions, etc that are on docs-build) which pulls in the content of the tag (the actual documentation doesn't change).

PR #2822, opened 2023-09-21
Get rid of proxy in DefaultKafkaConsumerFactory

artembilan:
The proxy in the `DefaultKafkaConsumerFactory` for `KafkaConsumer` is created only to intercept `close()` call to remove an instance from the `listeners`.
There is no need in such a proxy since simple `KafkaConsumer` class extension can handle that scenario.

The reason behind this change is to avoid a `Serializable` (`java.lang.reflect.Proxy`) header in the produced message for the `KafkaConsumer` where we still fail to serialize it because other properties of the proxy are not `Serializable`

* Introduce `DefaultKafkaConsumerFactory.ExtendedKafkaConsumer` to handle `listeners` interaction
* The `createRawConsumer()` might be considered as breaking change since now end-user must extend this `ExtendedKafkaConsumer` to be able to handle `listeners` same way as before
* Adjust `DefaultKafkaConsumerFactoryTests.listener()` test for the current code state

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2022 the original author or authors.
+ * Copyright 2016-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package org.springframework.kafka.core;
 
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
@@ -28,19 +29,13 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
 
-import org.aopalliance.aop.Advice;
-import org.aopalliance.intercept.MethodInterceptor;
-import org.aopalliance.intercept.MethodInvocation;
 import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
-import org.apache.kafka.common.Metric;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.serialization.Deserializer;
 
-import org.springframework.aop.framework.ProxyFactory;
-import org.springframework.aop.support.NameMatchMethodPointcutAdvisor;
 import org.springframework.beans.factory.BeanNameAware;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.lang.Nullable;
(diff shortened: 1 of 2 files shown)


sobychacko (approved):

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
```diff
@@ -445,68 +440,68 @@ private void checkInaccessible(Properties properties, Map<String, Object> modifi
 		}
 	}
 
-	@SuppressWarnings("resource")
 	protected Consumer<K, V> createKafkaConsumer(Map<String, Object> configProps) {
 		checkBootstrap(configProps);
 		Consumer<K, V> kafkaConsumer = createRawConsumer(configProps);
```
garyrussell, line 445: Perhaps we could throw an exception here, if listeners are present and the created consumer is not an instance of `ExtendedKafkaConsumer`?

garyrussell (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
garyrussell: On the other hand, Java doesn't support multiple inheritance (the user might already be extending some other class), so a hard failure would not be so good; perhaps log a warning that listeners must be invoked when closing a custom consumer.

Or, simply revert the code and have `ExtendedKafkaConsumer` wrap the proxy instead so that the consumer is not `Serializable`.

garyrussell: I agree; makes sense; maybe also deprecate that method?

artembilan: I would let it live for now until we really fail for some peculiar use-case to learn from.
If we deprecate and remove it (I guess there was a reason behind that method), we may never learn the mentioned use-case 😉

garyrussell (approved):

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java
```diff
@@ -445,68 +440,73 @@ private void checkInaccessible(Properties properties, Map<String, Object> modifi
...
-			}
-			String id = this.beanName + "." + clientId;
-			kafkaConsumer = createProxy(kafkaConsumer, id);
-			for (Listener<K, V> listener : this.listeners) {
-				listener.consumerAdded(id, kafkaConsumer);
-			}
+		if (!this.listeners.isEmpty() && !(kafkaConsumer instanceof ExtendedKafkaConsumer)) {
+			LOGGER.warn("The 'ConsumerFactory.Listener' configuration is ignored " +
+					"because the consumer is not an instance of 'ExtendedKafkaConsumer'." +
+					"Consider to extend 'ExtendedKafkaConsumer' or implement your own 'ConsumerFactory'.");
```
garyrussell, line 449:
```suggestion
					"Consider extending 'ExtendedKafkaConsumer' or implement your own 'ConsumerFactory'.");
```

Issue #2826, opened 2023-10-06
CommonDelegatingErrorHandler: Missing updateClassifier() in addDelegate()?


remcobIT:
While creating a CommonDelegatingErrorHandler I ran (probably) into a bug. I used addDelegate() to add a different errorhandler in case of a specific exception in the cause chain (I set causeChainTraversing to true), but the default errorhandler was triggered instead.

I came to know by checking this source code that in addDelegate() the private method updateClassifier() isn't called. So I used the method setErrorHandlers() instead and now it is working.

https://github.com/spring-projects/spring-kafka/blob/638ee013e1b742c1e0af21b565f1e49af94932da/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonDelegatingErrorHandler.java#L121C16-L121C16

Is this a bug in addDelegate()?

PR #2829, opened 2023-10-09
GH-2826: Fix CommonDelegatingErrorHandler

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2826

The `addDelegate` method did not update the classifier, so it failed to work when cause chain traversal is enabled.

**cherry-pick to 3.0.x, 2.9.x**


artembilan:
... and cherry-picked to `3.0.x`.
Back-ported to `2.9.x` as https://github.com/spring-projects/spring-kafka/commit/fda59f3a87cf53168ebe70449ee1cec184d2d464 after migrating back to Java 8 API for `Map.Entry`.

PR #2843, opened 2023-10-15
Upgrade org.apache.kafka:kafka-clients to 3.6.0

MarcusAdriano:
Upgrade org.apache.kafka:kafka-clients to 3.6.0 to fix [CVE-2023-43642](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-43642) for Spring-Boot 2.7.X.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,7 +59,7 @@ ext {
 	jaywayJsonPathVersion = '2.6.0'
 	junit4Version = '4.13.2'
 	junitJupiterVersion = '5.8.2'
-	kafkaVersion = '3.2.3'
+	kafkaVersion = '3.6.0'
 	log4jVersion = '2.17.2'
 	micrometerVersion = '1.9.14'
 	mockitoVersion = '4.5.1'


garyrussell:
We cannot upgrade to a new major/minor version of a dependency in a patch release.

The upcoming 3.1 release uses the 3.6.0 clients natively.

However, 2.9.x is fully compatible with the 3.6.x clients; we explain how to override the version prescribed by Boot here https://docs.spring.io/spring-kafka/docs/2.9.12/reference/html/#update-deps (when using Boot, the version in this project's dependencies is ignored; Boot dependencies rule).

If you are using the embedded Kafka broker then, with Boot versions less than 3.2.x, you also have to add the following to your POM...

```xml
<dependency>
	<groupId>org.apache.kafka</groupId>
	<artifactId>kafka-server-common</artifactId>
	<classifier>test</classifier>
	<scope>test</scope>
	<version>3.6.0</version>
</dependency>
```

This is because some classes were moved there in 3.6.0.

If you are using

PR #2858, opened 2023-10-24
GH-2731: fix count and time ack mode

### LINKED ISSUE #2731: AckCount is invalid when ackMode is Count and ackCount is less than maxPollRecords
**Expected Behavior**

Assume we have following settings:

application.properties
```css
spring.kafka.listener.ack-mode=count
spring.kafka.listener.ack-count=100
logging.level.org.springframework.kafka=debug
```

Expect to commit offset every 100 messages processed

SpringBoot Application:
```java
@RestController
@SpringBootApplication
public class SpringKafkaDemo2Application {

        SpringApplication.run(SpringKafkaDemo2Application.class, args);
    }

    @KafkaListener(id = "fooGroup", topics = "topic")
    public void listen(ConsumerRecord<String, String> record) throws Exception {
    }

    @Autowired
    private KafkaTemplate<Object, Object> template;

    @PostMapping(path = "/send/{what}")
    public void sendFoo(@PathVariable String what) {
        for (int i = 0; i < 500; i++) {
            this.template.send("topic", "message" + i);
        }
    }

}
```

**Current Behavior**

Committing the offset every time after 500 records processed:
```css
2023-07-06T18:07:36.680+08:00 DEBUG 26195 --- [ fooGroup-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Committing in COUNT because count 500 exceeds configured limit of 100
2023-07-06T18:07:36.680+08:00 DEBUG 26195 --- [ fooGroup-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Commit list: {topic-0=OffsetAndMetadata{offset=560, leaderEpoch=null, metadata=''}}
2023-07-06T18:07:36.680+08:00 DEBUG 26195 --- [ fooGroup-0-C-1] o.s.k.l.KafkaMessageListenerContainer    : Committing: {topic-0=OffsetAndMetadata{offset=560, leaderEpoch=null, metadata=''}}
```

**Context**
Although the larger the ackCount, the better the performance, but it will increase the chance of retransmission, so sometimes it is useful for the ackCount to be smaller than maxPollRecords

Wzy19930507:
fixes #2731

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
@@ -3060,12 +3060,12 @@ public class KafkaMessageListenerContainerTests {
 		final Map<TopicPartition, List<ConsumerRecord<Integer, String>>> records2 = new HashMap<>();
 		records2.put(topicPartition, Arrays.asList(
 				new ConsumerRecord<>("foo", 0, 2L, 1, "baz"),
-				new ConsumerRecord<>("foo", 0, 3L, 1, "qux"))); // commit (4 >= 3)
+				new ConsumerRecord<>("foo", 0, 3L, 1, "qux"))); // commit (3 = 3)
 		final Map<TopicPartition, List<ConsumerRecord<Integer, String>>> records3 = new HashMap<>();
 		records3.put(topicPartition, Arrays.asList(
 				new ConsumerRecord<>("foo", 0, 4L, 1, "fiz"),
-				new ConsumerRecord<>("foo", 0, 5L, 1, "buz"),
-				new ConsumerRecord<>("foo", 0, 6L, 1, "bif"))); // commit (3 >= 3)
+				new ConsumerRecord<>("foo", 0, 5L, 1, "buz"), // commit (3 = 3)
+				new ConsumerRecord<>("foo", 0, 6L, 1, "bif"))); // commit (1 when next poll returns no records)
 		ConsumerRecords<Integer, String> consumerRecords1 = new ConsumerRecords<>(records1);
 		ConsumerRecords<Integer, String> consumerRecords2 = new ConsumerRecords<>(records2);
 		ConsumerRecords<Integer, String> consumerRecords3 = new ConsumerRecords<>(records3);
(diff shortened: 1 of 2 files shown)


spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -3112,9 +3112,9 @@ public void testAckModeCount() throws Exception {
 		container.start();
 		assertThat(commitLatch.await(10, TimeUnit.SECONDS)).isTrue();
 		assertThat(recordCount.get()).isEqualTo(7);
-		verify(consumer).commitSync(Collections.singletonMap(topicPartition, new OffsetAndMetadata(4L)),
+		verify(consumer).commitSync(Collections.singletonMap(topicPartition, new OffsetAndMetadata(3L)),
 				Duration.ofSeconds(42));
-		verify(consumer).commitSync(Collections.singletonMap(topicPartition, new OffsetAndMetadata(7L)),
+		verify(consumer).commitSync(Collections.singletonMap(topicPartition, new OffsetAndMetadata(6L)),
```
garyrussell, line 3117: Thanks for the contribution; in general it looks good but, in this case, the 7th record's offset will never be committed; I believe we need to accelerate the commit based on count if the next poll returns no records. This was not needed before this change.

Restoring the
```java
		verify(consumer).commitSync(Collections.singletonMap(topicPartition, new OffsetAndMetadata(7L)),
				Duration.ofSeconds(42));
```
(after the test for offset 6) causes the test to fail.

We also need to adjust the comments on lines 3063, 3068.

garyrussell (changes requested):
Thanks.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
Wzy19930507: Hi, @garyrussell 


**config**
AckMode is AckMode.COUNT
AckCount is 100

**expect action**
consumer.poll()  return 101 record    // commit offset 100th
consumer.poll()  return 0 record  // commit offset 101th
consumer.poll()  return 101 record  // commit offset 201th
consumer.poll()  return 0 record  // commit offset 202th
It looks like the submissions are a little frequent

**Change**
Can I change it here? Avoid commit offset when manually calling `Consumer.wake()`
https://github.com/spring-projects/spring-kafka/blob/ad5e754bf1c0b9c2566fe1ab900d33467ae07dd9/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1414-L1420

garyrussell (approved):

Issue #2859, opened 2023-10-24
Admin is boot up by turning on observability


dogrizz:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.0.10

**Describe the bug**

When enabling observabiity it tries to bring up AdminClient to retrieve clusterId even if adminClient auto-create is set to false. 
When admin connection is blocked on other side it causes timeouts.
With default admin operation timeout  message processing goes to a crawl as it tries to retrieve it on each message.

**To Reproduce**
```

    @Bean
    fun kafkaListenerContainerFactory(consumerFactory: ConsumerFactory<String?, String?>): ConcurrentKafkaListenerContainerFactory<String, String> {
        val factory = ConcurrentKafkaListenerContainerFactory<String, String>()
        factory.containerProperties.isObservationEnabled = true
        factory.consumerFactory = consumerFactory
        return factory
    }

```
**Expected behavior**

I would expect that when adminClient is disabled clusterId will be ignored or at least that we have a possibility to customize:
https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordReceiverContext.java
To not to try reading clusterId.

**Sample**

A link to a GitHub repository with a [minimal, reproducible,  sample](https://stackoverflow.com/help/minimal-reproducible-example).

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

garyrussell:
I suppose we could add `setClusterId` to the `KafkaAdmin` which (when not null) would effectively disable the attempt to fetch the cluster id.

garyrussell:
In the meantime, you can subclass `KafkaAdmin` and override `clusterId()` to return some String.

```
factory.setContainerCustomizer(container -> {
	if (container.getListenerId().equals("obs3")) {
		((AbstractMessageListenerContainer<Integer, String>) container).setKafkaAdmin(myAdminNoClusterFetch);
	}
});
```

PR #2860, opened 2023-10-25
GH-2859: Option to Manually Set the Cluster ID

garyrussell:
Resolves https://github.com/spring-projects/spring-kafka/issues/2859

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/micrometer.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/micrometer.adoc
@@ -109,3 +109,6 @@ Starting with version 3.0.6, you can add dynamic tags to the timers and traces, 
 To do so, add a custom `KafkaListenerObservationConvention` and/or `KafkaTemplateObservationConvention` to the listener container properties or `KafkaTemplate` respectively.
 The `record` property in both observation contexts contains the `ConsumerRecord` or `ProducerRecord` respectively.
 
+The sender and receiver contexts' `remoteServiceName` properties are set to the Kafka `clusterId` property; this is retrieved by a `KafkaAdmin`.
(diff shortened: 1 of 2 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/micrometer.adoc
```diff
@@ -109,3 +109,6 @@ Starting with version 3.0.6, you can add dynamic tags to the timers and traces,
 To do so, add a custom `KafkaListenerObservationConvention` and/or `KafkaTemplateObservationConvention` to the listener container properties or `KafkaTemplate` respectively.
 The `record` property in both observation contexts contains the `ConsumerRecord` or `ProducerRecord` respectively.
 
+The sender and receiver contexts' `remoteServiceName` properties are set to the Kafka `clusterId` property; this is retrieved by a `KafkaAdmin`.
+If, for some reason - perhaps lack of admin permissions, you cannot retrieve the cluster id, starting with version 3.1, you can set a manual `clusterId` on the `KafkaAdmin` and inject it into `KafkaTemplate` s and listener containers.
+When it is `null` (default), the admin will invoke the `describeCluster` admin operation to retrieve it from the broker.
```
artembilan, line 114: Don't we need to adjust `KafkaAdmin.initialize()` logic for this setting then?
There is an uncoditiional code like this:
```
synchronized (this) {
	this.clusterId = adminClient.describeCluster().clusterId().get(this.operationTimeout,
								TimeUnit.SECONDS);
}
```

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/micrometer.adoc
garyrussell: Good catch.

Issue #2870, opened 2023-10-31
No early bean initialization from `EmbeddedKafkaContextCustomizer`


garyrussell:
See https://github.com/spring-projects/spring-framework/issues/31453#issuecomment-1787252754

While we don't support the embedded broker in native images; the problem seems to be that the broker is not shut down properly.

snicoll:
The problem is rather that the broker is starting "manually" as part of a callback that should not start any bean, see 

https://github.com/spring-projects/spring-kafka/blob/0dfde15edf4b66706fc411137c1ec624e0064d4b/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafkaContextCustomizer.java#L127-L128


> Customize the supplied ConfigurableApplicationContext after bean definitions have been loaded into the context but before the context has been refreshed.

I think it should rather contribute a `RootBeanDefinition`. Unfortunately, the contract of this method is a bit strange as it should provide a registry, not a  context since it has not been refreshed yet.

Here is an example in Spring Boot of what I mean: https://github.com/spring-projects/spring-boot/blob/8f2ec227389391fdd173db0ab64f26abd2752f20/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/web/client/TestRestTemplateContextCustomizer.java#L62

artembilan:
Just debugged one of the test using `@EmbeddedKafka`. The `EmbeddedKafkaZKBroker.destroy()` is called and all the brokers and Zookeeper are shouted down in normal JVM.
I fail to find that `@DisabledInAotMode` to mark our `@EmbeddedKafka`, but I see that mentioned `TestRestTemplateContextCustomizer` uses `if (AotDetector.useGeneratedArtifacts()) {`.
Maybe that would be enough for us as well.
At the same time I don't see too much difference with what we have so far with a `((DefaultSingletonBeanRegistry) beanFactory).registerDisposableBean(EmbeddedKafkaBroker.BEAN_NAME, embeddedKafkaBroker);` and what Spring Boot does:
```
	if (beanFactory instanceof BeanDefinitionRegistry registry) {
			registerTestRestTemplate(registry);
		}
```

Even if we are about migrating to the `RootBeanDefinition`, we still need a `BeanDefinitionRegistry` registry.
So, I don't understand what we are pursuing here.

Thanks

snicoll:
[...]

PR #2872, opened 2023-10-31
GH-2870: EmbeddedKafka: register BeanDefinition

artembilan:
Fixes https://github.com/spring-projects/spring-kafka/issues/2870

The `EmbeddedKafkaContextCustomizer` uses `beanFactory.initializeBean()` which is too early according to the `ApplicationContext` lifecycle since it is not refreshed yet for `ContextCustomizer`

* Rework the logic in the `EmbeddedKafkaContextCustomizer` to register a `BeanDefinition` for an `EmbeddedKafkaBroker` to include it into standard `ApplicationContext` lifecycle
* Also mark `@EmbeddedKafka` with a `@DisabledInAotMode` to disallow this kind of tests in native images

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,7 +75,7 @@ ext {
 	springBootVersion = '3.0.9' // docs module
 	springDataVersion = '2023.1.0-RC1'
 	springRetryVersion = '2.0.4'
-	springVersion = '6.1.0-RC1'
+	springVersion = '6.1.0-SNAPSHOT'
 	zookeeperVersion = '3.6.4'
 
 	idPrefix = 'kafka'
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
@@ -33,6 +33,7 @@ import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -91,6 +92,8 @@ public class EmbeddedKafkaKraftBroker implements EmbeddedKafkaBroker {
 	private final int partitionsPerTopic;
 
 	private final Properties brokerProperties = new Properties();
+
+	private final AtomicBoolean initialized = new AtomicBoolean();
 
 	private KafkaClusterTestKit cluster;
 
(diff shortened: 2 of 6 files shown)


artembilan:
OK. So, there is a problem with this fix.
After starting that embedded broker we expose several system props:
```
System.setProperty(this.brokerListProperty, getBrokersAsString());
```

The in test configuration this cannot be resolved:
```
		@Value("${" + EmbeddedKafkaBroker.SPRING_EMBEDDED_KAFKA_BROKERS + "}")
		private String brokerAddresses;
```
Just because we have not initialized environment yet.
Therefore trying to postpone an embedded broker initialization from that customizer leads to many other problems, where the mentioned `afterPropertiesSet()` is not a problem at all for the rest of application context used just in a test scope. 🤷

snicoll:
IMO, there is no rush to fix this, certainly post RC as I expect other lifecycle issues. The code that we have now is making sure that the broker starts (very) early and, from that perspective, it is a good thing.  

> where the mentioned afterPropertiesSet() is not a problem at all for the rest of application context used just in a test scope. 

Having it working doesn't necessarily means it's right. To make it right, the bits that mutate the environment should be separated from the bits that actually start the container. Then, to make sure things are initialized in the right order, a dependency should probably be defined between the embedded broker and whatever is using it first. At the moment, the broker became a bean like any other bean and the container is free to start it whenever it feels like.

I can understand why it's frustrating to have to spend more time to make it work again with the proper semantics but having it done initially wouldn't have lead to this issue at all.

garyrussell:
I would rather change the tests to get the broker address from the broker bean itself rather than using that `@Value` injection.

We should not be calling `afterPropetiesSet()` in the customizer - that's what `initializeBean()` did before.

artembilan:
There is more into that, Gary.
Our `System.setProperty(this.brokerListProperty, getBrokersAsString());` is based on the `spring.kafka.bootstrap-servers` and it contributes to Spring Boot auto-configuration.
Therefore, as Stephane said, the environment must be mutated before the application context is refreshed.
From here we cannot defer that `afterPropetiesSet()` to the application context lifecycle.

I guess if we would not expose this `EmbeddedKafkaBroker` as a bean, but just started it from the customizer as it is done with many other embedded services, then we would not have this discussion at all.
Well, in fact, when there is no Spring context around the test, we use an `EmbeddedKafkaCondition` logic based on JUnit lifecycle.

In the end this customizer is there really to expose a bean for target Spring configuration convenience in the test, but the Kafka server must be started and its options exposed to the environment before other beans got into the action.
The mix of concerns in the `EmbeddedKafkaBroker` implementation is really one-stop-shop for end-user convenience.
You may even not use any annotations, but just declare that `EmbeddedKafkaBroker` as a bean in your test configuration.
Or just plain:
```
	@BeforeAll
	static void setup() {
		embeddedKafka = new EmbeddedKafkaKraftBroker(1, 2, topic1, topic2, topic3, topic4, topic5, topic6);
		embeddedKafka.afterPropertiesSet();
```

So, why is that so bad to have a single convenient component which can be reused in different places? The end result is very beneficial for users of our framework.
Finally, we just talk about the test environment and it has no any effect on production code 🤷

PR #2875, opened 2023-11-02
Improved the immutable field TRUSTED_PACKAGES in the DefaultJackson2JavaTypeMapper class to a List.of method that returns an ImmutableList.

ch4570:
Improvements

1. The Arrays.asList method creates a List based on variable arguments, which can lead to heap pollution, and TRUST_PACKAGES is an immutable field, so it has been improved to utilize Immutable List.

2. Improved to call the empty method to clarify the intent, rather than directly comparing integers to see if the size of a Set is zero.


artembilan (changes requested):
```
> Task :spring-kafka:checkstyleMain FAILED
Error: eckstyle] [ERROR] /home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/support/mapping/DefaultJackson2JavaTypeMapper.java:19:8: Unused import - java.util.Arrays. [UnusedImports]
```

artembilan:
@ch4570 ,

Issue #2654, opened 2023-04-14
Incorrect retry topic suffix


Desire456:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

2.8.2

**Describe the bug**

Incorrect retry topic suffix (`retry-5000`) when attempts is equal to 2 in `@RetryableTopic` and `fixedDelayStrategy=SINGLE_TOPIC`

**To Reproduce**

```
@KafkaListener(
        id = "someId",
        topics = [GREETING_TOPIC],
        groupId = "groupId",
        autoStartup = "true"
    )
@RetryableTopic(
        backoff = Backoff(delay = 5000),
        attempts = "2",
        autoCreateTopics = "false",
        fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC
    )
```

**Expected behavior**

Retry topic suffix = `-retry`

Desire456:
The problem is here:
https://github.com/spring-projects/spring-kafka/blob/f351a7a4aa85deb9c56c3f599d2eceac08e0f35f/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopicPropertiesFactory.java#L272.
`this.backOffValues.size() > 1` 
Why > 1?
Is 1 attempt with fixed delay is not fixed delay according to this method? Or it's feature that I didn't realize?

garyrussell:
Confirmed that this is a bug (when `attempts = "2"`) but, unfortunately, we can't change the default behavior in a patch release; we could make it an opt-in, in a future release.

As a work around, see https://docs.spring.io/spring-kafka/docs/current/reference/html/#custom-naming-strategies

Here's an example:

```java
@SpringBootApplication
public class Kgh2654Application extends RetryTopicConfigurationSupport {

	public static void main(String[] args) {
		SpringApplication.run(Kgh2654Application.class, args);
	}

	@Bean
	NewTopic topic() {
		return TopicBuilder.name("kgh2654").partitions(1).replicas(1).build();
	}

	@Bean
	ApplicationRunner runner(KafkaTemplate<String, String> template) {
		return args -> {
			template.send("kgh2654", "foo");
		};
	}

	@Override
	protected RetryTopicComponentFactory createComponentFactory() {
		return new RetryTopicComponentFactory() {

			@Override
			public RetryTopicNamesProviderFactory retryTopicNamesProviderFactory() {
				return new CustomRetryTopicNamesProviderFactory();
			}

		};
	}

[...]

PR #2876, opened 2023-11-02
GH-2654: Incorrect retry topic suffix

Wzy19930507:
Resolves https://github.com/spring-projects/spring-kafka/issues/2654

* fix incorrect retry topic suffix when attempts is 2

* optimization method `isDelayWithReusedTopic`

* add retry topic integration tests when reuse retry topic

fix java doc in RetryTopic

* fix javadoc in `@RetryTopic`

* fix javadoc in `RetryTopicConfigurationProvider`

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
@@ -38,3 +38,9 @@ See xref:kafka/container-factory.adoc[Container Factory] for more information.
 You can now add a `Validator` to this deserializer; if the delegate `Deserializer` successfully deserializes the object, but that object fails validation, an exception is thrown similar to a deserialization exception occurring.
 This allows the original raw data to be passed to the error handler.
 See xref:kafka/serdes.adoc#error-handling-deserializer[Using `ErrorHandlingDeserializer`] for more information.
+
+[[x31-retryable]]
+=== Retryable Topics
+Change suffix -retry-5000 to -retry when `@RetryableTopic(backoff = @Backoff(delay = 5000), attempts = "2", fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC)`.
+If you want to keep suffix -retry-5000, use `@RetryableTopic(backoff = @Backoff(delay = 5000), attempts = "2")`.
+See xref:retrytopic/topic-naming.adoc[Topic Naming] for more information.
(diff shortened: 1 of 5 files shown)


artembilan (approved):
This is OK with me, but we still need @garyrussell 's approval.

And I guess since we have a workaround for older versions, we are not going to back-port the fix since it feels a bit as a breaking change.

Thanks

Wzy19930507:
Thanks for your review 💯

garyrussell (changes requested):
LGTM, but we should add a note to the `whats-new.adoc` about the behavior change - users will somehow need to migrate any messages in the "old" retry topic.

On the other hand, maybe the new behavior needs to be an opt-in property, to avoid the need for such migration.

We are getting close to GA (next week); if we decide to go with the opt-in approach, it would be ok to move it to 3.1.1.

Wzy19930507:
> On the other hand, maybe the new behavior needs to be an opt-in property, to avoid the need for such migration.

May we use this case replacement opt-in?
```
@RetryableTopic(
        backoff = Backoff(delay = 5000),
        attempts = "2",
        autoCreateTopics = "false",
        fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC
        
    )
```

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -38,3 +38,9 @@ See xref:kafka/container-factory.adoc[Container Factory] for more information.
 You can now add a `Validator` to this deserializer; if the delegate `Deserializer` successfully deserializes the object, but that object fails validation, an exception is thrown similar to a deserialization exception occurring.
 This allows the original raw data to be passed to the error handler.
 See xref:kafka/serdes.adoc#error-handling-deserializer[Using `ErrorHandlingDeserializer`] for more information.
+
+[[x31-retryable]]
+=== Retryable Topics
+Change suffix -retry-5000 to -retry when `@RetryableTopic(backoff = @Backoff(delay = 5000), attempts = "2", fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC)`.
+If you want to keep suffix -retry-5000, use `@RetryableTopic(backoff = @Backoff(delay = 5000), attempts = "2")`.
+See xref:retrytopic/topic-naming.adoc[Topic Naming] for more information.
```
garyrussell, line 46:
```suggestion
Change suffix `-retry-5000` to `-retry` when `@RetryableTopic(backoff = @Backoff(delay = 5000), attempts = "2", fixedDelayTopicStrategy = FixedDelayStrategy.SINGLE_TOPIC)`.
If you want to keep suffix `-retry-5000`, use `@RetryableTopic(backoff = @Backoff(delay = 5000), attempts = "2")`.
See xref:retrytopic/topic-naming.adoc[Topic Naming] for more information.
```

garyrussell (approved):

garyrussell:
Thanks!

Please rebase to main to address conflicts.

Wzy19930507:
Thanks you for review : )

PR #2881, opened 2023-11-05
fix CONSUMER_RECORDS_CLASS_NAME constant error in KafkaAvroBeanRegistrationAotProcessor.java

NathanQingyangXu:


spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaAvroBeanRegistrationAotProcessor.java
```diff
@@ -79,7 +80,7 @@ public BeanRegistrationAotContribution processAheadOfTime(RegisteredBean registe
 				}
 			}, method -> method.getName().equals("onMessage"));
 		}
-		if (avroTypes.size() > 0) {
+		if (!avroTypes.isEmpty()) {
```
NathanQingyangXu, line 83: the above cosmetic is minor; but I do think the change improves readability a little bit (IDEA agrees as well)

PR #2882, opened 2023-11-05
add a missing left quote in AbstractKafkaListenerEndpoint#getEndpointDescription()

NathanQingyangXu:


artembilan:
I don't treat this as critical therefore I'm not going to back-port it to previous versions.

Thanks

PR #2885, opened 2023-11-08
fix some typos

NathanQingyangXu:
fix some obvious typos in various javadocs, code comments or log statements

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHints.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHints.java
@@ -144,7 +144,7 @@ public class KafkaRuntimeHints implements RuntimeHintsRegistrar {
 					RangeAssignor.class,
 					RoundRobinAssignor.class,
 					StickyAssignor.class,
-					// standard partitions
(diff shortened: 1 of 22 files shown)


spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHints.java
```diff
@@ -144,7 +144,7 @@ public void registerHints(RuntimeHints hints, @Nullable ClassLoader classLoader)
 					RangeAssignor.class,
 					RoundRobinAssignor.class,
 					StickyAssignor.class,
-					// standard partitioners
+					// standard partitions
```
artembilan, line 147: This is wrong. 
The next classes in this list are really `Partitioner` implementations.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHints.java
NathanQingyangXu: yeah, you are totally right. Correction reverted back. thx

PR #2888, opened 2023-11-09
GH-2588: ARBP support batch recoverable

### LINKED ISSUE #2588: Provide ability to disable/limit retries in batch listener when using DefaultAfterRollbackProcessor
**Expected Behavior**
Configuration to disable/limit retries in the batch listener when using DB transactions and `DefaultAfterRollbackProcessor`
Example of listener `@KafkaListener(topics = BATCH_TOPIC, batch = "true") {}`

**Current Behavior**
Currently, the `DefaultAfterRollbackProcessor` retries the failing message infinitely if using batch listener with DB transactions.

**Context**
The main problem if we forget to handle the failing message inside the listener then the failed message will stuck until we take some manual action to skip this message. That's why we would like to have a config to disable/limit retries at the configurations level.
I tried to find a way to disable retries for batch listeners but unsuccessfully.
`new FixedBackOff(0, 0)`
`new DefaultAfterRollbackProcessor<>(deadLetterPublishingRecoverer, backOff);`

The workaround can be to force to use `recoverable=true` always  in order to limit retries

```
return new DefaultAfterRollbackProcessor<>(deadLetterPublishingRecoverer, backOff) {
            @Override
            public void process(
                    @NotNull List<ConsumerRecord<Object, Object>> consumerRecords,
                    @NotNull org.apache.kafka.clients.consumer.Consumer<Object, Object> consumer,
                    MessageListenerContainer container,
                    @NotNull Exception exception,
                    boolean recoverable,
                    @NotNull ContainerProperties.EOSMode eosMode) {
                // Force all messages to be recoverable.
                super.process(consumerRecords, consumer, container, exception, true, eosMode);
            }
        };
```

Wzy19930507:
Resolves #2588

This is a draft of support DLT batch recover for `DefaultAfterRollbackProcessor`. so we can evaluate if it's going in the right direction before doing a final review, adding docs, etc.

Function
* add opt-in property `batchRecoverAfterRollback` at `ContainerProperties`
* add method processBatch at `AfterRollbackProcessor`

Batch recoverable
* batchRecoverAfterRollback is true
* set transcation DLT to enable isCommitRecovered
* enable kafka transcation

exception handler logic
* in retry, seek to begin
* out of retry exception recover entire recods
* recover fail, seek to begin

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
@@ -51,6 +51,7 @@ import org.springframework.util.CollectionUtils;
  * @author Johnny Lim
  * @author Lukasz Kaminski
  * @author Kyuhyeok Park
+ * @author Wang Zhiyang
  */
 public class ContainerProperties extends ConsumerProperties {
 
@@ -545,10 +546,20 @@ public class ContainerProperties extends ConsumerProperties {
 		this.transactionManager = transactionManager;
 	}
 
+	/**
+	 * Recover batch records after rollback if true.
+	 * @return true to recover.
+	 * @since 3.2
+	 */
 	public boolean isBatchRecoverAfterRollback() {
 		return this.batchRecoverAfterRollback;
 	}
 
+	/**
+	 * enable the batch recover after rollback.
+	 * @param batchRecoverAfterRollback the batchRecoverAfterRollback to set.
+	 * @since 3.2
+	 */
 	public void setBatchRecoverAfterRollback(boolean batchRecoverAfterRollback) {
 		this.batchRecoverAfterRollback = batchRecoverAfterRollback;
 	}
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
@@ -258,6 +258,8 @@ public class ContainerProperties extends ConsumerProperties {
 
 	private PlatformTransactionManager transactionManager;
 
+	private boolean batchRecoverAfterRollback = false;
+
 	private int monitorInterval = DEFAULT_MONITOR_INTERVAL;
 
 	private TaskScheduler scheduler;
@@ -541,6 +543,14 @@ public class ContainerProperties extends ConsumerProperties {
 	 */
 	public void setTransactionManager(@Nullable PlatformTransactionManager transactionManager) {
 		this.transactionManager = transactionManager;
+	}
+
+	public boolean isBatchRecoverAfterRollback() {
+		return this.batchRecoverAfterRollback;
+	}
+
+	public void setBatchRecoverAfterRollback(boolean batchRecoverAfterRollback) {
+		this.batchRecoverAfterRollback = batchRecoverAfterRollback;
 	}
 
 	public int getMonitorInterval() {
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
@@ -50,6 +50,7 @@ import org.springframework.util.backoff.BackOffExecution;
  *
  * @author Gary Russell
  * @author Francois Rosiere
+ * @author Wang Zhiyang
  *
  * @since 1.3.5
  *
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2023 the original author or authors.
+ * Copyright 2018-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -216,7 +216,7 @@ public class DefaultAfterRollbackProcessor<K, V> extends FailedRecordProcessor
 			}
 			catch (Exception ex) {
 				SeekUtils.doSeeksToBegin((List) recordList, consumer, this.logger);
-				logger.error(ex, () -> "Recoverer threw an exception; re-seeking batch");
+				logger.error(ex, "Recoverer threw an exception; re-seeking batch");
 				throw ex;
 			}
 			return;
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
@@ -17,6 +17,7 @@
 package org.springframework.kafka.listener;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -24,9 +25,11 @@ import java.util.function.BiConsumer;
 
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
 
+import org.springframework.kafka.KafkaException;
 import org.springframework.kafka.core.KafkaOperations;
 import org.springframework.kafka.listener.ContainerProperties.EOSMode;
 import org.springframework.lang.Nullable;
@@ -60,7 +63,9 @@ public class DefaultAfterRollbackProcessor<K, V> extends FailedRecordProcessor
 
 	private final BackOff backOff;
 
-	private KafkaOperations<?, ?> kafkaTemplate;
+	private final KafkaOperations<?, ?> kafkaTemplate;
+
+	private final BiConsumer<ConsumerRecords<?, ?>, Exception> recoverer;
 
 	/**
 	 * Construct an instance with the default recoverer which simply logs the record after
@@ -143,6 +148,11 @@ public class DefaultAfterRollbackProcessor<K, V> extends FailedRecordProcessor
 		super.setCommitRecovered(commitRecovered);
 		checkConfig();
 		this.backOff = backOff;
+		this.recoverer = (crs, ex) -> {
+			if (recoverer != null && !crs.isEmpty()) {
+				crs.spliterator().forEachRemaining(rec -> recoverer.accept(rec, ex));
+			}
+		};
 	}
 
 	private void checkConfig() {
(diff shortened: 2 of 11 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
```diff
@@ -138,7 +141,7 @@ public DefaultAfterRollbackProcessor(@Nullable BiConsumer<ConsumerRecord<?, ?>,
 			BackOff backOff, @Nullable BackOffHandler backOffHandler, @Nullable KafkaOperations<?, ?> kafkaOperations,
 			boolean commitRecovered) {
 
-		super(recoverer, backOff, backOffHandler);
+		super(recoverer, backOff, backOffHandler, null);
```
Wzy19930507, line 144: may be use `CommonLoggingErrorHandler` first

garyrussell:
Can you provide a valid use case where this would be useful? 

Any transactional sends for the records before the failed one would have been rolled back, so it doesn’t make sense to to skip those records on the retry. If you are not doing transactional sends, then why use transactions?

I tried to convey this in my comments on the issue.

Wzy19930507:
> Any transactional sends for the records before the failed one would have been rolled back, so it doesn’t make sense to to skip those records on the retry. If you are not doing transactional sends, then why use transactions?

Exactly right, this is the most important point in this issue.

I had a rethink in the `BatchListenerFailedException` sence. Simple implementation is
- in retry range that seek to reocrds begin offsets
- retry exhaustion that return remaing record exclude recovered records.

spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
```diff
@@ -137,20 +144,29 @@ public DefaultAfterRollbackProcessor(@Nullable
...
 			BackOff backOff, @Nullable BackOffHandler backOffHandler, @Nullable KafkaOperations<?, ?> kafkaOperations,
 			boolean commitRecovered) {
-
-		super(recoverer, backOff, backOffHandler);
+		super(recoverer, backOff, backOffHandler, new CommonErrorHandler() { });
 		this.kafkaTemplate = kafkaOperations;
 		super.setCommitRecovered(commitRecovered);
 		checkConfig();
 		this.backOff = backOff;
+		this.recoverer = (crs, ex) -> {
```
Wzy19930507, line 152: it's invalid.
recoverer need transaction too
if use DLT,
need producer inTX, consumer inTX, DLT inTX, DLT consumer inTX

garyrussell:
One more time - we cannot support `BatchListenerFailedException` here - the transaction will have rolled back by now so any transactional sends for the records before the failed record will have been ignored; it is not correct to discard those records.

Yes, we could support recovering (sending to DLT) the entire batch, but we can't discard a partial batch here. Even that would have to be an opt-in feature.

Wzy19930507:
Hi, @garyrussell 
Could you do me a favor for build failed?

I didn't modify code at `GlobalEmbeddedKafkaTestExecutionListenerTests#testGlobalEmbeddedKafkaTestExecutionListener`
and ran `./gradlew check` success locally.  

but build failed and following exception is reported
image

garyrussell:
Re-running now; that test is sometimes flaky.

Wzy19930507:
Thanks

artembilan:
@Wzy19930507 ,

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -522,6 +528,7 @@ public Long getIdlePartitionEventInterval() {
 		return this.idlePartitionEventInterval;
 	}
 
+	@Deprecated(since = "3.1")
```
artembilan, line 531: `3.2`

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerUtils.java
```diff
@@ -144,6 +139,17 @@ public static void unrecoverableBackOff(BackOff backOff, Map<Thread, BackOffExec
 		}
 	}
 
+	public static long nextBackOff(BackOff backOff, Map<Thread, BackOffExecution> executions) {
```
artembilan, line 142: Not sure if this method supposed to be `public`, but if it is, then it has to have respective Javadoc.

spring-kafka/src/main/java/org/springframework/kafka/listener/SeekUtils.java
```diff
@@ -133,6 +133,17 @@ public static boolean doSeeks(List<ConsumerRecord<?, ?>> records, Consumer<?, ?>
 		return skipped.get();
 	}
 
+	public static void doSeeksToBegin(List<ConsumerRecord<?, ?>> records, Consumer<?, ?> consumer,
```
artembilan, line 136: Javadoc, please

spring-kafka/src/test/java/org/springframework/kafka/listener/CommitOnAssignmentTests.java
```diff
@@ -135,13 +135,48 @@ void testLatestOnlyNoTx() throws InterruptedException {
 		KafkaTransactionManager tm = new KafkaTransactionManager<>(pf);
 		Producer producer = mock(Producer.class);
 		given(pf.createProducer(any())).willReturn(producer);
-		CountDownLatch latch = new CountDownLatch(1);
 		props.setTransactionManager(tm);
```
artembilan, line 138: Can we consider to not use deprecated API then?
Although it is not clear why deprecation of that property is the part of this change.
It is really hard to review mixed concerns.

artembilan (changes requested):
Don't we also need to document the feature somehow?

Thanks

spring-kafka/src/test/java/org/springframework/kafka/listener/CommitOnAssignmentTests.java
Wzy19930507: I'll create anthoer issue to this pr

artembilan (changes requested):
Please, also add your name to the `@author` list of all the affected classes.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/SeekUtils.java
```diff
@@ -133,6 +133,23 @@ public static boolean doSeeks(List<ConsumerRecord<?, ?>> records, Consumer<?, ?>
 		return skipped.get();
 	}
 
+	/**
+	 * Seek records to begin position, optionally skipping the first.
+	 * @param records the records.
+	 * @param consumer the consumer.
+	 * @param logger a {@link LogAccessor} for seek errors.
```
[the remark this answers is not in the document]
Wzy19930507: yes, new method for `DefaultAfterRollbackProcessor.processBatch`

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerUtils.java
```diff
@@ -144,6 +139,17 @@ public static void unrecoverableBackOff(BackOff backOff, Map<Thread, BackOffExec
 		}
 	}
 
+	static long nextBackOff(BackOff backOff, Map<Thread, BackOffExecution> executions) {
```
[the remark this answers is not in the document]
Wzy19930507: I use `ListenerUtils.nextBackOff` in DefaultAfterRollbackProcessor.processBatch to retry exception

spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java
```diff
@@ -176,6 +186,53 @@ && isCommitRecovered() && this.kafkaTemplate.isTransactional()) {
...
+
+		if (recoverable && isCommitRecovered()) {
+			long nextBackOff = ListenerUtils.nextBackOff(this.backOff, this.backOffs);
+			if (nextBackOff != BackOffExecution.STOP) {
+				SeekUtils.doSeeksToBegin((List) recordList, consumer, this.logger);
+				try {
+					ListenerUtils.stoppableSleep(container, nextBackOff);
+				}
+				catch (InterruptedException e) {
+					Thread.currentThread().interrupt();
```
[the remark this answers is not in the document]
Wzy19930507: If new transaction, no rollback is required before `ListenerUtils.stoppableSleep(container, nextBackOff)` in method `processBatch`.
And i'm not sure it makes sense to roll back interrupt exceptions in a nested transaction scenario

spring-kafka/src/test/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessorTests.java
```diff
@@ -118,37 +123,25 @@ void testBatchBackOff() {
...
+		processor.processBatch(consumerRecords, records, consumer, container, illegalState, false, EOSMode.V2);
 		verify(backOff, times(2)).start();
 		verify(execution.get(), times(2)).nextBackOff();
 		processor.clearThreadState();
-		processor.process(records, consumer, container, illegalState, false, EOSMode.V2);
+		processor.processBatch(consumerRecords, records, consumer, container, illegalState, false, EOSMode.V2);
 		verify(backOff, times(3)).start();
 	}
 
-	void testEarlyExitBackOff() {
```
artembilan, line 136: Why removing this instead of adding `@Test` on the method?

spring-kafka/src/test/java/org/springframework/kafka/listener/TransactionalContainerTests.java
```diff
@@ -777,12 +785,133 @@ public void accept(ConsumerRecord<?, ?> record, Consumer<?, ?> consumer, Excepti
...
+		verify(afterRollbackProcessor, times(4)).processBatch(any(), any(), any(), any(), captor.capture(), anyBoolean(), any());
+		assertThat(captor.getValue()).isInstanceOf(ListenerExecutionFailedException.class)
+				.extracting(ex -> ((ListenerExecutionFailedException) ex).getGroupId())
+				.isEqualTo(group);
+		verify(afterRollbackProcessor, times(2)).clearThreadState();
+		verify(dlTemplate, times(5)).send(any(ProducerRecord.class));
+		verify(dlTemplate).sendOffsetsToTransaction(
+				eq(Collections.singletonMap(new TopicPartition(topic8, 0), new OffsetAndMetadata(4L))),
+				any(ConsumerGroupMetadata.class));
+		logger.info("Stop testBatchListenerMaxFailures");
```
artembilan, line 907: Why do we need this info in the logs of the test?

spring-kafka/src/test/java/org/springframework/kafka/listener/TransactionalContainerTests.java
```diff
@@ -841,6 +970,80 @@ public void testRollbackProcessorCrash() throws Exception {
 		logger.info("Stop testRollbackNoRetries");
 	}
 
+	@SuppressWarnings("unchecked")
+	@Test
+	public void testBatchListenerRecoverAfterRollbackProcessorCrash() throws Exception {
+		logger.info("Start testBatchListenerRollbackNoRetries");
```
artembilan, line 976: I really don't see reason in this extra logs for the tests...

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessorTests.java
Wzy19930507: You are right!

sobychacko (approved):

Wzy19930507:
@artembilan @garyrussell @sobychacko Thanks for review, this issue could not have been closed without your guidance.

PR #2892, opened 2023-11-11
make EmbeddedKafkaBroker.BROKER_LIST_PROPERTY take precedence to default one

NathanQingyangXu:
As per the javadoc:

> Instead of default spring.embedded.kafka.brokers system property, the address of the Kafka brokers can be exposed to any arbitrary and convenient property. For this purpose a spring.embedded.kafka.brokers.property (EmbeddedKafkaBroker.BROKER_LIST_PROPERTY) system property can be set before starting an embedded Kafka.

Literally it seems to imply when `spring.embedded.kafka.brokers.property ` is set, the default `spring.embedded.kafka.brokers` system property should not be set; but it seems regardless, the default is still set

Not 100% sure about my code changes and even my understanding is correct.

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
@@ -76,12 +76,6 @@ public class EmbeddedKafkaKraftBroker implements EmbeddedKafkaBroker {
 
 	private static final LogAccessor LOGGER = new LogAccessor(LogFactory.getLog(EmbeddedKafkaKraftBroker.class));
 
-	/**
-	 * Set the value of this property to a property name that should be set to the list of
-	 * embedded broker addresses instead of {@value #SPRING_EMBEDDED_KAFKA_BROKERS}.
-	 */
-	public static final String BROKER_LIST_PROPERTY = "spring.embedded.kafka.brokers.property";
-
 	public static final int DEFAULT_ADMIN_TIMEOUT = 10;
 
 	private final int count;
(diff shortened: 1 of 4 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
```diff
@@ -225,14 +219,17 @@ private void start() {
...
+		String brokersAsString = getBrokersAsString();
 		if (this.brokerListProperty != null) {
-			System.setProperty(this.brokerListProperty, getBrokersAsString());
+			System.setProperty(this.brokerListProperty, brokersAsString);
+		}
+		String globalBrokerListProperty = System.getProperties().containsKey(BROKER_LIST_PROPERTY)
+				? System.getProperties().getProperty(BROKER_LIST_PROPERTY)
+				: SPRING_EMBEDDED_KAFKA_BROKERS;
+		if (!globalBrokerListProperty.equals(this.brokerListProperty)) {
+			System.setProperty(globalBrokerListProperty, brokersAsString);
```
NathanQingyangXu, line 231: we have two ways to pass the broker list property:

- chaining setter
- system property

Seems both are working independently; for that reason, I renamed the latter as 'globalBrokerListProperty` to differentiate with `this.brokerListProperty`.

artembilan:
I agree about closing this as `Works as Designed`.
The motivation is indeed was to always have that default exposed independently of `spring.embedded.kafka.brokers.property` setting.

NathanQingyangXu:
Yeah, later on I realized that. Though it is a little bit confusing for the
doc mentioned "instead of...", but it seems the reality is "in addition
to...", for the default one is always there. But anyway.

PR #2893, opened 2023-11-11
remove unnecessary property setting statement in EmbeddedKafkaKraftBroker

NathanQingyangXu:
```
if (this.brokerListProperty != null) {
    System.setProperty(this.brokerListProperty, getBrokersAsString());
}
System.setProperty(this.brokerListProperty, getBrokersAsString());
```

The last statement seems unnecessary and could lead to exception thrown if `this.brokerListProperty` is null


artembilan:
This is covered in the https://github.com/spring-projects/spring-kafka/pull/2872.
But yeah... it is not clear when (and if) that one is going to be merged...

Issue #2806, opened 2023-09-11
Receiving an empty list when using RecordFilterStrategy on batch messages


artembilan:
### Discussed in https://github.com/spring-projects/spring-kafka/discussions/2805

<div type='discussions-op-text'>

<sup>Originally posted by **m-hetz** September 11, 2023</sup>
Hi!

I'm using `RecordFilterStrategy` to filter records based on headers. This works great on single messages, the filter works and records don't  reach the listener (annotated with `@KafkaListener`).

When using batch mode, filtering also works but when the entire batch is filtered, the listener is invoked with an empty list.
Is this expected behavior? According to [this](https://docs.spring.io/spring-kafka/api/org/springframework/kafka/listener/adapter/RecordFilterStrategy.html)  when returning an empty list all records are filtered

Is there a way to configure it not to reach the listener at all? (on an empty bulk)

Thanks!

```
val filter = object : RecordFilterStrategy<String, String> {
    override fun filter(consumerRecord: ConsumerRecord<String, String>): Boolean {
        val header = consumerRecord.headers().lastHeader("SKIP")
        if (header != null) {
            logger.info("Filtering header=" + consumerRecord.key())
            return true
        }
        return false
    }

    override fun filterBatch(records: MutableList<ConsumerRecord<String, String>>): MutableList<ConsumerRecord<String, String>> {
        val iterator: MutableIterator<ConsumerRecord<String, String>> = records.iterator()

        while (iterator.hasNext()) {
            if (this.filter(iterator.next())) {
                iterator.remove()
            }
        }
        if (records.size > 0) return records
        return mutableListOf()
    }
}
factory.setBatchListener(isBatchListener)
factory.setRecordFilterStrategy(filter)


@KafkaListener(topics = ["topic"], containerFactory = "messageContainerFactory")
fun listener(messages: List<ConsumerRecord<String, String?>>,
) {

    logger.info("Consuming messages, batch size=${messages.size}")
}
```

Sorry for the unformatted code, I tried editing the message but the code is still messed up</div>

chickenchickenlove:
Hi, @artembilan. 
I'm interested in solving this issue. 

```java
		if (consumerRecords.size() > 0 || consumerAware
				|| (!this.ackDiscarded && this.delegateType.equals(ListenerType.ACKNOWLEDGING))) {
			invokeDelegate(consumerRecords, acknowledgment, consumer);
		}
```
[...]

PR #2895, opened 2023-11-13
GH-2806 fix bug of receiving an empty list when using RecordFilterStrategy on batch messages

NathanQingyangXu:
https://github.com/spring-projects/spring-kafka/issues/2806

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchMessageListenerAdapter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchMessageListenerAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 the original author or authors.
+ * Copyright 2016-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchMessageListenerAdapter.java
```diff
@@ -82,11 +82,11 @@ public void onMessage(List<ConsumerRecord<K, V>> records, @Nullable Acknowledgme
...
 		/*
-		 *  An empty list goes to the listener if ackDiscarded is false and the listener can ack
-		 *  either through the acknowledgment
+		 *  An empty list goes to the listener if ackDiscarded is false, and the listener can ack
+		 *  either through the acknowledgment or through the consumer
 		 */
-		if (consumerRecords.size() > 0 || consumerAware
-				|| (!this.ackDiscarded && this.delegateType.equals(ListenerType.ACKNOWLEDGING))) {
+		if (consumerRecords.size() > 0
+				|| (!this.ackDiscarded && ((consumer != null && consumerAware) || (acknowledgment != null && this.delegateType.equals(ListenerType.ACKNOWLEDGING))))) {
```
NathanQingyangXu, line 89: not an ultimate fixing (`consumerAware` is always true) but might be sufficient.

NathanQingyangXu:
I closed this for it is much more complicated than I originally thought. Still learned a lot from the process.

PR #2897, opened 2023-11-15
fix many issues in DefaultKafkaProducerFactory#updateConfigs()

NathanQingyangXu:
There are many defects in the original DefaultKafkaProducerFactory#updateConfigs() as below:
```
#1 public void updateConfigs(Map<String, Object> updates) {
#2		updates.entrySet().forEach(entry -> {
#3			if (entry.getKey().equals(ProducerConfig.TRANSACTIONAL_ID_CONFIG)) {
#4				Assert.isTrue(entry.getValue() instanceof String, () -> "'" + ProducerConfig.TRANSACTIONAL_ID_CONFIG
#5						+ "' must be a String, not a " + entry.getClass().getName());
#6				Assert.isTrue(this.transactionIdPrefix != null
#7								? entry.getValue() != null
#8								: entry.getValue() == null,
#9						"Cannot change transactional capability");
#10				this.transactionIdPrefix = (String) entry.getValue();
#11			}
#12			else if (entry.getKey().equals(ProducerConfig.CLIENT_ID_CONFIG)) {
#13				Assert.isTrue(entry.getValue() instanceof String, () -> "'" + ProducerConfig.CLIENT_ID_CONFIG
#14						+ "' must be a String, not a " + entry.getClass().getName());
#15				this.clientIdPrefix = (String) entry.getValue();
#16			}
#17			else {
#18				this.configs.put(entry.getKey(), entry.getValue());
#19			}
#20		});
#21	}
```

It has following defects:

1. given the input is a Map, we can't rule out the possibility that both key and value could be null (e.g. in default HashMap); when key is null, Line #2 will thrown NPE; when value is null; #18 will throw exception for ConcurrentHashMap disallows null key or value (this.configs is based on this impl class);
2. from either Line #10, it implies null value for `ProducerConfig.TRANSACTIONAL_ID_CONFIG` is allowed; but when entry.getValue() returns null, the `entry.getValue() instanceof String` in Line #4 should have thrown NPE without reaching later statements
3. obviously in both Line #5 and Line #14, the intention was ` "' must be a String, not a " + entry.getValue().getClass().getName())`; `entry.getClass()` would always return `Map.Entry`;

I found I need to fix the legacy checkStyle issue for this gigantic class. To avoid code review difficulty, I split the PR commits into two, the latter of which is focused on legacy checkStyle manual labour. So to verify the code logic change alone, simply focus on the first commit.


spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -165,6 +165,7 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 
 	/**
 	 * Construct a factory with the provided configuration.
+	 *
```
garyrussell, line 170: We never have blank lines in method javadocs; they are allowed in class java docs.

garyrussell (changes requested):
Thanks; but it is not clear why you found it necessary for all those format changes.

artembilan:
How about this `./gradlew check -x test`?
Probably that `checkstyleMain` doesn't see our `src/checkstyle/checkstyle.xml` config.

NathanQingyangXu:
> How about this `./gradlew check -x test`? Probably that `checkstyleMain` doesn't see our `src/checkstyle/checkstyle.xml` config.

thanks. It works. Lesson learned.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -626,23 +626,31 @@ public boolean removePostProcessor(ProducerPostProcessor<K, V> postProcessor) {
...
-				Assert.isTrue(entry.getValue() instanceof String, () -> "'" + ProducerConfig.TRANSACTIONAL_ID_CONFIG
-						+ "' must be a String, not a " + entry.getClass().getName());
-				Assert.isTrue(this.transactionIdPrefix != null
-								? entry.getValue() != null
-								: entry.getValue() == null,
-						"Cannot change transactional capability");
-				this.transactionIdPrefix = (String) entry.getValue();
+		updates.forEach((key, value) -> {
+			if (key == null) {
+				return;
```
artembilan, line 631: Doesn't this mean that we are going to exit from a loop on a first key as null?
Therefore we just don't examine all the entries in the map.

artembilan (changes requested):

garyrussell (approved):

PR #2899, opened 2023-11-16
fix a bug in RoutingKafkaTemplate

NathanQingyangXu:
The following code has bug:
```
public RoutingKafkaTemplate(Map<Pattern, ProducerFactory<Object, Object>> factories) {
    ... ...
    Optional<Boolean> transactional = factories.values().stream()
	.map(fact -> fact.transactionCapable())
	.findFirst();
    Assert.isTrue(!transactional.isPresent() || !transactional.get(), "Transactional factories are not supported");
}
```
Suppose the factories havetwo elements, the first is not `transactionCapable` whereas the second is. then the above stream usage only covers the first factory due to the `map()` usage (`filter()` was the intention).
Created a unit testing case to prove the bug and cover our asses.


artembilan (approved):
Sounds good!

Let's see if @garyrussell agrees and we need to back-port this down to all supported versions!

garyrussell (approved):

PR #2904, opened 2023-11-17
make use of incoming Message's `replyTopic` header if needed

NathanQingyangXu:
as per reference doc: https://docs.spring.io/spring-kafka/docs/current/reference/html/#reply-message:

> Starting with version 2.5, the framework will detect if these headers are missing and populate them with the topic - either the topic determined from the @SendTo value or the incoming KafkaHeaders.REPLY_TOPIC header (if present)

Seems we overlooked the incoming `KafkaHeaders.REPLY_TOPIC` logic.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
@@ -80,6 +80,7 @@ import org.springframework.util.StringUtils;
  * @author Gary Russell
  * @author Artem Bilan
  * @author Venil Noronha
+ * @author Nathan Xu
  */
 public abstract class MessagingMessageListenerAdapter<K, V> implements ConsumerSeekAware {
 
@@ -466,7 +467,13 @@ public abstract class MessagingMessageListenerAdapter<K, V> implements ConsumerS
 	 * @since 2.1.3
 	 */
 	@SuppressWarnings("unchecked")
-	protected void sendResponse(Object result, String topic, @Nullable Object source, boolean returnTypeMessage) {
+	protected void sendResponse(Object result, @Nullable String topic, @Nullable Object source, boolean returnTypeMessage) {
+		if (topic == null && source instanceof Message) {
+			byte[] replyTopicBytes = getReplyTopic((Message<?>) source);
+			if (replyTopicBytes != null) {
+				topic = new String(replyTopicBytes, StandardCharsets.UTF_8);
+			}
+		}
 		if (!returnTypeMessage && topic == null) {
 			this.logger.debug(() -> "No replyTopic to handle the reply: " + result);
 		}
@@ -482,14 +489,14 @@ public abstract class MessagingMessageListenerAdapter<K, V> implements ConsumerS
 					iterableOfMessages = iterator.next() instanceof Message;
 				}
 				if (iterableOfMessages || this.splitIterables) {
-					((Iterable<V>) result).forEach(v -> {
+					for (V v : (Iterable<V>) result) {
 						if (v instanceof Message) {
-							this.replyTemplate.send((Message<?>) v);
+							this.replyTemplate.send(checkHeaders(v, topic, source));
 						}
 						else {
 							this.replyTemplate.send(topic, v);
 						}
-					});
+					}
 				}
 				else {
 					sendSingleResult(result, topic, source);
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -482,14 +489,14 @@ else if (result instanceof Message) {
...
-					((Iterable<V>) result).forEach(v -> {
+					for (V v : (Iterable<V>) result) {
 						if (v instanceof Message) {
 							this.replyTemplate.send((Message<?>) v);
 						}
 						else {
 							this.replyTemplate.send(topic, v);
 						}
-					});
+					}
```
NathanQingyangXu, line 499: lambda has to be switched to loop for `topic` is not effective final after logic change

artembilan (changes requested):
@NathanQingyangXu ,

thank you very much for so thoughtful contributions!

Would you mind double checking if PR build failure is related somehow to your changes: https://github.com/spring-projects/spring-kafka/actions/runs/6903843623/job/18783345784?pr=2904 ?

artembilan:
Can you, please, stop squashing your commits?
It is really much easier to review if there is a history of commits in the PR.
And sometime we just review the latest commit which might not include changes for other files.
When we merge PR, we squash to preserve only one commit in the `main` for the issue we are resolving.

Thanks for understanding!

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -482,14 +489,14 @@ else if (result instanceof Message) {
 					iterableOfMessages = iterator.next() instanceof Message;
```
NathanQingyangXu, line 489: the current approach assumes the `Iterable` is not `Iterable<?>` and treat the first element type as the indicator of the whole collection. but if we get a `List<Object>`, should we tell on a case-by-case basis instead?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -482,14 +489,14 @@ else if (result instanceof Message) {
...
 				}
 				if (iterableOfMessages || this.splitIterables) {
-					((Iterable<V>) result).forEach(v -> {
+					for (V v : (Iterable<V>) result) {
 						if (v instanceof Message) {
-							this.replyTemplate.send((Message<?>) v);
+							this.replyTemplate.send(checkHeaders(v, topic, source));
```
NathanQingyangXu, line 494: I hope my above change makes sense, not very sure though

NathanQingyangXu:
> Can you, please, stop squashing your commits? It is really much easier to review if there is a history of commits in the PR. And sometime we just review the latest commit which might not include changes for other files. When we merge PR, we squash to preserve only one commit in the `main` for the issue we are resolving.
> 
> Thanks for understanding!

a little bit late to read this comment. yeah, for sure. I totally understand. Will keep that in mind from now on. thanks.
btw, I am not very confident about my changes in this PR (seems a big overlooking if valid). I am a developer whose job requires good Spring Kafka experience and I think contributing directly is the best way to learn. Thanks for your patience and good product.

NathanQingyangXu:
> @NathanQingyangXu ,
> 
> thank you very much for so thoughtful contributions!
> 
> Would you mind double checking if PR build failure is related somehow to your changes: https://github.com/spring-projects/spring-kafka/actions/runs/6903843623/job/18783345784?pr=2904 ?

as usual, unit testing cases protected. Fixed the bug and now they would not complain.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
artembilan: Let's leave it wit this assumption for now!
We can look into that later.
Let's concentrate on missed headers logic in this your PR!

artembilan (changes requested):
Looks like this fix introduces some new functionality.
Any chances to have ti covered with unit test, please?

NathanQingyangXu:
New unit testing cases are desperately needed. My current concern is to
confirm my PR validity. Will add testing cases soon.

NathanQingyangXu:
During unit testing case development, I found this PR is invalid. For the following code snippet:
```
/**
* Return the default expression when no SendTo value is present.
* @return the expression.
* @since 2.2.15
*/
public static String getDefaultReplyTopicExpression() {
return PARSER_CONTEXT.getExpressionPrefix() + "source.headers['"
		+ KafkaHeaders.REPLY_TOPIC + "']" + PARSER_CONTEXT.getExpressionSuffix();
}
```
so if no reply topic is specified, the above default expression will ensure we got the value from `KafkaHeaders.REPLY_TOPIC` header.
There are other minior issues valid in this PR, but given the gist of it is invalid, I close this PR for now.

PR #2906, opened 2023-11-17
Improve Documentation about Acknowledgment.nack()

powibol:
From the current documentation, it was not clear to me what is the effect of calling `Acknowledgment.nack()` when a listener is assigned to multiple topics and/or partitions.

Looking at the implementation, I believe that calling `nack(Duration sleep)` will pause the entire listener for the specified sleep duration. Another option would have been that is only pauses reading from one partition or one topic, which other topics or partitions are still processed.

With this PR, I attempt to make this more clear on the documentation page and in the JavaDoc of `Acknowledgment.nack()`. 

Please feel free to discuss, improve and correct me in case I'm wrong. Thank you!


garyrussell:
This behavior is as-designed. Typically, `nack()` is used for transient errors (e.g. DB is down), in which case processing records from other partitions won't work either. Adding partition-level nacks would add quite a bit of complexity to an already complex component.

garyrussell (approved):

PR #2907, opened 2023-11-18
fix bug topic header is missing when ReplyingKafkaTemplate's listener's method returns Iterable<Message<?>>

NathanQingyangXu:
unit testing case attached showcasing the bug. without code changes, the new testing case would fail with Assert failure due to topic header missing.


NathanQingyangXu:
woa, found it is intentional as per doc:

> If the listener method returns Message<?> or Collection<Message<?>>, the listener method is responsible for setting up the message headers for the reply

so I close this PR.

PR #2908, opened 2023-11-18
fix some defects of setting default Message headers in MessagingMessageListenerAdapter

NathanQingyangXu:
This PR aims to fix some minor defects in MessagingMessageListenerAdapter, including:

- improved the `checkHeaders()` method
- added the logic to apply `checkHeaders()` to Message element in Collection type return of method annotated with `@SendTo`
- added a unit testing case covering default header setting for `Iterable<Message<?>>` return type

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
@@ -484,8 +484,9 @@ public abstract class MessagingMessageListenerAdapter<K, V> implements ConsumerS
 				}
 				if (iterableOfMessages || this.splitIterables) {
 					((Iterable<V>) result).forEach(v -> {
-						if (v instanceof Message) {
-							this.replyTemplate.send((Message<?>) v);
+						if (v instanceof Message<?> mv) {
+							Message<?> aReply = checkHeaders(mv, topic, source);
+							this.replyTemplate.send(aReply);
 						}
 						else {
 							this.replyTemplate.send(topic, v);
@@ -506,9 +507,9 @@ public abstract class MessagingMessageListenerAdapter<K, V> implements ConsumerS
 		MessageHeaders headers = reply.getHeaders();
 		boolean needsTopic = topic != null && headers.get(KafkaHeaders.TOPIC) == null;
 		boolean sourceIsMessage = source instanceof Message;
-		boolean needsCorrelation = sourceIsMessage && headers.get(this.correlationHeaderName) == null
+		boolean needsCorrelation = headers.get(this.correlationHeaderName) == null && sourceIsMessage
 				&& getCorrelation((Message<?>) source) != null;
-		boolean needsPartition = sourceIsMessage && headers.get(KafkaHeaders.PARTITION) == null
+		boolean needsPartition = headers.get(KafkaHeaders.PARTITION) == null && sourceIsMessage
 				&& getReplyPartition((Message<?>) source) != null;
 		if (needsTopic || needsCorrelation || needsPartition) {
 			MessageBuilder<?> builder = MessageBuilder.fromMessage(reply);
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -514,11 +515,11 @@ private Message<?> checkHeaders(Object result, String topic, @Nullable Object so
 			if (needsTopic) {
 				builder.setHeader(KafkaHeaders.TOPIC, topic);
 			}
-			if (needsCorrelation && sourceIsMessage) {
```
NathanQingyangXu, line 517: seems the second condtion has been implied in the first condition

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -514,11 +515,11 @@ private Message<?> checkHeaders(Object result, String topic, @Nullable Object so
...
+			if (needsCorrelation) {
 				builder.setHeader(this.correlationHeaderName,
 						((Message<?>) source).getHeaders().get(this.correlationHeaderName));
 			}
-			if (sourceIsMessage && reply.getHeaders().get(KafkaHeaders.REPLY_PARTITION) == null) {
```
NathanQingyangXu, line 521: confusing. should `REPLY_PARTITION` be relevant on `source` side? On `reply` side, seems only `PARTITION` header matters.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -501,24 +502,24 @@ else if (result instanceof Message) {
...
+	private Message<?> checkHeaders(Message<?> reply, @Nullable String topic, @Nullable Object source) { // NOSONAR (complexity)
 		MessageHeaders headers = reply.getHeaders();
-		boolean needsTopic = headers.get(KafkaHeaders.TOPIC) == null;
+		boolean needsTopic = topic != null && headers.get(KafkaHeaders.TOPIC) == null;
 		boolean sourceIsMessage = source instanceof Message;
-		boolean needsCorrelation = headers.get(this.correlationHeaderName) == null && sourceIsMessage;
-		boolean needsPartition = headers.get(KafkaHeaders.PARTITION) == null && sourceIsMessage
+		boolean needsCorrelation = sourceIsMessage && headers.get(this.correlationHeaderName) == null
+				&& ((Message<?>) source).getHeaders().get(this.correlationHeaderName) != null;
+		boolean needsPartition = sourceIsMessage && headers.get(KafkaHeaders.PARTITION) == null
```
NathanQingyangXu, line 511: I moved `xx != null` testing to the beginning for it is much more light-weight than `headers.get(...) != null`, so shot-circuiting might avoid some unnecessary cpu cycles. Someone even claims such nullness checking involves almost zero cost in openJDK.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -501,24 +502,24 @@ else if (result instanceof Message) {
 		}
 	}
 
-	private Message<?> checkHeaders(Object result, String topic, @Nullable Object source) { // NOSONAR (complexity)
-		Message<?> reply = (Message<?>) result;
+	private Message<?> checkHeaders(Message<?> reply, @Nullable String topic, @Nullable Object source) { // NOSONAR (complexity)
```
NathanQingyangXu, line 505: I think choosing `Message<?>` as the first argument type makes more sense, for this method is only meant to be used with that precondition.

NathanQingyangXu: not sure whether we can use `Nullable` to `topic`, but from its invocation context, seems we could never rule out the nullness possibility?

sobychacko:
@NathanQingyangXu Thanks for the PR! It is now merged upstream.

NathanQingyangXu:
> @NathanQingyangXu Thanks for the PR! It is now merged upstream.

thanks. I really appreciate that the PR got merged finally.

PR #2909, opened 2023-11-20
eliminate the usage of StringBuffer

NathanQingyangXu:
There are two usages of `StringBuffer`. This PR aims to either replace it with more appropriate and efficient `StringBuilder` or simply eliminate its usage if better alternative exists.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
@@ -22,6 +22,8 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.function.BiConsumer;
+import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
```diff
@@ -189,12 +191,9 @@ private static void listen(List<RetryListener> listeners, ConsumerRecords<?, ?>
 	 * @return the String.
 	 */
 	public static String recordsToString(ConsumerRecords<?, ?> records) {
-		StringBuffer sb = new StringBuffer();
-		records.spliterator().forEachRemaining(rec -> sb
-				.append(KafkaUtils.format(rec))
-				.append(','));
-		sb.deleteCharAt(sb.length() - 1);
```
NathanQingyangXu, line 196: the above implementation also throws exception when `records` is an empty one

garyrussell (approved):

PR #2913, opened 2023-11-23
GH-2852: support fixed transaction id suffix

### LINKED ISSUE #2852: Provide a way to customize a transactionIdSuffix
**Expected Behavior**

Provide strategy to customize a `transactionIdSuffix`.

**Current Behavior**

Currently, spring-kafka use [AtomicInteger](https://github.com/spring-projects/spring-kafka/blob/61bc2190eb592515e95d322f6d16b359cb5134d0/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java#L124) for transactionIdSuffix.

**Context**

In my environment, I work with restricted `transactionId` within a specific range(e.g. **topic-{1-5}**).
Currently, spring-kafka use [AtomicInteger](https://github.com/spring-projects/spring-kafka/blob/61bc2190eb592515e95d322f6d16b359cb5134d0/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java#L124) for `transactionIdSuffix`, it would be beneficial to have the flexibility to implement a custom strategy for `transactionIdSuffix`.
This could involve actions such as resetting the counter after reaching a limit or a storage for managing `transactionsIds`.

I'm happy to contribute a PR once I know what your opinion on this feature is.

Wzy19930507:
Add properties `maxCache` at `DefaultKafkaProducerFactory`, setting `maxCache` greater than zero can reuse `transactional.id`.

If maxCache is 3, the suffix corresponding to the first prefix is 0-2, corresponding to the second prefix is 3-5.

Add new interface to initialize transactionIdSuffix.

Resolves #2852

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
@@ -102,10 +102,39 @@ NOTE: If there is a `KafkaTransactionManager` (or synchronized) transaction in p
 Instead, a new "nested" transaction is used.
 
 [[transaction-id-prefix]]
-== `transactionIdPrefix`
+== `TransactionIdPrefix`
 
 With `EOSMode.V2` (aka `BETA`), the only supported mode, it is no longer necessary to use the same `transactional.id`, even for consumer-initiated transactions; in fact, it must be unique on each instance the same as for producer-initiated transactions.
 This property must have a different value on each application instance.
+
+[[transaction-id-suffix-fixed]]
+== `TransactionIdSuffix Fixed`
+
+Since 3.1, when setting `maxCache` greater than zero can reuse `transactional.id` within a specific range.
+When a transaction producer is requested and `transactional.id` all in use, throw a `NoProducerAvailableException`.
+User can use then use a RetryTemplate configured to retry that exception, with a suitably configured back off.
+
+[source, java]
+----
+public static class Config {
+
+    @Bean
+    public ProducerFactory<String, String> myProducerFactory() {
+        Map<String, Object> configs = producerConfigs();
+        configs.put(ProducerConfig.CLIENT_ID_CONFIG, "myClientId");
+        ...
+        DefaultKafkaProducerFactory<String, String> pf = new DefaultKafkaProducerFactory<>(configs);
+        ...
+        pf.setTransactionIdPrefix("my.txid.");
+        pf.setMaxCache(5);
+        return pf;
+    }
+
+}
+----
+When setting `maxCache` to 5, `transactional.id` is `my.txid.`++`{0-4}`+.
+
+IMPORTANT: When use `KafkaTransactionManager` in the `ConcurrentMessageListenerContainer`, `maxCache` must be greater than `concurrency`, also be careful nested transaction.
 
 [[tx-template-mixed]]
 == `KafkaTemplate` Transactional and non-Transactional Publishing
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
@@ -108,11 +108,11 @@ With `EOSMode.V2` (aka `BETA`), the only supported mode, it is no longer necessa
 This property must have a different value on each application instance.
 
 [[transaction-id-suffix-fixed]]
-== `TransactionIdSuffix Fixed`
+== `TransactionIdSuffix` Fixed
 
 Since 3.1, when setting `maxCache` greater than zero can reuse `transactional.id` within a specific range.
 When a transaction producer is requested and `transactional.id` all in use, throw a `NoProducerAvailableException`.
-User can use then use a RetryTemplate configured to retry that exception, with a suitably configured back off.
+We can then use a RetryTemplate configured to retry that exception, with a suitably configured back off.
 
 [source, java]
 ----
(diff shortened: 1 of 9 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -846,7 +861,20 @@ protected Producer<K, V> createTransactionalProducer(String txIdPrefix) {
...
+				BlockingQueue<String> suffixQueue = getSuffixCache(txIdPrefix, this.maxCache);
+				Assert.notNull(suffixQueue,
+						() -> "No suffix cache found for " + txIdPrefix + ", max cache" + this.maxCache);
+				suffix = suffixQueue.poll();
+				if (suffix == null) {
+					throw new NoProducerAvailableException("No transaction producer available for " + txIdPrefix);
+				}
+			}
+			else {
+				suffix = String.valueOf(this.transactionIdSuffix.getAndIncrement());
```
stillya, line 875: Nice work, and thanks for your effort! However, I think just incrementing may not be sufficient. The goal of [GH-2852](https://github.com/spring-projects/spring-kafka/issues/2852) is to allow users to set how `transactionIds` are issued. For example, with a topic having 9 partitions, 3 replicas of my service (each with 3 producers), and limits on 9 `txIds`, restarting my service becomes a bit of a pain. It takes time to distribute transaction IDs correctly(without getting `Fenced` all the time), so instead of that, I would prefer to use external storage as a source of truth.
I suggest introducing a new interface for generating `transactionIdSuffix` with a default increment implementation. WDYTH?

Wzy19930507: > I suggest introducing a new interface for generating transactionIdSuffix with a default increment implementation.
I like the proposal, but there are some scenarios to consider.

Because of maxCache can be modified dynamically, if we didn't set maxCache to begin with
, maybe suffix{0-9} was used. The suffix needs a fixed value to avoid `transaction.id` conflicts.

On the other hand, when it comes to reset(), some transaction ids are in use, we can't recycle them until they close.

The new interface may need to be restricted to KafkaTemplate scenarios where the listenerContainer is reset but some transaction ids are still unavailable, causing conumser to report an error.

Wzy19930507: The first service startup suffix increments from 0 and the second service startup increments from 9?

stillya: > Via the new interface, does the first instance of the service increment transactionIdSuffix from 0, and the second instance increments it from 3?

It actually depends on the implementation. It could distribute `transactionIdSuffix` between replicas as described, or it could provide the first free suffix from the `source of truth storage`. However, the default implementation remains the same as it is now.

> At the same time, we need to monitor the generated transaction.id. For example, provide a method to return the current transactionIdSuffix or publish an event carrying the current transactionIdSuffix at DefaultKafkaProducerFactory.destroy().

Certainly, we need this functionality. Therefore, the interface must include a contract to retrieve the `transactionId` and return it back.

> As an alternative, writing a unique ID in the transactionIdPrefix, such as UUID, is feasible?


In my environment, it's not an option because I have a mask for `transactionIds` like **topic-name-{0..n}**(e.g. mytopic-1).

Wzy19930507: > At the same time, we need to monitor the generated transaction.id. For example, provide a method to return the current transactionIdSuffix or publish an event carrying the current transactionIdSuffix at DefaultKafkaProducerFactory.destroy().

As a supplement,  interface `ProducerFactory.addListener` can subscribe all producer.

sobychacko:
Hi @Wzy19930507, I am taking a look at this PR. Could you let me know if you added the new interface that @stillya suggested? I don't see that in the PR yet.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
```diff
@@ -102,11 +102,40 @@ NOTE: If there is a `KafkaTransactionManager` (or synchronized) transaction in p
...
 
 With `EOSMode.V2` (aka `BETA`), the only supported mode, it is no longer necessary to use the same `transactional.id`, even for consumer-initiated transactions; in fact, it must be unique on each instance the same as for producer-initiated transactions.
 This property must have a different value on each application instance.
 
+[[transaction-id-suffix-fixed]]
+== `TransactionIdSuffix Fixed`
+
+Since 3.1, when setting `maxCache` greater than zero can reuse `transactional.id` within a specific range.
+When a transaction producer is requested and `transactional.id` all in use, throw a `NoProducerAvailableException`.
+User can use then use a RetryTemplate configured to retry that exception, with a suitably configured back off.
```
sobychacko, line 115: `We can then use a`

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
```diff
@@ -102,11 +102,40 @@ NOTE: If there is a `KafkaTransactionManager` (or synchronized) transaction in p
...
+        pf.setMaxCache(5);
+        return pf;
+    }
+
+}
+----
+When setting `maxCache` to 5, `transactional.id` is `my.txid.`++`{0-4}`+.
+
+IMPORTANT: When use `KafkaTransactionManager` in the `ConcurrentMessageListenerContainer`, `maxCache` must be greater than `concurrency`, also be careful nested transaction.
+
```
sobychacko, line 138: "also be careful about".
It would be beneficial if you could add a few sentences detailing why the `maxCache` needs to be greater than `concurrency`. Same thing about nested transactions.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -424,6 +424,15 @@ public final void setTransactionIdPrefix(String transactionIdPrefix) {
 		return this.transactionIdPrefix;
 	}
 
+	public final boolean initTransactionIdSuffix(int transactionIdSuffix) {
+		Assert.isTrue(transactionIdSuffix >= 0, "'transactionIdSuffix' initial value must greater than or equal 0");
+		return this.transactionIdSuffix.compareAndSet(0, transactionIdSuffix);
+	}
```
Wzy19930507, line 430: Hi, @sobychacko new interface is here. `DefaultKafkaProducerFactory` use only

stillya: Thank you, but I didn't quite mean that, so I've decided to lend a hand. Could you please take a [look](https://github.com/spring-projects/spring-kafka/pull/2930)?

Issue #2924, opened 2023-12-07
Unable to send null payload with KafkaTemplate#send(Message)  


pswrdf:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**
3.0.11

**Describe the bug**

Unable to send null message through `KafkaTemplate#send(Message)`.
Unable to create `GenericMessage` with null payload.

**To Reproduce**
```
spring:
  kafka:
    producer:
      value-serializer: org.apache.kafka.common.serialization.ByteArraySerializer
```
```
    @Bean
    public RecordMessageConverter jsonMessageConverter(ObjectMapper objectMapper) {
        return new ByteArrayJsonMessageConverter(objectMapper);
    }
```
```
       final var Object myValue = null;
       final var message = MessageBuilder.withPayload(myValue).setHeader(TOPIC, topic).build();
        kafkaTemplate.send(message);
```

**Expected behavior**

Kafka sends null message.
But we get: `IllegalArgumentException: Payload must not be null`

We could set `myValue=KafkaNull.INSTANCE;`
But JsonMessageConverter inherited classes cannot proccess KafkaNull.

jakabgyonci97:
Hello, I have a somewhat similar issue in version 3.1.1.
I'm writing a kafka message using _KafkaTemplate#send(Message)_. In my case there are different headers that I attach to the message, some value value null. When the send operation is performed I get the following error:

> java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "rawValue" is null

PR #2925, opened 2023-12-07
 Gh-2924 Fix send null payload with KafkaTemplate#send(Message), and …

pswrdf:
…JsonMessageConverter

fixes: https://github.com/spring-projects/spring-kafka/issues/2924

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/converter/JsonMessageConverter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/converter/JsonMessageConverter.java
@@ -93,12 +93,8 @@ public class JsonMessageConverter extends MessagingMessageConverter {
 
 	@Override
 	protected Object convertPayload(Message<?> message) {
-		Object payload = super.convertPayload(message);
-		if (payload != null) {
-			throw new UnsupportedOperationException("Select a subclass that creates a ProducerRecord value "
-					+ "corresponding to the configured Kafka Serializer");
-		}
-		return payload;
+		throw new UnsupportedOperationException("Select a subclass that creates a ProducerRecord value "
+				+ "corresponding to the configured Kafka Serializer");
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


sobychacko:
@pswrdf Thanks for the PR! Can you add a unit test to verify the changes? Or point to an existing test where this code is exercised.

artembilan (changes requested):
Hi @pswrdf !

Thank you for contribution.
Please, make yourself familiar with commit message rules we try to follow: https://cbea.ms/git-commit/.

I don't say that you need to change anything right now if you don't want, but just for the future to avoid comments from me like this 😉 

@sobychacko ,

This fix is supposed to be back-ported to `3.0.x`.
When you merge this, please, consider to include `Fixes: gh-2924` like into commit message.
This way Backport Bot GHA workflow will create a respective back-port issue for for current `3.0.x` milestone. However don't merge yet, please, I'm working in that branch right now to migrate it into GHA.

artembilan (changes requested):
So, now CI/CD infra is set up for `3.0.x`.
Therefore we are good to cherry-pick after merging this.

1. Add some unit test, please. See `BatchListenerConversionTests` where we use one of those `JsonMessageConverter` impls.
2. Add your name to the `@author` list of all the affected classes.
3. Add also `Fixes: gh-2924` into next commit message as I asked before.
4. Add `**Cherry-pick to `3.0.x`**` into the next commit message.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/support/converter/JsonMessageConverter.java
sobychacko, line 1: Missing `author` tag in this class.

artembilan: There is no need. 
This class has lost its changes.

pswrdf: it was reverted in the second commit.

I think it's done.

pswrdf:
I think it's done.

Would I squash commits? It needs to change email also.

artembilan:
... and cherry-picked to `3.0.x`

Issue #2800, opened 2023-09-07
Non-Blocking Retries: Consider Making it Easier to Route to Different DLTs by Exception Type


garyrussell:
Currently requires some tricky code to capture the DLT properties; see https://stackoverflow.com/questions/77058404/custom-deadletterpublishingrecoverer-is-not-used-when-errors-occurs-during-valid/77060340#77060340

breader124:
Hi, I'd like to start working on it if possible

garyrussell:
Sure, please go ahead.

breader124:
I've just opened the PR with changes aimed to introduce the described feature: #2929

PR #2929, opened 2023-12-10
GH-2800: Make it possible to route messages to custom DLT based on exceptions thrown

breader124:
This PR is related to #2800. It introduces changes making it easy for developers to route messages to different DLTs when the previously configured exception is thrown.


sobychacko:
@breader124 Thanks for the PR. I am going to take a look reviewing the changes. One general comment. Could you add some docs snippets related to the change? This seems like a bigger change that deserves some notes in the reference docs.

breader124:
Sure, I was thinking about modifications in the reference docs as well. The only thing I'd like to wait for until I start working on the docs is the confirmation of the API part of changes. Having the API stable I'll propose docs changes

sobychacko:
Since this PR is large in terms of the changes, could you provide a brief overview of the changes in a few sentences? Please provide a flow of the changes, the APIs you add, etc. It doesn't have to be very elaborate, but very briefly.

breader124:
The main difference that I introduced is the possibility to configure the routing to the custom DLTs based on the exceptions thrown during the message processing. This configuration can be provided in 2 ways (and these are the only new APIs):
- Using the `org.springframework.kafka.annotation.RetryableTopic#exceptionBasedDltRouting`. There you need to pass the array of routing rules (`org.springframework.kafka.retrytopic.ExceptionBasedDestinationDlt`). Each of them specifies what suffix extension should be used when creating the custom DLT and which exceptions should be handled by it
- Using the `org.springframework.kafka.retrytopic.RetryTopicConfigurationBuilder#dltRoutingRules` method, where you pass the Map describing the custom DLTs routing rules.

When the configuration is provided, then depending on the `autoCreateTopics` config property new DLTs might be created together with rest of the topic automatically. When these DLTs are being created, then their properties are placed before the properties of the generic purpose DLT inside the `org.springframework.kafka.retrytopic.DestinationTopicPropertiesFactory#createPropertiesForDefaultTopicStrategy` method. Thanks to this change they are considered earlier for routing than the generic purpose DLT is considered.

Lastly, when it comes to the actual routing part, I've modified the `org.springframework.kafka.retrytopic.DefaultDestinationTopicResolver`. There I try to match the exception, which has been thrown during the message processing with the ones configured for the particular custom DLT. This process traverse causes and it's not configurable now. When I'm not able to find any custom, dedicated DLT, then finally the processing comes to the generic purpose DLT and message will land there.

Besides all changes described above I also provided tests for the changed behaviour.

artembilan (changes requested):
This sounds reasonable, but unfortunately it introduces some breaking changes to public API. To avoid immediate upgrade in the target projects we cannot accept this change for the current patch version.

We are fully OK to schedule this for the next `3.2` generation, but not earlier.

Therefore your PR will sit here for a while.

Thanks for understanding.

sobychacko:
That's a great point, Artem. @ breader124 If you still want to go into the patch release, could you still make these changes without breaking an existing API? It might still be okay to add new API's as we have done in some other existing PR's.

breader124:
It's totally fine for me to wait for a while @artembilan, @sobychacko. I'll update the `@since` tags in the Javadocs then to make them point to the 3.2.0 version instead of the 3.1.1

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ExceptionBasedDestinationDlt.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Annotation allowing to specify additional DLT which will be chosen when message
+ * processing caused the configured exception to be thrown.
+ *
+ * @author Adrian Chlebosz
+ * @see org.springframework.kafka.retrytopic.ExceptionBasedDltRouting
+ * @since 3.2.0
+ */
+public @interface ExceptionBasedDestinationDlt {
```
sobychacko, line 27: Maybe `ExceptionBasedDltDestination`?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DefaultDestinationTopicResolver.java
```diff
@@ -144,18 +155,44 @@ public DestinationTopic getDestinationTopicByName(String mainListenerId, String
...
+			return false;
+		}
+
+		boolean isDltIntendedForCurrentExc = destination.usedForExceptions().stream()
+			.anyMatch(excType -> isDirectExcOrCause(e, excType));
+		boolean isGenericPurposeDlt = destination.usedForExceptions().isEmpty();
+		return isDltIntendedForCurrentExc || isGenericPurposeDlt;
+	}
+
+	private static boolean isDirectExcOrCause(Exception e, Class<? extends Throwable> excType) {
```
sobychacko, line 183: Is there a use case in which the traversal of the `cause` chain is necessary? Should we enable/disable this via config?

sobychacko:
@breader124 I did a first pass at the PR. It looks good. One general question: The corresponding issue that was created by the [SO thread ](https://stackoverflow.com/questions/77058404/custom-deadletterpublishingrecoverer-is-not-used-when-errors-occurs-during-inval/77060340#77060340), originally trying to route records from deserialization exceptions. Is that use case addressed by this PR?

breader124:
Yes, it's possible to configure the `@RetryableTopic` the way how it was done in one the tests (https://github.com/spring-projects/spring-kafka/pull/2929/files#diff-9ffe03fb2bdcbda33d0d0d2cd23f8ae88d5c7d5c3219296eb7c3e38e9e643aabR373). Then 
messages, which caused `DeserializationException`s will be routed to the different DLT

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DefaultDestinationTopicResolver.java
breader124: I've implemented it this way, because I think it's a natural behaviour, which might be expected by developers. Let's consider a situation, when someone wants to have a separate DLT per business exception. When this exception is wrapped by `TimestampedException` by Spring Kafka, then custom DLT routing won't kick in if we don't traverse causes. That's because mentioned business exception is one level deeper than the `TimestampedException`, it's its cause, which needs to be looked for.

If you find this important to provide the option to configure this causes traversing behaviour, then I'm open to implement it. I didn't want to introduce it from the beginning, because I'd like to avoid potentially unnecessary complexity.

artembilan:
@breader124 ,

Thanks

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopic.java
```diff
@@ -188,13 +196,15 @@ public Properties(Properties sourceProperties, String suffix, Type type) {
 		 * @param shouldRetryOn the exception classifications.
 		 * @param timeout the timeout.
 		 * @param autoStartDltHandler whether or not to start the DLT handler.
+		 * @param usedForExceptions the exceptions which destination is intended for
 		 * @since 2.8
 		 */
 		public Properties(long delayMs, String suffix, Type type,
```
artembilan, line 202: I suggest to have one more public ctor.
This way we won't have a breaking change in the point version.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopicContainer.java
```diff
@@ -65,9 +66,10 @@ public interface DestinationTopicContainer {
 	 * DLT for the given topic, or null if none is found.
 	 * @param mainListenerId the listener id.
 	 * @param topicName the topic name for which to look the DLT for
+	 * @param exc the exception which is being handled
 	 * @return The {@link DestinationTopic} instance corresponding to the DLT.
 	 */
 	@Nullable
-	DestinationTopic getDltFor(String mainListenerId, String topicName);
+	DestinationTopic getDltFor(String mainListenerId, String topicName, Exception exc);
```
artembilan, line 73: Cannot we have another method to avoid a breaking change?
It also would be great to have it as a `default` to avoid a immediate upgrade compilation problem.
We may have the current method deprecated then.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopicPropertiesFactory.java
```diff
@@ -86,7 +94,8 @@ public DestinationTopicPropertiesFactory(String retryTopicSuffix, String dltSuff
 			DltStrategy dltStrategy,
 			TopicSuffixingStrategy topicSuffixingStrategy,
 			SameIntervalTopicReuseStrategy sameIntervalTopicReuseStrategy,
-			long timeout) {
+			long timeout,
+			Map<String, Set<Class<? extends Throwable>>> dltRoutingRules) {
```
artembilan, line 98: Cannot we have another ctor with a delegating from this one using an empty map?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ExceptionBasedDltDestination.java
```diff
@@ -0,0 +1,48 @@
...
+
+package org.springframework.kafka.retrytopic;
+
+/**
+ * Annotation allowing to specify additional DLT which will be chosen when message
+ * processing caused the configured exception to be thrown.
+ *
+ * @author Adrian Chlebosz
+ * @see org.springframework.kafka.retrytopic.ExceptionBasedDltRouting
+ * @since 3.2.0
```
artembilan, line 25: I think these two has to be swapped according to the Checkstyle rules:
```
		<module name="AtclauseOrder">
			<property name="target" value="CLASS_DEF, INTERFACE_DEF, ENUM_DEF"/>
			<property name="tagOrder" value="@param, @author, @since, @see, @version, @serial, @deprecated"/>
		</module>
```

Also it would be great to have an `@author` in its own paragraph to really make it visible for an honor.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ExceptionBasedDltDestination.java
```diff
@@ -0,0 +1,48 @@
...
+
+	/**
+	 * Suffix extension used when constructing the name for the new DLT. It is placed
+	 * before the main suffix configured through the
+	 * ${@link org.springframework.kafka.annotation.RetryableTopic#dltTopicSuffix()}, so the
+	 * final name is the product of these two.
+	 *
+	 * @return the configured suffix extension
+	 */
+	String suffix();
```
artembilan, line 37: I wonder if this is really required and can be just empty by default...

Also: no blank lines in the method Javadocs: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ExceptionBasedDltRouting.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Annotation allowing to specify the custom DLT routing rules steered by exceptions
+ * which might be thrown during the processing.
+ *
+ * @author Adrian Chlebosz
+ * @see ExceptionBasedDltDestination
+ * @since 3.2.0
+ */
+public @interface ExceptionBasedDltRouting {
```
artembilan, line 27: Why do we need this annotation if `@RetryableTopic` could just have a property like that below?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
```diff
@@ -519,6 +526,20 @@ private BinaryExceptionClassifierBuilder classifierBuilder() {
...
+	 * Configure to set DLT routing rules causing the message to be redirected to the custom
+	 * DLT when the configured exception has been thrown during message processing.
+	 * The cause of the originally thrown exception will be traversed in order to find the
+	 * match with the configured exceptions.
+	 * @param dltRoutingRules specification of custom DLT name extensions and exceptions which should be matched for them
+	 * @return the builder
+	 * @since 3.2.0
+	 */
+	public RetryTopicConfigurationBuilder dltRoutingRules(Map<String, Set<Class<? extends Throwable>>> dltRoutingRules) {
+		this.dltRoutingRules = dltRoutingRules;
```
artembilan, line 539: This is not OK. Has to be a copy.
The point is to avoid internal mutation when that external map is edited.
Well, that `dltRoutingRules` could be final and here you could just use `putAll()`

artembilan (changes requested):
I also believe that this feature deserves some docs in the `\spring-kafka\spring-kafka-docs\src\main\antora\modules\ROOT\pages\retrytopic\` dir.

But yeah, that can be done in the end when we agree with all the changes from the code perspective.

Thanks

breader124:
Thanks for the review. I've just pushed the rebased code and will apply the changes you've suggested in the review over the weekend together with the docs changes suggestions

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ExceptionBasedDltDestination.java
breader124: From my point of view having the `suffix` required makes this API harder to misuse by design. `exceptions` is required, but when someone will be allowed to specify exceptions without the `suffix`, then it means that the general purpose DLT will be used even for this exc. I'm not sure when it could be an intended usage

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/ExceptionBasedDltRouting.java
breader124: It's not needed, I'll delete it

breader124:
@artembilan, @sobychacko, I've just rebased the code on top of the most recent `main` and pushed it. All the suggestions you had are now introduced and I think that the code is ready for the final round of the review

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopic.java
```diff
@@ -171,6 +173,13 @@
 	 */
 	String dltTopicSuffix() default RetryTopicConstants.DEFAULT_DLT_SUFFIX;
 
+	/**
+	 * The DLT routing allowing to redirect the message to the custom DLT based on the
+	 * exception thrown during the processing.
+	 * @return the exception based DLT routing
```
artembilan, line 179: `@since 3.2.0`

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
```diff
@@ -148,6 +153,8 @@ public RetryTopicConfiguration processAnnotation(String[] topics, Method method,
 				.autoStartDltHandler(autoStartDlt)
 				.setTopicSuffixingStrategy(annotation.topicSuffixingStrategy())
 				.sameIntervalTopicReuseStrategy(annotation.sameIntervalTopicReuseStrategy())
+				.timeoutAfter(timeout)
```
artembilan, line 156: Duplicated call. See bellow your new `dltRoutingRules`.

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/DestinationTopicPropertiesFactory.java
```diff
@@ -96,6 +131,7 @@ public DestinationTopicPropertiesFactory(String retryTopicSuffix, String dltSuff
 		this.sameIntervalTopicReuseStrategy = sameIntervalTopicReuseStrategy;
 		this.timeout = timeout;
 		this.destinationTopicSuffixes = new DestinationTopicSuffixes(retryTopicSuffix, dltSuffix);
+		this.dltRoutingRules = dltRoutingRules;
```
artembilan, line 134: This is not OK. Has to be a copy of the map.
This way we will avoid the problem when provided map is mutated somewhere outside.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/retrytopic/features.adoc
```diff
@@ -261,3 +261,41 @@ protected Consumer<DeadLetterPublishingRecovererFactory>
...
 
+[[exc-based-custom-dlt-routing]]
+== Routing of messages to custom DLTs based on thrown exceptions
+
+Starting with version 3.2.0, it's possible to route messages to custom DLTs based on the type of the exception, which has been thrown during their processing.
+In order to do that, you need to specify the routing.
+Routing customization consists of the specification of the additional destinations.
+Destinations in turn consist of two settings: the `suffix` and `exceptions`.
+When the exception type specified in `exceptions` has been thrown, the DLT containing the `suffix` will be considered as the target topic for the message before the general purpose DLT is considered.
+Take a look at the examples:
```
artembilan, line 272: I prefer to have an official neutral language in docs where it is depersonalized.
So, no phrases like `you`, `me`, `take a look` etc.

But... That's my personal preferences and I know there are many like that throughout the doc.

Also: this need to me mentioned in the `whats-new.adoc`

artembilan (changes requested):

sobychacko:
@breader124 There was a recent change made via another PR that modified `DestinationTopicPropertiesFactory.` Could you rebase your PR branch against the latest `main`?

breader124:
Sure, will do it and push tomorrow the latest

breader124:
@sobychacko, it's now rebased on top of the current `main`, all previously existing conflicts were solved

sobychacko:
@breader124 Thank you for this excellent PR. It is now merged upstream.

PR #2930, opened 2023-12-10
GH-2852: Provide a way to customize a transactionIdSuffix

stillya:
Introduce a new `TransactionSuffixManager` interface with `DefaultTransactionSuffixManager` implementation to restrict `transaction.id` in range for existing [PR](https://github.com/spring-projects/spring-kafka/pull/2913)

Resolves https://github.com/spring-projects/spring-kafka/issues/2852

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
@@ -126,8 +126,7 @@ public static class Config {
         ...
         DefaultKafkaProducerFactory<String, String> pf = new DefaultKafkaProducerFactory<>(configs);
         ...
-        DefaultTransactionIdSuffixStrategy ss = new DefaultTransactionIdSuffixStrategy();
-        ss.setMaxCache(5);
+        TransactionIdSuffixStrategy ss = new DefaultTransactionIdSuffixStrategy(5);
         pf.setTransactionIdSuffixStrategy(ss);
         return pf;
     }
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
@@ -102,10 +102,43 @@ NOTE: If there is a `KafkaTransactionManager` (or synchronized) transaction in p
 Instead, a new "nested" transaction is used.
 
 [[transaction-id-prefix]]
-== `transactionIdPrefix`
+== `TransactionIdPrefix`
 
 With `EOSMode.V2` (aka `BETA`), the only supported mode, it is no longer necessary to use the same `transactional.id`, even for consumer-initiated transactions; in fact, it must be unique on each instance the same as for producer-initiated transactions.
 This property must have a different value on each application instance.
+
+[[transaction-id-suffix-fixed]]
+== `TransactionIdSuffix Fixed`
+
+Since 3.2, a new `TransactionIdSuffixStrategy` interface was introduced to manage `transactional.id` suffix.
+The default implementation is `DefaultTransactionIdSuffixStrategy` when setting `maxCache` greater than zero can reuse `transactional.id` within a specific range, otherwise suffixes will be generated on the fly by incrementing a counter.
+When a transaction producer is requested and `transactional.id` all in use, throw a `NoProducerAvailableException`.
+User can then use a `RetryTemplate` configured to retry that exception, with a suitably configured back off.
+
+[source,java]
+----
+public static class Config {
+
+    @Bean
+    public ProducerFactory<String, String> myProducerFactory() {
+        Map<String, Object> configs = producerConfigs();
+        configs.put(ProducerConfig.CLIENT_ID_CONFIG, "myClientId");
+        ...
+        DefaultKafkaProducerFactory<String, String> pf = new DefaultKafkaProducerFactory<>(configs);
+        ...
+        DefaultTransactionIdSuffixStrategy ss = new DefaultTransactionIdSuffixStrategy();
+        ss.setMaxCache(5);
+        pf.setTransactionIdSuffixStrategy(ss);
+        return pf;
+    }
+
+}
+----
+When setting `maxCache` to 5, `transactional.id` is `my.txid.`++`{0-4}`+.
+
+IMPORTANT: When using `KafkaTransactionManager` with the `ConcurrentMessageListenerContainer` and enabling `maxCache`, it is necessary to set `maxCache` to a value greater than or equal to `concurrency`.
+If a `MessageListenerContainer` is unable to acquire a `transactional.id` suffix, it will throw a `NoProducerAvailableException`.
+When using nested transactions in the `ConcurrentMessageListenerContainer`, it is necessary to adjust the maxCache setting to handle the increased number of nested transactions.
 
 [[tx-template-mixed]]
 == `KafkaTemplate` Transactional and non-Transactional Publishing
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 the original author or authors.
+ * Copyright 2016-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -135,7 +135,7 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 
 	private final AtomicBoolean running = new AtomicBoolean();
 
-	private TransactionIdSuffixStrategy transactionIdSuffixStrategy = new DefaultTransactionIdSuffixStrategy();
+	private TransactionIdSuffixStrategy transactionIdSuffixStrategy = new DefaultTransactionIdSuffixStrategy(0);
 
 	private Supplier<Serializer<K>> keySerializerSupplier;
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
@@ -121,8 +121,6 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 
 	private final Map<String, Object> configs;
 
-	private final AtomicInteger transactionIdSuffix = new AtomicInteger();
-
 	private final Map<String, BlockingQueue<CloseSafeProducer<K, V>>> cache = new ConcurrentHashMap<>();
 
 	private final Map<Thread, CloseSafeProducer<K, V>> threadBoundProducers = new ConcurrentHashMap<>();
@@ -136,6 +134,8 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 	private final List<ProducerPostProcessor<K, V>> postProcessors = new ArrayList<>();
 
 	private final AtomicBoolean running = new AtomicBoolean();
+
+	private TransactionIdSuffixStrategy transactionIdSuffixStrategy = new DefaultTransactionIdSuffixStrategy();
 
 	private Supplier<Serializer<K>> keySerializerSupplier;
 
@@ -351,6 +351,16 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 	}
 
 	/**
+	 * Set the transaction suffix strategy.
+	 * @param transactionIdSuffixStrategy the strategy.
+	 * @since 3.2
+	 */
+	public void setTransactionIdSuffixStrategy(TransactionIdSuffixStrategy transactionIdSuffixStrategy) {
+		Assert.notNull(transactionIdSuffixStrategy, "'transactionIdSuffixStrategy' cannot be null");
+		this.transactionIdSuffixStrategy = transactionIdSuffixStrategy;
+	}
+
+	/**
 	 * If true (default), programmatically provided serializers (via constructor or
 	 * setters) will be configured using the producer configuration. Set to false if the
 	 * serializers are already fully configured.
@@ -404,7 +414,7 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 
 	/**
 	 * Set a prefix for the {@link ProducerConfig#TRANSACTIONAL_ID_CONFIG} config. By
-	 * default a {@link ProducerConfig#TRANSACTIONAL_ID_CONFIG} value from configs is used
+	 * default, a {@link ProducerConfig#TRANSACTIONAL_ID_CONFIG} value from configs is used
 	 * as a prefix in the target producer configs.
 	 * @param transactionIdPrefix the prefix.
 	 * @since 1.3
@@ -536,9 +546,9 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 		producerProperties.putAll(overrideProperties);
 		producerProperties = ensureExistingTransactionIdPrefixInProperties(producerProperties);
 		DefaultKafkaProducerFactory<K, V> newFactory = new DefaultKafkaProducerFactory<>(producerProperties,
-						getKeySerializerSupplier(),
-						getValueSerializerSupplier(),
-						isConfigureSerializers());
+				getKeySerializerSupplier(),
+				getValueSerializerSupplier(),
+				isConfigureSerializers());
 		newFactory.setPhysicalCloseTimeout((int) getPhysicalCloseTimeout().getSeconds());
 		newFactory.setProducerPerThread(isProducerPerThread());
 		for (ProducerPostProcessor<K, V> templatePostProcessor : getPostProcessors()) {
@@ -690,7 +700,7 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 		}
 		if (producerToClose != null) {
 			try {
-				producerToClose.closeDelegate(this.physicalCloseTimeout, this.listeners);
+				producerToClose.closeDelegate(this.physicalCloseTimeout);
 			}
 			catch (Exception e) {
 				LOGGER.error(e, "Exception while closing producer");
@@ -700,7 +710,7 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 			CloseSafeProducer<K, V> next = queue.poll();
 			while (next != null) {
 				try {
-					next.closeDelegate(this.physicalCloseTimeout, this.listeners);
+					next.closeDelegate(this.physicalCloseTimeout);
 				}
 				catch (Exception e) {
 					LOGGER.error(e, "Exception while closing producer");
@@ -711,7 +721,7 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 		this.cache.clear();
 		this.threadBoundProducers.values().forEach(prod -> {
 			try {
-				prod.closeDelegate(this.physicalCloseTimeout, this.listeners);
+				prod.closeDelegate(this.physicalCloseTimeout);
 			}
 			catch (Exception e) {
 				LOGGER.error(e, "Exception while closing producer");
@@ -769,7 +779,7 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 		this.globalLock.lock();
 		try {
 			if (this.producer != null && this.producer.closed) {
-				this.producer.closeDelegate(this.physicalCloseTimeout, this.listeners);
+				this.producer.closeDelegate(this.physicalCloseTimeout);
 				this.producer = null;
 			}
 			if (this.producer != null && expire(this.producer)) {
@@ -820,6 +830,9 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 	 * @since 2.2.13
 	 */
 	protected final boolean removeProducer(CloseSafeProducer<K, V> producerToRemove, Duration timeout) {
+		if (producerToRemove.closed) {
+			this.listeners.forEach(listener -> listener.producerRemoved(producerToRemove.clientId, producerToRemove));
+		}
 		return producerToRemove.closed;
 	}
 
@@ -846,7 +859,8 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 			}
 		}
 		if (cachedProducer == null) {
-			return doCreateTxProducer(txIdPrefix, "" + this.transactionIdSuffix.getAndIncrement(), this::cacheReturner);
+			String suffix = this.transactionIdSuffixStrategy.acquireSuffix(txIdPrefix);
+			return doCreateTxProducer(txIdPrefix, suffix, this::cacheReturner);
 		}
 		else {
 			return cachedProducer;
@@ -856,24 +870,28 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 	private boolean expire(CloseSafeProducer<K, V> producer) {
 		boolean expired = this.maxAge > 0 && System.currentTimeMillis() - producer.created > this.maxAge;
 		if (expired) {
-			producer.closeDelegate(this.physicalCloseTimeout, this.listeners);
+			producer.closeDelegate(this.physicalCloseTimeout);
 		}
 		return expired;
 	}
 
 	boolean cacheReturner(CloseSafeProducer<K, V> producerToRemove, Duration timeout) {
 		if (producerToRemove.closed) {
-			producerToRemove.closeDelegate(timeout, this.listeners);
+			this.removeTransactionProducer(producerToRemove, timeout, this.listeners);
 			return true;
 		}
 		else {
 			this.globalLock.lock();
 			try {
+				if (producerToRemove.epoch != this.epoch.get()) {
+					this.removeTransactionProducer(producerToRemove, timeout, this.listeners);
+					return true;
+				}
 				BlockingQueue<CloseSafeProducer<K, V>> txIdCache = getCache(producerToRemove.txIdPrefix);
 				if (producerToRemove.epoch != this.epoch.get()
 						|| (txIdCache != null && !txIdCache.contains(producerToRemove)
 						&& !txIdCache.offer(producerToRemove))) {
-					producerToRemove.closeDelegate(timeout, this.listeners);
+					this.removeTransactionProducer(producerToRemove, timeout, this.listeners);
 					return true;
 				}
 			}
@@ -882,6 +900,12 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 			}
 			return false;
 		}
+	}
+
+	private void removeTransactionProducer(CloseSafeProducer<K, V> producer, Duration timeout,
+			List<Listener<K, V>> listeners) {
+		this.transactionIdSuffixStrategy.releaseSuffix(producer.txIdPrefix, producer.txIdSuffix);
+		listeners.forEach(listener -> listener.producerRemoved(producer.clientId, producer));
 	}
 
 	private CloseSafeProducer<K, V> doCreateTxProducer(String prefix, String suffix,
@@ -899,10 +923,13 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 				newEx.addSuppressed(ex2);
 				throw newEx; // NOSONAR - lost stack trace
 			}
+			finally {
+				this.transactionIdSuffixStrategy.releaseSuffix(prefix, suffix);
+			}
 			throw new KafkaException("initTransactions() failed", ex);
 		}
 		CloseSafeProducer<K, V> closeSafeProducer =
-				new CloseSafeProducer<>(newProducer, remover, prefix, this.physicalCloseTimeout, this.beanName,
+				new CloseSafeProducer<>(newProducer, remover, prefix, suffix, this.physicalCloseTimeout, this.beanName,
 						this.epoch.get());
 		this.listeners.forEach(listener -> listener.producerAdded(closeSafeProducer.clientId, closeSafeProducer));
 		return closeSafeProducer;
@@ -923,7 +950,7 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 	}
 
 	@Nullable
-	protected BlockingQueue<CloseSafeProducer<K, V>> getCache(String txIdPrefix) {
+	protected BlockingQueue<CloseSafeProducer<K, V>> getCache(@Nullable String txIdPrefix) {
 		if (txIdPrefix == null) {
 			return null;
 		}
@@ -941,7 +968,7 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 	public void closeThreadBoundProducer() {
 		CloseSafeProducer<K, V> tlProducer = this.threadBoundProducers.remove(Thread.currentThread());
 		if (tlProducer != null) {
-			tlProducer.closeDelegate(this.physicalCloseTimeout, this.listeners);
+			tlProducer.closeDelegate(this.physicalCloseTimeout);
 		}
 	}
 
@@ -991,6 +1018,8 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 
 		final String txIdPrefix; // NOSONAR
 
+		final String txIdSuffix;  // NOSONAR
+
 		final long created; // NOSONAR
 
 		private final Duration closeTimeout;
@@ -1010,14 +1039,21 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 			this(delegate, removeConsumerProducer, null, closeTimeout, factoryName, epoch);
 		}
 
+		CloseSafeProducer(Producer<K, V> delegate, BiPredicate<CloseSafeProducer<K, V>, Duration> removeProducer,
+				@Nullable String txIdPrefix, Duration closeTimeout, String factoryName, int epoch) {
+
+			this(delegate, removeProducer, txIdPrefix, null, closeTimeout, factoryName, epoch);
+		}
+
 		CloseSafeProducer(Producer<K, V> delegate,
 				BiPredicate<CloseSafeProducer<K, V>, Duration> removeProducer, @Nullable String txIdPrefix,
-				Duration closeTimeout, String factoryName, int epoch) {
+				@Nullable String txIdSuffix, Duration closeTimeout, String factoryName, int epoch) {
 
 			Assert.isTrue(!(delegate instanceof CloseSafeProducer), "Cannot double-wrap a producer");
 			this.delegate = delegate;
 			this.removeProducer = removeProducer;
 			this.txIdPrefix = txIdPrefix;
+			this.txIdSuffix = txIdSuffix;
 			this.closeTimeout = closeTimeout;
 			Map<MetricName, ? extends Metric> metrics = delegate.metrics();
 			Iterator<MetricName> metricIterator = metrics.keySet().iterator();
@@ -1057,7 +1093,6 @@ public class DefaultKafkaProducerFactory<K, V> extends KafkaResourceFactory
 					}
 					callback.onCompletion(metadata, exception);
 				}
-
 			});
 		}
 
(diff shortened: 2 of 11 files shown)


sobychacko:
@stillya @Wzy19930507 It looks like we can use this PR instead of the previous one since this contains all the commits from the previous one (except for the docs commit). If that is the case, can we close the previous PR?

sobychacko:
Checkstyle failures in the PR build: https://github.com/spring-projects/spring-kafka/actions/runs/7160082920

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -926,7 +926,6 @@ else if (listener instanceof MessageListener) {
 			if (this.containerProperties.isLogContainerConfig()) {
 				this.logger.info(toString());
 			}
-			Map<String, Object> props = KafkaMessageListenerContainer.this.consumerFactory.getConfigurationProperties();
```
sobychacko, line 929: Why is this change?

sobychacko (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionSuffixManager.java
```diff
@@ -0,0 +1,110 @@
...
+	 * Retrieve the suffix for the transactional producer from the cache or generate a new one 
+	 * if caching is disabled.
+	 * @param txIdPrefix the transaction id prefix.
+	 * @return the suffix.
+	 * @throws NoProducerAvailableException if caching is enabled and no suffixes are available.
+	 */
+	@Override
+	public String retrieveSuffix(String txIdPrefix) {
+		BlockingQueue<String> cache = getSuffixCache(txIdPrefix);
+		if (cache == null) {
```
sobychacko, line 60: This may never be null since the value is computed in the `getSuffixCache` method.

sobychacko:
@stillya The gist of your change from a first-pass review I did is the following. Please confirm if this understanding is correct. The `DefaultKafkaProducerFactory` is now aware of a `TransactionSuffixManager.` with an implementation of `DefaultTransactionSuffixManager.` The contract for the `TransactionSuffixManager` allows three things - retrieve, return, and reset transaction suffixes.  When the transactional producer is created for the first time, the suffix manager primes the suffix cache with the `maxCache - 1` suffixes, which the producer factory retrieves to create the transactional producer. The suffix retrieval will continue if more producers are created via the factory. Once the suffix cache is full, If a request comes in for another producer creation, then the no producer exception is thrown. The suffix is returned to the cache when the producers are closed, or transaction exceptions occur, etc. When the producer factory is destroyed, then the suffix cache is reset. I think this is the main use case that we are addressing via this PR. Could you please confirm?

Some quick comments. The `TransactionSuffixManager` might sound a bit heavy in this context, as we expose this as a public API. I suggest using a lighter API name for both the interface and the implementation. 

When `maxCache` is zero, where is the default code that increments the suffix re-located?

Thanks!

Wzy19930507:
> It looks like we can use this PR instead of the previous one since this contains all the commits from the previous one (except for the docs commit). If that is the case, can we close the previous PR?

@sobychacko Sure, let’s continue in this PR, @stillya I’m sorry I misunderstood your meaning.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
stillya: I obtained it from previous PR, but this field isn't used because `ErrorHandlingUtils.checkDeserializer` itself goes to `ConsumerFactory` to grab configuration properties. This was done in this [commit](https://github.com/spring-projects/spring-kafka/commit/e9564134733e81ba530d83193b68c0c34cb7a556). Perhaps it would be better to extract these changes into another refactoring PR?

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionSuffixManager.java
stillya: It could be null if `txIdPrefix` is not passed or caching is not enabled(maxCache <= 0).

stillya:
> @stillya The gist of your change from a first-pass review I did is the following.
> 
> The `DefaultKafkaProducerFactory` is now aware of a `TransactionSuffixManager.` with an implementation of `DefaultTransactionSuffixManager.` The contract for the `TransactionSuffixManager` allows three things - retrieve, return, and reset transaction suffixes. When the transactional producer is created for the first time, the suffix manager primes the suffix cache with the `maxCache - 1` suffixes, which the producer factory retrieves to create the transactional producer. The suffix retrieval will continue if more producers are created via the factory. Once the suffix cache is full, If a request comes in for creating another transactional producer, then the `NoProducerAvailableException` is thrown. The suffix is returned to the cache when the producers are closed, or transaction exceptions occur, etc. When the producer factory is destroyed, then the suffix cache is reset. I think this is the main use case that we are addressing via this PR. Could you please confirm?

Yes, you're absolutely right.

> Some quick comments. The TransactionSuffixManager might sound a bit heavy in this context, as we expose this as a public API. I suggest using a lighter API name for both the interface and the implementation.

I agree that `TransactionSuffixManager` might sound a bit heavy for this context. Do you have any suggestions for a lighter, more intuitive name for both the interface and its implementation? Maybe `TransactionSuffixHandler` as an alternative?

> When maxCache is zero, where is the default code that increments the suffix re-located?

When maxCache is zero, so caching is disabled, then we go [here](https://github.com/spring-projects/spring-kafka/pull/2930/files#diff-9b2f6da3e4263fb745edb8cd99402d6f4dd559b812bdb00d8a6ef9b28ca9e5afR61) and just increment the counter as before.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionSuffixManager.java
```diff
@@ -0,0 +1,110 @@
...
+
+		String suffix = cache.poll();
+		if (suffix == null) {
+			throw new NoProducerAvailableException("No available transaction producer suffix for " + txIdPrefix);
+		}
+		return suffix;
+	}
+
+	@Override
+	public void returnSuffix(@Nullable String txIdPrefix, @Nullable String suffix) {
```
Wzy19930507, line 72: We need epoch when return suffix, prev epoch producer  maybe close after reset at next epoch.

Wzy19930507 (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
sobychacko: I suggest not including changes unrelated to the core features we are adding through the PR. Otherwise, it will be confusing if we have to re-review this code later.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionSuffixManager.java
```diff
@@ -0,0 +1,110 @@
...
+	 * if caching is disabled.
+	 * @param txIdPrefix the transaction id prefix.
+	 * @return the suffix.
+	 * @throws NoProducerAvailableException if caching is enabled and no suffixes are available.
+	 */
+	@Override
+	public String retrieveSuffix(String txIdPrefix) {
+		BlockingQueue<String> cache = getSuffixCache(txIdPrefix);
+		if (cache == null) {
+			return String.valueOf(this.transactionIdSuffix.getAndIncrement());
```
Wzy19930507, line 61: When txIdPrefix is null maybe not return suffix is better.

sobychacko:
> I agree that `TransactionSuffixManager` might sound a bit heavy for this context. Do you have any suggestions for a lighter, more intuitive name for both the interface and its implementation? Maybe `TransactionSuffixHandler` as an alternative?

`Handler` sounds very reasonable. How about `TransactionIdSuffixHandler` and `DefaultTransactionIdSuffixHandler`?

spring-kafka/src/main/java/org/springframework/kafka/core/TransactionSuffixManager.java
```diff
@@ -0,0 +1,48 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.core;
+
+/**
+ * The strategy for managing transactional producer suffixes.
```
sobychacko, line 20: If it's a real strategy, we must call it such. But are we providing a way to customize a strategy? Are we envisioning a situation where users provide a custom implementation in which they implement a different strategy? (i.e., non-incrementing integers for suffix, but rather a random string, for e.g.)

spring-kafka/src/main/java/org/springframework/kafka/core/TransactionSuffixManager.java
```diff
@@ -0,0 +1,48 @@
...
+	 */
+	String retrieveSuffix(String txIdPrefix);
+
+	/**
+	 * Return the suffix for the transactional producer back for reuse.
+	 *
+	 * @param txIdPrefix the transaction id prefix.
+	 * @param suffix the suffix.
+	 */
+	void returnSuffix(String txIdPrefix, String suffix);
```
sobychacko, line 42: `retrieve` and `return` in the API names might sound confusing. Can we change `returnSuffix` to something that is more explicit to reason about? maybe `releaseSuffixBack`? or something along those lines.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionSuffixManager.java
```diff
@@ -0,0 +1,110 @@
...
+	 * @param maxCache the maxCache to set
+	 */
+	public void setMaxCache(int maxCache) {
+		Assert.isTrue(maxCache >= 0, "'maxCache' must greater than or equal 0");
+		this.maxCache = maxCache;
+	}
+
+	@Nullable
+	private BlockingQueue<String> getSuffixCache(@Nullable String txIdPrefix) {
+		if (txIdPrefix == null || this.maxCache <= 0) {
```
Wzy19930507, line 98: When txIdPrefix is null maybe not return suffix is better.
At DefaultKafkaProducerFactory line 848 in this PR, When txIdPrefix is null will threw exception.

spring-kafka/src/main/java/org/springframework/kafka/core/TransactionSuffixManager.java
stillya: Yes, it's a real strategy. The main reason why I created this issue is to provide the ability to manage `txIds` using external storage, for example, because in my environment, it's a better strategy than everything else. So maybe it's better to call it `TransactionSuffixStrategy`/`DefaultTransactionSuffixStrategy`?

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionSuffixManager.java
stillya: I don't quite understand how `epoch` and `txId` are related in this case. If `epoch` is bumped, then the previous producer will be closed, and the suffix will be returned back.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionSuffixManager.java
stillya: Sounds reasonable; I will add an assert.

spring-kafka/src/main/java/org/springframework/kafka/core/TransactionSuffixManager.java
Wzy19930507: Strategy needs a strategy method. Here has three method `setMaxCache`,  `getSuffix`, `releaseSuffix`, not like strategy

spring-kafka/src/main/java/org/springframework/kafka/core/TransactionSuffixManager.java
sobychacko: I like those.

spring-kafka/src/main/java/org/springframework/kafka/core/TransactionSuffixManager.java
sobychacko: > Yes, it's a real strategy. The main reason why I created this issue is to provide the ability to manage `txIds` using external storage, for example, because in my environment, it's a better strategy than everything else. So maybe it's better to call it `TransactionSuffixStrategy`/`DefaultTransactionSuffixStrategy`?

`TransactionIdSuffixStrategy`/`DefaultTransactionIdSuffixStrategy`?

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionSuffixManager.java
Wzy19930507: such as https://github.com/spring-projects/spring-kafka/blob/2306ac5cbd47702f9a492541186944682edada9c/spring-kafka/src/test/java/org/springframework/kafka/core/DefaultKafkaProducerFactoryTests.java#L257-L288

stillya: I still don't get it. When `ProducerFactory#reset()` is called, **every** producer is closed, so calling close on the producer itself via `Producer#close()` will do nothing. Am I mistaken?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
stillya: Fixed.

spring-kafka/src/main/java/org/springframework/kafka/core/TransactionSuffixManager.java
stillya: Fixed.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionSuffixManager.java
stillya: Fixed.

sobychacko:
@stillya The PR CI build keeps failing due to some test failure. Could you take a look?

stillya:
> @stillya, The PR CI build keeps failing due to some test failure. Could you take a look? https://github.com/spring-projects/spring-kafka/actions/runs/7186689052

Yeah, sorry, broke one test by reverting old code. Fixed, now everything should pass.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionSuffixManager.java
stillya: I got what you meant, so I we do not need to return `suffix` back if `epoch` if bumped, because all `suffixes` already returned.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultTransactionIdSuffixStrategyTests.java
```diff
@@ -0,0 +1,113 @@
...
+	@Test
+	void testSetMaxCacheIsNegative() {
+		assertThatExceptionOfType(IllegalArgumentException.class).isThrownBy(() -> suffixStrategy.setMaxCache(-1));
+	}
+
+	@Test
+	void testSetMaxCacheIsZero() {
+		assertThatNoException().isThrownBy(() -> suffixStrategy.setMaxCache(0));
+	}
+}
```
sobychacko, line 113: It would be nice to make the test names consistent with the API name changes - `acquire` and `release.`. Also, since this is a new test and we are in JUnit 5, there is no need to prefix the test methods with the `test` literal. Cleaning those up in this new test class is good.

stillya: Fixed.

sobychacko (approved):

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionSuffixManager.java
Wzy19930507: Have a look at https://github.com/spring-projects/spring-kafka/blob/2306ac5cbd47702f9a492541186944682edada9c/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java#L762-L788

At line L763-L765, tx producer is diff from nomal producer,
tx producer is not managed by `DefaultKafkaProducerFactory`, so if producer is in transaction when call `DefaultKafkaProducerFactory#reset()`, producer will not close.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -880,14 +880,10 @@ boolean cacheReturner(CloseSafeProducer<K, V> producerToRemove, Duration timeout
 		else {
 			this.globalLock.lock();
 			try {
-				if (producerToRemove.epoch != this.epoch.get()) {
-					producerToRemove.closeDelegate(timeout, this.listeners);
-					return true;
-				}
```
Wzy19930507: Have a look at https://github.com/spring-projects/spring-kafka/blob/2306ac5cbd47702f9a492541186944682edada9c/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java#L762-L788

At line L763-L765, create tx producer is diff from nomal producer,
tx producer is not managed by `DefaultKafkaProducerFactory`, so producer suffix not release when call `DefaultKafkaProducerFactory#reset()` if producer is in transaction. (eg. use tx producer in `containerListener`)

Maybe recover this code is better, i'm not sure, please help take a look, @sobychacko @artembilan

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -867,7 +867,7 @@ protected Producer<K, V> createTransactionalProducer(String txIdPrefix) {
 	private boolean expire(CloseSafeProducer<K, V> producer) {
 		boolean expired = this.maxAge > 0 && System.currentTimeMillis() - producer.created > this.maxAge;
 		if (expired) {
-			closeTransactionProducer(producer, this.physicalCloseTimeout, this.listeners);
+			producer.closeDelegate(this.physicalCloseTimeout, this.listeners);
```
Wzy19930507: At https://github.com/spring-projects/spring-kafka/blob/2306ac5cbd47702f9a492541186944682edada9c/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java#L846-L853

`expire()` is called by A, in this case producer need close and return suffix

spring-kafka/src/main/java/org/springframework/kafka/core/NoProducerAvailableException.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * An exception thrown by no transaction producer available exception, when set
+ * {@link DefaultKafkaProducerFactory} maxCache greater than 0.
+ *
+ * @author Wang Zhiyang
+ * @since 3.1.1
+ *
+ */
+public class NoProducerAvailableException extends RuntimeException {
```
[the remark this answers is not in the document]
Wzy19930507: My mistake in prev commit

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
stillya: Sure, but `expire()` will call `closeDelegate`, which will, in turn, call `remover`, and in the case of a transactional producer, it is the `cacheReturner`.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultKafkaProducerFactoryTests.java
```diff
@@ -666,38 +674,6 @@ protected Producer createRawProducer(Map configs) {
...
-		pf.setMaxCache(2);
-		assertThat(pf.initTransactionIdSuffix(10)).isTrue();
-		Producer aProducer = pf.createProducer();
-		assertThat(configPassedToKafkaConsumer.get(ProducerConfig.TRANSACTIONAL_ID_CONFIG)).isEqualTo("tx.10");
-		assertThat(pf.initTransactionIdSuffix(20)).isFalse();
-		Producer bProducer = pf.createProducer();
-		assertThat(configPassedToKafkaConsumer.get(ProducerConfig.TRANSACTIONAL_ID_CONFIG)).isEqualTo("tx.11");
-		bProducer.close(Duration.ofSeconds(20));
-		pf.setMaxAge(Duration.ofMillis(10));
-		Thread.sleep(50);
```
Wzy19930507: Hi, @stillya , This unit test can test tx expire() scene. if you don't mind, recover it

artembilan:
FYI, we have pushed this change to the next `3.2` generation as a new feature request.
Too drastic to make a change in the current patch release.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -350,6 +350,15 @@ public void setValueSerializerSupplier(Supplier<Serializer<V>> valueSerializerSu
 		this.valueSerializerSupplier = valueSerializerSupplier(valueSerializerSupplier);
 	}
 
+	/**
+	 * Set the transaction suffix strategy.
+	 * @param transactionIdSuffixStrategy the strategy.
+	 * @since 3.1.1
+	 */
+	public void setTransactionIdSuffixStrategy(TransactionIdSuffixStrategy transactionIdSuffixStrategy) {
+		this.transactionIdSuffixStrategy = transactionIdSuffixStrategy;
```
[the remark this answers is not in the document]
stillya: Fixed.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
```diff
@@ -102,11 +102,45 @@ NOTE: If there is a `KafkaTransactionManager` (or synchronized) transaction in p
...
 With `EOSMode.V2` (aka `BETA`), the only supported mode, it is no longer necessary to use the same `transactional.id`, even for consumer-initiated transactions; in fact, it must be unique on each instance the same as for producer-initiated transactions.
 This property must have a different value on each application instance.
 
+[[transaction-id-suffix-fixed]]
+== `TransactionIdSuffix Fixed`
+
+Since 3.1, introduced a new interface `TransactionIdSuffixStrategy` to manage `transactional.id` suffix.
+The default implementation is `DefaultTransactionIdSuffixStrategy` when setting `maxCache` greater than zero can reuse `transactional.id` within a specific range, otherwise suffixes will be generated on the fly by incrementing a counter.
+When a transaction producer is requested and `transactional.id` all in use, throw a `NoProducerAvailableException`.
+User can then use a RetryTemplate configured to retry that exception, with a suitably configured back off.
```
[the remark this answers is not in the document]
stillya: Fixed.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
```diff
@@ -102,11 +102,45 @@ NOTE: If there is a `KafkaTransactionManager` (or synchronized) transaction in p
...
+
+    @Bean
+    public ProducerFactory<String, String> myProducerFactory() {
+        Map<String, Object> configs = producerConfigs();
+        configs.put(ProducerConfig.CLIENT_ID_CONFIG, "myClientId");
+        ...
+        DefaultKafkaProducerFactory<String, String> pf = new DefaultKafkaProducerFactory<>(configs);
+        ...
+        DefaultTransactionIdSuffixStrategy ss = new DefaultTransactionIdSuffixStrategy();
+		ss.setMaxCache(5);
```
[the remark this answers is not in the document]
stillya: Fixed.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionIdSuffixStrategy.java
```diff
@@ -0,0 +1,113 @@
...
+	/**
+	 * Acquire the suffix for the transactional producer from the cache or generate a new one
+	 * if caching is disabled.
+	 * @param txIdPrefix the transaction id prefix.
+	 * @return the suffix.
+	 * @throws NoProducerAvailableException if caching is enabled and no suffixes are available.
+	 */
+	@Override
+	public String acquireSuffix(String txIdPrefix) {
+		Assert.notNull(txIdPrefix, () -> "'txIdPrefix' must not be null");
```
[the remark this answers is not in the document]
stillya: Fixed.

spring-kafka/src/main/java/org/springframework/kafka/core/NoProducerAvailableException.java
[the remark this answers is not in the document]
stillya: Fixed and added `txIdPrefix`.

spring-kafka/src/main/java/org/springframework/kafka/core/NoProducerAvailableException.java
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.core;
+
+import org.springframework.lang.Nullable;
+
+/**
+ * An exception thrown by no transaction producer available exception, when set
```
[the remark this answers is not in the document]
stillya: Fixed.

spring-kafka/src/main/java/org/springframework/kafka/core/TransactionIdSuffixStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @author Ilya Starchenko
+ *
+ * @since 3.1.1
+ */
+public interface TransactionIdSuffixStrategy {
+
+	/**
+	 * Acquire the suffix for the transactional producer.
+	 *
```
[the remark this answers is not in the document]
stillya: Fixed.

spring-kafka/src/main/java/org/springframework/kafka/core/TransactionIdSuffixStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+	 * @param txIdPrefix the transaction id prefix.
+	 * @param suffix the suffix.
+	 */
+	void releaseSuffix(String txIdPrefix, String suffix);
+
+	/**
+	 * Clear all suffixes for the transactional producer.
+	 */
+	void reset();
```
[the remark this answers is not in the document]
stillya: I have refactored interface so that there is no actual need to reset `txId`. What we aim for is to acquire and release it. If someone manually closes the factory without closing the cached producer, it becomes the user's responsibility to close the producer and release the `suffix`.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -1159,22 +1194,30 @@ public void close(@Nullable Duration timeout) {
...
 					this.closed = this.removeProducer.test(this, timeout);
+					if (this.closed) {
+						this.delegate.close(timeout == null ? this.closeTimeout : timeout);
+					}
 				}
 			}
 		}
 
-		void closeDelegate(Duration timeout, List<Listener<K, V>> listeners) {
+		void closeDelegate(Duration timeout) {
```
stillya, line 1210: I modified the behavior of `CloseSafeProducer` to enhance the consistency of the close operation. Now, both `close()` and `closeDelegate()` methods invoke `removeProducer()`, ensuring that external logic (e.g., cache returning or listeners) behaves consistently for both operations. The key distinction is that `closeDelegate()` explicitly forces the closure of the producer. I believe this change makes the code more consistent and clearer. @artembilan, WDYT?

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
stillya: Rework this behaviour. Look at [here](https://github.com/spring-projects/spring-kafka/pull/2930#discussion_r1428924996).

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultKafkaProducerFactoryTests.java
```diff
@@ -320,12 +454,12 @@ protected Producer createKafkaProducer() {
 		pf.closeThreadBoundProducer();
 		assertThat(KafkaTestUtils.getPropertyValue(pf, "threadBoundProducers", Map.class).get(Thread.currentThread()))
 				.isNull();
-		verify(producer, times(3)).close(any(Duration.class));
+		verify(producer, times(2)).close(any(Duration.class));
```
stillya, line 457: I changed the `assert` because in the refactored `close`/`closeDelegate` methods, there is no need to call close one more time on an already closed producer. I am unsure why it was done previously. If I'm missing something, please let me know.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionIdSuffixStrategy.java
```diff
@@ -0,0 +1,109 @@
...
+			queue.add(suffix);
+		}
+	}
+
+	/**
+	 * Set the maximum size for transaction producer cache.
+	 * @param maxCache the maxCache to set
+	 */
+	public void setMaxCache(int maxCache) {
+		Assert.isTrue(maxCache >= 0, "'maxCache' must greater than or equal 0");
```
NathanQingyangXu, line 90: "'maxCache' must be greater than or equal to 0"?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
```diff
@@ -102,11 +102,45 @@ NOTE: If there is a `KafkaTransactionManager` (or synchronized) transaction in p
...
-== `transactionIdPrefix`
+== `TransactionIdPrefix`
 
 With `EOSMode.V2` (aka `BETA`), the only supported mode, it is no longer necessary to use the same `transactional.id`, even for consumer-initiated transactions; in fact, it must be unique on each instance the same as for producer-initiated transactions.
 This property must have a different value on each application instance.
 
+[[transaction-id-suffix-fixed]]
+== `TransactionIdSuffix Fixed`
+
+Since 3.2, introduced a new interface `TransactionIdSuffixStrategy` to manage `transactional.id` suffix.
```
NathanQingyangXu, line 113: seems the verbiage is unnatural:
```
Since 3.2, a new `TransactionIdSuffixStrategy` interface was introduced to manage `transactional.id` suffix.
```
may be better

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
```diff
@@ -102,11 +102,45 @@ NOTE: If there is a `KafkaTransactionManager` (or synchronized) transaction in p
...
+        return pf;
+    }
+
+}
+----
+When setting `maxCache` to 5, `transactional.id` is `my.txid.`++`{0-4}`+.
+
+IMPORTANT: When use `KafkaTransactionManager` in the `ConcurrentMessageListenerContainer` and enable `maxCache`, `maxCache` must be greater than or equal to `concurrency`.
+If some `MessageListenerContainer` cannot get the transaction, will throw `NoProducerAvailableException`.
+When use nested transactions in `ConcurrentMessageListenerContainer`, `maxCache` needs to increase the number of nested transactions.
```
NathanQingyangXu, line 141: the verbigage above has room of improvement. E.g.:
```
If some `MessageListenerContainer` cannot get the transaction, will throw `NoProducerAvailableException`.
```
Just one example.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionIdSuffixStrategy.java
NathanQingyangXu: should we assert `maxCache > 0` to align with the adoc?

stillya: I don't see any contradictions. The documentation states that if maxCache > 0, but the constraint specifies that it should not be negative. In this case, we could set maxCache to zero to disable caching.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
stillya: Tried to improve. WDYT?

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionIdSuffixStrategy.java
NathanQingyangXu: My first comment aims to point out the grammatical defect, my second one is about logic.

stillya: > My first comment aims to point out the grammatical defect
Fixed.

artembilan:
@stillya ,

Thanks

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
```diff
@@ -820,6 +830,9 @@ protected Producer<K, V> createKafkaProducer() {
 	 * @since 2.2.13
 	 */
 	protected final boolean removeProducer(CloseSafeProducer<K, V> producerToRemove, Duration timeout) {
+		if (producerToRemove.closed) {
```
artembilan, line 833: This method is `protected`, so, can be overridden in the target project.
Therefore, `this.listeners.forEach` might not be closed.

I'm not sure why you have changed that `closeDelegate()` implementation, but probably it is not what we really need for the functionality you trying to implement.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionIdSuffixStrategy.java
```diff
@@ -0,0 +1,109 @@
...
+		if (queue != null && !queue.contains(suffix)) {
+			queue.add(suffix);
+		}
+	}
+
+	/**
+	 * Set the maximum size for transaction producer cache.
+	 * @param maxCache the maxCache to set
+	 */
+	public void setMaxCache(int maxCache) {
```
artembilan, line 89: It looks like this cam go to the ctor arg.
Even if it is optional, it is better to make end-user explicit with setting it to `0`

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionIdSuffixStrategy.java
```diff
@@ -0,0 +1,109 @@
...
+	public String acquireSuffix(String txIdPrefix) {
+		Assert.notNull(txIdPrefix, "'txIdPrefix' must not be null");
+		BlockingQueue<String> cache = getSuffixCache(txIdPrefix);
+		if (cache == null) {
+			return String.valueOf(this.transactionIdSuffix.getAndIncrement());
+		}
+
+		String suffix = cache.poll();
+		if (suffix == null) {
+			throw new NoProducerAvailableException("No available transaction producer", txIdPrefix);
```
artembilan, line 67:
```suggestion
			throw new NoProducerAvailableException("No available transactional producer", txIdPrefix);
```

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionIdSuffixStrategy.java
stillya: I tried to follow the current style, as demonstrated [here](https://github.com/spring-projects/spring-kafka/blob/8d8f2b364d35fdd4f88289289fbf79a95207e5e1/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java#L156), for example.

artembilan: Well, you don't compare apples with apples.
That `maxAge` is indeed an optional on the `DefaultKafkaProducerFactory`, where `configs` are required. 
Therefore they go to the ctor.
Something similar is here: the main contract of this strategy is a cache size. Therefore it has to be as a ctor arg for expected enforcement.

I'm just trying to be reasonable instead of following other code blindly.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaProducerFactory.java
stillya: > This method is protected, so, can be overridden in the target project.
Therefore, this.listeners.forEach might not be closed.

I'm sorry, I didn't quite understand your point. What do you mean by "this.listeners.forEach might not be closed"?

> I'm not sure why you have changed that closeDelegate() implementation, but probably it is not what we really need for the functionality you trying to implement.

[This](https://github.com/spring-projects/spring-kafka/pull/2930#discussion_r1428924996) is the main reason why I change `closeDelegate()` implementation. The goal is to ensure that the behavior remains unchanged while providing a more user-friendly approach to closing the producer. Otherwise, I would need to duplicate the logic for the close operation to implement my feature.

artembilan: Sorry. I mean `not called`. 
Well, I see now that this `removeProducer` is `final`. So, it cannot be overridden.
I still don't understand what drove you to do that, but I probably will believe you as is.

artembilan (changes requested):
Please, consider to update copyright of all the affected classes to the current year.

spring-kafka/src/main/java/org/springframework/kafka/core/DefaultTransactionIdSuffixStrategy.java
stillya: Sure, I agree. It looks way better; I just wanted to note why I did it.
Thanks. Is it okay now?

artembilan (approved):
I think, I'm OK with this.

@sobychacko , you call.

Thanks

sobychacko:
Thanks, Artem! I will take a look and merge it today.

sobychacko:
@stillya @Wzy19930507 Thank you for this significant PR contribution. It is now merged upstream.

PR #2945, opened 2023-12-16
fix a default method defect in ConsumerFactory

NathanQingyangXu:
in ConsumerFactory interface, the following two methods seems counter-intuitive:
```
Consumer<K, V> createConsumer(@Nullable String groupId, @Nullable String clientIdPrefix,
			@Nullable String clientIdSuffix);
default Consumer<K, V> createConsumer(@Nullable String groupId, @Nullable String clientIdPrefix,
			@Nullable String clientIdSuffix, @Nullable Properties properties) {

		return createConsumer(groupId, clientIdPrefix, clientIdSuffix);
	}
```
seems we should switch the `default` relationship the other way.

Also fixed some grammatical and logging defects in other classes

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/ConsumerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/ConsumerFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 the original author or authors.
+ * Copyright 2016-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/ConsumerFactory.java
```diff
@@ -27,7 +27,7 @@
 import org.springframework.lang.Nullable;
 
 /**
- * The strategy to produce a {@link Consumer} instance(s).
+ * The strategy to produce a {@link Consumer} instance.
```
NathanQingyangXu, line 30: seems we never create multiple instances

sobychacko (approved):

PR #2961, opened 2023-12-18
Improve MockProducerFactory usability by wrapping  MockProducer with …

pszymczyk:
…CloseSafeMockProducer

Hi

In the current project, I've opted to use the `MockProducer` class and Spring support in the form of `MockProducerFactory`. Unfortunately, it turns out that after the first message is sent,  `KafkaTemplate` by default closes the provided `MockProducer` instance, making it impossible to send another message. I suggest a solution in which the factory wraps the MockProducer in a class that ignores the producer's closing.

Best regards,
Pawel

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/mock/MockProducerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/mock/MockProducerFactory.java
@@ -29,6 +29,7 @@ import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.ProducerFencedException;
+
 import org.springframework.core.log.LogAccessor;
 import org.springframework.kafka.core.DefaultKafkaProducerFactory;
 import org.springframework.kafka.core.ProducerFactory;
--- a/spring-kafka/src/main/java/org/springframework/kafka/mock/MockProducerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/mock/MockProducerFactory.java
@@ -29,7 +29,6 @@ import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.ProducerFencedException;
-
 import org.springframework.core.log.LogAccessor;
 import org.springframework.kafka.core.DefaultKafkaProducerFactory;
 import org.springframework.kafka.core.ProducerFactory;
@@ -80,8 +79,7 @@ public class MockProducerFactory<K, V> implements ProducerFactory<K, V> {
 	 * @param producerProvider the provider function.
 	 * @param defaultTxId the default transactional id.
 	 */
-	public MockProducerFactory(BiFunction<Boolean, String, MockProducer<K, V>> producerProvider,
-							   @Nullable String defaultTxId) {
+	public MockProducerFactory(BiFunction<Boolean, String, MockProducer<K, V>> producerProvider, @Nullable String defaultTxId) {
 
 		this.producerProvider = producerProvider;
 		this.defaultTxId = defaultTxId;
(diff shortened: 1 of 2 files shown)


sobychacko:
@pszymczyk Checkstyle failures in the PR build. Could you fix them? https://github.com/spring-projects/spring-kafka/actions/runs/7251864605/job/19755109702?pr=2961

sobychacko:
Build is still failing.

pszymczyk:
I've executed `./gradlew idea' and then execute reformat from Intellij Idea and it's still failing, to be honest I don't know how to fix it.

sobychacko:
Can you see the output on this build here? https://github.com/spring-projects/spring-kafka/actions/runs/7252844586/job/19758171575

pszymczyk:
I see:
```
 Error: [Task :spring-kafka:checkstyleMain] [ImportOrder] Wrong order for 'java.time.Duration' import.
Error: [Task :spring-kafka:checkstyleMain] [RegexpSinglelineJava] Line has leading space characters; indentation should be performed with tabs only.
```

pszymczyk:
Sorry but I have no idea what is wrong with following imports order:
```
import java.time.Duration;
import java.util.List;
import java.util.Map;
import java.util.concurrent.Future;
import java.util.function.BiFunction;
import java.util.function.Supplier;
```

spring-kafka/src/main/java/org/springframework/kafka/mock/MockProducerFactory.java
```diff
@@ -81,14 +99,96 @@ public Producer<K, V> createProducer() {
...
+		}
+
+		@Override
+		public void close() {
+			close(null);
+		}
+
+		@Override
+		public void close(Duration timeout) {
+			LOGGER.debug(() -> "The closing of delegate producer " + this.delegate + "has been skipped.");
```
sobychacko, line 191: @pszymczyk While it works for your use case, I wonder if there are other consequences by not closing the underlying producer. Can we only activate this feature (i.e., not closing the producer) based on a property or something along those lines?

pszymczyk: If you like to provide some configuration I would consider to add something in `KafkaTemplate` [send finally block](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java#L864). 

Currently, the `KafkaTemplate` provides clients the possibility to pass a custom implementation of the `ProducerFactory`. However, if you pass something other than the `DefaultKafkaProducerFactory` and your implementation does not implement something similar to the `CloseSafeProducer` wrapper, then your `KafkaProducer` will be closed just after the first message is sent, which basically makes no sense.

artembilan: Yes. I find that as a reasonable request.
We already have that `DefaultKafkaProducerFactory` API:
```
	/**
	 * Close the {@link Producer}(s) and clear the cache of transactional
	 * {@link Producer}(s).
	 * @since 2.2
	 */
	@Override
	public void reset() {
```
which lets you control such a lifecycle from the outside of the `KafkaTemplate`.
So, it is going to be safe to not close that cached producer in the `KafkaTemplate` after each send.

This way I suggest to back off the change to the `MockProducerFactory` and raise a new issue which we will consider for the next version.

Thanks

P.S. Removing that my other comment since it contradicts with what we have agreed here.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/mock/MockProducerFactory.java
pszymczyk: @artembilan sounds really good! I agree that `CloseSafeMockProducer` it's more like a workaround for the real problem I see in `KafkaTemplate` and `DefaultKafkaProducerFactory` indirect relationship, it's great that we all agree on that, I'm closing this PR and I will submit an issue with proper description.

PR #2963, opened 2023-12-20
GH-1189: Asynchronous server-side processing in a request/reply scenario

### LINKED ISSUE #1189: Asynchronous server-side processing in a request/reply scenario
Request/reply semantics requires that the server side listener method (annotated with `@SendTo`) return a response synchronously. There are use cases where server side request processing is asynchronous such that the listener method can't return a result immediately and must defer the response, releasing the consumer thread to handle other incoming requests. Once the result is available, an application thread would need to send it back to the client. This would be a feature similar to Servlet 3.0 Asynchronous Processing.

Wzy19930507:
Resolves #1189 
* support `Mono` and `Future`
* support kotlin suspend
* support auto-delect async reply
* `@SendTo` for `@KafkaHandler` after error is handled

---

Refactor `MessagingMessageListenerAdapter`

* move `BatchMessagingMessageListenerAdapter#invoke` and `RecordMessagingMessageListenerAdapter#invoke` to `MessagingMessageListenerAdapter`
* move `KafkaListenerErrorHandler` to `MessagingMessageListenerAdapter`
* add `@Nullable` to `KafkaListenerErrorHandler#handleError`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -60,6 +60,7 @@ ext {
 	junit4Version = '4.13.2'
 	junitJupiterVersion = '5.10.1'
 	kafkaVersion = '3.6.1'
+	kotlinCoroutinesVersion = '1.7.3'
 	log4jVersion = '2.21.1'
 	micrometerDocsVersion = '1.0.2'
 	micrometerVersion = '1.12.1'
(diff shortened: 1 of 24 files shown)


artembilan:
@Wzy19930507 ,

Thanks

Wzy19930507:
@artembilan @sobychacko Sorry, I accidentally shut down this PR. other one has been reopened.

Issue #2965, opened 2023-12-20
Closing MockProducer options in reference docs


sobychacko:
Include information on how to override the `MockProducer` and use it in `MockProducerFactory` in the testing section of the reference docs. 

See this for more details: https://github.com/spring-projects/spring-kafka/pull/2961#issuecomment-1864670561

PR #2966, opened 2023-12-20
GH-2965: Docs improvements in testing ref docs

sobychacko:
Fixes: #2965

Clarify how to override MockProducer and provide it via MockProducerFactory.

**Cherry-pick to `3.0.x`**

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
@@ -680,16 +680,21 @@ This is convenient for testing, when verifying multiple publishing on the same p
 
 Here is an example:
 
-[source, java]
+[source,java]
 ----
 @Bean
-ProducerFactory<String, String> mockProducerFactory() {
-    return new MockProducerFactory<>(() -> new MockProducer<>(false, new StringSerializer(), new StringSerializer()) {
+MockProducer<String, String> mockProducer() {
+    return new MockProducer<>(false, new StringSerializer(), new StringSerializer()) {
         @Override
         public void close() {
 
         }
-    });
+    };
+}
+
+@Bean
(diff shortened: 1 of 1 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
```diff
@@ -674,3 +674,22 @@ The transactional id is provided in case you wish to use a different `MockProduc
...
+When using the `MockProducer`, if you do not want to close the producer after each send, then you can provide a custom `MockProducer` implementation that overrides the `close` method that does not call the `close` method from the super class.
+This is convenient for testing, when verifying multiple publishing on the same producer without closing it.
+
+Here is an example:
+
+[source, java]
+----
+@Bean
+ProducerFactory<String, String> mockProducerFactory() {
+    return new MockProducerFactory<>(() -> new MockProducer<>(false, new StringSerializer(), new StringSerializer()) {
```
artembilan, line 687: This sample is a bit misleading.
When we really use a `MockProducer`, we perform some verification afterwards.
Therefore an instance for this `MockProducer` must be shared and accessible from the unit tests.
I probably won't mind to have it as a bean in the test configuration and then just refer to it from this supplier for `MockProducerFactory`.
Otherwise a configuration like this would always lead to a new `MockProducer` instance every time we call `KafkaTemplate.send()`.
Just because the `createProducer()` always call this `Supplier.get()`.

artembilan (changes requested):
I also don't see a reason to back-port it.
More over it is not going to be so easy because we have migrated to Antora exactly in this version and previous `3.0.x` is still plain AsciiDoc.

Thanks

sobychacko:
> I also don't see a reason to back-port it. More over it is not going to be so easy because we have migrated to Antora exactly in this version and previous `3.0.x` is still plain AsciiDoc.
> 
> Thanks

Can you remove the cherry-pick from the commit on merge?

PR #2967, opened 2023-12-22
GH-2967: Some minor improvements in RetryableTopicAnnotationProcessor

NathanQingyangXu:
some minor and easily verified code quality improvements.
after the changes, the warnings in IDEA decreased from 16 to 5.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
@@ -127,9 +127,7 @@ public class RetryableTopicAnnotationProcessor {
 		if (StringUtils.hasText(annotation.autoStartDltHandler())) {
 			autoStartDlt = resolveExpressionAsBoolean(annotation.autoStartDltHandler(), "autoStartDltContainer");
 		}
-		return RetryTopicConfigurationBuilder.newInstance()
-				.maxAttempts(resolveExpressionAsInteger(annotation.attempts(), "attempts", true))
-				.concurrency(resolveExpressionAsInteger(annotation.concurrency(), "concurrency", false))
+		RetryTopicConfigurationBuilder builder = RetryTopicConfigurationBuilder.newInstance()
 				.customBackoff(createBackoffFromAnnotation(annotation.backoff(), this.beanFactory))
 				.retryTopicSuffix(resolveExpressionAsString(annotation.retryTopicSuffix(), "retryTopicSuffix"))
 				.dltSuffix(resolveExpressionAsString(annotation.dltTopicSuffix(), "dltTopicSuffix"))
@@ -146,8 +144,17 @@ public class RetryableTopicAnnotationProcessor {
 				.autoStartDltHandler(autoStartDlt)
 				.setTopicSuffixingStrategy(annotation.topicSuffixingStrategy())
 				.sameIntervalTopicReuseStrategy(annotation.sameIntervalTopicReuseStrategy())
-				.timeoutAfter(timeout)
-				.create(getKafkaTemplate(resolveExpressionAsString(annotation.kafkaTemplate(), "kafkaTemplate"), topics));
+				.timeoutAfter(timeout);
+
+		Integer attempts = resolveExpressionAsInteger(annotation.attempts(), "attempts", true);
+		if (attempts != null) {
+			builder.maxAttempts(attempts);
+		}
+		Integer concurrency = resolveExpressionAsInteger(annotation.concurrency(), "concurrency", false);
+		if (concurrency != null) {
+			builder.concurrency(concurrency);
+		}
+		return builder.create(getKafkaTemplate(resolveExpressionAsString(annotation.kafkaTemplate(), "kafkaTemplate"), topics));
 	}
 
 	private SleepingBackOffPolicy<?> createBackoffFromAnnotation(Backoff backoff, BeanFactory beanFactory) { // NOSONAR
@@ -231,8 +238,8 @@ public class RetryableTopicAnnotationProcessor {
 
 	private String resolveExpressionAsString(String value, String attribute) {
 		Object resolved = resolveExpression(value);
-		if (resolved instanceof String sResolved) {
-			return sResolved;
+		if (resolved instanceof String str) {
+			return str;
 		}
 		else if (resolved != null) {
 			throw new IllegalStateException(THE_OSQ + attribute + "] must resolve to a String. "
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
```diff
@@ -173,9 +174,13 @@ private SleepingBackOffPolicy<?> createBackoffFromAnnotation(Backoff backoff, Be
...
 			}
-			policy.setInitialInterval(min);
+			if (min != null) {
+				policy.setInitialInterval(min);
+			}
 			policy.setMultiplier(multiplier);
-			policy.setMaxInterval(max > min ? max : ExponentialBackOffPolicy.DEFAULT_MAX_INTERVAL);
+			if (max != null && min != null && max > min) {
+				policy.setMaxInterval(max);
+			}
```
NathanQingyangXu, line 183: at least in theory, without nullness checking, NPE could be thrown while unboxing.

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
```diff
@@ -147,8 +144,17 @@ public RetryTopicConfiguration processAnnotation(String[] topics, Method method,
...
+
+		Integer attempts = resolveExpressionAsInteger(annotation.attempts(), "attempts", true);
+		if (attempts != null) {
+			builder.maxAttempts(attempts);
+		}
+		Integer concurrency = resolveExpressionAsInteger(annotation.concurrency(), "concurrency", false);
+		if (concurrency != null) {
+			builder.concurrency(concurrency);
+		}
+		return builder.create(getKafkaTemplate(resolveExpressionAsString(annotation.kafkaTemplate(), "kafkaTemplate"), topics));
```
NathanQingyangXu, line 157: there are many loose ends regarding nullness safety net. Not sure why there is no usage of the following two entries in the package-info.java within the package of `annotation`:
```
@org.springframework.lang.NonNullApi
@org.springframework.lang.NonNullFields
```
if we added them, many NPE issues in this package would be better exposed.

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
```diff
@@ -200,7 +210,6 @@ private EndpointHandlerMethod getDltProcessor(Method listenerMethod, Object bean
 				.orElse(RetryTopicConfigurer.DEFAULT_DLT_HANDLER);
 	}
 
-	@SuppressWarnings("deprecation")
```
NathanQingyangXu, line 203: seems leftover of previous JDK. In JDK17, I saw no compiling warnings after the removal of it.

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
```diff
@@ -378,8 +376,8 @@ private Object resolveExpression(String value) {
 	}
 
 	private String resolve(String value) {
-		if (this.beanFactory != null && this.beanFactory instanceof ConfigurableBeanFactory) {
-			return ((ConfigurableBeanFactory) this.beanFactory).resolveEmbeddedValue(value);
+		if (this.beanFactory instanceof ConfigurableBeanFactory cbf) {
+			return cbf.resolveEmbeddedValue(value);
```
NathanQingyangXu, line 380: of course, the original nullness checking is unnecessary for `instanceof` has included it already.

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
```diff
@@ -368,8 +366,8 @@ private List<Class<? extends Throwable>> resolveClasses(Class<? extends Throwabl
 	}
 
 	private Object resolveExpression(String value) {
-		String resolved = resolve(value);
 		if (this.expressionContext != null) {
+			String resolved = resolve(value);
```
NathanQingyangXu, line 370: seems the `resolved` is only used within the if block, so it makes more sense to move it within, so when `expressionContext` is null, no unnecessary overhead will incur.

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
sobychacko: Sure. If you are up to it, feel free to send another PR with that change.

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
```diff
@@ -173,9 +179,13 @@ private SleepingBackOffPolicy<?> createBackoffFromAnnotation(Backoff backoff, Be
 			if (backoff.random()) {
 				policy = new ExponentialRandomBackOffPolicy();
 			}
-			policy.setInitialInterval(min);
+			if (min != null) {
+				policy.setInitialInterval(min);
+			}
 			policy.setMultiplier(multiplier);
-			policy.setMaxInterval(max > min ? max : ExponentialBackOffPolicy.DEFAULT_MAX_INTERVAL);
+			if (max != null && min != null && max > min) {
```
sobychacko, line 186: What will' max' be if `max <= min`? Currently, we set it to `ExponentialBackOffPolicy.DEFAULT_MAX_INTERVAL` in that case. The new code seems to change that semantics.

sobychacko (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
NathanQingyangXu: Good comment! Below is the relevant code snippet:
```
public class ExponentialBackOffPolicy implements SleepingBackOffPolicy<ExponentialBackOffPolicy> {
        ... ...
	/**
	 * The maximum value of the backoff period in milliseconds.
	 */
	private long maxInterval = DEFAULT_MAX_INTERVAL;
```
so regardless of whether we set the same default value or not in `RetryableTopicAnnotationProcessor` above, the end result is the same. I avoided duplication to make it more elegant.

NathanQingyangXu: Let me know whether we need to revert back old behaviour. I am perfectly fine with either way.

sobychacko: But don't we need to set that value in `RetryableTopicAnnotationProcessor`? Otherwise, how does this `max` variable get to that value? I am fine with cleaning it up if it is redundant. I am trying to see the connection.

NathanQingyangXu: As I showed above, the policy class will set the fields to default values already just after the instance was initiated but before constructor is invoked, so after the default constructor was invoked, we have ended up with valid state already. 

Am I missing something here?

sobychacko: Ok, I got it. Thanks for the explanation.

sobychacko (approved):

PR #2969, opened 2023-12-22
GH-2968 `DefaultErrorHandler#handleBatchAndReturnRemaining` recovered invalid

### LINKED ISSUE #2968: DefaultErrorHandler#handleBatchAndReturnRemaining recovered invalid

For example:

Between 2.9 and 3.1.1

**Describe the bug**

DefaultErrorHandler#handleBatchAndReturnRemaining recovered invalid cause infinite loop when kafka listener threw `BatchListenerFailedException` and error record is first one of partition

```
		@KafkaListener(id = CONTAINER_ID_2, topics = "foo")
		public void foo2(List<String> in) {
			int index = 0;
			for (String str : in) {
				if ("qux".equals(str)) {    // Deliver 6 records from three partitions, qux is second partition first record
					throw new BatchListenerFailedException("test", index);
				}
				index++;
			}
		}
```

**Sample**
https://github.com/Wzy19930507/spring-boot-issue/tree/issue_231222

Wzy19930507:
Resolves #2968 

`DefaultErrorHandler#handleBatchAndReturnRemaining` recovered invalid and infinite loop when kafka listener threw BatchListenerFailedException and error record is first one in remaining list..

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedBatchProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedBatchProcessor.java
@@ -16,12 +16,10 @@
 
 package org.springframework.kafka.listener;
 
-import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -37,6 +35,7 @@ import org.apache.kafka.common.TopicPartition;
 
 import org.springframework.kafka.KafkaException;
 import org.springframework.kafka.KafkaException.Level;
+import org.springframework.kafka.support.KafkaUtils;
 import org.springframework.lang.Nullable;
 import org.springframework.util.backoff.BackOff;
 
@@ -50,6 +49,7 @@ import org.springframework.util.backoff.BackOff;
  *
  * @author Gary Russell
  * @author Francois Rosiere
+ * @author Wang Zhiyang
  * @since 2.8
  *
  */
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/FailedBatchProcessor.java
```diff
@@ -254,35 +246,33 @@ private <K, V> ConsumerRecords<K, V> seekOrRecover(Exception thrownException, @N
...
+						consumer)) {
+					remaining.remove(0);
 				}
-				Map<TopicPartition, List<ConsumerRecord<K, V>>> remains = new HashMap<>();
-				remaining.forEach(rec -> remains.computeIfAbsent(new TopicPartition(rec.topic(), rec.partition()),
-						tp -> new ArrayList<ConsumerRecord<K, V>>()).add((ConsumerRecord<K, V>) rec));
-				return new ConsumerRecords<>(remains);
 			}
+			catch (Exception e) {
+			}
```
sobychacko, line 256: Empty catch - can we address this?

sobychacko:
I went through the PR. It looks good. Could you explain why it is not a problem when the record that throws the error is not the last one in the partition?

Wzy19930507:
> I went through the PR. It looks good. Could you explain why it is not a problem when the record that throws the error is not the last one in the partition?

https://github.com/spring-projects/spring-kafka/blob/2cf5fbb571c91ef75241ad00fe8f32fff6607ccf/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedBatchProcessor.java#L257-L273

at line 262 - 265, if not the last one in the partition, we can recovered then remove it.

Wzy19930507:
> I went through the PR. It looks good. Could you explain why it is not a problem when the record that throws the error is not the last one in the partition?

my description mistake, error record must be first one in remaining list.

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedBatchProcessor.java
Wzy19930507: fixed

PR #2970, opened 2023-12-22
minor improvement error handler related

Wzy19930507:
* remove `FailedRecordProcessor.retryListeners`, reuse FailureTracker's retryListeners
* cleanup related to error handlers

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
@@ -90,8 +90,8 @@ class FailedRecordTracker implements RecoveryStrategy {
 			};
 		}
 		else {
-			if (recoverer instanceof ConsumerAwareRecordRecoverer) {
-				this.recoverer = (ConsumerAwareRecordRecoverer) recoverer;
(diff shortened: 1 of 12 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
```diff
@@ -150,7 +150,7 @@ public void setRetryListeners(RetryListener... listeners) {
 		this.retryListeners.addAll(Arrays.asList(listeners));
 	}
 
-	List<RetryListener> getRetryListeners() {
+	protected List<RetryListener> getRetryListeners() {
```
sobychacko, line 153: The class itself is package-protected. Why did you make this method `protected.`?

sobychacko (changes requested):

sobychacko:
@Wzy19930507 One comment. Otherwise, it looks good.

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
Wzy19930507: thanks for you review, package-protected is much better.

PR #2971, opened 2023-12-22
Update Spring Boot Version of the samples

jitokim:

### DIFF
--- a/samples/sample-01/pom.xml
+++ b/samples/sample-01/pom.xml
@@ -14,7 +14,7 @@
 	<parent>
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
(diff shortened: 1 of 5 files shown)


sobychacko:
@jitokim Spring Boot `3.2.1` is released. Why don't we use that instead of `3.2.0`?

jitokim:
> @jitokim Spring Boot `3.2.1` is released. Why don't we use that instead of `3.2.0`?

Actually, I wanted to switch to the 3.2.1 snapshot version after looking at the [release notes](https://github.com/spring-projects/spring-boot/releases).

But there was no version in the maven repo, so I made a mistake while lowering the version. 😢

artembilan:
No problem!
We are accepting your change.
Thank you!

And for the future we are going to apply Dependabot for these Maven projects.

PR #2976, opened 2023-12-25
minor improvement `DelegatingInvocableHandler` and AKLE relate

Wzy19930507:

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/DelegatingInvocableHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/DelegatingInvocableHandler.java
@@ -112,8 +112,8 @@ public class DelegatingInvocableHandler {
 		this.bean = bean;
 		this.resolver = beanExpressionResolver;
 		this.beanExpressionContext = beanExpressionContext;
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/DelegatingInvocableHandler.java
```diff
@@ -246,36 +246,32 @@ protected InvocableHandlerMethod findHandlerForPayload(Class<? extends Object> p
...
-					|| !methodParameter.hasParameterAnnotation(Header.class))
-					&& methodParameter.getParameterType().isAssignableFrom(payloadClass)) {
+			boolean isPayload = assignPayload(methodParameter, payloadClass);
+			if (isPayload) {
 				if (this.validator != null) {
 					this.payloadMethodParameters.put(handler, methodParameter);
 				}
-				return true;
 			}
+			return isPayload;
```
Wzy19930507, line 274: When `isPayload` is false, will call `findCandidate` then return null.

Wzy19930507, line 274: When isPayload is false, will call findCandidate then return null.

Issue #2991, opened 2024-01-17
False warning log for not committed offsets on CooperativeStickyAssignor strategy


hrabur:
spring-kafka:3.0.12 (code of `main` and `3.0.x` branches show the same problem)

Recently we switched to `partition.assignment.strategy: org.apache.kafka.clients.consumer.CooperativeStickyAssignor`, as a consequence, we started to observe false positive warning messages during the restart of our service

> These offsets could not be committed; partition(s) lost during rebalance: {}

The message led us to the following code in `spring-kafka`

https://github.com/spring-projects/spring-kafka/blob/659b066291051d804563f8fe2f11aabba131c75a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1714

We believe the the warning is false positive, because as seen from the log message the `uncommitted` map is empty (`... rebalance: {}`), i.e. all offsets are properly committed.

As it is seen in the code the warning log message is printed always, without taking into consideration if `uncommitted` is empty or not

PR #2994, opened 2024-01-17
GH-2991: Fix unwanted warning logs in KMLC

sobychacko:
Fixes: #2991

There is a warning log from `KafkaMessageListenerContainter#checkRebalanceCommits` that unconditionally prints that offsets could not be committed. Because of this unconditional check, the messages are very likely to be false positives. Fixing this by adding a check to see if there are indeed uncommitted records.


artembilan:
... and cherry-picked to `3.0.x`

PR #3009, opened 2024-01-31
Implementation Deserializer.deserialize(String, Headers, ByteBuffer)

Wzy19930507:
adapt apache kafka feature, reduce `CompletedFetch#parseRecord()` memory copy, details see https://issues.apache.org/jira/browse/KAFKA-14944

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2021 the original author or authors.
+ * Copyright 2019-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
```diff
@@ -75,4 +78,9 @@ public Object deserialize(String topic, Headers headers, byte[] data) {
 		return findDelegate(topic).deserialize(topic, headers, data);
 	}
 
+	@Override
```
sobychacko, line 81: Need the `since` tag.

sobychacko (changes requested):

sobychacko:
@Wzy19930507 one minor comment. I will do a detailed review today though.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicDeserializer.java
artembilan: No need: this is an impl of existing API.
Technically it was there even before, but relied on a `default` method.

So, all good here.

artembilan:
@Wzy19930507 ,

merged and thank you, but please, don't do not feature-related refactoring.
Better to do that work in separate PRs for easier and faster review.

Wzy19930507:
Thanks for the review and advice!
I will often remind myself.

Issue #609, opened 2018-03-20
Unable to read the 'paused' state of a AbstractMessageListenerContainer


patrickherrera:
Hi, I have a use case where I wish to pause consumption on a topic to allow a higher-priority topic to get ahead.

I can see that I can pause a listener container via the public pause() method and this seems to work, but for some reason the method to return the 'isPaused' state is protected.

Is there a reason why this is not public?  Could that be changed?

Thanks

garyrussell:
Yes, we could make it public but it doesn't really reflect whether the container is actually paused yet, just that `pause()` has been called (it takes a little while before the actual pause occurs).

It that's all you need, it's a simple fix; if you actually need to know the actual state, it would take a bit more work.

We should also consider publishing events when each consumer thread actually pauses.

garyrussell:
@patrickherrera The fix is on master and is available in 2.1.5.BUILD-SNAPSHOT.

We're currently planning on releasing 2.1.5.RELEASE next Wednesday.

patrickherrera:
Thanks guys, fantastic work!  I'll check it out right now

PR #3016, opened 2024-02-03
Remove AbstractMessageListenerContainer.isPaused

Wzy19930507:
Use AbstractMessageListenerContainer method `isPauseRequested` uniformly instead of `isPaused()`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -67,6 +67,7 @@ import org.springframework.util.StringUtils;
  * @author Marius Bogoevici
  * @author Artem Bilan
  * @author Tomaz Fernandes
+ * @author Wang Zhiyang
  */
 public abstract class AbstractMessageListenerContainer<K, V>
 		implements GenericMessageListenerContainer<K, V>, BeanNameAware, ApplicationEventPublisherAware,
@@ -269,6 +270,7 @@ public abstract class AbstractMessageListenerContainer<K, V>
 		return this.running;
 	}
 
+	@Deprecated(since = "3.2", forRemoval = true)
 	protected boolean isPaused() {
 		return this.paused;
 	}
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -300,7 +300,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 
 	@Override
 	public boolean isContainerPaused() {
-		return isPaused() && this.listenerConsumer != null && this.listenerConsumer.isConsumerPaused();
+		return isPauseRequested() && this.listenerConsumer != null && this.listenerConsumer.isConsumerPaused();
 	}
 
 	@Override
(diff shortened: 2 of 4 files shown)


Wzy19930507:
I executed `./gradlew check` again and executed `TestOOMError` multiple times, but can not reproduce the error in the image
image

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -269,10 +270,6 @@ public boolean isRunning() {
 		return this.running;
 	}
 
-	protected boolean isPaused() {
```
sobychacko, line 272: Although it is a protected method, the fact that it is not private means that a downstream implementation can technically access it and override it. Therefore, we can't remove the method. We need to deprecate it.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -299,7 +299,7 @@ public Map<String, Collection<TopicPartition>> getAssignmentsByClientId() {
 
 	@Override
 	public boolean isContainerPaused() {
-		return isPaused() && this.listenerConsumer != null && this.listenerConsumer.isConsumerPaused();
+		return isPauseRequested() && this.listenerConsumer != null && this.listenerConsumer.isConsumerPaused();
```
sobychacko, line 302: I would think that `isPaused` is more appropriate here. Thus, we need to reevaluate the removal of `isPaused` from the container API. What is your reasoning for this change?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1635,7 +1635,7 @@ private ConsumerRecords<K, V> doPoll() {
 						KafkaMessageListenerContainer.this.emergencyStop.run();
 					}
 					TopicPartition firstPart = this.remainingRecords.partitions().iterator().next();
-					boolean isPaused = isPaused() || isPartitionPauseRequested(firstPart);
+					boolean isPaused = isPauseRequested() || isPartitionPauseRequested(firstPart);
```
sobychacko, line 1638: `isPaused` maybe more appropriate semantically here.

sobychacko:
I am not sure whether we should remove the `isPaused` method from the container API as it makes more sense than the `isPauseRequested` counterpart.

Wzy19930507:
See #609 about `isPaused()` 
>  container is actually paused yet, just that pause() has been called (it takes a little while before the actual pause occurs).

look like same as,
```
	/**
	 * Return true if {@link #pause()} has been called; and all consumers in this container
	 * have actually paused.
	 * @return true if the container is paused.
	 * @since 2.1.5
	 */
	default boolean isContainerPaused() {
		throw new UnsupportedOperationException("This container doesn't support pause/resume");
	}
```

`isPaused()` use to adaptor resumed/paused, see [spring-integration-kafka#198](https://github.com/spring-projects/spring-integration-kafka/issues/198). but now container has two phase for paused/resume `isPauseRequested` and `isContainerPaused`.

artembilan:
Yes. I totally agree with the change.

@sobychacko , 
please, proceed with the detailed review.

I'm not sure how that `ConcurrentModificationException` is possible: we have all the `this.containers` guarded by the `this.lifecycleLock.lock()` and I don't see any modifications from the outside in that test...

sobychacko:
Looks like the tests are green now. I will do another pass of the review. Don't you think the `isPaused` removal deprecated first since its `protected`?

sobychacko:
Sorry, I missed that.

Issue #3011, opened 2024-01-31
Provide a way to support Consumer#enforceRebalance


sobychacko:
Currently, we can use the `RecordInterceptor` to access the consumer and trigger an `enforceRebalance`. Explore an option to see if we can support this operation as a first-class citizen from the container. Fore more details: https://kafka.apache.org/36/javadoc/org/apache/kafka/clients/consumer/Consumer.html#enforceRebalance()

PR #3021, opened 2024-02-08
GH-3011: Support enforced consumer rebalance

sobychacko:
Fixes: #3011

Kafka consumer API supports an enforced rebalance. Provide an option via the message listener container to trigger this operation.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 the original author or authors.
+ * Copyright 2016-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
 import java.util.regex.Pattern;
@@ -68,6 +69,7 @@ import org.springframework.util.StringUtils;
  * @author Artem Bilan
  * @author Tomaz Fernandes
  * @author Wang Zhiyang
+ * @author Soby Chacko
  */
 public abstract class AbstractMessageListenerContainer<K, V>
 		implements GenericMessageListenerContainer<K, V>, BeanNameAware, ApplicationEventPublisherAware,
@@ -134,6 +136,8 @@ public abstract class AbstractMessageListenerContainer<K, V>
 
 	@Nullable
 	private KafkaAdmin kafkaAdmin;
+
+	protected AtomicBoolean enforceRebalanceRequested = new AtomicBoolean();
 
 	/**
 	 * Construct an instance with the provided factory and properties.
@@ -617,6 +621,15 @@ public abstract class AbstractMessageListenerContainer<K, V>
 		finally {
 			this.lifecycleLock.unlock();
 		}
+	}
+
+	@Override
+	public void enforceRebalance() {
+		this.enforceRebalanceRequested.set(true);
+	}
+
+	protected boolean isEnforceRebalanceRequested() {
+		return this.enforceRebalanceRequested.get();
 	}
 
 	@Override
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -625,12 +625,6 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	}
 
 	@Override
-	public void enforceRebalance() {
-		throw new UnsupportedOperationException("enforceRebalance() is supposed to be implemented " +
-				"by concrete implementations only.");
-	}
-
-	@Override
 	public void pause() {
 		this.paused = true;
 	}
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -92,6 +92,8 @@ public abstract class AbstractMessageListenerContainer<K, V>
 
 	protected final ReentrantLock lifecycleLock = new ReentrantLock(); // NOSONAR
 
+	protected final AtomicBoolean enforceRebalanceRequested = new AtomicBoolean();
+
 	private final Set<TopicPartition> pauseRequestedPartitions = ConcurrentHashMap.newKeySet();
 
 	@NonNull
@@ -137,7 +139,6 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	@Nullable
 	private KafkaAdmin kafkaAdmin;
 
-	protected final AtomicBoolean enforceRebalanceRequested = new AtomicBoolean();
 
 	/**
 	 * Construct an instance with the provided factory and properties.
@@ -625,7 +626,8 @@ public abstract class AbstractMessageListenerContainer<K, V>
 
 	@Override
 	public void enforceRebalance() {
-		this.enforceRebalanceRequested.set(true);
+		throw new UnsupportedOperationException("enforceRebalance() is supposed to be implemented " +
+				"by concrete implementations only.");
 	}
 
 	@Override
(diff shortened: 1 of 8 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1739,6 +1748,18 @@ private void sleepFor(Duration duration) {
 			}
 		}
 
+		private void enforceRebalanceIfNecessary() {
+			final AtomicBoolean enforceRebalanceRequested = KafkaMessageListenerContainer.super.enforceRebalanceRequested;
```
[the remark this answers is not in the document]
sobychacko: good point. Let me look into that.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -135,6 +137,8 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	@Nullable
 	private KafkaAdmin kafkaAdmin;
 
+	protected final AtomicBoolean enforceRebalanceRequested = new AtomicBoolean();
```
artembilan, line 140: Wrong order of properties: has to be after `lifecycleLock`

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -619,6 +623,11 @@ public final void stop(boolean wait) {
 		}
 	}
 
+	@Override
+	public void enforceRebalance() {
+		this.enforceRebalanceRequested.set(true);
```
artembilan, line 628: I think we have to leave it here without any implementation.
The point is that we can simply set the value in that `protected final` from target implementations.

spring-kafka/src/test/java/org/springframework/kafka/listener/ContainerEnforceRebalanceTests.java
```diff
@@ -0,0 +1,144 @@
...
+		final MessageListenerContainer listenerContainer = registry.getListenerContainer("enforce-rebalance-grp");
+		assertThat(config.listenerLatch.await(10, TimeUnit.SECONDS)).isTrue();
+		assertThat(listenerContainer).isNotNull();
+		listenerContainer.enforceRebalance();
+		// The test is expecting partition revoke once and assign twice.
+		assertThat(config.partitionRevokedLatch.await(10, TimeUnit.SECONDS)).isTrue();
+		assertThat(config.partitionAssignedLatch.await(10, TimeUnit.SECONDS)).isTrue();
+		final List<? extends KafkaMessageListenerContainer<?, ?>> containers =
+				((ConcurrentMessageListenerContainer<?, ?>) listenerContainer).getContainers();
+		assertThat(containers.get(0).enforceRebalanceRequested).isFalse();
```
artembilan, line 73: I believe `thisOrParentContainer` doesn't work for you (yet), because you check the `enforceRebalanceRequested` on the child container, but directly on the one requested from the `registry`

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -2486,6 +2486,34 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
 		logger.info("Stop rebalance after failed record");
 	}
 
+	@Test
+	void enforceRabalanceOnTheConsumer() throws Exception {
+		ConsumerFactory<Integer, String> cf = mock();
+		ContainerProperties containerProps = new ContainerProperties("foo");
```
artembilan, line 2492: I said before that I don't like non-official language in open source.
Please, consider to use some other words.
Even if we have it somewhere in the code already, that doesn't mean we should continue to use what might be harmful one day.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1739,6 +1749,20 @@ private void sleepFor(Duration duration) {
 			}
 		}
 
+		private void enforceRebalanceIfNecessary() {
+			try {
+				if (KafkaMessageListenerContainer.this.enforceRebalanceRequested.get()) {
+					final String enforcedRebalanceReason = String.format("Enforced rebalance requested for container: %s",
+							KafkaMessageListenerContainer.this.getListenerId());
+					KafkaMessageListenerContainer.this.logger.info(enforcedRebalanceReason);
```
artembilan, line 1757: There is local `ListenerConsumer.logger` reference to this one.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -314,6 +314,15 @@ public boolean isInExpectedState() {
 		return isRunning() || isStoppedNormally();
 	}
 
+	@Override
+	public void enforceRebalance() {
+		super.enforceRebalance();
```
artembilan, line 319: Please, see my comment in your test below and consider to use this `this.thisOrParentContainer.enforceRebalanceRequested.set(true)` instead.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1739,6 +1749,20 @@ private void sleepFor(Duration duration) {
 			}
 		}
 
+		private void enforceRebalanceIfNecessary() {
+			try {
+				if (KafkaMessageListenerContainer.this.enforceRebalanceRequested.get()) {
```
artembilan, line 1754: Therefore then here: `KafkaMessageListenerContainer.this.thisOrParentContainer.enforceRebalanceRequested.get()`

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1739,6 +1749,20 @@ private void sleepFor(Duration duration) {
...
+					final String enforcedRebalanceReason = String.format("Enforced rebalance requested for container: %s",
+							KafkaMessageListenerContainer.this.getListenerId());
+					KafkaMessageListenerContainer.this.logger.info(enforcedRebalanceReason);
+					this.consumer.enforceRebalance(enforcedRebalanceReason);
+				}
+			}
+			finally {
+				KafkaMessageListenerContainer.this.enforceRebalanceRequested.set(false);
```
artembilan, line 1762: DITTO

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -625,7 +626,8 @@ public final void stop(boolean wait) {
 
 	@Override
 	public void enforceRebalance() {
-		this.enforceRebalanceRequested.set(true);
+		throw new UnsupportedOperationException("enforceRebalance() is supposed to be implemented " +
+				"by concrete implementations only.");
```
artembilan, line 630: This is redundant: you say the same in the `default` method of the `MessageListenerContainer` interface.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1751,15 +1751,15 @@ private void sleepFor(Duration duration) {
 
 		private void enforceRebalanceIfNecessary() {
 			try {
-				if (KafkaMessageListenerContainer.this.enforceRebalanceRequested.get()) {
+				if (KafkaMessageListenerContainer.this.thisOrParentContainer.enforceRebalanceRequested.get()) {
 					final String enforcedRebalanceReason = String.format("Enforced rebalance requested for container: %s",
 							KafkaMessageListenerContainer.this.getListenerId());
-					KafkaMessageListenerContainer.this.logger.info(enforcedRebalanceReason);
+					ListenerConsumer.this.logger.info(enforcedRebalanceReason);
```
artembilan, line 1757: I don't think we need `ListenerConsumer` prefix when we deal in the context of exactly `ListenerConsumer`.
I also think that those local `final` for variables give us a little more noise reading this code.

spring-kafka/src/test/java/org/springframework/kafka/listener/ContainerEnforceRebalanceTests.java
artembilan: I wonder why you have not fixed this part in the test...
Technically it is always `false` for any child containers in the `ConcurrentMessageListenerContainer`.
We need to be sure that it is `true` exactly in the `ConcurrentMessageListenerContainer` according to our new logic around `thisOrParentContainer`.

Or do I miss something?

artembilan (changes requested):

artembilan:
...and cherry-picked to `3.1.x` as https://github.com/spring-projects/spring-kafka/commit/e518f1ace98446dd71e220e086a62f7cf8eaae36 after minor conflict resolution.

PR #3033, opened 2024-02-13
Clarify retries in KSIQS docs

sobychacko:
* Clarify how KafkaStreamsInteractiveQueryService retries work in the docs

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/streams.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/streams.adoc
@@ -369,11 +369,9 @@ NOTE: `KafkaStreamsInteractiveQueryService` API in Spring for Apache Kafka only 
 
 When trying to retrieve the state store using the `KafkaStreamsInteractiveQueryService`, there is a chance that the state store might not be found for various reasons.
 If those reasons are transitory, `KafkaStreamsInteractiveQueryService` provides an option to retry the retrieval of the state store by allowing to inject a custom `RetryTemplate`.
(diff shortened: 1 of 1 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/streams.adoc
```diff
@@ -365,6 +365,30 @@ When calling this method, the user can specifially ask for the proper state stor
 
 NOTE: `KafkaStreamsInteractiveQueryService` API in Spring for Apache Kafka only supports providing access to local key-value stores at the moment.
 
+==== Retrying State Store Retrieval
+
+When trying to retrieve the state store using the `KafkaStreamsInteractiveQueryService`, there is a chance that the state store might not be found for various reasons.
+If those reasons are transitory, `KafkaStreamsInteractiveQueryService` provides an option to retry the retrieval of the state store by allowing to inject a custom `RetryTemplate`.
+By default, the `RetryTemmplate` that is used in `KafkaStreamsInteractiveQueryService` does not attempt any retries if the first attempt to retrieve the state store fails.
```
artembilan, line 372: > does not attempt any retries

That's not correct. See `RetryTemplate.retryPolicy` by default:
```
private volatile RetryPolicy retryPolicy = new SimpleRetryPolicy(3);
```

Or do you just talk about something else, like you are mentioning above about `back-off`?

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/streams.adoc
sobychacko: Got it. I want to reset that default to a single try by default, which I will do via another PR. For now, I synched the docs with the current state of the code.

Issue #3032, opened 2024-02-13
Provide a recovery callback for the execute method of RetryingDeserializer 


vspiliop:
**Expected Behavior**

Provide a recovery callback for the execute method of `org.springframework.kafka.support.serializer.RetryingDeserializer `. Maybe make necessary fields protected from private and create `class RetryingAndRecoveringDeserializer extends RetryingDeserializer`

**Current Behavior**

No such option at the moment.

I would like to be able to re-try for example HTTP 500 Schema Registry errors in a blocking way for e.g. one minute and then trigger the `ErrorHandlingDeserializer` so that we can re-consume and start the blocking re-tries and so on so forth. This is helpful for avoiding consumer group re-balancing, by retrying forever in a blocking fashion. But if it is not a recoverable error (e.g.  HTTP 400 Schema Registry), then I would like to skip the event consumed (by instructing `ErrorHandlingDeserializer ` to skip).

An example could be:

```
  @Bean
  public ErrorHandlingDeserializer<Object> retryingValueDeserializer(
      RetryTemplate retryRegistryhttp500ErrorsTemplate,
      SchemaRegistryClient schemaRegistryClient) {
    return new ErrorHandlingDeserializer<>(
        new RetryingAndRecoveringDeserializer<>(
            new KafkaAvroDeserializer(schemaRegistryClient),
            retryRegistryhttp500ErrorsTemplate,
            this::triggerErrorHandlingDeserializerToReconsumeOrSkip));
  }

  private <T> T triggerErrorHandlingDeserializerToReconsumeOrSkip(RetryContext context) {
    if (isRecoverable(context)) {
      throw new BlockingRetriesExhaustedRecoverableException();
    } else {
      throw new BlockingRetriesExhaustedNonRecoverableException();
    }
  }

  /**
   * Non-recoverable exceptions are not retried by the blocking retry template and retry count is
   * always 1.
   */
  private boolean isRecoverable(RetryContext context) {
    return context.getRetryCount() != 1;
  }

  @Bean
  public RetryTemplate retryRegistryhttp500ErrorsTemplate() {

    ExpressionRetryPolicy retryRegistry5xxErrors =
        new ExpressionRetryPolicy(
            "cause instanceof T(io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException) and cause?.errorCode >= 500");
    retryRegistry5xxErrors.setMaxAttempts(blockingRetries);

    return RetryTemplate.builder()
[...]

PR #3035, opened 2024-02-15
GH-3032: RecoveryCallback on RetryingDeserializer

sobychacko:
Fixes: #3032

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/RetryingDeserializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/RetryingDeserializer.java
@@ -22,6 +22,8 @@ import java.util.Map;
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.serialization.Deserializer;
 
+import org.springframework.lang.Nullable;
+import org.springframework.retry.RecoveryCallback;
 import org.springframework.retry.RetryOperations;
 import org.springframework.util.Assert;
 
@@ -30,18 +32,19 @@ import org.springframework.util.Assert;
  * deserialization in case of transient errors.
  *
  * @param <T> Type to be deserialized into.
- *
  * @author Gary Russell
  * @author Wang Zhiyang
- *
+ * @author Soby Chacko
  * @since 2.3
- *
  */
 public class RetryingDeserializer<T> implements Deserializer<T> {
 
 	private final Deserializer<T> delegate;
 
 	private final RetryOperations retryOperations;
+
+	@Nullable
+	private RecoveryCallback<T> recoveryCallback;
 
 	public RetryingDeserializer(Deserializer<T> delegate, RetryOperations retryOperations) {
 		Assert.notNull(delegate, "the 'delegate' deserializer cannot be null");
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/serializer/RetryingDeserializer.java
```diff
@@ -57,22 +60,31 @@ public void configure(Map<String, ?> configs, boolean isKey) {
...
 	public void close() {
 		this.delegate.close();
 	}
 
+	/**
+	 * Set a recovery callback to execute when the retries are exhausted.
+	 * @param recoveryCallback {@link RecoveryCallback} to execute
+	 * @since 3.1.2
+	 */
+	public void setRecoveryCallback(@Nullable RecoveryCallback<T> recoveryCallback) {
```
artembilan, line 86: Setters must follow just after constructors.

spring-kafka/src/test/java/org/springframework/kafka/support/serializer/RetryingDeserializerTests.java
```diff
@@ -54,6 +55,39 @@ void testRetry() {
 		rdes.close();
 	}
 
+	@Test
+	void retryDeserializerWithRecoveryCallback() throws InterruptedException {
+		NonRecoverableDeser delegate = new NonRecoverableDeser();
+		RetryingDeserializer<String> rdes = new RetryingDeserializer<>(delegate, new RetryTemplate());
```
artembilan, line 61: You also can just inject a lambda which would throw that exception.
I don't think it matters how many time we retry.
Only what is important for this test that `RecoveryCallback` is called.

spring-kafka/src/test/java/org/springframework/kafka/support/serializer/RetryingDeserializerTests.java
```diff
@@ -54,6 +55,39 @@ void testRetry() {
...
+		rdes.setRecoveryCallback(context -> {
+			delegate.latch.countDown();
```
artembilan, line 63: You can just sent here a result of `mock()` and then verify in the end that it was called.
There is no any race conditions: everything in the test is called sequentially.

spring-kafka/src/test/java/org/springframework/kafka/support/serializer/RetryingDeserializerTests.java
```diff
@@ -54,6 +55,39 @@ void testRetry() {
...
+		}
+
+		@Override
+		public String deserialize(String topic, byte[] data) {
+			latch.countDown();
+			throw new RuntimeException();
+		}
+
+		@Override
+		public void close() {
```
artembilan, line 86: The `Deserializer` is functional interface, so no need to override those `default` methods.
See my comment above about its simple lambda for the `RetryingDeserializer` ctor.

artembilan (changes requested):
Also missed an entry in the `whats-new.adoc`.

Thanks

sobychacko:
> Also missed an entry in the `whats-new.adoc`.
> 
> Thanks

Aren't we going to backport the change to `3.1.x` though?

spring-kafka/src/test/java/org/springframework/kafka/support/serializer/RetryingDeserializerTests.java
```diff
@@ -54,6 +59,19 @@ void testRetry() {
 		rdes.close();
 	}
 
+	@SuppressWarnings("unchecked")
+	@Test
+	void retryingDeserializerWithRecoveryCallback() throws Exception {
+		RetryingDeserializer<String> rdes = new RetryingDeserializer<>((s, b) -> {
+			throw new RuntimeException();
+		}, new RetryTemplate());
+		RecoveryCallback<String> recoveryCallback = mock(RecoveryCallback.class);
```
artembilan, line 68: I know that is going to be nit-picking, but I'll shoot it anyway.

The Mockito starting with some version does not need the type argument.
Just `RecoveryCallback<String> recoveryCallback = mock();` is enough to make it working.
And the code looks nicer.

spring-kafka/src/test/java/org/springframework/kafka/support/serializer/RetryingDeserializerTests.java
```diff
@@ -54,6 +59,19 @@ void testRetry() {
...
+		rdes.setRecoveryCallback(recoveryCallback);
+		rdes.deserialize("my-topic", "my-data".getBytes());
+		verify(recoveryCallback, times(1)).recover(any(RetryContext.class));
```
artembilan, line 71: The ` times(1)` is there by default:
```
    public static <T> T verify(T mock) {
        return MOCKITO_CORE.verify(mock, times(1));
    }
```
Why do you use this explicitly?

spring-kafka/src/test/java/org/springframework/kafka/support/serializer/RetryingDeserializerTests.java
```diff
@@ -54,6 +59,19 @@ void testRetry() {
...
+		rdes.close();
```
artembilan, line 72: There is nothing to close. See our delegate lambda.
IMHO the test should be as short as possible and be focused on the specific task.
If tests are too complicated and they are flaky, it is hard to find the reason and solution.
I know: this one is not, but better to have a good habit for less code - better support!

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/support/serializer/RetryingDeserializerTests.java
sobychacko: ok, will cleanup, i was just following original test patterns.

artembilan (approved):
LGTM.
Will merger when PR build is green.

vspiliop:
@artembilan How can I find when `spring-kafka v3.1.2` will be included in a 

```
  <parent>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-dependencies</artifactId>
    <version>...</version>
  </parent>
```

Should I assume in the next spring boot release? Is there a link with planned future spring boot releases?

Thanks again!

Issue #3001, opened 2024-01-24
Use application name as default clientId


notizklotz:
**Expected Behavior**

If `spring.application.name` is defined, it should be used as default Kafka clientId unless overriden by more specific configs. I get the behaviour that I think would be a good default by defining this in my application:

`spring.kafka.client-id=${spring.application.name}`

**Current Behavior**

The default clientIds are "producer", "adminclient", etc.

**Context**

We operate fairly large shared Kafka clusters and a couple of important metrics and log output by Kafka only includes the clientId and not the username. Especially for producers this means we see many of our customers implicitly connecting using "producer-1" as clientId as there is no way on Kafka server side to enforce a specific pattern. By using a more specific default this would ease problem analysis on server side. `spring.application.name` would be an ideal default because this is also used for similar use cases with other technologies.

I could have a look at creating a pull request for this, if this isn't something that has already been rejected in the past.

sobychacko:
@notizklotz I think we should still preserve `producer,` `consumer,` `admin client` etc. in the client-id name to distinguish them, but we can attach the `spring.application.name` for shared cluster environments where there are multiple apps connecting to Kafka. A PR would be greatly welcomed!

artembilan:
Note: the `spring.kafka.client-id` is the part of Spring Boot.
But Soby is right: this common property is out of use and those API-specific are there for consideration.

The "producer", "consumer" and "adminclient" comes from bean names auto-configured for those specific Kafka clients.

We may indeed consider to include `spring.application.name` into those properties.
For example, `DefaultKafkaProducerFactory.CloseSafeProducer`:
```
this.clientId = factoryName + "." + id;
```
And somewhere there we can resolve it like `this.applicationContext.getEnvironment().getProperty("spring.application.name");`

notizklotz:
I started implementing a proof of concept and it seems to work out well, especially for the cases which currently have a very generic clientId (Consumer without Consumer Group, Producer, Admin Client).

I chose to put the application name before the type (`myapp-consumer-1` instead of `consumer-myapp-1`) to differentiate it better from the Consumer with Consumer Group defaults, which have the Consumer Group name after the type. And it is more similar to the Kafka Streams defaults.

The following table with examples assumes `spring.application.name=myapp`

| Type                  | Before                                                                                                                                               | After                                                                                                                                                | Notes                                                                                                                                                                                                    |
[...]

PR #3048, opened 2024-02-19
GH-3001: default clientIds with application name

notizklotz:
Fixes: #GH-3001

Use Spring Boot's `spring.application.name` property as part of the default clientIds for Consumers, Producers and AdminClients. Helps with identifying problematic clients at server side.

* Only use as a fallback if clientId wasn't specified explicitly
* Do not use for Consumers with a specified groupId because KafkaConsumer will use the groupId as clientId which already is an identifiable default

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/connecting.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/connecting.adoc
@@ -58,3 +58,37 @@ These listeners can be used, for example, to create and bind a Micrometer `Kafka
 
 The framework provides listeners that do exactly that; see xref:kafka/micrometer.adoc#micrometer-native[Micrometer Native Metrics].
 
+[[default-client-id-prefixes]]
+== Default client ID prefixes
+
+Starting with version 3.2, for Spring Boot applications which define an application name using the `spring.application.name` property, this name is now used
+as a default prefix for auto-generated client IDs for these client types:
+
+- consumer clients which don't use a consumer group
+- producer clients
+- admin clients
+
+This makes it easier to identify these clients at server side for troubleshooting or applying quotas.
+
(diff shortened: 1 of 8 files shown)


sobychacko:
@notizklotz I am Starting to review the PR. There are some build failures due to checkstyle issues. See [here](https://github.com/spring-projects/spring-kafka/actions/runs/7960753296/job/22101177475?pr=3048). 
Also, you need to add some docs about these changes.

notizklotz:
> @notizklotz I am Starting to review the PR. There are some build failures due to checkstyle issues. See [here](https://github.com/spring-projects/spring-kafka/actions/runs/7960753296/job/22101177475?pr=3048). Also, you need to add some docs about these changes.

@sobychacko I fixed the checkstyle errors and added some docs to the "What’s new?" section

sobychacko:
@notizklotz Thanks. The problem with adding this in the `whats-new` doc is that this will be replaced for the next version (`3.3.0`). This deserves a separate small section inside the reference docs. Maybe add a new section for configuring `client.id`? One other thing - can you add your name as the `@author` in all the classes you modified and update any copyright years?

notizklotz:
@sobychacko Added section to reference docs, added author tag, checked copyright years, squashed commits and rebased to upstream main

sobychacko:
Thanks @notizklotz for the updates. One last comment - What do you think about extracting `spring.application.name` text literal into a common property since it is repeated in a handful of places?

notizklotz:
> Thanks @notizklotz for the updates. One last comment - What do you think about extracting `spring.application.name` text literal into a common property since it is repeated in a handful of places?

@sobychacko I considered that but decided against it, because I couldn't find an existing class where I think it would be a good fit. I didn't want to introduce a new class like "Constants" just for this literal. I also saw that the "client-id" literal is also copy/pasted across various classes and I thought to stick with the existing style.

sobychacko:
Fair enough. We can always refactor that later if need be.

Issue #3050, opened 2024-02-20
CommonDelegatingErrorHandler does not check delegates compatibility correctly


antonin-arquey:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.1.0

**Describe the bug**

When adding a new delegate to a `CommonDelegatingErrorHandler`, the check for compatibility with `ackAfterHandle` and `seeksAfterHandling` is not performed.

The following method does not check the `delegatesToCheck` but uses the current `this.delegates.values()` which does not yet contains the new delegates:

```java
@SuppressWarnings("deprecation")
private void checkDelegatesAndUpdateClassifier(Map<Class<? extends Throwable>,
			CommonErrorHandler> delegatesToCheck) {

        boolean ackAfterHandle = this.defaultErrorHandler.isAckAfterHandle();
        boolean seeksAfterHandling = this.defaultErrorHandler.seeksAfterHandling();
        this.delegates.values().forEach(handler -> {
	        Assert.isTrue(ackAfterHandle == handler.isAckAfterHandle(),
			        "All delegates must return the same value when calling 'isAckAfterHandle()'");
	        Assert.isTrue(seeksAfterHandling == handler.seeksAfterHandling(),
			        "All delegates must return the same value when calling 'seeksAfterHandling()'");
        });
        updateClassifier(delegatesToCheck);
}
```

**To Reproduce**

Create a `CommonDelegatingErrorHandler` and add an incompatible `CommonErrorHandler`, no exception will be thrown.

**Expected behavior**

The new delegates compatibility should be validated.

antonin-arquey:
opened a small PR to fix the compatibility check when adding a delegate

PR #3051, opened 2024-02-20
GH-3050: Delegating EH delegates compatibility

antonin-arquey:
Fixes: #3050

Correct CommonDelegatingErrorHandler validation for delegates compatibility.

Add documentation stating that delegates must be compatible with default error handler.


sobychacko:
@antonin-arquey, Could you add your name as an author in the classes you modified?

antonin-arquey:
@sobychacko added the missing author tags

PR #3064, opened 2024-02-23
GH-3057: ReactiveKafkaConsumerTemplate missing receiveBatch method

### LINKED ISSUE #3057: ReactiveKafkaConsumerTemplate missing receiveBatch method
**Expected Behavior**

There is a ReactiveKafkaConsumerTemplate#receiveBatch  method.

**Current Behavior**

The ReactiveKafkaConsumerTemplate class is missing the KafkaReceiver#receiveBatch method wrapper

**Context**

marcusvoltolim:
* Adding wrapper method `receiveBatch` in ReactiveKafkaConsumerTemplate.

Closes #3057

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/reactive/ReactiveKafkaConsumerTemplate.java
@@ -72,8 +72,8 @@ public class ReactiveKafkaConsumerTemplate<K, V> {
 		return this.kafkaReceiver.receive();
 	}
 
-	public Flux<ReceiverRecord<K, V>> receiveBatch() {
-		return this.kafkaReceiver.receiveBatch().concatMap(Function.identity());
+	public Flux<Flux<ReceiverRecord<K, V>>> receiveBatch() {
+		return this.kafkaReceiver.receiveBatch();
 	}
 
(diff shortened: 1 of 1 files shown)


sobychacko:
@marcusvoltolim Why are you changing the batch semantics in the API? The `batchReceives` has a return of `Flux<Flux<ReceiverRecord<K, V>>>` and you are flat mapping it. Why can't we let the user make that decision?

marcusvoltolim:
> @marcusvoltolim Why are you changing the batch semantics in the API? The `batchReceives` has a return of `Flux<Flux<ReceiverRecord<K, V>>>` and you are flat mapping it. Why can't we let the user make that decision?

I used `receiveAutoAck` as an example, which does the same thing.

sobychacko:
Thanks. But, since we are explicitly dealing with a batch API here, producing the output in the raw format might be beneficial without that flatmap operation.

marcusvoltolim:
> flatmap operatio



> Thanks. But, since we are explicitly dealing with a batch API here, producing the output in the raw format might be beneficial without that flatmap operation.

I agree, and removed this concatMap.

sobychacko:
Any chance you can verify the new API via a unit or integration test?

marcusvoltolim:
> Any chance you can verify the new API via a unit or integration test?

I try make this test!

sobychacko:
@marcusvoltolim Any updates? We have a release next week, and if you want this to be part of that release, then this needs to be addressed sooner so we can progress on the PR.

PR #3066, opened 2024-02-23
GH-2775: Deprecate ContainerProperties properties transactionManager

### LINKED ISSUE #2775: Offset is committed after the rollback of the DB transaction

Tested on  v3.0.8, previous versions are affected as well

**Describe the bug**

We are using `JpaTransactionManager` in `ConcurrentKafkaListenerContainerFactory` to start new DB transactions in consumers.
We've noticed that `spring-kafka` commits offset despite the error during the commit of the DB transaction.
This behavior is unexpected because an error has occurred and retries have been conducted as well.
Therefore, we are expecting that the offset will not be committed and it should work in the same way as when the consumer throws an exception.

Retries are working fine 

**To Reproduce**

Create a container factory and use `JpaTransactionManager` there
```
ConcurrentKafkaListenerContainerFactory<Object, Object> factory =

factory.getContainerProperties().setTransactionManager(jpaTransactionManager);
```

and consumer
```
        // to propagate transaction from spring-kafka to be able to rollback it
       @Transactional
       @KafkaListener(topics = ROLLING_BACK_TRANSACTIONAL_TOPIC)
        public void rollingBackTransactionalReceiver(SampleMessage message) {
           // the data will not be saved into DB
            persist(message);
            TransactionAspectSupport.currentTransactionStatus().setRollbackOnly();
        }

```

**Expected behavior**
Offset is not committed into the consumer group because of an error during the commit of the DB transaction

Wzy19930507:
Use `KafkaAwareTransactionManager` to properly manage Kafka transactions.

* Deprecated properties transactionManager and add kafkaAwareTransactionManager at ContainerProperties.
* Modify transaction unit test to `setKafkaAwareTransactionManager.

Resolves #2775

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/container-props.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/container-props.adoc
@@ -123,6 +123,10 @@ Also see `idleBeforeDataMultiplier`.
 |None
 |Used to override any arbitrary consumer properties configured on the consumer factory.
 
+|[[kafkaAwareTransactionManager]]<<kafkaAwareTransactionManager,`kafkaAwareTransactionManager`>>
+|`null`
+|See xref:kafka/transactions.adoc[Transactions].
+
 |[[listenerTaskExecutor]]<<listenerTaskExecutor,`listenerTaskExecutor`>>
 |`SimpleAsyncTaskExecutor`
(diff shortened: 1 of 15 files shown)


sobychacko:
This PR is merged via `https://github.com/spring-projects/spring-kafka/commit/cd4341ce89bfe794c9fd39c34e8fad0b80d4c8e2`.

Issue #2934, opened 2023-12-13
Is Non Blocking Retry (@Retryable) intended to be combined with Kafka Transactions (consumer initiated) ?


PaulFert:
### Discussed in https://github.com/spring-projects/spring-kafka/discussions/2918

<div type='discussions-op-text'>

<sup>Originally posted by **PaulFert** December  1, 2023</sup>
Hello,

I wanted to use both features for consume -> process -> produce applications

When i use Transaction with Blocking Retry, DLT Recover the flow is : 
![Kafka-Error-Handling-Transaction-Blocking-Retry-Transaction drawio](https://github.com/spring-projects/spring-kafka/assets/5616739/906a9214-ac09-45b8-a1cd-e725250f6a13)

When exception is raised from the listener, the transaction is rollback and then a new transaction is created for the send to DLT, OK fine.

On the oher side, when i use Transaction with Non Blocking Retry the flow is more like that : 
![Kafka-Error-Handling-Transaction-Non-Blocking-Retry-Transaction drawio](https://github.com/spring-projects/spring-kafka/assets/5616739/1fb46c39-aca3-4849-94ec-13baa905bd31)

When exception is raised from the listener, *the transaction is not rollback*, the message is send to retry topic and finally the transaction is comited. Then consumer offsets is commit *plus the messages out* and messages to retry topic. If an exception is raised i want my messages out to be not comited.

It's not what i expected, i expected a flow like that :
![Kafka-Error-Handling-Transaction-Non-Blocking-Retry-Transaction-expected drawio](https://github.com/spring-projects/spring-kafka/assets/5616739/6c22272a-bf35-486c-be09-f7c6576b336e)

Did i miss something with the configuration ? I use standard configuration, with one kafkatemplate, with transactional producer
Or maybe did i misunderstood the Non Blocking Retry pattern,  Transactions are not intended to be combined with ? 

Sorry for my poor english...

Testing with 'org.springframework.boot' version '3.1.1-SNAPSHOT', spring-kafka version 3.0.8

Thanks
</div>

Wzy19930507:
Use `@RetryTopic`, `DeadLetterPublishingRecoverer` will inject `CommonErrorHandler.
`RuntimeException` catch by `CommonErrorHandler`  in transaction, but not throw RuntimeException again.

1. exception catch by `CommonErrorHandler`
[...]

PR #3072, opened 2024-02-26
Non-Blocking Retries cannot combine with container transactions.

Wzy19930507:
Retry topic will break container txns atomicity

1. container txns commit success.
2. Message send to retryable topic retry again.

* Add adoc to notice.

See #2934

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
@@ -79,6 +79,9 @@ Instead, use a `KafkaTransactionManager` in the container to start the Kafka tra
 
 See xref:tips.adoc#ex-jdbc-sync[Examples of Kafka Transactions with Other Transaction Managers] for an example application that chains JDBC and Kafka transactions.
 
+IMPORTANT: xref:retrytopic.adoc[Non-Blocking Retries] cannot combine with xref:kafka/transactions.adoc#container-transaction-manager[Container Transactions].
+When listener code threw exception, container transactions commit success and record sending to the retryable topic.
+
 [[kafkatemplate-local-transactions]]
 == `KafkaTemplate` Local Transactions
 
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
@@ -80,7 +80,7 @@ Instead, use a `KafkaTransactionManager` in the container to start the Kafka tra
 See xref:tips.adoc#ex-jdbc-sync[Examples of Kafka Transactions with Other Transaction Managers] for an example application that chains JDBC and Kafka transactions.
 
 IMPORTANT: xref:retrytopic.adoc[Non-Blocking Retries] cannot combine with xref:kafka/transactions.adoc#container-transaction-manager[Container Transactions].
-When listener code threw exception, container transactions commit success and record sending to the retryable topic.
+When the listener code throws an exception, container transaction commit succeeds, and the record is sent to the retryable topic.
 
 [[kafkatemplate-local-transactions]]
 == `KafkaTemplate` Local Transactions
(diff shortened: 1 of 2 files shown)


sobychacko:
@ Wzy19930507 We need some more due diligence analysis on this issue. What you are saying is that retryable topics and nested transactions are not supported. Maybe the minimum we need to do is add a note in the doc saying that nested txns are not supported with retryable topics. I don't think we need to add the tests and the detailed docs you have in this PR. 

We can probably take your test as a POC for analyzing the issue (at a later time). Could you re-work the PR to only include the note in the docs?

Wzy19930507:
@sobychacko Thanks for your quick reply. I will re-work this PR.
Sorry for my an ambiguous description, let me see if I can describe the problem more precisely.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
```diff
@@ -79,6 +79,9 @@ Instead, use a `KafkaTransactionManager` in the container to start the Kafka tra
 
 See xref:tips.adoc#ex-jdbc-sync[Examples of Kafka Transactions with Other Transaction Managers] for an example application that chains JDBC and Kafka transactions.
 
+IMPORTANT: Container transactions use xref:retrytopic.adoc[Non-Blocking Retries] will break transaction atomicity.
+When listener code threw exception, container transactions commit success and record sending to the retryable topic.
```
artembilan, line 83: Isn't that what expected from asynchronous, non-blocking retries?
To be able to send to the retry topic we indeed has to commit the current transaction.
I'm not fully sure how this non-blocking retires work internally, but apparently we don't start a new transaction, when we consume from the next retry topic.

It might work some way with a `@Transactional` on the `@KafkaListener`, but pay attention that `@EnableTransactionManagement` has to be activated.

Either way: I would not say in the docs `break`.
It sounds like a bug which is not supposed to be there.

It is just a feature which cannot be combined with transactions due to async nature.

Do you see things differently?

Thanks

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
Wzy19930507: Totally agree with you, i'll re-test the `@Transactional` case.

Wzy19930507: Test result is non-blocking retries cannot be combined with transactions.

---

`@Transactional` case as follows:
When `PROPAGATION_REQUIRED` will infinite retry because of container transactions rollback. (un-expected).
When `PROPAGATION_REQUIRED_NEW`
1. If listener code fail and container code success, listener code rollback, container code commit success. (expected)
2. If listener code fail and container code fail, listener code rollback, container code rollback. (expected)
3. If listener code success and container code fail, listener code commit success, container code rollback then infinite retry. (un-expected)
Current `KafkaTransactionManager` only support `PROPAGATION_REQUIRES_NEW` and `PROPAGATION_REQUIRED`, the above case tests only these two cases.

PROPAGATION_REQUIRED_NEW un-expected Report see https://github.com/Wzy19930507/spring-kafka/tree/report_pr_3072

Please see if there are any omissions, thanks.

---

I'll modify docs to **non-blocking retries cannot be combined with transactions** first and see if there's a way to implement it.

sobychacko: @Wzy19930507 We might want to use present tense here: `When the listener code throws an exception, container transaction commit succeeds, and the record is sent to the retryable topic.`

sobychacko (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/transactions.adoc
Wzy19930507: Thanks for the correction, fix it.

PR #3080, opened 2024-02-28
GH-3049: Listener exceptions are not saved to the observation.

### LINKED ISSUE #3049: Listener exceptions not saved to the observation

- `org.springframework.kafka:spring-kafka:3.1.1`
- `org.apache.kafka:kafka-clients:3.6.1`

**Describe the bug**

When an exception is thrown from a listener, by default the exception is not saved in the current listener observation.

https://github.com/spring-projects/spring-kafka/blob/22764cbcdf045def5a8688f2aecfecfc78c1e614/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2759-L2794

`Observation#observe` would do it on its own, but then the provided supplier has to re-throw the exception rather than just return it, which it probably does for `ListenerConsumer#invokeInTransaction`.

**To Reproduce**

1. Make & register a message listener.
2. Send a message to the listener topic. (For example, `java.lang.NullPointerException`.)
3. Throw an exception from the listener.

**Expected behavior**

The error handler is invoked and all, but on top of that, the (original) exception is embedded into the observation, allowing downstream tracing code to handle it.

**Sample**

For simplicity's sake, I will only leave the listener code here. In the configuration, there is only a consumer group identifier. Let me know if I should make this into a repository.

```java
@KafkaListener(topics = "error")
public void onErrorMessage(@Payload String payload) throws Throwable {
    Class<?> errorClass = Class.forName(payload);
    throw (Throwable) errorClass.getDeclaredConstructor().newInstance();
}
```

Wzy19930507:
* Embedded the (original) exception into the observation, allowing downstream tracing code to handle it.
* Add unit test for observation Error and RuntimeException.

Resolves #3049

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -2762,37 +2762,37 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 					DefaultKafkaListenerObservationConvention.INSTANCE,
 					() -> new KafkaRecordReceiverContext(cRecord, getListenerId(), this::clusterId),
 					this.observationRegistry);
-			return observation.observe(() -> {
-				try {
+			try {
+				observation.observe(() -> {
 					invokeOnMessage(cRecord);
 					successTimer(sample, cRecord);
 					recordInterceptAfter(cRecord, null);
+				});
+			}
+			catch (RuntimeException e) {
+				failureTimer(sample, cRecord);
+				recordInterceptAfter(cRecord, e);
+				if (this.commonErrorHandler == null) {
+					throw e;
 				}
-				catch (RuntimeException e) {
-					failureTimer(sample, cRecord);
-					recordInterceptAfter(cRecord, e);
-					if (this.commonErrorHandler == null) {
-						throw e;
-					}
-					try {
-						invokeErrorHandler(cRecord, iterator, e);
-						commitOffsetsIfNeededAfterHandlingError(cRecord);
-					}
-					catch (KafkaException ke) {
-						ke.selfLog(ERROR_HANDLER_THREW_AN_EXCEPTION, this.logger);
-						return ke;
-					}
-					catch (RuntimeException ee) {
-						this.logger.error(ee, ERROR_HANDLER_THREW_AN_EXCEPTION);
-						return ee;
-					}
-					catch (Error er) { // NOSONAR
-						this.logger.error(er, "Error handler threw an error");
-						throw er;
-					}
+				try {
+					invokeErrorHandler(cRecord, iterator, e);
+					commitOffsetsIfNeededAfterHandlingError(cRecord);
 				}
-				return null;
-			});
+				catch (KafkaException ke) {
+					ke.selfLog(ERROR_HANDLER_THREW_AN_EXCEPTION, this.logger);
+					return ke;
+				}
+				catch (RuntimeException ee) {
+					this.logger.error(ee, ERROR_HANDLER_THREW_AN_EXCEPTION);
+					return ee;
+				}
+				catch (Error er) { // NOSONAR
+					this.logger.error(er, "Error handler threw an error");
+					throw er;
+				}
+			}
+			return null;
 		}
 
 		private void commitOffsetsIfNeededAfterHandlingError(final ConsumerRecord<K, V> cRecord) {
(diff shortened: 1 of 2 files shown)


Wzy19930507:
Unit text `TransactionalContainerTests.testBatchListenerRecoverAfterRollbackProcessorCrash` throw exception see https://github.com/spring-projects/spring-kafka/pull/3081.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2772,37 +2773,59 @@ private RuntimeException doInvokeRecordListener(final ConsumerRecord<K, V> cReco
 					DefaultKafkaListenerObservationConvention.INSTANCE,
 					() -> new KafkaRecordReceiverContext(cRecord, getListenerId(), this::clusterId),
 					this.observationRegistry);
-			return observation.observe(() -> {
+			RuntimeException runtimeException;
+			observation.start();
```
artembilan, line 2777: I don't think this is what we need here.
There is just enough to pull `try..catch` from that supplier in the `observe()` around this `observe()`.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
Wzy19930507: Fix it, hope i get your point, and the code is much cleaner.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2780,10 +2780,12 @@ private RuntimeException doInvokeRecordListener(final ConsumerRecord<K, V> cReco
 					}
 					catch (KafkaException ke) {
 						ke.selfLog(ERROR_HANDLER_THREW_AN_EXCEPTION, this.logger);
+						observation.error(ke);
 						return ke;
 					}
 					catch (RuntimeException ee) {
 						this.logger.error(ee, ERROR_HANDLER_THREW_AN_EXCEPTION);
+						observation.error(ee);
```
artembilan, line 2788: Thanks, but I meant to do this:
```
			try {
				observation.observe(() -> {
					invokeOnMessage(cRecord);
					successTimer(sample, cRecord);
					recordInterceptAfter(cRecord, null);
				});
			}
			catch (RuntimeException e) {
				failureTimer(sample, cRecord);
				recordInterceptAfter(cRecord, e);
					throw e;
				}
				try {
					invokeErrorHandler(cRecord, iterator, e);
					commitOffsetsIfNeededAfterHandlingError(cRecord);
				}
				catch (KafkaException ke) {
					ke.selfLog(ERROR_HANDLER_THREW_AN_EXCEPTION, this.logger);
					return ke;
				}
				catch (RuntimeException ee) {
					this.logger.error(ee, ERROR_HANDLER_THREW_AN_EXCEPTION);
					return ee;
				}
				catch (Error er) { // NOSONAR
					this.logger.error(er, "Error handler threw an error");
					throw er;
				}
			}
			return null;
```

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
Wzy19930507: Thanks, i'd like it. much much better.

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/ObservationTests.java
```diff
@@ -276,6 +325,20 @@ KafkaTemplate<Integer, String> customTemplate(ProducerFactory<Integer, String> p
...
+			KafkaTemplate<Integer, String> template = new KafkaTemplate<>(pf);
+			template.setObservationEnabled(true);
+			return template;
+		}
+
+		@Bean
+		KafkaTemplate<Integer, String> errorTemplate(ProducerFactory<Integer, String> pf) {
+			KafkaTemplate<Integer, String> template = new KafkaTemplate<>(pf);
+			template.setObservationEnabled(true);
+			return template;
```
artembilan, line 339: Looks like both of these new `KafkaTemplate` beans are just exactly the same what is `template` bean.

Any particular reason why do we need these two?
Why that `template` is not enough for new tests?

Thanks

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/ObservationTests.java
Wzy19930507: Just for test tag `spring.kafka.template.name`, is doesn't matter.
Should I keep it or remove?

Issue #3076, opened 2024-02-27
`KafkaMessageListenerContainer` doesn't properly clear threadState of `AfterRollbackProcessor`


rroesch1:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.1.1

**Describe the bug**

The `KafkaMessageListenerContainer` doesn't clear the threadState of `AfterRollbackProcessor` after a successfull processing attempt if `transactionManager` is set.

**To Reproduce**

* I have a `KafkaMessageListenerContainer` configured with a `BatchMessageListener` and `KafkaTransactionManager` (batched exactly-once processing)
* Additionally i set a `DefaultAfterRollbackProcessor` with `ExponentialBackOff`
* i **do not** set any CommonErrorHandler (`AbstractMessageListenerContainer#setCommonErrorHandler`)

Example snippet:
```java
    @Bean
    MessageListenerContainer messageListenerContainer(
            KafkaTemplate<byte[], byte[]> kafkaTemplate,
            ConsumerFactory<byte[], byte[]> consumerFactory,
            KafkaTransactionManager<byte[], byte[]> transactionManager) {

        // initialize MyEventListener which is a BatchMessageListener
        var myEventListener = new MyEventListener(kafkaTemplate);
        // initialize container properties
        var props = new ContainerProperties(sourceTopics);
        props.setMessageListener(myEventListener);
        props.setTransactionManager(transactionManager); // enable transaction manager for exactly-once processing
        // initialize back-off strategy for retries
        var backOff = new ExponentialBackOff(1000, 2);
        backOff.setMaxInterval(300000L); // set back-off limit to 5min
        var afterRollbackProcessor = new DefaultAfterRollbackProcessor<>(backOff);
        // initialize MessageListenerContainer
        var container = new KafkaMessageListenerContainer<>(consumerFactory, props);
        container.setAfterRollbackProcessor(afterRollbackProcessor);
        return container;
    }
```

**(Un-)expected behavior**

* `KafkaMessageListenerContainer` calls the `AfterRollbackProcessor` after each failed transacation (as expected)
* the `DefaultAfterRollbackProcessor` uses the configured `BackOff` after each failed transacttion (as expected)
* after a sucessfull processing attempt, `AfterRollbackProcessor#clearThreadState` is not being called (that's **unexpected** to me)
[...]

PR #3082, opened 2024-02-28
GH-3076: Set `batchFailed` state in `ListenerConsumer` even if no `commonErrorHandler` is configured

rroesch1:
This PR is naïve approach to solve #3076 (i am not sure about possible side effects/consequences of this change)


rroesch1:
@pivotal-cla This is an Obvious Fix

artembilan (changes requested):
The fix looks good!

Any chances we can have a unit test to cover such a situation?

And from there the fix won't be so `obvious`, so consider to sign the CLA, please.

rroesch1:
* Sorry for the late feedback but i had to clarify the CLA requirement with my employer.  My contributions should now be covered by a Corporate CLA.
* I added a unit test which should cover the situation described in #3076 (thus it's probably no longer an obvious fix)

artembilan (changes requested):
Please, add your name to the `@author` list of the affected classes.

rroesch1:
> Please, add your name to the `@author` list of the affected classes.

done

Issue #2287, opened 2022-05-31
RecordInterceptor lifecycle is wrong if record filtered


nikoncode:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**
2.8.6

**Describe the bug**

We are using RecordInterceptor to the instrument and measure Kafka handlers. If you are skipping record (by returning null from intercept) rest lifecycle methods `success/failure/afterRecord` will never be called.

**To Reproduce**

Create an always skipping record interceptor.

**Expected behavior**

I expect the same lifecycle for all cases: intercept + (failure OR success) + afterRecord.

nikoncode:
It might be related to earlyInterceptionLogic

nikoncode:
Lifecycle without early interceptor:
```
invoke(record) (returns null)
success(record) (not null record, even if invoke returned null)
afterRecord(record) (not null record, even if invoke returned null)
```


Lifecycle with early interceptor:
```
invoke(record) (returns null)
```

nikoncode:
There is one more problem related to this.

Let's imagine you have 2 interceptors returning always null combined using CompositeRecordInterceptor.

`intercept` method will handle only the first one, but other lifecycle methods `success/fail/afterRecord` will be called for both. It might be harmful if you rely on lifecycle (use thread local to measure execution time for example).

PR #3090, opened 2024-03-01
Polish KafkaMessageListenerContainer

Wzy19930507:
* remove `checkAckMode` because `ContainerProperties.ackCount` and `ContainerProperties.ackTime` have init value.
* change `ListenerConsumer.offsets` type to Map<TopicPartition, Long>.
* replace `ListenerConsumer.setupSubBatchPerPartition` to `ContainerProperties.isSubBatchPerPartition`.
* add method `ListenerConsumer.commitOffsets` and `ListenerConsumer.commitOffsetsInTransactions`
* remove properties `ListenerConsumer.this.commitRecovered`.
* refactor method `ListenerConsumer.processCommits`.
* refactor method  `ListenerConsumer.ackCurrent`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -173,8 +173,6 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 
 	private static final String RAWTYPES = "rawtypes";
 
-	private static final int DEFAULT_ACK_TIME = 5000;
-
 	private static final Map<String, Object> CONSUMER_CONFIG_DEFAULTS = ConsumerConfig.configDef().defaultValues();
 
 	private final AbstractMessageListenerContainer<K, V> thisOrParentContainer;
@@ -277,13 +275,8 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			else if (partitionsListenerConsumer.assignedPartitions != null) {
 				return Collections.unmodifiableCollection(partitionsListenerConsumer.assignedPartitions);
 			}
-			else {
-				return null;
-			}
 		}
-		else {
-			return null;
-		}
+		return null;
 	}
 
 	@Override
@@ -293,9 +286,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 		if (partitionsListenerConsumer != null) {
 			return Collections.singletonMap(partitionsListenerConsumer.getClientId(), getAssignedPartitions());
 		}
-		else {
-			return null;
-		}
+		return null;
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


sobychacko:
@Wzy19930507 There are test failures in the PR: https://github.com/spring-projects/spring-kafka/actions/runs/8115263868/job/22182761295?pr=3090. Could you take a look?

Wzy19930507:
introduced a unit test bug from https://github.com/spring-projects/spring-kafka/issues/2287, sorry for it.

When `AckMode.RECORD` and early is `false`, the last execute method is `RecordInterceptor.afterRecord`, but i forget a 'CountDownLatch' to wait 'RecordInterceptor.   afterRecord 'completes execution.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3038,45 +2954,44 @@ private void doSendOffsets(Producer<?, ?> prod, Map<TopicPartition, OffsetAndMet
 		private void processCommits() {
 			this.count += this.acks.size();
 			handleAcks();
-			AckMode ackMode = this.containerProperties.getAckMode();
-			if (!this.isManualImmediateAck) {
-				if (!this.isManualAck) {
-					updatePendingOffsets();
```
sobychacko, line 3044: Where did this code (`updatePendingOffsets`) move to? Do we not need this any longer? I see that you removed the method, but where is this logic now?

Wzy19930507: `updatePendingOffsets()` feel like duplicate in `processAck()`

https://github.com/spring-projects/spring-kafka/blob/5beb8fe607bd131ced632dbc7afa32e11a6ff2dd/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2024-L2034

sobychacko: But `processAck()` is not called from the hierarchy of `processCommits` where `updatePendingOffsets()` was originally called. I see that there are some duplicates, but still wondering whether it changes the logic a little bit?

sobychacko: @Wzy19930507 Any updates on this? This is the only pending item that needs an explanation before merging the PR.

Wzy19930507: Sorry for late.  `handleAcks` has three scenarios and include `updatePendingOffsets` logic

1. `!Thread.currentThread().equals(this.consumerThread)` --> keep `isManualImmediateAck` thread safe.
2. `isManualImmediateAck` --> call `ackImmediate`
3. not `isManualImmediateAck` --> call `addOffset`

`updatePendingOffsets` logic would be:
```
			if (!this.isManualImmediateAck) {
				if (!this.isManualAck) {
					updatePendingOffsets();
				}
```
Look like same as scenarios three.

Wzy19930507: Yes, Yes, totally right, your explanation is clearer!

Issue #3067, opened 2024-02-23
Spring Kafka does not allow multiple headers with same key


poznachowski:
I'm using Spring Kafka 3.1.2

Apache Kafka allows to have multiple headers with same key, but Spring Kafka limits that.

Sample project reproducting the issue is available here:
https://github.com/poznachowski/sample-spring-kafka-headers

When I use Listener with `ConsumerRecord` directly as payload I can see all the headers:

`ConsumerRecord payload: SomePayload[value=payload], headers: RecordHeaders(headers = [RecordHeader(key = testHeader, value = [118, 97, 108, 117, 101, 49]), RecordHeader(key = testHeader, value = [118, 97, 108, 117, 101, 50])], isReadOnly = false)`

But I'd like to use much more convenient Spring abstractions such as: `@Payload` and `@Headers`, but with this approach I'm getting only a single `testHeader` value:

`headers: {kafka_offset=11, testHeader=[B@67388d79, ...`

The reason for that is that `KafkaHeaderMapper` implementations uses `Map.put` method that effective replaces value with the latest.

sobychacko:
@poznachowski If we change this outright so that multiple values are allowed, it may break existing consumer applications relying on a single value for the header. For example, suppose a producer application sends two values for a key. In that case, a consumer application built with a prior version of Spring Kafka may stop working or result in incorrect behavior. Even if we introduce some special headers on the producer side to opt-in for this, how do the existing consumer apps know about it? It feels like we need to think through the consequences before implementing this feature.

artembilan:
Well, that is a `DefaultKafkaHeaderMapper` behavior.
Feel free to implement any custom one which would populate a list of values into the target message header.

I don't treat this is a bug since we never advertised such a behavior from the `DefaultKafkaHeaderMapper`.

We may revise the default logic into something similar what we have in the `DefaultHttpHeaderMapper` in Spring Integration.

Kafka record has always had a multi-entry capabilities...

sobychacko:
We had an internal discussion on this matter and this is something we can adjust in the `3.2.0` version. Any existing consumers who are expecting a single value for a certain header need to migrate (adjust) the code. You are welcome to send a PR. Please let us know if you are willing (or the time) to do it.

poznachowski:
Thank you for quick response. Everything makes sense. I'm willing to help with a PR.

sobychacko:
[...]

PR #3101, opened 2024-03-05
GH-3067: Draft of mapping multiple headers with same key

poznachowski:
Draft of mapping iterable headers

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
@@ -19,6 +19,7 @@ package org.springframework.kafka.support;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
@@ -26,6 +27,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
@@ -33,6 +35,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.header.internals.RecordHeader;
+import org.assertj.core.util.Streams;
 
 import org.springframework.messaging.MessageHeaders;
 import org.springframework.util.Assert;
@@ -48,11 +51,13 @@ import org.springframework.util.ClassUtils;
  * @author Gary Russell
  * @author Artem Bilan
  * @author Soby Chacko
+ * @author Grzegorz Poznachowski
  *
  * @since 1.3
- *
  */
 public class DefaultKafkaHeaderMapper extends AbstractKafkaHeaderMapper {
+
+	private static final String ITERABLE_HEADER_TYPE_PATTERN = "%s#%s";
 
 	private static final String JAVA_LANG_STRING = "java.lang.String";
 
@@ -96,6 +101,7 @@ public class DefaultKafkaHeaderMapper extends AbstractKafkaHeaderMapper {
 	 * {@code "!id", "!timestamp" and "*"}. In addition, most of the headers in
 	 * {@link KafkaHeaders} are never mapped as headers since they represent data in
 	 * consumer/producer records.
+	 *
 	 * @see #DefaultKafkaHeaderMapper(ObjectMapper)
 	 */
 	public DefaultKafkaHeaderMapper() {
@@ -110,6 +116,7 @@ public class DefaultKafkaHeaderMapper extends AbstractKafkaHeaderMapper {
 	 * {@code "!id", "!timestamp" and "*"}. In addition, most of the headers in
 	 * {@link KafkaHeaders} are never mapped as headers since they represent data in
 	 * consumer/producer records.
+	 *
 	 * @param objectMapper the object mapper.
 	 * @see org.springframework.util.PatternMatchUtils#simpleMatch(String, String)
 	 */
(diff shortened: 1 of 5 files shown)


artembilan (changes requested):
I'm OK with the solution.
Please, consider to run `gradlew check` locally, so you will see Checkstyle violations.
Thanks

artembilan (changes requested):
Please, also add your name to the `@author` list of the affected classes.
The `whats-new.adoc` deserves a new entry for this new feature.
And `headers.adoc` must be improved about this new behavior.

I also see that you did nothing for the `DefaultKafkaHeaderMapper`.
Probably the logic you have introduced can go into the `AbstractKafkaHeaderMapper` ?

poznachowski:
Yes of course. This was just a starter to make sure I'm on the right track.

sobychacko:
@poznachowski, Any updates on this? We have a release coming up next week. If you want this to be part of that release, we need to start further review on this PR sooner.

poznachowski:
I should have something ready tomorrow.

poznachowski:
I updated my MR with initial version (not polished) of `DefaultHeaderKafkaMapper`. It's more complicated than I thought.
The need of handling `spring_json_header_types` header makes it harder to extract common logic to the `AbstractKafkaHeaderMapper`.
I'm not really sure if my solution will hold. I see that `RecordHeaders` uses `List` internally, but is it enough to guarantee that Kafka ensures headers order?
Additionally, the change makes `DeliveryHeaderTests` fail. `retry_topic-attempts` header becomes now a list with 2 entries and `getNonBlockingRetryDeliveryAttempt` method fails. Haven't investigated it yet.

sobychacko:
@poznachowski We moved the milestone to `3.2.0-RC1`, which is the April release, so you have some time to polish things. Please let us know when it is ready for the actual review. The approach that you take in the PR is in the right direction. Could you polish and make sure that there are no checkstyle issues, etc.? Then we will do more formal reviews.

poznachowski:
@sobychacko sure, should have it ready somewhere beginning next week.

sobychacko:
@poznachowski Any updates on this PR?

poznachowski:
I'd like to say that I have all ready, but I found some Spring specific header inconsistencies and struggling with picking the right approach.

`retainAllRetryHeaderValues` flag in `DeadLetterPublishingRecovererFactory` allows to have multiple headers on Kafka-side for given header types (especially `RetryTopicHeaders.DEFAULT_HEADER_ATTEMPTS`). When these were mapped onto Spring `Message` only the last one was always retained allowing to simply use `@Header(name = RetryTopicHeaders.DEFAULT_HEADER_ATTEMPTS, required = false) Integer nonBlockingAttempts` (as in `DeliveryHeaderTests`). As I was unware of usage of such headers, the current state of my MR allows to map these to a potential List on Spring side.
This has developed into another problem. Spring was still able to map List of values onto a single element with usage of `CollectionToObjectConverter`. The converter is taking first element thought. To overcome this and allow potential backwards compatibility I decided to reverse the list order while mapping.
Current solution is also inconsistent as I decided to keep only last values for headers such as: `KafkaHeaders.DELIVERY_ATTEMPT, KafkaHeaders.LISTENER_INFO, KafkaUtils.KEY_DESERIALIZER_EXCEPTION_HEADER and KafkaUtils.VALUE_DESERIALIZER_EXCEPTION_HEADER` as they where already specified in the `DefaultKafkaHeaderMapper` mapper. When I tried to allow multiple values for `KafkaHeader.DELIVERY_ATTEMPT` multiple test issues arose.

Having said that I'd suggest to keep a list of Spring 'technical' Kafka headers that should never be mapped into a collection on the Spring side (only last values should be kept). This would avoid unnecessary changes in the code (especially in the retrytopic package). If this plus reversing list order (so `CollectionToObjectConverter` can pick up the last value for backwards compatibility) sounds fine to you I can proceed with polishing and providing appropriate documentation.

sobychacko:
@poznachowski Sorry for the delay in responding to you. The approach you mentioned for keeping the internal metadata type headers as only single-value item makes sense. As long as it doesn't bring any backward compatibility issues, it's ok to reverse the order. Please re-work the PR and when ready, remove the draft status, so that we can start a formal PR review. Thanks!

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -97,6 +102,7 @@ public class DefaultKafkaHeaderMapper extends AbstractKafkaHeaderMapper {
 	 * {@code "!id", "!timestamp" and "*"}. In addition, most of the headers in
 	 * {@link KafkaHeaders} are never mapped as headers since they represent data in
 	 * consumer/producer records.
+	 *
```
artembilan, line 105: Please, consider to not reformat Javadocs: no empty lines for methods.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -304,34 +304,84 @@ public void fromHeaders(MessageHeaders headers, Headers target) {
...
+					encodeToJson = true;
+				}
+				if (!encodeToJson && valueToAdd instanceof String stringValue) {
+					target.add(new RecordHeader(headerName, stringValue.getBytes(getCharset())));
 				}
 				else {
-					headers.put(headerName, headerValueToAddIn(header));
+					target.add(new RecordHeader(headerName, this.objectMapper.writeValueAsBytes(valueToAdd)));
 				}
+				jsonHeaders.put(headerIndex == null ?
```
artembilan, line 327: I'm not sure what you are doing here, but I think all the framework headers must be as a single value Kafka headers anyway. There are a lot of logic in the framework which relies on them.
We just must agree that we have to make a fix only for end-user headers which can be iterable.
Even if that JSON headers looks like a collection, it is really something what the framework rely on as as single header.
So, I don't expect too drastic changes in this area.
Thanks.

artembilan (changes requested):

sobychacko:
@poznachowski Are you going to continue working on this PR? We can also consider taking it on our end and start making additional changes on top of your PR. Please let us know the status.

sobychacko:
@poznachowski, I just wanted to see if there is an update on this PR. If you are busy, we can take this PR and work on it on our end. Please let us know. Thank you.

sobychacko:
@poznachowski, Any updates on this PR? We are still in the milestone stages of the `3.3.0` release, and if you update the PR, we can potentially include this feature in the RC1 timeline.

artembilan:
Superseded by https://github.com/spring-projects/spring-kafka/pull/3874.

PR #3105, opened 2024-03-06
GH-3012: Non-blocking retries support @KafkaListener on class level

Wzy19930507:
Fixes: #3012

* Non-blocking retries support `@KafkaListener` on class level, include `@RetryableTopic`, `@DltHandler`, `RetryTopicConfiguration`, `RetryTopicConfigurationSupport`.
* Doc and test Non-blocking retries support `@KafkaListener` on class level.

---

Part 1:  Support process annotation `@RetryableTopic` and `@DltHandler` when `@RetryableTopic` annotated on class, see #3107.
Part 2:  `RetryTopicConfigurer#processAndRegisterEndpoint` support `MultiMethodKafkaListenerEndpoint`, see #3112.
Part 3:  `@EnableKafka` integration feature that non-blocking retries support `@KafkaListener` on class level,  see #3105.

PR #3107, opened 2024-03-08
@RetryableTopic support @KafkaListener annotated on class part 1

Wzy19930507:
* `@RetryableTopic` support annotated on Class.
* Support process `@RetryableTopic` from Class in `RetryTopicConfigurationProvider`.
* Support process `@DltHandler` when `@RetryableTopic` annotated on the Class in `RetryableTopicAnnotationProcessor`.

part of #3105

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryTopicConfigurationProvider.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryTopicConfigurationProvider.java
@@ -106,6 +106,14 @@ public class RetryTopicConfigurationProvider {
 		return findRetryConfigurationFor(topics, method, null, bean);
 	}
 
+	/**
+	 * Find retry topic configuration.
+	 * @param topics the retryable topic list.
+	 * @param method the method that gets @RetryableTopic annotation.
+	 * @param clazz the class that gets @RetryableTopic annotation.
+	 * @param bean the bean.
+	 * @return the retry topic configuration.
+	 */
 	@Nullable
 	public RetryTopicConfiguration findRetryConfigurationFor(String[] topics, @Nullable Method method,
 			@Nullable Class<?> clazz, Object bean) {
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryTopicConfigurationProvider.java
```diff
@@ -96,19 +100,34 @@ public RetryTopicConfigurationProvider(@Nullable BeanFactory beanFactory, @Nulla
...
-		RetryableTopic annotation = MergedAnnotations.from(method, SearchStrategy.TYPE_HIERARCHY,
-					RepeatableContainers.none())
-				.get(RetryableTopic.class)
-				.synthesize(MergedAnnotation::isPresent)
-				.orElse(null);
+		return findRetryConfigurationFor(topics, method, null, bean);
+	}
+
+	@Nullable
+	public RetryTopicConfiguration findRetryConfigurationFor(String[] topics, @Nullable Method method,
```
sobychacko, line 110: Why does it need to be `public`? If it's valid, add the `since` tag and a note in `whats-new` as this is a new public API.

sobychacko (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryTopicConfigurationProvider.java
Wzy19930507: > note in whats-new as this is a new public API.

It doesn't seem useful for a user to know how to get `RetryTopicConfiguration`. Unless we write a separate section that describes how to manually register main and retry endpoints using the `RetryTopicConfigurer`.

Wzy19930507: Fix it.

sobychacko (approved):

Issue #3108, opened 2024-03-08
Unsafe access to Kafka consumer in seek operation


sobychacko:
When the seek API that uses the user-provided function to compute the offset to seek, there is a concurrency issue in which the Kafka consumer is used unsafely. See [this](https://github.com/spring-projects/spring-kafka/issues/3078#issuecomment-1983759811) for more details.

Issue #3078, opened 2024-02-28
Compute New Seek Position From Current Offset


eblocha:
**Expected Behavior**

I have a use-case where I only want to seek to a specified offset if it is before the consumer's current offset. Ideally I would like to have a method on `ConsumerSeekCallback` where I would provide a `Function<Long, Long>` that computes the new offset given the current one.

**Current Behavior**

Currently the framework only provides a way to seek to a specific offset unconditionally, or provide a relative offset from the current one. There is no way to access the current offset value when performing a seek.

**Context**

For my use-case, there are certain events that require re-processing records that were already processed before. However, I should not seek _forward_ in the Kafka partition, because the partition may contain other records that are unrelated to the event and shouldn't be skipped (sorry if this is a bit vague).

So, I need to know what the current partition offset is before I decide to actually seek from the event.

In order to try to accomplish this, I have the Kafka listener extend `AbstractConsumerSeekAware`, then use the methods from that to perform the seek when the event occurs.

I have considered using the `onIdleContainer` method to store the current offsets, but reading the source code it seems that this is only called after a certain time has elapsed, so these offsets are not guaranteed to be up-to-date by the time I decide to seek.

I have also considered using the `KafkaListenerEndpointRegistry`, however this only allows pausing/resuming/stopping containers, and does not provide access to the underlying `Consumer`.

sobychacko:
@eblocha, I want to understand this use case a bit more. The `seekRelative` method allows you to seek relatively from the current offset by setting `toCurrent` to be true. See the javadocs. 

```
/**
 * Perform a seek relative to the start, end, or current position. When called 
 * from {@link ConsumerSeekAware#onPartitionsAssigned(Map, ConsumerSeekCallback)}
 * or from {@link ConsumerSeekAware#onIdleContainer(Map, ConsumerSeekCallback)}
 * perform the seek immediately on the consumer. When called from elsewhere, queue
 * the seek operation. The queued seek will occur after any pending offset
 * commits. The consumer must be currently assigned the specified partition.
 * @param topic the topic.
 * @param partition the partition.
 * @param offset the offset; positive values are relative to the start, negative
 * values are relative to the end, unless toCurrent is true.
 * @param toCurrent true for the offset to be relative to the current position
 * rather than the beginning or end.
 * @since 2.3
 */
void seekRelative(String topic, int partition, long offset, boolean toCurrent);
```

[...]

PR #3109, opened 2024-03-08
GH-3108: Unsafe access to consumer in seek API

sobychacko:
Fixes: #3108

* When the seek API that uses the user-provided function to compute the offset to seek, there is a concurrency issue in which the Kafka consumer is used unsafely.

See this for details: https://github.com/spring-projects/spring-kafka/issues/3078#issuecomment-1983759811


spring-kafka/src/main/java/org/springframework/kafka/support/TopicPartitionOffset.java
```diff
@@ -198,6 +215,10 @@ public SeekPosition getPosition() {
 		return this.position;
 	}
 
+	public Function<Long, Long> getOffsetComputeFunction() {
+		return offsetComputeFunction;
```
artembilan, line 219:
```suggestion
		return this.offsetComputeFunction;
```

Issue #3114, opened 2024-03-11
Failing to handle deserialization exceptions in batch listener


ikercrg:
I am migrating Spring boot from version 2 to version 3.

In my former implementation, I was based on [v2.8.4 documentation](https://docs.spring.io/spring-kafka/docs/2.8.4/reference/html/#error-handling-deserializer) and all was working fine.

Currently, ListenerUtils.byteArrayToDeserializationException is removed and I have to use SerializationUtils.byteArrayToDeserializationException.

For this, I am basing on [v3.1.x documentation](https://docs.spring.io/spring-kafka/docs/current/reference/kafka/serdes.html#error-handling-deserializer), but I have problems.

The example given in the documentation, precisely this part, does not compile:

```
DeserializationException deserEx = SerializationUtils.byteArrayToDeserializationException(this.logger,
                    headers.get(i).get(SerializationUtils.VALUE_DESERIALIZER_EXCEPTION_HEADER));
```

Theoretically what is passed is a byte[], but the method only admits a Header. So we would need a cast for that. With ListenerUtils I was casting it to byte[]. But now, I cast to Header or RecordHeader, I get the following error -> Foreign deserialization exception header ignored; possible attack? Something like the following:

```
DeserializationException exception = SerializationUtils.byteArrayToDeserializationException(
            new LogAccessor(IndexingService.class),
            new RecordHeader(SerializationUtils.KEY_DESERIALIZER_EXCEPTION_HEADER,
                (byte[]) headersOfNullItem.get(SerializationUtils.VALUE_DESERIALIZER_EXCEPTION_HEADER)));
```

I have tried to create a DeserializationExceptionHeader, but it is package-protected.

Using 

`List<ConsumerRecord<String, my_object>> my_method()`

instead of what is said in the documentation

`List<Thing> in, @Header(KafkaHeaders.BATCH_CONVERTED_HEADERS) List<Map<String, Object>> headers`

works fine.

Issue opened as petition from @artembilan in Stackoverflow:

See https://stackoverflow.com/questions/78122348/failing-to-handle-deserialization-exceptions-in-batch-listener

sobychacko:
@ikercrg, You are right in that when you consume as `ConsumerRecord`, you can now use the `SerializationUtils#getExceptionHeader` method. 

```
	/**
	 * Extract a {@link DeserializationException} from the supplied header name, if
	 * present.
	 * @param record the consumer record.
	 * @param headerName the header name.
[...]

PR #3115, opened 2024-03-12
GH-3114: DeserializationException propagation

sobychacko:
Fixes: #3114

* Since DeserializationExceptionHeader is currently porpagated as a `byte[]`, it encounters some issues when processing the header especially in batch listeners. Fixing this by providing the deserialization header without `byte[]` conversion
* Adding test to verify
* Refactoring in SerializationTestUtils

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
@@ -31,6 +31,7 @@ import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.header.internals.RecordHeader;
 
+import org.springframework.kafka.support.serializer.SerializationUtils;
 import org.springframework.messaging.MessageHeaders;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
@@ -48,6 +49,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  *
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Soby Chacko
  *
  * @since 1.3
  *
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -31,6 +31,7 @@
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.header.internals.RecordHeader;
 
+import org.springframework.kafka.support.serializer.SerializationUtils;
```
artembilan, line 34: We need to be sure that there is no package tangle . Usually sub-packages uses their parents.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -314,6 +316,11 @@ public void toHeaders(Headers source, final Map<String, Object> headers) {
 			else if (headerName.equals(KafkaHeaders.LISTENER_INFO) && matchesForInbound(headerName)) {
 				headers.put(headerName, new String(header.value(), getCharset()));
 			}
+			else if ((headerName.equals(SerializationUtils.KEY_DESERIALIZER_EXCEPTION_HEADER) ||
+					headerName.equals(SerializationUtils.VALUE_DESERIALIZER_EXCEPTION_HEADER))
+					&& matchesForInbound(headerName)) {
```
artembilan, line 321: I believe this header has to be mapped unconditionally. It is not some what end-user provides.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaUtils.java
```diff
@@ -38,12 +38,32 @@
...
+ * @author Soby Chacko
  *
  * @since 2.2
  *
  */
 public final class KafkaUtils {
 
+	/**
+	 * Header name for deserialization exceptions.
+	 * @since 3.2
```
artembilan, line 50: According to the GH issue mandate this fix is supposed to be back-ported down to `3.0.x`

artembilan (changes requested):

PR #3137, opened 2024-03-15
GH-3135 : Adding new `ListenerContainerRegistry.getListenerContainersMatchingId(Predicate)` for easier usage. 

### LINKED ISSUE #3135: Adding new `KafkaListenerEndpointRegistry.getListenerContainersMatchingId(predicate: Predicate<String>)`
**Expected Behavior**
Would be nice to have method...

```java
kafkaListenerEndpointRegistry
            .getListenerContainersMatchingId( id -> id.startsWith(prefix))
           .forEach(MessageListenerContainer::start)
           
```

**Current Behavior**

```java
kafkaListenerEndpointRegistry
            .getListenerContainerIds()
            .filter( id -> id.startsWith(prefix))
            .forEach(id -> kafkaListenerEndpointRegistry.getListenerContainer(id).start())
```

**Context**

I run application that starts/stops listeners at certain time(office hours)

We have listeners with ids such as...

- `listener_product_purchase`
- `listener_product_purchase-retry-0`
- `listener_product_purchase-retry-1`
- `listener_product_purchase-dlt`

... this, generated by @RetryableTopic configuration.
Just some syntactic sugar.

JooHyukKim:
Resolves #3135

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
@@ -85,10 +85,11 @@ as a default prefix for auto-generated client IDs for certain client types.
 See xref:kafka/connecting.adoc#default-client-id-prefixes[Default client ID prefixes] for more details.
 
 [[get-listener-containers-matching]]
-=== ListenerContainerRegistry's new getListenerContainersMatching(Predicate<String> idMatcher) method
+== Retrieve MessageListenerContainers with by matching a Predicate
 
-Introduces a new method `getListenerContainersMatching` since version 3.2, which returns all `MessageListenerContainer`
-instances whose ID matches a specified (non-null) `Predicate<String>`. This addition provides a flexible way to
-retrieve containers by more complex criteria than just a direct ID match.
+Starting with version 3.2, the `ListenerContainerRegistry` introduces the `getListenerContainersMatching` method.
+This method filters `MessageListenerContainer` instances based on a `Predicate<String>` applied to their IDs.
+It's designed for scenarios requiring more nuanced selection than simple ID matching,
+offering a powerful tool for dynamic container management.
 
-See xref:kafka/receiving-messages/kafkalistener-lifecycle.adoc#get-listener-containers-matching[ `@KafkaListener` Lifecycle Management API Docs] for more details.
+See xref:kafka/receiving-messages/kafkalistener-lifecycle.adoc#get-listener-containers-matching[`@KafkaListener` Lifecycle Management's API Docs] for more information.
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
@@ -87,22 +87,8 @@ See xref:kafka/connecting.adoc#default-client-id-prefixes[Default client ID pref
 [[get-listener-containers-matching]]
 == Enhanced Retrieval of MessageListenerContainers
 
-Introduced in version 3.2, the `ListenerContainerRegistry` provides advanced methods for dynamically filtering `MessageListenerContainer` instances, enhancing management capabilities.
-
-Two filtering methods are available:
-
-- **Filter by ID**: Using `getListenerContainersMatching(Predicate<String> idMatcher)`, you can filter containers by their IDs, ideal for ID-centric selection criteria.
-
-- **Advanced Filter by ID and Properties**: The `getListenerContainersMatching(BiPredicate<String, MessageListenerContainer>)` method allows for more sophisticated filtering, taking into account both the container's ID and its properties. This enables detailed filtering based on the container's state, configuration, or other relevant characteristics.
-
-```java
-// Example: Filter containers by ID
-Collection<MessageListenerContainer> filteredContainers =
-    registry.getListenerContainersMatching(id -> id.startsWith("specificPrefix-"));
-
-// Example: Advanced filter by ID and container state
-Collection<MessageListenerContainer> advancedFilteredContainers =
-    registry.getListenerContainersMatching((id, container) -> id.startsWith("specificPrefix-") && container.isRunning());
-```
+`ListenerContainerRegistry` provides two new API's dynamically find and filter `MessageListenerContainer` instances.
+`getListenerContainersMatching(Predicate<String> idMatcher)` to filter by ID and the other is
+`getListenerContainersMatching(BiPredicate<String, MessageListenerContainer> matcher)` to filter by ID and container properties.
 
 See xref:kafka/receiving-messages/kafkalistener-lifecycle.adoc#retrieving-message-listener-containers[`@KafkaListener` Lifecycle Management's API Docs] for more information.
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
@@ -83,3 +83,12 @@ See xref:kafka/seek.adoc#seek[Seek API Docs] for more details.
 For Spring Boot applications which define an application name, this name is now used
 as a default prefix for auto-generated client IDs for certain client types.
 See xref:kafka/connecting.adoc#default-client-id-prefixes[Default client ID prefixes] for more details.
+
+[[get-listener-containers-matching]]
+=== ListenerContainerRegistry's new getListenerContainersMatching(Predicate<String> idMatcher) method
+
+Introduces a new method `getListenerContainersMatching` since version 3.2, which returns all `MessageListenerContainer`
+instances whose ID matches a specified (non-null) `Predicate<String>`. This addition provides a flexible way to
+retrieve containers by more complex criteria than just a direct ID match.
+
+See xref:kafka/receiving-messages/kafkalistener-lifecycle.adoc#get-listener-containers-matching[ `@KafkaListener` Lifecycle Management API Docs] for more details.
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
@@ -85,11 +85,24 @@ as a default prefix for auto-generated client IDs for certain client types.
 See xref:kafka/connecting.adoc#default-client-id-prefixes[Default client ID prefixes] for more details.
 
 [[get-listener-containers-matching]]
-== Retrieve MessageListenerContainers with by matching a Predicate
+== Enhanced Retrieval of MessageListenerContainers
 
-Starting with version 3.2, the `ListenerContainerRegistry` introduces the `getListenerContainersMatching` method.
-This method filters `MessageListenerContainer` instances based on a `Predicate<String>` applied to their IDs.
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistry.java
```diff
@@ -115,6 +118,25 @@ public MessageListenerContainer getListenerContainer(String id) {
...
+	 * @see #getListenerContainer(String)
+	 */
+	@Override
+	public Collection<MessageListenerContainer> getListenerContainersMatching(Predicate<String> idMatcher) {
+		Assert.notNull(idMatcher, "'idMatcher' cannot be null");
+		List<MessageListenerContainer> matchingContainers = this.listenerContainers.entrySet()
+			.stream()
+			.filter(entry -> idMatcher.test(entry.getKey()))
+			.map(Map.Entry::getValue)
+			.collect(Collectors.toList());
```
artembilan, line 136: There is just `toList()` API.

spring-kafka/src/main/java/org/springframework/kafka/config/KafkaListenerEndpointRegistry.java
```diff
@@ -115,6 +118,25 @@ public MessageListenerContainer getListenerContainer(String id) {
 		return this.listenerContainers.get(id);
 	}
 
+	/**
+	 * Return all {@link MessageListenerContainer} instances with id matching the predicate or
+	 * empty {@link Collection} if no such container exists.
+	 * @param idMatcher the predicate to match container id with
+	 * @return the containers or empty {@link Collection} if no container with that id exists
+	 * @see #getListenerContainerIds()
+	 * @see #getListenerContainer(String)
```
artembilan, line 127: `@since 3.2`

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerContainerRegistry.java
```diff
@@ -41,6 +42,16 @@ public interface ListenerContainerRegistry {
...
 	MessageListenerContainer getListenerContainer(String id);
 
+	/**
+	 * Return all {@link MessageListenerContainer} instances with id matching the predicate or
+	 * empty {@link Collection} if no such container exists.
+	 * @param idMatcher the predicate to match the container id with
+	 * @return the containers or empty {@link Collection} if no container with that id exists
+	 * @see #getListenerContainerIds()
+	 * @see #getListenerContainer(String)
+	 */
```
artembilan, line 52: `@since 3.2`

spring-kafka/src/test/java/org/springframework/kafka/config/KafkaListenerEndpointRegistryTests.java
```diff
@@ -47,4 +63,77 @@ void unregister() {
...
+	@DisplayName("getListenerContainersMatching should return unmodifiable list")
+	@Test
+    void testGetListenerContainersMatchingReturnsUnmodifiableList() {
+        // Given
+        KafkaListenerEndpointRegistry registry = new KafkaListenerEndpointRegistry();
+        registerListenerWithId(registry, "foo");
+        // When
+        Collection<MessageListenerContainer> listeners = registry.getListenerContainersMatching(s -> true);
+        // Then
+        assertThrows(UnsupportedOperationException.class, () -> listeners.add(mock(MessageListenerContainer.class)));
```
artembilan, line 87: Why don't use `assertThatExceptionOfType(UnsupportedOperationException.class)` instead?

It also probably going to fail on Checkstyle since we have a limited set of allowed `static` method imports.

spring-kafka/src/test/java/org/springframework/kafka/config/KafkaListenerEndpointRegistryTests.java
```diff
@@ -47,4 +63,77 @@ void unregister() {
...
+			// Corner case: No names match as the predicate always returns false
+			Arguments.of(names, (Predicate<String>) id -> false, 0)
+		);
+	}
+
+	private void registerWithListenerIds(KafkaListenerEndpointRegistry registry, List<String> names) {
+		names.forEach(name -> registerListenerWithId(registry, name));
+	}
+
+	private void registerListenerWithId(KafkaListenerEndpointRegistry registry, String id) {
```
artembilan, line 130: Looks like these both methods can be `static`.

artembilan (changes requested):

Also consider to mention this new feature in the `whats-new.adoc` and in the `kafkalistener-lifecycle.adoc`.

Thanks

spring-kafka/src/test/java/org/springframework/kafka/config/KafkaListenerEndpointRegistryTests.java
JooHyukKim: Right, looking now there is no pre-existing usage of `assertThatExceptionOfType()`

JooHyukKim:
I've updated the author list and added documentation for the new feature in `whats-new.adoc` and `kafkalistener-lifecycle.adoc`.

I noticed there seems to be limited documentation on `KafkaListenerEndpointRegistry`, particularly methods like `getListenerContainers()` and `getListenerContainer(String)`. I can
 address this in a future PR to enhance docs further. WDYT, @artembilan?

artembilan:
Sure! You can add those couple sentences just in this PR. We have a release on Monday, so it will be reviewed immediately .
Thanks

JooHyukKim:
Added some more doc as discussed 👍🏼 cc. @artembilan

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -83,3 +83,13 @@ See xref:kafka/seek.adoc#seek[Seek API Docs] for more details.
 For Spring Boot applications which define an application name, this name is now used
 as a default prefix for auto-generated client IDs for certain client types.
 See xref:kafka/connecting.adoc#default-client-id-prefixes[Default client ID prefixes] for more details.
+
+[[get-listener-containers-matching]]
+== Retrieve MessageListenerContainers with by matching a Predicate
+
+Starting with version 3.2, the `ListenerContainerRegistry` introduces the `getListenerContainersMatching` method.
+This method filters `MessageListenerContainer` instances based on a `Predicate<String>` applied to their IDs.
```
artembilan, line 91: After your great docs I wonder if we could extend this new feature to something like this `BiPredicate<String, MessageListenerContainer>`. This way we may filter desired containers not only by their ids, but via some other option form the container itself.

WDYT?

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
JooHyukKim: Makes sense 👍🏼
I wonder if we should have both or just `BiPredicate<String, MessageListenerContainer>` 🤔
Extended feature sounds very powerful, but might be too verbose in someway, knowing that `id` is the key used to access the `MessageContainerListener`.

JooHyukKim: Since the `String, MessageListenerContainer` pair in `BiPredicate` technically represents the key-value pair stored in the registry, how about having both?

JooHyukKim: Should we do a separate PR or do everything in this one?

artembilan: This one is OK.
Thanks

JooHyukKim: Alright, hopefully it would probably take an hour or two

spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerContainerRegistry.java
```diff
@@ -41,6 +44,29 @@ public interface ListenerContainerRegistry {
...
+	 * Return all {@link MessageListenerContainer} instances that satisfy the given bi-predicate.
+	 * The {@code BiPredicate<String, MessageListenerContainer>} takes the container id and the container itself as arguments.
+	 * This allows for more sophisticated filtering, including properties or state of the container itself.
+	 * @param idAndContainerMatcher the bi-predicate to match the container id and the container
+	 * @return the containers that match the bi-predicate criteria or an empty {@link Collection} if no matching containers exist
+	 * @since 3.2
+	 * @see #getListenerContainerIds()
+	 * @see #getListenerContainersMatching(Predicate)
+	 */
+	Collection<MessageListenerContainer> getListenerContainersMatching(BiPredicate<String, MessageListenerContainer> idAndContainerMatcher);
```
artembilan, line 68: Let's see if you can follow the rule no more than 120 symbols per line!

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -83,3 +83,26 @@ See xref:kafka/seek.adoc#seek[Seek API Docs] for more details.
 For Spring Boot applications which define an application name, this name is now used
 as a default prefix for auto-generated client IDs for certain client types.
 See xref:kafka/connecting.adoc#default-client-id-prefixes[Default client ID prefixes] for more details.
+
+[[get-listener-containers-matching]]
+== Enhanced Retrieval of MessageListenerContainers
+
+Introduced in version 3.2, the `ListenerContainerRegistry` provides advanced methods for dynamically filtering `MessageListenerContainer` instances, enhancing management capabilities.
```
artembilan, line 90: This file is fully specific to particular version we mention in the beginning of the file.
More over the content of the file is just a list of topics with links to details.
You do show enough in the `kafkalistener-lifecycle.adoc`, but that all must not be duplicated in `What's New` file.

spring-kafka/src/test/java/org/springframework/kafka/config/KafkaListenerEndpointRegistryTests.java
```diff
@@ -47,4 +60,123 @@ void unregister() {
...
+	private static void registerListenerWithId(KafkaListenerEndpointRegistry registry, String id) {
+		KafkaListenerEndpoint endpoint = mock(KafkaListenerEndpoint.class);
+		@SuppressWarnings("unchecked")
+		KafkaListenerContainerFactory<MessageListenerContainer> factory = mock(KafkaListenerContainerFactory.class);
+		given(endpoint.getId()).willReturn(id);
+		MessageListenerContainer container = mock(MessageListenerContainer.class);
+		given(container.isRunning()).willReturn(true);
+		given(factory.createListenerContainer(endpoint)).willReturn(container);
+		registry.registerListenerContainer(endpoint, factory);
+	}
```
artembilan, line 181: Every member (including the last method of the last class) in the class must be surrounded with blank lines for much better code readability.

artembilan (changes requested):
The API is great: only some cosmetic concerns in regards of our code style rules.
Thanks

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
JooHyukKim: True, thank you for pointing out!

PR #3147, opened 2024-03-21
Remove org.apache.kafka native runtime hints

sobychacko:
* Since several of the native runtime hints that Spring for Apache Kafka provides for org.apahce.kafka classes are available via the graalvm-reachability-metadata repository (https://github.com/oracle/graalvm-reachability-metadata), we are no longer keeping these runtime hints in Spring for Apache Kafka since version 3.2.0. This applies to both kafka-client classes and Kafka Streams classes.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHints.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHints.java
@@ -21,33 +21,10 @@ import java.util.zip.CRC32C;
 
 import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.clients.consumer.Consumer;
-import org.apache.kafka.clients.consumer.CooperativeStickyAssignor;
-import org.apache.kafka.clients.consumer.RangeAssignor;
-import org.apache.kafka.clients.consumer.RoundRobinAssignor;
-import org.apache.kafka.clients.consumer.StickyAssignor;
 import org.apache.kafka.clients.producer.Producer;
-import org.apache.kafka.clients.producer.RoundRobinPartitioner;
 import org.apache.kafka.common.message.CreateTopicsRequestData.CreatableTopic;
 import org.apache.kafka.common.protocol.Message;
-import org.apache.kafka.common.serialization.ByteArrayDeserializer;
-import org.apache.kafka.common.serialization.ByteArraySerializer;
-import org.apache.kafka.common.serialization.ByteBufferDeserializer;
-import org.apache.kafka.common.serialization.ByteBufferSerializer;
-import org.apache.kafka.common.serialization.BytesDeserializer;
-import org.apache.kafka.common.serialization.BytesSerializer;
-import org.apache.kafka.common.serialization.DoubleDeserializer;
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHints.java
```diff
@@ -140,34 +117,10 @@ public void registerHints(RuntimeHints hints, @Nullable ClassLoader classLoader)
...
 					AppInfo.class,
-					// standard assignors
-					CooperativeStickyAssignor.class,
-					RangeAssignor.class,
-					RoundRobinAssignor.class,
-					StickyAssignor.class,
 					// standard partitioners
 					org.apache.kafka.clients.producer.internals.DefaultPartitioner.class,
-					RoundRobinPartitioner.class,
 					org.apache.kafka.clients.producer.UniformStickyPartitioner.class,
```
artembilan, line 122: Why do we still keep this Kafka client classes here?

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHints.java
```diff
@@ -200,15 +142,10 @@ public void registerHints(RuntimeHints hints, @Nullable ClassLoader classLoader)
 
 		Stream.of(
 				"sun.security.provider.ConfigFile",
-				"org.apache.kafka.streams.processor.internals.StreamsPartitionAssignor",
-				"org.apache.kafka.streams.errors.DefaultProductionExceptionHandler",
-				"org.apache.kafka.streams.processor.FailOnInvalidTimestamp",
-				"org.apache.kafka.streams.processor.internals.assignment.HighAvailabilityTaskAssignor",
 				"org.apache.kafka.streams.processor.internals.assignment.StickyTaskAssignor",
 				"org.apache.kafka.streams.processor.internals.assignment.FallbackPriorTaskAssignor",
```
artembilan, line 146: Do we really still need to keep these two, as well?

artembilan (changes requested):
There is also a typo in the commit message about Spring for Apache Kafka version.
I believe it has to be the current one - `3.2`

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHints.java
sobychacko: ya, they are missing from the metadata repo. Once they are added to the metadata, then we can remove them.

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHints.java
sobychacko: Missing in the metadata repo.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/aot/KafkaRuntimeHints.java
artembilan: So, all good. Leaving as is for now.

Issue #2381, opened 2022-08-24
Support kafka parallel-consumer


bojanv55:
**Expected Behavior**

Enable parallel processing of the messages in single consumer.

**Current Behavior**

Single-threaded processing of the messages.

**Context**

Are there any plans to include `ParallelStreamProcessor` as an option for spring-kafka? https://github.com/confluentinc/parallel-consumer It should handle automatically multiple threads, acking and other stuff (synchronization on same key if needed etc.).

Currently I can configure this manually, but would be easier I guess If I would only need to implement  `void poll(Consumer<PollContext<K, V>> usersVoidConsumptionFunction);`, and the rest is taken care by the spring.

garyrussell:
No plans currently; but contributions are welcome!

garyrussell:
That said; it's easy enough to use it. from Spring - just set up a `@KafkaListener` as normal; set the auto startup to "false" and then invoke the listener from the parallel consumer:
```java
@SpringBootApplication
public class Kgh2381Application {

	private static final Logger log = LoggerFactory.getLogger(Kgh2381Application.class);

	public static void main(String[] args) {
		SpringApplication.run(Kgh2381Application.class, args);
	}

	@KafkaListener(id = "kgh2381", topics = "kgh2381", autoStartup = "false")
	void listen(String in) {
		log.info(in);
	}

	@Bean
	public NewTopic topic() {
		return TopicBuilder.name("kgh2381").partitions(1).replicas(1).build();
	}

	@Bean
	ApplicationRunner runner(KafkaListenerEndpointRegistry registry, ConsumerFactory<String, String> cf,
			KafkaTemplate<String, String> template) {

		return args -> {
			MessageListener messageListener = (MessageListener) registry.getListenerContainer("kgh2381")
					.getContainerProperties().getMessageListener();
			Consumer<String, String> consumer = cf.createConsumer("group", "");
			var options = ParallelConsumerOptions.<String, String>builder()
					.ordering(ProcessingOrder.KEY)
					.consumer(consumer)
					.maxConcurrency(10)
					.build();
			ParallelStreamProcessor<String, String> processor = ParallelStreamProcessor
					.createEosStreamProcessor(options);
			processor.subscribe(List.of("kgh2381"));
			processor.poll(context -> messageListener.onMessage(context.getSingleConsumerRecord(), null, consumer));
			IntStream.range(0, 10).forEach(i -> template.send("kgh2381", "foo" + i));
		};
	}

}
```

[...]

PR #3161, opened 2024-03-24
Introduce Confluent ParallelConsumer to Spring Kafka

chickenchickenlove:
### Introduction
Hi, Spring Team. 
This is chickenchickenlove, huge fan of you guys.
I saw this issue ([Support kafka parallel-consumer](https://github.com/spring-projects/spring-kafka/issues/2381)), and this comment as well! 
> No plans currently; but contributions are welcome!

So, i'd like to implement that integrate `Parallel Consumer` to `spring-kafka`
This is skeleton code.

Please take a look when you have some free time. 🙇‍♂️ 
Thank you for your time!

### Background
- Sometimes, Confluent `Parallel consumer` will be more helpful than kafka consumer, because parallel consumer have good parallel processing (If you want to know more, See [this](https://github.com/confluentinc/parallel-consumer?tab=readme-ov-file#51-illustrative-performance-example)).
- Thus, i think it will be good option to `spring-kafka` as well. 

### How To use
```java
@EnableKafka
@EnableParallelConsumer
@SpringBootApplication
public class SpringkafkasampleApplication {

		SpringApplication.run(SpringkafkasampleApplication.class, args);
	}
}
```
- Annotated `@EnableParallelConsumer` to your Spring application. then, some spring bean related with `parallel consumer`, will be registered to `ApplicationContext`

```java
@Component
public class MyTestClass implements ParallelConsumerCallback<String, String> {

    private final List<String> topics;
    public MyTestClass(List<String> topics) {
        this.topics = topics;
    }

    @Override
    public void accept(PollContext<String, String> context) {
        context.getConsumerRecordsFlattened().forEach(kvConsumerRecord -> System.out.println(kvConsumerRecord));
    }

    @Override
    public List<String> getTopics() {
        return this.topics;
    }
}
```
- and then, you should implement interface `ParallelConsumerCallback` and register their concrete class as spring bean by annotating `@Component` or `@Configuration`.
- then, `ParallelConsumerFactory` will start `ParallelConsumer` and `ParallelConsumer` process records with `ParallelConsumerCallback` concrete class. 

### ETC
- How about follow the way like `@KafkaListener`? 
  - IMHO, it is unnecessary. because i think `ParallelConsumer` can allow only one callback. thus, it is not suitable for `ParallelConsumer`.
- Can we remove `ParallelConsumerImportSelector`? 
  - Yes. if `org.springframework.boot.autoconfigure.kafka.KafkaAutoConfiguration.java` class support to register spring bean, `ParallelConsumerImportSelector` can be removed.

Thank you for your time, Again.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,6 +73,7 @@ ext {
 	springRetryVersion = '2.0.5'
 	springVersion = '6.1.5'
 	zookeeperVersion = '3.8.4'
+	parallelConsumerVersion = '0.5.2.8'
 
 	idPrefix = 'kafka'
 
@@ -288,6 +289,9 @@ project ('spring-kafka') {
 		optionalApi ('com.fasterxml.jackson.module:jackson-module-kotlin') {
 			exclude group: 'org.jetbrains.kotlin'
 		}
+
+		// Parallel Consumer
+		api "io.confluent.parallelconsumer:parallel-consumer-core:$parallelConsumerVersion"
 
 		// Spring Data projection message binding support
 		optionalApi ("org.springframework.data:spring-data-commons") {
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableParallelConsumer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/EnableParallelConsumer.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2014-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.annotation;
+
+import org.springframework.context.annotation.Import;
+import org.springframework.kafka.config.ParallelConsumerConfiguration;
+import org.springframework.kafka.config.ParallelConsumerImportSelector;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * If you want to import {@link ParallelConsumerConfiguration} to your application,
+ * you just annotated {@link EnableParallelConsumer} to your spring application.
+ * @author ...
+ * @since 3.2.0
+ */
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Import(ParallelConsumerImportSelector.class)
+public @interface EnableParallelConsumer {
+}
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/ParallelConsumerConfig.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/ParallelConsumerConfig.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2014-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.config;
+
+import java.time.Duration;
+import java.util.HashMap;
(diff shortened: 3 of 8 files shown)


sobychacko:
@chickenchickenlove Thank you for the PR and the kind words. Next month, we will release the first `RC` version of `3.2.0`. Since this is an entirely new feature and involves dependencies from `io.confluent` (as opposed to `org.apache` for the regular consumer), this needs a detailed review to come up with a plan for supporting parallel consumers. I think you are in the right direction, but we will have an internal discussion on this matter. Depending on how things go, we may have to defer this feature to the `3.3.x` release, but we will let you know.

chickenchickenlove:
Hello, @sobychacko !
Thanks for your time 🙇‍♂️ 

I see! however, that's sounds great.  👍 
If after internal discussions, spring-kafka team decides to introduce `ParallelConsumer` to `spring-kafka`, and there is anything more to do, Please let me know. I'm more than willing to help.

Also, Please let me know in case of a decision not to introduce `ParallelConsumer` to spring-kafka!"

Thanks for your time, again 🙇‍♂️

sobychacko:
@tagorenathv We will look at this after the GA for `3.2.0` (this month) as part of the next minor version.

BartlomiejSzczotka:
hello @sobychacko any plans to implement it in near future?

sobychacko:
@BartlomiejSzczotka Yes, we will still look at this for the `3.3.0` version. It may not be possible for it to be ready in time for the upcoming M1 release, but it will likely be in the M2 timeframe. Thanks!

chickenchickenlove:
Hi, @sobychacko ! 
FYI, this PR is currently a draft because we haven't reached a consensus yet.
This means that it reflects only my perspective.
I believe there are better ways to support `ParallelConsumer`, and we can find them through discussion. 

Also, i made new commits for improving code quality at April 27, 28! 

So, Take your times and please feel free to talk anything, any thoughs! 
Once we reach a consensus, i will write test codes for this feature as well!

Thanks a lot to look at this PR! 😃

chickenchickenlove:
Hi, @sobychacko .Gently ping. 🙇‍♂️

It's been a while. 
Is the integration of the parallel consumer still under review?
Please let me know 🙏

sobychacko:
@chickenchickenlove For various reasons on our end, we decided to postpone this feature to either the next minor (`3.4.0`) or the next major release (`4.0.0`). Either way, we will start working on it in a few months.

chickenchickenlove:
Thanks checking 👍 
I will be waiting for cue 😃

chickenchickenlove:
Hi, @sobychacko ! 
I've heard that `4.0.x` is coming in a while.
So, I'm curious to know what you think about bringing `ParallelConsumer` to `Spring-Kafka`.
Please let me know, when you have time. 🙇‍♂️

artembilan:
Spring for Apache Kafka `4.0` is going to be based on Apache Kafka `4.0`.
I wonder how that https://github.com/confluentinc/parallel-consumer is going to be compatible.
Perhaps it would be better to wait until it has released respective version based on Apache Kafka `4.0` as well...

chickenchickenlove:
@artembilan Thanks for your comment 🙇‍♂️ 
The latest `parallel-consumer` uses `kafka 3.7.1`. ([link](https://github.com/confluentinc/parallel-consumer/blob/a478d1f6caa273df5a3260dc8b05a21597ad9a81/pom.xml#L102))
Ah, Yes, I agree with you! 

I have a question. 
If `spring-kafka` version becomes `4.x.x` and, `parallel-consumer` supports apache kafka `4.0.0`, in this situation, 
could I try to implement `parallel-consumer`, should I postpone it? 

Looking back, It might be better for `spring-kafka` not to support `parallel-consumer`, 
because `parallel-consumer` is not as active as `spring-kafka`.
So, the slow updates of parallel-consumer could actually become an obstacle for spring-kafka in adopting newer versions of Apache Kafka.

WDYT?

artembilan:
That's true, too.
We also don't want to have a module based on the out-dated and not supported library.
So, perhaps we just should not do this and continue to advertise our concurrency and reactive streams support in listener container.

sobychacko:
@chickenchickenlove We have an action item on our end to re-evaluate parallel consumer as part of the 4.0.0 work. I suggest, leave this open and we can make a decision on it once we actively start working on the 4.0.x changes.

artembilan:
We are probably going to reject this since we are already based on the latest Apache Kafka `4.0`, but that library is not.

leonardotorresaltez:
Following this discussion, I would like to strongly encourage continuing the work on integrating Confluent Parallel Consumer with Spring Kafka.

While the proposed queues-based approach is promising, it currently only addresses the unordered processing use case. A very important scenario remains uncovered: key-level ordering with parallelism.

This is precisely where Parallel Consumer brings unique value:

- It guarantees ordering per key while still allowing high throughput.
- It solves a problem that is non-trivial to implement correctly at the application level.

Covering key-level ordering through queues would most likely require adding significant and heavy capabilities to the broker, which seems misaligned with Kafka’s long-standing philosophy of keeping the broker lightweight and simple, pushing complexity to clients instead.

For many real-world enterprise use cases (e.g. per-customer, per-account, or per-entity processing), key-level ordering is a hard requirement, and Parallel Consumer is already a mature, well-tested solution in this space.

For these reasons, continuing the integration of Parallel Consumer into Spring Kafka would provide substantial value, complementing the queues work rather than overlapping with it.

Thank you for considering this perspective.

sobychacko:
@leonardotorresaltez parallel-consumer as I can understand it, does not support Kafka 4.0.x (at least yet). That would be the biggest blocker even if someone from the community want to work on upgrading the PR we had.

PR #3162, opened 2024-03-24
Add `checkstyle` rule on `empty-line-separater`

JooHyukKim:
### Motivation

This PR adds `checkstyle` rule to enforce blank lines around things like class, methods, and constructors, etc..., except `field` making our codebase cleaner and more consistent.

### Key settings:

- Mandatory blank lines between class members, "excluding" fields.
- Only single blank lines are allowed for separation.

### Additional Note

Idea from recent [PR comment](https://github.com/spring-projects/spring-kafka/pull/3137#discussion_r1531001776). Figured why not have it checked during build?

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
@@ -204,7 +204,6 @@ public class EmbeddedKafkaKraftBroker implements EmbeddedKafkaBroker {
 		}
 	}
 
-
 	private void start() {
 		if (this.cluster != null) {
 			return;
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/assertj/KafkaConditions.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -89,7 +89,6 @@ public final class KafkaConditions {
 	public static Condition<ConsumerRecord<?, ?>> partition(int partition) {
 		return new ConsumerRecordPartitionCondition(partition);
 	}
-
 
 	public static class ConsumerRecordKeyCondition<K> extends Condition<ConsumerRecord<K, ?>> {
 
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/condition/EmbeddedKafkaCondition.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/condition/EmbeddedKafkaCondition.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2023 the original author or authors.
+ * Copyright 2019-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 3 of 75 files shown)


artembilan (changes requested):
Wow! I didn’t know there is such a rule. I prefer to have blank lines around fields as well, though. 
Thank you for great job!

JooHyukKim:
> Wow! I didn’t know there is such a rule. I prefer to have blank lines around fields as well, though. Thank you for great job!

@artembilan if we do cover `fields` as well, there will be (I think) hundreds of lines of line changes. Are we okay with that?

Asking this because some OSS teams prefer not to push thru such changes for sake of easier debugging around Git.

artembilan:
Yes, I’m Ok with that. As long as your change is only about blank lines around class members and nothing more. So, we know to review only Checkstyle config - and everything else is just a follow up change to satisfy that rule.

JooHyukKim:
> Yes, I’m Ok with that. As long as your change is only about blank lines around class members and nothing more. So, we know to review only Checkstyle config - and everything else is just a follow up change to satisfy that rule.

Alright 👍🏼 Will follow up with changes in a day.

artembilan:
@JooHyukKim ,

May I know, please, how have you fixed all those problems with the code after applying this rule?
What I mean that I'd like to apply this rule in my other projects where I also comment too much that blank lines must be around all the class members.
This way I'd follow your experience.

Thanks again!

JooHyukKim:
@artembilan , I did it all manually but with the help of IntelliJ. Basically what I did was ....

1. Add check style rule, and
2. Run checkstyle locally, then
3. Fix violations

... and that's it. I guess [ moving fast + leveraging IntelliJ ] was my approach. 
Below is an example with photo which I labeled each step...

(1) run checkstyle via IntelliJ
(2) Then click to navigate then fix

image

I wish I could tell you more (maybe a automated tool)🥲, but hope it helps 👍🏼!

artembilan:
That’s great! Thank for sharing your experience and sorry for making you to do that job 🙇‍♂️.

So, in Spring Integration project we use this IntelliJ IDEA editor config: https://github.com/spring-projects/spring-integration/blob/main/src/idea/spring-framework.xml. It seems for me it is very close to what we expect with the respective Checkstyle config. I’ll come back tomorrow with my findings when I add this blank lines rule.

Another feature wanted to share is: https://github.com/spring-io/spring-javaformat.
Very convenient tool to move fast, although I’m not always agree with its decision on method call chain.
We may consider to move there, though, next month when we are in GA phase, so no new features, but refactoring is OK.

artembilan:
So, for Spring Integration I went a head with rule like this:
```
		<module name="EmptyLineSeparator">
			<property name="tokens"
					  value="CLASS_DEF, INTERFACE_DEF, ENUM_DEF, STATIC_INIT, INSTANCE_INIT, METHOD_DEF, CTOR_DEF, VARIABLE_DEF, RECORD_DEF, COMPACT_CTOR_DEF "/>
			<property name="allowMultipleEmptyLines" value="false"/>
		</module>
```

Then I modified the mentioned [spring-framework.xml](https://github.com/spring-projects/spring-integration/blob/main/src/idea/spring-framework.xml) file with these extra rules:
```
                <option name="KEEP_BLANK_LINES_IN_DECLARATIONS" value="1"/>
		<option name="KEEP_BLANK_LINES_IN_CODE" value="1"/>
		<option name="KEEP_BLANK_LINES_BETWEEN_PACKAGE_DECLARATION_AND_HEADER" value="1"/>
		<option name="BLANK_LINES_AFTER_CLASS_HEADER" value="1"/>
		<option name="BLANK_LINES_AFTER_ANONYMOUS_CLASS_HEADER" value="1"/>
		<option name="BLANK_LINES_BEFORE_CLASS_END" value="1"/>
```

The ran code reformat for the whole project with a `*.java` mask.
And only after that the `gradlew check -x test` command to be sure in a new Checkstyle rule.
It failed only for a couple use-cases where we declare classes in the method body 😄

JooHyukKim:
Good to know it worked! I guess we could consider applying to this project also (or any other Spring projects..!)
Thank you for sharing 😆 @artembilan

PR #3163, opened 2024-03-25
[Suggestion] Refactor KafkaListenerConfigUtils to Use Enum for Constants Management

JoeCP17:
This pull request introduces a significant refactor to the KafkaListenerConfigUtils class, transitioning it from an abstract class to an enum named KafkaListenerConfigTypes. 

This change aims to enhance the management of configuration constants within our project, leveraging the strengths of enum in Java for a more efficient and type-safe way to handle constants.

## Key Changes
KafkaListenerConfigUtils is now an enum KafkaListenerConfigTypes, encapsulating the bean names as enum constants.

Introduced a nested `KAFKA_CONFIG_PATH` class within `KafkaListenerConfigTypes` to maintain the original constants'values.

## Rationale
The primary motivation behind this change is to utilize enums for constant management, which offers several advantages

### Type Safety
Enums provide compile-time type checking, reducing the risk of assigning incorrect values to constants.

### Singleton Pattern 
By their nature, enums in Java implement the singleton pattern, ensuring that constant values are instantiated only once throughout the JVM lifecycle.

### Impact
This change does not introduce any breaking changes to existing functionality.
All references to KafkaListenerConfigUtils constants have been updated to use KafkaListenerConfigTypes accordingly.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaBootstrapConfiguration.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaBootstrapConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


artembilan (changes requested):
Sorry, I cannot accept this change since it is more an abusing of `enum` feature over simple string constant. Even if it is a bit type safe internally for the framework, it is introducing some inconvenience for end-users like your change in the `AliasPropertiesTests` with that `@Bean(KafkaListenerConfigTypes.KAFKA_CONFIG_PATH.KAFKA_LISTENER_ENDPOINT_REGISTRY_BEAN_NAME)`.

We really turn to `enum` only if there are more than just plain string constant in the logic. The rest is a simple pattern to follow: just use some util class with those simple string constants.

artembilan:
Closed as `Works as Designed`.

Issue #3166, opened 2024-03-27
BatchInterceptor is not working properly with retries


RuslanHryn:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.1.3, also reproducible on older versions as well

**Describe the bug**
No invocation of `org.springframework.kafka.listener.BatchInterceptor#success` after successful retry.
Also, only one invocation of `org.springframework.kafka.listener.BatchInterceptor#failure` for the first failure but no invocations for failed retries.

**To Reproduce**

Just use `BatchInterceptor` with retries.

Here is a test to reproduce the issue: https://github.com/RuslanHryn/kafka-batch-Interceptor-issue/blob/main/src/test/java/org/example/kafkabatchinterceptorissue/KafkaBatchInterceptorIssueTests.java

**Expected behavior**
`org.springframework.kafka.listener.BatchInterceptor#success` -  will be invoked after a successful retry.
`org.springframework.kafka.listener.BatchInterceptor#failure` - will be invoked for every listener failure including retries.
**Sample**

here is the reproducer: https://github.com/RuslanHryn/kafka-batch-Interceptor-issue

sobychacko:
@RuslanHryn Thanks for the report and the test to reproduce the issue. We will triage and get back to you.

PR #3167, opened 2024-03-28
GH-3166: BatchInterceptor issues with retries

sobychacko:
Fixes: #3166

* When retries are enabled, batch interceptor is not invoking the intercept methods for failures on retries and the possible eventual success method call. Addressing this issue.

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
@@ -4019,7 +4019,6 @@ public class KafkaMessageListenerContainerTests {
 		inOrder.verify(interceptor).setupThreadState(eq(consumer));
 		inOrder.verify(consumer).poll(Duration.ofMillis(ContainerProperties.DEFAULT_POLL_TIMEOUT));
 		inOrder.verify(interceptor).intercept(any(), eq(consumer));
-		inOrder.verify(interceptor).success(any(), eq(consumer));
 		inOrder.verify(consumer).commitSync(eq(Map.of(new TopicPartition("foo", 0), new OffsetAndMetadata(2L))),
 				any(Duration.class));
 		container.stop();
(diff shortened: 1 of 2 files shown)


spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -4019,7 +4019,6 @@ public ConsumerRecords<Integer, String> intercept(ConsumerRecords<Integer, Strin
 		inOrder.verify(interceptor).setupThreadState(eq(consumer));
 		inOrder.verify(consumer).poll(Duration.ofMillis(ContainerProperties.DEFAULT_POLL_TIMEOUT));
 		inOrder.verify(interceptor).intercept(any(), eq(consumer));
-		inOrder.verify(interceptor).success(any(), eq(consumer));
```
sobychacko, line 4022: This is because we are no longer invoking the `success` method because of [this](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2356). Since that is returned because the records are null, the intercept calls are not being reached. This might be the right thing, because earlier we were calling the `success` intercept call unnecessarily.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -4240,6 +4239,82 @@ public void clearThreadState(Consumer<?, ?> consumer) {
...
+		ConsumerRecord<Integer, String> firstRecord = new ConsumerRecord<>("test-topic", 0, 0L, 1, "data-1");
+		ConsumerRecord<Integer, String> secondRecord = new ConsumerRecord<>("test-topic", 0, 1L, 1, "data-2");
+		Map<TopicPartition, List<ConsumerRecord<Integer, String>>> records = new HashMap<>();
+		records.put(new TopicPartition("test-topic", 0), List.of(firstRecord, secondRecord));
+		ConsumerRecords<Integer, String> consumerRecords = new ConsumerRecords<>(records);
+		AtomicInteger invocation = new AtomicInteger(0);
+		given(consumer.poll(any(Duration.class))).willAnswer(i -> {
+			Thread.sleep(50);
+			if (invocation.get() == 0) {
+				invocation.getAndIncrement();
```
artembilan, line 4257: This definitely can be a part of that `if`: no need to call `get` and then `getAndIncrement`. The last one would contribute exactly same value for that `if`.
Also I'm not sure that `Thread.sleep(50)`. Even if it is small enough, we still need to be sure that we need such a delay.
Why to make test slower for nothing?

artembilan (changes requested):
Just one minor nit-pick in the new unit test.

I guess we are still going to back-port the fix into all our supported version?

sobychacko:
@artembilan I added the backport labels to the original issue. I will tackle the test comments soon.

artembilan (approved):
LGTM.

Will merge when build is green.

PR #3168, opened 2024-03-28
Add an option to leave consumer group when closing Streams

juwit:
👋 Hi Spring-Kafka Team !

When developping with Kafka Streams, I figured that when shutting down an app, the consumer didn't seem to proactively leave the Kafka consumer group.
This seems to be related to a `leaveGroup` parameter that wasn't used in the `org.apache.kafka.streams.KafkaStreams#close()` method.

So this PR adds support for this parameter, and documentation of the parameters in the *Apache Kafka Streams Support* section.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
@@ -55,6 +55,7 @@ import org.springframework.util.Assert;
  * @author Nurettin Yilmaz
  * @author Denis Washington
  * @author Gary Russell
+ * @author Julien Wittouck
  *
  * @since 1.1.4
  */
@@ -99,6 +100,8 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 	private int phase = Integer.MAX_VALUE - 1000; // NOSONAR magic #
 
 	private Duration closeTimeout = DEFAULT_CLOSE_TIMEOUT;
+
+	private boolean leaveGroupOnClose = false;
 
 	private KafkaStreams kafkaStreams;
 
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -225,6 +227,14 @@ public void setCloseTimeout(int closeTimeout) {
 		this.closeTimeout = Duration.ofSeconds(closeTimeout); // NOSONAR (sync)
 	}
 
+	/**
```
sobychacko, line 230: This needs an `@since` tag. In addition, please add your name as an `author` of this class.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -225,6 +227,14 @@ public void setCloseTimeout(int closeTimeout) {
...
+	 * Specify if the group should be leaved when stopping the Streams. Defaults to false.
```
sobychacko, line 231: `Specify if the consumer should leave the group when stopping Kafka Streams`?

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -383,7 +393,10 @@ public void stop() {
 			if (this.running) {
 				try {
 					if (this.kafkaStreams != null) {
-						this.kafkaStreams.close(this.closeTimeout);
+						this.kafkaStreams.close(new KafkaStreams.CloseOptions()
+								.timeout(closeTimeout)
```
sobychacko, line 397: `this.closeTimeout`.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -383,7 +393,10 @@ public void stop() {
...
+								.leaveGroup(leaveGroupOnClose)
```
sobychacko, line 398: `this.leaveGroupOnClose`

sobychacko (changes requested):

juwit:
Thanks for the review @sobychacko 🙏
I updated the code with your suggestions.

sobychacko (approved):

Issue #2609, opened 2023-03-08
More and proper tags for observation spans


cmergenthaler:
**Expected Behavior**

The current tags for tracing spans are very limited and contain only a few relevant information. [Opentelemetry](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/messaging/) defines all the recommended and required attributes a single message span should include. It would be nice to have them predefined by spring-kafka automatically instead of using the customizer explicitly. Even though customizing might work, we lack a few information on the customizer to follow otel standardization properly (used broker, consumergroup, partition for producer, etc.).

**Current Behavior**

The only tags currently included are spring related tags (e.g. `spring.kafka.template`) and do not follow any standard. There are other spring projects that already have proper otel tags defined (e.g. [spring-data-mongodb](https://github.com/spring-projects/spring-data-mongodb/blob/main/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/MongoObservation.java)). 

I think it would be beneficial to have consistent standard support over different spring projects. 

**Context**

To workaround this one can implement their own KafkaTemplateObservationConvention and define the necessary tags:

```java
@Component
public class KafkaTemplateObservationConvention extends KafkaTemplateObservation.DefaultKafkaTemplateObservationConvention {

    @Override
    public KeyValues getLowCardinalityKeyValues(KafkaRecordSenderContext context) {
        context.setRemoteServiceName(context.getBroker()); //peer.service -> should be the broker which is not available here
        return KeyValues.of(KafkaTemplateObservation.TemplateLowCardinalityTags.BEAN_NAME.asString(),
                context.getBeanName())
                .and("messaging.system", "kafka")
                .and("messaging.destination.kind", "topic")
                //.and("messaging.kafka.partition", context.getCarrier().partition().toString()) //Partition is always null
                //.and("messaging.kafka.message.offset", context.getCarrier().offset()) //Offset is not exposed
                .and("messaging.destination.name", context.getDestination());
    }

    @Override
    public String getContextualName(KafkaRecordSenderContext context) {
        return context.getDestination() + " publish";
    }
}
```

As you can see some required/recommended information is missing on the customizer.

cmergenthaler:
I'll create a PR for that soon :)

artembilan:
>  `return context.getDestination() + " publish";`

I believe previously it was `send`...
That's why we have it like that everywhere.

cmergenthaler:
No problem, I will just change that. 

[...]

PR #3169, opened 2024-03-28
GH-2609: providing more tracing tags

Wzy19930507:
Fixes: #2609

detail see #2616.

* fix review.
* polish `ObservationTests`.
* add doc in `whats-new.adoc`.


spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
```diff
@@ -89,18 +241,44 @@ public static class DefaultKafkaListenerObservationConvention implements KafkaLi
...
+		public KeyValues getHighCardinalityKeyValues(KafkaRecordReceiverContext context) {
+			KeyValues keyValues = KeyValues.of(
+					ListenerHighCardinalityTags.MESSAGING_PARTITION.withValue(context.getPartition()),
+					ListenerHighCardinalityTags.MESSAGING_OFFSET.withValue(context.getOffset())
+			);
+
+			if (StringUtils.hasText(context.getClientId())) {
+				keyValues = keyValues
+						.and(ListenerHighCardinalityTags.MESSAGING_CLIENT_ID.withValue(context.getClientId()))
+						.and(ListenerHighCardinalityTags.MESSAGING_CONSUMER_ID.withValue(getConsumerId(context)));
```
artembilan, line 266: According to the `getConsumerId()` logic, it can live without `clientId`, but it is never going to reach there since it is restricted here via that `if()`.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
```diff
@@ -89,18 +241,44 @@ public static class DefaultKafkaListenerObservationConvention implements KafkaLi
...
+		@NonNull
+		public KeyValues getHighCardinalityKeyValues(KafkaRecordReceiverContext context) {
+			KeyValues keyValues = KeyValues.of(
+					ListenerHighCardinalityTags.MESSAGING_PARTITION.withValue(context.getPartition()),
+					ListenerHighCardinalityTags.MESSAGING_OFFSET.withValue(context.getOffset())
+			);
+
+			if (StringUtils.hasText(context.getClientId())) {
+				keyValues = keyValues
+						.and(ListenerHighCardinalityTags.MESSAGING_CLIENT_ID.withValue(context.getClientId()))
```
artembilan, line 265: Too many `context.getClientId()` call.
Consider to extract it into a variable and also propagate it down to that `getConsumerId()` method.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordReceiverContext.java
```diff
@@ -28,16 +28,24 @@
...
 	private final String listenerId;
 
+	private final String clientId;
+
+	private final String groupId;
+
 	private final ConsumerRecord<?, ?> record;
 
-	public KafkaRecordReceiverContext(ConsumerRecord<?, ?> record, String listenerId, Supplier<String> clusterId) {
+	public KafkaRecordReceiverContext(ConsumerRecord<?, ?> record, String listenerId, String clientId, String groupId,
```
artembilan, line 47: This is a breaking change in the public API.
Let's consider to have a separate constructor, even if we would never call the old one from the framework.

artembilan (changes requested):
Thank you for looking into this!
Please, find some my requests about improvements.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordReceiverContext.java
Wzy19930507: Fix it.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
Wzy19930507: Fix it.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
Wzy19930507: Fix it.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
```diff
@@ -42,19 +49,32 @@ public Class<? extends ObservationConvention<? extends Context>> getDefaultConve
...
+
 	};
 
 	/**
 	 * Low cardinality tags.
+	 *
+	 * @author Christian Mergenthaler
+	 * @author Wang Zhiyang
+	 *
+	 * @since 3.2
```
artembilan, line 77: This inner class was there before, so this `@since` is misleading.
Plus we prefer to have `@author` only on top level class, kinda one set per Java file 😄

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
```diff
@@ -64,18 +84,151 @@ public enum ListenerLowCardinalityTags implements KeyName {
...
+
+	}
+
+	/**
+	 * High cardinality tags.
+	 *
+	 * @author Wang Zhiyang
+	 * @author Christian Mergenthaler
+	 *
+	 * @since 3.2
```
artembilan, line 167: This is correct, but again: no authors on the inner entities.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
```diff
@@ -64,18 +84,151 @@ public enum ListenerLowCardinalityTags implements KeyName {
...
+			public String asString() {
+				return "messaging.source.kind";
+			}
+
+		},
+
+		/**
+		 * Messaging the consumer group.
+		 */
+		MESSAGING_CONSUMER_GROUP {
```
artembilan, line 149: I believe all of these new `enum` values must be marked with `@since 3.2`.
Because they are addition to the existing class.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
```diff
@@ -89,18 +242,45 @@ public static class DefaultKafkaListenerObservationConvention implements KafkaLi
...
 
 		@Override
 		public String getContextualName(KafkaRecordReceiverContext context) {
 			return context.getSource() + " receive";
 		}
 
-		@Override
-		public String getName() {
-			return "spring.kafka.listener";
+		private String getConsumerId(KafkaRecordReceiverContext context, @Nullable String clientId) {
```
artembilan, line 279: Looks like this new method can be `static`.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordReceiverContext.java
```diff
@@ -28,16 +28,43 @@
...
+	/**
+	 * Construct a kafka record receiver context.
+	 * @param record 		the consumer record.
+	 * @param listenerId	the container listener id.
+	 * @param clientId		the kafka client id.
+	 * @param groupId		the consumer group id.
+	 * @param clusterId		the kafka cluster id.
+	 * @since 3.2
+	 */
+	public KafkaRecordReceiverContext(ConsumerRecord<?, ?> record, String listenerId, String clientId, String groupId,
```
artembilan, line 66: New ctor, so `@since 3.2`

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordReceiverContext.java
```diff
@@ -60,6 +89,14 @@ public String getListenerId() {
 		return this.listenerId;
 	}
 
+	public String getGroupId() {
+		return this.groupId;
+	}
+
+	public String getClientId() {
+		return this.clientId;
```
artembilan, line 97: Apparently both these new props can be `null`, so let's add `@Nullable` on their methods!

artembilan (changes requested):
Just some minor clean ups according to our code style.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
Wzy19930507: Sorry for this misleading, fix it. 😸

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
Wzy19930507: Good point. 👍

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordReceiverContext.java
Wzy19930507: Already have.

artembilan: Thanks.

If you agree with me, there is no need to say that via `Done/Fixed/OK` comments.
I'll see the change I've asked before in your next commit 😄

Wzy19930507: Thanks! You are right!

Issue #3020, opened 2024-02-08
Allows to initialize/inject the Topology object without starting the Kafka streams.


foal:
**Expected Behavior**

Get the Topology from StreamsBuilderFactoryBean without starting the streams engine. Topology have to be created with properties and KafkaStreamsInfrastructureCustomizer and all futures updated if any.

It is necessary to run the tests with Kafka TopologyTestDriver.

**Current Behavior**

I can inject the StreamsBuilder, but I then I need to create and initialize the topology manually.

**Context**

Manual work around:

```
    @Autowired
    private StreamsBuilder streamBuilder;
    @Autowired
    private KafkaStreamsConfiguration streamConfig;
    @Autowired
    private KafkaStreamsInfrastructureCustomizer infrastructureCustomizer;

///....

        Topology topology = streamBuilder.build(streamConfig.asProperties());
        infrastructureCustomizer.configureTopology(topology);

```
But of course, it aligned with current version of StreamsBuilderFactoryBean only.

sobychacko:
That is an interesting request. Could you elaborate on the use case that requires it?

artembilan:
I wonder why do you need `StreamsBuilderFactoryBean` if you create it manually.
There is also a PR like this https://github.com/spring-projects/spring-kafka/pull/1853 which does something with `TopologyTestDriver`.

The `StreamsBuilderFactoryBean` can also be configured for the `setAutoStartup(false)`.

So, we indeed need more info how to proceed.
Perhaps the fully formed test would be great start to look into.

Thanks

foal:
I do not create `StreamsBuilderFactoryBean` manually.  In the real application all configured automatically. 
The issue is in the tests. I have tree options:
 - Use `Spring Boot test` using `Testcontainers` - the Kafka Streams run exactly in the same way how it does in the application but tests take **much more times, resources** (the application works with 2 Kafka clusters) and **very difficult and unstable** due to asynchronous nature of the Kafka.
 - Use the plain `JUnit` test with Kafka `TopologyTestDriver`. Much faster and more predictable but I need to **duplicate the configuration, topology setup** (a have it as a set of beans too) etc. to run it. Also, it doesn't test the configuration from properties files and other things that handled in the real application by Spring.
 - Try to use the `Spring Boot test` but instead start the Kafka Streams use the `TopologyTestDriver` - it is my case. The reasonable balance between 100% identity with real application from the 1st case and lightweight and simplifying of testing from the 2nd case.

[...]

PR #3172, opened 2024-03-30
Create topology before kafka streams start.

chickenchickenlove:
### Introduction
Hi, Spring kafka Team.
This is chickenchickenlove, huge fan of you guys.
I saw this issue (https://github.com/spring-projects/spring-kafka/issues/3020). i want to resolve this problem by myself!

### Background
- https://github.com/spring-projects/spring-kafka/issues/3020
- If `Topology` of `kafka streams` are initialized before `StreamsBuilderFactoryBean#start()`, it will be very helpful to Test code for `spring kafka-streams`. `TopologyTestDriver` is very efficient to test comparing to `TestContainers`. 

### To reviewer
- I have tiny PR to solve Issue(https://github.com/spring-projects/spring-kafka/issues/3020).
- It might be a fairly simple fix, but it could look bad. IMHO, it's a better idea to register `StreamsBuilder` as a `Bean` in a different scope, however, it will introduce many changes to `StreamsBuilderFactoryBean`, `AbstractBuilderFactoryBean` as well. I have made modifications to minimize the changes. 
- When you have some free time, Please take a look 🙇‍♂️.

</br>

### Changes
- Point of initializing `kafka-streams topology` move to `StreamsBuilderFactoryBean#isAutoStartup()`.
</br>

### `@PostConstruct` or `StreamsBuilderFactoryBean#afterPropertiesSet()` Does not work well in this case. Why? 
- Because they are called just before bean is initialized. however, codes to make topology will be injected after `StreamsBuilderFactoryBean#afterPropertiesSet()` are called.
1. `StreamBuilder` is initialized in that time by this [code](https://github.com/spring-projects/spring-kafka/blob/c26f26e4598f40c0bf32b5294b110de438b3faf6/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java#L322-L336).  (`StreamsBuilderFactoryBean#afterPropertiesSet()` -> `StreamsBuilderFactoryBean#createInstance()` -> finally, `StreamsBuilder` are intialized!)
2. However, codes to make topology, are not injected `StreamsBuilder` in this time. you can refer to this topology code below.
```java
@Component
public class MyComponent {

    private static final Serde<String> STRING_SERDE = Serdes.String();
    private final StreamsBuilder streamsBuilder;

    public MyComponent(StreamsBuilder streamsBuilder) {
        this.streamsBuilder = streamsBuilder;
        buildPipeline();
    }

    public void buildPipeline() {
        KStream<String, String> messageStream = streamsBuilder
                .stream("input-topic", Consumed.with(STRING_SERDE, STRING_SERDE));
        ...
        wordCounts.toStream().to("output-topic");
    }
}
```
3. MyComponent#buildPipeline()` is executed after `StreamsBuilderFactoryBean#afterPropertiesSet()` called. After this step, `StreamsBuilder` can return valid `topology`. Before this step, `StreamsBuilder` will return invalid `topology` always.

~~### Why am i used to `isAutoStart()`?~~
~~- AFAIK, There is no hooking point for this. even, `ContextRefreshedEvent` will be called after `SmartLifeCycle#isAutoStart()`~~
~~- IMHO, this is best simple modification. In common use case, Spring will call `SmartLifeCycle#isAutoStart()` only once.  I did send request to spring team whether if they can implement new hooking point such like it. ([link](https://github.com/spring-projects/spring-framework/issues/32554))~~
</br>

### Any Other workaround? 
~~- IMHO, `StreamsBuilder` should be registered as bean other scope (not in `StreamsBuilderFactoryBean scope`). But it will involves a lot of changes to the `StreamsBuilderFactoryBean`.~~
~~- I was wrong. IMHO, there is no workaround, i guess 😢 .~~
- Finally, i used `SmartInitializingSingleton.afterSingletonsInstantiated()` to resolve this issue.

</br>

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
@@ -33,6 +33,7 @@ import org.apache.kafka.streams.processor.StateRestoreListener;
 import org.apache.kafka.streams.processor.internals.DefaultKafkaClientSupplier;
 
 import org.springframework.beans.factory.BeanNameAware;
+import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.beans.factory.config.AbstractFactoryBean;
 import org.springframework.context.SmartLifecycle;
 import org.springframework.core.log.LogAccessor;
@@ -61,7 +62,7 @@ import org.springframework.util.Assert;
  * @since 1.1.4
  */
 public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilder>
-		implements SmartLifecycle, BeanNameAware {
+		implements SmartLifecycle, BeanNameAware, SmartInitializingSingleton {
 
 	/**
 	 * The default {@link Duration} of {@code 10 seconds} for close timeout.
@@ -338,12 +339,6 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 
 	@Override
 	public boolean isAutoStartup() {
-		try {
-			this.topology = getObject().build(this.properties);
-		}
-		catch (Exception e) {
-			throw new RuntimeException(e);
-		}
 		return this.autoStartup;
 	}
 
@@ -363,8 +358,6 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 				try {
 					Assert.state(this.properties != null,
 							"streams configuration properties must not be null");
-					this.infrastructureCustomizer.configureTopology(this.topology);
-					LOGGER.debug(() -> this.topology.describe().toString());
 					this.kafkaStreams = new KafkaStreams(this.topology, this.properties, this.clientSupplier);
 					this.kafkaStreams.setStateListener(this.stateListener);
 					this.kafkaStreams.setGlobalStateRestoreListener(this.stateRestoreListener);
@@ -434,6 +427,18 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 		}
 		finally {
 			this.lifecycleLock.unlock();
+		}
+	}
+
+	@Override
+	public void afterSingletonsInstantiated() {
+		try {
+			this.topology = getObject().build(this.properties);
+			this.infrastructureCustomizer.configureTopology(this.topology);
+			LOGGER.debug(() -> this.topology.describe().toString());
+		}
+		catch (Exception e) {
+			throw new RuntimeException(e);
 		}
 	}
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
@@ -339,7 +339,8 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 	public boolean isAutoStartup() {
 		try {
 			this.topology = getObject().build(this.properties);
-		} catch (Exception e) {
+		}
+		catch (Exception e) {
 			throw new RuntimeException(e);
 		}
 		return this.autoStartup;
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2023 the original author or authors.
+ * Copyright 2017-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -337,6 +337,11 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 
 	@Override
 	public boolean isAutoStartup() {
+		try {
+			this.topology = getObject().build(this.properties);
+		} catch (Exception e) {
+			throw new RuntimeException(e);
+		}
 		return this.autoStartup;
 	}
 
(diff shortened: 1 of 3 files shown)


sobychacko:
@chickenchickenlove There are checkstyle issues in the build. Please take a look.

chickenchickenlove:
Aye, i will check it. thanks for your comment 🙇‍♂️

chickenchickenlove:
Hi, @sobychacko !
Thanks for your comments 🙇‍♂️ 
I fixed failed test and checkstyle! 

Please take a look when you have some free time! 🙇‍♂️

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -337,6 +337,12 @@ protected StreamsBuilder createInstance() {
 
 	@Override
 	public boolean isAutoStartup() {
+		try {
+			this.topology = getObject().build(this.properties);
+		}
+		catch (Exception e) {
+			throw new RuntimeException(e);
+		}
```
sobychacko, line 345: It might be better to add this to the `afterPropertiesSet()` method instead of here. Doing it here adds more responsibilities to the `isAutoStartup()` method, which should technically only return the current `autoStartup` status. Also, we need to call the `configureTopology(topology)` method immediately after the topology creation so that some clients, such as the `TopologyTestDriver`, may use a fully configured topology.

sobychacko: Also, please add your name as an `author` of this class.

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
```diff
@@ -92,20 +92,56 @@ public void testCleanupStreams() throws IOException {
 	}
```
sobychacko, line 92: Can you add your name as an `author` of this class?

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryLateConfigTests.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2020 the original author or authors.
+ * Copyright 2018-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
```
sobychacko, line 5: Missing `author` tag.

sobychacko (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
chickenchickenlove: Aye, i added my name classes below.
1. `StreamsBuilderFactoryBean`
2. `StreamsBuilderFactoryBeanTests`
3. `StreamsBuilderFactoryLateConfigTests`
Please take a look 🙇‍♂️

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
sobychacko: Looks like this comment is not addressed? https://github.com/spring-projects/spring-kafka/pull/3172/files#r1546662939

sobychacko (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
chickenchickenlove: @sobychacko, I'm so sorry, i missed this comments. 

> It might be better to add this to the afterPropertiesSet() method instead of here. Doing it here adds more responsibilities to the isAutoStartup() method, which should technically only return the current autoStartup status. Also, we need to call the configureTopology(topology) method immediately after the topology creation so that some clients, such as the TopologyTestDriver, may use a fully configured topology.

IMHO, `afterPropertiesSet()` is not proper workaround. 
`streamBuilder#stream()` seems to be called  after `StreamsBuilderFactoryBean#afterPropertiesSet()` are executed.  

When `StreamsBuilderFactoryBean#afterPropertiesSet()` is called, no stream declarations are made in the `StreamBuilder` yet. Therefore, executing `StreamsBuilder.build()` at the time of the afterPropertiesSet() call does not create any topology.

chickenchickenlove: And i agree that `isAutoStartup()` has much responsibilities in this case.
If we move `configureTopology(topology)` to `isAutoStartup()`, much responsibilities as well. 

However, as i said above, we cannot build `streamsBuilder` properly in scope of `afterPropertiesSet()`. 

I have one more idea. Creating `StreamsBuilder`.
Remove logic to create`StreamsBuilder` and register `StreamsBuilder` as bean separately. But it must involve much changes.

chickenchickenlove: @sobychacko May I try writing a PR again in this direction? 
> Remove logic to createStreamsBuilder and register StreamsBuilder as bean separately.

Or do you have any other good ideas? Please let me know. 🙇‍♂️

artembilan: Just commented on your issue in Spring Framework.
Let's see if `SmartInitializingSingleton.afterSingletonsInstantiated()` implementation help us somehow!

chickenchickenlove: @artembilan Hi! 
Thanks your comment. `SmartInitializingSingleton.afterSingletonsInstantiated()` works well. 
I modified code to use `SmartInitializingSingleton.afterSingletonsInstantiated()` and test code as well. 

Please Take a look. 🙇‍♂️ 

CC. @sobychacko

sobychacko (approved):

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
```diff
@@ -102,12 +103,35 @@ protected StreamsBuilder createInstance() {
...
+		// Given
+		streamsBuilderFactoryBean = new StreamsBuilderFactoryBean(kafkaStreamsConfiguration) {
+			@Override
+			protected StreamsBuilder createInstance() {
+				return spy(super.createInstance());
+			}
+		};
+		streamsBuilderFactoryBean.afterPropertiesSet();
+		StreamsBuilder builder = streamsBuilderFactoryBean.getObject();
+		builder.stream(Pattern.compile("foo"));
```
sobychacko, line 124: Try not to use names like `foo` in the new code that we add.

sobychacko (changes requested):

sobychacko:
@chickenchickenlove The new updates look good, but the PR build still has some checkstyle issues. See more details here: https://github.com/spring-projects/spring-kafka/actions/runs/8532654028/job/23411166278?pr=3172.
Once that is addressed, this is ready to be merged.

chickenchickenlove:
@sobychacko 
Thanks for your time 🙇‍♂️ 
I resolved the lint error and change from `foo` to `test-topic`.
Please Could you take a look?

sobychacko:
@chickenchickenlove Thanks for the PR. It is now merged upstream.

injae-kim:
Nice work @chickenchickenlove !!

Issue #3176, opened 2024-04-02
Setting the Kafka Streams property `default.dsl.store` has no effect


cedric-schaller:
Between 2.9.x and 3.2.x (main)

**Describe the bug**

In Kafka Streams, all DSL operators use RocksDB stores by default. With [KIP-591](https://cwiki.apache.org/confluence/display/KAFKA/KIP-591%3A+Add+Kafka+Streams+config+to+set+default+state+store) (introduced with Kafka 3.2), a new Kafka Streams property called `default.dsl.store` was proposed, which allows to set the default built-in store type for the whole application. When using Spring for Apache Kafka and setting `default.dsl.store` to `in_memory`, this has no effect and RocksDB is still used for all the stores. 

I have already fixed this bug and I will create a pull request soon. 

**To Reproduce**

In order to reproduce the bug, one needs to set the property `default.dsl.store` to `in_memory`. Despite this setting, RocksDB files will still be created under the directory used as `state.dir`. 

A minimal reproducible sample is available under the chapter "Sample" of this bug report. 

**Expected behavior**

Spring for Apache Kafka must comply with the setting and use in-memory stores when asked to do so. 

As a side remark, [KIP-591](https://cwiki.apache.org/confluence/display/KAFKA/KIP-591%3A+Add+Kafka+Streams+config+to+set+default+state+store) has been deprecated with Kafka 3.7 in favor [KIP-954](https://cwiki.apache.org/confluence/display/KAFKA/KIP-954%3A+expand+default+DSL+store+configuration+to+custom+types). However, despite the setting `default.dsl.store` now being deprecated, it is still available and should therefore be supported until it is completely removed, especially in spring-kafka 2.9.x -> 3.1.x, where the mechanism introduced by [KIP-954](https://cwiki.apache.org/confluence/display/KAFKA/KIP-954%3A+expand+default+DSL+store+configuration+to+custom+types) is not yet available. 

It is also important to mention that [KIP-954](https://cwiki.apache.org/confluence/display/KAFKA/KIP-954%3A+expand+default+DSL+store+configuration+to+custom+types) is already properly supported by Spring for Apache Kafka and is in no way affected by the bug described here. 

**Sample**

A minimal reproducible sample can be found under https://github.com/cedric-schaller/default-dsl-store-bug. When building it, the test `DefaultDslStoreBugApplicationIT` will fail, indicating that a RocksDB state store is being used despite the instruction to use `in_memory` state stores by default. 

[...]

PR #3177, opened 2024-04-02
GH-3176: Take property default.dsl.store into account

cedric-schaller:
GH-3176: Setting the Kafka Streams property default.dsl.store has no effect

This is a proposal to fix GH-3176 (https://github.com/spring-projects/spring-kafka/issues/3176)

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
@@ -27,7 +27,9 @@ import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.streams.KafkaClientSupplier;
 import org.apache.kafka.streams.KafkaStreams;
 import org.apache.kafka.streams.StreamsBuilder;
+import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.Topology;
+import org.apache.kafka.streams.TopologyConfig;
 import org.apache.kafka.streams.errors.StreamsUncaughtExceptionHandler;
 import org.apache.kafka.streams.processor.StateRestoreListener;
 import org.apache.kafka.streams.processor.internals.DefaultKafkaClientSupplier;
@@ -58,6 +60,7 @@ import org.springframework.util.Assert;
  * @author Gary Russell
  * @author Julien Wittouck
  * @author Sanghyeok An
+ * @author Cédric Schaller
  *
  * @since 1.1.4
  */
@@ -328,7 +331,7 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 				Assert.state(this.properties != null,
 						"streams configuration properties must not be null");
 			}
-			StreamsBuilder builder = new StreamsBuilder();
+			StreamsBuilder builder = createStreamBuilder();
 			this.infrastructureCustomizer.configureBuilder(builder);
 			return builder;
 		}
@@ -439,6 +442,17 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 		}
 		catch (Exception e) {
 			throw new RuntimeException(e);
+		}
+	}
+
+	private StreamsBuilder createStreamBuilder() {
+		if (this.properties == null) {
+			return new StreamsBuilder();
+		}
+		else {
+			StreamsConfig streamsConfig = new StreamsConfig(this.properties);
+			TopologyConfig topologyConfig = new TopologyConfig(streamsConfig);
+			return new StreamsBuilder(topologyConfig);
 		}
 	}
 
--- a/spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanInMemoryStateStoreTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanInMemoryStateStoreTests.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.config;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.kafka.common.serialization.Serdes;
+import org.apache.kafka.streams.StreamsBuilder;
+import org.apache.kafka.streams.StreamsConfig;
+import org.apache.kafka.streams.kstream.KStream;
+import org.apache.kafka.streams.kstream.KTable;
+import org.apache.kafka.streams.kstream.Materialized;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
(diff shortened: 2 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -1,5 +1,5 @@
 /*
```
sobychacko, line 1: Need to add your name as an author.

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanInMemoryStateStoreTests.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright 2018-2024 the original author or authors.
```
sobychacko, line 2: Just 2024 will be enough.

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanInMemoryStateStoreTests.java
```diff
@@ -0,0 +1,96 @@
...
+	private static Path stateStoreDir;
+
+	@BeforeAll
+	static void beforeAll() throws IOException {
+		stateStoreDir = Files.createTempDirectory(StreamsBuilderFactoryBeanInMemoryStateStoreTests.class.getSimpleName());
+	}
+
+	@Test
+	void testStateStoreIsInMemory() {
+		assertThat(stateStoreDir).isEmptyDirectory();
```
sobychacko, line 64: Isn't this always true regardless of the change? Maybe you want to test whether some data was produced in the in-memory state store?

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanInMemoryStateStoreTests.java
```diff
@@ -0,0 +1,96 @@
...
+			props.put(StreamsConfig.STATE_DIR_CONFIG, stateStoreDir.toString());
+
+			// Property introduced with KIP-591 (Kafka 3.2) and deprecated (but still supported) with Kafka 3.7
+			props.put(StreamsConfig.DEFAULT_DSL_STORE_CONFIG, "in_memory");
+			return new KafkaStreamsConfiguration(props);
+		}
+
+		@Bean
+		public KTable<?, ?> table(StreamsBuilder builder) {
+			KStream<Object, Object> stream = builder.stream(Pattern.compile("foo"));
```
sobychacko, line 90: We are trying to avoid things like `foo` and `bar` in the project for new code.

sobychacko (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanInMemoryStateStoreTests.java
```diff
@@ -0,0 +1,96 @@
...
+		@Bean(name = KafkaStreamsDefaultConfiguration.DEFAULT_STREAMS_CONFIG_BEAN_NAME)
+		public KafkaStreamsConfiguration kStreamsConfigWithInMemoryStateStores() {
+			Map<String, Object> props = new HashMap<>();
+			props.put(StreamsConfig.APPLICATION_ID_CONFIG, "should-be-stored-in-memory");
+			props.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, this.brokerAddresses);
+			props.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.StringSerde.class);
+			props.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.StringSerde.class);
+			props.put(StreamsConfig.STATE_DIR_CONFIG, stateStoreDir.toString());
+
+			// Property introduced with KIP-591 (Kafka 3.2) and deprecated (but still supported) with Kafka 3.7
```
sobychacko, line 83: How will this work after the property is removed?

sobychacko (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanInMemoryStateStoreTests.java
cedric-schaller: I am more than happy to change this. I tried to aligned with the code I saw but I definitely prefer proper naming. Done.

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanInMemoryStateStoreTests.java
cedric-schaller: No, removing the change will lead to a test failure. 

By default, state stores (and therefore the `state.dir` when using RocksDB) are kept across restarts of the application so that the state does not have to be reconstructed from the Kafka topic. If you want to remove the state store directory when Kafka Streams is shutting down, you have to pass a `CleanupConfig` to the `StreamsBuilderFactoryBean` with `onStop` set to `true`, as it is the case e.g. in `StreamsBuilderFactoryBeanTests:120` for the test called `testCleanupStreams()`. 

At first, I tried to test the fact that we indeed had an in-memory store, but this involved using `ReflectionUtils`, which in my opinion was suboptimal. So checking the non-existence of a RocksDB directory felt like the cleanest way to do this. However, if you have an idea how to test that an in-memory store is used without using reflection, I am happy to modify this test.

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanInMemoryStateStoreTests.java
cedric-schaller: This would be a breaking change, and therefore I would assume it to be part of a major `kafka-streams` release. At that point in time, this code would not longer compile and the test would either need to be completely removed or (in my view the better option) the line 83 could be changed to set `StreamsConfig.DSL_STORE_SUPPLIERS_CLASS_CONFIG` instead. This property was introduced with [KIP-954](https://cwiki.apache.org/confluence/display/KAFKA/KIP-954%3A+expand+default+DSL+store+configuration+to+custom+types) as a replacement for `default.dsl.store`. 

For the `spring-kafka` releases 2.9.x, 3.0.x and 3.1.x, the Kafka Streams API will remain stable and therefore the property `default.dsl.store` will keep being supported.

sobychacko: Makes sense. Btw, we cannot backport the changes into `2.9.x`, as we are already out of the [OSS support timeline](https://spring.io/projects/spring-kafka#support) for that version.

sobychacko:
@cedric-schaller There are merge conflicts in the PR currently since we merged another change that modified the `StreamsBuilderFactoryBean.` Could you rebase the PR against the latest `main`? Thanks!

cedric-schaller:
@sobychacko I have now finalized the PR. Please let me know if you expect anything else from me.

sobychacko (approved):

sobychacko:
@cedric-schaller Thanks. I started a PR build; we can merge these changes once that is green.

cedric-schaller:
@sobychacko I would appreciate cherry-picking this fix to 3.0.x and 3.1.x as we use 3.1.x on our project and need this feature. Would you like me to create 2 PR for this, one for each branch?

artembilan:
@cedric-schaller , see the issue for this this PR and from there you can follow with connected issues.
And this one points that the fix has made it into a `3.1.x` branch: https://github.com/spring-projects/spring-kafka/issues/3181.

And we have releases today.

Hope that helps somehow.

cedric-schaller:
@artembilan I am sorry, I missed these commits. Thanks for answering, and I am looking forward to using the new release!

sobychacko:
@cedric-schaller We removed the property usage (`StreamsConfig.DEFAULT_DSL_STORE_CONFIG`) from the main branch test you added (`3.2.0-SNAPSHOT`) as this has caused some deprecation and other related compiler warnings in our builds. I verified the behavior using the strategies in KIP-954.  Please take a look at the related [commit](https://github.com/spring-projects/spring-kafka/commit/4a46f0d2c3f637d6791f9c7e863f7dd1c1d34363). In the `3.1.x` and `3.0.x` branches, the property usage is intact in the test. Thanks!

PR #3185, opened 2024-04-09
Update message-listener-container.adoc

cenkuyan-tomtom:
The PR suggests a correction to the wording , which clarifies the meaning.


sobychacko:
@cenkuyan-tomtom I think `as long as` is the correct usage here. Because we are saying that the commits are processed when all the records from the previous poll have been processed, as long as the ackTime since the last commit has been exceeded. By changing it to `as soon as,` we are implying that we will commit the offsets regardless of whether all the records are processed, which is not the case.

sobychacko:
I am closing now since the current semantics in wording look right.

PR #3212, opened 2024-04-23
Fix ackDiscarded.

chickenchickenlove:
### Background
- I found that the `descriptions` and `methods` related to `ackDiscarded` behave differently during investigating [issue](https://github.com/spring-projects/spring-kafka/issues/2806). 
https://github.com/spring-projects/spring-kafka/blob/1d0a7d305aec6e0976027267e80c3effd80e870c/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java#L335-L341
- From java docs, when `setRecordFilterStrategy()` in use, `ackDiscarded` should be `true`. however, `ackDiscarded` is always `false` even if some `BeanPostProcessor` call `setRecordFilterStrategy()` to use `RecordFilter`. 

### Changes
- create new method `useRecordFilterStrategy()` to bind `setRecordFilterStrategy()` and `setAckDiscarded()` to follow java docs.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
@@ -340,6 +340,11 @@ public abstract class AbstractKafkaListenerEndpoint<K, V>
 		this.ackDiscarded = ackDiscarded;
 	}
 
+	public void useRecordFilterStrategy(RecordFilterStrategy<? super K, ? super V> recordFilterStrategy) {
+		setRecordFilterStrategy(recordFilterStrategy);
+		setAckDiscarded(true);
+	}
+
 	@Nullable
 	@Override
 	public String getClientIdPrefix() {
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
@@ -62,6 +62,7 @@ import org.springframework.util.ObjectUtils;
  * @author Gary Russell
  * @author Artem Bilan
  * @author Wang Zhiyang
+ * @author Sanghyeok An
  *
  * @see MethodKafkaListenerEndpoint
  */
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
@@ -340,11 +340,6 @@ public abstract class AbstractKafkaListenerEndpoint<K, V>
 		this.ackDiscarded = ackDiscarded;
 	}
 
-	public void useRecordFilterStrategy(RecordFilterStrategy<? super K, ? super V> recordFilterStrategy) {
-		setRecordFilterStrategy(recordFilterStrategy);
-		setAckDiscarded(true);
-	}
-
 	@Nullable
 	@Override
 	public String getClientIdPrefix() {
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
```diff
@@ -340,6 +340,11 @@ public void setAckDiscarded(boolean ackDiscarded) {
 		this.ackDiscarded = ackDiscarded;
 	}
 
+	public void useRecordFilterStrategy(RecordFilterStrategy<? super K, ? super V> recordFilterStrategy) {
+		setRecordFilterStrategy(recordFilterStrategy);
+		setAckDiscarded(true);
```
artembilan, line 345: This new method does not make too much sense since we simply can do this `setAckDiscarded(true)` in the existing `setRecordFilterStrategy()`.
WDYT?

artembilan (changes requested):

sobychacko:
@chickenchickenlove, Can you add your name as an author to the classes and update any copyright years?

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
chickenchickenlove: I agree with you, i made new commit to follow your guideline. 
It is good point! 👍 

I was just worried that including `setAckDiscarded(true)` in `setFilterStrategy(...)` might bring unintended side effects. For example, a developer might not expect that `ackDiscarded` would change when updating `setFilterStrategy()`, since it's not specified in the method signature.

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
```diff
@@ -326,6 +326,7 @@ public void setReplyTemplate(KafkaTemplate<?, ?> replyTemplate) {
 	@SuppressWarnings("unchecked")
 	public void setRecordFilterStrategy(RecordFilterStrategy<? super K, ? super V> recordFilterStrategy) {
 		this.recordFilterStrategy = (RecordFilterStrategy<K, V>) recordFilterStrategy;
+		setAckDiscarded(true);
```
artembilan, line 329: Looking deeper, I don't think this fix is correct.
See more further:
```
	/**
	 * Create an instance with the supplied strategy and delegate listener.
	 * @param delegate the delegate.
	 * @param recordFilterStrategy the filter.
	 * @param ackDiscarded true to ack (commit offset for) discarded messages when the
	 * listener is configured for manual acks.
	 */
	public FilteringMessageListenerAdapter(MessageListener<K, V> delegate,
			RecordFilterStrategy<K, V> recordFilterStrategy, boolean ackDiscarded) {
```
Therefore it is up to end-user to decide if `ackDiscarded` or not.

So, what the problem we are trying to fix here, please?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
chickenchickenlove: I see 👍 

How about fix the java docs of `setAckDiscarded `'s `Set to true if the {@link #setRecordFilterStrategy(RecordFilterStrategy)} is in use.` ?

For example, `Set to true if the {@link #setRecordFilterStrategy(RecordFilterStrategy)} is in use and ack discarded messages`?

artembilan: How about:
```
Set to true if the {@link #setRecordFilterStrategy(RecordFilterStrategy)} should ack discarded messages.
```
?

chickenchickenlove: It's more suitable. (sorry i'm not good at english well🥲 ).
may i fix it?

chickenchickenlove: Thanks a lot 🙇‍♂️ 
I updated It!

artembilan (approved):
This is good to go. 
Do we have an opened issue to connect this PR into?

chickenchickenlove:
Thanks for you to review this 🙇‍♂️ 
> Do we have an opened issue to connect this PR into?

No, there is no opened issue! I found it while investigating other issue.

PR #3216, opened 2024-04-24
GH-2806 : Receiving an empty list when using RecordFilterStrategy on batch messages

chickenchickenlove:
### Motivation:
- https://github.com/spring-projects/spring-kafka/issues/2806
- In the current `batch` mode, even if the `RecordFilterStrategy` filters all records resulting in an `Empty List` being returned, the `KafkaListener` is still invoked. In contrast, in `single record` mode, if `record` are filtered, the `KafkaListener` is not called. This difference in behavior between the two modes can cause confusion for users.

### Modifications:
- Add public method `isAnyManualAck()` to `Acknowledgment` to verify that `manualAck` is needed on `FilteringBatchMessageListenerAdapter`. 
- Modify `FilteringBatchMessageListenerAdapter`.
  - add field `consumerAware` as `final` (IMHO, we don't need to calculate it every single call `onMessage()`.
  - add logic (if `empty list` and `manual Ack == true`, `KafkaListener` will be invoked. if `empty list` and `manual Ack == false`, `KafkaListener` will not be invoked even if `listener` is kind of `ConsumerAware`. In detail, See Discussion section below.)

### Result:
- Closes https://github.com/spring-projects/spring-kafka/issues/2806
- When the `RecordFilterStrategy` filters all records and returns an Empty List, the `KafkaListener` is invoked only if it is in manual `ACK` mode.

### Discussion
- When using a `ConsumerAware` Listener, commits can be made using `Consumer.commitSync()` and `Consumer.commitAsync()`. However, when using a `ConsumerAwareAckListener`, it seems possible that commits using the `Consumer` and commits using `Ack` could be processed simultaneously. That situation seems quite ambiguous.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
@@ -27,3 +27,62 @@ public void listen(Thing thing) {
 }
 ----
 
+### Ignore empty batch when you use `batch` mode.
+Starting with version `3.3.0`, we support ignoring empty batch that result from filtering by `RecordFilterStrategy`.
+When implementing `RecordFilterStrategy`, you can configure it through `ignoreEmptyBatch()`. default is `false`, that means `KafkaListener` will be invoked even if all `ConsumerRecord` are filtered out.
+
+If `true` is returned, the `KafkaListener` will not be invoked when all `ConsumerRecord` s are filtered out. However, commit to broker, will still be executed.
+If `false` is returned, the `KafkaListener` will be invoked when all `ConsumerRecord` s are filtered out.
+
+Let's look some examples.
+
+[source,java]
+----
+public class IgnoreEmptyBatchRecordFilterStrategy implements RecordFilterStrategy {
+    ...
+    @Override
+    public List<ConsumerRecord<String, String>> filterBatch(
+            List<ConsumerRecord<String, String>> consumerRecords) {
+        return List.of();
+    }
+
+    @Override
+    public boolean ignoreEmptyBatch() {
+        return true;
+    }
+};
+
+// NOTE: ignoreEmptyBatchRecordFilterStrategy is bean name of IgnoreEmptyBatchRecordFilterStrategy instance.
+@KafkaListener(id = "filtered", topics = "topic", filter = "ignoreEmptyBatchRecordFilterStrategy")
+public void listen(Thing thing) {
+    ...
+}
+----
+In this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `true` as result of `ignoreEmptyBatch()`.
+thus `KafkaListener#listen(...)` never will be invoked at all.
+
+
+[source,java]
+----
+public class NotIgnoreEmptyBatchRecordFilterStrategy implements RecordFilterStrategy {
+    ...
+    @Override
+    public List<ConsumerRecord<String, String>> filterBatch(
+            List<ConsumerRecord<String, String>> consumerRecords) {
+        return List.of();
+    }
+
+    @Override
+    public boolean ignoreEmptyBatch() {
+        return false;
+    }
+};
+
+// NOTE: notIgnoreEmptyBatchRecordFilterStrategy is bean name of NotIgnoreEmptyBatchRecordFilterStrategy instance.
+@KafkaListener(id = "filtered", topics = "topic", filter = "notIgnoreEmptyBatchRecordFilterStrategy")
+public void listen(Thing thing) {
+    ...
+}
+----
+However, in this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `false` as result of `ignoreEmptyBatch()`.
+thus `KafkaListener#listen(...)` always will be invoked.
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
@@ -28,7 +28,7 @@ public void listen(Thing thing) {
 ----
 
 [[x32-ignore-empty-batch-messages]]
-=== Ignore empty batch when you use `batch` mode.
+=== Ignore empty batch in `batch` mode.
 
 Starting with version 3.3, Ignoring empty batches that result from filtering by `RecordFilterStrategy` is supported. +
 When implementing `RecordFilterStrategy`, it can be configured through `ignoreEmptyBatch()`. +
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
@@ -89,4 +89,4 @@ public void listen(List<Thing> things) {
 }
 ----
 However, in this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `false` as result of `ignoreEmptyBatch()`. +
-thus `KafkaListener#listen(...)` always will be invoked.
+Thus `KafkaListener#listen(...)` always will be invoked.
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
@@ -27,16 +27,13 @@ public void listen(Thing thing) {
 }
 ----
 
-[[x32-ignore-empty-batch-messages]]
-=== Ignore empty batch in `batch` mode.
+Starting with version 3.3, Ignoring empty batches that result from filtering by `RecordFilterStrategy` is supported.
+When implementing `RecordFilterStrategy`, it can be configured through `ignoreEmptyBatch()`.
+The default setting is `false`, indicating `KafkaListener` will be invoked even if all `ConsumerRecord` s are filtered out.
 
-Starting with version 3.3, Ignoring empty batches that result from filtering by `RecordFilterStrategy` is supported. +
-When implementing `RecordFilterStrategy`, it can be configured through `ignoreEmptyBatch()`. +
-The default setting is `false`, indicating `KafkaListener` will be invoked even if all `ConsumerRecord` s are filtered out. +
-
-If `true` is returned, the `KafkaListener` [underline]#will not be invoked# when all `ConsumerRecord` are filtered out. +
+If `true` is returned, the `KafkaListener` [underline]#will not be invoked# when all `ConsumerRecord` are filtered out.
 However, commit to broker, will still be executed. +
-If `false` is returned, the `KafkaListener` [underline]#will be invoked# when all `ConsumerRecord` are filtered out. +
+If `false` is returned, the `KafkaListener` [underline]#will be invoked# when all `ConsumerRecord` are filtered out.
 
 Here are some examples.
 
@@ -62,9 +59,8 @@ public void listen(List<Thing> things) {
     ...
 }
 ----
-In this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `true` as result of `ignoreEmptyBatch()`. +
-Thus `KafkaListener#listen(...)` never will be invoked at all. +
-
+In this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `true` as result of `ignoreEmptyBatch()`.
+Thus `KafkaListener#listen(...)` never will be invoked at all.
 
 [source,java]
 ----
@@ -88,5 +84,5 @@ public void listen(List<Thing> things) {
     ...
 }
 ----
-However, in this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `false` as result of `ignoreEmptyBatch()`. +
+However, in this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `false` as result of `ignoreEmptyBatch()`.
 Thus `KafkaListener#listen(...)` always will be invoked.
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
@@ -27,14 +27,18 @@ public void listen(Thing thing) {
 }
 ----
 
-### Ignore empty batch when you use `batch` mode.
-Starting with version `3.3.0`, we support ignoring empty batch that result from filtering by `RecordFilterStrategy`.
-When implementing `RecordFilterStrategy`, you can configure it through `ignoreEmptyBatch()`. default is `false`, that means `KafkaListener` will be invoked even if all `ConsumerRecord` are filtered out.
+[[x32-ignore-empty-batch-messages]]
+=== Ignore empty batch when you use `batch` mode.
 
-If `true` is returned, the `KafkaListener` will not be invoked when all `ConsumerRecord` s are filtered out. However, commit to broker, will still be executed.
-If `false` is returned, the `KafkaListener` will be invoked when all `ConsumerRecord` s are filtered out.
+Starting with version 3.3, Ignoring empty batches that result from filtering by `RecordFilterStrategy` is supported. +
+When implementing `RecordFilterStrategy`, it can be configured through `ignoreEmptyBatch()`. +
+The default setting is `false`, indicating `KafkaListener` will be invoked even if all `ConsumerRecord` s are filtered out. +
 
-Let's look some examples.
+If `true` is returned, the `KafkaListener` [underline]#will not be invoked# when all `ConsumerRecord` are filtered out. +
+However, commit to broker, will still be executed. +
+If `false` is returned, the `KafkaListener` [underline]#will be invoked# when all `ConsumerRecord` are filtered out. +
+
+Here are some examples.
 
 [source,java]
 ----
@@ -54,12 +58,12 @@ public class IgnoreEmptyBatchRecordFilterStrategy implements RecordFilterStrateg
 
 // NOTE: ignoreEmptyBatchRecordFilterStrategy is bean name of IgnoreEmptyBatchRecordFilterStrategy instance.
 @KafkaListener(id = "filtered", topics = "topic", filter = "ignoreEmptyBatchRecordFilterStrategy")
-public void listen(Thing thing) {
+public void listen(List<Thing> things) {
     ...
 }
 ----
-In this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `true` as result of `ignoreEmptyBatch()`.
-thus `KafkaListener#listen(...)` never will be invoked at all.
+In this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `true` as result of `ignoreEmptyBatch()`. +
+Thus `KafkaListener#listen(...)` never will be invoked at all. +
 
 
 [source,java]
@@ -80,9 +84,9 @@ public class NotIgnoreEmptyBatchRecordFilterStrategy implements RecordFilterStra
 
 // NOTE: notIgnoreEmptyBatchRecordFilterStrategy is bean name of NotIgnoreEmptyBatchRecordFilterStrategy instance.
 @KafkaListener(id = "filtered", topics = "topic", filter = "notIgnoreEmptyBatchRecordFilterStrategy")
-public void listen(Thing thing) {
+public void listen(List<Thing> things) {
     ...
 }
 ----
-However, in this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `false` as result of `ignoreEmptyBatch()`.
+However, in this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `false` as result of `ignoreEmptyBatch()`. +
 thus `KafkaListener#listen(...)` always will be invoked.
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -3396,16 +3396,28 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			private final ConsumerRecords<K, V> records;
 
 			private final List<ConsumerRecord<K, V>> recordList;
+			private boolean isAnyManualAck = false;
 
 			private volatile boolean acked;
 
 			private volatile int partial = -1;
 
 			ConsumerBatchAcknowledgment(ConsumerRecords<K, V> records,
-					@Nullable List<ConsumerRecord<K, V>> recordList) {
-
+										@Nullable List<ConsumerRecord<K, V>> recordList) {
 				this.records = records;
 				this.recordList = recordList;
+			}
+
+			ConsumerBatchAcknowledgment(ConsumerRecords<K, V> records,
+										@Nullable List<ConsumerRecord<K, V>> recordList,
+										boolean isAnyManualAck) {
+				this(records, recordList);
+				this.isAnyManualAck = isAnyManualAck;
+			}
+
+			@Override
+			public boolean isAnyManualAck() {
+				return this.isAnyManualAck;
 			}
 
 			@Override
@@ -3416,7 +3428,8 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 				}
 				if (!this.acked) {
 					Map<TopicPartition, List<Long>> offs = ListenerConsumer.this.offsetsInThisBatch;
-					Map<TopicPartition, List<ConsumerRecord<K, V>>> deferred = ListenerConsumer.this.deferredOffsets;
+					Map<TopicPartition, List<ConsumerRecord<K, V>>> deferred =
+							ListenerConsumer.this.deferredOffsets;
 					for (TopicPartition topicPartition : this.records.partitions()) {
 						if (offs != null) {
 							offs.remove(topicPartition);
@@ -3431,22 +3444,23 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			@Override
 			public void acknowledge(int index) {
 				Assert.isTrue(index > this.partial,
-						() -> String.format("index (%d) must be greater than the previous partial commit (%d)", index,
-								this.partial));
+							  () -> String.format(
+									  "index (%d) must be greater than the previous partial commit (%d)", index,
+									  this.partial));
 				Assert.state(ListenerConsumer.this.isManualImmediateAck,
-						"Partial batch acknowledgment is only supported with AckMode.MANUAL_IMMEDIATE");
+							 "Partial batch acknowledgment is only supported with AckMode.MANUAL_IMMEDIATE");
 				Assert.state(this.recordList != null,
-						"Listener must receive a List of records to use partial batch acknowledgment");
+							 "Listener must receive a List of records to use partial batch acknowledgment");
 				Assert.isTrue(index >= 0 && index < this.recordList.size(),
-						() -> String.format("index (%d) is out of range (%d-%d)", index, 0,
-								this.recordList.size() - 1));
+							  () -> String.format("index (%d) is out of range (%d-%d)", index, 0,
+												  this.recordList.size() - 1));
 				Assert.state(Thread.currentThread().equals(ListenerConsumer.this.consumerThread),
-						"Partial batch acknowledgment is only supported on the consumer thread");
+							 "Partial batch acknowledgment is only supported on the consumer thread");
 				Map<TopicPartition, List<ConsumerRecord<K, V>>> offsetsToCommit = new LinkedHashMap<>();
 				for (int i = this.partial + 1; i <= index; i++) {
 					ConsumerRecord<K, V> record = this.recordList.get(i);
 					offsetsToCommit.computeIfAbsent(new TopicPartition(record.topic(), record.partition()),
-							tp -> new ArrayList<>()).add(record);
+													tp -> new ArrayList<>()).add(record);
 				}
 				if (!offsetsToCommit.isEmpty()) {
 					processAcks(new ConsumerRecords<>(offsetsToCommit));
@@ -3457,9 +3471,9 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			@Override
 			public void nack(int index, Duration sleep) {
 				Assert.state(Thread.currentThread().equals(ListenerConsumer.this.consumerThread),
-						"nack() can only be called on the consumer thread");
+							 "nack() can only be called on the consumer thread");
 				Assert.state(!ListenerConsumer.this.asyncReplies,
-						"nack() is not supported with out-of-order commits");
+							 "nack() is not supported with out-of-order commits");
 				Assert.isTrue(!sleep.isNegative(), "sleep cannot be negative");
 				Assert.isTrue(index >= 0 && index < this.records.count(), "index out of bounds");
 				ListenerConsumer.this.nackIndex = index;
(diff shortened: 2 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
```diff
@@ -85,4 +85,7 @@ default boolean isOutOfOrderCommit() {
 		return false;
 	}
 
+	default boolean isAnyManualAck() {
```
artembilan, line 88: I don't think this method makes sense since `Acknowledgment` abstraction is there exactly for the manual acks.
See the logic in the `doInvokeOnMessage()`.
What is your exact motivation do introduce this option over here?
Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3463,15 +3477,14 @@ public void nack(int index, Duration sleep) {
...
-					else {
+					} else {
 						break;
 					}
 				}
 				Map<TopicPartition, List<ConsumerRecord<K, V>>> newRecords = new HashMap<>();
 				for (ConsumerRecord<K, V> cRecord : toAck) {
 					newRecords.computeIfAbsent(new TopicPartition(cRecord.topic(), cRecord.partition()),
-							tp -> new LinkedList<>()).add(cRecord);
+											   tp -> new LinkedList<>()).add(cRecord);
```
artembilan, line 3487: I think if we get rid of that `Acknowledgment.isAnyManualAck()` all the refactoring changes in the class are void.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/Acknowledgment.java
chickenchickenlove: As you said, `Acknowledgment abstraction is there exactly for the manual acks` is true! 
i reverted this method `isAnyManualAck()`.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: Thank you for confirming that we are on the same page about `Acknowledgment`.
Would you mind to revert changes in this class since they are just code style refactoring and apparently not the one we are following. Especially that `} else {` block.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
chickenchickenlove: No, it's not true. 
If `ack != null` is true, it's equivalent to `anyManualAck`, so the refactoring is still valid.

Previously, even if the filter resulted in an empty list, the listener was always called. With this PR, the listener is now only called when a Manual Ack is required. since consumerAware was always true, the listener was called even in cases that did not require manual Ack. In this case, regardless of `consumerAware == true`, if it is not a manual Ack and the list is empty, the listener will not be called.

```java
// FilteringBatchMessageListenerAdapater#onMessage() 
...
if (consumerRecords.isEmpty()) {
			if (acknowledgment != null) {
				invokeDelegate(consumerRecords, acknowledgment, consumer);
			}
		}
		else {
			if (this.consumerAware
				|| (!this.ackDiscarded && this.delegateType.equals(ListenerType.ACKNOWLEDGING))) {
				invokeDelegate(consumerRecords, acknowledgment, consumer);
			}
			else {
				if (this.ackDiscarded && acknowledgment != null) {
					acknowledgment.acknowledge();
				}
			}
		}
...
```

chickenchickenlove:
I see! i reverted all and make new commits.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchMessageListenerAdapter.java
```diff
@@ -54,6 +56,7 @@ public FilteringBatchMessageListenerAdapter(BatchMessageListener<K, V> delegate,
 
 		super(delegate, recordFilterStrategy);
 		this.ackDiscarded = false;
+		this.consumerAware = this.delegateType.equals(ListenerType.ACKNOWLEDGING_CONSUMER_AWARE) || this.delegateType.equals(ListenerType.CONSUMER_AWARE);
```
artembilan, line 59: Can we optimize the code here a little bit and call other constructor:
```
this(delegate, recordFilterStrategy, false);
```
?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchMessageListenerAdapter.java
```diff
@@ -79,19 +83,21 @@ public void onMessage(List<ConsumerRecord<K, V>> records, @Nullable Acknowledgme
...
 		else {
-			if (this.ackDiscarded && acknowledgment != null) {
-				acknowledgment.acknowledge();
+			if (this.consumerAware
+				|| (!this.ackDiscarded && this.delegateType.equals(ListenerType.ACKNOWLEDGING))) {
+				invokeDelegate(consumerRecords, acknowledgment, consumer);
+			}
+			else {
+				if (this.ackDiscarded && acknowledgment != null) {
+					acknowledgment.acknowledge();
```
artembilan, line 99: I don't think this is correct.
We are in this `else` branch when `!consumerRecords.isEmpty()`, therefore in my understanding we always have to call `invokeDelegate()`.
It sounds more like we are looking for something like this:
```
	if (!consumerRecords.isEmpty() || acknowledgment != null || consumerAware
				|| (!this.ackDiscarded && this.delegateType.equals(ListenerType.ACKNOWLEDGING))) {
			invokeDelegate(consumerRecords, acknowledgment, consumer);
```

However I believe that `this.delegateType.equals(ListenerType.ACKNOWLEDGING)` covers `acknowledgment != null`.

I probably need to look close what is the request in that issue.

artembilan (changes requested):

artembilan:
I think the problem comes from the `MethodKafkaListenerEndpoint.createMessageListenerInstance()`:
```
		if (isBatchListener()) {
			BatchMessagingMessageListenerAdapter<K, V> messageListener = new BatchMessagingMessageListenerAdapter<>(
					this.bean, this.method, this.errorHandler);
```
where we have:
```
public class BatchMessagingMessageListenerAdapter<K, V> extends MessagingMessageListenerAdapter<K, V>
		implements BatchAcknowledgingConsumerAwareMessageListener<K, V> {
```
and that leads to the:
```
		if (listener instanceof AcknowledgingConsumerAwareMessageListener
				|| listener instanceof BatchAcknowledgingConsumerAwareMessageListener) {
			listenerType = ListenerType.ACKNOWLEDGING_CONSUMER_AWARE;
		}
```
So, the logic in that `FilteringBatchMessageListenerAdapter` always falls to the `consumerAware` as `true`.
Therefore sounds like we cannot achieve the requested logic with existing flags.
Not sure, though, if that would be convenient to introduce a new one exactly for this use-case for batch filtering.

Maybe `RecordFilterStrategy` could be improved with extra `boolean` method to implement?
Like:
```
default boolean ignoreEmptyBatch() {
    return false;
}
```

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchMessageListenerAdapter.java
chickenchickenlove: > However I believe that this.delegateType.equals(ListenerType.ACKNOWLEDGING) covers acknowledgment != null.

IMHO, Thus, `acknowledgment != null` has more narrow than `this.delegateType.equals(ListenerType.ACKNOWLEDGING)`. 
`delegateType` is determined when spring context started with this codes.
https://github.com/spring-projects/spring-kafka/blob/5962e756814df74222368c73dbde6fbd578f9e9e/spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerUtils.java#L54-L76

However, `Acknowledgment` instance is created only `isAnyManualAck == true`. 

https://github.com/spring-projects/spring-kafka/blob/5962e756814df74222368c73dbde6fbd578f9e9e/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2380-L2403

`isAnyManualAck` is determined on `Constructor`.
`isAnyManualAck` consider only `AckMode.MANUAL` and `AckMode.MANUAL_IMMEDIATE`.

https://github.com/spring-projects/spring-kafka/blob/5962e756814df74222368c73dbde6fbd578f9e9e/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L853-L855

Thus, `acknowledgment != null` has more narrow than `this.delegateType.equals(ListenerType.ACKNOWLEDGING)`.

chickenchickenlove: Why we should have to call `invokeDelegate()` always? 
It loos like that `KafkaMessageListener` will ack batch eventually when(`AckMode != MANUAL_IMMEDIATE`) and (`AckMODE != MANUAL`) and (`!autoCommit`) condition is satisfied,

https://github.com/spring-projects/spring-kafka/blob/5962e756814df74222368c73dbde6fbd578f9e9e/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2301-L2322

IMHO, From codes above, If `consumerRecords.isEmpty()` is `true`, we don't need to invoke listener always. 
However, Condition which you said (it looks like that same with current code on main branch, `!consumerRecords.isEmpty()` equals `consumerRecords.size() > 0 `, right?), yes, we have to call `invoke` always.
```java
if (!consumerRecords.isEmpty() || acknowledgment != null || consumerAware || (!this.ackDiscarded && this.delegateType.equals(ListenerType.ACKNOWLEDGING))) { invokeDelegate(consumerRecords, acknowledgment, consumer);
```

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchMessageListenerAdapter.java
chickenchickenlove: Of course. I update it!

chickenchickenlove:
Thank you for your analysis a lot 🙇‍♂️🙇‍♂️🙇‍♂️🙇‍♂️!
The logic remains the same eventually,  but I think the direction you proposed is better because it gives `users` a choice.

chickenchickenlove:
```java
	default boolean ignoreEmptyBatch() {
		return false;
	}
```

I added `ignoreEmptyBatch()` to `RecordFilterStrategy` interface, and use it on `FilteringBatchMessageListenerAdapter` instead of `consumerRecords.isEmpty()`. 

This way, those who want to be as it is can do so without modifying their codes. Meanwhile, it provides a choice for users who have considered this to be an issue until now.

What do you think?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchMessageListenerAdapter.java
```diff
@@ -71,27 +71,31 @@ public FilteringBatchMessageListenerAdapter(BatchMessageListener<K, V> delegate,
...
-		 *  An empty list goes to the listener if ackDiscarded is false and the listener can ack
-		 *  either through the acknowledgment
-		 */
-		if (consumerRecords.size() > 0 || consumerAware
-				|| (!this.ackDiscarded && this.delegateType.equals(ListenerType.ACKNOWLEDGING))) {
-			invokeDelegate(consumerRecords, acknowledgment, consumer);
+
+		if (recordFilterStrategy.ignoreEmptyBatch()) {
+			if (acknowledgment != null) {
+				invokeDelegate(consumerRecords, acknowledgment, consumer);
```
artembilan, line 87: I don't think this logic is correct. I believe the request was about an option to fully ignore an empty batch as it is right now with a `FilteringMessageListenerAdapter` independently of the manual ack settings.
Therefore this your change is a bit off: looks like you always are going to push an the batch into listener independently of its state. 
My idea with this `ignoreEmptyBatch` was exactly opposite: if it is `true`, we don't call delegate and call `acknowledgment.acknowledge()` if it is present. Otherwise you are right: the container is going to take care about committing offset properly for the whole batch.
We call `invokeDelegate()` in all those cases we talked about and when `ignoreEmptyBatch` is `false`.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
```diff
@@ -58,4 +58,8 @@ default List<ConsumerRecord<K, V>> filterBatch(List<ConsumerRecord<K, V>> record
 		return records;
 	}
 
+	default boolean ignoreEmptyBatch() {
```
artembilan, line 61: For any new public API we need Javadoc, plus `@since`

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
chickenchickenlove: Thanks for your comment 🙇‍♂️ 
I added java docs for `ignoreEmptyBatch()`!

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchMessageListenerAdapter.java
chickenchickenlove: >  I believe the request was about an option to fully ignore an empty batch as it is right now with a FilteringMessageListenerAdapter independently of the manual ack settings.

I read the request again, and your point is correct. Yes, i make wrong PR. I'm sorry for that.  😓 
I added this logic! 
```java
// I added this.
if (recordFilterStrategy.ignoreEmptyBatch() &&
			consumerRecords.isEmpty() &&
			acknowledgment != null) {
			acknowledgment.acknowledge();
		}
else if (consumerRecords.size() > 0 || this.consumerAware
             || (!this.ackDiscarded && this.delegateType.equals(ListenerType.ACKNOWLEDGING))) {
	     invokeDelegate(consumerRecords, acknowledgment, consumer);
}
...
```
I expect that we are able to meet those requirements with this logic. 
What do you think?

artembilan: Yes, that sounds good. If batch is empty and manual ack, then check for `ignoreEmptyBatch()` and don't call delegate.
I think we all are now on the same page.
👍

chickenchickenlove: Yeah, i think so too 👍 
Is there anything else I need to work on? 

from [comment](https://github.com/spring-projects/spring-kafka/issues/2806#issuecomment-2069432409), 
After waiting for the `3.3.x` version to start, the next step for this PR, looks like it will start 🤔.

artembilan (changes requested):
We need some tests for this new feature. And, yes, I also think that this should go to `3.3` already . So, please, fix Javadoc respectively. And add some doc, too.

chickenchickenlove:
@artembilan , thanks for your comments 🙇‍♂️ 
I added a couple of test cases to test new public API.
When you have free time, please take a look 🙇‍♂️

chickenchickenlove:
I added `spring-kafka-docs` as well to `Filtering Messages` section.
image

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
```diff
@@ -27,3 +27,62 @@ public void listen(Thing thing) {
 }
 ----
 
+### Ignore empty batch when you use `batch` mode.
+Starting with version `3.3.0`, we support ignoring empty batch that result from filtering by `RecordFilterStrategy`.
+When implementing `RecordFilterStrategy`, you can configure it through `ignoreEmptyBatch()`. default is `false`, that means `KafkaListener` will be invoked even if all `ConsumerRecord` are filtered out.
```
artembilan, line 32: One sentence per line rule.
Plus it is not clear why you start a new sentence with lower case.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
```diff
@@ -27,3 +27,62 @@ public void listen(Thing thing) {
...
+
+If `true` is returned, the `KafkaListener` will not be invoked when all `ConsumerRecord` s are filtered out. However, commit to broker, will still be executed.
+If `false` is returned, the `KafkaListener` will be invoked when all `ConsumerRecord` s are filtered out.
+
+Let's look some examples.
```
artembilan, line 37: `Let's` is also not a technical documentation word 😄

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
```diff
@@ -27,3 +27,62 @@ public void listen(Thing thing) {
...
+};
+
+// NOTE: notIgnoreEmptyBatchRecordFilterStrategy is bean name of NotIgnoreEmptyBatchRecordFilterStrategy instance.
+@KafkaListener(id = "filtered", topics = "topic", filter = "notIgnoreEmptyBatchRecordFilterStrategy")
+public void listen(Thing thing) {
+    ...
+}
+----
+However, in this case, `IgnoreEmptyBatchRecordFilterStrategy` always returns empty list and return `false` as result of `ignoreEmptyBatch()`.
+thus `KafkaListener#listen(...)` always will be invoked.
```
artembilan, line 88: Lower case in the beginning of the sentence again.

If you wish, we will clean up all of these docs issue on merge.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchMessageListenerAdapter.java
```diff
@@ -71,22 +72,24 @@ public FilteringBatchMessageListenerAdapter(BatchMessageListener<K, V> delegate,
 
 		super(delegate, recordFilterStrategy);
 		this.ackDiscarded = ackDiscarded;
+		this.consumerAware = this.delegateType.equals(ListenerType.ACKNOWLEDGING_CONSUMER_AWARE) || this.delegateType.equals(ListenerType.CONSUMER_AWARE);
```
artembilan, line 75: The code line  length `! > 120` symbols.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
```diff
@@ -58,4 +60,14 @@ default List<ConsumerRecord<K, V>> filterBatch(List<ConsumerRecord<K, V>> record
 		return records;
 	}
 
+	/**
+	 * Returns a boolean value that determines whether {@link FilteringBatchMessageListenerAdapter} invoke the {@link BatchMessageListener} when all {@link ConsumerRecord}
```
artembilan, line 64: Method Javadocs must be imperative, like a command what yo do.
The length of the like not more than `120`.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
```diff
@@ -58,4 +60,14 @@ default List<ConsumerRecord<K, V>> filterBatch(List<ConsumerRecord<K, V>> record
...
+	 * have been filtered and return an EmptyList. The default is not to invoke the {@link BatchMessageListener} (false).
```
artembilan, line 65: What is `EmptyList`?
This is not a word, I believe better to say `empty list`.
I mean this is an official documentation of Spring: it has to be as professional as possible.
I also believe that default `false` means to call delegate with an empty list.

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/FilteringAdapterTests.java
```diff
@@ -72,4 +75,173 @@ public void testBatchFilterAckDiscard() throws Exception {
...
+			@Override
+			public List<ConsumerRecord<String, String>> filterBatch(
+					List<ConsumerRecord<String, String>> consumerRecords) {
+				// SUT
+				return List.of();
+			}
+
+			@Override
+			public boolean ignoreEmptyBatch() {
+				// SUT
```
artembilan, line 96: What is `SUT`, please?

I'm not sure that we need obvious comments in tests at all.

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/FilteringAdapterTests.java
```diff
@@ -72,4 +75,173 @@ public void testBatchFilterAckDiscard() throws Exception {
...
+				return false;
```
artembilan, line 180: Not sure why we need this method impl since its `default` variant in the interface does exactly same `false` logic.

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/FilteringAdapterTests.java
```diff
@@ -72,4 +75,173 @@ public void testBatchFilterAckDiscard() throws Exception {
...
+			public List<ConsumerRecord<String, String>> filterBatch(
+					// SUT
+					List<ConsumerRecord<String, String>> consumerRecords) {
+				return consumerRecords;
+			}
+
+			@Override
+			public boolean ignoreEmptyBatch() {
+				// SUT
+				return false;
```
artembilan, line 224: DITTO

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/FilteringAdapterTests.java
```diff
@@ -72,4 +75,173 @@ public void testBatchFilterAckDiscard() throws Exception {
...
+			return null;
+		}).given(listener).onMessage(any(List.class), any(Acknowledgment.class));
+
+		// When :
+		adapter.onMessage(consumerRecords, ack, null);
+
+		// Then
+		assertThat(latch.await(10, TimeUnit.SECONDS)).isTrue();
+		verify(ack, never()).acknowledge();
+	}
```
artembilan, line 246: Every member in the class, including last method of the last inner class, has to be surrounded with blank lines.

artembilan (changes requested):
The review is a bit thorough, but there is no rush with the fixes.
We have like a month yet until with start a new `3.3` version.

Thanks

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/FilteringAdapterTests.java
chickenchickenlove: I see! i fixed it 🙇‍♂️

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/FilteringAdapterTests.java
chickenchickenlove: Yes, you are right.  👍 
My intention was to explicitly show the test environment. 
But, it is better to do use `default method` to clarify the intent of the test. 

I fixed it.

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/FilteringAdapterTests.java
chickenchickenlove: Fixed it!

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/FilteringAdapterTests.java
chickenchickenlove: I agree! 
I removed all of them from test.

FYI, `SUT` is `System Under Test`.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringBatchMessageListenerAdapter.java
chickenchickenlove: Thanks for your comment 🙇‍♂️ 
I checked all the code and fixed it! 

I think it would be better to include this rule to `checkStyle`. I will make new PR for this.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
chickenchickenlove: Thanks for your comment.  🙇‍♂️ 
I've revised it to be more imperative, referencing other Javadocs. 
Could you take a look and see if it's okay?

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/FilteringAdapterTests.java
artembilan: Sure! Another request from me. No reason to say `fixed`, `done` etc. if you are agreed with me and will push that in the next commit or so. It is totally fine to argue if there is a disagreement or I am wrong. The point is that I have many other PRs to review or so and such a `Fixed` email is unnecessary distraction . 😅

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
chickenchickenlove: > I also believe that default false means to call delegate with an empty list.
Sorry, i made a mistake 😓😓😓.. I fixed it! 

DITTO, `EmptyList`!

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
```diff
@@ -27,3 +27,62 @@ public void listen(Thing thing) {
...
+
+    @Override
+    public boolean ignoreEmptyBatch() {
+        return false;
+    }
+};
+
+// NOTE: notIgnoreEmptyBatchRecordFilterStrategy is bean name of NotIgnoreEmptyBatchRecordFilterStrategy instance.
+@KafkaListener(id = "filtered", topics = "topic", filter = "notIgnoreEmptyBatchRecordFilterStrategy")
+public void listen(Thing thing) {
```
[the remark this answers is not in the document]
chickenchickenlove: Good point! 
I fixed it. 👍

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
```diff
@@ -27,3 +27,62 @@ public void listen(Thing thing) {
 }
 ----
 
+### Ignore empty batch when you use `batch` mode.
```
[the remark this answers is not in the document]
chickenchickenlove: Yes, i want to emphasize as a title. 
I emphasized it as a title using ASCIIDoc syntax! 
Thanks 🙇‍♂️

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
```diff
@@ -27,3 +27,62 @@ public void listen(Thing thing) {
...
+Starting with version `3.3.0`, we support ignoring empty batch that result from filtering by `RecordFilterStrategy`.
```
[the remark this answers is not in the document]
chickenchickenlove: I fixed it all! 
Thanks your comment 🙇‍♂️

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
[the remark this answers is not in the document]
artembilan: Ok. However I don’t think this new option deserves the whole new title: it is still a part of filtering algorithm. So, just a new paragraph with those couple samples and that’s it.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
chickenchickenlove: I see. 
I added `+` at end of the sentence to make new line!

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
chickenchickenlove: I see!! 😄😄

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
chickenchickenlove: I fixed it! 

> If you wish, we will clean up all of these docs issue on merge.


If the reviewers are okay with it, I'll go ahead and finalize it.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/filtering.adoc
artembilan: No, it had to be a new line in the code, not final text. See ASCIIDoc manual: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

spring-kafka/src/test/java/org/springframework/kafka/listener/adapter/FilteringAdapterTests.java
chickenchickenlove: Oh, I'm really sorry. 😓 
I thought I had to do it that way...! 
Keep in my mind! i will left comment if it is needed. 

Sorry to bother you again 😅

artembilan (changes requested):
It looks like your changes have slipped somehow from our sight.
@chickenchickenlove ,
let us know if there is anything else to do in this PR.

I can only suggest to rebase your branch to the latest `main` and add a short section into the `whats-new.adoc` with a link to your comprehensive explanation in the `filtering.adoc`.

The first milestone for `3.3` is due next Monday.

Thank you!

chickenchickenlove:
Hi, @artembilan ! long time no see 😄. 
Thanks for notifying me! 
I followed your comments about doing rebase and add description!
Please take another look, when you have free time 🙇‍♂️

artembilan (approved):
Pulling your changes locally for final review and merge...

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/f91f8a918688ccf855207aa66ef5703da0568770.

@chickenchickenlove ,

thank you very much for the contribution (again)!

A couple notes: 
 - it is better to use `@SuppressWarnings("unchecked")` if you cannot fix cast warnings in the tests.
 - Please, consider to use in the future an issue number for the branch to PR eventually, e.g. this one could be `GH-2806`. It is much easier to handle that on merge then
 - The `mock()` is there (without `Class` arg) for those to mitigate generics warning and easier to code.
 - The `final` is not necessary in most cases for local variables. Java makes them "effectively final" anyway.

artembilan:
Well, merged, so closing

chickenchickenlove:
Thanks for your time and looking this PR! 
Also, i will keep your advice in my mind and will do better next contribution 👍 
Thanks a lot, again 🙇‍♂️

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
```diff
@@ -58,4 +60,16 @@ default List<ConsumerRecord<K, V>> filterBatch(List<ConsumerRecord<K, V>> record
 		return records;
 	}
 
+	/**
+	 * Determine whether {@link FilteringBatchMessageListenerAdapter} should invoke
+	 * the {@link BatchMessageListener} when all {@link ConsumerRecord}s in a batch have been filtered out
+	 * resulting in empty list. By default, do invoke the {@link BatchMessageListener} (return false).
+	 * @return true for {@link FilteringBatchMessageListenerAdapter} to {@link BatchMessageListener}
```
injae-kim, line 67:
```
{@link FilteringBatchMessageListenerAdapter} to {@link BatchMessageListener}
-> {@link FilteringBatchMessageListenerAdapter} to not invoke {@link BatchMessageListener}
```
`not invoke` is correct doc?

artembilan: Fixed: https://github.com/spring-projects/spring-kafka/commit/5c3412204517303aacdbabdbb1c2b62866d3d9f2.
Thank you!

PR #3220, opened 2024-04-26
DRAFT : Add checkstyle: max line length 119.

chickenchickenlove:
### Background
- Recently, i have learned that `spring-kafka` project has a line length convention. It would be better to include that rules to `checkStyle`.  (The code line length ! > 120 symbols.)

### To Reviewer
After adding new rule to `checkStyle`, `checkStyle` throw a lot of `style error`.  If you guys want to introduce this rule to `checkStyle`, i'm willing to fix all of `style error`. 
However, you guys don't want it, i will close this PR. 

image

### DIFF
--- a/src/checkstyle/checkstyle.xml
+++ b/src/checkstyle/checkstyle.xml
@@ -8,6 +8,11 @@
 		<property name="file" value="${config_loc}/checkstyle-suppressions.xml"/>
 	</module>
 
+	<module name="LineLength">
+		<property name="max" value="119"/>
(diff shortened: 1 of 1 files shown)


artembilan (changes requested):
Thank you for the effort , but we cannot enforce such a rule. Sometimes it is just impossible to break at 120. It is just recommendation to make code more readable and easy to review. Remember this is an Open Source, so perfectly looking code is our goal. 

My vote is to reject this change.

Thanks for understanding!

Issue #3145, opened 2024-03-19
Add test for next generation consumer-group rebalance protocol


sobychacko:
Starting with `3.7.0`, the Kafka client now supports a new consumer group rebalance protocol and categorizes the consumer-group strategy as legacy (classic) vs. consumer. Spring for Apache Kafka supports the new protocol transparently; however, we need to verify this by adding an integration test. See more details [here](https://cwiki.apache.org/confluence/display/KAFKA/The+Next+Generation+of+the+Consumer+Rebalance+Protocol+%28KIP-848%29+-+Early+Access+Release+Notes).

chickenchickenlove:
I am investigating whether I can resolve this issue on my own. However, it seems currently unachievable due to the limitations of `KafkaTestKit`, which is used to create `EmbeddedKafka`.

I have created an issue with the `apache/Kafka` Team to see if there are any other workarounds.
I will let you know if there are any updates. 😅

sobychacko:
@chickenchickenlove This is an issue we need to tackle before the GA release, so we were planning to take a look at this, glad that you started on it. Please let us know if disabling `kraft` as @artembilan suggested works.

chickenchickenlove:
@sobychacko, @artembilan thanks for you guys comments. 
I didn't know about `@EmbeddedKafka(kraft = false)`. i'm going to try to test it ASAP.

This is [issue](https://issues.apache.org/jira/projects/KAFKA/issues/KAFKA-16648?filter=reportedbyme) which i reported. 
I tested KIP-848 rebalancing with this [code](https://github.com/chickenchickenlove/kraft-test). (docker-compose and simple main class). and it did work well. 

However, i have problem when i used KafkaTestKit with @EmbeddedKafka. 
I think [here](https://github.com/apache/kafka/blob/7c0a302c4da9d53a8fddc504a9fac8d8afecbec8/core/src/test/java/kafka/testkit/KafkaClusterTestKit.java#L305-L312) is the thing. 

NOTE: We cannot set `controller.quorum.voters` the way we want. (always, `0.0.0.0:0`) 

I may not be very knowledgeable about `KRAFT`, so this might not be accurate. 
It seems that in that state, there is no quorum available for voting.

The `AsyncKafkaConsumer` sends a `HeartBeat` to the `Broker` using the `consumer group protocol`, but the `Broker` cannot assign `Topic Partitions` because there are no `quorum voters` available to assign.

[...]

PR #3237, opened 2024-05-06
GH-3145 : Add test for next generation consumer-group rebalance protocol

chickenchickenlove:
### Motivation:
- https://github.com/spring-projects/spring-kafka/issues/3145
- Currently, i believe that we cannot adequately test the `new Consumer Rebalancing Protocol` using `EmbeddedKafka`. because we cannot set `controller.quorum.voters`. this is because `controller.quorum.voters` are hard-coded as `0.0.0.0:0` and `new consumer rebalancing protocol` support `KRaft` only. so, we cannot test `new consumer Rebalancing protocol` by using `EmbeddedKafka`. 
- We can use `TestContainers` to test it. however, it is not appropriate to add a new dependency to `spring-kafka` just for that test. thus, i would like to add only a sample code for writing a Test Case using `TestContainer`.

### Modifications
- Add `sample-07` to `/samples/` to test `new consumer rebalancing protocol` by using `testContainers`.

### To Reviewer
1. `new consumer protocol` does not support `consumer.enforceRebalance()`.
2. `new consumer protocol` does not support `Zookeeper mode`.  (It's not explicitly mentioned. but you can check it from this issue : https://issues.apache.org/jira/browse/KAFKA-16657)
3. I guess that `new consumer protocol` has tiny concurrent issue. (I reported it via https://issues.apache.org/jira/browse/KAFKA-16670)

In case of 3, my work around is using dummy `consumer` to make broker load metadata from `__consumer_offsets`. 
```java
			Consumer<Object, Object> dummyConsumer = new KafkaConsumer<>(props);
			Thread.sleep(5000);
			dummyConsumer.close();
```

### Test Scenario
1. `spring-kafka consumer` subscribe `hello-topic`. (1st rebalancing occurs)
2. `rawKafkaConsumer` subscribe `hello-topic`. (2nd rebalancing occurs. this is because `new consumer rebalancing protocol` does not support `consumer.enforceRebalance()`) 
3. `rawKafkaConsumer` is closed. (3rd rebalancing occurs).

### Result:
- Might be close https://github.com/spring-projects/spring-kafka/issues/3145

### Releated
- https://issues.apache.org/jira/browse/KAFKA-16657
- https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol


sobychacko:
Thanks, @chickenchickenlove, for this PR. We will take a look at this today.

sobychacko:
@chickenchickenlove Sorry about the delay in reviewing. I took a look. Here are some general comments.

1. Although it is a sample, adding the author name, copyrights, and checkstyle-related formatting is still recommended.
2. Please rename the test with more descriptive names.

I was able to follow your sample, but could you please clarify the connection of `Consumer#enforceRebalance` in your comments above?

Thanks!

chickenchickenlove:
@sobychacko thanks for your comments 🙇‍♂️ 

> Although it is a sample, adding the author name, copyrights, and checkstyle-related formatting is still recommended.
> Please rename the test with more descriptive names.

I make new commit to apply your review. 
There is no `checkStyle` rule for `/samples/`, so i did copy `NewConsumerRebalancProtocol` under `spring-kafka/src/test/.../` and execute `gradle :spring-kafka:checkStyleTest`, and no error occurs. 
I renamed test function as well! 

> I was able to follow your sample, but could you please clarify the connection of Consumer#enforceRebalance in your comments above?

Sorry to make you confused. 
My intention is as follows:

1. `Since consumer.enforceRebalance()` is not supported, I cannot trigger `consumer rebalancing` using the consumer API.
2. I thought that adding a new Consumer would be the simplest way to trigger `consumer rebalancing`, so I added a `rawConsumer`.


When you have free time, Please take another look 🙇‍♂️

samples/sample-07/src/test/java/com/example/sample07/NewConsumerRebalanceProtocolTest.java
```diff
@@ -0,0 +1,325 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.kafka.annotation.EnableKafka;
+import org.springframework.kafka.annotation.KafkaListener;
+import org.springframework.kafka.config.ConcurrentKafkaListenerContainerFactory;
+import org.springframework.kafka.config.KafkaListenerEndpointRegistry;
+import org.springframework.kafka.core.*;
```
artembilan, line 37: No asterisk imports, even in tests.

samples/sample-07/src/test/java/com/example/sample07/NewConsumerRebalanceProtocolTest.java
```diff
@@ -0,0 +1,325 @@
...
+			// https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol
+			// https://cwiki.apache.org/confluence/display/KAFKA/The+Next+Generation+of+the+Consumer+Rebalance+Protocol+%28KIP-848%29+-+Early+Access+Release+Notes
+			broker.addEnv("KAFKA_CFG_GROUP_COORDINATOR_REBALANCE_PROTOCOLS", "classic,consumer");
+			broker.addEnv("KAFKA_CFG_TRANSACTION_PARTITION_VERIFICATION_ENABLE", "false");
+
+			// Set Port forwarding. 12000 -> 9094.
+			String portFormat = String.format("%d:%d/%s", 12000, 9094, InternetProtocol.TCP.toDockerNotation());
+			final List<String> portsBinding = Collections.singletonList(portFormat);
+			broker.setPortBindings(portsBinding);
+			broker.start();
```
artembilan, line 308: Please, look into Spring Boot Testcontainers Support: https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.testcontainers.

I would expect that community would like to see some convenient high-level API.

samples/sample-07/src/test/java/com/example/sample07/NewConsumerRebalanceProtocolTest.java
```diff
@@ -0,0 +1,325 @@
...
+
+			Thread.sleep(5000);
+			dummyConsumer.close();
+
+			return genericContainer;
+		}
+
+		GenericContainer setUpBroker() {
+			final DockerImageName imageName = DockerImageName.parse(KAFKA_IMAGE_NAME);
+			final GenericContainer broker = new GenericContainer(imageName);
```
artembilan, line 271: Why do we use `org.testcontainers:kafka` then if we don't rely on its specific API?

samples/sample-07/src/test/java/com/example/sample07/NewConsumerRebalanceProtocolTest.java
```diff
@@ -0,0 +1,325 @@
...
+			broker.addEnv("KAFKA_CFG_TRANSACTION_STATE_LOG_MIN_ISR", "1");
+			broker.addEnv("KAFKA_CFG_TRANSACTION_STATE_LOG_REPLICATION_FACTOR", "1");
+			broker.addEnv("KAFKA_CFG_OFFSETS_TOPIC_REPLICATION_FACTOR", "1");
+			broker.addEnv("KAFKA_CFG_DEFAULT_REPLICATION_FACTOR", "1");
+
+			// KIP-848. If you want more, see links below.
+			// https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol
+			// https://cwiki.apache.org/confluence/display/KAFKA/The+Next+Generation+of+the+Consumer+Rebalance+Protocol+%28KIP-848%29+-+Early+Access+Release+Notes
+			broker.addEnv("KAFKA_CFG_GROUP_COORDINATOR_REBALANCE_PROTOCOLS", "classic,consumer");
+			broker.addEnv("KAFKA_CFG_TRANSACTION_PARTITION_VERIFICATION_ENABLE", "false");
```
artembilan, line 302: Cannot all of these env vars be extracted into a properties files?
And then we can use `broker.getEnvMap().putAll()`

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTests.java
```diff
@@ -0,0 +1,13 @@
+package com.example.sample07;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.context.SpringBootTest;
+
+@SpringBootTest
+class Sample07ApplicationTests {
```
artembilan, line 7: Why do we still need this class?
Or why don't do the logic in your new class exactly in this one?
So, we won't need to fight for naming 😄

samples/sample-07/src/test/java/com/example/sample07/NewConsumerRebalanceProtocolTest.java
```diff
@@ -0,0 +1,325 @@
...
+		}
+
+		@Bean
+		ConsumerFactory<Integer, String> consumerFactory() {
+			final Map<String, Object> props = getConsumerProperties();
+			return new DefaultKafkaConsumerFactory<>(props);
+		}
+
+		@Bean
+		public ThreadPoolTaskExecutor threadPoolTaskExecutor() {
```
artembilan, line 241: We don't need beans from here and above since they have to be auto-configured by Spring Boot.
When you switch to the `@ServiceConnection` on the container bean definition, everything will be picked up by auto-configuration.

artembilan (changes requested):

artembilan:
@chickenchickenlove ,

> Since `consumer.enforceRebalance()` is not supported, I cannot trigger consumer rebalancing using the consumer API.

I'm not sure in your statement.
See `MessageListenerContainer`:
```
	/**
	 * Alerting the consumer to trigger an enforced rebalance. The actual enforce will happen
	 * when the next poll() operation is invoked.
	 * @since 3.1.2
	 * @see org.apache.kafka.clients.consumer.KafkaConsumer#enforceRebalance()
	 */
	default void enforceRebalance() {
```

sobychacko:
I think what he means is that the new consumer (non-classic) does not support explicitly triggering `enforceRebalance`.

artembilan:
Ah! I see now in the `AsyncKafkaConsumer`:
```
    public void enforceRebalance(String reason) {
        log.warn("Operation not supported in new consumer group protocol");
    }
```

samples/sample-07/src/test/java/com/example/sample07/NewConsumerRebalanceProtocolTest.java
chickenchickenlove: @artembilan , sorry to say that it is impossible to use high-level API like `@ServiceConnection` currently. 

1. `apache/kafka` support `server side consumer rebalancing protocol` from `apache/kafka:3.7.0`
2. `testContainers` support `cp/conflunetic-kafka` only.
3.  There is no `cp/confluent-kafka` version compatible with `apache/kafka:3.7.0`. (See more, please click [link](https://docs.confluent.io/platform/current/installation/versions-interoperability.html))
4. The low-level code in `KafkaContainer`  is coupled with a lot of Confluent stuff, including the name of the vars (https://github.com/testcontainers/testcontainers-java/issues/8107#issuecomment-1900537709)

Considering the above four points, I am unable to use `KafkaContainer`.
Thus i should use `GenericContainer` instead of `KafkaContainer`.
The `@ServiceConnection` you recommended, depends on `KafkaContainer`.

As you can see [this](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.testing.testcontainers.service-connections), 
There is no ConnectionDetails instance for `GenericContainer`. 
image

This is why i cannot use high-level API of `testContainers` like `@ServiceConnection`.
I used `@TestContainers` to the best of your advice. If you have a better way, please let me know. 😄

chickenchickenlove: In the docs of `test-containers:kafka`, they support `apache/kafka` with `org.testcontainers.kafka.KafkaContainer`.  ([link](https://java.testcontainers.org/modules/kafka/))

However, there is no `org.testcontainers.kafka.KafkaContainer` and maintainer said `there is mismatch between the docs and the implementation` ([link](https://github.com/testcontainers/testcontainers-java/issues/8576#issuecomment-2077955213)).

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
```diff
@@ -0,0 +1,335 @@
...
+			propsDummy.setProperty(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG,
+									StringDeserializer.class.getName());
+			propsDummy.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, BROKER);
+			propsDummy.setProperty(ConsumerConfig.GROUP_ID_CONFIG, "hello");
+			propsDummy.setProperty(ConsumerConfig.GROUP_PROTOCOL_CONFIG, "consumer");
+			propsDummy.setProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+			Consumer<Object, Object> dummyConsumer = new KafkaConsumer<>(propsDummy);
+
+			Thread.sleep(5000);
+			dummyConsumer.close();
```
artembilan, line 289: This looks like a mix of concerns.
Why cannot we just rely on the auto-configuration for the `ConsumerFactory` and then have such a dummy consumer created from the `ApplicationRunner`?

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
```diff
@@ -0,0 +1,335 @@
...
+
+			final Map<String, Object> props = getConsumerProperties();
+			return new DefaultKafkaConsumerFactory<>(props);
+		}
+
+		@Bean
+		public ThreadPoolTaskExecutor threadPoolTaskExecutor() {
```
artembilan, line 296: We don't need this bean. Spring Boot auto-configures for us one: `TaskExecutionAutoConfiguration`.
If we need some custom its properties, they can be set from the `application.properties`: `spring.task.execution` prefix.

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
```diff
@@ -0,0 +1,335 @@
...
+			final ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+			executor.setCorePoolSize(5);
+			executor.setQueueCapacity(20);
+			executor.setMaxPoolSize(10);
+			return executor;
+		}
+
+		@Bean
+		public GenericContainer<?> brokerContainer() {
```
artembilan, line 305: I don't think this suppose to be a bean. 
See `@Container` and read its Javadocs how Testcontainers manage their lifecycle.

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
```diff
@@ -0,0 +1,335 @@
...
+
+	static Map<String, Object> getConsumerProperties() {
+		final Map<String, Object> props = new HashMap<>();
+
+		props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, BytesDeserializer.class.getName());
+		props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class.getName());
+		props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, BROKER);
+		props.put(ConsumerConfig.GROUP_ID_CONFIG, GROUP_ID);
+		props.put(ConsumerConfig.GROUP_PROTOCOL_CONFIG, "consumer");
+		props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
artembilan, line 209: These all could go to the `application.properties`

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
```diff
@@ -0,0 +1,335 @@
...
+		@Bean
+		ConcurrentKafkaListenerContainerFactory<Integer, String>
+		kafkaListenerContainerFactory(ConsumerFactory<Integer, String> consumerFactory) {
+
+			final ConcurrentKafkaListenerContainerFactory<Integer, String> factory =
+					new ConcurrentKafkaListenerContainerFactory<>();
+
+			factory.setConsumerFactory(consumerFactory);
+
+			factory.getContainerProperties().setConsumerRebalanceListener(new ConsumerAwareRebalanceListener() {
```
artembilan, line 261: The `ConsumerAwareRebalanceListener` could be just a bean and Spring Boot takes care about it to be injected into the auto-configured `kafkaListenerContainerFactory`.

artembilan (changes requested):
Some README.adoc required for this new sample to explain what is going on and why.

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
chickenchickenlove: > I guess that new consumer protocol has tiny concurrent issue. (I reported it via https://issues.apache.org/jira/browse/KAFKA-16670)
> In case of 3, my work around is using dummy consumer to make broker load metadata from __consumer_offsets.

I don't think so. it is intended. 😄 
As i said in PR description, this is work around to solve tiny issue. 

1. Right after the `broker` starts, `broker` seems not load metadata from `__consumer_offsets`. Once the metadata is loaded, the `broker` can function as a normal coordinator.
2. When a `consumer` is created, `consumer` sends a `findCoordinator` request to the `bootstrap server`.
3. Upon receiving `findCoordinator` request, the `broker` schedules a metadata load from `__consumer_offsets`. However, at this point, since the `broker` has not loaded the metadata yet.

and then, `consumer.subscribe()` and `consumer.poll()` will be executed. 
If `broker` does not load metadata yet, `consumer` received response of first `consumer.poll() and `Response` said `this coordinator is invalid`. 
And then, `consumer` try to find valid `coordinator`, but stuck in that status forever. 

This is problem which what i said. 
To solve this problem, I create `dummyConsumer` to make broker schedule loading metadata.  (and `dummyConsumer` will be closed after sleeping 5 sec, because it's only needed to trigger schedule). 

This is why i make other `consumer` without `ConsumerFactory`! 🤔 
What do you think?

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
chickenchickenlove: This is not for consumer managed by `spring-kafka`. 

I make two `consumers` which not are managed by `spring-kafka`.
One is `rawConsumer` for triggering `consumer rebalance`. 
The other one is `dummyConsumer` for making broker scheduling to load metadata. (https://github.com/spring-projects/spring-kafka/pull/3237#discussion_r1596833689)

Thus, we don't need to `application.properties`. 
I think `application.properties` is for the consumer managed by `spring-kafka`, right?  (for example, `spring.kafka.admin.client-id`) 
If you want to migrate this to properties, how about making `dummy-consumer.properties` and use it? 
```java
	static Map<String, String> getPropertiesFromFile() {
		final Resource resource = new ClassPathResource(BROKER_PROPERTIES_FILE_PATH);
		try {
			final Properties properties = PropertiesLoaderUtils.loadAllProperties(resource.getFilename());
			return properties.stringPropertyNames().stream()
							.collect(Collectors.toMap(
										s -> s,
										s -> (String) properties.get(s)));
		}
		catch (IOException e) {
			throw new RuntimeException(e);
		}
	}
```

samples/sample-07/src/main/java/com/example/sample07/Sample07Application.java
```diff
@@ -0,0 +1,38 @@
...
+ * @author Sanghyeok An.
+ *
+ * @since 3.3
+ */
+
+@SpringBootApplication
+public class Sample07Application {
+
+	public static void main(String[] args) {
+		SpringApplication.run(Sample07Application.class, args);
```
artembilan, line 35: Looking into this from high level, I wonder what we are trying to achieve with this sample.
Where are real consumers (`@KafkaListener` method)  for end-user purpose.
Having the test is OK, but it does not give end-users too much value since this sample really does not demonstrate (yet) how really develop an application with that new consumer group management algorithm.

artembilan (changes requested):

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
chickenchickenlove: Ah, i missed one. 
I will move the settings for consumer which are managed by `spring-kafka` to `application.properties`. However, it seems necessary to leave the settings for `dummyConsumer` and `rawConsumer` as they are.

What do you think?

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
artembilan: > And then, consumertry to find validcoordinator`, but stuck in that status forever.

If that is the problem of Apache Kafka by itself for this new algorithm, how about to postpone such a sample until it is solved?
I wonder how end-user are supposed to solve this problem in the applications?
If you still insist having this sample, how about to rework it the way end-user would implement it in their own applications?
No one is going to do test stuff in production.
The sample purpose is to get it is and have your own application up-and-running.

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
chickenchickenlove: - https://github.com/spring-projects/spring-kafka/pull/3237#discussion_r1596833689
- https://github.com/spring-projects/spring-kafka/pull/3237#discussion_r1596642357

I tried to do it. 
However, there is a major reason why i cannot use it.

1. As i commented on [this](https://github.com/spring-projects/spring-kafka/pull/3237#discussion_r1596833689), `broker` should load metadata before `consumer.subscribe()` is executed. 
2. If using `@Container`, `broker` container will be started successfully. however, there is no hooking point to execute `dummyConsumer` to make `broker` schedule load to `metadata` before `consumer` which are managed by `spring-kafka` execute `subscribe()`. 

We should make broker schedule to load metadata before `consumer` which is managed by `spring-kafka` are registered as spring bean. However, AFAIK, there is no annotation to define the execution order at the time of registering `Spring beans` on `@Configuration`.

I have this difficulty. Do you have good idea? 🙇‍♂️

artembilan: Yeah... That's too much involved.
I would give up already making this working 😄 .

As I said in other my comment, it might be a time to postpone such a sample until Apache Kafka has that coordinator fixed. 😢 

@sobychacko , WDYT?

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
chickenchickenlove: Um... I see! 
I will try to find common way.  
If i fail to find a good common way, it might be better to close this PR. 🤔

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
chickenchickenlove: As you said, it seems to be better to postpone 😂😂.
Let me check that there is common way as last.
Because i'm not professional to kafka and still learning about it. 
Thus some mis-understanding can be existed. so, i will take another look to check mis-understanding made by me as last 😭

samples/sample-07/src/test/java/com/example/sample07/NewConsumerRebalanceProtocolTest.java
sobychacko: Could you clarify? In the above comments, you say that - `testContainers support cp/conflunetic-kafka only.` and then `In the docs of test-containers:kafka, they support apache/kafka with org.testcontainers.kafka.KafkaContainer. `. Which one do they support? Incidentally, We recently switched to `apache/kafka` from the old `wurstmeister/kafka` for some AOT testing projects we have.

chickenchickenlove: Docs says
- `org.testcontainers.containers.KafkaContainer` supports `confluentinc/cp-kafka`.
- `org.testcontainers.kafka.KafkaContainer` supports `apache/kafka`.

However, there is no `org.testcontainers.kafka.KafkaContainer`. (Please, see image below, `testContainers.kafka:1.19.7`) 
image

So I looked it up. I was able to find this issue on `testcontainers`' github.
It's kind of well-known bug. i think it means some `testContainers` version cannot use `org.testcontainers.kafka.KafkaContainer`. 

By the way, it seems to be fixed (`1.19.8`, deployed may 9).
image

When I wrote this PR, the latest version was `1.19.7`. sorry to make you confused. 🙇‍♂️

sobychacko: Thanks for the clarification!

samples/sample-07/src/test/java/com/example/sample07/Sample07ApplicationTest.java
chickenchickenlove: Hey, @artembilan 
I found root cause of `kafka`, so tried to create a new PR on `apache/kafka`.
However it seems to have already been fixed by this PR (https://github.com/apache/kafka/pull/15698) 

The problem was caused by not retrying the error response (as opposed to the log say).
However, that PR is not included in `apache/kafka:3.7.0`.  😓 

Thus, we should postpone this PR. 
May I leave this PR in an open status and postpone it?

Thanks in advance!

artembilan: Yes. We can keep it opened until the next Apache Kafka `2.7.1`.
Thank you!

samples/sample-07/src/main/java/com/example/sample07/Sample07Application.java
sobychacko: The PR in the test configuration currently has a [KafkaListener](https://github.com/spring-projects/spring-kafka/pull/3237/files#diff-5dd4447d779db9588dae1e82edfcb5e45f9cd18234a7a8851f2456cc6a02d2c5R237) where the underlying [container factory](https://github.com/spring-projects/spring-kafka/pull/3237/files#diff-5dd4447d779db9588dae1e82edfcb5e45f9cd18234a7a8851f2456cc6a02d2c5R253) uses a consumer factory where the new [consumer group protocol ](https://github.com/spring-projects/spring-kafka/pull/3237/files#diff-5dd4447d779db9588dae1e82edfcb5e45f9cd18234a7a8851f2456cc6a02d2c5R284)is used. Since this feature is in its early access form (and not recommended for production usage yet), it might still be beneficial for any users who want to play around with this in a test format. In other words, we don't expect any users to use it in a real application yet until Apache Kafka makes that recommendation, but users might be interested in trying this out via Spring Kafka, and this sample could help with that — just my 2 cents.

chickenchickenlove:
@artembilan 
Thanks for your comments and your times. 🙇‍♂️ 
I want to sync my understanding and your comments. 

### My Understanding
1. Replace `testContainers` to `docker-compose` with `spring-boot`. 
2. No need to test `New Consumer Rebalance` in detail. 

### Direction
- Just make `docker-compose.yaml` and use it with `spring-boot docker-compose`. 
- Simplify test codes. because we don't need to check each `consumer rebalance`. 
Thus, it seems sufficient to retain only the logic where the Producer supplies one message, and the Consumer checks if it can poll that message.

If i misunderstood about your intention, please let me know and give more detail when you have free time 🙇‍♂️

artembilan:
> Thus, it seems sufficient to retain only the logic where the Producer supplies one message, and the Consumer checks if it can poll that message.

Sounds like a plan.

We need to have our consumer in a new consumer group algorithm and that's it.
Perhaps cause some rebalance adding a consumer on the fly.
Kinda cover all the functionality suggested by this new algorithm.

chickenchickenlove:
Hi, @artembilan, @sobychacko ! I created new commit to apply your comments. 
I want to left some comments for new commit.  😃 

1. I cannot config `Consumer` via `application.yaml`. this is because `spring-kafka` does not support `spring.kafka.consumer.group-protocol`. thus, i have to config `Consumer` by manually. 
2. Anyway, this sample codes will not work properly now. this is because `Consumer`'s problem which i said before. we should wait for `kafka` to be updated to > `3.7.0`(https://github.com/apache/kafka/pull/15698). Also, `spring-kafka` should have it as well. 

I realized that there is new issue for `spring-kafka` or `spring-auto-configure` to support `spring.kafka.consumer.group-protocol`. 

When you have some free time, please take another look 🙇‍♂️. 
If this sample is okay, I will wait until `kafka` and `spring-kafka` are updated as well. 
If there's anything that needs to be fixed before `kafka` and `spring-kafka` being updated, please let me know, and I'll address it immediately. 

Thanks in advance! 🙇‍♂️

sobychacko:
@chickenchickenlove Good point, created this issue: https://github.com/spring-projects/spring-kafka/issues/3254

sobychacko:
The sample clean up looks good. I think it might be beneficial to have a `README` explaining what this sample is all about. I thought you had one before, but not seeing in the latest changes?

sobychacko (changes requested):
See my comments about the README.

samples/sample-07/README.adoc
```diff
@@ -0,0 +1,50 @@
+== Sample 7
+
+This sample demonstrated the application of the new consumer rebalance protocol in Spring Kafka.
```
sobychacko, line 3: demonstrates.

sobychacko: Spring for Apache Kafka.

samples/sample-07/README.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+The new consumer rebalance protocol refers to the Server Side rebalance protocol proposed in KIP-848.
```
sobychacko, line 5: Link to KIP-848.

sobychacko (changes requested):
See the inline comments in README.

sobychacko (changes requested):
@chickenchickenlove We can use the property like below even today:
`spring.kafka.consumer.properties[group.protocol]=CONSUMER`.

chickenchickenlove:
@sobychacko thanks for your quick comments 👍 
I make new commit to apply your comments.

Two points seem to work well as you recommended. (Even if new consumer rebalancing protocol)

image

- `spring-autoconfigure` put `group.protocol=consumer` to `ConsumerProperties` as expected.

image

- Consumer rebalancing have been completed successfully via `new rebalancing protocol`. 
- The first attempt fails. This is due to the reasons discussed earlier.
- Every attempt after the second always succeeds. 


When you have free time, please take another look! 🙇‍♂️

sobychacko (approved):

samples/sample-07/src/main/java/com/example/sample07/Sample07Application.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * New consumer rebalance protocol sample which purpose is only to demonstrate the application
+ * of the New Consumer Rebalance Protocol in Spring Kafka.
+ *
+ * @author Sanghyeok An.
+ *
+ * @since 3.3
+ */
+
+@EnableKafka
```
artembilan, line 32: We don't need this annotation since we rely on Spring Boot.

samples/sample-07/src/main/resources/application.yaml
```diff
@@ -0,0 +1,17 @@
...
+      lifecycle-management: start_and_stop
+      start:
+        command: up
+      stop:
+        command: down
+        timeout: 10s
+  kafka:
+    consumer:
+      bootstrap-servers: localhost:10000
+      group-id: sample07
```
artembilan, line 13: I see you use some other groups in your `@KafkaListener` methods.
Why do we still need it here?

samples/sample-07/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+}
+
+dependencies {
+	implementation 'org.springframework.boot:spring-boot-starter'
+	implementation 'org.springframework.kafka:spring-kafka'
+	developmentOnly 'org.springframework.boot:spring-boot-docker-compose'
+
+	testImplementation 'org.springframework.boot:spring-boot-starter-test'
+	testImplementation 'org.springframework.kafka:spring-kafka-test'
+	testImplementation 'org.springframework.boot:spring-boot-testcontainers'
```
artembilan, line 27: Why do we need all of these test dependencies if we don't have any tests in this project?
And that raises the question: why we don't have any tests in this project?

artembilan (changes requested):

artembilan:
I see PR has been merged.
Not a bit deal: we always can have follow up PRs to improve the stuff.
For example, we have merged it into the current `3.2`, but I see `@since 3.3` 🤷

sobychacko:
@chickenchickenlove Thanks for the PR with this sample. We went back and forth a lot, but thanks for being patient during the review process. We are glad that this sample is in. 

@artembilan Sorry, I noticed your latest review only after merging. I will follow up with the cleanup.

chickenchickenlove:
Hi, @artembilan  !
I'm sorry for the inconvenience caused by the unclear code. 
You have taught me a lot (But i am still lacking 😅 ). Would it be okay to write a few more improvement PR later? 

@sobychacko 
You're welcome. I'm actually glad to have the opportunity to write such PR! 
Rather, I am grateful to @sobychacko and @artembilan  for thoroughly checking my inadequate PRs and being patient with me as well. Thanks again!!

Issue #3242, opened 2024-05-09
Documentation for configuring custom `ConsumerFactory` and `ProducerFactory` beans when using Spring Boot is incorrect


wilkinsona:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

All supported versions.

**Describe the bug**

Please see https://github.com/spring-projects/spring-boot/issues/40174.

**To Reproduce**

Please see https://github.com/spring-projects/spring-boot/issues/40174.

**Expected behavior**

The custom consumer and or producer factories are used by the default listener container when following Spring Kafka's documentation. To achieve this the documentation should be updated to define the beans as `ConsumerFactory<?, ?>` and `ProducerFactory<?, ?>`.

PR #3243, opened 2024-05-09
GH-3242: Clarify docs on custom Consumer/Producer factories

sobychacko:
Fixes: #3242

**Auto-cherry-pick to `3.1.x` & `3.0.x`**


artembilan:
> Auto-cherry-pick to 3.1.x & 3.0.x

The Antora docs are there starting with version `3.1.`. So, this is not going to be back-port very easy.
I would stay just with `3.1.x` back-port, but if you insist, feel free make a fix in the `3.0.x`.
Mean while I'll merge this only to `3.1.x`.

PR #3253, opened 2024-05-16
Adding more whats-new for kakfa-client 3.7.0

sobychacko:
* Clarify the usage of the new consumer-group protocol in the 3.7.0 version of the client
* Add a note on the usage of `EmbeddedKafka` in `kraft` mode

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
@@ -18,10 +18,14 @@ It is only recommended to use for testing purposes in this version.
 Therefore, Spring for Apache Kafka supports this new consumer group protocol only to the extent of such testing level support available in the `kafka-client` itself.
 By default, Spring for Apache Kafka uses the classic consumer group protocol and when testing the new consumer group protocol, that needs to be opted-in via the `group.protocol` property on the consumer.
 
-When using the `EmbeddedKafka` with the 3.7.0 version of the `kafka-client`, the `kraft` mode is disabeld by default.
-This is due to certain instabilities observed while using `EmbeddedKafka` with `kraft` enabled, especially when testing the new consumer group protocol.
-In addition, there were some other race conditions observed, while running multiple `KafkaListener` methods with `EmbeddedKafka` in `kraft` mode in this version.
-Due to these reasons, when trying `EmbeddedKafka` with `kraft`, that needs to be explicitly opted-in by setting `kraft` to `true`.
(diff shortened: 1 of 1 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -11,6 +11,17 @@ For changes in earlier version, see xref:appendix/change-history.adoc[Change His
...
 
 This version requires 3.7.0 `kafka-clients`.
+The 3.7.0 version of Kafka client introduces the new consumer group protocol.
+Fore more details and it's limitations see https://cwiki.apache.org/confluence/display/KAFKA/The+Next+Generation+of+the+Consumer+Rebalance+Protocol+%28KIP-848%29+-+Early+Access+Release+Notes[KIP-848].
+The new consumer group protocol is an early access release and not meant to be used in production.
+It is only recommended to use for testing purposes in this version.
+Therefore, Spring for Apache Kafka supports this new consumer group protocol only to the extent of such testing level support available in the `kafka-client` itself.
+By default, Spring for Apache Kafka uses the classic consumer group protocol and when testing the new consumer group protocol, that needs to be opted-in via the `group.protocol` property on the consumer.
+
+When using the `EmbeddedKafka` with the 3.7.0 version of the `kafka-client`, the `kraft` mode is disabeld by default.
```
artembilan, line 21: This sentence does not reflect reality.
We have it `kraft = false` compiled-in. Independently of the Kafka client.

I would suggest to have a dedicated bullet for the `Testing Support Changes` and explain there why we turned it off.
But I believe that has nothing to do with Kafka client `3.7.0`

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
sobychacko: ya, i hear your point. I actually thought about it. However, the new consumer-group protocol testing with `EmbeddedKafka` in `kraft` mode may still be relevant here. What do you think?

artembilan: Well, there is catch 22. The new consumer group protocol works only in KRaft mode. 
And unfortunately, only with real broker, not one created by the `KafkaClusterTestKit`.
Then we also saw some flaws for classic group management in KRaft mode.
So, this or other way `kraft` mode is not recommended with `@EmbeddedKafka`.

Issue #3254, opened 2024-05-17
Add group.protocol property to ConsumerProperties


sobychacko:
The `3.7.0` kafka client introduced the new consumer-group protocol. Provide a way to set the property (`group.protocol`) on the `ConsumerProperties`.

PR #3255, opened 2024-05-17
GH-3254: Group protocol in ConsumerProperties

sobychacko:
Fixes: #3254

* Add group.protocol property to ConsumerProperties

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2022 the original author or authors.
+ * Copyright 2019-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -35,6 +35,7 @@ import org.springframework.util.StringUtils;
  * Common consumer properties.
  *
  * @author Gary Russell
+ * @author Soby Chacko
  * @since 2.3
  *
  */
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
```diff
@@ -501,6 +508,23 @@ public void setCheckDeserExWhenValueNull(boolean checkDeserExWhenValueNull) {
...
+	 * Return the container's group protocol.
+	 * @return Container group protocol set via the group.protocol Consumer property
+	 */
+	public String getGroupProtocol() {
+		return this.groupProtocol;
+	}
+
+	/**
+	 * Set the container group protocol.
+	 *
```
artembilan, line 521: No blank lines in method Javadocs.
The `@since` is supposed to be there on the public API.
And why is this property is not set on the target `KafkaConsumer` in this PR?

spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerProperties.java
```diff
@@ -116,6 +117,12 @@ public class ConsumerProperties {
 
 	private boolean checkDeserExWhenValueNull;
 
+	/**
+	 * Container group protocol that will be set on the Kafka Consumer.
+	 * Default to classic consumer protocol.
+	 */
+	private String groupProtocol;
```
artembilan, line 124: Isn't there some `enum` to rely on?
Or we might introduce one as an inner in this `ConsumerProperties` class...

artembilan (changes requested):

Issue #3277, opened 2024-05-24
On kotlin application, Spring kafka 3.2.0 doesn't take the message conversion logic, because of the wrong coroutine detection on MessagingMessageListenerAdapter


huisam:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

For example:

3.2.0

Between 3.1.4 and 3.2.0

**Describe the bug**

On the MessagingMessageListenerAdapter class there is a bug on the `determineInferredType` method.

```kotlin
boolean isCoroutines = KotlinDetector.isKotlinType(methodParameter.getParameterType());
isNotConvertible |= isCoroutines;
```

if you determine the coroutine function, you should have to use 
```kotlin
KotlinDetector.isSuspendingFunction(Method method)
```
not `isKotlinType` method

because on kotlin, `@KafkaListener` method parameter type is always kotlin type. So the `conversionNeeded` flag changes to false.
This occurs messageConverters not working normally.

**To Reproduce**

```kotlin
@KafkaListener(topics = [TEST_KAFKA_TOPIC], groupId = TEST_KAFKA_TOPIC_GROUP)
fun onMessage(message: TestMessage) {
    execute()
}

data class TestMessage(val key: String)
```

**Expected behavior**

consumes normally and execute the `execute()` method but it occurs `MethodArgumentNotValidException` exception 

```
org.springframework.messaging.handler.annotation.support.MethodArgumentNotValidException: Could not resolve method parameter at index 0 in public void com.company.test.spring.kafka.TestKafkaService.onMessage(com.company.test.spring.kafka.TestMessage): 1 error(s): [Error in object 'message': codes []; arguments []; default message [Payload value must not be empty]] 
```

**Sample**

sample is on the issue

artembilan:
Looks like we have overseen this scenario when we tested Kotlin support for `@KafkaListener`:
```
		@KafkaListener(id = "kotlin", topics = ["kotlinTestTopic1"], containerFactory = "kafkaListenerContainerFactory")
		fun listen(value: String) {
			this.received = value
			this.latch1.countDown()
		}
```
That `String` is indeed not a Kotlin type 😄 .

Feel free to contribute the fix as you have just explained.
And I guess with respective fix for that `EnableKafkaKotlinTests.`test listener`()`  test and its configuration with the mentioned `listen()` method.

Thanks

huisam:
@artembilan Thank you for your feedback on my opinion. 

[...]

PR #3283, opened 2024-05-26
GH-3277 : Fix coroutine detection login on MessagingMessageListenerAdapter

huisam:
## Motivation:

* https://github.com/spring-projects/spring-kafka/issues/3277
* `isCoroutines` is a method unit, but on the determineInferredType method it determine by each parameter type. So I think that determine by parameter type is more nice way to improve it

## Modifications:

* Change variable naming `isCoroutines` to `isKotlinContinuation` variable 
* Change the coroutine detection logic 

## DIscussion

* Feel free to discuss about the test code improvement way

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AdapterUtils.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AdapterUtils.java
@@ -35,6 +35,7 @@ import reactor.core.publisher.Mono;
  *
  * @author Gary Russell
  * @author Wang Zhiyang
+ * @author Huijin Hong
  * @since 2.5
  *
  */
@@ -124,4 +125,13 @@ public final class AdapterUtils {
 		return CompletableFuture.class.isAssignableFrom(resultType);
 	}
 
+	/**
+	 * Return the true when type is {@code Continuation}.
+	 * @param  parameterType {@code MethodParameter} parameter type.
+	 * @return type is {@code Continuation}.
+	 * @since 3.2
+	 */
+	public static boolean isKotlinContinuation(Class<?> parameterType) {
+		return "kotlin.coroutines.Continuation".equals(parameterType.getName());
+	}
 }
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -74,6 +73,7 @@
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 
+import kotlin.coroutines.Continuation;
```
artembilan, line 76: This is not OK because Kotlin is `optional` dependency and non-Kotlin projects are gong to fail with `ClassNotFoundException`.

I wonder why you don't use `KotlinDetector.isSuspendingFunction()`.
Looks like it has exact logic you are trying to achieve:
```
        if (isKotlinType(method.getDeclaringClass())) {
            Class<?>[] types = method.getParameterTypes();
            if (types.length > 0 && "kotlin.coroutines.Continuation".equals(types[types.length - 1].getName())) {
                return true;
            }
        }
```
It might be extracted from the loop of these params for some optimization, but still the same outcome.

artembilan (changes requested):

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
huisam: Oh, That's a good point. I missed about considering non-Kotlin projects.
I'll use the `isSuspendingFunction` as I first suggested.

huisam:
I add my name on `@author` to affected class. Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/AdapterUtils.java
```diff
@@ -124,4 +125,13 @@ public static boolean isCompletableFuture(Class<?> resultType) {
 		return CompletableFuture.class.isAssignableFrom(resultType);
 	}
 
+	/**
+	 * Return the true when type is {@code Continuation}.
+	 * @param  parameterType {@code MethodParameter} parameter type.
+	 * @return type is {@code Continuation}.
+	 * @since 3.2
```
artembilan, line 132: All good, but this version has been released already. 
Therefore:
```suggestion
	 * @since 3.2.1
```

artembilan (approved):
LGTM.
Will merge when build is green.
Thank you!

Issue #3291, opened 2024-06-01
Possible inconsistency in DLT topic naming convention


Watlas:
**Expected Behavior**

DLT topics should be created consistently, using a standardized suffix, preferably with a dash, for example, all ending with "-dlt".

**Current Behavior**

By default, DLT topics are created with ".dlt" at the end. However, when using the retry non-blocking solution, topics are created with "-dlt". This causes inconsistency in the naming of the topics.

![image](https://github.com/spring-projects/spring-kafka/assets/57519850/4e1ac540-772e-4493-8ccc-a2ffa1966a31)

**Context**

This inconsistency can cause issues when migrating from one solution to another. For example, if a user wants to migrate from a solution that uses blocking retry (which creates topics with ".dlt") to a non-blocking retry solution (which creates topics with "-dlt"), the old ".dlt" topics will become obsolete. The new system will no longer communicate with these topics due to the difference in naming.

[According to](https://github.com/apache/kafka/blob/2.3.0/core/src/main/scala/kafka/admin/TopicCommand.scala#L147-L148), Kafka recommends not mixing dashes and dots in topic names to avoid conflicts. This mix can lead to issues in identifying and managing topics, further complicating system maintenance and monitoring.

**Suggestiont**

Standardize the creation of DLT topics to use a consistent naming convention, preferably using a dash suffix, such as "-dlt". This will ensure compatibility and avoid conflicts when transitioning between different retry solutions.

PR #3292, opened 2024-06-01
GH-3291 - Possible inconsistency in DLT topic naming convention

Watlas:
**Standardize DLT Topic Naming Convention**

_[Issue Reference](https://github.com/spring-projects/spring-kafka/issues/3291)_

**Description:**

This pull request addresses the inconsistency in the naming convention of DLT topics when using different retry mechanisms. By standardizing the suffix to "-dlt", we aim to ensure compatibility and avoid conflicts during migration between retry solutions.

**Expected Behavior:**

DLT topics should be created consistently, using a standardized suffix, preferably with a dash, for example, all ending with "-dlt".

**Current Behavior:**

By default, DLT topics are created with ".dlt" at the end. However, when using the retry non-blocking solution, topics are created with "-dlt". This causes inconsistency in the naming of the topics.

**Context:**

This inconsistency can cause issues when migrating from one solution to another. For example, if a user wants to migrate from a solution that uses blocking retry (which creates topics with ".dlt") to a non-blocking retry solution (which creates topics with "-dlt"), the old ".dlt" topics will become obsolete. The new system will no longer communicate with these topics due to the difference in naming.

Additionally, Kafka recommends not mixing dashes and dots in topic names to avoid conflicts. This mix can lead to issues in identifying and managing topics, further complicating system maintenance and monitoring.

**Solution:**

Standardize the creation of DLT topics to use a consistent naming convention, preferably using a dash suffix, such as "-dlt". This will ensure compatibility and avoid conflicts when transitioning between different retry solutions.

**Implementation Details:**

- Updated the naming convention for DLT topics to use the "-dlt" suffix consistently.
- Adjusted relevant configurations and documentation to reflect this change.

Please review the changes and provide feedback. If there are any suggestions or concerns regarding this implementation, I am open to further discussion.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/annotation-error-handling.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/annotation-error-handling.adoc
@@ -614,7 +614,7 @@ The framework provides the `DeadLetterPublishingRecoverer`, which publishes the 
 The recoverer requires a `KafkaTemplate<Object, Object>`, which is used to send the record.
 You can also, optionally, configure it with a `BiFunction<ConsumerRecord<?, ?>, Exception, TopicPartition>`, which is called to resolve the destination topic and partition.
 
-IMPORTANT: By default, the dead-letter record is sent to a topic named `<originalTopic>.DLT` (the original topic name suffixed with `.DLT`) and to the same partition as the original record.
+IMPORTANT: By default, the dead-letter record is sent to a topic named `<originalTopic>.-dlt` (the original topic name suffixed with `.-dlt`) and to the same partition as the original record.
 Therefore, when you use the default resolver, the dead-letter topic **must have at least as many partitions as the original topic.**
 
 If the returned `TopicPartition` has a negative partition, the partition is not set in the `ProducerRecord`, so the partition is selected by Kafka.
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
@@ -64,6 +64,7 @@ import org.springframework.util.ObjectUtils;
  *
  * @author Gary Russell
  * @author Tomaz Fernandes
+ * @author Watlas
  * @since 2.2
  *
  */
@@ -75,7 +76,7 @@ public class DeadLetterPublishingRecoverer extends ExceptionClassifier implement
 	protected final LogAccessor logger = new LogAccessor(LogFactory.getLog(getClass())); // NOSONAR
 
 	private static final BiFunction<ConsumerRecord<?, ?>, Exception, TopicPartition>
-		DEFAULT_DESTINATION_RESOLVER = (cr, e) -> new TopicPartition(cr.topic() + ".DLT", cr.partition());
+		DEFAULT_DESTINATION_RESOLVER = (cr, e) -> new TopicPartition(cr.topic() + "-dlt", cr.partition());
 
 	private static final long FIVE = 5L;
 
--- a/spring-kafka/src/test/java/org/springframework/kafka/annotation/BatchListenerConversionTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/annotation/BatchListenerConversionTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2023 the original author or authors.
+ * Copyright 2017-2024 the original author or authors.
  *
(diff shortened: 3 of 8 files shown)


sobychacko:
Could you add the note as part of the what-new section? In it, indicate that if any users want the `.dlt` behavior, they need to opt-in for that. Feel free to update the PR. Then we can do further review. Thanks!

Watlas:
@sobychacko  Done.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/annotation-error-handling.adoc
```diff
@@ -614,7 +614,7 @@ The framework provides the `DeadLetterPublishingRecoverer`, which publishes the
 The recoverer requires a `KafkaTemplate<Object, Object>`, which is used to send the record.
 You can also, optionally, configure it with a `BiFunction<ConsumerRecord<?, ?>, Exception, TopicPartition>`, which is called to resolve the destination topic and partition.
 
-IMPORTANT: By default, the dead-letter record is sent to a topic named `<originalTopic>.DLT` (the original topic name suffixed with `.DLT`) and to the same partition as the original record.
+IMPORTANT: By default, the dead-letter record is sent to a topic named `<originalTopic>.-dlt` (the original topic name suffixed with `.-dlt`) and to the same partition as the original record.
```
sobychacko, line 617: <originalTopic>-dlt and -dlt.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
```diff
@@ -75,7 +76,7 @@ public class DeadLetterPublishingRecoverer extends ExceptionClassifier implement
 	protected final LogAccessor logger = new LogAccessor(LogFactory.getLog(getClass())); // NOSONAR
 
 	private static final BiFunction<ConsumerRecord<?, ?>, Exception, TopicPartition>
-		DEFAULT_DESTINATION_RESOLVER = (cr, e) -> new TopicPartition(cr.topic() + ".DLT", cr.partition());
+		DEFAULT_DESTINATION_RESOLVER = (cr, e) -> new TopicPartition(cr.topic() + "-dlt", cr.partition());
```
sobychacko, line 79: `-DLT` looks better. That way, it is almost identical to the old naming convention, except for the new delimiter.

sobychacko: This means that all the tests in this PR need to be modified.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/annotation-error-handling.adoc
Watlas: I will fix

spring-kafka/src/main/java/org/springframework/kafka/listener/DeadLetterPublishingRecoverer.java
Watlas: @sobychacko  I think if we use "**-DLT**", it might cause compatibility issues with the retry topics that currently use "**dlt**" in lowercase. Don't you think so? Also, I don't see much need to use uppercase here.

sobychacko: Thats OK then. Let's keep it as it is.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -7,3 +7,13 @@
 This section covers the changes made from version 3.2 to version 3.3.
 For changes in earlier version, see xref:appendix/change-history.adoc[Change History].
 
+
+[[x33-dlt-topic-naming]]
+=== DLT Topic Naming Convention
+
+The naming convention for DLT topics has been standardized to use the "-dlt" suffix consistently. This change ensures compatibility and avoids conflicts when transitioning between different retry solutions. Users who wish to retain the ".dlt" suffix behavior need to opt-in explicitly by setting the appropriate DLT name property.
```
sobychacko, line 14: One more change is needed. It should be `Users who wish to retain the ".DLT..."`

sobychacko (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
Watlas: @sobychacko done.

PR #3226, opened 2024-04-30
Fix typos in docs and classes

bky373:
### Changes
- **annotation-error-handling.adoc**
    - `records` -> `things` (Change to a more appropriate value)
- **events.adoc**
    - `ConsumerStartingEvent` -> `ConsumerStartedEvent` (To remove duplicates)
- **listener-annotation.adoc**
    - `ProducerConfig` -> `ConsumerConfig` (Change to a more appropriate value)
    - `embeddedKafka.getBrokersAsString()` -> `"localhost:9092"` (It's not necessary)
-  **sending-messages.adoc**
    -  `failedProducerRecord` -> `producerRecord` (`failedProducerRecord` property does not exist)
- **serdes.adoc**
    - `contentMessageConverter` -> `contentTypeConverter` (`contentMessageConverter` does not exist)
    - `WHen` -> `When` (typo)
- **replytopic/features.adoc**
    - `fixedBackoff` -> `fixedBackOff` (`fixedBackoff` does not exist)
    - `customBackOff` -> `customBackoff` (`customBackOff` does not exist)
    - `DeadLetterPublishingRecover` -> `DeadLetterPublishingRecoverer` (typo)
- **ReplyingKafkaTemplate.java**
    - `start()ed` -> `started` (typo)
- **BackOffValuesGenerator.java**
    - `checkBackOffPolicyTipe` -> `checkBackOffPolicyType` (typo)
- **RetryTopicConfigurer.java**
    - `fixedBackoff` -> `fixedBackOff` (`fixedBackoff` does not exist)

This document is truly awesome! It's been incredibly helpful for me to learn Spring Kafka.
I hope this PR is a small way to show my appreciation for all the help I've received.
I'd love to contribute code in the future. Thanks again!

PR #3293, opened 2024-06-02
Update misleading contents in ConsumerPartitionPausedEvent and features.adoc

bky373:
## Changes
- **ConsumerPartitionPausedEvent**
  - Correct misleading method name and content.
    - `ConsumerPausedEvent` -> `ConsumerPartitionPausedEvent` (This prevents confusion with the existing event `ConsumerPausedEvent`).
    - `getPartitions()` -> `getPartition()` (Since the event is specific to a single partition, use the singular form)
- **features.adoc**
  - `DeadLetterPublishingRecovererer` -> `DeadLetterPublishingRecoverer` 
  - This reverts the incorrect change made in [PR #3226](https://github.com/spring-projects/spring-kafka/pull/3226).

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerPartitionPausedEvent.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerPartitionPausedEvent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2021 the original author or authors.
+ * Copyright 2018-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,6 +22,7 @@ import org.apache.kafka.common.TopicPartition;
  * An event published when a consumer partition is paused.
  *
  * @author Tomaz Fernandes
+ * @author Borahm Lee
  * @since 2.7
  *
  */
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerPartitionPausedEvent.java
```diff
@@ -47,13 +48,13 @@ public ConsumerPartitionPausedEvent(Object source, Object container, TopicPartit
 	 * Return the paused partition.
 	 * @return the partition.
 	 */
-	public TopicPartition getPartitions() {
+	public TopicPartition getPartition() {
```
sobychacko, line 51: Since this is a public API, it might be better to deprecate this method, mark it for removal, and add the new one `getPartition()`. In a future version, we can completely remove this.

sobychacko (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerPartitionPausedEvent.java
bky373: Thanks for your comment! I just [made a change](https://github.com/spring-projects/spring-kafka/pull/3293/commits/804ae8501d20de71a211faffa5ddf5f7b65b3e83) and check it please.

spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerPartitionPausedEvent.java
```diff
@@ -46,14 +47,24 @@ public ConsumerPartitionPausedEvent(Object source, Object container, TopicPartit
...
 	 */
+	@Deprecated(since = "3.3", forRemoval = true)
 	public TopicPartition getPartitions() {
 		return this.partition;
 	}
 
+	/**
+	 * Return the paused partition.
+	 * @return the partition.
+	 */
```
sobychacko, line 60: Need the `@since` tag here.

sobychacko (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerPartitionPausedEvent.java
bky373: Oh Thanks for letting me know. I've added it.

Issue #2302, opened 2022-06-09
consumer group specific offset seeking for AbstractConsumerSeekAware 


akemalsaglam:
**Expected Behavior**

We want to be able to seek offset for specific consumer group by using AbstractConsumerSeekAware. 

**Current Behavior**

regarding to below implementation it is clear that we can seek offset for all assigned partitions in a topic regardless of different consumer group ids.
  
```
@Override
    public void onPartitionsAssigned(Map<TopicPartition, Long> assignments, ConsumerSeekCallback callback) {
        super.onPartitionsAssigned(assignments, callback);
    }

    @Override
    public void seekToTimestamp(long time) {
        getSeekCallbacks().forEach((tp, callback) -> {
            callback.seekToTimestamp(tp.topic(), tp.partition(), time);
        });
    }
```

**Context**
For our use case there might be more than one group instance which is assigned same partition in a topic. Below example might be useful to describe our case:

1. we have one topic let's name it:  product.feed.fullexport, it has 12 partitions.
2. 10 different micro-services with different group ids are listening same topic and each has 12 concurrent consumers.
3. when we want to seek offset by using above ConsumerSeekCallback implementation for one of the micro-service, it affects all assigned partitions and listening consumer instances regardless to expected group id.

Is there any way to seek offset in a partition but only for specific group id?

artembilan:
I don't see a behavior you are describing:
```
@SpringBootApplication
public class KafkaGh2302Application {

	public static void main(String[] args) {
		SpringApplication.run(KafkaGh2302Application.class, args);
	}


	@Bean
	public NewTopic topic() {
		return new NewTopic("seekExample", 3, (short) 1);
	}

	@Component
	public static class Listener extends AbstractConsumerSeekAware {

		@KafkaListener(id = "seekExample", topics = "seekExample", concurrency = "3")
		public void listen(String payload) {
			System.out.println("Listener received: " + payload);
		}

		public void seekToStart() {
			getSeekCallbacks().forEach((tp, callback) -> callback.seekToBeginning(tp.topic(), tp.partition()));
		}

	}

	@Component
	public static class Listener2 extends AbstractConsumerSeekAware {

		@KafkaListener(id = "seekExample2", topics = "seekExample", concurrency = "3")
		public void listen(String payload) {
			System.out.println("Listener2 received: " + payload);
		}

		public void seekToStart() {
[...]

PR #3318, opened 2024-06-18
GH-2302: Enable consumer seek only on matching group Id

bky373:
## Backgrounds
- Fixes: https://github.com/spring-projects/spring-kafka/issues/2302
- Using `@KafkaListener`, listeners with different group IDs can exist within the same class.
- Using class which extends `AbstractConsumerSeekAware`, when seek is performed in one listener, it is executed in all listeners, even if it is not desired.

## Changes
- Add a new default method (`matchGroupId()`) to `ConsumerSeekAware`. This will allow setting the callback and performing seeks only for the desired group ID. **(default: `false`)**
- Override `matchGroupId()` in `AbstractConsumerSeekAware` to return `true`. It would be nice to set this to `false` so that users should specify the groupId themselves, but in this case the callback from `KafkaMessageListenerContainer.ListenerConsumer.setupSeeks()` would not be registered, so I set it to `true`. And this is also to ensure backwards compatibility.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2024 the original author or authors.
+ * Copyright 2019-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,7 +33,6 @@ import org.springframework.lang.Nullable;
  * having to keep track of the callbacks itself.
  *
  * @author Gary Russell
- * @author Borahm Lee
  * @since 2.3
  *
  */
@@ -47,59 +46,43 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	@Override
 	public void registerSeekCallback(ConsumerSeekCallback callback) {
-		if (matchGroupId()) {
-			this.callbackForThread.put(Thread.currentThread(), callback);
-		}
+		this.callbackForThread.put(Thread.currentThread(), callback);
 	}
 
 	@Override
 	public void onPartitionsAssigned(Map<TopicPartition, Long> assignments, ConsumerSeekCallback callback) {
-		if (matchGroupId()) {
-			ConsumerSeekCallback threadCallback = this.callbackForThread.get(Thread.currentThread());
-			if (threadCallback != null) {
-				assignments.keySet()
-						.forEach(tp -> {
-							this.callbacks.put(tp, threadCallback);
-							this.callbacksToTopic.computeIfAbsent(threadCallback, key -> new LinkedList<>())
-									.add(tp);
-						});
-			}
-		}
-	}
-
-	@Override
-	public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
-		if (matchGroupId()) {
-			partitions.forEach(tp -> {
-				ConsumerSeekCallback removed = this.callbacks.remove(tp);
-				if (removed != null) {
-					List<TopicPartition> topics = this.callbacksToTopic.get(removed);
-					if (topics != null) {
-						topics.remove(tp);
-						if (topics.size() == 0) {
-							this.callbacksToTopic.remove(removed);
-						}
-					}
-				}
+		ConsumerSeekCallback threadCallback = this.callbackForThread.get(Thread.currentThread());
+		if (threadCallback != null) {
+			assignments.keySet().forEach(tp -> {
+				this.callbacks.put(tp, threadCallback);
+				this.callbacksToTopic.computeIfAbsent(threadCallback, key -> new LinkedList<>()).add(tp);
 			});
 		}
 	}
 
 	@Override
-	public void unregisterSeekCallback() {
-		if (matchGroupId()) {
-			this.callbackForThread.remove(Thread.currentThread());
-		}
+	public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
+		partitions.forEach(tp -> {
+			ConsumerSeekCallback removed = this.callbacks.remove(tp);
+			if (removed != null) {
+				List<TopicPartition> topics = this.callbacksToTopic.get(removed);
+				if (topics != null) {
+					topics.remove(tp);
+					if (topics.size() == 0) {
+						this.callbacksToTopic.remove(removed);
+					}
+				}
+			}
+		});
 	}
 
 	@Override
-	public boolean matchGroupId() {
-		return true;
+	public void unregisterSeekCallback() {
+		this.callbackForThread.remove(Thread.currentThread());
 	}
 
 	/**
 	 * Return the callback for the specified topic/partition.
-	 *
 	 * @param topicPartition the topic/partition.
 	 * @return the callback (or null if there is no assignment).
 	 */
@@ -110,7 +93,6 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	/**
 	 * The map of callbacks for all currently assigned partitions.
-	 *
 	 * @return the map.
 	 */
 	protected Map<TopicPartition, ConsumerSeekCallback> getSeekCallbacks() {
@@ -119,7 +101,6 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	/**
 	 * Return the currently registered callbacks and their associated {@link TopicPartition}(s).
-	 *
 	 * @return the map of callbacks and partitions.
 	 * @since 2.6
 	 */
@@ -129,7 +110,6 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	/**
 	 * Seek all assigned partitions to the beginning.
-	 *
 	 * @since 2.6
 	 */
 	public void seekToBeginning() {
@@ -138,7 +118,6 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	/**
 	 * Seek all assigned partitions to the end.
-	 *
 	 * @since 2.6
 	 */
 	public void seekToEnd() {
@@ -147,7 +126,6 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	/**
 	 * Seek all assigned partitions to the offset represented by the timestamp.
-	 *
 	 * @param time the time to seek to.
 	 * @since 2.6
 	 */
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2023 the original author or authors.
+ * Copyright 2019-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,6 +33,7 @@ import org.springframework.lang.Nullable;
  * having to keep track of the callbacks itself.
  *
  * @author Gary Russell
+ * @author Borahm Lee
  * @since 2.3
  *
  */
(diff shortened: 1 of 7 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
```diff
@@ -46,43 +47,59 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	@Override
 	public void registerSeekCallback(ConsumerSeekCallback callback) {
-		this.callbackForThread.put(Thread.currentThread(), callback);
+		if (matchGroupId()) {
+			this.callbackForThread.put(Thread.currentThread(), callback);
```
artembilan, line 51: I'm sorry, I don't see how is this logic different from what you have suggested as an option on the `@KafkaListener`?
I still don't see `groupId` matching though...

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
bky373: You're right. Using a flag to control the behavior is similar. The main difference is that the PR approach don't use the top-level property of @KafkaListener. It avoids the somewhat awkward and unrelated task of having @KafkaListener determine whether to use seek. Instead, the decision of whether to apply seek by group id has been moved to `ConsumerSeekAware`, which is more fitting for this role.

bky373: And `groupId` can be matched like [this](https://github.com/bky373/spring-kafka/commit/b93ff92a4e2e2ec69329783c7bc515494062cd21#diff-d59d8f0f475e848f3ab4a6d6768c513aa1e6b45d61a762cec1ff50fc39164980R107). Am I misunderstanding something?

artembilan: Oh! That is smart!
I think such a logic for the `KafkaUtils.getConsumerGroupId()` has to be hidden from end-user and its result propagated to that `matchGroupId()` method impl.
Or do you have any other use-case in mind?

Thank you for explanation!

spring-kafka/src/test/java/org/springframework/kafka/listener/ConsumerSeekAwareTests.java
```diff
@@ -80,8 +78,7 @@ class CSA extends AbstractConsumerSeekAware {
 		var revoke1 = (Callable<Void>) () -> {
 			if (!first.getAndSet(true)) {
 				csa.onPartitionsRevoked(Collections.singletonList(map1.keySet().iterator().next()));
-			}
-			else {
+			} else {
```
sobychacko, line 81: `else` needs to be on new line.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConsumerSeekAwareTests.java
```diff
@@ -96,8 +93,7 @@ class CSA extends AbstractConsumerSeekAware {
 		var revoke2 = (Callable<Void>) () -> {
 			if (first.getAndSet(false)) {
 				csa.onPartitionsRevoked(Collections.singletonList(map1.keySet().iterator().next()));
-			}
-			else {
+			} else {
```
sobychacko, line 96: `else` needs to be on a new line.

sobychacko (changes requested):

sobychacko:
@bky373 I see that the general idea is what we discussed. We need to add some documentation, explaining how end users can match the group id.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
sobychacko: That's a great point, Artem. Ya, the end users should not worry about calling `KafkaUtils.getConsumerGroupId()`.

bky373: Thanks! It seems like a good idea to hide `KafkaUtils.getConsumerGroupId()` from the user. It's late here, so I'll try to work on this tomorrow or by the end of today.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConsumerSeekAwareTests.java
bky373: Thanks! I'll also fix them.

sobychacko:
@bky373 I think we may need to take a slightly different approach in this PR. While the solution you have in the PR currently satisfies the immediate use cases at hand, this may not be flexible enough for other wider set of requirements. Take a look at this code snippet you provided in the issue. 

```
@Component
    public static class Listener extends AbstractConsumerSeekAware {

        @KafkaListener(id = "seekExample", topics = "seekExample", concurrency = "3")
        public void listen(String payload) {
            System.out.println("Listener received: " + payload);
        }

        @KafkaListener(id = "seekExample3", topics = "seekExample", concurrency = "3")
        public void listen3(String payload) {
            System.out.println("Listener3 received: " + payload);
        }

        public void seekToStart() {
            getSeekCallbacks().forEach((tp, callback) -> callback.seekToBeginning(tp.topic(), tp.partition()));
        }
    }
```

While you register the callbacks based on the new method's boolean return value, which works as expected, this is not flexible enough. What if the user wants to seek other listeners based on some other conditions during runtime? The callback registration happens at startup only once, and the callbacks cannot be re-registered during the runtime. Moreover, the seeks must happen on the listener threads, which the PR is doing currently. A better solution is a hybrid of what we discussed in the [issue](https://github.com/spring-projects/spring-kafka/issues/2302). 

After the `getSeekCallbacks(),` the user needs to do some filtering operations to get the group IDs. Currently, the `ConsumerSeekCallback` returned doesn't give you the group ID. This is where you need to make the core changes so that the `ConsumerSeekCallback` adds a new public API for retrieving the group ID. Could you rework the PR to accommodate this new design? Thanks!

CC @artembilan for any further comments.

bky373:
Okay, I see. I guess we'll need a flexible implementation.
I'll have to think about this a bit more and work on it
Thank you so much for working with me on this.

bky373:
@sobychacko 
I have added a `groupId` getter to `ConsumerSeekCallback`. This approach also minimizes changes, and based on the new tests created, users can utilize it as needed in their preferred manner. Your feedback on this would be greatly appreciated. Thank you.

spring-kafka/src/test/java/org/springframework/kafka/listener/AbstractConsumerSeekAwareTests.java
```diff
@@ -0,0 +1,132 @@
...
+				latch2.countDown();
+			}
+
+			public void seekToBeginningForGroup(String groupIdForSeek) {
+				getCallbacksAndTopics().forEach((cb, topics) -> {
+					if (groupIdForSeek.equals(cb.getGroupId())) {
+						topics.forEach(tp -> cb.seekToBeginning(tp.topic(), tp.partition()));
+					}
+				});
+			}
```
bky373, line 128: I also noticed something awkward when inserting values into `AbstractConsumerSeekAware.callbacks` using `onPartitionsAssigned()`, and I'm not sure if this was intentional. 

Currently, `TopicPartition` serves as the key and `ConsumerSeekCallback` as the value. However, if there are different group IDs, a `TopicPartition` should be allowed to have multiple callbacks. This relates to why I chose to use `getCallbacksAndTopics()` instead of `getSeekCallbacks()`. 

While this issue isn't addressed in the current PR, we can discuss it in another issue or PR. In summary, a beneficial change would be as follows:
Current: `Map<TopicPartition, ConsumerSeekCallback> callbacks`
Proposed: `Map<TopicPartition, List<ConsumerSeekCallback>> callbacks`

bky373: Please take a look this later. https://github.com/spring-projects/spring-kafka/issues/3328

spring-kafka/src/test/java/org/springframework/kafka/listener/AbstractConsumerSeekAwareTests.java
```diff
@@ -0,0 +1,132 @@
+package org.springframework.kafka.listener;
```
sobychacko, line 1: The class has checkstyle violations and missing copyright section, author tag etc.

sobychacko (changes requested):

sobychacko:
@bky373 I took a look at your latest changes. It looks good to me. The changes look very simple, and it also makes the API flexible. I also get the other issue you have encountered, and it is a good idea to work on it in a separate PR. I added some review comments in the new test class.

spring-kafka/src/test/java/org/springframework/kafka/listener/AbstractConsumerSeekAwareTests.java
bky373: Thanks! I fixed it. [a724510](https://github.com/spring-projects/spring-kafka/pull/3318/commits/a724510a8d86588d73863ac7ebe3867c30a60af7)

bky373:
> @bky373 I see that the general idea is what we discussed. We need to add some documentation, explaining how end users can match the group id.

@sobychacko The implementation has changed, but should we add documentation on how to enable seek for only certain groupIds in `seek.adoc`?

sobychacko:
Yes, I think we need to add this in the `seek` section of the ref docs. Also, since we are adding a new public API, we need to mention this in the `whats-new` section.

spring-kafka/src/test/java/org/springframework/kafka/listener/AbstractConsumerSeekAwareTests.java
```diff
@@ -0,0 +1,152 @@
...
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+/**
+ * @author Borahm Lee
+ * @since 3.3
+ */
+@DirtiesContext
+@SpringJUnitConfig
+@EmbeddedKafka(topics = {AbstractConsumerSeekAwareTests.TOPIC}, partitions = 3)
+public class AbstractConsumerSeekAwareTests {
```
sobychacko, line 52: We probably don't need the public modifier here since this is a new test class we are adding and we are using JUnit 5.

spring-kafka/src/test/java/org/springframework/kafka/listener/AbstractConsumerSeekAwareTests.java
```diff
@@ -0,0 +1,152 @@
...
+	Config config;
+
+	@Autowired
+	KafkaTemplate<String, String> template;
+
+	@Autowired
+	MultiGroupListener multiGroupListener;
+
+	@Test
+	public void seekForAllGroups() throws Exception {
```
sobychacko, line 66: We can drop the `public` modifier.

spring-kafka/src/test/java/org/springframework/kafka/listener/AbstractConsumerSeekAwareTests.java
```diff
@@ -0,0 +1,152 @@
...
+		MultiGroupListener.latch1 = new CountDownLatch(2);
+		MultiGroupListener.latch2 = new CountDownLatch(2);
+
+		multiGroupListener.seekToBeginning();
+		assertThat(MultiGroupListener.latch1.await(10, TimeUnit.SECONDS)).isTrue();
+		assertThat(MultiGroupListener.latch2.await(10, TimeUnit.SECONDS)).isTrue();
+	}
+
+	@Test
+	public void seekForSpecificGroup() throws Exception {
```
sobychacko, line 81: We can drop the `public` modifier.

sobychacko (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/listener/AbstractConsumerSeekAwareTests.java
bky373: Thanks. I've dropped them. [d7ce604](https://github.com/spring-projects/spring-kafka/pull/3318/commits/d7ce6044f3962d23507850e64bbb7f0cc6219d86)

bky373:
> Yes, I think we need to add this in the seek section of the ref docs. Also, since we are adding a new public API, we need to mention this in the whats-new section.

@sobychacko 
Please check the documentation! [2f3cdb7](https://github.com/spring-projects/spring-kafka/pull/3318/commits/2f3cdb7ed72fb03ae86abea62a6f4ef885141478)

sobychacko (approved):

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1906,7 +1907,7 @@ private void wrapUp(@Nullable Throwable throwable) {
 				this.consumerSeekAwareListener.onPartitionsRevoked(partitions);
 				this.consumerSeekAwareListener.unregisterSeekCallback();
 			}
-			this.logger.info(() -> getGroupId() + ": Consumer stopped");
+			this.logger.info(() -> KafkaMessageListenerContainer.this.getGroupId() + ": Consumer stopped");
```
artembilan, line 1910: Why don't use that `this.consumerGroupId` instead?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2693,7 +2694,7 @@ private RuntimeException doInvokeRecordListener(final ConsumerRecord<K, V> cReco
 			Observation observation = KafkaListenerObservation.LISTENER_OBSERVATION.observation(
 					this.containerProperties.getObservationConvention(),
 					DefaultKafkaListenerObservationConvention.INSTANCE,
-					() -> new KafkaRecordReceiverContext(cRecord, getListenerId(), getClientId(), getGroupId(),
+					() -> new KafkaRecordReceiverContext(cRecord, getListenerId(), getClientId(), KafkaMessageListenerContainer.this.getGroupId(),
```
artembilan, line 2697: DITTO

artembilan (changes requested):
Please, find some nit-picks from me.
In general it looks good and very close to merge.

Thank you!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
bky373: There is no reason other than to preserve the existing get method. But I think it is okay to use `this.consumerGroupId` as it is initialized when creating the ListenerConsumer and can't be changed. 

I did the same for the override method `getGropuId()` ([link](https://github.com/spring-projects/spring-kafka/pull/3318/files#diff-49bc63455f16e2f0b1fe17ed1ec885135d494883d5a21a3da53d031098d09a33R3333)). Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/ConsumerSeekAware.java
```diff
@@ -39,15 +41,17 @@ public interface ConsumerSeekAware {
 	 * {@code ConcurrentMessageListenerContainer} or the same listener instance in multiple
 	 * containers listeners should store the callback in a {@code ThreadLocal} or a map keyed
 	 * by the thread.
+	 *
```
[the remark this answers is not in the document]
bky373: Yes, I reverted it. Thanks for your guidance!

spring-kafka/src/test/java/org/springframework/kafka/listener/AbstractConsumerSeekAwareTests.java
```diff
@@ -88,8 +88,8 @@ void seekForSpecificGroup() throws Exception {
 		MultiGroupListener.latch2 = new CountDownLatch(2);
 
 		multiGroupListener.seekToBeginningForGroup("group2");
-		assertThat(MultiGroupListener.latch1.getCount()).isEqualTo(2);
 		assertThat(MultiGroupListener.latch2.await(10, TimeUnit.SECONDS)).isTrue();
+		assertThat(MultiGroupListener.latch1.await(100, TimeUnit.MICROSECONDS)).isFalse();
```
bky373, line 92: The previous assert validation was incorrect because it had no wait time, so I fixed it.

bky373, line 92: The previous assert validation was incorrect because it had no wait time, so I fixed it.
Check it please.

Issue #3319, opened 2024-06-19
getUnregisteredListenerContainer return null


aliciz:
version: 3.1.1

I am encountering an issue where the getUnregisteredListenerContainer method from the KafkaListenerEndpointRegistry returns null. This occurs in the scenario where I am not using the @KafkaListener annotation but instead, I am manually registering custom containers.

```java
public void pause(String id) {
        MessageListenerContainer listenerContainer = registry.getUnregisteredListenerContainer(id);
        if (listenerContainer == null) {
            log.warn("listenerContainer not exists id:{}", id);
            return;
        }
        listenerContainer.pause();
}
```

![image](https://github.com/spring-projects/spring-kafka/assets/45386902/16647d07-b955-49e9-bd39-74b780d78a3d)

aliciz:
container define like this:

```java
    @Bean(name = KafkaBeanName.DEVICE_DATA_LISTENER_CONTAINER)
    public ConcurrentMessageListenerContainer<String, DeviceData> deviceDataListenerContainer(@Qualifier(KafkaBeanName.DEFAULT_LISTENER_CONTAINER_FACTORY) ConcurrentKafkaListenerContainerFactory<String, DeviceData> factory, IDeviceDataListener deviceDataListener) {
        String topic = KafkaTopic.getDeviceDataTopicByAppId(ttliotSdkProperties.getAppId());
        log.info("afka container init topic:{}", topic);
        ConcurrentMessageListenerContainer<String, DeviceData> container = factory.createContainer(topic);
        Properties props = container.getContainerProperties().getKafkaConsumerProperties();
        props.put(ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG, ListenerContainerInterceptor.class.getName());
        props.put(JsonDeserializer.VALUE_DEFAULT_TYPE, DeviceData.class.getName());
        container.setupMessageListener(deviceDataListener);
        container.setRecordInterceptor(new DeviceDataRecordInterceptor());
        return container;
    }
```

You are correct; the unregisteredContainers list does contain the container, but getUnregisteredListenerContainer still returns null.

![image](https://github.com/spring-projects/spring-kafka/assets/45386902/e726cbfd-96a5-4fff-924a-192330b4052b)


Can't it be written like this?
```java
@Override
@Nullable
public MessageListenerContainer getUnregisteredListenerContainer(String id) {
  MessageListenerContainer container = this.unregisteredContainers.get(id);
  if (container == null) {
[...]

PR #3320, opened 2024-06-19
GH-3319: Fix NPE in the `KafkaListenerEndpointRegistry`

artembilan:
The `KafkaListenerEndpointRegistry.getUnregisteredListenerContainer()` returns `null` when container is already present in the internal `unregisteredContainers` cache

* Fix `KafkaListenerEndpointRegistry.getUnregisteredListenerContainer()` to return a container instance from the `unregisteredContainers` cache

**Auto-cherry-pick to `3.2.x` & `3.1.x`**


sobychacko:
Do you think we need to update the Javadoc on [ListenerContainerRegistry](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/ListenerContainerRegistry.java#L80) since we are not returning `null` any longer?

artembilan:
It still will be null if there is no such a bean.

PR #3341, opened 2024-07-03
GH-3328: Enhance AbstractConsumerSeekAware with Extended callback for Multi-Group Listeners

### LINKED ISSUE #3328: Mismatch Between Registered and Found Callbacks in Multi-group Listener Cases

For example:

3.2

**Describe the bug**
- When using `AbstractConsumerSeekAware` in a multi-group listener scenario, there are cases where the number of registered callbacks differs from the number of discovered callbacks. Let's consider the examples below.
- **Case 1:**
  - Topic partitions: 1
  - Consumer groups: 2 (group1, group2)
  - Registered callbacks: 2 (group1 thread callback and group2 thread callback)
  - Discovered callbacks: 1 (either group1 thread callback or group2 thread callback — only one callback exists assigned to topic partition 0)

- **Case 2 - with concurrency:**
    - Topic partitions: 2
    - Consumer groups: 2 (group1, group2)
    - Concurrency: 2 per group
    - Registered callbacks: 4
        - group1-0 thread callback, group1-1 thread callback, group2-0 thread callback, group2-1 thread callback
    - Discovered callbacks: 2
        - 1 thread assigned to partition 0 and 1 thread assigned to partition 1 (uncertain whether they belong to group1 or group2 threads)

**Expected behavior**

- The number of discovered callbacks should be equal to the number of registered callbacks.

**Sample**

- [009d8b9](https://github.com/bky373/spring-kafka/commit/009d8b91a916dd14f600697090f77a29623a810c)

bky373:
- When using `AbstractConsumerSeekAware` in a multi-group listeners scenario, there are cases where the number of registered callbacks differs from the number of discovered callbacks.
- This is due to the value type of `callbacks` Map in `AbstractConsumerSeekAware` class being simply `ConsumerSeekCallback`. This causes some callbacks looking at the same partition to be missing.

## Changes
- Change the value type of `callbacks` Map in `AbstractConsumerSeekAware` class from `ConsumerSeekCallback` to `List<ConsumerSeekCallback>`. Also modify some methods, test codes and docs that are affected by this change.
- Add test codes to verify that the callbacks registered via `registeredSeekCallback()` and the ones you can get via `getSeekCallbacks()` match completely.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/seek.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/seek.adoc
@@ -187,16 +187,17 @@ public class SeekToLastOnIdleListener extends AbstractConsumerSeekAware {
     */
     public void rewindAllOneRecord() {
         getSeekCallbacks()
-            .forEach((tp, callback) ->
-                callback.seekRelative(tp.topic(), tp.partition(), -1, true));
+            .forEach((tp, callbacks) ->
+                    callbacks.forEach(callback -> callback.seekRelative(tp.topic(), tp.partition(), -1, true))
+            );
     }
 
     /**
     * Rewind one partition one record.
     */
     public void rewindOnePartitionOneRecord(String topic, int partition) {
-        getSeekCallbackFor(new TopicPartition(topic, partition))
-            .seekRelative(topic, partition, -1, true);
+        getSeekCallbacksFor(new TopicPartition(topic, partition))
+            .forEach(callback -> callback.seekRelative(topic, partition, -1, true));
     }
 
 }
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/seek.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/seek.adoc
@@ -184,25 +184,20 @@ public class SeekToLastOnIdleListener extends AbstractConsumerSeekAware {
 
     /**
     * Rewind all partitions one record.
-    * As of version 3.3, for multi-group listeners, it's recommended to use `getTopicsAndCallbacks()`
-    * in {@link org.springframework.kafka.listener.AbstractConsumerSeekAware}
-    * to handle seeks more accurately across different consumer groups.
     */
     public void rewindAllOneRecord() {
-        getSeekCallbacks()
-            .forEach((tp, callback) ->
-                callback.seekRelative(tp.topic(), tp.partition(), -1, true));
+        getTopicsAndCallbacks()
+            .forEach((tp, callbacks) ->
+                    callbacks.forEach(callback -> callback.seekRelative(tp.topic(), tp.partition(), -1, true))
+            );
     }
 
     /**
     * Rewind one partition one record.
-    * As of version 3.3, for multi-group listeners, it's recommended to use `getSeekCallbacksFor(TopicPartition)`
-    * in {@link org.springframework.kafka.listener.AbstractConsumerSeekAware}
-    * to handle seeks more accurately across different consumer groups.
     */
     public void rewindOnePartitionOneRecord(String topic, int partition) {
-        getSeekCallbackFor(new TopicPartition(topic, partition))
-            .seekRelative(topic, partition, -1, true);
+        getSeekCallbacksFor(new TopicPartition(topic, partition))
+            .forEach(callback -> callback.seekRelative(topic, partition, -1, true));
     }
 
 }
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/seek.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/seek.adoc
@@ -184,20 +184,25 @@ public class SeekToLastOnIdleListener extends AbstractConsumerSeekAware {
 
     /**
     * Rewind all partitions one record.
+    * As of version 3.3, for multi-group listeners, it's recommended to use `getTopicsAndCallbacks()`
+    * in {@link org.springframework.kafka.listener.AbstractConsumerSeekAware}
+    * to handle seeks more accurately across different consumer groups.
     */
     public void rewindAllOneRecord() {
         getSeekCallbacks()
-            .forEach((tp, callbacks) ->
-                    callbacks.forEach(callback -> callback.seekRelative(tp.topic(), tp.partition(), -1, true))
-            );
+            .forEach((tp, callback) ->
+                callback.seekRelative(tp.topic(), tp.partition(), -1, true));
     }
 
     /**
     * Rewind one partition one record.
+    * As of version 3.3, for multi-group listeners, it's recommended to use `getSeekCallbacksFor(TopicPartition)`
+    * in {@link org.springframework.kafka.listener.AbstractConsumerSeekAware}
+    * to handle seeks more accurately across different consumer groups.
     */
     public void rewindOnePartitionOneRecord(String topic, int partition) {
-        getSeekCallbacksFor(new TopicPartition(topic, partition))
-            .forEach(callback -> callback.seekRelative(topic, partition, -1, true));
+        getSeekCallbackFor(new TopicPartition(topic, partition))
+            .seekRelative(topic, partition, -1, true);
     }
 
 }
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
@@ -87,7 +87,6 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	/**
 	 * Return the callbacks for the specified topic/partition.
-	 *
 	 * @param topicPartition the topic/partition.
 	 * @return the callbacks (or null if there is no assignment).
 	 */
@@ -98,7 +97,6 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	/**
 	 * The map of callbacks for all currently assigned partitions.
-	 *
 	 * @return the map.
 	 */
 	protected Map<TopicPartition, List<ConsumerSeekCallback>> getSeekCallbacks() {
@@ -107,7 +105,6 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	/**
 	 * Return the currently registered callbacks and their associated {@link TopicPartition}(s).
-	 *
 	 * @return the map of callbacks and partitions.
 	 * @since 2.6
 	 */
@@ -117,7 +114,6 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	/**
 	 * Seek all assigned partitions to the beginning.
-	 *
 	 * @since 2.6
 	 */
 	public void seekToBeginning() {
@@ -126,7 +122,6 @@ public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	/**
 	 * Seek all assigned partitions to the end.
-	 *
 	 * @since 2.6
 	 */
 	public void seekToEnd() {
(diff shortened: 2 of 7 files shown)


spring-kafka/src/test/java/org/springframework/kafka/listener/AbstractConsumerSeekAwareTests.java
```diff
@@ -62,6 +71,22 @@ class AbstractConsumerSeekAwareTests {
...
+			Map<ConsumerSeekCallback, List<TopicPartition>> callbacksAndTopics = multiGroupListener.getCallbacksAndTopics();
+				Set<ConsumerSeekCallback> registeredCallbacks = callbacksAndTopics.keySet();
+				Set<TopicPartition> registeredTopicPartitions = callbacksAndTopics.values().stream().flatMap(Collection::stream).collect(Collectors.toSet());
+
+				Map<TopicPartition, List<ConsumerSeekCallback>> topicsAndCallbacks = multiGroupListener.getSeekCallbacks();
+				Set<TopicPartition> getTopicPartitions = topicsAndCallbacks.keySet();
+				Set<ConsumerSeekCallback> getCallbacks = topicsAndCallbacks.values().stream().flatMap(Collection::stream).collect(Collectors.toSet());
+
+				assertThat(registeredCallbacks).containsExactlyInAnyOrderElementsOf(getCallbacks).isNotEmpty();
+				assertThat(registeredTopicPartitions).containsExactlyInAnyOrderElementsOf(getTopicPartitions).hasSize(3);
```
bky373: When `concurrency = 2`, if the partitions are all allocated to only the first consumer in the group,
callbacks will be `two`, and `four` if the partitions are allocated split within the group.

spring-kafka/src/test/java/org/springframework/kafka/listener/AbstractConsumerSeekAwareTests.java
```diff
@@ -62,6 +71,22 @@ class AbstractConsumerSeekAwareTests {
...
+			Map<ConsumerSeekCallback, List<TopicPartition>> callbacksAndTopics = multiGroupListener.getCallbacksAndTopics();
+			Set<ConsumerSeekCallback> registeredCallbacks = callbacksAndTopics.keySet();
+			Set<TopicPartition> registeredTopicPartitions = callbacksAndTopics.values().stream().flatMap(Collection::stream).collect(Collectors.toSet());
+
+			Map<TopicPartition, List<ConsumerSeekCallback>> topicsAndCallbacks = multiGroupListener.getSeekCallbacks();
+			Set<TopicPartition> getTopicPartitions = topicsAndCallbacks.keySet();
+			Set<ConsumerSeekCallback> getCallbacks = topicsAndCallbacks.values().stream().flatMap(Collection::stream).collect(Collectors.toSet());
+
+			assertThat(registeredCallbacks).containsExactlyInAnyOrderElementsOf(getCallbacks).isNotEmpty();
+			assertThat(registeredTopicPartitions).containsExactlyInAnyOrderElementsOf(getTopicPartitions).hasSize(3);
```
bky373, line 86: When `concurrency = 2`, if the partitions are all allocated to only the first consumer in the group,
callbacks will be `two`, and `four` if the partitions are allocated split within the group.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
```diff
@@ -82,20 +86,20 @@ public void unregisterSeekCallback() {
...
 	/**
-	 * Return the callback for the specified topic/partition.
+	 * Return the callbacks for the specified topic/partition.
 	 * @param topicPartition the topic/partition.
-	 * @return the callback (or null if there is no assignment).
+	 * @return the callbacks (or null if there is no assignment).
 	 */
 	@Nullable
-	protected ConsumerSeekCallback getSeekCallbackFor(TopicPartition topicPartition) {
+	protected List<ConsumerSeekCallback> getSeekCallbacksFor(TopicPartition topicPartition) {
```
sobychacko, line 94: This is a breaking change. Since there is no easy path for us to deprecate first and then remove later in this case, I suggest calling this in the `whats-new` section.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
```diff
@@ -82,20 +86,20 @@ public void unregisterSeekCallback() {
...
 		return this.callbacks.get(topicPartition);
 	}
 
 	/**
 	 * The map of callbacks for all currently assigned partitions.
 	 * @return the map.
 	 */
-	protected Map<TopicPartition, ConsumerSeekCallback> getSeekCallbacks() {
+	protected Map<TopicPartition, List<ConsumerSeekCallback>> getSeekCallbacks() {
```
sobychacko, line 102: It should be mentioned in the `whats-new` section since this is a breaking change.

sobychacko (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
artembilan: I think it is better to leave an old method yet, even if it is deprecated.
Probably could just return a first item from the list.
This way no breaking changes and the is a migration path.
The name of method are definitely different, therefore no way we would break end-user code.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
artembilan: This one is a bit harder.
May be let's call a new method as `getAllSeekCallbacks()`?
With a `TODO` that it is supposed to be deprecated and removed in the future when we remove the old one?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
bky373: Thanks for your great suggestion!

I'll make the changes and add new method (`getSeekCallbacksFor()`) in the `whats-new` section as you mentioned.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
bky373: `getAllSeekCallbacks()` is good. however, I don't understand a bit the below comment.
> With a TODO that it is supposed to be deprecated and removed in the future when we remove the old one?

Do you mean to add `getAllSeekCallbacks()` first, and then deprecate it later after modifying the `getSeekCallbacks()` method internals?

artembilan: Sorry for not being Cristal clear. So, the plan is:
1. Deprecate existing method
2. Introduce new one with that “All” awkward name.
3. Remove that deprecated method in the next release.
4. Next release Introduce a new method with the desired contract and the name of removed before.
5. Deprecate that “All” awkward method.
6. And then remove it eventually .

This way we have a smooth migration path without breaking changes for target projects .

However I feel like with this releases cycle we would be far in the future where at the same time we may already start a major 4.0 version. And that’s where we can skip some of the backward compatibility steps and have relief with freedom to introduce a breaking change 🤪

bky373: Thanks for explanation.

Actually, Instead of using the old `callbacks` field, I added the `topicToCallbacks` field. 
> [3570f7a](https://github.com/spring-projects/spring-kafka/pull/3341/commits/3570f7a53a6fc9102302b5b80bb1e7610695017c)

This way, not only will existing features work the same, but it will also naturally accommodate new features.

However, this seems to make it more complicated, so please feel free to take a look and let me know what you think. 

Feel free to comment on the javaDocs as well.
The `whats-new` section documentation will be written after the code modifications are complete.

bky373: To summarize, I like your strategy above, but maybe we could make a clearer distinction between `deprecated` and `new` methods. To that end, I did something like this.

- **To be deprecated**
  - `callbacks` field
  - `getSeekCallbackFor()`
  - `getSeekCallbacks()`

- **To be introduced**
  - `topicToCallbacks` field
  - `getSeekCallbacksFor()`
  - `getTopicsAndCallbacks()`

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/seek.adoc
```diff
@@ -184,6 +184,9 @@ public class SeekToLastOnIdleListener extends AbstractConsumerSeekAware {
 
     /**
     * Rewind all partitions one record.
+    * As of version 3.3, for multi-group listeners, it's recommended to use `getTopicsAndCallbacks()`
```
artembilan, line 187: There is no reason to mention this in the docs since the version of the doc is exact match to the code.
So, better to just adjust these samples in the doc to whatever is expected from the code.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
```diff
@@ -33,15 +34,23 @@
...
  */
 public abstract class AbstractConsumerSeekAware implements ConsumerSeekAware {
 
 	private final Map<Thread, ConsumerSeekCallback> callbackForThread = new ConcurrentHashMap<>();
 
+	/*
+	* @deprecated Replaced by the {@link #topicToCallbacks}.
+	* see https://github.com/spring-projects/spring-kafka/pull/3341
+	*/
+	@Deprecated(since = "3.3", forRemoval = true)
```
artembilan, line 49: There is no reason in deprecating `private` props: it is totally OK to rework internals to a new state.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractConsumerSeekAware.java
```diff
@@ -54,7 +63,8 @@ public void onPartitionsAssigned(Map<TopicPartition, Long> assignments, Consumer
 		ConsumerSeekCallback threadCallback = this.callbackForThread.get(Thread.currentThread());
 		if (threadCallback != null) {
 			assignments.keySet().forEach(tp -> {
-				this.callbacks.put(tp, threadCallback);
+				this.callbacks.put(tp, threadCallback); // TODO: This should be removed when the existing `callbacks` field disappears.
```
artembilan, line 66: Therefore no need in this TODO if you just remove that `callbacks` in favor of `topicToCallbacks`.

artembilan (changes requested):
Looks good so far!
Just some minor suggestions to perfect it.
Thanks

bky373:
> Looks good so far! Just some minor suggestions to perfect it. Thanks

@artembilan 
Thanks for your review! 

Also one minor suggestion: I'd like to rename the `callbacksToTopic` Map in `AbstractConsumerSeekAware` to `callbackToTopics`, so that it's consistent with the newly added `topicToCallbacks` Map. (Or should I rename the one I'm adding?) But I'm not sure if it's a good idea to change it, and if so, if I should do it here in PR.

artembilan (changes requested):
The `callbackToTopics` sounds reasonable.

Looks like that's the last concern we have in this PR.

Thanks

bky373:
> The `callbackToTopics` sounds reasonable.
> 
> Looks like that's the last concern we have in this PR.
> 
> Thanks

Thanks! I still have some work like `whats-new` documentation, but It's getting very late where I am, so I'll try to finish it tomorrow, thanks as always.

borahm-lee:
> The `callbackToTopics` sounds reasonable.
> 
> Looks like that's the last concern we have in this PR.
> 
> Thanks

@artembilan 
Sorry for the delay. I've changed the names and added a description to the `whats-new` section. Please check it out!

artembilan (approved):
Pulling locally for final review and merge.

PR #3347, opened 2024-07-08
Gh 3340

LokeshAlamuri:
Code changes to publish an event once all the Kafka MessageListner Container instances are stopped.


spring-kafka/src/main/java/org/springframework/kafka/event/ConcurrentContainerStoppedEvent.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2018-2020 the original author or authors.
```
artembilan, line 2: The date supposed to be current year.

spring-kafka/src/main/java/org/springframework/kafka/event/ConcurrentContainerStoppedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.event;
+
+/**
+ * An event published when a container is stopped.
+ *
+ * @author Gary Russell
+ * @since 2.2
```
artembilan, line 23: Your name must be here instead.
Plus version since is `3.3`

spring-kafka/src/main/java/org/springframework/kafka/event/ConcurrentContainerStoppedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+
+	private static final long serialVersionUID = 1L;
+
+	/**
+	 * Construct an instance with the provided source and container.
+	 * @param source the container instance that generated the event.
+	 * @param container the container or the parent container if the container is a child.
+	 * @since 2.2.1
+	 */
+	public ConcurrentContainerStoppedEvent(Object source, Object container) {
```
artembilan, line 36: This is only for the concurrent container. 
There is just no `parent`.
Therefore no reason in the second argument.
Please, also revise the Javadoc.

spring-kafka/src/main/java/org/springframework/kafka/event/ConcurrentContainerStoppedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+		super(source, container);
+	}
+
+	@Override
+	public String toString() {
+		return "ContainerStoppedEvent [source=" + getSource() + "]";
```
artembilan, line 42: Looks like this class is total copy/paste of existing.
Please, revise the intention of this class.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -704,6 +711,13 @@ protected void publishContainerStoppedEvent() {
 		}
 	}
 
+	protected void publishConcurrentContainerStoppedEvent() {
```
artembilan, line 714: This is not correct.
It has to be only in the `ConcurrentMessageListenerContainer`.
More over I think that event has to be a part of logic of the overridden `publishContainerStoppedEvent()`.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -202,15 +202,24 @@ public boolean isContainerPaused() {
...
-			return false;
-		}
-		for (MessageListenerContainer container : this.containers) {
-			if (container.isRunning()) {
+		this.lifecycleLock.lock();
+		try {
+			if(isRunning()) {
+				return true;
+			}
+			if(this.containers.isEmpty() && this.stoppedContainers.get() < this.concurrency) {
```
artembilan, line 210: I still don't understand this condition: how it is `true` if we don't have containers?

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -392,6 +402,8 @@ && getContainerProperties().isRestartAfterAuthExceptions()
 				exec = new SimpleAsyncTaskExecutor(getListenerId() + ".authRestart");
 			}
 			exec.execute(this::start);
+		} else if (this.concurrency == stoppedContainersCount) {
```
artembilan, line 405: According to code style the `else` supposed to be on a new line.
Please, run `gradlew check` before pushing to the PR.

artembilan (changes requested):
The `ConcurrentContainerStoppedEvent` has to be mentioned in the `whats-new.adoc` and described in the `events.adoc`.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
LokeshAlamuri: containers are cleared once the stop containers is called. But, still the containers will be running in the background. In that scenario, it will be true.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
LokeshAlamuri: As per the earlier discussion, you have suggested not to make any changes related to the 'ContainerStoppedEvent'. So, I have added new method to publish 'ConcurrentContainerStoppedEvent'. I can move code 'publishConcurrentContainerStoppedEvent()' to ConcurrentMessageListenerContainer. But, we can not have same code to publish both 'ContainerStoppedEvent'  and 'ConcurrentMessageListenerContainer'.

artembilan: That's not correct.
To not break the current behavior we still have to emit both events from the `ConcurrentMessageListenerContainer`.
And that new even is really just a responsibility of the `ConcurrentMessageListenerContainer`: don't mix concerns of the encapsulation.

We might just don't need a new method, but override that `publishContainerStoppedEvent()` in the `ConcurrentMessageListenerContainer` to publish `ConcurrentContainerStoppedEvent` as well.

spring-kafka/src/main/java/org/springframework/kafka/event/ConcurrentContainerStoppedEvent.java
LokeshAlamuri: I have updated ConcurrentContainerStoppedEvent as per the comments. Please review.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
artembilan: OK. I'm not sure what you try to achieve here, but looks like you are mixing concerns in the PR.
Can we concentrate on one problem at a time?
Looks like we have agreed about a `ConcurrentContainerStoppedEvent`, but you do something else in the PR.
It is very hard to review.

spring-kafka/src/main/java/org/springframework/kafka/event/ConcurrentContainerStoppedEvent.java
artembilan: Thanks.
There is no need to push the change and comment about it.
I just got to emails from you talking about same stuff: one is commit pushed to the PR and another your comment mentioning just pushed commit.
From my review perspective it is a duplication of the event happening on the PR.
So, if you agree with me, no need to say that you agree: just make a change and push it.

What is more important is to add more info into commit messages.
This way, over time we might extract the context of the change properly.

artembilan (changes requested):
Here is an article I try to follow about commit messages: https://cbea.ms/git-commit/

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
LokeshAlamuri: Please find the following points.

1) I have not made any code changes that will alter the 'ContainerStoppedEvent'.
	
2) In single method we cannot publish both 'ContainerStoppedEvent' and 'ConcurrentContainerStoppedEvent'. **'ConcurrentContainerStoppedEvent' has to be published from the 'childStopped' method in 'org.springframework.kafka.listener.ConcurrentMessageListenerContainer' once all the Containers are stopped.** We cannot publish both.

3) Following events are published from their own methods in KafkaMessageListenerContainer.   
   ConsumerPartitionResumedEvent, ConsumerStoppingEvent, ConsumerStoppedEvent etc....

In similar way, i have written a new method 'publishConcurrentContainerStoppedEvent' in 'org.springframework.kafka.listener.AbstractMessageListenerContainer'. Please let me know if you want me to move this method to 'org.springframework.kafka.listener.ConcurrentMessageListenerContainer'.

Please give your idea.


**Apart from that, i will remove all the changes that are not related to publishing 'ConcurrentContainerStoppedEvent'.**

artembilan: Right. As I said: the `ConcurrentContainerStoppedEvent` is a responsibility only of the `ConcurrentMessageListenerContainer`.
Therefore the method to emit (and respective logic) must be only in the `ConcurrentMessageListenerContainer`.
I'm not sure why you still insist keeping it in the `AbstractMessageListenerContainer`.
This is just wrong by the logic.

Well, if this new event must be emitted from another place, then the method override might not be the case.
So, no need to do any changes to the `publishContainerStoppedEvent()`, but just emit `ConcurrentContainerStoppedEvent` from where you think is appropriate.
However I still don't believe that the place has to be `AbstractMessageListenerContainer`...

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -19,4 +19,8 @@ A new method, `getGroupId()`, has been added to the `ConsumerSeekCallback` inter
 This method allows for more selective seek operations by targeting only the desired consumer group.
 For more details, see xref:kafka/seek.adoc#seek[Seek API Docs].
 
+=== ConcurrentContainerStoppedEvent
```
artembilan, line 22: Has to have a section id. Something like:
```
[x33-concurrent-container-stopped-event]]
```

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -19,4 +19,8 @@ A new method, `getGroupId()`, has been added to the `ConsumerSeekCallback` inter
...
+
+A new event `ConcurrentContainerStoppedEvent` will be published once the parent container i.e ConcurentContainerMessageListenerContainer is stopped.
```
artembilan, line 24: The `ConcurentContainerMessageListenerContainer` must be present as a code snippet.
I also don't find that this sentence is correct.
May be something like:
```
The `ConcurentContainerMessageListenerContainer` emits now a `ConcurrentContainerStoppedEvent` when all of its child containers are stopped.
```

`i.e.` has to be with the period.
And better to not use it at all, but rather the whole phrase you wanted to express here.
We also may go on where exactly comma is supposed to be around this `i.e.`: https://www.bachelorprint.com/commas/comma-before-or-after-ie/

😄

spring-kafka/src/main/java/org/springframework/kafka/event/ConcurrentContainerStoppedEvent.java
artembilan: Looks like this concern has not been addressed yet.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -379,19 +381,30 @@ public void childStopped(MessageListenerContainer child, Reason reason) {
...
+				if (exec == null) {
+					exec = new SimpleAsyncTaskExecutor(getListenerId() + ".authRestart");
+				}
+				exec.execute(this::start);
 			}
-			exec.execute(this::start);
+		}
+	}
+
+	protected void publishConcurrentContainerStoppedEvent() {
```
artembilan, line 404: I don't think this is supposed to be `protected`.
Just `private` will do.

artembilan (changes requested):
I don't see any confirmations of the change with tests.
Are you planning to do that?

Thanks

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
```diff
@@ -142,7 +145,7 @@ protected Consumer<Integer, String> createKafkaConsumer(String groupId, String c
 
 		final CountDownLatch latch = new CountDownLatch(3);
 		final Set<String> listenerThreadNames = new ConcurrentSkipListSet<>();
-		final List<String> payloads = new ArrayList<>();
+		final List<String> payloads = Collections.synchronizedList(new ArrayList<>());
```
artembilan, line 148: Any particular reason to wrap?

artembilan (changes requested):
Thank you for the update!
Looks like we are very close for merging this.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
LokeshAlamuri: We need a synchronized ArrayList. This is shared in all the MessageListener instances.

artembilan: And why is that?
What is the problem with non-synchornized here?
We just add data into the list and then we verify it like this:
```
assertThat(payloads).containsExactlyInAnyOrder("foo", "bar", "qux");
```

So, no matter who and when adds items to the list, we are just save according to our expectations.

artembilan: Might be, but the code is like that for years and we haven't seen any errors yet.
If you say it is possible, I would leave it as is and prefer to see that error eventually.
How this change is relevant to the problem of the PR?
Why make changes which are going to be hard to track in the future?
The contributors come and go, but the code is on us to support, therefore we want to be confident what we are going to deal with.

LokeshAlamuri: It is possible for the error since ArrayList add method is not thread-safe. Please let me know if you want me to change that to plain ArrayList. I will do that.

artembilan: That's correct, but we are not in that situation in this logic. See Javadoc of `ArrayList`:
```
 * <p><strong>Note that this implementation is not synchronized.</strong>
 * If multiple threads access an {@code ArrayList} instance concurrently,
 * and at least one of the threads modifies the list structurally, it
 * <i>must</i> be synchronized externally.  (A structural modification is
 * any operation that adds or deletes one or more elements, or explicitly
 * resizes the backing array; merely setting the value of an element is not
 * a structural modification.)  This is typically accomplished by
 * synchronizing on some object that naturally encapsulates the list.
```
>  merely setting the value of an element is not a structural modification

So, just adding as we do here is OK.

Please, revert this change in favor of the concurrent performance and possible error bubbling in the future for our education.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -374,24 +376,38 @@ protected void doStop(final Runnable callback, boolean normal) {
...
 	public void childStopped(MessageListenerContainer child, Reason reason) {
 		if (this.reason == null || reason.equals(Reason.AUTH)) {
 			this.reason = reason;
 		}
-		if (Reason.AUTH.equals(this.reason)
-				&& getContainerProperties().isRestartAfterAuthExceptions()
-				&& this.concurrency == this.stoppedContainers.incrementAndGet()) {
+		int startedContainersCount = this.startedContainers.decrementAndGet();
+		if (startedContainersCount == 0) {
+			publishConcurrentContainerStoppedEvent();
```
artembilan, line 391: I wonder if we really should emit this event here unconditionally.
Looks like in case of `Reason.AUTH` we restart the container immediately.
Therefore we go to the state when children are running, so `ConcurrentContainerStoppedEvent` is loosing its purpose.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
LokeshAlamuri: **ConcurrentContainer restart happens only when all the containers are stopped**. Whenever all the containers are stopped, then as per the contract ConcurrentContainerStopped event has to be published. 

Please give your comments.

artembilan: That's true, but are they really stopped if we restart it immediately?
That's my concern if we should rethink this situation and don't emit that `ConcurrentContainerStoppedEvent` if we restart immediately.

According to your original request about stopping the whole application context on that event, such a restart might cause some problem.
So, why don't we change the logic to be sure that this is exactly a moment for `ConcurrentContainerStoppedEvent`?

LokeshAlamuri: I got your point. I am good to make change so that `ConcurrentContainerStoppedEvent` is not published when AUTH error occurred. But, I am having one concern and one suggestion.

Concern:
	Whatever may be the scenario, as per the contract it is good to publish `ConcurrentContainerStoppedEvent` when all the containers are stopped.
	
Suggestion:
	Error reason could be added to `ConcurrentContainerStoppedEvent`, so that application user would take a right decision.

public ConcurrentContainerStoppedEvent(Object source, Reason reason)


Please give your idea.

LokeshAlamuri: > but are they really stopped if we restart it immediately?

Yes, childStopped API will be called only when the container  stopped.

artembilan: Yes, that sounds good to have that `Reason` propagated down to the `ConcurrentContainerStoppedEvent`.
Thank you!

LokeshAlamuri: shall we propagate all reasons for each container as below. As of now, only Auth reason is propagated. Is that sufficient.

public ConcurrentContainerStoppedEvent(Object source, ConsumerStoppedEvent.Reason[] reason) -- all reasons will be propagated.

or 

public ConcurrentContainerStoppedEvent(Object source, Reason reason) -- only Auth reason propagated. Other will not be propagated.

artembilan: If we propagate only `Reason.AUTH`, then no reason in this `reason` property for the event.
I don't think we have to propagate all of them.
The logic is like this:
```
	if (this.reason == null || reason.equals(Reason.AUTH)) {
			this.reason = reason;
		}
```
So, any first reason is going to be use (or AUTH) when it overrides the value of that property.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
```diff
@@ -216,6 +222,10 @@ protected Consumer<Integer, String> createKafkaConsumer(String groupId, String c
 					assertThat(children).contains((KafkaMessageListenerContainer<Integer, String>) e.getSource());
 				}
 			}
+			else if (e instanceof ConcurrentContainerStoppedEvent) {
+				assertThat(e.getSource()).isSameAs(container);
+				assertThat(e.getContainer(MessageListenerContainer.class)).isSameAs(container);
```
artembilan, line 227: Can we also assert a new `reason` here, please?

artembilan (changes requested):

artembilan (approved):
After looking into this one more time, I think it is ready for final local review and merge.
I'll address that minor test concern myself.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/1ab37ab7628aaef7931ee97fe9b22b667757f391 after some minor clean up.

@LokeshAlamuri ,

PR #3377, opened 2024-07-19
Fence child containers after ConcurrentContainer stops

LokeshAlamuri:
Containers are not restricted from starting after ConcurrentContainer stopped or restarted. These changes would fix this issue.


spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerFencedException.java
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Exception throw when container is started when it is not a member of
+ * ConcurrentMessageListenerContainer.
```
artembilan, line 19: I'm not sure in this conclusion.
The `KafkaMessageListenerContainer` can be used by itself, without being a member of the `ConcurrentMessageListenerContainer`.

I think we talked about failing to start a child container when its parent is stopped.
Not sure also if we definitely need a dedicated exception on the matter.
I feel like regular `IllegalStateException` is fully enough.
Because indeed we are in the illegal state to attempt to start that child container when its parent is stopped.

What do I miss, please?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerFencedException.java
LokeshAlamuri: From design perspective I thought, It would be better if we can provide a specific exception for this scenario. But, `IllegalStateException` would also works here. I will do that change here if you suggest.

LokeshAlamuri: I think new exception would be benefical for the app developers. It improves the error handling. IllegalStateException would work, but it is really generic. From framework perspective, a seperate exception could be better. Please give your comments. I will modify accordingly.

LokeshAlamuri: > The `KafkaMessageListenerContainer` can be used by itself, without being a member of the `ConcurrentMessageListenerContainer`.
> 
> I think we talked about failing to start a child container when its parent is stopped.

KafkaMessageListenerContainer can be used individually. Given changes does not change any modify in this regard. I feel expected spring-kafka system should be as follows.

Assumption: CMain -- ConcurrentContainer. C0, C1 are the child containers. Choosen concurrency 2.

Flow:

1) CMain -- started. 
   This would inturn starts C0 and C1.
   
   CMain -- running.
   C0 -- running
   C1 -- running.
   
2) C0 stopped. This should be allowed.
   
   CMain -- not running.
   C0 -- not running
   C1 -- running.
   
3) C0 started. This should be allowed. Since, CMain is not stopped explicitly.
   
   CMain -- running.
   C0 -- running
   C1 -- running.

4) C0 stopped. This should be allowed. Since, CMain is not stopped explicitly.
   
   CMain -- not running.
   C0 -- not running
   C1 -- running.

5) C1 stopped. This should be allowed. Since, CMain is not stopped explicitly.
   
   CMain -- not running.
   C0 -- not running
   C1 -- not running.

6) C0 started. This should be allowed. Since, CMain is not stopped explicitly.
   
   CMain -- not running.
   C0 -- running
   C1 -- not running.

7) C1 started. This should be allowed. Since, CMain is not stopped explicitly.
   
   CMain -- running.
   C0 -- running
   C1 -- running.

8) CMain -- stopped explicitly. 
   This would in turn stops C0 and C1. C0 and C1 would be cleared.
   
   CMain -- not running.
   C0 -- not running. delinked
   C1 -- not running. delinked.

9) **C0 started. This should not be allowed. Since, CMain is stopped explicitly. Container Should be Fenced.**
   

10) CMain -- started. 
   This would inturn starts C2 and C3.
   
   CMain -- running.
   C2 -- running
   C3 -- running.

   C0 -- not running. delinked
   C1 -- not running. delinked.

artembilan: The custom exception makes sense if it is intended to be used in the flow control.
In this case it is really abnormal to try to start the container which has not connection with its parent.
Therefore such a logic is wrong and has to be fixed in the target project.
So, `IllegalStateException` would just do whatever we need to indicated is wrong.

I'm not sure in your use-cases explanation, but looking into behavior one more time, I feel like we have to modify `ConcurrentMessageListenerContainer.doStop(final Runnable callback, boolean normal)` to call `destroy()` on all those child containers before `this.containers.clear();`.
That  `destroy()` should call not only `stop();` as it is right now in the `MessageListenerContainer`, but also mark some flag like `destroyed`.
Then when we try to call `start()` on that child container this new flag would give us the mentioned `IllegalStateException` indicating that container cannot be restarted after being destroyed.

Does that make sense to you?

Thank you for being patient with my review!

LokeshAlamuri: Thank you. I have taken the core idea from your comments and modified the logic to solve this issue. I have planned to use `containerAllowedToStart `flag to verify if the container is allowed to start. Please consider. 

I have not used here `destroy` here for the following reasons.

1) `destroy` does not support to take the callback. 
2) A race could occur to occur to start the container immediately after stopped, if destroy is called before clearing the containers and after stopping the containers.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -355,6 +355,7 @@ protected void doStop(final Runnable callback, boolean normal) {
 				}
 			}
 			for (KafkaMessageListenerContainer<K, V> container : this.containers) {
+				container.setAllowedToStart(false);
```
artembilan, line 358: I would really be more specific here and call it `setOrphaned()` (or `fenced`) as you call it in the exception message.

artembilan (changes requested):
I'm OK with the rest of logic.
Thank you!

spring-kafka/src/main/java/org/springframework/kafka/listener/MessageListenerContainer.java
```diff
@@ -290,6 +290,16 @@ default void childStopped(MessageListenerContainer child, ConsumerStoppedEvent.R
 	default void childStarted(MessageListenerContainer child) {
 	}
 
+	/**
+	 * Verify if this container is allowed to start. Eg scenario: If this container has a parent container, and it is stopped,
+	 * implementation of this API could return false to indicate container is not allowed to start.
+	 * @return true if a container is allowed to start.
+	 * @since 2.7.3
+	 */
+	default boolean isAllowedToStart() {
```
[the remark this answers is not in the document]
LokeshAlamuri: > More over that Spring Kafka version is out of support: https://spring.io/projects/spring-kafka#support.
> So, even if we will decide to back-port the fix, it won't make it into that EOL version anyway.

I have taken the changes from origin/main branch. This issue exists in latest code also. Could you please inform why there might be chances that this fix will not be included in the EOL version.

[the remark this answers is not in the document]
LokeshAlamuri: > More over that Spring Kafka version is out of support: https://spring.io/projects/spring-kafka#support. So, even if we will decide to back-port the fix, it won't make it into that EOL version anyway.

My mistake. I have not update the version to 3.3. As per your comments, I have removed the API.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -732,4 +742,10 @@ protected Properties propertiesFromConsumerPropertyOverrides() {
 		return props;
 	}
 
+	public boolean isContainerFenced() {
```
artembilan, line 745: Again: I don't see a reason in any new API like this.
The functionality is totally connected with the `start()` and if you fail there, then something is off in your logic.
Don't give the framework users to make some choice which would be possible with this new public API.
Just throw that `IllegalStateException` from the start.
If you meet such an exception, then something has to be fixed in the target project.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -275,6 +277,14 @@ public boolean isRunning() {
 		return this.running;
 	}
 
+	protected void setFenced(boolean fenced) {
+		this.fenced = fenced;
+	}
+
+	public boolean isFenced() {
```
artembilan, line 284: I even don't see a reason in this one.
That is fully internal logic of the container architecture.
And if you have to deal with this new public API, then the logic in your project is off.
The consumer of the framework must use it in a proper way.
So, any child containers must be controlled outside of the parent container, when that parent container has abandoned them.

Would you sharing with us some idea what is the use-case that we would need to deal with child lifecycle individually, but not via its parent?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -506,7 +512,7 @@ public final void start() {
 		checkGroupId();
 		this.lifecycleLock.lock();
 		try {
-			if (!isRunning()) {
+			if (!isRunning() && this.isContainerFenced()) {
 				Assert.state(this.containerProperties.getMessageListener() instanceof GenericMessageListener,
```
artembilan, line 516: I would just add here the line of code instead:
```
Assert.state(!this.fenced, "Container Fenced. It is not allowed to start.");
```

Otherwise we may debate if the name of that new method is correct and its usage reflect the purpose.

See `Assert.state()` implementation for details.

artembilan (changes requested):

artembilan (approved):
LGTM.
Will be merged when PR build is green.
Thank you for outstanding contribution!

LokeshAlamuri:
> LGTM. Will be merged when PR build is green. Thank you for outstanding contribution!

Thanks a lot!!

Issue #3396, opened 2024-07-29
Fix retry topic documentation for class-level  configuration


artembilan:
This [chapter](https://docs.spring.io/spring-kafka/reference/retrytopic/how-the-pattern-works.html) says:

> At this time this functionality doesn’t support class level `@KafkaListener` annotations. 

Which apparently is not true any more according to the info in some [next chapter](https://docs.spring.io/spring-kafka/reference/retrytopic/retry-config.html):

> Since 3.2, `@RetryableTopic` support for `@KafkaListener` on a class would be:

Plus as you see that `@KafkaListener` has to be  a code snippet.

PR #3397, opened 2024-07-30
GH-3396 : Fix retry topic documentation for class-level configuration

dkswnkk:
Fixes: https://github.com/spring-projects/spring-kafka/issues/3396

This pull request updates the documentation to reflect the support for class-level @KafkaListener annotations in the retry topic configuration, available since version 3.2. It corrects the outdated statement and adds a relevant code example.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/retrytopic/how-the-pattern-works.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/retrytopic/how-the-pattern-works.adoc
@@ -13,22 +13,6 @@ IMPORTANT: By using this strategy you lose Kafka's ordering guarantees for that 
 
 IMPORTANT: You can set the `AckMode` mode you prefer, but `RECORD` is suggested.
 
-IMPORTANT: This functionality now supports class level `@KafkaListener` annotations since version 3.2.
-
-[source, java]
-----
-@RetryableTopic(listenerContainerFactory = "my-retry-topic-factory")
-@KafkaListener(topics = "my-annotated-topic")
-public class ClassLevelRetryListener {
-
-    @KafkaHandler
-    public void processMessage(MyPojo message) {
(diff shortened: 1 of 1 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/retrytopic/how-the-pattern-works.adoc
```diff
@@ -13,7 +13,21 @@ IMPORTANT: By using this strategy you lose Kafka's ordering guarantees for that
 
 IMPORTANT: You can set the `AckMode` mode you prefer, but `RECORD` is suggested.
 
-IMPORTANT: At this time this functionality doesn't support class level `@KafkaListener` annotations.
+IMPORTANT: This functionality now supports class level `@KafkaListener` annotations since version 3.2.
```
artembilan, line 16: Thank you for looking into this!
However  I believe there is not need in this section here at all.
It is really mentioned and explained in the next chapter just after this intro: https://docs.spring.io/spring-kafka/reference/retrytopic/retry-config.html#using-the-retryabletopic-annotation.

So, please, just remove this `IMPORTANT` and that's it.
You can wrap that `@KafkaListener` in the next chapter into a code snippet though 😄

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/retrytopic/how-the-pattern-works.adoc
dkswnkk: Ah, I see, it's clearly explained in the next chapter, but I wrote it redundantly 😅
Thank you for the quick feedback!

artembilan:
@dkswnkk ,

thank you for contribution!
Will be merged after greed PR build.
Looking forward for more!

Issue #3402, opened 2024-08-02
KafkaAdmin clusterId configuration is ignored when observability is enabled and bootstrap supplier is not set


varmenise:
Version: org.springframework.kafka:spring-kafka:jar:3.2.0

**Describe the bug**

If you enable observability and define clutserId inside the KafkaAdmin configuration, unless you specify an observability supplier, which matches the producer bootstrap server, the code will re-create a KafkaAdmin passing the defined properties as well as the producer bootstrap config.

```
		if (this.kafkaAdmin != null) {
			Object producerServers = this.producerFactory.getConfigurationProperties()
					.get(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG);
			String adminServers = this.kafkaAdmin.getBootstrapServers();
			if (!producerServers.equals(adminServers)) {
				Map<String, Object> props = new HashMap<>(this.kafkaAdmin.getConfigurationProperties());
				props.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, producerServers);
				int opTo = this.kafkaAdmin.getOperationTimeout();
				this.kafkaAdmin = new KafkaAdmin(props);
				this.kafkaAdmin.setOperationTimeout(opTo);
			}
		}
```
See https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java#L493

Unfortunately if you have specified a cluster id with 

`kafkaAdmin.setClusterId(clusterId);`

this will not be part of the property map, so the code above will basically override this configuration and clusterId will be null. 

**To Reproduce**

1) enable observability
2) create a KafkaAdminConfiguration and specify bootstrap property or no property at all, just a clusterId

    @Bean
```
    public KafkaAdmin kafkaAdmin() {
        Map<String, Object> configs = new HashMap<>();
        configs.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServer);
        KafkaAdmin kafkaAdmin = new KafkaAdmin(configs);
        kafkaAdmin.setClusterId(clusterId);
        //kafkaAdmin.setBootstrapServersSupplier(() -> bootstrapServer);
        return kafkaAdmin;
    }
```
3) send an event

**Expected behavior**

A kafka admin with properties, bootstrap and clusterId is set whereas clusterId is null.

PR #3403, opened 2024-08-02
GH-3402: kafkaAdmin clusterId configuration is ignored if observability is enabled and bootstrap supplier is not set

varmenise:
fixes GH-3402 (https://github.com/spring-projects/spring-kafka/issues/3402)

* Re-set clusterId after creating new KafkaAdmin

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
@@ -213,6 +213,14 @@ public class KafkaAdmin extends KafkaResourceFactory
 		this.clusterId = clusterId;
 	}
 
+	/**
+	 * Get the clusterId property.
+	 * @since 3.3.0
+	 */
+	public String getClusterId() {
+		return clusterId;
+	}
+
 	@Override
 	public Map<String, Object> getConfigurationProperties() {
 		Map<String, Object> configs2 = new HashMap<>(this.configs);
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
Any chances to prove the problem with some unit test, please ?

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -213,6 +213,14 @@ public void setClusterId(String clusterId) {
 		this.clusterId = clusterId;
 	}
 
+	/**
+	 * Get the clusterId property.
+	 * @since 3.3.0
+	 */
+	public String getClusterId() {
+		return clusterId;
```
artembilan, line 221: Requires `this.`

artembilan (changes requested):
I think I'm OK with this.
Although some unit test to cover would be nice.
Or modify existing one `ObservationIntegrationTests`.

Please, also run `./gradlew check` locally to be sure in all the passing tests and Checkstyle rules.

Then add you name to the `@author` list of the affected classes.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -213,6 +213,14 @@ public void setClusterId(String clusterId) {
 		this.clusterId = clusterId;
 	}
 
+	/**
+	 * Get the clusterId property.
+	 * @since 3.3.0
```
artembilan, line 218: Since we have agreed that this is a bug and there is no easy workaround (but fix is simple), then this one is going to be back-ported to supported versions. Therefore this `@since` is `3.1.8`

artembilan (changes requested):

varmenise:
I ll fix this as I get some time, thanks for reviewing!

artembilan:
@varmenise ,
gentle reminder.
We have a couple days left before release on Monday.

Thanks

sobychacko:
Merged via: https://github.com/spring-projects/spring-kafka/commit/f11c7caf081e25c8a513c888f22b9144b07a9404

Backporte to `3.2.x` an `3.1.x`.

varmenise:
@sobychacko @artembilan  Thanks for finalizing this and keeping my name on it.
Sorry I missed your earlier messages; I was away for a while.
Appreciate the collaboration!

Issue #3395, opened 2024-07-28
Use the original key in reply messages when using ReplyingKafkaTemplate and SendTo


neshkeev:
**Expected Behavior**

Response messages in addition to the `kafka_correlationId` Header should have the same keys as their corresponding request messages when constructing the Request-Reply semantics on top of `ReplyingKafkaTemplate` and `SendTo`.

**Current Behavior**

The response messages have no keys (`null`)

**Context**

When one sends a message/request it's only natural to expect that the response message have same key as the original message (request). Such an explicit formalization can allow engineers to reason about the workload. Also it might come in handy when there are different consumers listening the reply topic.

**Here is a brief example of what this issue is about**

Sender:
```java
void sender(ReplyingKafkaTemplate<String, String, String> template) {
    final ProducerRecord<String, String> record = new ProducerRecord<>(
            "mytopic",
            "key",
            "value"
    );

    var replay = template.sendAndReceive(record, responseTimeout);

    // THIS ASSERTION FAILS
    assert "key".equals(replay.get().key()) : "The response should have the same key";
}
```

Listener:
```java
@SendTo
@Listener(topics="mytopic")
String listen(String value) {
    return "Hello, " + value;
}
```

If you run this snippet, you will get the assertion error, because the key of the replying message is null.

artembilan:
This makes sense.
Feel free to contribute the fix: https://github.com/spring-projects/spring-kafka/blob/main/CONTRIBUTING.adoc
Thanks

sobychacko:
I am taking a look at this issue. It's straightforward with a single return record, as in the example above. I wonder what the semantics are if the return type is a batch or a collection. In the case of a collection, the framework may not detect what keys to provide, and it is up to the application to provide the keys by wrapping them in a `Message.` What do you think?

PR #3404, opened 2024-08-02
GH-3395: Use Original Key in Reply Record

sobychacko:
Fixes: #3395

When using `ReplyingKafkaTemplate,` include the original key from the request record if such a key exists.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/sending-messages.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/sending-messages.adoc
@@ -650,8 +650,7 @@ public Message<?> messageReturn(String in) {
 }
 ----
 
-[[record-key-in-reply]]
-== Preserving Request Record Key in Reply
+=== Original Record Key in Reply
 
 Starting with version 3.3, the Kafka record key from the incoming request (if it exists) will be preserved in the reply record.
 This is only applicable for single record request/reply scenario.
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/sending-messages.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/sending-messages.adoc
@@ -650,6 +650,13 @@ public Message<?> messageReturn(String in) {
 }
 ----
 
+[[record-key-in-reply]]
+== Preserving Request Record Key in Reply
+
+Starting with version 3.3, the Kafka record key from the incoming request (if it exists) will be preserved in the reply record.
+This is only applicable for single record request/reply scenario.
+When the listener is batch or when the return type is a collection, it is up to the application to specify which keys to use by wrapping the reply record in a `Message` type.
+
 [[aggregating-request-reply]]
 == Aggregating Multiple Replies
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
@@ -644,6 +644,7 @@ public abstract class MessagingMessageListenerAdapter<K, V> implements ConsumerS
 
 	protected void asyncSuccess(@Nullable Object result, String replyTopic, Message<?> source,
 			boolean returnTypeMessage) {
+
 		if (result == null) {
 			if (this.logger.isDebugEnabled()) {
 				this.logger.debug("Async result is null, ignoring");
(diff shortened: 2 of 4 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/sending-messages.adoc
```diff
@@ -650,6 +650,13 @@ public Message<?> messageReturn(String in) {
 }
 ----
 
+[[record-key-in-reply]]
+== Preserving Request Record Key in Reply
```
artembilan, line 654: Why do we need a dedicated section for this single short paragraph?
Why it cannot be included in the end if the previous one?
At least it looks like both of them talks about similar context.
Where that `Message` indeed covers what you try to say about multi-value reply.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -638,12 +638,12 @@ private void sendReplyForMessageSource(Object result, String topic, Message<?> s
 			builder.setHeader(this.correlationHeaderName, correlationId);
 		}
 		setPartition(builder, source);
+		setKey(builder, source);
 		this.replyTemplate.send(builder.build());
 	}
 
 	protected void asyncSuccess(@Nullable Object result, String replyTopic, Message<?> source,
 			boolean returnTypeMessage) {
-
```
artembilan, line 646: This is really a good style to make code more readable.
There is might no reason to do that for single-like method header, but if it is multi-line, then better to separate body for easy review.
Please, revert this change and fix your IDE to preserver such a blank line on refactoring.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -719,11 +717,24 @@ private void setPartition(MessageBuilder<?> builder, Message<?> source) {
...
+	}
+
 	@Nullable
 	private byte[] getReplyPartition(Message<?> source) {
 		return source.getHeaders().get(KafkaHeaders.REPLY_PARTITION, byte[].class);
 	}
 
+	@Nullable
+	private Object getReplyKeyFromRequest(Message<?> source) {
+		return source.getHeaders().get(KafkaHeaders.RECEIVED_KEY);
```
artembilan, line 735: I don't see a reason in this one-line method.
More over it feel like `static`.
Way, either way: I'd prefer to not have it making me thinking "where else this method can be used" 😄

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -662,7 +663,6 @@ protected void acknowledge(@Nullable Acknowledgment acknowledgment) {
 
 	protected void asyncFailure(Object request, @Nullable Acknowledgment acknowledgment, Consumer<?, ?> consumer,
 			Throwable t, Message<?> source) {
-
```
artembilan, line 665: Why haven't you reverted this?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -676,7 +676,6 @@ protected void asyncFailure(Object request, @Nullable Acknowledgment acknowledgm
 
 	protected void handleException(Object records, @Nullable Acknowledgment acknowledgment, Consumer<?, ?> consumer,
 			Message<?> message, ListenerExecutionFailedException e) {
-
```
artembilan, line 679: DITTO, please.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/sending-messages.adoc
```diff
@@ -650,6 +650,12 @@ public Message<?> messageReturn(String in) {
 }
 ----
 
+=== Original Record Key in Reply
```
artembilan, line 653: Why do you still think that this short paragraph deserves its own title?

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/sending-messages.adoc
sobychacko: I don't see a problem with a separate section for this, but I am neutral either way.

Issue #3338, opened 2024-06-29
ConcurrentMessageListenerContainer isChildRunning API is returning false even though active MessageListenerContainer instances are processing messages.


LokeshAlamuri:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.1.3

**Application**

SpringBoot application uses spring-kafka module to consume and process the messages. MessageListenerContainer concurrency is 4. CommonContainerStoppingErrorHandler is chosen to stop the Container on any failures. Assume it would take 60 seconds to complete processing any message.

**Describe the bug**

One of the MessageListenerContainer instance throws a fatal exception. CommonContainerStoppingErrorHandler would stop the MessageListenerContainerContainer. After this point, we if try to check the MessageListenerContainer instances state immediately using following code snippet, it would show no MessageListenerContainer instances are running. Because, the state is updated prematurely as not running.

**code:**
org.springframework.kafka.listener.ListenerContainerRegistry listenerContainerRegistry;

listenerContainerRegistry.getListenerContainers().stream().
                filter(messageListenerContainer -> messageListenerContainer.isChildRunning()).count() -- **This is returning (zero)0. Ideally it should return zero only after stopping processing of messages in all the instances.**

Ideally, it would take some time to properly complete processing of the messages in other MessageListenerContainer instances.

**To Reproduce**

Any spring-kafka consumer application

**Expected behavior**

org.springframework.kafka.listener.ListenerContainerRegistry listenerContainerRegistry;

listenerContainerRegistry.getListenerContainers().stream().
                filter(messageListenerContainer -> messageListenerContainer.isChildRunning()).count() -- This should not return (zero)0.

The above code should return zero after receiving closed notification from all the MessageListenerContainer instances.

**Problem**
From the application perspective, state of the Kafka-Consumer component is reported incorrectly. This is a great problem if at all we plan to shutdown the application based on the reported state. But because of this bug, application would be stopped prematurely. 

As of now, MessageListnerContainer APIs are not reporting the right status regarding when exactly all the MessageListnerContainer instances are completed processing. 

**Sample**

https://github.com/LokeshAlamuri/SpringBootKafkaConsumerDemo/blob/main/src/main/java/com/example/springboot/ConsumerStoppedEventListener.java

**Root Cause** 

As per my analysis, I could see that 'isChildRunning' API is returning false based on the org.springframework.kafka.listener.ConcurrentMessageListenerContainer 
[...]

Issue #3371, opened 2024-07-13
Fence the MessageLisnterContainer restart once the ConcurrentMessageListenerContainer is stopped


LokeshAlamuri:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.3

**Describe the bug**

1) Start the `ConcurrentMessageListenerContainer` with concurrency 2.
2) Obtain one of the `KafkaMessageListenerContainer` from the `ConcurrentMessageListenerContainer`.
3) Stop `KafkaMessageListenerContainer`.
4) Stop `ConcurrentMessageListenerContainer`
5) Start previous `KafkaMessageListenerContainer`. **This container would start running even though `ConcurrentMessageListenerContainer` has stopped. It is even holding the reference of `ConcurrentMessageListenerContainer`.**

**To Reproduce**

I have modified the Junit in `org.springframework.kafka.listener.ConcurrentMessageListenerContainerTests` to reproduce.

        `@Test
	public void testAutoCommit() throws Exception {
		this.logger.info("Start auto");
		Map<String, Object> props = KafkaTestUtils.consumerProps("test1", "true", embeddedKafka);
		AtomicReference<Properties> overrides = new AtomicReference<>();
		DefaultKafkaConsumerFactory<Integer, String> cf = new DefaultKafkaConsumerFactory<Integer, String>(props) {
			@Override
			protected Consumer<Integer, String> createKafkaConsumer(String groupId, String clientIdPrefix,
					String clientIdSuffixArg, Properties properties) {

				overrides.set(properties);
				return super.createKafkaConsumer(groupId, clientIdPrefix, clientIdSuffixArg, properties);
			}

		};
		ContainerProperties containerProps = new ContainerProperties(topic1);
		containerProps.setLogContainerConfig(true);
		containerProps.setClientId("client");

		final CountDownLatch latch = new CountDownLatch(3);
		final Set<String> listenerThreadNames = new ConcurrentSkipListSet<>();
		final List<String> payloads = new ArrayList<>();
		containerProps.setMessageListener((MessageListener<Integer, String>) message -> {
			ConcurrentMessageListenerContainerTests.this.logger.info("auto: " + message);
			listenerThreadNames.add(Thread.currentThread().getName());
			payloads.add(message.value());
			latch.countDown();
		});

		ConcurrentMessageListenerContainer<Integer, String> container =
				new ConcurrentMessageListenerContainer<>(cf, containerProps);
		container.setConcurrency(2);
[...]

PR #3406, opened 2024-08-03
Fix ConcurrentContainer lifecycle issues

LokeshAlamuri:
This commit would fix the [issue](https://github.com/spring-projects/spring-kafka/issues/3338).

1) 'isChildRunning' API would return true only after all the containers are actually stopped. 
2) Add 'stopAbnormally' in a Lock.
3) Call `childStarted` in ConcurrentContainer from KafkaMessageListenerContainer right before publishing ConsumerStartedEvent.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -71,6 +71,7 @@ import org.springframework.util.StringUtils;
  * @author Wang Zhiyang
  * @author Soby Chacko
  * @author Sanghyeok An
+ * @author Lokesh Alamuri
  */
 public abstract class AbstractMessageListenerContainer<K, V>
 		implements GenericMessageListenerContainer<K, V>, BeanNameAware, ApplicationEventPublisherAware,
@@ -275,10 +276,6 @@ public abstract class AbstractMessageListenerContainer<K, V>
 
 	@Override
 	public boolean isRunning() {
-		return this.running;
-	}
-
-	protected boolean canStop() {
 		return this.running;
 	}
 
@@ -614,7 +611,7 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	public final void stop(boolean wait) {
 		this.lifecycleLock.lock();
 		try {
-			if (canStop()) {
+			if (isRunning()) {
 				if (wait) {
 					final CountDownLatch latch = new CountDownLatch(1);
 					doStop(latch::countDown);
@@ -650,7 +647,7 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	public void stop(Runnable callback) {
 		this.lifecycleLock.lock();
 		try {
-			if (canStop()) {
+			if (isRunning()) {
 				doStop(callback);
 			}
 			else {
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -278,6 +278,10 @@ public abstract class AbstractMessageListenerContainer<K, V>
 		return this.running;
 	}
 
+	protected boolean canStop() {
+		return this.running;
+	}
+
 	protected void setFenced(boolean fenced) {
 		this.fenced = fenced;
 	}
@@ -610,7 +614,7 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	public final void stop(boolean wait) {
 		this.lifecycleLock.lock();
 		try {
-			if (isRunning()) {
+			if (canStop()) {
 				if (wait) {
 					final CountDownLatch latch = new CountDownLatch(1);
 					doStop(latch::countDown);
@@ -646,7 +650,7 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	public void stop(Runnable callback) {
 		this.lifecycleLock.lock();
 		try {
-			if (isRunning()) {
+			if (canStop()) {
 				doStop(callback);
 			}
 			else {
@@ -660,8 +664,14 @@ public abstract class AbstractMessageListenerContainer<K, V>
 
 	@Override
 	public void stopAbnormally(Runnable callback) {
-		doStop(callback, false);
-		publishContainerStoppedEvent();
+		this.lifecycleLock.lock();
+		try {
+			doStop(callback, false);
+			publishContainerStoppedEvent();
+		}
+		finally {
+			this.lifecycleLock.unlock();
+		}
 	}
 
 	protected void doStop(Runnable callback) {
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
@@ -213,7 +213,7 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 					return true;
 				}
 			}
-			if ((!isRunning() && this.startedContainers.get() > 0)) {
+			if (this.startedContainers.get() > 0) {
 				return true;
 			}
 		}
@@ -243,7 +243,7 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 	 */
 	@Override
 	protected void doStart() {
-		if (!isRunning() && this.containers.stream().allMatch(container -> !container.isRunning())) {
+		if (!isRunning()) {
 			checkTopics();
 			ContainerProperties containerProperties = getContainerProperties();
 			TopicPartitionOffset[] topicPartitions = containerProperties.getTopicPartitions();
@@ -351,7 +351,7 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 	@Override
 	protected void doStop(final Runnable callback, boolean normal) {
 		final AtomicInteger count = new AtomicInteger();
-		if (canStop()) {
+		if (isRunning()) {
 			boolean childRunning = isChildRunning();
 			setRunning(false);
 			if (!childRunning) {
@@ -392,7 +392,6 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 		this.lifecycleLock.lock();
 		try {
 			if (this.containers.contains(child) || this.stoppedContainers.contains(child)) {
-				setRunning(!this.stoppedContainers.contains(child));
 				this.startedContainers.incrementAndGet();
 			}
 		}
@@ -408,7 +407,6 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 			if (!this.containers.contains(child) && !this.stoppedContainers.contains(child)) {
 				return;
 			}
-			setRunning(false);
 			if (this.reason == null || reason.equals(Reason.AUTH)) {
 				this.reason = reason;
 			}
@@ -549,11 +547,6 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 		this.startedContainers.set(0);
 		this.stoppedContainers.clear();
 		this.reason = null;
-	}
-
-	@Override
-	protected boolean canStop() {
-		return !this.containers.isEmpty() || isRunning();
 	}
 
 	@Override
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
@@ -380,6 +380,10 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 				}
 			}
 			setStoppedNormally(normal);
+			// All the containers are stopped before calling stop API
+			if (this.startedContainers.get() == 0) {
+				publishConcurrentContainerStoppedEvent(this.reason);
+			}
 		}
 	}
 
@@ -408,7 +412,10 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 			}
 			int startedContainersCount = this.startedContainers.decrementAndGet();
 			if (startedContainersCount == 0) {
-				publishConcurrentContainerStoppedEvent(this.reason);
+				if (!isRunning()) {
+					this.containers.clear();
+					publishConcurrentContainerStoppedEvent(this.reason);
+				}
 				boolean restartContainer = Reason.AUTH.equals(this.reason)
 						&& getContainerProperties().isRestartAfterAuthExceptions();
 				this.reason = null;
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
@@ -65,6 +65,8 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 
 	private final List<KafkaMessageListenerContainer<K, V>> containers = new ArrayList<>();
 
+	private final List<KafkaMessageListenerContainer<K, V>> stoppedContainers = new ArrayList<>();
+
 	private final List<AsyncTaskExecutor> executors = new ArrayList<>();
 
 	private final AtomicInteger startedContainers = new AtomicInteger();
@@ -204,13 +206,19 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 
 	@Override
 	public boolean isChildRunning() {
-		if (!isRunning()) {
-			return false;
-		}
-		for (MessageListenerContainer container : this.containers) {
-			if (container.isRunning()) {
+		this.lifecycleLock.lock();
+		try {
+			for (MessageListenerContainer container : this.containers) {
+				if (container.isRunning()) {
+					return true;
+				}
+			}
+			if ((!isRunning() && this.startedContainers.get() > 0)) {
 				return true;
 			}
+		}
+		finally {
+			this.lifecycleLock.unlock();
 		}
 		return false;
 	}
@@ -235,7 +243,7 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 	 */
 	@Override
 	protected void doStart() {
-		if (!isRunning()) {
+		if (!isRunning() && this.containers.stream().allMatch(container -> !container.isRunning())) {
 			checkTopics();
 			ContainerProperties containerProperties = getContainerProperties();
 			TopicPartitionOffset[] topicPartitions = containerProperties.getTopicPartitions();
@@ -245,7 +253,7 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 						+ topicPartitions.length);
 				this.concurrency = topicPartitions.length;
 			}
-			this.startedContainers.set(0);
+			clearState();
 			setRunning(true);
 
 			for (int i = 0; i < this.concurrency; i++) {
@@ -343,7 +351,7 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 	@Override
 	protected void doStop(final Runnable callback, boolean normal) {
 		final AtomicInteger count = new AtomicInteger();
-		if (isRunning()) {
+		if (canStop()) {
 			boolean childRunning = isChildRunning();
 			setRunning(false);
 			if (!childRunning) {
@@ -372,6 +380,7 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 						});
 					}
 				}
+				this.stoppedContainers.add(container);
 			}
 			this.containers.clear();
 			setStoppedNormally(normal);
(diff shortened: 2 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -277,6 +277,10 @@ public boolean isRunning() {
 		return this.running;
 	}
 
+	protected boolean canStop() {
+		return this.running;
```
artembilan, line 281: Why `isRunning()` not enough?
It is totally OK to have an extra logic in the overridden method in that `ConcurrentMessageListenerContainer`.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -204,14 +206,20 @@ public boolean isContainerPaused() {
...
-		for (MessageListenerContainer container : this.containers) {
-			if (container.isRunning()) {
+		this.lifecycleLock.lock();
+		try {
+			for (MessageListenerContainer container : this.containers) {
+				if (container.isRunning()) {
+					return true;
+				}
+			}
+			if ((!isRunning() && this.startedContainers.get() > 0)) {
```
artembilan, line 216: I don't understand this extra logic.
The purpose of the `isChildRunning()` is to report if any child container is active at the moment.
Why do we need to check `!this.isRunning()` and then number of started containers?
Didn't we discuss with your in other PR (#3377 ) that it is abnormal to have child container running and stopped parent one?

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -235,7 +243,7 @@ public boolean isChildRunning() {
 	 */
 	@Override
 	protected void doStart() {
-		if (!isRunning()) {
+		if (!isRunning() && this.containers.stream().allMatch(container -> !container.isRunning())) {
```
artembilan, line 246: This again sounds like a contradiction to what we have with fenced child container.
Why would one be running if parent is stopped?

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerGroupSequencer.java
```diff
@@ -152,24 +152,22 @@ public synchronized void onApplicationEvent(ListenerContainerIdleEvent event) {
 	}
 
 	private synchronized void stopParentAndCheckGroup(MessageListenerContainer parent) {
-		if (parent.isRunning()) {
```
artembilan, line 155: Why do you remove this condition?

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerTests.java
```diff
@@ -816,4 +816,270 @@ private void testAckOnErrorWithManualImmediateGuts(String topic, boolean ackOnEr
...
+	@Test
+	public void testIsChildRunning() throws Exception {
+		this.logger.info("Start isChildRunning");
+		Map<String, Object> props = KafkaTestUtils.consumerProps("test1", "true", embeddedKafka);
+		AtomicReference<Properties> overrides = new AtomicReference<>();
+		DefaultKafkaConsumerFactory<Integer, String> cf = new DefaultKafkaConsumerFactory<Integer, String>(props) {
+
+			@Override
+			protected Consumer<Integer, String> createKafkaConsumer(String groupId, String clientIdPrefix,
+																	String clientIdSuffixArg, Properties properties) {
```
artembilan, line 828: This does not look like OK code formatting?
Can you revise it, please, so it looks nice from review perspective?

artembilan (changes requested):
> ConcurrentContainer start would be permitted only after all the containers running status is false.

Would you mind to revise the logic in a way that they are idempotent?
So, if `start()` has been called before that does not mean that we cannot call it again.
Same for `stop()`.

Probably if you think about these lifecycle hooks as idempotent operations, then the logic would e much simple. Or the problem will go away at all.

I also curious how this fix is correlated with your fenced one before.
At a glance they contradict each other.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
LokeshAlamuri: Assume concurrency == 2.
Cmain -- concurrent container
C0, C1 -- child containers
   
1. `start` the concurrent container. 

   Cmain -- running
   C0 -- running
   C1 -- running
2. `stop` container C0 manually.
   
   Cmain -- not running
   C0 -- not running
   C1 -- running
   
3. `stop` 'Cmain' container. 
   
   As per the earlier condition concurrent container running status is false. So, it will not be stopped. 
   
   New condition has to be added to verify if really child containers are cleared. This is equal to `stop` is called prior or not.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
LokeshAlamuri: This condition is to prevent `run` call on ConcurrentContainer before stopping the child containers. This is to verify if really all the containers are stopped.

This is similar to earlier logic. If container is in running status , subsequent `run` calls will be ignored.

Earlier, running status of the concurrent container is not set to false, even if one of the container is stopped. I have made the changes to set  concurrent container running status to false, even if one of the container is stopped. 
   
  In this case, condition to verify if concurrent is allowed to start needs to be changed. If all the containers running status is set to 'false' then only `run` is allowed.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
artembilan: > stop container C0 manually.

> Cmain -- not running
> C0 -- not running
> C1 -- running

But this situation is not correct.
The `Cmain` has to be running until any of its children is running.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
artembilan: > In this case, condition to verify if concurrent is allowed to start needs to be changed. If all the containers running status is set to 'false' then only run is allowed.

That's also not what I think about this concurrent container logic.
It is in running state when any of its child containers running.
When we start concurrent container, it suppose to start all of its children.
If some of them are running already, then idempotent.
As we state before: we just don't allow orphaned child containers to be restarted.
Probably much robust solution is to stop all the children when we call start of their parent.
This way any new start would give us a fresh state. Kinda total renew.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
LokeshAlamuri: The condition in the Line:216 is the critical logic. `running` status of the ConcurrentContainer will be set to `false` for the following conditions.

1) ConcurrentContainer itself is stopped manually.
2) One of the child container is stopped for any reason.
 
`startedContainers` count will be decremented only when the actual container exits. So, i have verified if the `ConcurrentContainer` is stopped and it is having any child containers processing still messages. In this case, it should return `true`.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerGroupSequencer.java
LokeshAlamuri: I have removed the condition to verify `running` is true. That is not correct, since the status of the `ConcurrentContainer` would be set to `false` if one of the container is stopped. Earlier, it worked since `running` status of the `ConcurrentContainer` is not set to `false` if any of the container is stopped and it is not set to true after it is started.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
LokeshAlamuri: But as per the API definition,

From interface:
org.springframework.context.Lifecycle Check whether this component is currently running.
In the case of a container, this will return true only if all components that apply are currently running

LokeshAlamuri: > The `Cmain` has to be running until any of its children is running.

`Cmain` ConcurrentContainer is not stopped. Only it's running status will be set to `false` to indicate that one or all containers are stopped as per the API definition.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
LokeshAlamuri: > When we start concurrent container, it suppose to start all of its children.
> If some of them are running already, then idempotent.
> As we state before: we just don't allow orphaned child containers to be restarted.
> Probably much robust solution is to stop all the children when we call start of their parent.
> This way any new start would give us a fresh state. Kinda total renew.

This is already happening. I have not modified any thing here.

LokeshAlamuri: > That's also not what I think about this concurrent container logic.
> It is in running state when any of its child containers running.

I have replied to this comment. I have set the `running` status to `false` as per the API definition

artembilan: I see that you refer to `Lifecycle.isRunning()`.
I'm not sure what that `container` means there, but that does not compile in my head.
How we suppose to treat container as not running if any of its child still does its work?
That really means we have a leak.

I still believe that we have to return `true` if any of the children are running.
Or leave it as is, because we just don't change concurrent `running` state even if we stop all its children manually.

LokeshAlamuri: > I still believe that we have to return `true` if any of the children are running.

I am good with this. 
But one query, what is your suggestion to know from the API, if any of the container is stopped. Do we need to get all the containers and verify if any of it is stopped.

LokeshAlamuri: > Or leave it as is, because we just don't change concurrent `running` state even if we stop all its children manually.

I think here we must set the `running` status to `false` if all the containers are stopped. Let us not leave as it is.

artembilan: Well, that's `start()` of the concurrent container.
I think the best solution would be to stop all the currently running and go ahead with the rest of the logic where we really re-create child container. 

I might agree that we can move concurrent container to not running state if all of its children are stopped.
If we have an API to notify via `thisOrParentContainer` that child is stopped.

> Do we need to get all the containers and verify if any of it is stopped.

Why do we need to check if they are stopped?
I think iteration over `this.containers` and checking `isRunning()` is enough.
Is `isChildRunning()` OK?

LokeshAlamuri: > Well, that's `start()` of the concurrent container.
> I think the best solution would be to stop all the currently running and go ahead with the rest of the logic where we really re-create child container.

As per the current code,

Start API would start new child containers.
Stop API would stop any containers running.

I feel it is correct. Let us not change any thing regarding this. I have added new conditions verify if really ConcurrentContainer is really stopped or not.

As per the earlier code, second `run` is ignored. APP dev can directly `stop` and `run` the ConcurrentContainer if he plans to run again. But, good practise to wait untill `isChildRunningAPI` returns `false`. This indicates all the resources are released.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
LokeshAlamuri: > > stop container C0 manually.
> 
> > Cmain -- not running
> > C0 -- not running
> > C1 -- running
> 
> But this situation is not correct. The `Cmain` has to be running until any of its children is running.

It could be defined like this.

1) set `running` true if all the contianers are running. Indicates `green` and All is Well.
2) set `running` false if any or all the containers are not running. Indicates `red` and all is not well. In this case, App dev can iterate over the containers to verify what are all stopped.
3) `isChildRunningAPI` returns `false` only when all the message processing is actually stopped. All the resources are released properly.

From the application operations perspective, the above definition looks good.

But, If ConcurrentContainer `running` status has to be set to `false` only when all the containers are stopped. Then it looks as below.

1) set `running` true if all the contianers are running. `running` status true,  indicates `green` and All is Well.
2) 'running` state remains `true` even if one or more containers stopped. `running` status true, indicates `green` and All is Well. But, not correct!!

App dev always has to iterate over the containers to track the app condition. `running` status is not useful.

artembilan: The problem with `false` just for one child container that the rest are still running and we don't know about that from parent container since, as you said, it has to return `false`.
If that is what you want to implement here, then it is not OK to allow to stop any child container individually.
If we allow (and I don't see why not), then `false` for parent container would lead to the resource leak in the next `start()` call. 
Pay attention that we don't stop any running containers over there, but just create new instances.
Plus pay attention that `this.startedContainers.set(0);` is wrong here, because we don't take into account those running containers at the moment.
And all of that just because we decided to return `false` for the situation when at least one child container is stopped.

LokeshAlamuri: > The problem with `false` just for one child container that the rest are still running and we don't know about that from parent container since, as you said, it has to return `false`.

If it returns `false`, indicates one, more or all containers are stopped. We can know, what containers are running by iterating over the containers.

LokeshAlamuri: > If that is what you want to implement here, then it is not OK to allow to stop any child container individually.
> If we allow (and I don't see why not), then `false` for parent container would lead to the resource leak in the next `start()` call.

As per the existing logic running containers are stopped, if `stop` is called on `ConcurrentContainer`. There is no resource leak here. Please suggest here.

LokeshAlamuri: > Plus pay attention that `this.startedContainers.set(0);` is wrong here, because we don't take into account those running containers at the moment.

this.startedContainers.set(0) -- Indicates initially no child container has started. Once started, it will be incremented and gets decremented once stopped.

LokeshAlamuri: > The problem with `false` just for one child container that the rest are still running and we don't know about that from parent container since, as you said, it has to return `false`.

As mentioned earlier, I am good to implement as per your comments.

set `running` status to `false` only if all the containers. otherwise, set to `true`

 I will verify one more time and confirm this. I thought  of putting my views. Looking for your final review of my earlier comments before I implement as you mentioned.

artembilan:
@LokeshAlamuri ,

let's step back and try to understand what is the problem at all.

Would you mind explaining the reasoning behind your work?
I thought you have implemented `fenced` before for the situation when child container can be started when its parent is stopped.
So, that mean that we are able to stop any particular child container and I don't see a reason why don't allow to that.
But I feel like this is fully different story and that is not what you are trying to achieve here.
However I still cannot fully understand what you are doing and why.

Thanks

LokeshAlamuri:
> I thought you have implemented `fenced` before for the situation when child container can be started when its parent is stopped.
> So, that mean that we are able to stop any particular child container and I don't see a reason why don't allow to that.

Fenced child container issue is very clear and straightforward. Assume `ConcurrentContainer` is stopped. But, if we are having reference to child container,  it is still possible to start the child container. It is not correct. Because, it is having reference to the `ConcurrentContainer`. It is nothing but corrupting the `ConcurrentContainer`.  I have provided Junit clearly how this scenario could be replicated.

LokeshAlamuri:
> But I feel like this is fully different story and that is not what you are trying to achieve here.
> However I still cannot fully understand what you are doing and why.

I am here trying to fix the `ConcurrentContainer` lifecycle issues and especially `isChildRunning` API. As discussed earlier, `isChildRunning` API should return `false` only when all the containers are stopped. This is the most critical thing for any application. It indicates, when exactly the spring-kafka component is completely stopped.

artembilan:
>  It is nothing but corrupting the ConcurrentContainer.

How does it corrupt it with its start if we don't stop ConcurrentContainer?
You can chose some child container to be stopped for any reason and then start it back when ever you need.

> It indicates, when exactly the spring-kafka component is completely stopped.

Right. And that's what I mean with having ConcurrentContainer running until at least one of its child container is running.
Or I even feel more natural to have ConcurrentContainer always running until its `stop()` explicitly called.
This way we still maintain the same set of `this.containers` and its up to target application to decide what to do with every child container.
Why do you find this logic as wrong?

LokeshAlamuri:
> > It indicates, when exactly the spring-kafka component is completely stopped.
> 
> Right. And that's what I mean with having ConcurrentContainer running until at least one of its child container is running. Or I even feel more natural to have ConcurrentContainer always running until its `stop()` explicitly called. This way we still maintain the same set of `this.containers` and its up to target application to decide what to do with every child container. Why do you find this logic as wrong?

Looks good to me. Only one issue, if we keep the `running`  status of the `ConcurrentContainer`  to `true`, even after one or all the containers are stopped, from the API perspective imagine once, who wants to track the system, they cannot use the `running` status any more and have to always iterate over the containers to see every thing is functioning properly or not. 

But, If we have set the `running`  status of the `ConcurrentContainer`  to `false`,  from the API perspective it tells ok something is wrong and then they can have the logic to handle the situation. `running` status to `true` indicates all is good. In this model, `running` flag is  really useful.

LokeshAlamuri:
> > It is nothing but corrupting the ConcurrentContainer.
> 
> How does it corrupt it with its start if we don't stop ConcurrentContainer? You can chose some child container to be stopped for any reason and then start it back when ever you need.

As mentioned in my previous comment, I am not trying to stop `ConcurrentContainer`.  I am only trying to stop the fenced containers from starting once again since they are holding the reference to the `ConcurrentContainer`. That is the fix we gave  for the issue https://github.com/spring-projects/spring-kafka/issues/3371.

artembilan:
>  set the running status of the ConcurrentContainer to false, if any of the container fails

Right, but then it is chicken-egg problem.
I see the situation when other child containers are still running, however the whole ConcurrentContainer says that nothing is running, so it is OK to call its run and we are in a leakage situation where ConcurrentContainer spawns new containers without stopping others.
Even if we fix this via refresh the whole state in the start, I still feel like this is an overhead and pointless number of operations.
With `running` controlled just only by the ConcurrentContainer, then it is OK to deal with every its children individually.
And we really refresh the state when we restart the whole ConcurrentContainer explicitly.
The combination of `isRunning()` and `isChildRunning()` should cover all the possible situations and simplify target logic.

LokeshAlamuri:
> > set the running status of the ConcurrentContainer to false, if any of the container fails
> 
> Right, but then it is chicken-egg problem. I see the situation when other child containers are still running, however the whole ConcurrentContainer says that nothing is running, so it is OK to call its run and we are in a leakage situation where ConcurrentContainer spawns new containers without stopping others. Even if we fix this via refresh the whole state in the start, I still feel like this is an overhead and pointless number of operations. With `running` controlled just only by the ConcurrentContainer, then it is OK to deal with every its children individually. And we really refresh the state when we restart the whole ConcurrentContainer explicitly. The combination of `isRunning()` and `isChildRunning()` should cover all the possible situations and simplify target logic.

Ok then i will keep the `running` status logic as before and will revert the necessary changes in my PR. Once done, we can have  further discussion. Thank you.

LokeshAlamuri:
Gentle reminder. I have updated the PR as per our discussion.

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -204,14 +206,20 @@ public boolean isContainerPaused() {
...
-		for (MessageListenerContainer container : this.containers) {
-			if (container.isRunning()) {
+		this.lifecycleLock.lock();
+		try {
+			for (MessageListenerContainer container : this.containers) {
+				if (container.isRunning()) {
+					return true;
+				}
+			}
+			if (this.startedContainers.get() > 0) {
```
artembilan, line 216: How we can reach here if we have at least one running child container?
Otherwise, if we don't, how this property is positive?
Isn't there something off?
Why do we need `this.startedContainers` at all if we fully can control that via ` this.containers`?

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
```diff
@@ -380,30 +389,47 @@ protected void doStop(final Runnable callback, boolean normal) {
 
 	@Override
 	public void childStarted(MessageListenerContainer child) {
-		this.startedContainers.incrementAndGet();
+		this.lifecycleLock.lock();
+		try {
+			if (this.containers.contains(child) || this.stoppedContainers.contains(child)) {
+				this.startedContainers.incrementAndGet();
```
artembilan, line 395: This logic is somehow does not compile in my head.
Why do we need to track those `stoppedContainers` at all?
Does not look like that brings any benefit over `setFenced()` introduced before.

In my feeling the `this.containers` collection is totally enough to track all the lifecycle of children if their `start/stop` is done manually. 
Well, I still see a performance benefit via `this.startedContainers` counter, but this `stoppedContainers` smells like an overhead.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
LokeshAlamuri: This is the critical condition and is reachable if the `isChildRunning` is called if the child containers are still processing the messages. `this.containers` are cleared after the `stop` is called. We cannot rely on it completely. `this.startedContainers` could be used to track the actual child count and would be reached to `zero` after all the containers are stopped.

Please find the junit I have submitted with this PR to test this condition. 

```
public void testIsChildRunning() throws Exception {

.....
// This returns true since one container is still processing message. Key validation for this test case.
		assertThat(container.isChildRunning()).isTrue();
```

spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
LokeshAlamuri: When the container stop is called, all the `this.containers` are cleared. Assume not all containers really called childstopped.  Now, earlier child containers would invoke the childstopped API. How can we know that these child containers really belong to ConcurrentContainer for the active run or just stopped run. So, even though `stop` is called on ConcurrentContainer, we need to maintain the previous containers untill a second run starts or all the child containers are really stopped.

`this.stoppedContainers` would be useful here.

artembilan: Right. So, we must not do that `this.containers.clear();` in the end of `ConcurrentMessageListenerContainer.doStop(final Runnable callback, boolean normal)`.
And reset it when we do a new `ConcurrentMessageListenerContainer.start()`.
This way we would always have a track of children until we go fresh start 😄

artembilan (changes requested):

LokeshAlamuri:
I have updated the PR as per our discussion. Please review and give your comments.

artembilan:
Thanks for the great contribution @LokeshAlamuri; looking forward for more!

We will try to back-port the fix to those supported versions...

artembilan:
OK. We cannot back-port it to previous versions.
Apparently there is still something missing there.
So, leaving it only in the `main`.

Feel free to contribute PRs for those supported versions.
However they will make only for the next release: we do current today.

Hope the problem you are trying to fix is not that critical if we deal with container lifecycle properly.

PR #3411, opened 2024-08-09
Clean up `processSeeks()` duplicate code

bky373:
In `processTimestampSeeks()`, we use `seekIterator` to pre-remove unassigned partitions, thus eliminating duplicate code running in the following loop.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -3011,8 +3011,14 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			while (offset != null) {
 				traceSeek(offset);
 				try {
-					SeekPosition position = offset.getPosition();
 					TopicPartition topicPartition = offset.getTopicPartition();
+					if (assigned == null || !assigned.contains(topicPartition)) {
+						this.logger.warn("No current assignment for partition " + topicPartition +
+								" due to partition reassignment prior to seeking.");
+						offset = this.seeks.poll();
+						continue;
+					}
+					SeekPosition position = offset.getPosition();
 					Long whereTo = offset.getOffset();
 					Function<Long, Long> offsetComputeFunction = offset.getOffsetComputeFunction();
 					if (position == null) {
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3011,14 +3011,8 @@ private void processSeeks() {
 			while (offset != null) {
 				traceSeek(offset);
 				try {
```
sobychacko, line 3013: Are these based on a different PR? The latest `main` branch line numbers don't match and the code looks different.

bky373: No, I worked on the remote main branch.

bky373: But I worked on the same branch as the work [here](https://github.com/spring-projects/spring-kafka/pull/3410), which I've force-pushed now.

sobychacko: Ok understood. My bad. I was looking at a different branch. Sorry.

bky373:
Hmm, according to L3028 in the same class, late addition can occur during a while loop, so it might be a good idea to close this PR.

sobychacko:
Yes, feel free to close this PR, marking it as invalid if you see a path for any issue. It might still be a good idea to refactor that code into a utility method since the code is almost identical in the conditional check - `if (assigned == null || !assigned.contains(topicPartition)) {`

bky373:
> Yes, feel free to close this PR, marking it as invalid if you see a path for any issue. It might still be a good idea to refactor that code into a utility method since the code is almost identical in the conditional check - `if (assigned == null || !assigned.contains(topicPartition)) {`

Your suggestion sounds great. I'll update the PR with a commit that uses a common method.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3012,9 +3016,9 @@ private void processSeeks() {
...
 					TopicPartition topicPartition = offset.getTopicPartition();
-					if (assigned == null || !assigned.contains(topicPartition)) {
-						this.logger.warn("No current assignment for partition " + topicPartition +
-								" due to partition reassignment prior to seeking.");
+					if (!isPartitionAssigned(assigned, topicPartition)) {
+						this.logger.warn("No current assignment for partition '" + topicPartition +
```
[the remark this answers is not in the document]
bky373: Okay moved. Thanks!

Issue #3409, opened 2024-08-07
Error handler throws an exception on blocking retries


lm231290:
version 3.1.2 - 3.2.2

When using blocking retries an back-off (e.g `new FixedBackOff(5000, 2)`), there is an error message in the log:
```
...KafkaMessageListenerContainer    : Error handler threw an exception
org.springframework.kafka.KafkaException: Seek to current after exception
...
```
I see two confusing things here:
1. the log states that it is an exception of Error Handler, and occurred while seeking. Even though it has a real exception in `Caused by`, the beginning is wrong in my opinion because the Error Handler didn't fail handling my errors
2. the log is printed **after** the back-off interval (5s in my example), right before the next attempt

To reproduce:
- Create DefaultErrorHandler as a Spring bean
- Set a back-off with a notable interval
- Set this error handler in KafkaListenerContainerFactory used for the listener

**Expected behavior**

Either no error message at all, or log it **right after** the attempt fails, **before** the back-off pause, **and** with less confusing message

**Sample**

https://github.com/lm231290/kafka-tests

There is a `retry.log` file there with my execution example
Or you can run ListenerTest.test_single and see the log messages and the timestamps

artembilan:
I didn’t look into code, but if you find it confusing then something is really wrong. We will be happy to review any contribution. Otherwise we will look into this when we can.
Thanks.

sobychacko:
I briefly looked at the issue by running your sample app. It looks like it is doing the right thing, but I agree that something is amiss, as the messages are confusing. The exception is thrown specifically from here: https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2719
The exception from the listener is wrapped inside a `KafkaException`, and the default level there is `ERROR.` We will take a closer look.

sobychacko:
The normal expectation is that when you have enabled retry, you see the exceptions only after all the retries are exhausted and the record fails. So, we will have to debug more and see whats going on.

lm231290:
@sobychacko Yeah, that is a good point too, to have the error logged only after the attempts are exhausted. Before that I'd prefer to see warnings or infos

sobychacko:
As a workaround, you can set the log level on the error handler to `TRACE.` That should remove the superfluous logging. 
```
defaultErrorHandler.setLogLevel(KafkaException.Level.TRACE);
```
[...]

PR #3421, opened 2024-08-14
GH-3409: Improve exceptions during retry

sobychacko:
Fixes: #3409

- During error handling, records in retry are throwing `KafkaException` after seeking that causes the exception to get logged which maybe misleading. It seems to indicate that the error handling process itself thew the exception, while the exception was thrown in order to prevent accidental committing of a not-yet recovered record. Change this exception during seeks in retry while handling error from `KafkaException` to a new framework-only used exception - `RecordInRetryException` that simply logs the message at INFO level.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -2213,7 +2213,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 					commitOffsetsIfNeededAfterHandlingError(records);
 				}
 				catch (RecordInRetryException rire) {
-					rire.selfLog(this.logger);
+					this.logger.info("Record in retry and not yet recovered");
 					return rire;
 				}
 				catch (KafkaException ke) {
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2212,6 +2212,10 @@ private RuntimeException doInvokeBatchListener(final ConsumerRecords<K, V> recor
 					invokeBatchErrorHandler(records, recordList, e);
 					commitOffsetsIfNeededAfterHandlingError(records);
 				}
+				catch (RecordInRetryException rire) {
+					rire.selfLog(this.logger);
```
artembilan, line 2216: I wouldn't do this, but just `this.logger.info("BLAH-BLAH", rire)`

spring-kafka/src/main/java/org/springframework/kafka/listener/RecordInRetryException.java
```diff
@@ -0,0 +1,52 @@
...
+	 * @param cause {@link Throwable}
+	 */
+	RecordInRetryException(String message, @Nullable Throwable cause) {
+		super(message, cause);
+		this.message = message;
+	}
+
+	void selfLog(LogAccessor logger) {
+		logger.info(this.message);
+	}
```
artembilan, line 51: Just nit-pic: every member in the class must be surrounded with blank lines, even the last method of the last inner class.

artembilan (changes requested):
And yes: I don't think there is anything to mention in the docs 😄

Djaler:
It was unexpected for me to not see listener error logs after upgrading spring in our app.

We have error handler without max retries, so now we don't see errors logs.

PR #3422, opened 2024-08-15
No Issue : Internal Refactoring to improve code readability.

chickenchickenlove:
### Motivation:
- There are some codes using `*.size() > 0` or `*.size() == 0` in `spring-kafka`.
- There is code where the constructor could be reused, but it was not. 
- There is code where a `null` value can occur, but it does not indicate that `null` is possible.

### Modifications:
- Replace `a.size() > 0` with `!a.isEmpty()`.
- Replace `a.size() == 0` with `a.isEmpty()`.
- Reuse constructor already existed to remove duplicate constructor logic. 
- Add @Nullable annotations to field, and method. Also, add warning log. 

### Result:
- There is only one side effect which developer can observe.
  -  when `BatchMessagingMessageListenerAdapter#setBatchMessageConverter()` is executed, new warning log can be occur.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
@@ -60,6 +60,7 @@ import org.springframework.messaging.support.MessageBuilder;
  * @author Gary Russell
  * @author Dariusz Szablinski
  * @author Biju Kunjummen
+ * @author Sanghyeok An
  * @since 1.1
  */
 public class BatchMessagingMessageConverter implements BatchMessageConverter {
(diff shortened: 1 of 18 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
```diff
@@ -60,12 +60,14 @@
...
  * @author Dariusz Szablinski
  * @author Biju Kunjummen
+ * @author Sanghyeok An
  * @since 1.1
  */
 public class BatchMessagingMessageConverter implements BatchMessageConverter {
 
 	protected final LogAccessor logger = new LogAccessor(LogFactory.getLog(getClass())); // NOSONAR
 
+	@Nullable
```
chickenchickenlove, line 70: `RecordMessageConverter` can be null.  Please refer to codes below.

```java
// BatchMessagingMessageConverter.java
	public BatchMessagingMessageConverter() {
		this(null);
	}

	public BatchMessagingMessageConverter(RecordMessageConverter recordConverter) {
		this.recordConverter = recordConverter;
		if (JacksonPresent.isJackson2Present()) {
			this.headerMapper = new DefaultKafkaHeaderMapper();
		}
	}
```

Then, `getRecordMessageConverter()` can return `null`.
`BatchMessagingMessageListenerAdapter#setBatchMessageConverter()` call `getRecordMessageConverter()` internally to set messageConverter. 

If user created `BatchMessagingMessageConverter` by using default constructor, there is no message converter even if `BatchMessagingMessageListenerAdapter#setBatchMessageConverter()` is completed without any warning logs.

Issue #3466, opened 2024-08-29
Wrong comparison of server bootstrap addresses in KafkaTemplate.afterSingletonsInstantiated


bleipold1:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.2.3

**Describe the bug**

In line 490 of KafkaTemplate are the bootstrap server addresses of the producer factory and the bootstrap server addresses of the kafka admin bean compared  with each other. In most cases this will fail, because producerServers is a list of strings and adminServers always string. Even if the string represents the same list.

**To Reproduce**

**Expected behavior**

If producerServers and adminServers represent the same bootstrap servers, the if branch should not be entered.

sobychacko:
@bleipold1 Thanks for the report. We will look into this soon.

sobychacko:
@bleipold1 While it looks like an issue, we are curious how you encountered this issue. If they don't match (even by type), then we get into the conditional for creating a new `KafkaAdmin`. 

Also, this only happens if you have a Boot application because you always get the bootstrap server config from Boot as a `List`.

PR #3471, opened 2024-08-29
GH-3466: Optimize KafkaAdmin creation in KafkaTemplate

sobychacko:
Fixes: #3466

https://github.com/spring-projects/spring-kafka/issues/3466

Improve bootstrap-server config comparison to avoid unnecessary KafkaAdmin recreation. This addresses inconsistencies between List<String> and String configurations for bootstrap servers.

The change ensures that List versions of bootstrap-server configs are converted to regular Strings by removing brackets. This allows for consistent comparison between producer and admin configurations.

This optimization is particularly relevant for Spring Boot scenarios where configs may be provided in different formats but represent the same underlying values.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -80,6 +80,7 @@ import org.springframework.messaging.converter.SmartMessageConverter;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
+import org.springframework.util.StringUtils;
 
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -506,19 +508,22 @@ else if (this.micrometerEnabled) {
 		}
 	}
 
+	private String removeLeadingAndTrailingBrackets(String str) {
```
artembilan, line 511: This should be `static` and somewhere in the end of class.
Probably also no need to make its code multi-line: looks like that last `']'` would fit into line length.

artembilan (changes requested):

PR #3477, opened 2024-08-31
Refactor code

LokeshAlamuri:
Following code is refactored.

1) `generateFromSleepingBackOffPolicy` method in `org.springframework.kafka.retrytopic.BackOffValuesGenerator`

Removed the code to set the `maxBackOffPeriod`, if the BackOffPolicy is `UniformRandomBackOffPolicy`. This is not required since, when sleeper is added to the `UniformRandomBackOffPolicy` `maxBackOffPeriod` is not modified. This could be an issue with the earlier versions of spring-retry. But, latest version of spring-kafka 3.3 and spring-retry 2.0.8, this is not an issue.

2) Modified `addIfAbsent` in `org.springframework.kafka.retrytopic.ListenerContainerFactoryResolver.Cache` to use `putIfAbsent`, rather than explictly checking for the key, if it is not there than adding it to the `HashMap`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/BackOffValuesGenerator.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/BackOffValuesGenerator.java
@@ -28,7 +28,6 @@ import org.springframework.retry.backoff.FixedBackOffPolicy;
 import org.springframework.retry.backoff.NoBackOffPolicy;
 import org.springframework.retry.backoff.Sleeper;
 import org.springframework.retry.backoff.SleepingBackOffPolicy;
-import org.springframework.retry.backoff.UniformRandomBackOffPolicy;
 import org.springframework.retry.support.RetrySynchronizationManager;
 
 /**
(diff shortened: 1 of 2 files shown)


sobychacko:
@LokeshAlamuri Changes look good. Did you test to ensure that the code you removed in `BackOffValuesGenerator` has no unintended side effects? How did you verify? Also, please add your name to the modified classes, improve the commit message title, etc. Thanks!

artembilan (approved):
We can merge this fix.
The code in Spring Retry is like this:
```
	public UniformRandomBackOffPolicy withSleeper(Sleeper sleeper) {
		UniformRandomBackOffPolicy res = new UniformRandomBackOffPolicy();
		res.minBackOffPeriodSupplier(minBackOffPeriod);
		res.maxBackOffPeriodSupplier(maxBackOffPeriod);
		res.setSleeper(sleeper);
		return res;
	}
```
So, both properties are copied to a new returned instance.

@LokeshAlamuri ,

thank you for contribution!

PR #3484, opened 2024-09-13
GH-3483: Allow @Override of KafkaAdmin createAdmin()

### LINKED ISSUE #3483: Allow @Override of KafkaAdmin createAdmin()
**Expected Behavior**

The Spring KafkaAdmin class has a package-private method "createAdmin" that returns the `org.apache.kafka.clients.admin.AdminClient` type:

```java
AdminClient createAdmin() {
    return AdminClient.create(this.getAdminConfig());
}
```

It'd be great if this method was protected, and returned the `org.apache.kafka.clients.admin.Admin` type. Additionally, the code in KafkaAdmin would need to be updated to use the `Admin` interface, instead of the `KafkaAdmin` class.

It's currently possible to do something similar the default producer/consumer factories, as they have methods like this:

```java
	protected Consumer<K, V> createKafkaConsumer(Map<String, Object> configProps) {
		checkBootstrap(configProps);
		Consumer<K, V> kafkaConsumer = createRawConsumer(configProps);
		if (!this.listeners.isEmpty() && !(kafkaConsumer instanceof ExtendedKafkaConsumer)) {
			LOGGER.warn("The 'ConsumerFactory.Listener' configuration is ignored " +
					"because the consumer is not an instance of 'ExtendedKafkaConsumer'." +
					"Consider extending 'ExtendedKafkaConsumer' or implement your own 'ConsumerFactory'.");
		}

		for (ConsumerPostProcessor<K, V> pp : this.postProcessors) {
			kafkaConsumer = pp.apply(kafkaConsumer);
		}
		return kafkaConsumer;
	}
```

**Current Behavior**

KafkaAdmin cannot be extended/overridden to use another type of class implementing the  `org.apache.kafka.clients.admin.Admin` interface.

**Context**

I'm using [OKafka](https://github.com/oracle/okafka), which has implementations of the Kafka interfaces, e.g., the Admin interface, and I would like to use these implementations with Spring Kafka. Based on my understanding of the code, some modifications would need to be made as described - I'd also be interested in any workarounds, if they are available.

anders-swanson:
Fixes #3483 

Changes the KafkaAdmin#createAdmin() method to be protected, returning the Admin interface instead of the AdminClient class. 

This change allows user of spring-kafka to supply their own Admin interface implementations by extending the KafkaAdmin class.


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -389,7 +390,7 @@ public Map<String, TopicDescription> describeTopics(String... topicNames) {
 		}
 	}
 
-	AdminClient createAdmin() {
+	protected Admin createAdmin() {
```
sobychacko, line 393: This will now be part of the API - we need to add a `@since` tag.

sobychacko (changes requested):
1. Please add author tag to the classes you modified. 
2. I think since we are exposing the `createAdmin` method as part of the `KafkaAdmin` API, we need to add some ref docs for it.
3. Also, add a section in `whats-new`.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
anders-swanson: Added a JavaDoc, thanks.

anders-swanson:
@sobychacko thanks for the review, I've updated the PR with the requested documentation 👍

Issue #3276, opened 2024-05-23
@RetryableTopic not working for asynchronous @KafkaListener return types


matthiasValuecloud:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.2.0

**Describe the bug**

When using the `@RetryableTopic` annotation on a Listener that returns a `CompletableFuture`, no retry or dlt events are sent when the `CompletableFuture` completes with an exception.

**To Reproduce**

```
@RetryableTopic(
            attempts = "5",
            kafkaTemplate = "retryTemplate",
            autoCreateTopics = "false",
            backoff = @Backoff(delay = 6000)
    )
@KafkaListener(
        topics = "my-test-topic",
        properties = {
                "key.deserializer=io.confluent.kafka.serializers.KafkaAvroDeserializer",
                "value.deserializer=io.confluent.kafka.serializers.KafkaAvroDeserializer",
                "specific.avro.reader=true"
        },
        id = "myId"
)
public CompletableFuture<Void> receive(ConsumerRecord<Key, Value> record) {
    log.info("Received request (id {}) on topic {}", record.key().getId(), record.topic());

    CompletableFuture<Void> result = processor.process(record.key(), record.value());
    result.whenComplete((v, ex) -> {
       if (ex != null) {
           log.error("Error processing request: {}", ex.getMessage());
       }
    });
    return result;
}
```

**Expected behavior**

A `CompletableFuture`, which results in an exception, should be retried with the given properties and marked for ack on my-topic.

matthiasValuecloud:
Well i can try, but i am really not comfortable around such a huge project and how to contribute properly. I will read through the guidelines and see what i can do :D

sobychacko:
@matthiasValuecloud Sounds good. Please let us know if you end up working on it. Here are the contribution guidelines: https://github.com/spring-projects/spring-kafka/blob/main/CONTRIBUTING.adoc.

matthiasValuecloud:
Haven't had the chance on working on it so far. I apologize.

artembilan:
Moving to Backlog since we haven't have a chance to look into this.

chickenchickenlove:
@matthiasValuecloud Hi!
If you are okay, may i take a look this issue?

artembilan:
Be our guest, @chickenchickenlove !
We have some other priorities to dig out, but will more than happy to review a possible fix or discuss some specific details .
Thank you for looking into this!

[...]

PR #3523, opened 2024-09-30
GH-3276: Support async retry with @RetryableTopic

chickenchickenlove:
### Motivation:
From now on, when the listener end point method returns `CompletableFuture<?>` and `Mono<?>` successfully and both `CompletableFuture<?>` and `Mono<?>` fail to be completed due to unexpected error, these failure record never retried even if a developer set `@Retryable` topic. 

If we supports async retry with @RetryableTopic, we expect that spring-kafka's will be improved.

### Changes
- Add new method for sending `ConsumerRecord` when both `CompletableFuture<?>` and `Mono<?>` was failed. 
- Add test cases and scenario test cases. 

### Idea
The Kafka's `Consumer` has soft lock, so multiple threads can not access the consumer simultaneously.
It means that the thread in `ThreadExecutor` which handle `CompletableFuture` or `Mono` cannot access the consumer. 

my idea is very simple. 
the `KafkaMessageListenerContainer` has `ConcurrentLinkedDeque` as its field. 
Then, when `MessagingMessageListenerContainer` add `asyncSuccess()` and `asyncFailure()` to `CompletableFuture<?>` and `Mono<?>` as hooks. 

If these future object failed to be completed, future object will call `MessagingMessageListenerContainer.asyncFailure()`. 
When `MessagingMessageListenerContainer.asyncFailure()` is called, `FailedRecordTuple` will be put to `ConcurrentLinkedDeque`. 

Then, next loop, `KafkaMessageListenerContainer` will call `handleAsyncFailure()`. 
then, `KafkaMessageListenerContainers` follows `FailedRecordTracker`'s logic. 

### Result
- Fixes : https://github.com/spring-projects/spring-kafka/issues/3276

### Background
I investigate how to support `@RetryableTopic` for async.  
I think it is difficult task for me to solve it. 😓
Anyway, challenges are a good thing, so I dug a little deeper. 😅 
First of all, i want to discuss  with the reviewers about which direction i should take for the implementation.

I implemented an ErrorHandler that sends messages to `Retry` and `DLT` topics when an error occurs, passing it as a callback to `MessagingMessageListenerAdapter` and inserting it into `MessagingMessageListeneerAdapter#asyncFailure()` to send messages to Retry and DLT topics. Although this method has some issues, it can achieve the intended purpose.

You can check successful test cases below.
[shouldRetryFirstFutureTopic](https://github.com/chickenchickenlove/spring-kafka/blob/fdc59f065f19d9db6ede26de62a3453a59ce4f24/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncRetryTopicClassLevelIntegrationTests.java#L132-L155)
[shouldRetryFirstMonoTopic](https://github.com/chickenchickenlove/spring-kafka/blob/fdc59f065f19d9db6ede26de62a3453a59ce4f24/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncRetryTopicClassLevelIntegrationTests.java#L158-L181)
[shouldRetrySecondFutureTopic](https://github.com/chickenchickenlove/spring-kafka/blob/fdc59f065f19d9db6ede26de62a3453a59ce4f24/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncRetryTopicClassLevelIntegrationTests.java#L184-L187)
[shouldRetrySecondMonoTopic](https://github.com/chickenchickenlove/spring-kafka/blob/fdc59f065f19d9db6ede26de62a3453a59ce4f24/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncRetryTopicClassLevelIntegrationTests.java#L191-L195)
[shouldRetryThirdFutureTopicWithTimeout](https://github.com/chickenchickenlove/spring-kafka/blob/fdc59f065f19d9db6ede26de62a3453a59ce4f24/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncRetryTopicClassLevelIntegrationTests.java#L199-L237)
[shouldRetryThirdMonoTopicWithTimeout](https://github.com/chickenchickenlove/spring-kafka/blob/fdc59f065f19d9db6ede26de62a3453a59ce4f24/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncRetryTopicClassLevelIntegrationTests.java#L240-L279)
[shouldRetryFourthFutureTopicWithNoDlt](https://github.com/chickenchickenlove/spring-kafka/blob/fdc59f065f19d9db6ede26de62a3453a59ce4f24/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncRetryTopicClassLevelIntegrationTests.java#L282-L286)
[shouldRetryFourthMonoTopicWithNoDlt](https://github.com/chickenchickenlove/spring-kafka/blob/fdc59f065f19d9db6ede26de62a3453a59ce4f24/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncRetryTopicClassLevelIntegrationTests.java#L289-L292)
[shouldGoStraightToDltInFuture](https://github.com/chickenchickenlove/spring-kafka/blob/240929-async-retry/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncRetryTopicClassLevelIntegrationTests.java#L317-L320)
[shouldGoStraightToDltInMono](https://github.com/chickenchickenlove/spring-kafka/blob/240929-async-retry/spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncRetryTopicClassLevelIntegrationTests.java#L324C7-L327)

However, i realized that it is hard for `spring-kafka` to support `back-off` to async kafka listener endpoint.
image

The `Backoff time` is determined when `KafkaMessageListenerContainer` calls `invokeErroHandler()`.
At that time, `FailedRecord` is computed by `Thread`.
```java
// FailedRecordTracker.java
...
// NOTE: currentThread.
Thread currentThread = Thread.currentThread();
Map<TopicPartition, FailedRecord> map = this.failures.computeIfAbsent(currentThread, t -> new HashMap<>());
TopicPartition topicPartition = new TopicPartition(record.topic(), record.partition());
FailedRecord failedRecord = getFailedRecordInstance(record, exception, map, topicPartition);
this.retryListeners.forEach(rl ->
		rl.failedDelivery(record, exception, failedRecord.getDeliveryAttempts().get()));
long nextBackOff = failedRecord.getBackOffExecution().nextBackOff();
...
```
However, it is hard to imagine that `CompletableFuture` and `Mono` always will be executed in same thread. 
it always will affects `nextBackOff`.
Thus, `spring-kafka` cannot support `BackOff Retry` in async mode. 

IMHO, I think that we have 3 options.

#### 1. `@RetryableTopic` does not support `backoff` features in `async` mode.
in this case, new kafka header is needed to distinguish between `sync` and `async`. 
the other one is that always use minimum backoff time \to avoid `KafkaBackoffException` 

#### 2. Introduce [micrometer/context-propagation](https://github.com/micrometer-metrics/context-propagation) to support `backoff` features in async mode.
For example, https://spring.io/blog/2023/03/28/context-propagation-with-project-reactor-1-the-basics. 
I believe that we can infer proper `nextBackOffTime` by using context-propagation to store and restore state in each `thread`.

#### 3.  Big refactoring.
But i have no idea. maybe `SyncKafkaMessageListenerContainer` and `AsyncKafkaMessageListenerContainer`? 

Thanks for reading this description. 
What do you think? Please let me know your opinion.
Thanks in advance! 🙇‍♂️


chickenchickenlove:
Sounds good, @artembilan 👍 
I would like to check one thing for confirmation to sync our opinions. (Sorry for my pool english 😓 )

# My Understanding 1
1. Call `asyncFailure()` with `Consumer`.
2. `asyncFailure()` calls `asyncRetryCallback` with `Consumer`
3. `asyncRetryCallback` calls `FailedRecordTracker.restore()` with `Consumer`. 
And then, we should investigate that `Consumer` can be replaced with `Thread.currentThread()`.

# My Understanding 2
1. Call `asyncFailure()` with `Consumer`.
2. `asyncFailure()` calls `FailedRecordTracker` directly to get `nextBackOff`
3. After then, `asyncFailure` produce failed record to retry or dlt topic. 

Which understanding is closer to what you described? 
When you have time, please let me know 🙇‍♂️

artembilan:
The `asyncFailure()` is already called with a `Consumer`.
I just followed a description you have provided so far:

> I implemented an `ErrorHandler` that sends messages to Retry and DLT topics when an error occurs, passing it as a callback to `MessagingMessageListenerAdapter`.

So, I assume that your impl deals already with a `FailedRecordTracker`.
Sorry, didn't look into the code yet.

I'm not sure what is `asyncRetryCallback`, but my understanding is to follow a logic of the `DefaultErrorHandler`:
```
		try {
			return getFailureTracker().recovered(record, thrownException, container, consumer);
		}
```
And as you see, the `Consumer` is there already.

chickenchickenlove:
Thanks for your description. 🙇‍♂️ 
I'm going to go back and dig it further. 
If it works well, I'll complete the test code and get back to you. If it doesn’t go well, I'll also investigate whether there is other options.


FYI, `asyncRetryCallback` is almost same with [KafkaMessageListenerContainer#invokeErrorHandler()](https://github.com/spring-projects/spring-kafka/blob/c44c339c6ccecc815860cc4ba7b158a2f3adbb68/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2830-L2874). 

From now on, function calls flow is : `asyncFailure()` -> `asyncRetryCallback` -> `KafkaMessageListenerContainer.invokeErrorHandler()` -> ... -> `FailedRecordTracker.recovered()`

chickenchickenlove:
@artembilan , 
There is an issue when attempting an async retry using a `Consumer`. 😢 
`LegacyKafkaConsumer` has a soft lock. because it is not safe for multi-threaded access. ([code](https://github.com/apache/kafka/blob/8fc6d8f30b19983db5e98b9e914f971f8ba91b39/clients/src/main/java/org/apache/kafka/clients/consumer/internals/LegacyKafkaConsumer.java#L1215-L1230)). 

For example, we have
- `Consumer A`: subscribe Topic A and poll messages. 
- `Main Thread`: is in charge of `Consumer A` main loop. (`fetch`, `poll`, and Etc.)
- `SubThread in ThreadPool Executor` : is in charge of executing `CompleatbleFuture` or `Mono` instance and async retry. 

In that case, `ConcurrentModificationException` can occurs in [here](https://github.com/apache/kafka/blob/8fc6d8f30b19983db5e98b9e914f971f8ba91b39/clients/src/main/java/org/apache/kafka/clients/consumer/internals/LegacyKafkaConsumer.java#L1215-L1230). 

image

</br>

Initially, only the `Main Thread` will own `Consumer A`, as there is no `async failure`, so there is no need to call `partitionsFor()`.

However, once async failures occurs, a `ConcurrentModificationException()` may be thrown due to a race condition between the thread(`Sub Thread`) calling `partitionsFor()` and the thread(`Main Thread`) calling `consumer.poll()` or `consumer.pause()`.

Thus, it can not be proper workaround. 

How about adding a `ConcurrentLinkedQueue` as a field in `KafkaMessageListenerContaine`r and using it for `SubThread ` to send a failure record to the `Main Thread`? Then, `Main Thread` will convert it to `FailedRecord` and send it to Retry Topic or DLT Topic. 

I imagined adding a method like `performAsyncFailureSendIfNeeded()` either before or after [pollAndInvoke()](https://github.com/spring-projects/spring-kafka/blob/c44c339c6ccecc815860cc4ba7b158a2f3adbb68/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1290-L1357), and calling that method to handle the async failures in the `ConcurrentLinkedQueue`.


What do you think?

chickenchickenlove:
> I imagined adding a method like performAsyncFailureSendIfNeeded() either before or after [pollAndInvoke()](https://github.com/spring-projects/spring-kafka/blob/c44c339c6ccecc815860cc4ba7b158a2f3adbb68/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1290-L1357), and calling that method to handle the async failures in the ConcurrentLinkedQueue.

If so, it seems to work well.
I implemented the code and test cases. 
All test cases for async retry (`CompletableFuture`, `Mono`) passed!

If it's alright with you, I would like to proceed in that direction and submit the code for your review.
Please let me know 🙇‍♂️ 

If so, it seems to work well. 
I implemented codes and test codes. 
all test cases for async retry (`CompletableFuture`, `Mono`) passed!

chickenchickenlove:
Hi, @artembilan ! 
I added new commit for code and scenario test cases! 
Also, i updated description of this PR. 

When you have time, could you take a look? 🙇‍♂️ 
Thanks in advance!!

spring-kafka/src/main/java/org/springframework/kafka/listener/MessageListener.java
```diff
@@ -26,8 +30,12 @@
  *
  * @author Marius Bogoevici
  * @author Gary Russell
+ * @author Sanghyeok An
  */
 @FunctionalInterface
 public interface MessageListener<K, V> extends GenericMessageListener<ConsumerRecord<K, V>> {
 
+	default void setCallbackForAsyncFailureQueue(Consumer<FailedRecordTuple<K, V>> asyncRetryCallback) {
```
artembilan, line 38: Can we hide this behind `java.util.function.BiConsumer` abstraction?

I may see a value in the `FailedRecordTuple`, but let's try to push it down to the `KafkaMessageListenerContainer` internals!

artembilan (changes requested):
I see what is going on now.
Thank you for digging further!
Now I know what is the power of single-threaded processors and how to handle its behavior.

So, I have only one simple concern about public API and after that it would be great to mention this feature in the docs.

Re. error handling of the error handler.
See `invokeErrorHandler()`:
```
				try {
					handled = this.commonErrorHandler.handleOne(rte, cRecord, this.consumer,
							KafkaMessageListenerContainer.this.thisOrParentContainer);
				}
				catch (Exception ex) {
					this.logger.error(ex, "ErrorHandler threw unexpected exception");
				}
```
So, that should be enough for now to close all the gaps we have so far.

spring-kafka/src/main/java/org/springframework/kafka/listener/MessageListener.java
chickenchickenlove: IMHO, we need to use `FailedRecordTuple`. 
Otherwise, it's hard to manage the `ConsumerRecord` and the `Exception` queues.!

As you mentioned, I changed the code to use `BiConsumer` and removed dependency on `FailedRecordTuple` , except in `KafkaMessageListenerContainer`. 
Now, `FailedRecordTuple` has become an internal record.

Does this make sense to you?
Please let me know!

chickenchickenlove:
@artembilan , sorry to bother you.

> So, I have only one simple concern about public API and after that it would be great to mention this feature in the docs.
> Re. error handling of the error handler.
> See invokeErrorHandler():
> So, that should be enough for now to close all the gaps we have so far.
> 

I can't understand your comments 😢. 
When you have time, could you give me more detail and context...? 
Thanks in advance 🙇‍♂️

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1294,6 +1307,16 @@ public void run() {
...
 			this.lastReceive = System.currentTimeMillis();
 			while (isRunning()) {
+
+				try {
+					handleAsyncFailure();
+				}
+				catch (Exception e) {
+					// TODO: Need to improve error handling.
+					// TODO: Need to determine how to handle a failed message.
+					this.logger.error("Failed to process re-try messages. ");
```
artembilan, line 1317: The logging is enough. 
Exactly same as we do for sync retry logic.
That's what I meant in the my comment from yesterday where you got doubts.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -665,16 +670,34 @@ protected void asyncFailure(Object request, @Nullable Acknowledgment acknowledgm
...
+				// For CompletableFuture Object
+				handleException(request, acknowledgment, consumer, source,
+								new ListenerExecutionFailedException(createMessagingErrorMessage(
+										"Async Fail", source.getPayload()), t.getCause()));
+			}
+			else {
+				// For Mono Object.
+				handleException(request, acknowledgment, consumer, source,
+								new ListenerExecutionFailedException(createMessagingErrorMessage(
+										"Async Fail", source.getPayload()), t));
```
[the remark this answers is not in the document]
chickenchickenlove: When `CompletableFuture<?>` fails to complete due to an unexpected errors, it throws a `CompletionException`.  By the way, when `Mono<?>` fails to complete due to an unexpected errors, it throws the root cause exception. 

It makes difference when `KafkaMessageListenerContainer` executes retry.
For example:

`KafkaMessageListenerContainer.handleAsyncFailure()`  -> ... -> `FailedRecordTracker.recovered()` -> `FailedRecordTracker.attemptRecovery()` -> `this.recoverer.accept(...)` -> `DeadLetterPublishingRecoverer.accept(...)` -> `this.destinationResolver.apply(...)` -> `DeadLetterPublishingRecoverereFactory.destinationResolver(...)` -> `this.destinationTopicResolver.resolveDestinationTopic(...)`.

In this case, `destinationTopicHolder.getSourceDestination().shouldRetryOn(attempt, maybeUnwrapException(e))` returns `false` when the exception is a `CompletionException`.
This is because `BinarayExceptionClassifier.classify(...)` and its superclass `SubclassClassifier` return false for `CompletetionException`. 


```java
	@Override
	public DestinationTopic resolveDestinationTopic(String mainListenerId, String topic, Integer attempt, Exception e,
													long originalTimestamp) {
		DestinationTopicHolder destinationTopicHolder = getDestinationHolderFor(mainListenerId, topic);
		return destinationTopicHolder.getSourceDestination().isDltTopic()
				? handleDltProcessingFailure(destinationTopicHolder, e)
				: destinationTopicHolder.getSourceDestination().shouldRetryOn(attempt, maybeUnwrapException(e)) // This is false with CompletionException.
						&& isNotFatalException(e)
						&& !isPastTimout(originalTimestamp, destinationTopicHolder)
					? resolveRetryDestination(mainListenerId, destinationTopicHolder, e)
					: getDltOrNoOpsDestination(mainListenerId, topic, e); // thus, CompletionException gets the DLT Topics as next destination.
	}
```
Therefore, when `CompletableFuture` throws a `CompletionException`, `resolveDestinationTopic()` always return DLT topics because `destinationTopicHolder.getSourceDestination().shouldRetryOn(attempt, maybeUnwrapException(e))` is false. 


In short, 
1. `CompletableFuture` and `Mono` throw different exception when they fail to complete.
2. It makes difference when async retry. (`CompletableFuture` -> Always DLT,  `Mono` -> retryable topic and DLT).
3. Also, I think `CompletionException` is not root cause and the classifier seems to not consider root cause when they classify.

Thus, we should unwrap the `CompletionException` once in `asyncFailure()`.

spring-kafka/src/main/java/org/springframework/kafka/listener/MessageListener.java
```diff
@@ -26,8 +28,13 @@
  *
  * @author Marius Bogoevici
  * @author Gary Russell
+ * @author Sanghyeok An
  */
 @FunctionalInterface
 public interface MessageListener<K, V> extends GenericMessageListener<ConsumerRecord<K, V>> {
 
+	default void setCallbackForAsyncFailureQueue(
```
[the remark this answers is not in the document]
chickenchickenlove: Fixes to `setCallbackForAsyncFail(...)`!

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -869,6 +892,10 @@ private boolean rawByParameterIsType(Type parameterType, Type type) {
 		return parameterType instanceof ParameterizedType pType && pType.getRawType().equals(type);
 	}
 
+	public void putInAsyncFailureQueue(BiConsumer<ConsumerRecord<K, V>, RuntimeException> callbackForAsyncFailureQueue) {
```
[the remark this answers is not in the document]
chickenchickenlove: @artembilan 
IMHO, we need to this method. 

The method `setCallbackForAsyncFailure` is in `RecordMessagingMessageListenerAdapter`, which is a concrete implementation of `MessageListener` interface.
However, `MessagingMessageListenerAdapter` is not a concrete class of `MessageListener`.
Therefore, `RecordMessagingMessageListenerAdapter` cannot access fields in its superclass, `MessagingMessageListenerAdapter`.

But yes, as you mentioned, the method name needs to be changed because it gives the wrong impression.

```Java
// RecordMessagingMessageListenerAdapter.java
@Override
public void setCallbackForAsyncFailure(
		BiConsumer<ConsumerRecord<K, V>, RuntimeException> asyncRetryCallback) {
	super.setCallbackForAsyncFailure(asyncRetryCallback);
}

// MessagingMessageListenerAdapter.java
protected void setCallbackForAsyncFailure(BiConsumer<ConsumerRecord<K, V>, RuntimeException> asyncRetryCallback) {
		this.callbackForAsyncFailureQueue = asyncRetryCallback;
}

```
What do you think? Please give me your opinion 🙇‍♂️

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -895,6 +899,15 @@ else if (listener instanceof MessageListener) {
 				this.wantsFullRecords = false;
 				this.pollThreadStateProcessor = setUpPollProcessor(false);
 				this.observationEnabled = this.containerProperties.isObservationEnabled();
+
+				if (!AopUtils.isAopProxy(listener)) {
+					BiConsumer<ConsumerRecord<K, V>, RuntimeException> callbackForAsyncFailureQueue =
+							(cRecord, ex) -> {
+								FailedRecordTuple<K, V> failedRecord = new FailedRecordTuple<>(cRecord, ex);
+								this.failedRecords.addLast(failedRecord);
```
artembilan, line 907: Feels like this could be simplified to a single line of code:
```
(cRecord, ex) -> this.failedRecords.addLast(new FailedRecordTuple<>(cRecord, ex))
```
Or might be even better to extract the lambda to a private method reference for easily to digest this logic.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -895,6 +899,15 @@ else if (listener instanceof MessageListener) {
...
+							};
+					this.listener.setCallbackForAsyncFailure(callbackForAsyncFailureQueue);
```
artembilan, line 909: We just worked with @sobychacko about async observation propagation.
And we have figured out that it is really just a `MessagingMessageListenerAdapter` responsibility.
The other `GenericMessageListener` impls are out of that async logic scope and has to deal with an observation on its own, if that does not fit into a normal logic of the listener container.

Perhaps something similar here for those `@RetryableTopic` variants?
I mean no need to worry about `GenericMessageListener` contract and extra hook on it.
Just have a setter on the `MessagingMessageListenerAdapter` and propagate such a callback only in that case.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1435,6 +1457,21 @@ protected void pollAndInvoke() {
...
+		protected void handleAsyncFailure() {
+			List<FailedRecordTuple> copyFailedRecords = new ArrayList<>();
+			while (!this.failedRecords.isEmpty()) {
+				FailedRecordTuple failedRecordTuple = this.failedRecords.pollFirst();
+				copyFailedRecords.add(failedRecordTuple);
+			}
+
+			// If any copied and failed record fails to complete due to an unexpected error,
+			// We will give up on retrying with the remaining copied and failed Records.
+			if (!copyFailedRecords.isEmpty()) {
```
artembilan, line 1469: I don't think we need this `if` here: if collection is empty its `forEach()` is going to be void.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1435,6 +1457,21 @@ protected void pollAndInvoke() {
...
+				copyFailedRecords.forEach(failedRecordTuple ->
+											this.invokeErrorHandlerBySingleRecord(failedRecordTuple.record(), failedRecordTuple.ex()));
```
artembilan, line 1471: Can we look into something like `this::invokeErrorHandlerBySingleRecord` instead?
Where contract of that method would be `invokeErrorHandlerBySingleRecord(FailedRecordTuple)`.
Either way, we prefer to not use `this.` prefix for method calls.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3913,4 +3986,6 @@ private static class StopAfterFenceException extends KafkaException {
 
 	}
 
+	record FailedRecordTuple<K, V>(ConsumerRecord<K, V> record, RuntimeException ex) { };
```
artembilan, line 3989: I believe this is supposed to be `private` to not let other classes to see it.

spring-kafka/src/main/java/org/springframework/kafka/listener/MessageListener.java
```diff
@@ -26,8 +28,13 @@
  *
  * @author Marius Bogoevici
  * @author Gary Russell
+ * @author Sanghyeok An
  */
 @FunctionalInterface
 public interface MessageListener<K, V> extends GenericMessageListener<ConsumerRecord<K, V>> {
 
+	default void setCallbackForAsyncFailure(
```
artembilan, line 36: As I mentioned in the container review, we might consider just have this contract on the `MessagingMessageListenerAdapter` and that's it.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
```diff
@@ -0,0 +1,1241 @@
+/*
+ * Copyright 2018-2024 the original author or authors.
```
artembilan, line 2: I don't think that such a past is correct to indicate when this class was created: https://stackoverflow.com/questions/9039698/which-years-to-write-into-copyright-header.

Someone used the first year as the one when project has started, but that still is not `2018`.
Therefore I suggest to do just `2024`.
When we change this class in `2025`, some our tool is going to do for us in this class `2024-2025`.
This way we would follow jurisdiction: the year when entity was created and the year when entity was last changed.

Please, see if all other your new classes have to be fixed respectively.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
```diff
@@ -0,0 +1,1241 @@
...
+
+	@EnableKafka
+	@Configuration
+	static class KafkaConsumerConfig {
+
+		@Autowired
+		EmbeddedKafkaBroker broker;
+
+		@Bean
+		KafkaAdmin kafkaAdmin() {
```
artembilan, line 1182: Why do we need admin for this test suite?

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
```diff
@@ -0,0 +1,1241 @@
...
+				if (container.getListenerId().startsWith("manual")) {
+					container.getContainerProperties().setAckMode(AckMode.MANUAL);
+					container.getContainerProperties().setAsyncAcks(true);
+				}
+			});
+			return factory;
+		}
+
+		@Bean
+		TaskScheduler sched() {
```
artembilan, line 1230: Why do we need scheduler?
Is that something what retry topic relies on?

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
```diff
@@ -0,0 +1,1241 @@
...
+
+		// When
+		kafkaTemplate.send(TEST_TOPIC1, shortFailedMsg);
+		kafkaTemplate.send(TEST_TOPIC1, longSuccessMsg);
+
+		// Then
+		assertThat(awaitLatch(latchContainer.countDownLatch1)).isTrue();
+		assertThat(awaitLatch(latchContainer.dltCountdownLatch1)).isTrue();
+
+		waitAWhile(executor, 10000);
```
artembilan, line 222: 10 seconds is too long to sleep for nothing in the unit test.
Now image we have 1000 of tests in the project and each of it does such a sleep for nothing.
Plus take into account how really many threads we can use on CI/CD to make some of them to sleep so long.
Cannot we look into the way to condition it via some async barrier instead?

Do we really need so many tests to cover this functionality?
Do they really all have to be so complex?

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoFutureRetryTopicClassLevelIntegrationTests.java
```diff
@@ -0,0 +1,971 @@
...
+/**
+ * @author Sanghyeok An
+ * @since 3.3.0
+ */
+
+@SpringJUnitConfig
+@DirtiesContext
+@EmbeddedKafka
+@TestPropertySource(properties = { "five.attempts=5", "kafka.template=customKafkaTemplate"})
+public class AsyncMonoFutureRetryTopicClassLevelIntegrationTests {
```
artembilan, line 101: What is the difference between this class and the next `AsyncMonoRetryTopicScenarioTests`?
Same question for all pairs of classes you have introduced here.
I even wonder if we don't have duplication in coveredge: I mean some tests might just do exactly the same what was done before.
Like this one is fully covered by the `CompletableFuture` one.
From the `MessagingMessageListenerAdapter` logic both `CompletableFuture` and `Mono` call the same `asyncFailure()`. So, your new code really does not introduce anything else `Mono`-specific.

My main concern that these new tests are too long to understand what is going on and they are possible too slow according to those 10 seconds sleep everywhere.

artembilan (changes requested):
Sorry for so lengthy review, but in my feeling we are close for merge, so I'd like to clear up things around.

Thank you for incredible work you've done with this flaky use-case!

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoFutureRetryTopicClassLevelIntegrationTests.java
chickenchickenlove: As you mentioned, yes, its test cases is same with the one with `CompletableFuture`. 
When i started to handle this PR, i concerned that the difference of `Mono<?>` and `CompletableFuture<?>` cause an unexpected situation. 
As i concerned about it, it became true. 

image

The one `if ~ else` code which you pointed is the problem that i can find via both `CompletableFuture<?>` and `Mono<?>`. 
```java
handleException(...);
```

If we simplify that code as one line, only `CompletableFuture` test cases will fail and `Mono` test cases will success.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
chickenchickenlove: The reason why i use `waitAWhile(executor, 10000)` is for a more accurate test.
I was concerned about the following scenario.

This is the expected case.
1. Initialize the latch (latch = 3), set the retry count to 3, and initialize the DLT latch to 3.
2. Consumer Failed -> latch 2
3. Consumer Retry 1 -> latch 1
4. Consumer Retry 2 -> latch 0
5. Consumer DLT -> DLT latch 0

What about this case? 
1. Initialize the latch (latch = 3), set the retry count to 3, and initialize the DLT latch to 3.
2. Consumer Failed -> latch 2
3. Consumer Retry 1 -> latch 1
4. Consumer Retry 2 -> latch 0
5. Consumer Retry 3 (Unexpected retry) -> latch 0  (latch should raise complain, but not)
7.  Consumer Retry 4 (Unexpected retry) -> latch 0 (latch should raise complain, but not)
8.  Consumer Retry 5 (Unexpected retry) -> latch 0 (latch should raise complain, but not)
9.  Consumer DLT -> DLT latch 0

What about this case? 
1. Initialize the latch (latch = 3), set the retry count to 3, and initialize the DLT latch to 3.
2. Consumer Failed -> latch 2
3. Consumer Retry 1 -> latch 1
4. Consumer Retry 2 -> latch 0
5. Consumer DLT -> DLT latch 0
6. Consumer DLT (Unexpected dlt) -> DLT latch 0 (latch should raise complain, but not)


When the latch overflows, it doesn't raise any complaints. 
In such case, we might consider the test a success even though it actually fails.

However, this could be due to my misunderstanding about the async barrier, as well. 
As you mentioned, I think 10 sec sleep is too long for a unit test case. 

We can remove `waitAWhile(exeuctor, 10000)` or decrease waiting time to 1000 ~ 3000. (it also too long, i guess. but too short value is meaningless i think)

What do you think?

artembilan: I believe we just need one latch for the number of retries. And some holders (or mocks) for interim actions if we want to be sure that something happens in between. The problem with your tests that they are too complex to digest and understand the logic we expect. In the end it is going to be some burden to maintain such a complexity in the tests from now on. Especially when we have so non-reliable timing barriers and for so long period. 

In general the point of tests is to call real code and verify it outcome , but not more code which does not make thing easier to reflect on.
I also cannot tell what and how to make simpler in these tests since I’m not get well with this retry topic feature. 

Maybe @sobychacko will judge us with his fresh perspective on these new tests?

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
chickenchickenlove: `TaskScheduler` need for `RetryTopicConfiguration`! 
https://github.com/spring-projects/spring-kafka/blob/6f4adb2146a7d27416fdb7810bd0771eeb1022b1/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationSupport.java#L310-L333

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
chickenchickenlove: I think so, too. 
But the problem is that multiple type casts are required. 
However, I prefer type casting over introducing unnecessary contracts to the interface, which would cause more significant changes! 😅 

I removed callback interface from both `MessageListener` and `GenericMessageListener` interface!

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
chickenchickenlove: Yes, Please give me you guys opinions! 

FYI, 
If we remove `waitAWhile(executor, 10000)` from these test cases, It doesn't affect the test result. 
I added `waitAWhile(executor, 10000)` due to the concern mentioned before, but in reality, that concern did not occur. 😅

sobychacko: @artembilan @chickenchickenlove I will take a look today.

sobychacko: @chickenchickenlove Do you see the test ever wait for 10 seconds while running it? We have other tests where 10 seconds might be a default wait time, but they rarely wait for that long to get the latch released. If this test indeed takes 10 seconds, then that might be an issue.

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncCompletableFutureRetryTopicClassLevelIntegrationTests.java
```diff
@@ -0,0 +1,966 @@
...
+
+		@Autowired
+		CountDownLatchContainer container;
+
+		@KafkaHandler
+		public CompletableFuture<Void> listen(String message, @Header(KafkaHeaders.RECEIVED_TOPIC) String receivedTopic) {
+			return CompletableFuture.supplyAsync(() -> {
+				container.countDownLatch1.countDown();
+				try {
+					Thread.sleep(1);
```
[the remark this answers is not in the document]
chickenchickenlove: Wow, this is something I hadn't thought of at all. 
Basically, I didn't consider it to be a problem.
The compiler did complain about the `CheckedException` when `Thread.sleep(1)` was used. 
So, i just caught it to silence the compiler.

I removed `Thread.sleep(1)`. 
We don't need it.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncCompletableFutureRetryTopicScenarioTests.java
```diff
@@ -0,0 +1,1263 @@
...
+		// When
+		kafkaTemplate.send(TEST_TOPIC0, shortFailedMsg1);
+		kafkaTemplate.send(TEST_TOPIC0, shortFailedMsg2);
+		kafkaTemplate.send(TEST_TOPIC0, shortFailedMsg3);
+
+		// Then
+		assertThat(awaitLatch(latchContainer.countDownLatch0)).isTrue();
+		assertThat(awaitLatch(latchContainer.dltCountdownLatch0)).isTrue();
+
+		waitAWhile(executor, 10000);
```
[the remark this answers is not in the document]
chickenchickenlove: I removed `waitAWhile(executor, 10000)` from all scenario tests! 

> what is the difference (at a glance) between those integration tests and these scenarios, please?

@artembilan My bad, i missed it. 🙇‍♂️ 
1. The `RetryTopicClassLevelIntegrationTests` class is base class of `Async(CompletableFuture | Mono)RetryTopicClassLevelIntegrationTests`. 
2. These async retry integration test cases follow the rules of the `RetryTopicClassLevelIntegrationTests`. In other words, they demonstrate that the async retry feature works well in the synchronous conditions. 
3. However, in the async condition, the message with offset 1 can fail earlier than the message with offset 0. this situation never occur in the synchronous condition. because, `ConsumerRecord` is invoked one by one.
4. The async scenario tests demonstrate that the async retry feature works well with the situation which i mentioned at 3.

I am concerned that async retry cases may affect the `offset` because the async retry feature follows the rules of `FailedRecordTracker` and `SeekUtils`. As a result of seeking the wrong offset, i am worried about the cases where a message coulde be re-consumed infinitely or some offsets could be skipped to be retried.

the async scenario test cases demonstrate this.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
chickenchickenlove: @sobychacko , @artembilan Thanks for you guys comments.
I fixed it! 

At the first, I imagined a special edge case. 
If the async retry feature might affect the consumer offset (for example, if offset 2 success, offset 1 fails), 
I was worried about a case where a message could be re-consumed infinitely due to seeking the wrong offset. 

However, after reconsidering, we don't need `waitAWhile(...)` to check this edge case. 
```java
protected final List<String> receivedMsgs = new ArrayList<>();
private final List<String> receivedTopics = new ArrayList<>();
```
because in these test cases, the listener writes all the messages and topics it receives.
So, we can check this edge case by asserting that both `receivedMsgs` and `receivedTopics` are equal  to the `expectedValue`.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncCompletableFutureRetryTopicClassLevelIntegrationTests.java
```diff
@@ -0,0 +1,966 @@
...
+		@SuppressWarnings("unchecked")
+		Collection<NewTopic> weededTopics = (Collection<NewTopic>) method.get().invoke(admin);
+		AtomicInteger weeded = new AtomicInteger();
+		weededTopics.forEach(topic -> {
+			if (topic.name().equals(THIRD_TOPIC) || topic.name().equals(FOURTH_TOPIC)) {
+				assertThat(topic).isExactlyInstanceOf(NewTopic.class);
+				weeded.incrementAndGet();
+			}
+		});
+		assertThat(weeded.get()).isEqualTo(2);
```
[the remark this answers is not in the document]
chickenchickenlove: The `AsyncCompletableFutureRetryTopicClassLevelIntegrationTests` demonstrates that the async retry feature works well in `RetryTopicClassLevelIntegrationTests`. It means the synchronous condition. (`ConsumerRecord` succeed or fail one by one)

So, I did copy and paste `RetryTopicClassLevelIntegrationTests` and replace `KafkaListener`'s return type to `Mono<?>` or `CompletableFuture<?>`.

In the `RetryTopicClassLevelIntegrationTests`, `weeded` seems to demonstrate that some new topics should be created. 
image

As a result of `admin.newTopics()`, instance `weededTopics` has two types (`TopicForRetryable`, `NewTopic`).
I think that the writer of `RetryTopicClassLevelIntegrationTests` want to verify whether expected `NewTopic`s are created. 

If there are no major issues, I would like to follow the assumptions of `RetryTopicClassLevelIntegrationTests` as well. 
This is because I believe that `AsyncRetry` should also satisfy the `synchronous retry` condition where offsets are processed sequentially, given that AsyncRetry may or may not process offsets in sequence.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncCompletableFutureRetryTopicScenarioTests.java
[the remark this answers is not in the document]
chickenchickenlove: I think the async retry feature should include not only `CompletableFuture<?>` and `Mono<?>` test cases  based on `RetryTopicClassLevelIntegrationTests` but also `Async*ScenarioTests`. 

`CompletableFuture<?>` and `Mono<?>` test cases  based on `RetryTopicClassLevelIntegrationTests` demonstrate sequential failures (e.g., offset 0 fail -> offset 1 fail -> ....). 

However, `Async*ScenarioTests` demonstrate non-sequential failures. 

For example, i imagine the following situations.
1. put `Mono<?>` instance with Offset 0 to Executor.
2. put `Mono<?>` instance with Offset 1 to Executor. 
3. put `Mono<?>` instance with Offset 2 to Executor. 
4. `Mono<?>` instance with Offset 2 succeed after 100ms.
5. `Mono<?>` instance with Offset 0 failed after 500ms.
6. `Mono<?>` instance with Offset 0 should be retry. (Don't retry offset 1, 2)
7. `Mono<?>` instance with Offset 1 failed after 600ms.
8. `Mono<?>` instance with Offset 1 should be retry. (Don't retry offset 0, 2)


```java
kafkaTemplate.send(TEST_TOPIC3, TestTopicListener3.LONG_FAIL_MSG);
kafkaTemplate.send(TEST_TOPIC3, TestTopicListener3.LONG_FAIL_MSG);
kafkaTemplate.send(TEST_TOPIC3, TestTopicListener3.SHORT_SUCCESS_MSG);
kafkaTemplate.send(TEST_TOPIC3, TestTopicListener3.SHORT_SUCCESS_MSG);
kafkaTemplate.send(TEST_TOPIC3, TestTopicListener3.SHORT_SUCCESS_MSG);
```

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3299,6 +3385,10 @@ private Collection<ConsumerRecord<K, V>> getHighestOffsetRecords(ConsumerRecords
 					.values();
 		}
 
+		private BiConsumer<ConsumerRecord<K, V>, RuntimeException> getCallbackForAsyncFailure() {
```
artembilan, line 3388: Well. I meant this contract:
```
private void callbackForAsyncFailure(ConsumerRecord<K, V> cRecord, RuntimeException ex) {
    this.failedRecords.addLast(new FailedRecordTuple<>(cRecord, ex));
}
```

Then it could be used like this:
```
adapterListener.setCallbackForAsyncFailure(this::callbackForAsyncFailure);
```

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -869,6 +886,17 @@ private boolean rawByParameterIsType(Type parameterType, Type type) {
 		return parameterType instanceof ParameterizedType pType && pType.getRawType().equals(type);
 	}
 
+	/**
+	 * Sets the retry callback for failures of both {@link CompletableFuture} and {@link Mono}.
```
artembilan, line 890: Method JavaDocs have to be imperative, like commands.
Therefore `Set the retry callback...`.
Also `@since 3.3` still has to be present on the public API.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncCompletableFutureRetryTopicClassLevelIntegrationTests.java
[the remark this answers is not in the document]
artembilan: I still don't see how this `weeded` are relevant to async handling.
That is really not a part of retry topic behavior, but rather its configuration.
So, having same verifications in two different places is redundant.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
```diff
@@ -0,0 +1,1178 @@
...
+
+		private final List<String> receivedTopics = new ArrayList<>();
+
+		@KafkaHandler
+		public Mono<Void> listen(String message, @Header(KafkaHeaders.RECEIVED_TOPIC) String receivedTopic) {
+			this.receivedMsgs.add(message);
+			this.receivedTopics.add(receivedTopic);
+			return Mono.fromCallable(() -> {
+				container.countDownLatch0.countDown();
+				throw new RuntimeException("Woooops... in topic " + receivedTopic);
```
artembilan, line 549: This is still not OK.
The `countDown()` has to happen after we throw an exception.
So, it has to go to the `finally`.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
```diff
@@ -0,0 +1,1178 @@
...
+		private final List<String> receivedTopics = new ArrayList<>();
+
+		@KafkaHandler
+		public Mono<String> listen(String message, @Header(KafkaHeaders.RECEIVED_TOPIC) String receivedTopic) {
+			this.receivedMsgs.add(message);
+			this.receivedTopics.add(receivedTopic);
+
+			return Mono.fromCallable(() -> {
+				try {
+					Thread.sleep(Integer.parseInt(message));
```
artembilan, line 578: Why do we have to sleep on every message?
I don't believe that there is any value in stalling threads over here.

I understand that you may copied/pasted tests from somewhere, but can we give it more thoughts to avoid some undesired slowness?

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
chickenchickenlove: @artembilan 
Sorry for my lack of description.  🙇‍♂️ 
As i mentioned before, it is my intent to test async case. (I didn't copy/paste in case of **scenarioTest!)

The `TestTopicListener1` which you pointed out is used for the test `firstShortFailAndLastLongSuccessRetryTest`.

In the test code, you can see this.
```java
		// Given
		String longSuccessMsg = "3";
		String shortFailedMsg = "1";

...

		// When
		kafkaTemplate.send(TEST_TOPIC1, shortFailedMsg);
		kafkaTemplate.send(TEST_TOPIC1, longSuccessMsg);

```
It means that
1. The msg with offset 0 is invoked to listener method. it put `Mono<?>` `to Exectutor`
2. The msg with offset 1 is invoked to listener method. it put `Mono<?>` `to Exectutor`
3. The msg offset 0 is succeed after 1ms sleep.  (or this step can occur between step 1 and step 2)
4. The msg offset 1 is succeed after 3ms sleep. 


For other example, 
```java
		kafkaTemplate.send(TEST_TOPIC3, TestTopicListener3.LONG_FAIL_MSG);
		kafkaTemplate.send(TEST_TOPIC3, TestTopicListener3.LONG_FAIL_MSG);
		kafkaTemplate.send(TEST_TOPIC3, TestTopicListener3.SHORT_SUCCESS_MSG);
		kafkaTemplate.send(TEST_TOPIC3, TestTopicListener3.SHORT_SUCCESS_MSG);
		kafkaTemplate.send(TEST_TOPIC3, TestTopicListener3.SHORT_SUCCESS_MSG);
```
It means that
1. The msg with offset 0 is invoked to listener method. it put `Mono<?>` to `Exectutor`.
2. The msg with offset 1 is invoked to listener method. it put `Mono<?>` to `Exectutor`.
3. The msg with offset 2 is invoked to listener method. it put `Mono<?>` to `Exectutor`.
4. The msg with offset 3 is invoked to listener method. it put `Mono<?>` to `Exectutor`.
5. The msg with offset 4 is invoked to listener method. it put `Mono<?>` to `Exectutor`.
6. The msg with offset 2 succeed after 1ms sleep. (or it can occur between step 3 and step 4)
7. The msg with offset 3 succeed after 1ms sleep. (or it can occur between step 4 and step 5)
8. The msg with offset 4 succeed after 1ms sleep. (or it can occur between step 5 and step 6)
9. The msg with offset 0 failed after 5000ms sleep.
10. The msg with offset 1 failed after 5000ms sleep.
11. The msg with offset 0 at retry topic is invoked to listener method.  (or it can occur between step 9 and 10)
12. The msg with offset 1 at retry topic is invoked to listener method. 

This is what i consider a typical async case. 
The reason I set `LONG_FAIL_MSG` to 5000ms is that I wanted to give enough time for the `KafkaMessageListenerContainer.run()` loop to pass 1-2 times.  
(Looking at it now, it seems excessive, so I'll reduce it for now! 😅 )

A change in the loop means that a new `ConsumerRecord` has been fetched, which implies moving to the next offset regardless of the async fail. This was exactly what I wanted to test above, and it was the expected behavior in the async retry process.

What do you think? 
Please let me know! 🙇‍♂️

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -895,6 +902,15 @@ else if (listener instanceof MessageListener) {
 				this.wantsFullRecords = false;
 				this.pollThreadStateProcessor = setUpPollProcessor(false);
 				this.observationEnabled = this.containerProperties.isObservationEnabled();
+
+				if (!AopUtils.isAopProxy(this.genericListener) &&
+					this.genericListener instanceof AbstractDelegatingMessageListenerAdapter<?>) {
```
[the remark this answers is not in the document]
chickenchickenlove: In the runtime, `KafkaBackOffAwareMessageListenerAdapter` is set for `this.genericListener` and `this.listener`. 
The `KafkaBackOffAwareMessageListenerAdapter` has reference of `RecordMessagingMessageListenerAdapter` as its `delegate`.

Please refer to a image below 🙇‍♂️ 
image

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncMonoRetryTopicScenarioTests.java
artembilan: Yeah, I see. Thanks for the explanation ! Let it leave for now as you have designed! 
From my experience it is flaky to rely on time between assertions: there is still no guarantee which thread is going to be executed first and so on. We do use `CountDownLatch` instead, but I agree that it is too complicated for your use-cases here. In the end we may just tag these tests somehow to run them conditionally.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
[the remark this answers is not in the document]
artembilan: Ok. What I mean that it might be better to have a generic solution. We with @sobychacko are working for async observation and gave sumilar condition almost at the same line of code. So, I’m anticipating to combine them in the end. Therefore this clarification is useful . Thanks

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/AsyncCompletableFutureRetryTopicScenarioTests.java
```diff
@@ -68,6 +69,7 @@
  * @since 3.3.0
  */
 
+@Tag("async-retry-flaky-test")
```
artembilan, line 72: I’m sorry, I wasn’t clear: we can add tags eventually if these tests will prove to be sporadically failing.

chickenchickenlove: My bad 🙇‍♂️ 
Fixed it!

chickenchickenlove:
@artembilan Hi. 
I have a question. 

Don't we need test codes for this situation? 🤔 
- Consumer Listener method which subscribe `topicA` returns `CompletableFuture<?>` or `Mono<?>` 
- The `topicA` without `RetryTopicConfiguration`.

I didn't wrote any test codes for this situation.
I haven't looked into it yet, but it seems there is no test code for this case. 

Do you think I should write it? What are your thoughts?
Please let me know. Thanks a lot 🙇‍♂️

artembilan:
> Don't we need test codes for this situation? 🤔

>    Consumer Listener method which subscribe topicA returns CompletableFuture<?> or Mono<?>
>    The topicA without RetryTopicConfiguration.


Sorry, I'm not totally sure in your question.
Why doesn't the `AsyncListenerTests` meet your expectations?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -665,16 +671,27 @@ protected void asyncFailure(Object request, @Nullable Acknowledgment acknowledgm
...
-							"Async Fail", source.getPayload()), t));
+							new ListenerExecutionFailedException(createMessagingErrorMessage(
+									"Async Fail", source.getPayload()), cause));
 		}
 		catch (Throwable ex) {
 			this.logger.error(t, () -> "Future, Mono, or suspend function was completed with an exception for " + source);
 			acknowledge(acknowledgment);
+			if (canAsyncRetry(request, ex) &&
+				Objects.nonNull(this.callbackForAsyncFailure)) {
+				ConsumerRecord<K, V> record = (ConsumerRecord<K, V>) request;
```
artembilan, line 684:
```
/home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java:684: warning: [unchecked] unchecked cast

> Task :spring-kafka:compileJava
				ConsumerRecord<K, V> record = (ConsumerRecord<K, V>) request;
```

Consider to add a `@SuppressWarnings("unchecked")`

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -665,16 +671,27 @@ protected void asyncFailure(Object request, @Nullable Acknowledgment acknowledgm
...
+				this.callbackForAsyncFailure.accept(record, (RuntimeException) ex);
+			}
 		}
 	}
 
+	private boolean canAsyncRetry(Object request, Throwable exception) {
```
artembilan, line 690: Looks like has to be `static`.

artembilan (changes requested):

chickenchickenlove:
> Sorry, I'm not totally sure in your question.
> Why doesn't the AsyncListenerTests meet your expectations?

Sorry for lack of explanation 🙇‍♂️ 
Yes, `AsyncListenerTest` seems to meet my expectations. 

For your information, 
In the `AsyncListenerTest`, `this.genericListener` in `KafkaMessageListenerContainer` is `RecordMessagingMessageListenerAdapter`. so, it doesn't have `asyncFailureCallback`. 
So, even if `Object errorResult = this.errorHandler.handleError(...);` throws error, it will never call `asyncFailureCallback.`

However, in the `AsyncMonoRetryTopicScenarioTests`, `this.genericListener` is `KafkaBackoffAwareMessageListenerAdapter`. so, it has `asyncFailureCallback`. so it can call `asyncFailureCallback` after `Object errorResult = this.errorHandler.handleError(...);` throws error.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
chickenchickenlove: Github seems to have UI error. 😓 
Fixed it, but Github UI don't be aware of it.
image

artembilan:
> In the `AsyncListenerTests`, `this.genericListener` in `KafkaMessageListenerContainer` is `RecordMessagingMessageListenerAdapter`. so, it doesn't have `asyncFailureCallback`.

That's exactly what I meant about checking against `this.genericListener` first and then against `delegate`.
However this might be really good sign to not do that for regular async listener.
The error over there is handled by the `MessagingMessageListenerAdapter` itself.
There is nothing to do with retries, so all good so far.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
artembilan: Not sure what you'd like to see here.
The `@SuppressWarnings("unchecked")` is in and build is OK.
You might be wondering why this conversation is not closed.
Just because you made a change on a different line.
So, so far so good.
Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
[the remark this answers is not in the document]
artembilan: So, I think your previous solution was correct.
We have to set callback only in case of a `KafkaBackoffAwareMessageListenerAdapter` which is indeed an indicator that we are in retry topic scenario.
This way we won't break all other use-cases where such an error handling would be unexpected.

artembilan (changes requested):

chickenchickenlove:
> That's exactly what I meant about checking against this.genericListener first and then against delegate.
> However this might be really good sign to not do that for regular async listener.
> The error over there is handled by the MessagingMessageListenerAdapter itself.
> There is nothing to do with retries, so all good so far.

As per your advice, I'm glad I debugged the AsyncListener test code. 
Thanks to that, I feel more confident about writing more stable code.  🙇‍♂️

artembilan (approved):
Pulling locally for final review and possible merge.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/9eafa610e8239d79320f5889af5120331902b4a2.

@chickenchickenlove ,

thank you for awesome contribution as always!

So, I've accepted those new tests as well, although they are indeed some kind of slow.
We would need to think about filtering them if they will be some maintenance burden eventually.

A couple remarks about contribution overall:

1. It is very convenient to have a branch named after GH issue.
2. It is better to mention in the commit message an issue your are fixing.
3. Adding something like `Fixed: #3276` would close that issue automatically on merge.
4. The first commit of the PR becomes a description of the PR.
5. It is great to have a good commit message to preserve a history and motivation of the change.
The PR is fully not mentioned in the commit history, but looking into the commit I would see what is going on and why without looking for an issue and related PR.

See more info here: https://cbea.ms/git-commit/

chickenchickenlove:
Thanks for your time and awesome review, always! 🙇‍♂️ 
I will look into ways to shorten the test times by using a latch, as you suggested!. 
Thanks again, always!

artembilan:
So, yeah, we really got some failing tests already from this story:
```
AsyncCompletableFutureRetryTopicScenarioTests > oneLongSuccessMsgBetween100ShortFailMsg(TestTopicListener5, MyCustomDltProcessor) FAILED
    Expecting value to be true but was false
        at java.base@17.0.12/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at java.base@17.0.12/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
        at java.base@17.0.12/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.base@17.0.12/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:500)
        at app//org.springframework.kafka.retrytopic.AsyncCompletableFutureRetryTopicScenarioTests.oneLongSuccessMsgBetween100ShortFailMsg(AsyncCompletableFutureRetryTopicScenarioTests.java:429)
```

chickenchickenlove:
I see 🥲.
Sorry for inconvenience.
I will fix it by using latch in coming weekend.

Issue #3448, opened 2024-08-20
childContainer of previous run is stopping ConcurrentContainer after a new start


LokeshAlamuri:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

For example:

3.3

**Describe the bug**

If a ConcurrentContainer is stopped, then the child container should not be allowed to stop ConcurrentContainer. But, there are some scenarios where it is possible. 

**Scenario:**

Concurrency: 2

ConcurrentContainer:: CMain
Child containers: C0, C1.

1) `ConcurrentContainer` started.
    
	CMain -- running.
	C0 -- running.
	C1 -- running.
	
2) `ConcurrentContainer` stopped.
	
	CMain -- not running.
	C0 -- delinked. (message processing is happening)
	C1 -- delinked.

3) `ConcurrentContainer` started. This is permitted since `stop` is called before. Nothing wrong here. It should be allowed. Only the practise is not correct.
   
   	CMain -- running.
	C2 -- running.
	C3 -- running.

	C0 -- delinked. (message processing is happening)

4) `C0` has thrown error while processing. This would stop the running `ConcurrentContainer` !!!!
    
	CMain -- not running.
	C2 -- delinked.
	C3 -- delinked.

	C0 -- delinked. 

**To Reproduce**

```
@Test
	public void testFencedContainerFailed() throws Exception {
		this.logger.info("Start testFencedContainerFailed");
		Map<String, Object> props = KafkaTestUtils.consumerProps("test1", "true",
				embeddedKafka);
		AtomicReference<Properties> overrides = new AtomicReference<>();
		DefaultKafkaConsumerFactory<Integer, String> cf = new DefaultKafkaConsumerFactory<Integer, String>(props) {
			@Override
			protected Consumer<Integer, String> createKafkaConsumer(String groupId, String clientIdPrefix,
																	String clientIdSuffixArg, Properties properties) {
				overrides.set(properties);
				return super.createKafkaConsumer(groupId, clientIdPrefix, clientIdSuffixArg, properties);
			}
		};
		ContainerProperties containerProps = new ContainerProperties(topic1);
		containerProps.setLogContainerConfig(true);
		containerProps.setClientId("client");
		containerProps.setAckMode(ContainerProperties.AckMode.RECORD);

		final CountDownLatch secondRunLatch = new CountDownLatch(5);
		final Set<String> listenerThreadNames = new ConcurrentSkipListSet<>();
[...]

PR #3537, opened 2024-10-05
Forbid fenced Container to stop ConcurrentContainer

LokeshAlamuri:
[DRAFT]
Fixes #GH-3448

**Issue**: Fenced Child Container could stop the running ConcurrentContainer 
**Fix**: Configure KafkaMessageListenerContainer (KMLC) to use ConcurrentMessagleListenerContainerRef instead of
ConcurrentContainer. Internally, ConcurrentContainerRef checks if KMLC is fenced when stop operations are called on Concurrent Container. If KMLC is fenced, suppress the `stop` related operations. If KMLC is not fenced, delegate the stop call to ConcurrentContainer.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -193,6 +193,14 @@ public abstract class AbstractMessageListenerContainer<K, V>
 		}
 	}
 
+	/**
+	 *	To be used only with {@link ConcurrentMessageListenerContainerRef}.
+	 */
+	AbstractMessageListenerContainer() {
+		this.containerProperties = null;
+		this.consumerFactory = null;
+	}
+
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
 		this.applicationContext = applicationContext;
@@ -280,6 +288,10 @@ public abstract class AbstractMessageListenerContainer<K, V>
 
 	protected void setFenced(boolean fenced) {
 		this.fenced = fenced;
+	}
+
+	boolean isFenced() {
+		return this.fenced;
 	}
 
 	@Deprecated(since = "3.2", forRemoval = true)
@@ -722,7 +734,7 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	protected void publishContainerStoppedEvent() {
 		ApplicationEventPublisher eventPublisher = getApplicationEventPublisher();
 		if (eventPublisher != null) {
-			eventPublisher.publishEvent(new ContainerStoppedEvent(this, parentOrThis()));
+			eventPublisher.publishEvent(new ContainerStoppedEvent(this, parentContainerOrThis()));
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


sobychacko:
@LokeshAlamuri Thanks for the PR. Have you run into it as a bug in your application? We see that you have a test on the related issue. Could you let me know if that is enough to reproduce the issue? As @artembilan mentioned, this PR has many changes that touch the critical elements of the framework. Since we are about to go into an RC phase in our release cycle, we may not want to introduce any potential regression issues in core components unless they are critical bugs. Because of that reason, I would vote to revisit this issue in the next minor version after `3.3.x`, but we can discuss it. Thanks!

LokeshAlamuri:
> I'm sorry. That's too many changes. What is really the problem? And why would `KMLC` call the stop on its parent at all?

[sobychacko](https://github.com/sobychacko) [artembilan](https://github.com/artembilan) Thanks a lot for allocating your time and reviewing the issue and **draft version of PR**  I have submitted. As mentioned in the issue I have reported, this is possible in the following scenario.

**Configuration**:

1) `CommonContainerStoppingErrorHandler` is chosen to stop the Container on failure.
2) Chose concurreny is 2.

**Scenario**:

1) ConcurrentContainer is started.
2) ConcurrentContainer is stopped.
3) ConcurrentContainer is started. Assume, ConcurrentContainer is started immediately before all the KMLC are stopped.
4) Now if one of the KMLC of previous run has some error while processing previous record, it would stop the running ConcurrentContainer. This is not supposed to happen.




This is really a complex issue which would require changes in multiple files and APIs as well. This PR I have submitted could help in understanding the design issue and possible fix. The issue is not only with the `stop` but other APIs as well. I personally feel, ConcurrentContainer instance is freely passed around ErrorHandlers and Events. This is causing the real issue.

LokeshAlamuri:
> Have you run into it as a bug in your application? We see that you have a test on the related issue. Is that enough to reproduce the issue?

Yes, the Junits I have submitted would be 100% sufficient to reproduce the issue.

LokeshAlamuri:
> this PR has many changes that touch the critical elements of the framework. Since we are about to go into an RC phase in our release cycle, we may not want to introduce any potential regression issues in core components unless they are critical bugs. Case in point - a recent change in this area accidentally slowed down one of our tests in a downstream framework, which we only noticed a few weeks later due to the subtle nature of the issue. Because of that reason, I would vote to revisit this issue in the next minor version after `3.3.x`, but we can discuss it. Thanks!

I shall follow your suggestions and provide the information required.

artembilan:
@LokeshAlamuri ,

can you confirm, please, that this piece of code from the `KafkaMessageListenerContainer` is guilty in the picture:
```
		/**
		 * Handle exceptions thrown by the consumer outside of message listener
		 * invocation (e.g. commit exceptions).
		 * @param e the exception.
		 */
		protected void handleConsumerException(Exception e) {
			if (e instanceof RetriableCommitFailedException) {
				this.logger.error(e, "Commit retries exhausted");
				return;
			}
			try {
						this.commonErrorHandler.handleOtherException(e, this.consumer,
								KafkaMessageListenerContainer.this.thisOrParentContainer, this.isBatchListener);
				}
				else {
					this.logger.error(e, "Consumer exception");
				}
			}
			catch (Exception ex) {
				this.logger.error(ex, "Consumer exception");
			}
		}
```
So, whenever the `KMLC` handles an exception, it calls that `handleOtherException()` providing its `KafkaMessageListenerContainer.this.thisOrParentContainer`.
Can it be just replaced into just this whenever the `KMLC` is `fenced`?

LokeshAlamuri:
We can do that. But, still the container could be fenced when it is inside error handler. In that case, suggested fix will not work. This is a bit design issue, where problem is not just with `errorhandlers` and `stop` API.  It could be a `pause` call on ConcurrentContainer and inside the EventListeners also.

artembilan:
> This is a bit design issue, where problem is not just with errorhandlers and stop API

I see, so even if we check for `fenced` just before calling those API, there is still a race condition when we call API, but `fenced` is set after that, and target user code would suffer from the problem you are describing.

So, you probably right: something like proxy for the container to push to those APIs where we would check for `fenced` on each container hooks.

LokeshAlamuri:
Closing PR as per comments.

Issue #3407, opened 2024-08-05
KafkaHeaders.DELIVERY_ATTEMPT is not added for batch listeners


lm231290:
Version 3.3.2

When using blocking retries for a batch listener, the attempt number is not coming in DELIVERY_ATTEMPT header even if setting `setDeliveryAttemptHeader(true)`

The method `KafkaMessageListenerContainer.ListenerConsumer.internalHeaders` is called only in 
`ConsumerRecord<K, V> checkEarlyIntercept(ConsumerRecord<K, V> recordArg)` which is used for a regular listener, and is not called in 
`ConsumerRecords<K, V> checkEarlyIntercept(ConsumerRecords<K, V> nextArg)` which is used for a batch listener

lm231290:
If the whole batch fails, the retry behavior would be pretty much same as if the first record in a batch fails, is my assumption correct?
So If the header is added for `batchListenerFailedException`, I think it can also be added for any other (whole batch) errors as it would for batchListenerFailedException.index = 0

artembilan:
I'm not sure what you mean with a `BatchListenerFailedException` analogy, but I'd be happy to review some contribution for better understanding how this request could be fixed.
Thanks

artembilan:
The logic around this header is not that simple as it sounds. 
See `FailedRecordTracker` for details.

lm231290:
Seems like my analogy with BatchListenerFailedException is incorrect.

So my suggestion would be to call 
`KafkaMessageListenerContainer.internalHeaders` or 
`KafkaMessageListenerContainer.checkEarlyIntercept` from 
`KafkaMessageListenerContainer.doInvokeBatchListener` and 
`KafkaMessageListenerContainer.invokeBatchListenerInTx`. 

Currently `KafkaMessageListenerContainer.checkEarlyIntercept` is called only from 
`KafkaMessageListenerContainer.invokeRecordListenerInTx` and 
`KafkaMessageListenerContainer.doInvokeWithRecords`

artembilan:
That's correct. But the problem with the rest of missing API that we deal with the whole batch.
Therefore it is not clear how to inject that `KafkaHeaders.DELIVERY_ATTEMPT` header and handle it respectively.

If you have anything in mind, feel free to open pull request and we gladly review it.

lm231290:
I am not so deep into the code, are there any contributors who could take care of it? Maybe the issue should be promoted as a feature request?

artembilan:
[...]

PR #3539, opened 2024-10-07
GH-3407 : Support KafkaHeaders.DELIVERY_ATTEMP for batch listeners.

chickenchickenlove:
### Motivation:
The developer expects to get the `KafkaHeaders.DELIVERY_ATTEMPT` header from the received `ConsumerRecord`. 
It works well with the `SingleRecordListener`. 
However, it does not work well with the `BatchListener` even when `containerProperties.setDeliveryAttemptHeader(true)` is set. 
If the `KafkaContainerListener` can include the `KafkaHeaders.DELIVERY_ATTEMPT` header for each `ConsumerRecord` with the `BatchListener`, it will be more helpful to people who they expect.

### Modifications
- Add `DeliveryAttemptAwareRetryListener` class.
- Add unit tests and integration tests for `DeliveryAttemptAwareRetryListener`. 

### To Reviewer
With current `spring-kafka` codes, it is hard to integrate it to `FailedRecordTracker`. 
By the way, the function `ErrorHandlingUtils.retryBatch()` has deliveryAttempt count as their local scope and has same reference for `ConsumerRecord`. 
Thus, the `DeliveryAttemptAwareRetryListener` is easier way to achieve the goal. 
Because, in each fail to invocation, `ErrorHandlingUtils.retryBatch()` calls `listeners.forEach(listener -> listener.failedDelivery(records, thrownException, attempt)`.

If the developer want to use this feature, they should create `ContainerFactory` and `DeliveryAttemptAwareRetryListener`. also, they set `DeliveryAttemptAwareRetryListener` as `RetryListener`. 
Please refer to code below.

```java
final FixedBackOff fixedBackOff = new FixedBackOff(1000L, MAX_ATTEMPT_COUNT1);
DefaultErrorHandler errorHandler = new DefaultErrorHandler(fixedBackOff);
errorHandler.setRetryListeners(new DeliveryAttemptAwareRetryListener());

ConcurrentKafkaListenerContainerFactory<String, String> factory = new ConcurrentKafkaListenerContainerFactory<>();
factory.setConsumerFactory(consumerFactory);
factory.setCommonErrorHandler(errorHandler);
```

### Result:
- Fixes https://github.com/spring-projects/spring-kafka/issues/3407


spring-kafka/src/main/java/org/springframework/kafka/listener/DeliveryAttemptAwareRetryListener.java
```diff
@@ -0,0 +1,78 @@
...
+
+public class DeliveryAttemptAwareRetryListener implements RetryListener {
+
+	@Override
+	public void failedDelivery(ConsumerRecord<?, ?> record, Exception ex, int deliveryAttempt) {
+		// Pass
+	}
+
+	/**
+	 * Called after a delivery failed for batch records.
```
artembilan, line 48: The method JavaDoc must be imperative. So, `Call after`.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeliveryAttemptAwareRetryListener.java
```diff
@@ -0,0 +1,78 @@
...
+	 * If the {@link KafkaHeaders}.DELIVERY_ATTEMPT header already exists in the {@link ConsumerRecord}'s headers,
+	 * it will be removed. Then, the provided `deliveryAttempt` is added to the {@link ConsumerRecord}'s headers.
+	 *
```
artembilan, line 51: No blank lines in the method JavaDoc

spring-kafka/src/main/java/org/springframework/kafka/listener/DeliveryAttemptAwareRetryListener.java
```diff
@@ -0,0 +1,78 @@
...
+	 */
+	@Override
+	public void failedDelivery(ConsumerRecords<?, ?> records, Exception ex, int deliveryAttempt) {
+		for (ConsumerRecord<?, ?> record : records) {
+
+			Headers headers = record.headers();
+			int headerCount = 0;
+			Iterable<Header> iterator = record.headers().headers(KafkaHeaders.DELIVERY_ATTEMPT);
+			for (Header header : iterator) {
+				headerCount += 1;
```
artembilan, line 64: I think just `headers.remove(KafkaHeaders.DELIVERY_ATTEMPT)` is enough.
No need to do this extra loop.
You just don't check the content of the header, but fully override it.

spring-kafka/src/test/java/org/springframework/kafka/listener/DeliveryAttemptAwareRetryListenerIntegrationTest.java
```diff
@@ -0,0 +1,308 @@
...
+
+/**
+ * @author Sanghyeok An
+ * @since 3.3.0
+ */
+
+@SpringJUnitConfig
+@DirtiesContext
+@EmbeddedKafka
+class DeliveryAttemptAwareRetryListenerIntegrationTest {
```
artembilan, line 66: All the test must be called `*Tests`

spring-kafka/src/test/java/org/springframework/kafka/listener/DeliveryAttemptAwareRetryListenerTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+import org.springframework.kafka.support.KafkaHeaders;
+
+/**
+ * @author Sanghyeok An
+ * @since 3.3
+ *
+ */
+
+class DeliveryAttemptAwareRetryListenerTest {
```
artembilan, line 40: DITTO.

artembilan (changes requested):
I see now how this makes sense.
Thank you for fixing it!

Please, consider to mention a new feature in the docs, including a short entry with a link into target doc in the `whats-new.adoc`.

chickenchickenlove:
@artembilan thanks for your time and comments! 
I made a new commit to apply your comments.
When you have time, please take another look 🙇‍♂️

spring-kafka/src/test/java/org/springframework/kafka/listener/DeliveryAttemptAwareRetryListenerIntegrationTests.java
```diff
@@ -0,0 +1,308 @@
...
+
+			ConcurrentKafkaListenerContainerFactory<String, String> factory = new ConcurrentKafkaListenerContainerFactory<>();
+			factory.setConsumerFactory(consumerFactory);
+			factory.setCommonErrorHandler(errorHandler);
+
+
+			final ContainerProperties containerProperties = factory.getContainerProperties();
+			containerProperties.setDeliveryAttemptHeader(true);
+
+			return factory;
```
artembilan, line 302: Apparently a common code for both these two bean could be extracted to `private` method for better readability.

artembilan (changes requested):

chickenchickenlove:
@artembilan thanks for your time for looking into changes, again! 🙇‍♂️ 
I made a new commit to apply your comments.
When you have time, please take another look! 

Always thanks to read my code 🙇‍♂️

spring-kafka/src/test/java/org/springframework/kafka/listener/DeliveryAttemptAwareRetryListenerIntegrationTests.java
```diff
@@ -0,0 +1,280 @@
...
+
+		@Autowired
+		EmbeddedKafkaBroker broker;
+
+		@Bean
+		ProducerFactory<String, String> producerFactory() {
+			Map<String, Object> props = KafkaTestUtils.producerProps(
+					this.broker.getBrokersAsString());
+			props.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class);
+			props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, StringSerializer.class);
```
artembilan, line 233: The `value.serializer` is a `StringSerializer` by default in the `KafkaTestUtils.producerProps()`.
And why do we really need a `StringSerializer` for key in these tests?
Why cannot we just rely on the defaults from the `KafkaTestUtils.producerProps()`?

spring-kafka/src/test/java/org/springframework/kafka/listener/DeliveryAttemptAwareRetryListenerIntegrationTests.java
```diff
@@ -0,0 +1,308 @@
...
+					"groupId");
+			props.put(
+					ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG,
+					StringDeserializer.class);
+			props.put(
+					ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG,
+					StringDeserializer.class);
+			props.put(
+					ConsumerConfig.ALLOW_AUTO_CREATE_TOPICS_CONFIG, false);
+			props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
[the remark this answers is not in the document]
artembilan: Something similar for `KafkaTestUtils.consumerProps()`.
Why we just cannot rely on the defaults from there?
And why do we really need that `ConsumerConfig.ALLOW_AUTO_CREATE_TOPICS_CONFIG`?

chickenchickenlove:
@artembilan , It's my bad! 
I cleaned up the code that I thought was useless.
When you have time, please take another look 🙇‍♂️

artembilan:
Thank you, @chickenchickenlove , again for contribution!

Please, take a look into this article about commit messages: https://cbea.ms/git-commit/.
Also take into account that first commit message becomes PR description automatically. 
Just for a sake to save some yours and ours time:
- first - no need to write separate PR description
- second - no need to write extra commit message on merge
- and last - the commit message is exactly what we always need to be good. The PR easily might be lost, but commit is there forever.

Now imaging time is passing and we don't have contacts with you to remind us whatever was done in that commit years ago.
but if you write a good commit message, it will be merged there into a history with respective honor outcome.

chickenchickenlove:
Thanks for your time 🙇‍♂️ 
As you mentioned, I will try to write good commit message 👍

PR #3551, opened 2024-10-09
No Issue : Refactor test codes and Remove useless full package name.

chickenchickenlove:
#### 1. Remove useless full package name.
In the `MessagingMessageListenerAdapter` class, `org.springframework.messaging.converter.MessageConversionException` is already imported. however, the `MessageConversionException` was declared with full package names.

#### 2. Refactor test codes
Many test classes don't use `KafkaTestUtils.consumerProps()` and `KafkaTestUtils.producerProps()` even if they can. 
Also, test code for `KafkaStreams` can be integrated using `KafkaTestUtils.streamsProps()`.

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2022 the original author or authors.
+ * Copyright 2016-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -45,6 +45,7 @@ import org.apache.kafka.common.serialization.IntegerSerializer;
 import org.apache.kafka.common.serialization.StringDeserializer;
 import org.apache.kafka.common.serialization.StringSerializer;
 
+import org.apache.kafka.streams.StreamsConfig;
 import org.springframework.beans.DirectFieldAccessor;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.kafka.test.EmbeddedKafkaBroker;
@@ -57,6 +58,7 @@ import org.springframework.util.Assert;
  * @author Gary Russell
  * @author Hugo Wood
  * @author Artem Bilan
+ * @author Sanghyeok An
  */
 public final class KafkaTestUtils {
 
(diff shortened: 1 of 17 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
```diff
@@ -82,6 +84,16 @@ public static Map<String, Object> consumerProps(String group, String autoCommit,
 		return consumerProps(embeddedKafka.getBrokersAsString(), group, autoCommit);
 	}
 
+	/**
+	 * Set up test properties for an {@code <Integer, String>} consumer.
+	 * @param brokers the bootstrapServers property.
+	 * @param group the group id.
+	 * @return the properties.
```
sobychacko, line 91: You need to add the `since` tag in the new API method.

sobychacko (changes requested):
See the comment on adding the `since` tag.

spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
sobychacko, line 91: Need the `since` tag here since it is a new API method.

sobychacko (changes requested):

spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
```diff
@@ -82,6 +84,16 @@ public static Map<String, Object> consumerProps(String group, String autoCommit,
 		return consumerProps(embeddedKafka.getBrokersAsString(), group, autoCommit);
 	}
 
+	/**
+	 * Set up test properties for an {@code <Integer, String>} consumer.
```
artembilan, line 88: I understand that it is a copy/paste, but what is that `@code` about?
Can we just take closer look to see what exactly Gary has meant with these JavaDocs?

artembilan (changes requested):

spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
chickenchickenlove: Thanks for your comments!
It's my bad. 
I fixed JavaDocs for `streamsProps()`, because it does not consider which Deserializer use to!
However, I think `public static Map<String, Object> consumerProps(String brokers, String group)` has a proper description, because it invokes `public static Map<String, Object> consumerProps(String brokers, String group, String autoCommit)`.

What do you think?

artembilan: Yeah... I see now. It talks about exactly what (de)serializers we configure over there for `Producer/Consumer` respectively.

Issue #3589, opened 2024-10-25
Duplicate header processing in `BatchMessagingMessageConverter#processRecord()`


borahm-lee:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

For example:

- 3.3.0-SNAPSHOT

**Describe the bug**

- Originally, logs were supposed to be left once per record list when there is no `headerMapper`, but now they are left as many times as there are records. 
- Similarly, putting `listenerInfo` on `rawHeaders` is being attempted as many times as there are records.

**To Reproduce**

- Run `BatchMessageConverterTests#testNoMapper()`.

**Expected behavior**

- It only logs once when there is no `headerMapper`.
- Putting `listenerInfo` to `rawHeaders` is also attempted only once.

**Sample**

> Currently, the logs in the code are at the `DEBUG` level, but I changed them to `WARN` for ease of reading.
- Test before fixing:
    ![image](https://github.com/user-attachments/assets/dc372c0c-a2e6-4ed9-b484-0cd8a60600f4)

- Test after fixing:
     ![image](https://github.com/user-attachments/assets/f044a6b2-9925-445c-ba9a-8276d1746b4a)

I'm opening a PR right away to address this issue.

sobychacko:
PR merged.

PR #3590, opened 2024-10-25
GH-3589: Refactor `toMessage` batch method and fix logging/header issues

bky373:
## Description

- Fixes: #3589
- Remove duplicate logging and listenerInfo operation that occurs for each record in `processRecord()`.
- And refactor the tasks in `toMessage()` as follows:
    - `addRecordInfo()`: Adds information such as `type`, `payload`, `key`, etc. from the `record` to the given lists.
    - `addToConvertedHeaders()`: Adds the converted value of `record.headers()` to `convertedHeaders`.
    - Extract the value of the `KafkaHeaders.LISTENER_INFO` key from `convertedHeaders`, and put it to `rawHeaders` after the loop ends.
    - If there's no `headerMapper`, log a message so that users can refer to `KafkaHeaders.NATIVE_HEADERS`. As there's no need to check `boolean logged` every time in the loop, the `logged` variable has been removed, and the check for the existence of `headerMapper` and `natives` is now performed only once after the loop ends.

> [!NOTE]
> Also, I changed the logging level from `DEBUG` to `WARN` when there is no header mapper. `INFO` might be better, please let me know what you think. -> Sorry, I'll keep `DEBUG` because it's logged every time the listener poll.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
@@ -185,7 +185,7 @@ public class BatchMessagingMessageConverter implements BatchMessageConverter {
 			}
 		}
 		if (this.headerMapper == null) {
-			this.logger.warn(() ->
+			this.logger.debug(() ->
 					"No header mapper is available; Jackson is required for the default mapper; "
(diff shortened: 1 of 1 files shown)


sobychacko:
Merged via https://github.com/spring-projects/spring-kafka/commit/5fb1562d95b4b9ff816e56fee49d99293d5a3acb.

Issue #3557, opened 2024-10-14
Lower the replication factor for transactions topic when using embedded Kafka broker for tests


dlyash:
**Expected Behavior**

It would be nice if the embedded broker started by `@EmbeddedKafka` (or the similar JUnit Rule) defaulted the replication factor for the transaction state topic to the number of brokers, or a `min(<number-of-brokers>, 3)`. 

**Current Behavior**
The embedded broker starts with the default configuration of `transaction.state.log.replication.factor = 3` which doesn't make sense for a single-broker cluster, which is quite common in tests:
```
org.apache.kafka.common.errors.InvalidReplicationFactorException: Replication factor: 3 larger than available brokers: 1.
```

**Context**
It's not a big deal, as it can be resolved by explicitly setting broker properties like follows. But a meaningful default would save some time for thousands of developers googling the problem again and again.
```
@EmbeddedKafka(brokerProperties={
    "transaction.state.log.replication.factor=1"
})
```

sobychacko:
@dlyash The tests we have in the framework for transactions that use `EmbeddedKafka' do exactly that - i.e., setting that property `transaction.state.log.replication.factor` to 1. I wonder whether we should document that and not make any code changes primarily because retrieving the number of brokers consistently might be challenging. What do you think?

sobychacko:
Ok, you can give it a try and see if that works and is stable. If so, feel free to submit a PR. Thanks!

Dltmd202:
I created fix PR https://github.com/spring-projects/spring-kafka/pull/3602~!

PR #3602, opened 2024-10-27
Lower the replication factor for transactions topic on @EmbeddedKafka Test

Dltmd202:
Fixes: #3557 

Expected Behavior

It would be nice if the embedded broker started by @EmbeddedKafka (or the similar JUnit Rule) defaulted the replication factor for the transaction state topic to the number of brokers, or a min(<number-of-brokers>, 3).

Current Behavior
The embedded broker starts with the default configuration of transaction.state.log.replication.factor = 3 which doesn't make sense for a single-broker cluster, which is quite common in tests:

```
org.apache.kafka.common.errors.InvalidReplicationFactorException: Replication factor: 3 larger than available brokers: 1.
```

Fix

* When writing test code using EmbeddedKafka, adjust the number of brokers for the transaction topic to the minimum.
* The topic has been made constant

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafkaContextCustomizer.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafkaContextCustomizer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2023 the original author or authors.
+ * Copyright 2017-2024 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


sobychacko:
@Dltmd202 There are some checkstyle errors. Can you fix that? In addition to that, could you add your name as the author to all the classes you changed? Thanks!

Dltmd202:
Of course. I will reflect it right away! Thank you

sobychacko:
Thanks for the PR! It is now merged upstream.

PR #3642, opened 2024-11-19
GH-2128: Update Javadoc for nack method to correct maxPollInterval to pollTimeout

sebastto:
Correct the Javadoc to specify `pollTimeout` instead of `maxPollInterval` for the `nack()` methods.

See https://github.com/spring-projects/spring-kafka/issues/2128


sebastto:
Force push to adhere to 50 char limit on first line of commit message

PR #3643, opened 2024-11-22
Allow configuring observation registry directly

cfredri4:
This changes allows configuring the observation registry directly, instead of it being fetched from the application context. This is to allow observability when KafkaTemplate/KafkaMessageListenerContainer are used without an application context.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -153,7 +153,7 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V>, ApplicationCo
 
 	private KafkaTemplateObservationConvention observationConvention;
 
-	private ObservationRegistry observationRegistry = ObservationRegistry.NOOP;
+	private ObservationRegistry observationRegistry;
 
 	@Nullable
 	private Function<ProducerRecord<?, ?>, Map<String, String>> micrometerTagsProvider;
@@ -457,6 +457,15 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V>, ApplicationCo
 	}
 
 	/**
+	 * Configure the {@link ObservationRegistry} to use for recording observations.
+	 * @param observationRegistry the observation registry to use.
+	 * @since 3.3
+	 */
+	public void setObservationRegistry(ObservationRegistry observationRegistry) {
+		this.observationRegistry = observationRegistry;
+	}
+
+	/**
 	 * Return the {@link KafkaAdmin}, used to find the cluster id for observation, if
 	 * present.
 	 * @return the kafkaAdmin
@@ -478,7 +487,7 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V>, ApplicationCo
 
 	@Override
 	public void afterSingletonsInstantiated() {
-		if (this.observationEnabled && this.applicationContext != null) {
+		if (this.observationRegistry == null && this.observationEnabled && this.applicationContext != null) {
 			this.observationRegistry = this.applicationContext.getBeanProvider(ObservationRegistry.class)
 					.getIfUnique(() -> this.observationRegistry);
 			if (this.kafkaAdmin == null) {
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -372,14 +372,16 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 		}
 		GenericMessageListener<?> listener = (GenericMessageListener<?>) messageListener;
 		ListenerType listenerType = determineListenerType(listener);
-		ObservationRegistry observationRegistry = ObservationRegistry.NOOP;
-		ApplicationContext applicationContext = getApplicationContext();
-		if (applicationContext != null && containerProperties.isObservationEnabled()) {
-			ObjectProvider<ObservationRegistry> registry =
-					applicationContext.getBeanProvider(ObservationRegistry.class);
-			ObservationRegistry reg = registry.getIfUnique();
-			if (reg != null) {
-				observationRegistry = reg;
+		ObservationRegistry observationRegistry = containerProperties.getObservationRegistry();
+		if (observationRegistry == null) {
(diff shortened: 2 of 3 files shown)


sobychacko:
@ cfredri4 I think this needs more discussion. These changes modify the behavior of setting the default `NOOP` observation registry for both `KafkaTemplate` and the listener container. Why do you need to provide the `ObservationRegistry` as a non-bean, and why is the application context absent? Since there are some design changes in the PR, we will discuss this internally soon, but you can just give some more context around the changes. Thanks.

cfredri4:
> @cfredri4 I think this needs more discussion. These changes modify the behavior of setting the default NOOP observation registry for both KafkaTemplate and the listener container.

This was for simplicity's sake as the [registry is nullable and null and NOOP are treated the same](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java#L168). But I can change back to NOOP if desired.

> Why do you need to provide the ObservationRegistry as a non-bean, and why is the application context absent? Since there are some design changes in the PR, we will discuss this internally soon, but you can just give some more context around the changes. Thanks.

I have two situations where this change would be beneficial:

1. We sometimes use Spring Kafka in applications which are not Spring applications, i.e. they do not have a Spring application context. Here we need to create a mock `ApplicationContext` which implements `getBeanProvider` to return the registry; this is messy and fragile as it's dependent on internal logic in how KafkaTemplate/KafkaMessageListenerContainer works.

2. We sometimes have the need to use a different observation registry for Spring Kafka than what is in the application context. Here we need to do similar to # 1 above but then also make sure that this is done after Spring has already initialized KafkaTemplate/KafkaMessageListenerContainer as this time they _are_ Spring beans.

Both of these situations would be trivial had it been possible to configure the observation registry directly.
Exactly the same thing can already be done for the `KafkaAdmin`, so to me it makes sense to allow it also for the observation registry.

sobychacko:
@cfredri4 We are ok with the changes. However, could you re-work your solution so we don't break any existing behavior? Specifically, the default `ObservationRegistry.NOOP` should be restored in both `KafkaTemplate` and `KafkaMessageListenerContainer`. Thanks!

sobychacko:
Also, please add the author tag with your name in all the classes you changed. Thanks!

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -456,6 +456,15 @@ public void setObservationConvention(KafkaTemplateObservationConvention observat
 		this.observationConvention = observationConvention;
 	}
 
+	/**
+	 * Configure the {@link ObservationRegistry} to use for recording observations.
+	 * @param observationRegistry the observation registry to use.
+	 * @since 3.3
```
artembilan, line 462: `3.3.1`

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -456,6 +456,15 @@ public void setObservationConvention(KafkaTemplateObservationConvention observat
...
+	 */
+	public void setObservationRegistry(ObservationRegistry observationRegistry) {
+		this.observationRegistry = observationRegistry;
```
artembilan, line 465: `Assert.notNull(observationRegistry, "'observationRegistry' must not be null");`

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -372,14 +372,16 @@ protected void doStart() {
...
+		if (observationRegistry == null) {
+			ApplicationContext applicationContext = getApplicationContext();
+			if (applicationContext != null && containerProperties.isObservationEnabled()) {
+				ObjectProvider<ObservationRegistry> registry =
+						applicationContext.getBeanProvider(ObservationRegistry.class);
+				ObservationRegistry reg = registry.getIfUnique();
+				if (reg != null) {
+					observationRegistry = reg;
+				}
 			}
```
artembilan, line 385: `else {
    observationRegistry = ObservationRegistry.NOOP;
}`

artembilan (changes requested):
Please, consider to address reviews by tomorrow.
We have release on Monday, so might be great to have this in.

Thank you!

cfredri4:
I have pushed some updates, please check.

artembilan (approved):
LGTM

artembilan (changes requested):
Well, sorry for the noise, but apparently there are some Checkstyle violations:
```
Error: eckstyle] [ERROR] /home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java:496: Line matches the illegal pattern 'Trailing whitespace'. [Regexp]

Error: eckstyle] [ERROR] /home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java:45:1: Wrong order for 'io.micrometer.observation.ObservationRegistry' import. [ImportOrder]
> Task :spring-kafka:checkstyleMain
Error: eckstyle] [ERROR] /home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java:81:8: Unused import - org.springframework.beans.factory.ObjectProvider. [UnusedImports]
```

Please, consider to use Gradle `check` task before pushing commits to the PR.

Thanks

cfredri4:
Fixed.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -372,14 +372,15 @@ protected void doStart() {
...
-				observationRegistry = reg;
+		ObservationRegistry observationRegistry = containerProperties.getObservationRegistry();
+		if (observationRegistry.isNoop()) {
+			ApplicationContext applicationContext = getApplicationContext();
+			if (applicationContext != null && containerProperties.isObservationEnabled()) {
+				ObservationRegistry reg = applicationContext.getBeanProvider(ObservationRegistry.class)
+						.getIfUnique();
+				if (reg != null) {
+					observationRegistry = reg;
+				}
```
artembilan, line 383: Why don't you use similar to `KafkaTemplate` approach here instead:
```
.getIfUnique(() -> observationRegistry);
```
?

PR #3652, opened 2024-11-26
Change "synchronized" to reentrant lock for virtual-threads

omercelikceng:
This commit ensures that the method is friendly for virtual threads to avoid blocking and pinning. The lock is acquired at the beginning of the method and released in a finally block to ensure it is always released, even if an exception occurs.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
@@ -32,6 +32,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
@@ -78,6 +80,7 @@ import org.springframework.util.Assert;
  * @author Sanghyeok An
  * @author Valentina Armenise
  * @author Anders Swanson
+ * @author Omer Celik
  *
  * @since 1.3
  */
@@ -94,6 +97,8 @@ public class KafkaAdmin extends KafkaResourceFactory
 	private static final LogAccessor LOGGER = new LogAccessor(LogFactory.getLog(KafkaAdmin.class));
 
 	private static final AtomicInteger CLIENT_ID_COUNTER = new AtomicInteger();
+
+	private final Lock clusterIdLock = new ReentrantLock();
 
 	private final Map<String, Object> configs;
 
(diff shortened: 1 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -267,12 +272,16 @@ public final boolean initialize() {
 			}
 			if (adminClient != null) {
 				try {
-					synchronized (this) {
+					try {
```
sobychacko, line 275: Can we use only one `try` block? I think, you can move the unlocking to the `finally` block on the main `try`.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/JsonSerializer.java
```diff
@@ -147,30 +152,36 @@ public void setUseTypeMapperForKey(boolean isKey) {
...
-		}
-		Assert.state(!this.setterCalled
-				|| (!configs.containsKey(ADD_TYPE_INFO_HEADERS) && !configs.containsKey(TYPE_MAPPINGS)),
-				"JsonSerializer must be configured with property setters, or via configuration properties; not both");
-		setUseTypeMapperForKey(isKey);
-		if (configs.containsKey(ADD_TYPE_INFO_HEADERS)) {
-			Object config = configs.get(ADD_TYPE_INFO_HEADERS);
-			if (config instanceof Boolean configBoolean) {
-				this.addTypeInfo = configBoolean;
+		try {
```
sobychacko, line 155: Forgot to remove the `synchronized` keyword from the method signature?

sobychacko (changes requested):
See my in-line comments.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
omercelikceng: Hi @sobychacko .
If I do what you suggest, I will end up locking the method, including the addOrModifyTopicsIfNeeded method. I reviewed the addOrModifyTopicsIfNeeded method, and I think it contains many operations. In my opinion, this method should not be included within the lock.

sobychacko: That makes sense. I guess you can try to extract that inner try block into a separate private method - something like `updateClusterId(...)` and then call that in the current method. That way, we can avoid stacking one `try` block on another and thus look less cluttered.

omercelikceng: You're right, I'll fix it right away.

omercelikceng:
> See my in-line comments.

@sobychacko I've made the updates; could you review them again?

sobychacko:
@omercelikceng Thanks for the PR. Merged if via https://github.com/spring-projects/spring-kafka/commit/3c97a3bce1ed8141f7b5ff8390c0a68f6f1e232b.

Issue #3660, opened 2024-12-08
EmbeddedKafkaBroker consume with seekToEnd does not reliably seek to the end


coekie:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.3.0

**Describe the bug**

Calling `EmbeddedKafkaBroker#consumeFromAnEmbeddedTopic` with `seekToEnd` set to `true` may cause the consumer to miss messages that were sent to the topic _after_ calling that method, because the seeking is done asynchronously.
That seems unexpected and leads to flaky, unpredictable test results.

That is because `consumeFromEmbeddedTopics` calls the kafka `Consumer#seekToEnd` method which is documented as "This function evaluates lazily, seeking to the final offset in all partitions only when poll(Duration) or position(TopicPartition) are called.".

**To Reproduce**

Call `consumeFromAnEmbeddedTopic` with `seekToEnd`, then send a message to the topic, and then try to consume it. If the consumer sees the message or not depends on timing.

Concrete use case: Calling consumeFromAnEmbeddedTopic at the start of a test, then calling application code that is expected to send the message, and then using `KafkaTestUtils.getSingleRecord` in the test may randomly fail.
Workarounds: Calling `Thread.sleep` or `Consumer.position` after calling `consumeFromAnEmbeddedTopic`.

**Expected behavior**

The method should seek to the end before returning.

**Sample**

I will submit a PR including a failing test case.

PR #3661, opened 2024-12-08
GH-3660: Fix EmbeddedKafkaBroker seekToEnd

coekie:
Fixes #GH-3660

> Problem:
> consumeFromEmbeddedTopics calls Consumer.seekToEnd, which "evaluates lazily, seeking to the final offset in all partitions only when poll(Duration) or position(TopicPartition) are called". This means that the consumer may or may not see future messages, depending on timing.
> 
> This fix calls `position` so that the seek happens before consumeFromEmbeddedTopics returns. That was it is ensured that any message sent to the topic after the call to consumeFromEmbeddedTopics are seen by the consumer.

I did not see tests yet for the consume methods in EmbeddedKafkaBroker, so I'm not entirely sure if this is the approach you would want for the tests. The tests are inspired by tests in KafkaTestUtilsTests.

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
@@ -71,6 +71,7 @@ import org.springframework.util.Assert;
  * @author Adrian Chlebosz
  * @author Soby Chacko
  * @author Sanghyeok An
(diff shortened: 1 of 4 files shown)


sobychacko:
@coekie Changes look good. Can you add your name as an author to all the classes you changed?

coekie:
@sobychacko Great. I just added myself as author (and tweaked the commit/PR title)

sobychacko:
Thanks for the PR! Merged upstream.

Issue #3662, opened 2024-12-09
Allow kafka-clients 3.9.x with Spring-Kafka 3.3.x


bgK:
The test artifacts of `kafka-clients` versions 3.7.0 to 3.8.1 include a `junit-platform.properties` file. JUnit only supports a single `junit-platform.properties`. As a result, the file provided by `kafka-clients` can override the application provided file depending on classpath order. This has been fixed in 3.9.0 with [KAFKA-17121](https://issues.apache.org/jira/browse/KAFKA-17121).

In our application the `junit-platform.properties` file enables an extention for creating a test database.
When upgrading from Spring-Kafka 3.1.x to 3.3.x, `kafka-clients` was updated and the extension was no longer executed.
As a workaround, we downgraded `kafka-clients` to 3.6.2. This seems to work even though it's not documented as supported.

It would be nice if Spring-Kafka 3.3.x could support `kafka-clients` version 3.9.0 as a proper fix.

sobychacko:
@bgK Did you try upgrading your application to use `3.9.0` kafka client? We ran into some other issues with `3.9.0`, that we will fix, but generally speaking you should be able to update to `3.9.0` on the application side.

bgK:
@sobychacko I'm getting the following error with `kafka-clients` 3.9.0 when starting the embedded Kafka broker:
```
java.lang.NoSuchMethodError: 'kafka.testkit.KafkaClusterTestKit$Builder kafka.testkit.KafkaClusterTestKit$Builder.setConfigProp(java.lang.String, java.lang.String)'

	at org.springframework.kafka.test.EmbeddedKafkaKraftBroker.lambda$start$0(EmbeddedKafkaKraftBroker.java:219)
	at java.base/java.util.concurrent.ConcurrentHashMap.forEach(ConcurrentHashMap.java:1605)
	at java.base/java.util.Properties.forEach(Properties.java:1467)
	at org.springframework.kafka.test.EmbeddedKafkaKraftBroker.start(EmbeddedKafkaKraftBroker.java:219)
	at org.springframework.kafka.test.EmbeddedKafkaKraftBroker.afterPropertiesSet(EmbeddedKafkaKraftBroker.java:204)
	at org.springframework.kafka.test.context.EmbeddedKafkaContextCustomizer.customizeContext(EmbeddedKafkaContextCustomizer.java:135)
	at org.springframework.test.context.support.AbstractContextLoader.customizeContext(AbstractContextLoader.java:189)
[...]

PR #3665, opened 2024-12-12
GH-3662: Support Kafka 3.9.0+ in EmbeddedKafkaKraftBroker

sobychacko:
Fixes: #3662
Issue: https://github.com/spring-projects/spring-kafka/issues/3662

Add compatibility for both Kafka 3.8.0 and 3.9.0+ by handling different method signatures for setConfigProp:
- 3.9.0+: setConfigProp(String, Object)
- 3.8.0: setConfigProp(String, String)

The change uses reflection to detect Kafka version and call appropriate method.

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
@@ -18,7 +18,6 @@ package org.springframework.kafka.test;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.nio.file.Files;
 import java.time.Duration;
@@ -58,6 +57,8 @@ import org.apache.kafka.common.utils.Utils;
 
 import org.springframework.core.log.LogAccessor;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+import org.springframework.util.ReflectionUtils;
 
 /**
  * An embedded Kafka Broker(s) using KRaft.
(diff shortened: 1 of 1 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
```diff
@@ -243,6 +247,37 @@ private void start() {
...
 
+	private static void setConfigProperty(Object clusterBuilder, Object key, Object value) {
+		try {
+			boolean isKafka39OrLater = isClassicConsumerPresent();
+			Class<?> builderClass = clusterBuilder.getClass();
+
+			if (isKafka39OrLater) {
+				// For Kafka 3.9.0+: setConfigProp(String, Object)
+				Method setConfigMethod = builderClass.getMethod("setConfigProp", String.class, Object.class);
+				setConfigMethod.invoke(clusterBuilder, (String) key, value);
```
artembilan, line 258: Also see `ReflectionUtils.findMethod()` and `ReflectionUtils.invokeMethod()`.
You won't need to deal with any exceptions then!

spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
```diff
@@ -87,6 +92,24 @@ public class EmbeddedKafkaKraftBroker implements EmbeddedKafkaBroker {
 
 	public static final int DEFAULT_ADMIN_TIMEOUT = 10;
 
+	private static final boolean IS_KAFKA_39_OR_LATER;
+
+	private static final Method SET_CONFIG_METHOD;
+
+	static {
+		IS_KAFKA_39_OR_LATER = ClassUtils.isPresent(CLASS_EXISTS_ONLY_IN_390, EmbeddedKafkaKraftBroker.class.getClassLoader());
```
artembilan, line 100: `isPresent()` doesn't throw an exception: can be used directly in the property declaration line.
Also: I don't see a reason in the `CLASS_EXISTS_ONLY_IN_390` constant. You simply can use that class name in this method call.

spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaKraftBroker.java
```diff
@@ -243,6 +266,17 @@ private void start() {
 		System.setProperty(SPRING_EMBEDDED_KAFKA_BROKERS, getBrokersAsString());
 	}
 
+	private static void setConfigProperty(KafkaClusterTestKit.Builder clusterBuilder, Object key, Object value) {
```
artembilan, line 269: Nit-pick: can we cast `key` into a `String` where we call this method?
That way that cast operation would happen only in one place.

artembilan (changes requested):

Issue #3697, opened 2024-12-31
Optimize a `MessagingMessageListenerAdapter` returning `null` from the `DelegatingInvocableHandler.invoke()`


artembilan:
Right now the  `DelegatingInvocableHandler.invoke()` always returns an `InvocationResult` even if `result == null`.
In the end we may just have a log message like:
```
Async result is null, ignoring
```
which is not always true for the regular `void @KafkaHandler` method.

Changing the logic in the mention method to:
```
		if (result != null) {
			return new InvocationResult(result, this.handlerSendTo.get(handler),
					this.handlerReturnsMessage.get(handler));
		}
		else {
			return null;
		}
```
would make some optimization where we would not try to resolve any reply attempts and creation of many interim objects like `CompletableFuture`.

Since this would be a breaking change, the issue is scheduled for the next `4.0` major version.

ivamly:
Hi, @artembilan!

Could I work on this issue?

Thanks!

sobychacko:
@ivamly Yes, please go ahead.

sobychacko:
See the comment on the latest PR.

ivamly:
I'll make another PR soon, no problem.

PR #3710, opened 2025-01-16
GH-3697: DelegatingInvocableHandler.invoke() returning null

ivamly:
Fixes: #3697 

Change the logic in the  DelegatingInvocableHandler.invoke()


ivamly:
Thank for the help! I'll do it

sobychacko:
@ivamly Even though it was merged, I realized that this change needs to wait until we break into the new version (`4.0.x`). We cannot put this in the current versions since it is a breaking change. Sorry about the confusion. @artembilan made that clear in the original description. I will reopen your PR, and then we can tackle this once we branch off into `4.0.x`.

sobychacko:
Actually, i cannot reopen this PR since it was already merged, and we had to back it out. Can you issue another PR based on your latest changes? You can also wait until we branch into the `4.0.x` version. Thanks!

Issue #3726, opened 2025-01-29
KafkaMessageListenerContainer.getAssignedPartitions is not safe


tbarabanov:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.2.1

**Describe the bug**

`KafkaMessageListenerContainer.getAssignedPartitions()` is not safe due to the fact that different threads can iterate/modify any of the fields `partitionsListenerConsumer.definedPartitions`, `partitionsListenerConsumer.assignedPartitions` simultaneously, but collection types of these fields are not designed for such scenarios. Thus at least ConcurrentModificationException can be thrown.

**Expected behavior**

`KafkaMessageListenerContainer.getAssignedPartitions()` method is public and thus expected to be safe.

artembilan:
More info in StackOverflow question: https://stackoverflow.com/questions/79383650/is-kafkamessagelistenercontainer-getassignedpartitions-thread-safe.

@tbarabanov ,

feel free to contribute the fix: https://github.com/spring-projects/spring-kafka/blob/main/CONTRIBUTING.adoc!

PR #3728, opened 2025-01-30
GH-3726: Avoid ConcurrentModificationException in KafkaMessageListenerContainer

tbarabanov:
Fixes possible `ConcurrentModificationException` in `KafkaMessageListenerContainer` that can happen in case if any of `partitionsListenerConsumer.definedPartitions` or `partitionsListenerConsumer.assignedPartitions` fields are being iterated/modified simultaneously.

For instance by consumer thread performing rebalancing and arbitrary application thread calling `getAssignedPartitions` method in `KafkaMessageListenerContainer`.

To achieve desired effect I'm using synchronized wrappers. I would prefer modern concurrent collections such as `ConcurrentHashMap`, but `definedPartitions` field has type of `LinkedHashMap` and its `keySet` preserves keys order. 
Thus 1) no thread safe data structures available from std lib with same characteristics 2) possibly some users rely on the order. Anyway do not think that there are real life cases that can put `definedPartitions` or `assignedPartitions` under pressure.

Fixes GH-3726 (https://github.com/spring-projects/spring-kafka/issues/3726)


artembilan (changes requested):
Please, add your name to the `@author` list of that changed class - and I'll merge it shortly.

Thank you!

tbarabanov:
Done. Thx!

artembilan:
Merged and back-ported to `3.2.x`.

Issue #3734, opened 2025-02-10
Configure Gradle toolchain auto-provisioning in build


suneetk92:
**Expected Behavior**

Gradle provides a [Auto-provisioning](https://docs.gradle.org/8.12/userguide/toolchains.html#sec:provisioning) mode where it can automatically download the toolchain that matches the requirements of the build.

**Current Behavior**

User have to manually install the required JDK to sucessfully build the project.

**Context**

Similar to [spring-framework](https://github.com/spring-projects/spring-framework/commit/8708a21143cf345726f1f22ae5f991a3fccfa74e)

PR #3735, opened 2025-02-10
GH-3734: Configure Gradle toolchain auto-provisioning

suneetk92:
Fixes: https://github.com/spring-projects/spring-kafka/issues/3734

This commit configures the Foojay resolver plugin in the Gradle build to auto-provision a required JDK distribution if it's not present already on the host.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -7,6 +7,7 @@ pluginManagement {
 
 plugins {
 	id 'io.spring.develocity.conventions' version '0.0.22'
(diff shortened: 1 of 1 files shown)


sobychacko:
@suneetk92 Thanks for the PR. We are trying to see how to proceed with this from a framework standpoint since we might want to be consistent across the portfolio. Once that is clear, we can review and merge.

sobychacko:
Thanks @suneetk92. The PR is now merged upstream.

PR #3736, opened 2025-02-10
Fix typo in `CONTRIBUTING` and `README`

ngocnhan-tran1996:

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -113,10 +113,10 @@ From the command line you can check this using:
 log --graph --pretty=oneline
 ----
 
-As this may cause lots of typing, we recommend creating a global alias, e.g. `git log` for this:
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.adoc
```diff
@@ -113,10 +113,10 @@ From the command line you can check this using:
 log --graph --pretty=oneline
 ----
 
-As this may cause lots of typing, we recommend creating a global alias, e.g. `git logg` for this:
+As this may cause lots of typing, we recommend creating a global alias, e.g. `git log` for this:
 
 ----
-git config --global alias.logg 'log --graph --pretty=oneline'
+git config --global alias.log 'log --graph --pretty=oneline'
```
artembilan, line 119: I don't think this is a correct change.
There is already `git log`, and this section really talks about a custom one to include those options.
Therefore we cannot call a new alias as `log` and have to chose other name.

artembilan (changes requested):

Issue #3741, opened 2025-02-13
Metric collection for errors always report the same error "ListenerExecutionFailedException" regardless of what's thrown in the consumer method


renannprado:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.3.2

**Describe the bug**

I've followed [this part](https://docs.spring.io/spring-kafka/reference/kafka/micrometer.html#observation) of the documentation to integrate micrometer in the Kafka listener's container:

```java
@Bean
KafkaListenerContainerFactory<ConcurrentMessageListenerContainer<String, MyPayload>> myPayloadKafkaListenerContainerFactory(
    ConsumerFactory<String, MyPayload> MyPayloadKafkaConsumerFactory,
    ObservationRegistry observationRegistry
) {
    ConcurrentKafkaListenerContainerFactory<String, MyPayload> factory = new ConcurrentKafkaListenerContainerFactory<>();
    factory.setConsumerFactory(myPayloadKafkaConsumerFactory);
    factory.setConcurrency(1);
    // the relevant part
    factory.getContainerProperties().setObservationEnabled(true);
    factory.getContainerProperties().setObservationRegistry(observationRegistry);

    return factory;
}
````

Then after throwing an exception from the consumer:

```java
@RetryableTopic(
    kafkaTemplate = "myProcessRecoverKafkaTemplate",
    attempts = "3",
    backoff = @Backoff(
        delay = 2000, maxDelay = 10000, multiplier = 5
    )
)
@KafkaListener(
    id = "processmyProcess",
    topics = "my-process",
    containerFactory = "myProcessKafkaListenerContainerFactory"
)
public void processPayload(
    ConsumerRecord<String, MyPayload> data
) {
    if (1 == 1) {
        throw new CustomException();
    }
}
```

This is what I see in the metric:

```
spring_kafka_listener_seconds_count{error="ListenerExecutionFailedException",messaging_kafka_consumer_group="processMethod-retry-10000",messaging_operation="receive",messaging_source_kind="topic",messaging_source_name="my-method-retry-10000",messaging_system="kafka",spring_kafka_listener_id="processMethod-retry-10000-0"} 6

```

Without configuring micrometer observation:

```
spring_kafka_listener_seconds_sum{exception="ListenerExecutionFailedException",name="my-process-retry-4000-0",result="failure"} 0.02504
```

I tracked to this piece of code:

[...]

PR #3756, opened 2025-02-16
GH-3741: Fix metric tag to show underlying exception type

sobychacko:
Fixes: #3741
Issue: https://github.com/spring-projects/spring-kafka/issues/3741

When exceptions occur in Kafka listeners, the metrics currently show `ListenerExecutionFailedException` in both the `error` tag (when using observation) and `exception` tag (when using micrometer without observation), rather than the actual underlying exception.

* Modify ListenerContainer to pass actual exception to failure metrics
* Update MessagingMessageListenerAdapter to report cause to observation
* Add MicrometerMetricsTests to verify both observation and non-observation metrics
* Fix ObservationTests to verify correct error reporting in metrics

This ensures metrics show the actual underlying exception while maintaining existing span behavior.

Signed-off-by: Soby Chacko <soby.chacko@broadcom.com>

**Auto-cherry-pick to `3.3.x` & `3.2.x`**

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
@@ -16,7 +16,6 @@
 
 package org.springframework.kafka.support.micrometer;
 
-import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
@@ -25,8 +24,6 @@ import io.micrometer.core.instrument.Timer;
 import io.micrometer.core.instrument.observation.DefaultMeterObservationHandler;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import io.micrometer.observation.ObservationRegistry;
-import io.micrometer.tracing.test.simple.SimpleTracer;
-import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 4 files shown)


spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
```diff
@@ -0,0 +1,210 @@
...
+import org.springframework.kafka.test.context.EmbeddedKafka;
+import org.springframework.kafka.test.utils.KafkaTestUtils;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Soby Chacko
+ * @since 3.3.3
```
artembilan, line 55:
```suggestion
 * @since 3.2.7
```

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
```diff
@@ -0,0 +1,210 @@
...
+
+		assertThat(timer).isNotNull();
+	}
+
+	@Configuration
+	@EnableKafka
+	static class Config {
+
+		@Bean
+		KafkaAdmin admin(EmbeddedKafkaBroker broker) {
```
artembilan, line 105: I wonder why do we need this bean?

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
```diff
@@ -0,0 +1,210 @@
...
+			return new SimpleMeterRegistry();
+		}
+
+		@Bean
+		ObservationListener observationListener() {
+			return new ObservationListener();
+		}
+
+		@Bean
+		SimpleTracer simpleTracer() {
```
artembilan, line 166: Do we use this bean somewhere?

artembilan (changes requested):
Just a couple remarks.
Thanks

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
sobychacko: sorry - copy error from the other test :)

artembilan:
Yeah...
Looks like we have some breaking changes between `3.2.x` and `3.3.x`.
We just cannot cherry-pick this fix over there down to `3.2.x`.

Let's see if we want to have this fix in `3.2.x` at all!
Otherwise just aim it maximum for `3.3.x` and that's it.

PR #3766, opened 2025-02-24
feat: Replace LinkedList with ArrayList

### LINKED ISSUE #3764: Improve performance of acknowledge(int index) / override createRecordList implementation
**Expected Behavior**

Acknowledging an index in a batch should ideally have constant time, but be no worse than linear.

**Current Behavior**

Acknowledging an index in a batch has quadratic time `N(N+1)/2` ~ `N^2`

**Context**

Batch consumers operate on a `LinkedList` of records. If the consumer uses `MANUAL_IMMEDIATE` ack mode, and the listener invokes `acknowledgement.acknowledge(index)` where `index` is relatively big (eg. when processing batches of 100k), performance takes hit because of the linear lookup `records.get(i)` in a loop [here](https://github.com/spring-projects/spring-kafka/blob/v3.3.2/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L3587-L3588)

Maybe the most robust solution would be to offer clients the ability to customise how the record list is created? eg.

eg. 
```java
ConsumerFactory#setRecordListFactory(Function<ConsumerRecords<K, V>,List<ConsumerRecord<K, V>> factory);
```

This way we could pass in other implementations to replace [`createRecordList`](https://github.com/spring-projects/spring-kafka/blob/v3.3.2/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2237C55-L2243)

eg.

```java
private List<ConsumerRecord<K, V>> createRecordList(final ConsumerRecords<K, V> records) {
  Iterator<ConsumerRecord<K, V>> iterator = records.iterator();
  ConsumerRecord<K, V>[] recordsArray = (ConsumerRecord<K, V>[]) Array.newInstance(ConsumerRecord.class, records.count());
  int index = 0;
  while (iterator.hasNext()) {
    recordsArray[index] = iterator.next();
    index += 1;
  }
  return List.of(recordsArray);
}
```

I also tried to directly replace the existing `createRecordList` with the above implementation; however, it caused one test to fail (`SubBatchPerPartitionTests > withFilter()`) because `List.of` produces an immutable list. `List.of` could be replaced with `Arrays.asList` (and yield an `ArrayList`), but then filtering would be less efficient (.. i presume, but haven't verified).

janeklb:
Initialise the "batch of records" list container as an ArrayList rather than LinkedList. 

Using a LinkedList has an impact on performance as outlined here https://github.com/spring-projects/spring-kafka/issues/3764

(fixes #3764)

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -16,7 +16,6 @@
 
 package org.springframework.kafka.listener;
 
-import java.lang.reflect.Array;
 import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.AbstractMap.SimpleEntry;
@@ -2239,15 +2238,9 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 		}
 
 		private List<ConsumerRecord<K, V>> createRecordList(final ConsumerRecords<K, V> records) {
-			Iterator<ConsumerRecord<K, V>> iterator = records.iterator();
-			@SuppressWarnings("unchecked") ConsumerRecord<K, V>[] recordsArray =
-					(ConsumerRecord<K, V>[]) Array.newInstance(ConsumerRecord.class, records.count());
-			int index = 0;
-			while (iterator.hasNext()) {
-				recordsArray[index] = iterator.next();
-				index += 1;
-			}
-			return Arrays.asList(recordsArray);
+			List<ConsumerRecord<K, V>> recordList = new ArrayList<>(records.count());
+			records.forEach(recordList::add);
+			return recordList;
 		}
 
 		/**
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -173,6 +173,7 @@ import org.springframework.util.StringUtils;
  * @author Sanghyeok An
  * @author Christian Fredriksson
  * @author Timofey Barabanov
+ * @author Janek Lasocki-Biczysko
  */
 public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 		extends AbstractMessageListenerContainer<K, V> implements ConsumerPauseResumeEventPublisher {
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -16,6 +16,7 @@
 
 package org.springframework.kafka.listener;
 
+import java.lang.reflect.Array;
 import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.AbstractMap.SimpleEntry;
@@ -2238,11 +2239,14 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 
 		private List<ConsumerRecord<K, V>> createRecordList(final ConsumerRecords<K, V> records) {
 			Iterator<ConsumerRecord<K, V>> iterator = records.iterator();
-			List<ConsumerRecord<K, V>> list = new LinkedList<>();
+			@SuppressWarnings("unchecked") ConsumerRecord<K, V>[] recordsArray =
+					(ConsumerRecord<K, V>[]) Array.newInstance(ConsumerRecord.class, records.count());
+			int index = 0;
 			while (iterator.hasNext()) {
-				list.add(iterator.next());
+				recordsArray[index] = iterator.next();
+				index += 1;
 			}
-			return list;
+			return Arrays.asList(recordsArray);
 		}
 
 		/**
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2025 the original author or authors.
+ * Copyright 2016-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,7 +16,6 @@
 
 package org.springframework.kafka.listener.adapter;
 
-import java.util.Arrays;
 import java.util.List;
 
 import org.apache.kafka.clients.consumer.ConsumerRecord;
@@ -33,7 +32,6 @@ import org.springframework.kafka.listener.BatchMessageListener;
  *
  * @author Gary Russell
  * @author Sanghyeok An
- * @author Janek Lasocki-Biczysko
  */
 public interface RecordFilterStrategy<K, V> {
 
@@ -51,10 +49,9 @@ public interface RecordFilterStrategy<K, V> {
 	 * @return the filtered records.
 	 * @since 2.8
 	 */
-	@SuppressWarnings("unchecked")
 	default List<ConsumerRecord<K, V>> filterBatch(List<ConsumerRecord<K, V>> records) {
-		var recordsArray = records.stream().filter(record -> !this.filter(record)).toArray(ConsumerRecord[]::new);
-		return Arrays.asList(recordsArray);
+		records.removeIf(this::filter);
+		return records;
 	}
 
 	/**
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2024 the original author or authors.
+ * Copyright 2016-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -32,6 +32,7 @@ import org.springframework.kafka.listener.BatchMessageListener;
  *
  * @author Gary Russell
  * @author Sanghyeok An
+ * @author Janek Lasocki-Biczysko
  */
 public interface RecordFilterStrategy<K, V> {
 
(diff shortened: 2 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2238,11 +2239,14 @@ protected void doInTransactionWithoutResult(TransactionStatus status) {
 
 		private List<ConsumerRecord<K, V>> createRecordList(final ConsumerRecords<K, V> records) {
 			Iterator<ConsumerRecord<K, V>> iterator = records.iterator();
-			List<ConsumerRecord<K, V>> list = new LinkedList<>();
+			@SuppressWarnings("unchecked") ConsumerRecord<K, V>[] recordsArray =
+					(ConsumerRecord<K, V>[]) Array.newInstance(ConsumerRecord.class, records.count());
+			int index = 0;
 			while (iterator.hasNext()) {
-				list.add(iterator.next());
+				recordsArray[index] = iterator.next();
```
artembilan, line 2246: Cannot we do like this instead:
```
recordsArray[index++]
```
?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2238,11 +2239,14 @@ protected void doInTransactionWithoutResult(TransactionStatus status) {
...
+				index += 1;
```
janeklb, line 2247:
```suggestion
				recordsArray[index++] = iterator.next();
```

artembilan: Thanks.
You can apply this change yourself from this UI.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: Thanks!
That is great discussion and I agree that paradox of choice may cause confusions.

So, I'm accepting your variant of the code to make it more readable.

Please, add your name to the `@author` list of this changed class.

janeklb:
> We got a failed test with your change:
> 
> ```
>  SubBatchPerPartitionTests > withFilter() FAILED
>     org.opentest4j.AssertionFailedError: 
>     Expecting value to be true but was false
>         at app//org.springframework.kafka.listener.SubBatchPerPartitionTests.withFilter(SubBatchPerPartitionTests.java:121)
> ```
> 
> And you indeed mentioned that in the issue. WDYT?

Apologies -- I raised the PR too quickly. I assumed using `Arrays.asList` would have fixed the failing test (because `List.of` produces an immutable list and the failure seemed to be related to mutation..), but I only re-ran the tests locally after pushing this PR. Will investigate and let you know shortly.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2238,11 +2240,14 @@ protected void doInTransactionWithoutResult(TransactionStatus status) {
...
 			}
-			return list;
+			return List.of(recordsArray);
```
artembilan, line 2250: No, that's not what I was talking about.
To minimize braking change, please, let's consider to use here `new ArrayList()` instead.
And revert that change in the `RecordFilterStrategy`.

The point is that mostly this list is going to be used as is without any extra intermediate objects to mitigate its immutability. And if we look at this from the high point of view, it is really up to target application to deal with data properly, while we should let them to do with this list whatever they want and need.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
```diff
@@ -50,8 +51,7 @@ public interface RecordFilterStrategy<K, V> {
 	 * @since 2.8
 	 */
 	default List<ConsumerRecord<K, V>> filterBatch(List<ConsumerRecord<K, V>> records) {
-		records.removeIf(this::filter);
-		return records;
+		return records.stream().filter(record -> !this.filter(record)).toList();
```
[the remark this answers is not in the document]
janeklb: Please let me know how to proceed

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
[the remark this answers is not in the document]
artembilan: With `mutable` list, please.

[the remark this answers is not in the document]
janeklb: Oh, i should have explained this change..

`records.removeIf(this::filter);` does _not_ work with `ArrayList`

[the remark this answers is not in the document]
janeklb: Ok -- it is now returning a mutable list (`ArrayList` via `Arrays.asList`)

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2238,11 +2240,14 @@ protected void doInTransactionWithoutResult(TransactionStatus status) {
...
+			@SuppressWarnings("unchecked") ConsumerRecord<K, V>[] recordsArray =
+					(ConsumerRecord<K, V>[]) Array.newInstance(ConsumerRecord.class, records.count());
+			int index = 0;
 			while (iterator.hasNext()) {
-				list.add(iterator.next());
+				recordsArray[index] = iterator.next();
+				index += 1;
 			}
-			return list;
+			return Arrays.asList(recordsArray);
```
artembilan, line 2250: Why not `new ArrayList()`?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
```diff
@@ -49,9 +51,10 @@ public interface RecordFilterStrategy<K, V> {
 	 * @return the filtered records.
 	 * @since 2.8
 	 */
+	@SuppressWarnings("unchecked")
 	default List<ConsumerRecord<K, V>> filterBatch(List<ConsumerRecord<K, V>> records) {
-		records.removeIf(this::filter);
-		return records;
+		var recordsArray = records.stream().filter(record -> !this.filter(record)).toArray(ConsumerRecord[]::new);
+		return Arrays.asList(recordsArray);
```
artembilan, line 57: Why do we need this change?
Why that `removeIf()` doesn't work?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
janeklb: `Arrays.asList` is

```java
  public static <T> List<T> asList(T... a) {
        return new ArrayList<>(a);
    }
```

unfortunately `recordsArray` is just an array here, and `ArrayList` constructor requires a `Collection`..

I'm not exactly sure how it works, but it seems that java does some magic to transform `T... a` into `Collection<T>`

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
janeklb: I had `removeIf` here before, but tests failed because it was throwing an exception 

```
Caused by: java.lang.UnsupportedOperationException: remove
	at java.base/java.util.Iterator.remove(Iterator.java:102) ~[?:?]
	at java.base/java.util.Collection.removeIf(Collection.java:582) ~[?:?]
	at org.springframework.kafka.listener.adapter.RecordFilterStrategy.filterBatch(RecordFilterStrategy.java:57) ~[main/:?]
```

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: Right. Let's populate that `new ArrayList()` manually then!
Should not be technically too much difference with this `asList()` in performance.
I worry more about breaking change you made for `RecordFilterStrategy` and potential breaking changes in the target applications due to non-consistent result object behavior.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
janeklb: Stepping through it in a debugger showed that `Arrays$ArrayItr` does not support `remove`

image

artembilan: Therefore my request for the `new ArrayList()` instead.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordFilterStrategy.java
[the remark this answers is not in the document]
janeklb: My mistake -- the `ArrayList` returned by `Arrays.asList` is a different class (it's a private class in `Arrays`) than the "regular" `ArrayList`. I have taken your cue and adjusted. Thank you!

janeklb:
Turns out my comment here https://github.com/spring-projects/spring-kafka/issues/3764#issuecomment-2678719679 was totally wrong. I misunderstood the implementation of `ArrayList#add(E element)`. Your suggestion @artembilan https://github.com/spring-projects/spring-kafka/issues/3764#issuecomment-2678684528 works perfect 👍🏻

artembilan (approved):
LGTM.
Will be merged (and back-ported) when all tests pass.

Thank you for your contribution!

janeklb:
> LGTM. Will be merged (and back-ported) when all tests pass.
> 
> Thank you for your contribution!

Wonderful, thank you for helping me getting over the line. When should I expect a new 3.3.x release?

artembilan:
> When should I expect a new 3.3.x release?

See our Release Calendar for March: https://spring.io/projects#release-calendar

PR #3771, opened 2025-02-26
Polishing

ngocnhan-tran1996:


spring-kafka/src/main/java/org/springframework/kafka/support/Suffixer.java
```diff
@@ -42,7 +42,7 @@ public String maybeAddTo(String source) {
 		if (!StringUtils.hasText(this.suffix)) {
 			return source;
 		}
-		return source != null && StringUtils.hasText(source) // Only suffix if there's text
+		return StringUtils.hasText(source) // Only suffix if there's text
```
sobychacko, line 45: When making a code change, please add your name as the author to the class. Thanks!

sobychacko (changes requested):

PR #3778, opened 2025-03-04
Handle null group id in listener observation

cfredri4:
This change fixes an NPE when group id is null and observation is enabled.


cfredri4:
@sobychacko done.

Issue #3786, opened 2025-03-07
Kafka ProducerRecord may end up with duplicated trace headers


GFriedrich:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**
3.2.* up to the latest 3.3.*

**Describe the bug**
I found a bit of a weird behaviour when enabling the observation on Spring Integrations and the Spring Kafka binder.
In this case the Spring Integration already adds a `traceparent` header (e.g. when using W3C traces) to the internal `Message` but later on the Spring Kafka binder will also add another `traceparent` header.

**To Reproduce**
* Enable the observation for Spring Integrations via `spring.integration.management.observation-patterns`
* Enable Spring Kafka binder observations via `spring.cloud.stream.kafka.binder.enable-observation`
* Send a message to a `StreamBridge`
* See that the produced record on Kafka has two `traceparent` headers

**Expected behavior**
There should be just one `traceparent` header on a produced record.

**Sample**
see https://github.com/spring-cloud/spring-cloud-stream/issues/3095#issuecomment-2707075861

**Additional info**
This issue originates from https://github.com/spring-cloud/spring-cloud-stream/issues/3095 but it was suggested in the issue over there that the underlying problem should be fixed in the Spring Kafka library instead.

artembilan:
@sobychacko ,

let me know if you'd like to discuss this.

PR #3789, opened 2025-03-13
GH-3786: Remove ProducerRecord duplicated traceparent header

sobychacko:
Fixes: #3786
Issue link: https://github.com/spring-projects/spring-kafka/issues/3786

When tracing is enabled, the KafkaRecordSenderContext was adding a new traceparent header without removing existing ones, resulting in multiple traceparent headers in the same record. This commit fixes the issue by Updating KafkaRecordSenderContext to remove existing traceparent headers before adding new ones.

**Auto-cherry-pick to `3.3.x` & `3.2.x`**

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordSenderContext.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordSenderContext.java
@@ -21,6 +21,7 @@ import java.util.function.Supplier;
 
 import io.micrometer.observation.transport.SenderContext;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.Headers;
 
 /**
@@ -43,8 +44,8 @@ public class KafkaRecordSenderContext extends SenderContext<ProducerRecord<?, ?>
 	public KafkaRecordSenderContext(ProducerRecord<?, ?> record, String beanName, Supplier<String> clusterId) {
 		super((carrier, key, value) -> {
 			Headers headers = record.headers();
-			// For traceparent context headers, ensure there's only one
-			if ("traceparent".equals(key)) {
+			Iterable<Header> existingHeaders = headers.headers(key);
+			if (existingHeaders.iterator().hasNext()) {
 				headers.remove(key);
 			}
 			headers.add(key, value == null ? null : value.getBytes(StandardCharsets.UTF_8));
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordSenderContext.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordSenderContext.java
@@ -21,7 +21,6 @@ import java.util.function.Supplier;
 
 import io.micrometer.observation.transport.SenderContext;
 import org.apache.kafka.clients.producer.ProducerRecord;
-import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.Headers;
 
 /**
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordSenderContext.java
```diff
@@ -39,8 +41,15 @@ public class KafkaRecordSenderContext extends SenderContext<ProducerRecord<?, ?>
 	private final ProducerRecord<?, ?> record;
 
 	public KafkaRecordSenderContext(ProducerRecord<?, ?> record, String beanName, Supplier<String> clusterId) {
-		super((carrier, key, value) -> record.headers().add(key,
-				value == null ? null : value.getBytes(StandardCharsets.UTF_8)));
+		super((carrier, key, value) -> {
+			Headers headers = record.headers();
+			// For traceparent context headers, ensure there's only one
+			if ("traceparent".equals(key)) {
```
artembilan, line 47: This is not correct.
Different vendors provides different header name.
I have so far know here: https://github.com/spring-cloud/spring-cloud-stream-binder-aws-kinesis/blob/main/spring-cloud-stream-binder-kinesis/src/main/java/org/springframework/cloud/stream/binder/kinesis/KinesisMessageChannelBinder.java#L575-L577.

But that does not mean that anything else custom could be provided from the context supplier.

I think it is totally safe to replace whatever was asked from the supplier.
The point is to not mislead with extra entry from what might came from the previous observation.
We have to provide here what can be set from keys supplier.

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/ObservationTests.java
```diff
@@ -449,6 +456,65 @@ void kafkaAdminNotRecreatedIfBootstrapServersSameInProducerAndAdminConfig(
...
+				record,
+				"test-bean",
+				() -> "test-cluster"
+		);
+		context.getSetter().set(record, "traceparent", updatedTraceParent);
+
+		Iterable<Header> traceparentHeaders = record.headers().headers("traceparent");
+		List<String> headerValues = StreamSupport.stream(traceparentHeaders.spliterator(), false)
+				.map(header -> new String(header.value(), StandardCharsets.UTF_8))
+				.collect(Collectors.toList());
```
artembilan, line 479: No reason in this. Just `toList()` is enough.

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/ObservationTests.java
```diff
@@ -449,6 +456,65 @@ void kafkaAdminNotRecreatedIfBootstrapServersSameInProducerAndAdminConfig(
...
+
+		// Verify there's only one traceparent header and it contains the updated value
+		assertThat(headerValues).containsExactly(updatedTraceParent);
+	}
+
+	// https://github.com/spring-cloud/spring-cloud-stream/issues/3095#issuecomment-2707075861
+	// https://github.com/spring-projects/spring-kafka/issues/3786
```
artembilan, line 486: I don't think there is a reason in this comments.
We always can find out that from the commit history.

artembilan (changes requested):
Some review.
Thanks

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaRecordSenderContext.java
```diff
@@ -39,8 +42,15 @@ public class KafkaRecordSenderContext extends SenderContext<ProducerRecord<?, ?>
 	private final ProducerRecord<?, ?> record;
 
 	public KafkaRecordSenderContext(ProducerRecord<?, ?> record, String beanName, Supplier<String> clusterId) {
-		super((carrier, key, value) -> record.headers().add(key,
-				value == null ? null : value.getBytes(StandardCharsets.UTF_8)));
+		super((carrier, key, value) -> {
+			Headers headers = record.headers();
+			Iterable<Header> existingHeaders = headers.headers(key);
+			if (existingHeaders.iterator().hasNext()) {
```
artembilan, line 48: Why do we need to do this check?
Why just plain `remove()` not enough?

artembilan (changes requested):

artembilan:
Auto-cherry-pick has failed.
Please, take look.
Thanks

PR #3809, opened 2025-03-19
Refactor EmbeddedKafkaZKBroker to extract "num.partitions" to a constant

ghgh415263:
The "num.partitions" property is used twice in the EmbeddedKafkaZKBroker class. It would be more manageable if we define it as a single constant.
I considered this PR to be minor, so I didn’t create an issue for it.

I appreciate you taking the time to read this

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaZKBroker.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaZKBroker.java
@@ -104,6 +104,8 @@ public class EmbeddedKafkaZKBroker implements EmbeddedKafkaBroker {
 
 	public static final int DEFAULT_ZK_CONNECTION_TIMEOUT = DEFAULT_ZK_SESSION_TIMEOUT;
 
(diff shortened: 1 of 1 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaZKBroker.java
```diff
@@ -104,6 +104,8 @@ public class EmbeddedKafkaZKBroker implements EmbeddedKafkaBroker {
 
 	public static final int DEFAULT_ZK_CONNECTION_TIMEOUT = DEFAULT_ZK_SESSION_TIMEOUT;
 
+	private static final String NUM_PARTITIONS = "num.partitions";
```
artembilan, line 107: This is broker property.
It has nothing to do with the client application.
Why would one need to have it as a top-level API, since this is only a part if this class?
There is a dedicate constructor argument for this config.

On the other hand, with Apache Kafka `4.0` we are aim to remove Zookeeper support from the project.
So, your effort for this property will be eliminated.

Please, elaborate the reasoning for such a change.

PR #3815, opened 2025-03-20
GH-3814: Upgrade Apache Kafka client dependencies to version 4.0.0

sobychacko:
Fixes: #3814

Issue link: https://github.com/spring-projects/spring-kafka/issues/3814

This commit upgrades the Apache Kafka client to 4.0.0 with the following changes:

- Upgrade Kafka client dependencies to 4.0.0 in build.gradle
- Remove ZooKeeper-based broker implementation as Kafka 4.0 fully transitions to KRaft mode
  - Delete EmbeddedKafkaZKBroker class and related tests
  - Remove ZooKeeper dependency
  - Refactor EmbeddedKafkaRule to use KRaft exclusively
  - Update KafkaClusterTestKit imports to use new packages

- Update ConsumerRecords constructor calls to include the new required Map parameter
- Add implementations for new Producer interface methods:
  - registerMetricForSubscription
  - unregisterMetricFromSubscription

- Update KafkaStreamBrancher to use new split() and branch() methods
- Remove deprecated partitioner classes from runtime hints
- Remove deprecated sendOffsetsToTransaction method that used String consumerGroupId
- Update BrokerAddress to use org.apache.kafka.server.network.BrokerEndPoint
- Update DeserializationExceptionHandler to use new ErrorHandlerContext

The commit also includes test modifications to address limitations with static port assignments in KRaft mode and adjustments to replication factors in tests.


build.gradle
```diff
@@ -355,16 +355,14 @@ project ('spring-kafka-test') {
 	description = 'Spring Kafka Test Support'
 
 	dependencies {
+		implementation "org.apache.logging.log4j:log4j-slf4j-impl:$log4jVersion"
```
artembilan, line 358: Has to be `api` .
That's exactly how Java libraries are designed with Gradle.

spring-kafka/src/main/java/org/springframework/kafka/support/KafkaStreamBrancher.java
```diff
@@ -84,22 +88,37 @@ public KafkaStreamBrancher<K, V> defaultBranch(Consumer<? super KStream<K, V>> c
 
 	/**
 	 * Terminating method that builds branches on top of given {@code KStream}.
+	 * Applies each predicate-consumer pair sequentially to create branches.
+	 * If a default consumer exists, it will handle all records that don't match any predicates.
+	 *
```
artembilan, line 93: No blank lines in method Javadocs.

artembilan (changes requested):
Just couple simple nit-picks.
This is really cool!
And change is not that bad: mostly tests adjustments because of new ctor for `ConsumerRecords` 😄 

The removal of `EmbeddedKafkaRule` is a bit destructive. But who uses JUnit 4 these days!
So, or we revise it for KRaft or just bite a bullet.

The docs has to be changed as well.
That might be some effort, so consider to do that in the separate issue/PR.

Thanks

PR #3821, opened 2025-03-24
Adding whats-new section for 4.0.0

sobychacko:


artembilan (approved):
This is all good, Soby!

Thank you!

No need to address in this PR, but probably as separate one(s):

- Mention JSpecify migration
- Walk though all the docs to migrate them to this our new reality about Kafka 4.0.

Glad to see that such a migration was really smooth and we have a plenty of time to do anything else!

PR #3823, opened 2025-03-25
GH-3808: ReplyingKafkaTemplate observation on reply

### LINKED ISSUE #3808: Enable the tracing on the reply of a ReplyingKafkaTemplate
Spring Kafka supports observation and, more specifically, tracing on Kafka templates and Kafka listeners, but tracing is not supported on the reply of a ReplyingKafkaTemplate.

As the ReplyingKafkaTemplate consumes the replies using a  batch listener, it should be possible to reconstruct the tracing context (traceId and baggages) for each record when iterating over these records. This can be done by starting an observation for each record (see ReplyingKafkaTemplate#onMessage), similar as it's done in KafkaMessageListenerContainer#doInvokeRecordListener.

If we agree on this request, I could help by filing a PR.

Any help or suggestions to avoid having to handle it manually for each reply would be more than welcome.

frosiere:
Closes #3808

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -104,6 +104,7 @@ import org.springframework.util.StringUtils;
  * @author Gurps Bassi
  * @author Valentina Armenise
  * @author Christian Fredriksson
+ * @author Francois Rosiere
  */
 public class KafkaTemplate<K, V> implements KafkaOperations<K, V>, ApplicationContextAware, BeanNameAware,
 		ApplicationListener<ContextStoppedEvent>, DisposableBean, SmartInitializingSingleton {
@@ -468,9 +469,9 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V>, ApplicationCo
 	/**
 	 * Return the {@link ObservationRegistry} used by the template.
 	 * @return the observation registry
-	 * @since 3.3.5
+	 * @since 3.2.9
 	 */
-	public ObservationRegistry getObservationRegistry() {
+	protected ObservationRegistry getObservationRegistry() {
 		return this.observationRegistry;
 	}
 
@@ -545,7 +546,7 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V>, ApplicationCo
 	/**
 	 * Return the cluster id, if available.
 	 * @return the cluster id.
-	 * @since 3.3.5
+	 * @since 3.2.9
 	 */
 	@Nullable
 	protected String clusterId() {
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
@@ -466,6 +466,15 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V>, ApplicationCo
 	}
 
 	/**
+	 * Return the {@link ObservationRegistry} used by the template.
+	 * @return the observation registry
+	 * @since 3.3.5
+	 */
+	public ObservationRegistry getObservationRegistry() {
+		return this.observationRegistry;
+	}
+
+	/**
 	 * Return the {@link KafkaAdmin}, used to find the cluster id for observation, if
 	 * present.
 	 * @return the kafkaAdmin
@@ -533,8 +542,13 @@ public class KafkaTemplate<K, V> implements KafkaOperations<K, V>, ApplicationCo
 		return removeLeadingAndTrailingBrackets(adminServers == null ? "" : adminServers);
 	}
 
+	/**
+	 * Return the cluster id, if available.
+	 * @return the cluster id.
+	 * @since 3.3.5
+	 */
 	@Nullable
-	private String clusterId() {
+	protected String clusterId() {
 		if (this.kafkaAdmin != null && this.clusterId == null) {
 			this.clusterIdLock.lock();
 			try {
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -465,6 +465,15 @@ public void setObservationRegistry(ObservationRegistry observationRegistry) {
 		this.observationRegistry = observationRegistry;
 	}
 
+	/**
+	 * Return the {@link ObservationRegistry} used by the template.
+	 * @return the observation registry
+	 * @since 3.3.5
```
sobychacko, line 471: We can potentially backport this to 3.2.x. If that's the case, `@since` tag needs to be updates.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
```diff
@@ -465,6 +465,15 @@ public void setObservationRegistry(ObservationRegistry observationRegistry) {
...
+	 */
+	public ObservationRegistry getObservationRegistry() {
```
artembilan, line 473: Is it really has to be `public`?
Isn't just `protected` enough for our requirements in the `ReplyingKafkaTemplate`?

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
```diff
@@ -502,39 +506,50 @@ private static <K, V> CorrelationKey defaultCorrelationIdStrategy(
...
-				correlationId = this.binaryCorrelation
-						? new CorrelationKey(correlationHeader.value())
-						: new String(correlationHeader.value(), StandardCharsets.UTF_8);
-			}
-			if (correlationId == null) {
-				this.logger.error(() -> "No correlationId found in reply: " + KafkaUtils.format(record)
-						+ " - to use request/reply semantics, the responding server must return the correlation id "
-						+ " in the '" + this.correlationHeaderName + "' header");
+			ContainerProperties containerProperties = this.replyContainer.getContainerProperties();
+			Observation observation = KafkaListenerObservation.LISTENER_OBSERVATION.observation(
```
artembilan, line 510: Do we really need to do this manually here?
Why enabling observation on the provided `GenericMessageListenerContainer` is not enough?
See more info in docs: https://docs.spring.io/spring-kafka/reference/kafka/micrometer.html#observation.

With that it feels like nothing to do in the code, unless we would like to improve docs for `ReplyingKafkaTemplate` mentioning such an observation option in the `ContainerProperties`: https://docs.spring.io/spring-kafka/reference/kafka/container-props.html#observationEnabled

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
frosiere: Fair point, I can adapt.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
frosiere: Good point, protected visibility is indeed enough.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
frosiere: Observation is only supported on single record listeners. Batch listeners do not currently support observation. This would require another user contract to enable creating/building one observation per batch record (a batch may contain records associated to multiple observations). 
Since the ReplyingKafkaTemplate is consuming the reply through a batch listener, there's no way to automate it. I wanted to have a utility class but, in the end, the code wasn't smaller.
Did I missed something?

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/ObservationTests.java
```diff
@@ -511,6 +520,20 @@ public void onSuccess(ProducerRecord<Integer, String> producerRecord, RecordMeta
...
+	void testReplyingKafkaTemplateObservation(
+			@Autowired ReplyingKafkaTemplate<Integer, String, String> template,
+			@Autowired ObservationRegistry observationRegistry) {
+		AtomicReference<KafkaRecordReceiverContext> replyObservationContext = new AtomicReference<>();
+		template.sendAndReceive(new ProducerRecord<>(OBSERVATION_TEST_4, "test")).thenAccept(replyRecord -> {
+			Observation.Context observationContext = observationRegistry.getCurrentObservation().getContext();
+			assertThat(observationContext).isInstanceOf(KafkaRecordReceiverContext.class);
+			replyObservationContext.set((KafkaRecordReceiverContext) observationContext);
+		});
+		Awaitility.await().atMost(Duration.ofSeconds(60)).until(() ->
```
artembilan, line 533: See AssertJ API for `CompletableFuture` instead.
That `template.sendAndReceive()` returns for us exactly one we would like to verify.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java
frosiere: Updated to 3.2.9

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/ObservationTests.java
frosiere: Much better, fixed.

Issue #3847, opened 2025-04-17
Verify support for the new Consumer Rebalance protocol in Spring Kafka


sobychacko:
[KIP-848 ](https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol)added support for the new consumer rebalance protocol in Apache Kafka 4.0.0. For the most part, the new protocol is a seamless drop-in replacement in end-user applications using the client, as most of the heavy lifting is done on the server side. 

Verify this feature is supported in Spring Kafka 4.0.0 and introduce the necessary steps if needed.

PR #3848, opened 2025-04-17
GH-3847: Add support for Kafka 4.0 new consumer rebalance protocol (K…

sobychacko:
…IP-848)

Fixes: #3847

Issue link: https://github.com/spring-projects/spring-kafka/issues/3847

KIP details: https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol

* Add custom assignor warning in DefaultKafkaConsumerFactory for group.protocol=consumer
* Add ConsumerAwareRebalanceListenerTests to verify incremental rebalancing (next gen rebalance) and the legacy rebalancer
* Reference Docs updates

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/rebalance-listeners.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/rebalance-listeners.adoc
@@ -72,6 +72,9 @@ To enable the new protocol, configure the `group.protocol` property:
 spring.kafka.consumer.properties.group.protocol=consumer
 ----
 
+Keep in mind that, the above property is a Spring Boot property.
+If you are not using Spring Boot, you may want to set it manually as shown below.
+
 Alternatively, set it programmatically:
 
 [source, java]
@@ -82,10 +85,8 @@ ConsumerFactory<String, String> factory = new DefaultKafkaConsumerFactory<>(prop
 ----
 
 The new protocol works seamlessly with `ConsumerAwareRebalanceListener`.
-Due to incremental rebalancing, `onPartitionsAssigned` may be called multiple times with smaller partition sets, unlike the single callback typical of the legacy protocol (`group.protocol=classic`).
+Due to incremental rebalancing, `onPartitionsAssigned` may be called multiple times with smaller partition sets, unlike the single callback typical of the legacy protocol.
 
 The new protocol uses server-side partition assignments, ignoring client-side custom assignors set via `spring.kafka.consumer.partition-assignment-strategy`.
 A warning is logged if a custom assignor is detected.
-To use custom assignors, set `group.protocol=classic`.
-
-Test applications with `group.protocol=consumer` in a non-production environment to assess rebalancing impacts, and monitor callback frequency for performance tuning.
+To use custom assignors, set `group.protocol=classic` (which is the default if you don't specify a value for `group.protocol`).
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/rebalance-listeners.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/rebalance-listeners.adoc
@@ -59,3 +59,33 @@ If you implement `ConsumerRebalanceListener` you should override the default met
 This is because the listener container will call its own `onPartitionsRevoked` from its implementation of `onPartitionsLost` after calling the method on your implementation.
 If you implementation delegates to the default behavior, `onPartitionsRevoked` will be called twice each time the `Consumer` calls that method on the container's listener.
 
+[[new-rebalalcne-protocol]]
+== Kafka 4.0 Consumer Rebalance Protocol
+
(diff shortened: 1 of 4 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/rebalance-listeners.adoc
```diff
@@ -59,3 +59,33 @@ If you implement `ConsumerRebalanceListener` you should override the default met
...
+== Kafka 4.0 Consumer Rebalance Protocol
+
+Spring for Apache Kafka 4.0 supports Apache https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol[Kafka 4.0’s new consumer rebalance protocol] (KIP-848), which enhances performance with server-driven, incremental partition assignments.
+This reduces rebalancing downtime for consumer groups.
+
+To enable the new protocol, configure the `group.protocol` property:
+
+[source, properties]
+----
+spring.kafka.consumer.properties.group.protocol=consumer
```
artembilan, line 72: This is Spring Boot property style.
Might be good idea to mention that.
Or just don't mess up our docs with Boot to avoid cyclic dependency 😄

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/rebalance-listeners.adoc
```diff
@@ -59,3 +59,33 @@ If you implement `ConsumerRebalanceListener` you should override the default met
...
+
+[source, java]
+----
+Map<String, Object> props = new HashMap<>();
+props.put("group.protocol", "consumer");
+ConsumerFactory<String, String> factory = new DefaultKafkaConsumerFactory<>(props);
+----
+
+The new protocol works seamlessly with `ConsumerAwareRebalanceListener`.
+Due to incremental rebalancing, `onPartitionsAssigned` may be called multiple times with smaller partition sets, unlike the single callback typical of the legacy protocol (`group.protocol=classic`).
```
artembilan, line 85: Is `classic` a default value for that property?
Should we mention that?
Is there a clue that is will be changed in the future?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/rebalance-listeners.adoc
```diff
@@ -59,3 +59,33 @@ If you implement `ConsumerRebalanceListener` you should override the default met
...
+
+The new protocol uses server-side partition assignments, ignoring client-side custom assignors set via `spring.kafka.consumer.partition-assignment-strategy`.
+A warning is logged if a custom assignor is detected.
+To use custom assignors, set `group.protocol=classic`.
+
+Test applications with `group.protocol=consumer` in a non-production environment to assess rebalancing impacts, and monitor callback frequency for performance tuning.
```
artembilan, line 91: Why do we need this sentence?

spring-kafka/src/test/java/org/springframework/kafka/listener/ConsumerRebalanceProtocolTests.java
```diff
@@ -0,0 +1,174 @@
...
+		// longer duration.
+		Awaitility.await().timeout(Duration.ofSeconds(30))
+				.untilAsserted(() -> assertThat(assignedPartitionSet.size() == 6).isTrue());
+
+		container1.stop();
+		container2.stop();
+	}
+
+	@Configuration
+	@EnableKafka
```
artembilan, line 146: Looks like Spring Kafka infrastructure is out of use in this test suite.
More over, I think this test environment (according to test methods body) has nothing to do with Spring at all...

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/rebalance-listeners.adoc
sobychacko: Reworked the docs a bit.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConsumerRebalanceProtocolTests.java
sobychacko: I'm not sure what you mean. Although this feature does not require any major changes in Spring Kafka, since it is a major feature in 4.0.0 Kafka, I think we need to verify that it works seamlessly from our side. It's more like a verification that there is no regression in Spring Kafka, regardless of the rebalance protocol used. If any issues are reported in this area, we can potentially use this test as a springboard to debug.  With that said, we can also convert this into a sample app in our examples, and then remove the test.

artembilan: I mean what is a purpose of `@EnableKafka` over here if there is no any `@KafkaListener`?
Why do we need a `@Configuration` at all if Kafka configs can be created as not beans?

therefore my general question: what does Spring do in this test suite at all?

sobychacko: Ah ok. got it. I removed the use of `@EnableKafka`, but left the Spring `@Configuration` for the convenience of injecting `EmbeddedKafka`. I think its ok to use it like that for the test, but if you prefer strongly, feel free to rework before the merge. I can also take a look at that later today.

artembilan (approved):
LGTM.
Will be merged when PR build is green.

Issue #3807, opened 2025-03-18
Necessity of KafkaHandler on single method class


artembilan:
### Discussed in https://github.com/spring-projects/spring-kafka/discussions/3804

<div type='discussions-op-text'>

<sup>Originally posted by **vitorsalves** March 15, 2025</sup>
Hi everyone,
Quick question, in a recent project i had to implement serval KafkaListener all with similar annotations and i wanted to combine them in to one (In my case it was @Component, @ConditionalOnProperty and @KafkaListener). However even if all my classes only had a singular method i still had to add @KafkaHandler. While this isn't a huge issue, it does feel a bit redundant.

I understand very well the need of @KafkaHandler in a multimethod @KafkaListener, but in the case a singular method class it seems unnecessary. So i ask, is there a reason that in the case of a class-level annotated KafkaListener with only a singular method, KafkaHandler is still required?</div>

chickenchickenlove:
Hi, @artembilan . 
Would it be okay if i look into this issue?

artembilan:
Sure thing, @chickenchickenlove !

chickenchickenlove:
Appreciate it, as always! 😃 
If it turns out to be toooooo difficult for me to handle, I will let you know!

PR #3865, opened 2025-04-24
spring-projectsGH-3807: Necessity of KafkaHandler on single method class

chickenchickenlove:
Fixes: https://github.com/spring-projects/spring-kafka/issues/3807
Issue link: https://github.com/spring-projects/spring-kafka/issues/3807

From now on, the single public method without method-level `@KafkaHandler`, `@KafkaListener` in class which has class-level `@KafkaListener` will be registered as `MethodKafkaListenerEndpoint` so that it consumes messages.


spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaHandlerParser.java
```diff
@@ -0,0 +1,81 @@
...
+ * Find the appropriate handler method when the target class has a class-level {@link KafkaListener}
+ * annotation and contains a single public method without a {@link KafkaHandler} annotation.
+ *
+ * @author Sanghyeok An
+ *
+ * @since 4.0
+ *
+ * @see KafkaListenerAnnotationBeanPostProcessor
+ */
+public class KafkaHandlerParser {
```
artembilan, line 39: This logic doesn't bring any value to be a new class.
I think ` MethodIntrospector.selectMethods` logic in the `KafkaListenerAnnotationBeanPostProcessor` has to be just adjusted to check for `@KafkaHandler` or `@KafkaListener` on methods or the method is single and doesn't matter what annotations on it.

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/annotation/AliasPropertiesTests.java
```diff
@@ -220,6 +228,18 @@ public void listen1(String in) {
 
 		}
 
+		@Component
+		@MyListener(id = "onSinglePublicMethod", value = "alias.tests")
+		public static class ClassLevelListenerWithPublicMethodWithoutHandlerAnnotation {
```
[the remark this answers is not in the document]
chickenchickenlove: The `@MyListener` annotation is a custom annotation that inherits from @KafkaListener.
I believe users can use both `@KafkaListener` and custom annotations that inherit from it.

So, we also test about inherited Annotation of `@KafkaListener` as well.

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerTests.java
```diff
@@ -0,0 +1,239 @@
...
+			return new KafkaListenerEndpointRegistry();
+		}
+
+		@Bean
+		public EmbeddedKafkaBroker embeddedKafka() {
+			return new EmbeddedKafkaKraftBroker(1, 1, TEST_TOPIC);
+		}
+
+		@Bean
+		public KafkaListenerContainerFactory<?> kafkaListenerContainerFactory() {
```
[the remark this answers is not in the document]
chickenchickenlove: AFAIK, `@EnableKafka` don't create `KafkaListenerContainerFactory` 
Actually, If I delete this code lines, `Spring Context` throws error `Failed to load ApplicationContext` because it couldn't find `No bean named 'kafkaListenerContainerFactory' available`. 

Also, we use custom consumer configuration for each test cases. 
So, I think it would be better to create `kafkaListenerContainerFactory` and `DefaultKafkaConsumerFactory`.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -411,8 +412,23 @@ public Object postProcessAfterInitialization(final Object bean, final String bea
...
-					List<Method> multiMethods = new ArrayList<>(methodsWithHandler);
-					processMultiMethodListeners(classLevelListeners, multiMethods, targetClass, bean, beanName);
+					Set<Method> publicMethods = MethodIntrospector.selectMethods(targetClass,
+							(ReflectionUtils.MethodFilter) method ->
+									Modifier.isPublic(method.getModifiers()));
+
+					if (methodsWithHandler.isEmpty() && publicMethods.size() == 1 && !hasMethodLevelListeners) {
+						// Case when target class has class-level @KafkaListener annotation and
+						// has only single public method without @KafkaHandler.
+						// See, GH-3807 for more details.
```
artembilan, line 422: I don't think we need this comment here. The behavior is clear from the logic of these `if..else`

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerTests.java
```diff
@@ -0,0 +1,234 @@
...
+			return KafkaTestUtils.producerProps(this.broker.getBrokersAsString());
+		}
+
+		@Component
+		@KafkaListener(id = "classAndMethodLevelListener", topics = "default-listener.tests")
+		public static class ClassAndMethodLevelListener {
+
+			final CountDownLatch latch = new CountDownLatch(1);
+
+			@KafkaHandler
```
artembilan, line 167: Why do we need to test this here?
I think the `@KafkaHandler` is covered in other tests, e.g. `EnableKafkaIntegrationTests`.

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerTests.java
```diff
@@ -0,0 +1,234 @@
...
+			final CountDownLatch latch = new CountDownLatch(1);
+
+			public void listen(String in) {
+				this.latch.countDown();
+			}
+		}
+
+		@Component
+		@KafkaListener(id = "otherClassLevelListenerWithSinglePublicMethod", topics = TEST_TOPIC)
+		public static class OtherClassLevelListenerWithSinglePublicMethod {
```
artembilan, line 186: Why do we need a second variant with exactly the same signature as `ClassLevelListenerWithSinglePublicMethod`?

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerTests.java
```diff
@@ -0,0 +1,234 @@
...
+
+			public void listen1(String in) {
+				this.latch.countDown();
+			}
+
+			private void listen2(String in) {
+				this.latch.countDown();
+			}
+
+			private void listen3(String in) {
```
artembilan, line 226: This variant does not make sense since we don't take into account private methods.

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerTests.java
```diff
@@ -0,0 +1,234 @@
...
+		// WHEN
+		this.template.send(TEST_TOPIC, "foo");
+
+		// THEN
+		assertThat(this.classLevel.latch.await(10, TimeUnit.SECONDS)).isTrue();
+		assertThat(this.classLevelListenerWithSinglePublicMethod.latch.await(10, TimeUnit.SECONDS)).isTrue();
+		assertThat(this.otherClassLevelListenerWithSinglePublicMethod.latch.await(10, TimeUnit.SECONDS)).isTrue();
+		assertThat(this.classLevelListenerWithSinglePublicMethodAndPrivateMethod.latch.await(10, TimeUnit.SECONDS)).isTrue();
+
+		assertThat(this.classLevelListenerWithDoublePublicMethod.latch.await(10, TimeUnit.SECONDS)).isFalse();
```
artembilan, line 101: This is not OK.
I'm really against any blocking in the tests for nothing.
I mean this always going to add up 10 seconds to the whole test suite execution.
Now imaging we have like 3000 tests in the project and every one would contribute some delay.

I believe the configuration when we have ambiguity with several public methods has to be reject with an `IllegalStateException`.
No silent ignoring to make am impression that my configuration is OK.

artembilan (changes requested):
This is very close to the merge phase.
Please, consider to mention this feature in the `class-level-kafkalistener.adoc` and `whats-new.adoc`.

Thanks

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerTests.java
chickenchickenlove: I wanted to verify that even if beans are modified by a BeanPostProcessor, different classes with the same conditions (including the same topic) but different names can still independently handle messages without any issues.
Personally, I was able to confirm through a test case that this scenario works as expected without problems.
To reduce the overall test execution time, I will remove this test case.

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerTests.java
chickenchickenlove: @artembilan thanks for your comments. 🙇‍♂️ 
> I believe the configuration when we have ambiguity with several public methods has to be reject with an IllegalStateException.

I understand your point clearly.
However, The important thing is that the changes I made in this PR are not related to the case where there are multiple public methods. 
So we don't need to consider ambiguous conditions because in that case any public method will not be registered as kafka listeners.
```java
if (hasClassLevelListeners) {
  if (methodsWithHandler.isEmpty() && publicMethods.size() == 1 && !hasMethodLevelListeners) {
    // added this time.
  }
  else {
    // original	
  }
}
```

I think we would be better not to throw an error if there are several public methods when there is no `@KafkaHandler` method.
This is because after registering the class as a bean using `@KafkaListener`, there might be cases where other beans inject it and call its public methods. For example, `getter` methods.
If there are already users using it that way, it would be difficult to ensure backward compatibility.

```java
@KafkaListener
class MyTakser {
  
  // This method is for @KafkaHandler implicitly.
  public void listen (String message) {
    // do something with kafka messages.
  }

  // This method is not for @KafkaHandler.
  public void getCurrentState() {
    // return current state
  }
  
  ...
}
```

IMHO, If we should something to do for this, I think it would be better to notify by using warning log.
For example, 
```
The bean {BEAN_NAME} which has @KafkaListener is registered without a proper listener method.
```

WDYT?

artembilan: I think I see the picture better now. What if our class is extended and new methods appear there? Which one to choose from hierarchy then?

I am sorry to say that , but I feel more like better to leave it as is and rely only on the `@KafkaHandler` presence .

The issue simply could be turned into docs explaining why we require this or other annotation. We really cannot predict what purpose of the method is, so being explicit is much better, than trying to come up with other restrictions. In totally ignored the fact that class could be extended 🤦.

And between us: I’m not a fun of annotations in Java. 😉

Still, that is my opinion on the problem of this issue, so I’m opened for other arguments to continue your effort with the code.

CC @sobychacko

chickenchickenlove: @artembilan 
Thank you for your detailed explanation 🙇‍♂️ 
I prefer to use explicit one as well. annotations too, because of implicit 😉.

No worries at all — I'm happy to revert all the commits I made if needed! 😄 

How about reconsidering it from this perspective?

### 1. In the case of a class with only a single public method
I believe we don't need to consider class extension in this case.
If the class was originally designed with a `single public method` intended to act as an implicit `@KafkaHandler`, it should be fine to assume that even if the class is extended, the same usage will be maintained.

If an extended class introduces a new public method, it will be ignored because there are two public methods. 
If an extended class override parent's public method for a completely different purpose, then it should be considered the user's responsibility.

### 2. In the case of a class with multiple public methods
If there are multiple public methods and no explicit `@KafkaHandler` annotation is present, no method would be registered as a listener. it is current registering logic.


Therefore, I believe we don't need to worry about this case either.
For these reasons, I think this PR still can have value. 

After you read my comments and still believe this PR is invalid, I'm happy to revert all the commits and will create commits to update `class-level-kafkalistener.adoc`! 
Please, feel free to talk your opinion.

Thanks for your time 🙇‍♂️

artembilan:
I think I prefer the fix in the: https://github.com/spring-projects/spring-kafka/pull/3870.

since we talk about the situation when we may have the annotation on the class , but no on its methods (even if it is single ), we cannot judge what is a goal of that end-user design. May be they add method annotations in the sub-classes. So, we cannot make a decision for them and some applications might be broken if we apply this PR change.
For me it is better to document how explicit `@KafkaHandler` is important. Rather then make an assumption about those single methods in the class.

this is kinda similar to what we have with `@RestController` on the class and `@RequestMapping` on the methods: https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-controller/ann-requestmapping.html

chickenchickenlove:
@artembilan 
I got your point (especially `@RestController`, `@RequestMapping`)
I agree with your opinion.  👍 
I reverted all commits and added tiny comments to `class-level-kafkalistener`.

When you have time, PTAL 🙇‍♂️

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/class-level-kafkalistener.adoc
```diff
@@ -2,6 +2,10 @@
 = `@KafkaListener` on a Class
 
 When you use `@KafkaListener` at the class-level, you must specify `@KafkaHandler` at the method level.
+If you do not use `@KafkaHandler`, no method will be used as a listener.
+
+IMPORTANT: This is because a class annotated at the class-level may be extended or inherited in various ways.
```
artembilan, line 7: While I'm not disagree with the info, it feels a bit awkward.
Even the first new sentence is confusing.

How about to have a text like this after that  "you must specify `@KafkaHandler` ":

```
If no `@KafkaHandler` on any methods of this class or its sub-classes, the framework will reject such a configuration.
The `@KafkaHandler` annotation is required for explicit and concise purpose of the method.
Otherwise it is hard to make a decision about this or other method without extra restrictions.
```

artembilan (changes requested):

artembilan:
@vitorsalves,

Please. share your opinion and we will proceed from there.
Thanks

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/class-level-kafkalistener.adoc
chickenchickenlove: Nice! thanks for fixing it. 🙇‍♂️ 
I updated.

artembilan (approved):
LGTM, but let’s wait for @vitorsalves approval since our solution is not what was requested.

vitorsalves:
@chickenchickenlove @artembilan Personally, I would prefer not to have to declare @KafkaHandler in a @KafkaListener with just one method. But I understand the point about the ambiguity that this can bring and the need to have @KafkaHandler declared.

However [#3870](https://github.com/spring-projects/spring-kafka/pull/3870) actually helps in one of the things that bothered me the most, which was the application letting you declare @KafkaListener at the class level but not making it clear that @KafkaHandler was mandatory.

Thanks for the time and effort. I really appreciate you guys taking my idea into consideration, even if there was a better solution!

PR #3870, opened 2025-04-26
GH-3869: Should fail bean registration when no method …

### LINKED ISSUE #3869: Should fail bean registration when no method listeners are registered.
**Expected Behavior**

Should fail bean registration when no method listeners are registered.

**Current Behavior**
```java
@Component
@KafkaListener
class MyListener { 

  // There is no @KafkaHandler.
  public void listen1(String message) { ... }

  // There is no @KafkaHandler.
  public void listen2(String message) { ... }
}
```

If a user registers a `@KafkaListener` as bean but there are no methods annotated with `@KafkaHandler`, 
`spring-kafka` creates `MultiMethodKafkaListenerEndpoint` without methods silently. 
Later, After `KafkaConsumer` polls records, `spring-kafka` realizes that there are no available methods and just throws a runtime error. 
In this case, user only becomes aware of the problem at runtime, even though the application is already malfunctioning.

I think It would be better to throw an `IllegalStateException` immediately when no methods annotated with `@KafkaHandler` are found so that the `KafkaListener` bean fails to be registered, following the `fast-fail` principle.

- As-Is
  - There are no actions at all
- To-Be
  - Throws an `IllegalStateException` immediately when no methods annotated with @KafkaHandler are found

**Context**

When I was writing test code for my PR, I encountered a situation where I only realized at runtime that the listener had not been properly registered.
Since I was not very familiar with this behavior, it took me some time to debug the issue.
I believe that new users of Spring Kafka might experience a similar problem.

If a `KafkaListener` is declared at the class level but no listener methods can be found,
I think it would be better to fail the Spring Boot application startup from a `fast-fail` perspective.

chickenchickenlove:
- Fixes #3869 

I created an issue and a PR regarding an improvement I thought of from a fast-fail perspective.
Please feel free to let me know if you think the changes are unnecessary.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -407,16 +407,30 @@ public class KafkaListenerAnnotationBeanPostProcessor<K, V>
 					this.logger.debug(() -> annotatedMethods.size() + " @KafkaListener methods processed on bean '"
 							+ beanName + "': " + annotatedMethods);
 				}
+				Set<Method> methodsWithHandler = MethodIntrospector.selectMethods(targetClass,
+						(ReflectionUtils.MethodFilter) method ->
+								AnnotationUtils.findAnnotation(method, KafkaHandler.class) != null);
+				boolean hasMethodLevelKafkaHandlerAnnotation = !methodsWithHandler.isEmpty();
 				if (hasClassLevelListeners) {
-					Set<Method> methodsWithHandler = MethodIntrospector.selectMethods(targetClass,
-							(ReflectionUtils.MethodFilter) method ->
-									AnnotationUtils.findAnnotation(method, KafkaHandler.class) != null);
 					List<Method> multiMethods = new ArrayList<>(methodsWithHandler);
 					processMultiMethodListeners(classLevelListeners, multiMethods, targetClass, bean, beanName);
 				}
+				throwErrorIfNoListenerMethods(bean, hasMethodLevelListeners,
+											hasClassLevelListeners, hasMethodLevelKafkaHandlerAnnotation);
 			}
 		}
 		return bean;
+	}
+
+	private void throwErrorIfNoListenerMethods(Object bean, boolean hasMethodLevelListeners,
+				boolean hasClassLevelListeners, boolean hasMethodLevelKafkaHandlerAnnotation) {
+		if (hasMethodLevelListeners) {
+			return;
+		}
+
+		if (hasClassLevelListeners && !hasMethodLevelKafkaHandlerAnnotation) {
+			throw new IllegalStateException("No kafka listener methods found on bean type: " + bean.getClass());
+		}
 	}
 
 	/*


artembilan (changes requested):
Please, supply the test for coverage .
Thanks

chickenchickenlove:
@artembilan thanks for comments even if holiday!

I added new test code to cover worst case.
other happy cases can be discovered other test codes which already exist.

When you have time, Please take a look 🙇‍♂️ 
Thanks again to look into my PR even if holiday!

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -407,18 +407,32 @@ public Object postProcessAfterInitialization(final Object bean, final String bea
...
+											hasClassLevelListeners, hasMethodLevelKafkaHandlerAnnotation);
 			}
 		}
 		return bean;
 	}
 
+	private void throwErrorIfNoListenerMethods(Object bean, boolean hasMethodLevelListeners,
+				boolean hasClassLevelListeners, boolean hasMethodLevelKafkaHandlerAnnotation) {
+		if (hasMethodLevelListeners) {
+			return;
```
artembilan, line 428: This is a bit awkward style.
Why do we need to go to this method at all, if we meet this condition?
I fully don't see a reason in this method.
The `if (!hasMethodLevelListeners && hasClassLevelListeners && !hasMethodLevelKafkaHandlerAnnotation)` and `throw new IllegalStateException` is enough to have in place.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -407,18 +407,32 @@ public Object postProcessAfterInitialization(final Object bean, final String bea
...
+		}
+
+		if (hasClassLevelListeners && !hasMethodLevelKafkaHandlerAnnotation) {
+			throw new IllegalStateException("No kafka listener methods found on bean type: " + bean.getClass());
```
artembilan, line 432: I think just `on bean` is enough, without any `getClass()` extraction.
The bean instance might lead to the place where it declared.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
chickenchickenlove: I agree with you and addressed it!
I anticipated that the conditions could become more complex because of #3865 .
So, I extracted them into a separate method to clearly separate and handle the problematic conditions.

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessorTest.java
```diff
@@ -0,0 +1,76 @@
...
+	static class TestConfig {
+
+		@Bean
+		public KafkaListenerAnnotationBeanPostProcessor<Object, Object> kafkaListenerAnnotationBeanPostProcessor() {
+			return new KafkaListenerAnnotationBeanPostProcessor<>();
+		}
+
+		@Component
+		@KafkaListener
+		static class BuggyListener {
```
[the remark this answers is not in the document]
chickenchickenlove: I didn't know that. Thanks a lot 🙇‍♂️

chickenchickenlove:
I added two `withMessage(...)` to test cases. 
Because a `BeanCreationException` may occur, but I want to explicitly show that it is caused within our test target `NoHandlerMethodListener`.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -407,13 +407,19 @@ public Object postProcessAfterInitialization(final Object bean, final String bea
...
-					Set<Method> methodsWithHandler = MethodIntrospector.selectMethods(targetClass,
-							(ReflectionUtils.MethodFilter) method ->
-									AnnotationUtils.findAnnotation(method, KafkaHandler.class) != null);
 					List<Method> multiMethods = new ArrayList<>(methodsWithHandler);
 					processMultiMethodListeners(classLevelListeners, multiMethods, targetClass, bean, beanName);
 				}
+
+				if (!hasMethodLevelListeners && hasClassLevelListeners &&
+					!hasMethodLevelKafkaHandlerAnnotation) {
+					throw new IllegalStateException("No kafka listener methods found on bean type.");
```
artembilan, line 421: Sorry, I probably was not clear enough .
I mean like this:
```suggestion
					throw new IllegalStateException("No Kafka listener methods in bean: " + bean);
```

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
artembilan: I think we should not take into a consideration that `BeanCreationException`. Might be the fact that it is not thrown upstream.
Why my suggestion does not work for you then?

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -407,13 +407,19 @@ public Object postProcessAfterInitialization(final Object bean, final String bea
 					this.logger.debug(() -> annotatedMethods.size() + " @KafkaListener methods processed on bean '"
 							+ beanName + "': " + annotatedMethods);
 				}
+				Set<Method> methodsWithHandler = MethodIntrospector.selectMethods(targetClass,
+						(ReflectionUtils.MethodFilter) method ->
+								AnnotationUtils.findAnnotation(method, KafkaHandler.class) != null);
```
artembilan, line 412: I think no need to do this extraction outside of the `if (hasClassLevelListeners) {` branch.

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessorTests.java
```diff
@@ -0,0 +1,63 @@
...
+
+	@Test
+	void ctx_should_be_fail_to_register_bean_when_no_listener_methods_exist() {
+
+		AnnotationConfigApplicationContext ctx = new AnnotationConfigApplicationContext();
+		ctx.register(TestConfig.class);
+
+		assertThatExceptionOfType(BeanCreationException.class)
+				.isThrownBy(ctx::refresh)
+				.withMessageContaining("No kafka listener methods found on bean type.")
```
artembilan, line 44: We also can check for our `IllegalStateException` as a cause.

artembilan (changes requested):

PR #3874, opened 2025-04-27
GH-3067: Spring Kafka support multiple headers with same key.

### LINKED ISSUE #3067: Spring Kafka does not allow multiple headers with same key

chickenchickenlove:
- Fixes #3067 

I know that there is an existing PR(https://github.com/spring-projects/spring-kafka/pull/3101) related to this issue, but it seems to have been suspended for long time. 
Therefore, I created a new PR to address it.
If you are planning to maintain the existing one, I am happy to close mine.

I approached the problem differently from the existing PR and proposed an alternative solution.
For backward compatibility, user can configure `MultiValueKafkaHeaderMapper` to `KafkaListenerContainerFactory`.
For example, 
```java
ConcurrentKafkaListenerContainerFactory<Integer, String> factory =

MultiValueKafkaHeaderMapper headerMapper = new MultiValueKafkaHeaderMapper();

// Create new header mapper.
headerMapper.addSingleValueHeader(SINGLE_VALUE_HEADER);
MessagingMessageConverter converter = new MessagingMessageConverter(headerMapper);

// Set header mapper.
```


sobychacko:
@chickenchickenlove We were planning to resuscitate the other PR before the next milestone (`4.0.0-M3`) along with some other refactorings, so please stay tuned for that. Then we can compare your approach in this PR, but I will try first to rebase that PR on top of the latest main and see how it goes.

chickenchickenlove:
@sobychacko Thanks for your time!! 
I didn't get your point clearly.
From what I understand, this PR is no longer needed since you will be taking on a new task.
Could you please confirm if my understanding is correct?

sobychacko:
Yes, we will need to start from the other PR first. But we will keep this open to compare.

spring-kafka/src/main/java/org/springframework/kafka/support/MultiValueKafkaHeaderMapper.java
```diff
@@ -0,0 +1,129 @@
...
+ * This Header Mapper manages header values as a list,
+ * except for certain reserved headers.
+ * Other behaviors are identical to {@link DefaultKafkaHeaderMapper}.
+ *
+ * @author Sanghyeok An
+ *
+ * @since 4.0.0
+ *
+ */
+public class MultiValueKafkaHeaderMapper extends DefaultKafkaHeaderMapper {
```
artembilan, line 40: I don't think that I agree about extra abstraction on the matter.
I think the existing `HeaderMapper` must support iterables for end-user headers.
So, we should expect some pattern for those custom headers which we can treat as iterable.
And I think that one should be on the `in`  (from Kafka) only.
If we produce and some header is a collection, then populate it into Kafka headers as is.
On the other side, for convenience, we still should support first value mapping by default.
And map only those to collection values, which we have opted-in.

Make sense?

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/MultiValueKafkaHeaderMapper.java
chickenchickenlove: @artembilan 
There are parts that I understand and parts that I don't.
It's been a while since I last worked with `Kafka` in my work, so I might be misunderstanding some aspects.
Please keep that in mind as you read this.

How can `Spring Kafka` anticipate a `custom header pattern` that is iterable?
Even if `Spring Kafka` expects a `custom header` to be iterable, users might always expect the header to be a single value.
What I mean is, users may have already written their code accordingly, like `@Header("blahblah") String value`.

To handle such cases, I think it would be necessary to add a public method that allows users to configure the pattern for iterable custom headers as a workaround.

If the headers that should be treated as iterable are managed with a whitelist approach,
I believe `DefaultKafkaHeaderMapper` could support iterable headers without requiring additional abstraction as you mentioned before.


I believe that by moving part of the logic from the `MultiValueKafkaHeaderMapper` class which I implement to `DefaultKafkaHeaderMapper` and simply adding a `whitelist` field,
spring kafka can support multi-header values while maintaining backward compatibility.

Overall, it makes sense to me 👍

artembilan: I believe the `@Header("blahblah") String value` should work. That annotation processor is able to extract single value from an iterable header.

Yes, I would expect some configuration property on the `DefaultKafkaHeaderMapper` to opt-in to those Kafka headers which have to be mapped to `Message` header as `List`.
We cannot use `white` and `black` words though. So, we need to come up with other name. Like `headerAsListPattern`. Or even vararg `String... headerAsListPatterns`. Where pattern has to follow `org.springframework.util.PatternMatchUtils` expectations.

On the mapping from `Message` to Kafka there is no need to think about pattern.
The trigger is simple: if `Message` is an iterable, then we put it into multi-header in Kafka record.

chickenchickenlove: Nice. 
Thanks for your comments. They are always a great help to my learning.
I’ll be keeping an eye on your PRs to learn from your code. 🙇‍♂️ 
Rr, If you’re too busy with other tasks and don’t have the capacity to handle this, please feel free to let me know.

I have one question.
Could you tell me why the terms `white` and `black` cannot be used?
If this is a convention not allowed in Spring contributions, I’d like to keep it in mind for any future PRs I submit.

artembilan: Because they are related to racism and our team follows ethic rules to use only neutral official language.
There are many other words we have to avoid: `execute`, `kill`, `abort`, `master`, `slave` etc.
I even prefer to not use `foo` and `bar`, since this is not professional language.

We do have a capacity to review your PR.
This is long-standing not critical task, so we don't have a priority to work on it.
But contribution is always welcome!

sobychacko:
Now that we are going ahead with this PR, it is better to close the [other one](https://github.com/spring-projects/spring-kafka/pull/3101) in favor of this one.

spring-kafka/src/main/java/org/springframework/kafka/support/MultiValueKafkaHeaderMapper.java
chickenchickenlove: Oh, I see, My bad.
Because I'm not familiar that kinds of topic. Because our country is made up of a single race, ethnicity.
Thanks for your comments! I'm always going to keep in mind. 

Anyway, I will take a crack ASAP.

chickenchickenlove:
Hi, @artembilan , @sobychacko . 
I made an draft commit following artembilan comment! 
I want to get advise from you guys. 🙇‍♂️ 

However, header is not being formatted from bytes to the desired type at the time the method is invoked.
Because [AbstractNamedValueMethodArgumentResolver](https://github.com/spring-projects/spring-framework/blob/f8e7bf782503a76fd77618a1188d19d9bb5102e8/spring-messaging/src/main/java/org/springframework/messaging/handler/annotation/support/AbstractNamedValueMethodArgumentResolver.java#L115-L126) cannot handle it properly. 
because the If condition below is `False`

```java
if (parameter != nestedParameter || !ClassUtils.isAssignableValue(parameter.getParameterType(), arg)) { 
    ...
}
```

So, end user always get raw headers with `byte[]`. 
Should we implement `HandlerMethodArgumentResolver` or `AbstractNamedValueMethodArgumentResolver` to resolve `@Header List<String>`, `@Headers Map<String, Object>`? 
If you have any good idea, please let me know.

artembilan:
> This is because the type of myHeader is List. so !ClassUtils.isAssignableValue(parameter.getParameterType(), arg) is False.

I think you are mixing concerns in this PR.
The `@KafkaListener` with its `@Header` support is just one of the users of `HeaderMapper` abstraction result.

We have to concentrate on the `HeaderMapper` logic implementation here.
The argument resolver is different story and should be revised (if any) when we done here.

If you do provide a list header with `byte[]` as values, then it is not a listener method responsibility to convert them to typed list.
We really never advertised that we will be able to convert items in the list to expected generic type.

chickenchickenlove:
@artembilan 
Thank you for pointing it out. It's a bit clearer now.  😄 
I implemented both `DefaultKafkaHeaderMapper` and  `SimpleKafkaHeaderMapper` to handle inbound multi-value headers.

Also, I created two integration test classes to demonstrate that these header mappers can handle in/outbound multi-value headers while still respecting single-value headers as well.

When you have time, Please take another look. 🙇‍♂️

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -79,6 +80,12 @@ public abstract class AbstractKafkaHeaderMapper implements KafkaHeaderMapper {
 
 	private Charset charset = StandardCharsets.UTF_8;
 
+	private final List<HeaderMatcher> matchersForListValue = new ArrayList<>();
+
+	private final Set<String> cachedHeadersForListValue = new LinkedHashSet<>();
+
+	private final Set<String> cachedHeadersForSingleValue = new LinkedHashSet<>();
```
chickenchickenlove, line 87: inline comment; 

I added two caches for single-value and multi-value to optimize performance. 
I assume that headers in `ConsumerRecord` will have lower cardinality. 
From this assume, we can save our CPU usage.

artembilan: The idea is OK, but not perfect . It might be good to compare some performance and make sure it worth to sacrifice memory for these caches.

another optimization could be with eviction policy to avoid the case of “too many headers” and don’t keep too old headers in the cache.

with that in mind , I would suggest to consider caching as a separate bonus task.
It doesn’t feel crucial for the big picture in hands.

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -97,6 +104,20 @@ public AbstractKafkaHeaderMapper(String... patterns) {
...
 	protected AbstractKafkaHeaderMapper(boolean outbound, String... patterns) {
+		this(outbound, new ArrayList<>(), patterns);
+	}
+
+	/**
+	 * Construct a mapper that will match the supplied patterns (outbound) and all headers
+	 * (inbound). For outbound mapping, certain internal framework headers are never
+	 * mapped. For inbound mapping, Headers that match the pattern specified in
+	 * {@code patternsForListValue} will be appended to the values under the same key.
+	 *
```
artembilan, line 115: No blank lines in method Javadocs, please.

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -97,6 +104,20 @@ public AbstractKafkaHeaderMapper(String... patterns) {
...
+	 * @param outbound true for an outbound mapper.
+	 * @param patternsForListValue the patterns for multiple values at the same key.
```
artembilan, line 117: In terms of Apache Kafka, it is a multi-value headers. Can we follow that style instead ?
I mean the Kafka requirements first and we should map to its concepts even if internally it is a `List` for us. 
so, my suggestion is to rename this property to something like `multiValuePatterns`.
Remember, this is an Open Source? So, code is visible and has to be as clear as possible?

Plus, why bother constructor?
This is an optional feature, so has to be just a setter .
My requirements are to expose via constructor, if we cannot live without that property.
This way the setter for new property could be a varargs even if you set into a `Set` internally .

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -287,6 +313,34 @@ private String mapRawIn(String header, byte[] value) {
 		return null;
 	}
 
+	/**
+	 * Check whether the header value should be mapped to multiple values.
+	 * @param headerName the header name.
+	 * @return True for multiple values at the same key.
+	 */
+	protected boolean isHeaderForListValue(String headerName) {
```
artembilan, line 321: May be `mapAsMultiValue()`?
@sobychacko , what’s your idea for naming here? 🥹

artembilan (changes requested):
Some review so far I could pull from the phone. Will come back on Monday!
Thank you for doing all of this, @chickenchickenlove !

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -191,6 +193,16 @@ public void addRawMappedHeader(String name, boolean toString) {
 		this.rawMappedHeaders.put(name, toString);
 	}
 
+	/**
+	 * Add patterns for matching multi-value headers under the same key.
+	 * @param patterns the patterns for header.
+	 */
+	public void addHeaderPatternsForMultiValue(String ... patterns) {
```
artembilan, line 200: This is not correct name from Java Beans specification perspective.
Really has to be `setMultiValueHeaderPatterns`.

Not sure why you use `for` over there.

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -191,6 +193,16 @@ public void addRawMappedHeader(String name, boolean toString) {
...
+		for (String pattern : patterns) {
+			this.headerMatchersForMultiValue.add(new SimplePatternBasedHeaderMatcher(pattern));
```
artembilan, line 202: I believe we can use Java `Stream` API and then `addAll` on the targer collection.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -324,12 +326,37 @@ else if (!(headerName.equals(JSON_TYPES)) && matchesForInbound(headerName)) {
...
 	}
 
+	/**
+	 * Handle non-reserved headers in {@link DefaultKafkaHeaderMapper}.
+	 * @param headerName the header name.
+	 * @param header the header instance.
+	 * @param headers the target headers.
+	 * @since 4.0.0
+	 */
+	protected void handleHeader(String headerName, Header header, final Map<String, Object> headers) {
```
artembilan, line 342: This is mapping from Kafka headers, So, I believe this name is better in this context: `fromUserHeader`.
Since standard headers have been already mapped before.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -324,12 +326,37 @@ else if (!(headerName.equals(JSON_TYPES)) && matchesForInbound(headerName)) {
...
+		if (!this.doesMatchMultiValueHeader(headerName)) {
```
artembilan, line 343: No `this.` prefix for method calls.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -324,12 +326,37 @@ else if (!(headerName.equals(JSON_TYPES)) && matchesForInbound(headerName)) {
...
+			headers.put(headerName, headerValueToAddIn(header));
+		}
+		else {
+			Object values = headers.getOrDefault(headerName, new ArrayList<>());
+			if (values instanceof List) {
```
artembilan, line 348: I'm not fully sure in this logic.
Right now you mean: if there is a header with requested name and it is not a list, then we override its value.
Is it really expected?
Why we just cannot always override withe list since this is exactly what is asked by the respective pattern?
Otherwise I feel like we cannot override since the value is already there.

spring-kafka/src/test/java/org/springframework/kafka/support/DefaultKafkaHeaderMapperForMultiValueTests.java
```diff
@@ -0,0 +1,449 @@
...
+ *
+ * @since 4.0.0
+ */
+
+@SpringJUnitConfig
+@DirtiesContext
+@EmbeddedKafka(partitions = 1, topics = {
+		DefaultKafkaHeaderMapperForMultiValueTests.TEST_TOPIC,
+		DefaultKafkaHeaderMapperForMultiValueTests.RETRY_TOPIC
+})
```
artembilan, line 79: Why do we need fully-blown Kafka broker environment, since we just would like to test `DefaultKafkaHeaderMapper`?

spring-kafka/src/test/java/org/springframework/kafka/support/DefaultKafkaHeaderMapperForMultiValueTests.java
```diff
@@ -0,0 +1,449 @@
...
+					if (isHeadersExpected(headers, headerName)) {
+						latchForSingleValueHeadersInRetry.countDown();
+					}
+				}
+				throw new RuntimeException("Woooops... in topic " + message);
+			}
+
+		}
+
+		static class MyCustomDltProcessor {
```
artembilan, line 409: Why do we need to over-complicate the logic of the test with retryable topics?

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/support/DefaultKafkaHeaderMapperForMultiValueTests.java
chickenchickenlove: @artembilan 

Thanks for your comment!
I believe that the changes to the `HeaderMapper` have a significant impact on many parts of the system.
The integration tests I’ve written for multi-value headers cover the following:

1. Verifying that multi-value headers are correctly resolved on the inbound side, properly sent on the outbound side, and can be successfully received.
2. Ensuring there are no regressions for headers expected to have single values. In scenarios involving retries and DLT, additional reserved headers appear, so it’s important to verify that their values remain intact.

If you think that these test cases are too over engineering, how about revising it as follows?

1. Add simple unit tests.
2. Keep a simple integration test for inbound/outbound (e.g., `DefaultKafkaHeaderMapperTests#testForCommonCase`, `SimpleKafkaHeaderMapperTests#testForCommonCase`).

WDYT? Please let me know your opinion. 🙇‍♂️

spring-kafka/src/test/java/org/springframework/kafka/support/DefaultKafkaHeaderMapperForMultiValueTests.java
chickenchickenlove: Please refer to this comments! (https://github.com/spring-projects/spring-kafka/pull/3874#discussion_r2074443025)

artembilan: Why existing tests are not enough to ensure no regression for single-value headers ?

I worry more about complexity of the test to digest and it support in the future. Plus, all these integration tests contribute to the whole project build performance.

I would prefer to be reasonable: unit test just for the task in hands. And modification of some existing integration tests via adding multi-value mapping. When you try that, though, you will feel my struggle about complex tests 😜

chickenchickenlove: Okay! Thanks for your comments! 
I understand your concerns.
I’ll think about how to integrate it into the existing integration tests and revise the code again.
Thank you as always. 🙇‍♂️

chickenchickenlove:
@artembilan 
### 1
I didn’t find proper integration test cases to verify the multiple-header feature 
Because they would pollute the original meaning of the integration tests and might confuse developers. 
So I decided to add unit tests! 

### 2
Also, I don't implement multi value headers for both `DefaultKafkaHeaderMapper.fromHeaders(...)` and `SimpleKafkaHeaderMapper.fromHeaders(...)`. 
Because, the class `Message<?>` and `MessageHeaders` don't support multiple header under same key and method `fromHeaders(...)` are only called when parameter type is `MessageHeaders`. 

So, I think it makes sense to leave it as it is or just add documentation for the `Message<>` case in the Spring Kafka adocs!

When you have time, please take another look. 🙇‍♂️

artembilan:
> Because they would pollute the original meaning of the integration tests and might confuse developers

isn’t that from where I’ve started composing about your complex test? 🤪

> Because, the class Message<?> and MessageHeaders don't support multiple header under same key

I Thought we discussed that before : we map to list from ConsumerRecord headers. And create multi-value header in ProducerRecord from list. It might be the case that this already supported. Will take a look closer tommorow.

chickenchickenlove:
> I Thought we discussed that before : we map to list from ConsumerRecord headers. And create multi-value header in ProducerRecord from list. It might be the case that this already supported. Will take a look closer tommorow.

Some code try to map `Message<?>` to `ProducerRecord`.  
For example, `MessagingMessageConverter.fromMessage(...)`. 

https://github.com/spring-projects/spring-kafka/blob/529b5d0331dc0872bf96db74119e0c9b5bad68e4/spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java#L226-L234

That method got `Message<?>` as its parameter. 

https://github.com/spring-projects/spring-kafka/blob/529b5d0331dc0872bf96db74119e0c9b5bad68e4/spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java#L254-L259

Then, it attempts to map headers and build `ProducerRecord`. 
In this case, we cannot do anything because `RecordHeader` in Kafka package has `List<Header>` as delegate and `MessageHeaders` implement `Map<String, Object>`. 


So `HeaderMapper.fromMessage(...)` cannot map `MessageHeaders` in `spring` to `Headers` in `kafka` in case of multi value headers. 

Fortunately, In terms of framework internals, when `KafkaListener` returns `CompletableFuture<Message<?>>`, `HeaderMapper.fromHeaders(...)` seems to be called. 
By the way, `ReactiveKafkaProducerTemplate` and `ReplyingKafkaTemplate` don't support `ProducerRecord`. 
Therefore, in terms of `ReactiveKafkaProducerTemplate` and `ReplyingKafkaTemplate` don't support multi-value headers because of limitation of `Message<?>` in `spring`.

I think the easy way to address it is that provide new public method supporting `ProducerRecord` to `ReactiveKafkaProducerTemplate` and `ReplyingKafkaTemplate`. 

Or, we can suggest new implementation of `Message` and `MessageHeaders` in spring.

If I missed something, please let me know! 🙇‍♂️ 
WDYT?

artembilan:
> In this case, we cannot do anything because RecordHeader in Kafka package has List<Header> as delegate and MessageHeaders implement Map<String, Object>.

I don't understand what you are talking about.
The API is exactly what we need:
```
    /**
     * Creates and adds a header, to the end, returning if the operation succeeded.
     *
     * @param key of the header to be added.
     * @param value of the header to be added.
     * @return this instance of the Headers, once the header is added.
     * @throws IllegalStateException is thrown if headers are in a read-only state.
     */
    Headers add(String key, byte[] value) throws IllegalStateException;
```

Now if the value in the `MessageHeaders` is an `Iterable` we just have to iterate it in the `DefaultKafkaHeaderMapper.fromHeaders()`:
```
		headers.forEach((key, value) -> {
			if (matches(key, value)) {
				Iterable<?> valuesToMap;
				if (value instanceof Iterable<?> iterable) {
					valuesToMap = iterable;
				}
				else {
					valuesToMap = List.of(value);
				}
				for (Object rawValue : valuesToMap) {
					Object valueToAdd = headerValueToAddOut(key, rawValue);
					if (valueToAdd instanceof byte[]) {
						target.add(new RecordHeader(key, (byte[]) valueToAdd));
					}
					else {
						try {
							String className = valueToAdd.getClass().getName();
							boolean encodeToJson = this.encodeStrings;
							if (this.toStringClasses.contains(className)) {
								valueToAdd = valueToAdd.toString();
								className = JAVA_LANG_STRING;
								encodeToJson = true;
							}
							if (!encodeToJson && valueToAdd instanceof String) {
								target.add(new RecordHeader(key, ((String) valueToAdd).getBytes(getCharset())));
							}
							else {
								target.add(new RecordHeader(key, headerObjectMapper.writeValueAsBytes(valueToAdd)));
							}
							jsonHeaders.put(key, className);
						}
						catch (Exception e) {
							logger.error(e, 
									() -> "Could not map " + key + " with type " + rawValue.getClass().getName());
						}
					}
				}
			}
		});
```

What do I miss?

artembilan:
Yes. The code you show is what I have in mind.
Why you still talking about multi-value headers in the `Message`?
Yes, we had that discussion in the begging when you had started looking into this.

So, let me re-iterate it again!
Since we cannot have multi-value headers in the `Message`, we can suggest to treat a single entry in the map as a `Iterable`.
When we send a message to Kafka, we covert it into a `ProducerRecord`. And its `Headers` structure allows multi-value entries.
Apparently that is not too hard to convert a single `Message`'s `Iterable` header into a multi-value in the `ProducerRecord`.
Doesn't look like we need some special pattern matching treatment. The `Iterable` is enough trigger to determine that user would like to have a multi-value header in the Kafka record.

Now the consumer side.
Since all the `ConsumerHeaders` are really multi-value, we cannot assume which one is for single value and which has to be treated as multi-value.
I mean one record comes with a single entry for a `key`, but the next record has like three `key` headers.
Therefore, for consistency and easy to digest those values downstream I suggest a pattern matching for consumer headers.
This way, end-user would opt-in what headers they would like to have as a `List` in their application.
If not opt-in, behavior as right now: we map only the first entry for multi-value header.

The `HeaderMapper` has two contracts: `fromHeaders` - when we send data to Kafka.
And `toHeaders` when we receive data from Kafka.

So, why do you try to find difference between `MessageHeaders` and `Headers` interfaces?

chickenchickenlove:
@artembilan 

Thanks for your detail explanation! 🙇‍♂️🙇‍♂️🙇‍♂️🙇‍♂️🙇‍♂️ 
It was very helpful for me to understand our assumptions!

IMHO, We don't iterate Iterable headers in `fromHeaders(...)`. 
In `spring-kafka` test case, counterexample exists. 😓 

https://github.com/spring-projects/spring-kafka/blob/529b5d0331dc0872bf96db74119e0c9b5bad68e4/spring-kafka/src/test/java/org/springframework/kafka/support/DefaultKafkaHeaderMapperTests.java#L170-L185

Because such counterexamples exist, `Spring Kafka` users may be using `Iterable` values in `Message<?>` instances. Therefore, IMHO, `fromHeader()` should consider the same opt-in mechanism that `toHeader()` uses.

I have updated commit for it.
What are your thoughts?

artembilan:
> Therefore, IMHO, fromHeader() should consider the same opt-in mechanism that toHeader() uses.

That's sounds like legit.
According to that `DefaultKafkaHeaderMapperTests.testMimeTypeInHeaders()` we do support iterables even right now, but we serialize them to JSON.
And then back from JSON on the consumer side.
However that ends up just as a single Kafka record header.
Which is, essentially, not what is expected by this issue.

So, yeah, even if we support right now via JSON, that doesn't mean it works for every use-case, especially if consumer side of Kafka topic is not Spring application.
May we consider then apply pattern matching for producer side  as well?

chickenchickenlove:
@artembilan 
Sure!
I made a new commit and added test code.
Please refer to the code snippet below.
This is to implement pattern matching on the producer side! 

https://github.com/chickenchickenlove/spring-kafka/blob/13267dc32281608ea8b65852ff591b1575ee0be4/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java#L269-L282

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -251,6 +264,40 @@ protected Object headerValueToAddOut(String key, Object value) {
 		return valueToAdd;
 	}
 
+	/**
+	 * Check whether the header value should be mapped to multiple values.
+	 * @param headerName the header name.
+	 * @return True for multiple values at the same key.
+	 */
```
artembilan, line 271: `@since 4.0`

spring-kafka/src/test/java/org/springframework/kafka/support/DefaultKafkaHeaderMapperTests.java
```diff
@@ -33,6 +33,7 @@
 import org.junit.jupiter.api.Test;
 
 import org.springframework.core.log.LogAccessor;
+import org.springframework.kafka.retrytopic.RetryTopicHeaders;
```
artembilan, line 36: Why do we need this type in this unit test?

artembilan (changes requested):

artembilan:
I’ll review the rest tomorrow from big screen.
Thanks.

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -251,6 +264,40 @@ protected Object headerValueToAddOut(String key, Object value) {
...
+	protected void fromUserHeader(String headerName, Header header, final Map<String, Object> headers) {
+		if (!doesMatchMultiValueHeader(headerName)) {
+			headers.put(headerName, headerValueToAddIn(header));
+		}
+		else {
+			@SuppressWarnings("unchecked")
+			List<Object> headerValues = (List<Object>)
+					headers.computeIfAbsent(headerName, key -> new ArrayList<>());
+			headerValues.add(headerValueToAddIn(header));
+			headers.put(headerName, headerValues);
```
artembilan, line 297: You just did `computeIfAbsent`. why do we need do this `put` yet?

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -266,30 +267,18 @@ public void fromHeaders(MessageHeaders headers, Headers target) {
...
-						jsonHeaders.put(key, className);
+				if (doesMatchMultiValueHeader(key)) {
+					Iterable<?> valuesToMap;
+					if (rawValue instanceof Iterable<?> iterable) {
+						valuesToMap = iterable;
 					}
-					catch (Exception e) {
-						logger.error(e, () -> "Could not map " + key + " with type " + rawValue.getClass().getName());
+					else {
+						valuesToMap = List.of(rawValue);
```
artembilan, line 276: Now, since we are very close to the final solution, let's talk about optimization!
This `fromHaders` is very active action performed on each message to produce to Kafka.
Therefore, any extra object is performance degradation.
I would prefer to not have this `List.of()` for a single value.
Even if `if..else` would have some duplication, it is still would be better by performance.
So, let' do `forEach` if `Iterable`, and just one for a single-value header!

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -324,12 +313,42 @@ else if (!(headerName.equals(JSON_TYPES)) && matchesForInbound(headerName)) {
...
+					className = JAVA_LANG_STRING;
+					encodeToJson = true;
+				}
+				if (!encodeToJson && valueToAdd instanceof String) {
+					target.add(new RecordHeader(key, ((String) valueToAdd).getBytes(getCharset())));
+				}
+				else {
+					target.add(new RecordHeader(key, headerObjectMapper.writeValueAsBytes(valueToAdd)));
+				}
+				jsonHeaders.put(key, className);
```
artembilan, line 344: So, if original header is an iterable we call this `fromHeader` that many times.
While `target.add(new RecordHeader())` is OK , because Kafka supports multi-value headers. but this `jsonHeaders` is a single-value map, so we override the entry every time we iterate.
Might be OK to have only one type per iterabel, but can we optimize it a little bit with a `putIfAbsent()`?
And probably document somewhere that all the items of the iterable header must be of the same type.

spring-kafka/src/main/java/org/springframework/kafka/support/SimpleKafkaHeaderMapper.java
```diff
@@ -94,27 +96,43 @@ public static SimpleKafkaHeaderMapper forInboundOnlyWithMatchers(String... patte
...
-				Object valueToAdd = headerValueToAddOut(key, value);
-				if (valueToAdd instanceof byte[] && matches(key, valueToAdd)) {
-					target.add(new RecordHeader(key, (byte[]) valueToAdd));
+				if (doesMatchMultiValueHeader(key)) {
+					Iterable<?> valuesToMap;
+					if (value instanceof Iterable<?> iterable) {
+						valuesToMap = iterable;
+					}
+					else {
+						valuesToMap = List.of(value);
```
artembilan, line 105: DITTO

artembilan (changes requested):
After you fix the latest review feedback, please, consider to document a new functionality, including `What's New` entry.
Thanks

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/headers.adoc
```diff
@@ -179,3 +179,27 @@ MessagingMessageConverter converter() {
...
 If using Spring Boot, it will auto configure this converter bean into the auto-configured `KafkaTemplate`; otherwise you should add this converter to the template.
 
+[[multi-value-header]]
+== Support multi-value header
+
+Spring for Apache Kafka 4.0 supports multi-value header.
+To use multi-value header, the patterns for multi-value should be configured to `HeaderMapper`.
+If no patterns are configured to `HeaderMapper`, iterable header values under same header name will be handled like single value.
+
+If there are use cases in which multi-value headers are required, use this:
```
artembilan, line 189: The `use this` is a command for specific person.
That is not a technical documentation sentence.

Either way, I don't find the text here as useful as it could be.
I would expect an explanation about multi-value headers in Kafka and then how their are handled by default and so on.
Then other other hand, there should be an explanation how iterable message headers are processed now and what we do if there is respective pattern matching.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/headers.adoc
```diff
@@ -179,3 +179,27 @@ MessagingMessageConverter converter() {
...
+
+[source, java]
+----
+@Bean
+public void KafkaListenerContainerFactory<Integer, String> containerFactory() {
+    ConcurrentKafkaListenerContainerFactory<Integer, String> factory =
```
artembilan, line 195: We use header mapper not only in the listener container.
And such a factory is not only one way to configure that.
Therefore I would expect to not have such a misleading bean definition in this docs section.
More over, I would say there is no need in the code snippet just for a `setMultiValueHeaderPatterns()` at all.
Better to explain what are those patterns and give examples of possible values.
Now, you provide only whole header names.
What if I would like to have both mentioned headers mapped with one pattern?
I guess something like `test-multi-*` would be enough, but I'm not sure.
WDYT?

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultKafkaHeaderMapper.java
```diff
@@ -324,12 +311,42 @@ else if (!(headerName.equals(JSON_TYPES)) && matchesForInbound(headerName)) {
...
+				if (this.toStringClasses.contains(className)) {
+					valueToAdd = valueToAdd.toString();
+					className = JAVA_LANG_STRING;
+					encodeToJson = true;
+				}
+				if (!encodeToJson && valueToAdd instanceof String) {
+					target.add(new RecordHeader(key, ((String) valueToAdd).getBytes(getCharset())));
+				}
+				else {
+					target.add(new RecordHeader(key, headerObjectMapper.writeValueAsBytes(valueToAdd)));
```
artembilan, line 340: Looks like we can move `target.add(new RecordHeader(key, ` out of `if..else`.
And have there only `value` calculated.

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -70,3 +70,9 @@ Several deprecated items have been removed:
 
 Spring for Apache Kafka 4.0 supports Kafka 4.0’s new consumer rebalance protocol - https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol[KIP-848].
 For details, see xref:kafka/receiving-messages/rebalance-listeners.adoc#new-rebalalcne-protocol[New Consumer Rebalace Protocol docs].
+
+[[x40-multi-value-header]]
+=== Support multi-value header
+
+Spring for Apache Kafka 4.0 supports multi-value header for Kafka Record.
+For details, see xref:kafka/headers.adoc#multi-value-header[Support multi-value header].
```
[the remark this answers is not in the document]
chickenchickenlove: A journey of a thousand miles begins with a single step. 😃 
I modified 'see' to 'More details are available in ...'.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/headers.adoc
chickenchickenlove: I agree. 
I also added note section following java doc of `PatternMatchUtils.simpleMatch(...)`.
```
Regular expressions are *not* supported; only the `*` wildcard is allowed in simple patterns—supporting direct equality and forms such as:

- `xxx*`
- `*xxx`
- `*xxx*`
- `xxx*yyy`
```

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -70,3 +70,9 @@ Several deprecated items have been removed:
 
 Spring for Apache Kafka 4.0 supports Kafka 4.0’s new consumer rebalance protocol - https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol[KIP-848].
 For details, see xref:kafka/receiving-messages/rebalance-listeners.adoc#new-rebalalcne-protocol[New Consumer Rebalace Protocol docs].
+
+[[x40-multi-value-header]]
+=== Support multi-value header
+
+Spring for Apache Kafka 4.0 supports multi-value header for Kafka Record.
```
artembilan, line 77: We don’t need to repeat version since the whole chapter is about 4.0 news. Better to say what class(es) got a new option and then link to the target chapter.

spring-kafka/src/test/java/org/springframework/kafka/support/DefaultKafkaHeaderMapperTests.java
```diff
@@ -332,6 +334,115 @@ void inboundJson() {
...
+
+		@SuppressWarnings("unchecked")
+		List<byte[]> multiValueWildeCardHeader1Values = (List<byte[]>) mappedHeaders.get(multiValueWildeCardHeader1);
+		assertThat(multiValueWildeCardHeader1Values).contains(new byte[] { 0, 0, 0, 6 },
+															new byte[] { 0, 0, 0, 7 });
+
+		@SuppressWarnings("unchecked")
+		List<byte[]> multiValueWildeCardHeader2Values = (List<byte[]>) mappedHeaders.get(multiValueWildeCardHeader2);
+		assertThat(multiValueWildeCardHeader2Values).contains(new byte[] { 0, 0, 0, 8 },
+															new byte[] { 0, 0, 0, 9 });
```
artembilan, line 398: Can we extract all of these “data under test” to variables?
I think the code would be easier to read.

spring-kafka/src/test/java/org/springframework/kafka/support/DefaultKafkaHeaderMapperTests.java
```diff
@@ -332,6 +334,115 @@ void inboundJson() {
...
+		mapper.setMultiValueHeaderPatterns("test-multi-*",
+											multiValueHeader3,
+											"*-prefix-match-multi*");
+
+		// WHEN
+		Headers results = new RecordHeaders();
+		mapper.fromHeaders(message.getHeaders(), results);
+
+		// THEN
+		assertThat(results).contains(
```
artembilan, line 433: I think we can utilize `RecordHeaders` API for much cleaner assertion. This `contains` is just not readable.

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/headers.adoc
```diff
@@ -179,3 +179,61 @@ MessagingMessageConverter converter() {
...
 If using Spring Boot, it will auto configure this converter bean into the auto-configured `KafkaTemplate`; otherwise you should add this converter to the template.
 
+[[multi-value-header]]
+== Support multi-value header
+
+Spring for Apache Kafka 4.0 supports multi-value header where same logical header key appears more than once in a Kafka record.
+
+By default, when multiple headers share the same name, the `HeaderMapper` treats them as a single value and serialises the collection to JSON.
+
+* **Producer side:** `DefaultKafkaHeaderMapper` writes the JSON bytes, while `SimpleKafkaHeaderMapper` ignore it.
```
sobychacko, line 189: I wonder whether we should show an example of how a producer application adds multi headers. For ex. when reading the docs, it might be beneficial, if I can see how multi headers are produced. What do you think?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/headers.adoc
```diff
@@ -179,3 +179,61 @@ MessagingMessageConverter converter() {
...
+* **Consumer side:** the mapper exposes the header as a single value—the **last occurrence wins**; earlier duplicates are silently discarded.
```
sobychacko, line 190: Same comment as above for the producer. A quick example maybe?

sobychacko (changes requested):

sobychacko:
@chickenchickenlove Great job on the PR. I have one question related to some minor PR comments I added above: Do you think we should also add an end-to-end integration test (using `EmbeddedKafka`) where the multi-headers are propagated from the producer side to the consumer side?

chickenchickenlove:
@sobychacko 

Thanks for your comments. 🙇‍♂️
I prefer to add integration test cases for any features I develop.
I'm not perfect, but I believe that robust test code contributes to the overall stability of spring-kafka and helps catch my mistakes.

However, since this is an open-source framework, there's a dilemma: adding too many integration tests could affect the overall test execution time. This might slow down the overall development speed of spring-kafka. 

So, I'm not sure if adding integration tests for this feature is good approach for the project. 
If you would like me to add some integration test for simple case, I'd be happy to do so.
Just let me know. 

FYI, in an earlier commit, I added a complicated integration test case instead of a proper unit test. 😓 
(you can refer to this commit https://github.com/spring-projects/spring-kafka/pull/3874/commits/065bedc3ee6ca1501b2f0a6ef999c80ad26df886#diff-faf64a7271386c5e4f0f289951773b7db66bb1546642cab5213b546466d90765).

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/headers.adoc
```diff
@@ -179,3 +179,81 @@ MessagingMessageConverter converter() {
 
 If using Spring Boot, it will auto configure this converter bean into the auto-configured `KafkaTemplate`; otherwise you should add this converter to the template.
 
+[[multi-value-header]]
+== Support multi-value header
+
+Spring for Apache Kafka 4.0 supports multi-value header where same logical header key appears more than once in a Kafka record.
```
artembilan, line 185: Usually we say `Starting with 4.0,`. Otherwise it sounds like in `5.0` there won’t be such a support 😜.
Plus no need to mention our project name in every sentence since it is kinda obvious that this doc is about Spring for Apache Kafka.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/headers.adoc
```diff
@@ -179,3 +179,81 @@ MessagingMessageConverter converter() {
...
+
+By default, when multiple headers share the same name, the `HeaderMapper` treats them as a single value and serialises the collection to JSON.
```
artembilan, line 187: This sentence does not reflect reality. The message headers abstraction doesn’t have multi-value headers notion.
You probably talk about a collection header which just serialized to a single Kafka header as JSON.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/headers.adoc
```diff
@@ -179,3 +179,81 @@ MessagingMessageConverter converter() {
...
+mapper.setMultiValueHeaderPatterns("test-multi-value1", "test-multi-value2");
+
+// Wildcard patterns for test-multi-value1, test-multi-value2
+mapper.setMultiValueHeaderPatterns("test-multi-*");
+----
+
+Any header whose name matches one of the supplied patterns is
+
+* **Producer side:** written as separate Kafka headers, one per element.
+* **Consumer side:** reconstructed as a `List<?>` and returned unchanged to the application.
```
artembilan, line 243: I’m not sure about this “unchanged” statement.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -70,3 +70,9 @@ Several deprecated items have been removed:
 
 Spring for Apache Kafka 4.0 supports Kafka 4.0’s new consumer rebalance protocol - https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol[KIP-848].
 For details, see xref:kafka/receiving-messages/rebalance-listeners.adoc#new-rebalalcne-protocol[New Consumer Rebalace Protocol docs].
+
+[[x40-multi-value-header]]
+=== Support multi-value header
+
+The `DefaultKafkaHeaderMapper` and `SimpleKafkaHeaderMapper` supports multi-value header for Kafka Record.
```
artembilan, line 77: Since you use several objects , no need in a singular verb. Plus looks like “mapping” word is missed after “header”.

spring-kafka/src/test/java/org/springframework/kafka/support/DefaultKafkaHeaderMapperTests.java
```diff
@@ -378,6 +522,19 @@ void ensureNullHeaderValueHandledGraciously() {
 		verify(mockHeader, never()).key();
 	}
 
+	private List<byte[]> extractHeaderValues(Headers headers, String headerName) {
+		return StreamSupport.stream(headers.headers(headerName).spliterator(), false)
+							.map(Header::value)
+							.toList();
```
artembilan, line 528: I believe AssertJ can do assertion not only on List. I mean with AssertJ API it might turn out we don’t need this method.

artembilan (changes requested):

artembilan (approved):
LGTM so far.
Will take a final look on Monday and possibly merge.
Thank you again !

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/headers.adoc
```diff
@@ -179,3 +179,80 @@ MessagingMessageConverter converter() {
...
+- `xxx*`
+- `*xxx`
+- `*xxx*`
+- `xxx*yyy`
+====
+
+[IMPORTANT]
+====
+All elements collected under the same multi‑value header **must be of the same Java type**.
+Mixing, for example, `String` and `byte[]` values under a single header key is not supported and will lead to a conversion error.
```
artembilan, line 257: I think this is true only if producer side is our `DefaultKafkaHeaderMapper` when each item of the collection header is JSON-seriailized. If producer is some other framework, then we cannot claim this, since there won't be any dedicated type for those items, so everything is going come back to the consumer application as `List<byte[]>`.

sobychacko (approved):
@chickenchickenlove After looking at the tests again, I am ok with what you have in the PR for testing. No need for additional end-to-end testings that may complicate the PR. Since, Artem did a very thorough review on the PR, I am approving it. Looks good to me.

artembilan:
And let us know if that is OK with you to contribute docs next time!
Looks like that part has been more challenging than code.
I don't mean to reject your contribution; but just looking for some ways to optimize our collaboration.
So, if docs are hard for your, not a problem: just let us know and we will do that ourselves.
Meanwhile you can concentrate on something what is more comfortable for your to do.

Thank you again for your great contributions!

artembilan (approved):
LGTM. Will merge when build is green.
Thank you again for a great contribution!

chickenchickenlove:
@artembilan @sobychacko 
Thanks for you guys time!  🙇‍♂️ 

@artembilan 
Thank you for the thoughtful suggestion.

I agree that documentation is just as important as code, and I’m eager to improve in that area even though writing docs is still harder for me than writing code (and I’m certainly not perfect at coding either  😅 ).

My main concern is taking up too much of the reviewers’ time with doc revisions. 
If you ever feel that a particular section might require extensive editing to me, please feel free to suggest an alternative approach early on(as you mentioned eariler!).

I’m happy to start with smaller documentation tasks, learn from you guys feedback, and gradually take on larger ones as my writing improves—without putting an undue burden on the review team.

Thanks again for your guidance and for the opportunity to contribute! 🙇‍♂️

PR #3878, opened 2025-04-29
GH-3873: Deprecate JUnit 4 utilities in the project

### LINKED ISSUE #3873: Deprecate JUnit 4 utilities in the project
Also, migrate JUnit 4 tests to Jupiter.
More info in: https://github.com/spring-projects/spring-framework/issues/34794

chickenchickenlove:
- Fixes #3873 

The class that implements `MethodRule` in Junit4 is called whenever each test method is executed. 
Before executing the test method, it configures the log level.
Then, after executing the test method, it returns the log level back. 

Like `MethodRule` in Junit4, the interface `InvocationInterceptor` in Junit5 provides `interceptTestMethod(...)`. 
So, I migrated `MethodRule` to `InvocationInterceptor`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -57,6 +57,7 @@ ext {
 	hibernateValidationVersion = '8.0.2.Final'
 	jacksonBomVersion = '2.18.3'
 	jaywayJsonPathVersion = '2.9.0'
+	junit4Version = '4.13.2'
 	junitJupiterVersion = '5.12.2'
 	kafkaVersion = '4.0.0'
 	kotlinCoroutinesVersion = '1.10.2'
@@ -377,6 +378,9 @@ project ('spring-kafka-test') {
 		api 'org.junit.platform:junit-platform-launcher'
 		optionalApi "org.hamcrest:hamcrest-core:$hamcrestVersion"
 		optionalApi "org.mockito:mockito-core:$mockitoVersion"
+		optionalApi ("junit:junit:$junit4Version") {
+			exclude group: 'org.hamcrest', module: 'hamcrest-core'
+		}
 		optionalApi "org.apache.logging.log4j:log4j-core:$log4jVersion"
 	}
 }
--- a/build.gradle
+++ b/build.gradle
@@ -57,7 +57,6 @@ ext {
 	hibernateValidationVersion = '8.0.2.Final'
 	jacksonBomVersion = '2.18.3'
 	jaywayJsonPathVersion = '2.9.0'
-	junit4Version = '4.13.2'
 	junitJupiterVersion = '5.12.2'
 	kafkaVersion = '4.0.0'
 	kotlinCoroutinesVersion = '1.10.2'
@@ -378,9 +377,6 @@ project ('spring-kafka-test') {
 		api 'org.junit.platform:junit-platform-launcher'
 		optionalApi "org.hamcrest:hamcrest-core:$hamcrestVersion"
 		optionalApi "org.mockito:mockito-core:$mockitoVersion"
-		optionalApi ("junit:junit:$junit4Version") {
-			exclude group: 'org.hamcrest', module: 'hamcrest-core'
-		}
 		optionalApi "org.apache.logging.log4j:log4j-core:$log4jVersion"
 	}
 }
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
@@ -56,8 +56,7 @@ If this is not possible for some reason, note that the `consumeFromEmbeddedTopic
 Since it does not have access to the consumer properties, you must use the overloaded method that takes a `seekToEnd` boolean parameter to seek to the end instead of the beginning.
 ====
 
-NOTE: The `EmbeddedKafkaRule` JUnit 4 rule has been removed in version 4.0.
-For JUnit 4, you should use the `EmbeddedKafkaKraftBroker` directly or migrate to JUnit 5 with the `@EmbeddedKafka` annotation.
+NOTE: `Spring Kafka` no longer supports JUnit4. Please consider migrating from JUnit4 to JUnit5.
 
 The `EmbeddedKafkaBroker` class has a utility method that lets you consume for all the topics it created.
 The following example shows how to use it:
@@ -121,7 +120,7 @@ The following example configuration creates topics called `cat` and `hat` with f
 
 [source, java]
 ----
-@ExtendWith(SpringExtension.class)
+@SpringJUnitConfig
 @EmbeddedKafka(
 		partitions = 5,
 		topics = {"cat", "hat"}
@@ -229,7 +228,7 @@ The following example shows how to use it:
 
 [source, java]
 ----
-@ExtendWith(SpringExtension.class)
+@SpringJUnitConfig
 @DirtiesContext
 @EmbeddedKafka(partitions = 1,
          topics = {
@@ -292,7 +291,7 @@ In addition, the broker properties are loaded from the `broker.properties` class
 Property placeholders are resolved for the `brokerPropertiesLocation` URL and for any property placeholders found in the resource.
 Properties defined by `brokerProperties` override properties found in `brokerPropertiesLocation`.
 
-You can use the `@EmbeddedKafka` annotation with JUnit 4 or JUnit 5.
+You can use the `@EmbeddedKafka` annotation with JUnit 5.
 
 [[embedded-kafka-junit5]]
 == `@EmbeddedKafka` Annotation with JUnit5
@@ -337,7 +336,7 @@ The following example shows how to do so:
 =====
 [source, java]
 ----
-@ExtendWith(SpringExtension.class)
+@SpringJUnitConfig
 @SpringBootTest(properties = "spring.autoconfigure.exclude="
     + "org.springframework.cloud.stream.test.binder.TestSupportBinderAutoConfiguration")
 public class MyApplicationTests {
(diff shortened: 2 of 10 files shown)


spring-kafka-test/src/main/java/org/springframework/kafka/test/extensions/Log4j2LevelAdjuster.java
```diff
@@ -14,32 +14,34 @@
...
  * where we don't want to generate a large log in a tight inner loop.
  *
  * @author Dave Syer
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Sanghyoek An
  *
  */
-public class Log4j2LevelAdjuster implements MethodRule {
+public class Log4j2LevelAdjuster implements InvocationInterceptor {
```
sobychacko, line 44: Where is this class used with Junit5?

artembilan: This is not correct.
The class is public and for end-user consumption.
It is totally wrong to change the class contract.
We develop here a library for target projects, so whatever is `public` cannot be changed like this.
We have to revert the change here and deprecate it.
We may come up with JUnit 5 variant. But that is a different story.
You may take a inspiration from Spring AMQP: https://github.com/spring-projects/spring-amqp/blob/main/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/LogLevels.java

build.gradle
```diff
@@ -378,9 +377,6 @@ project ('spring-kafka-test') {
 		api 'org.junit.platform:junit-platform-launcher'
 		optionalApi "org.hamcrest:hamcrest-core:$hamcrestVersion"
 		optionalApi "org.mockito:mockito-core:$mockitoVersion"
-		optionalApi ("junit:junit:$junit4Version") {
```
artembilan, line 381: This is not correct.
We are talking about deprecation of the utility, but not total removal.
Therefore dependency and deprecated classes must still be here.
Again: we develop library, so whatever we do here might break target projects using our library.

Removing it temporary locally is a good way to check if we still have any JUnit 4 tests left. 😄

spring-kafka-test/src/test/java/org/springframework/kafka/test/extensions/AddressableEmbeddedBrokerTests.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package org.springframework.kafka.test.rule;
+package org.springframework.kafka.test.extensions;
```
artembilan, line 17: Doesn't look like this class logic does something with log adjustment.
Might be better to move it to some other package.
Probably to the root one alongside with the `EmbeddedKafkaKraftBrokerTests`

artembilan (changes requested):

spring-kafka-test/src/main/java/org/springframework/kafka/test/extensions/Log4j2LevelAdjuster.java
sobychacko: Maybe we can deprecate it on the `3.3.x` line and then still remove it in 4.0.0. Thoughts?

artembilan: Yeah... That would be inconsistent with Spring Framework plans.
They do deprecate JUnit 4 in version 7.0.

chickenchickenlove: @sobychacko 

In Junit4, `Log4j2LevelAdjuster` is used like this.
```java
class MyTests {

  @Rule
  private Log4j2LevelAdjuster adjuster = new Log4j2LevelAdjuster(MyTests.class);

  ...
}
```

In Junit5, `Log4j2LevelAdjuster` can be used like this. 
```java
class MyTests {

  @RegisterExtension
  private Log4j2LevelAdjuster adjuster = new Log4j2LevelAdjuster(MyTests.class);

  ...
}
```

chickenchickenlove:
@artembilan 

- There are no test cases using `Junit4` in `spring-kafka` project. So, I don't need to fix any test case. 
- I tried to replace [Components to Deprecate](https://github.com/spring-projects/spring-framework/issues/34794) with `JUnit5` one. 
  - However, there were no usages of deprecated components in the current `spring-kafka` codebase. 
  - However, many parts of the test documentation were written using `JUnit4`, so I updated them.
- Added `@Deprecated` to concreate class of JUnit4 interface.
- Update Java docs to use `Junit5` instead of `Junit4`.

I’m not sure if I’m on the right track with my work.
The work I just did was based on the following assumptions:

- Migrate all `JUnit4` test code used in spring-kafka to JUnit 5.
- Identify the `JUnit4-related deprecated components` in Spring and migrate them to Spring `JUnit5` equivalents (including both documentation and code).
- For any `JUnit 4-based public interfaces` exposed to users, add `@Deprecated` warnings.

Is my understanding correct?
If I’ve misunderstood anything, I would appreciate it if you could correct me.
I’m sorry for bothering you. 
This kind of work is new to me, so there are many parts that I find confusing. 

Sorry again and Thanks for your time. 🙇‍♂️

spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
```diff
@@ -121,15 +121,19 @@ The following example configuration creates topics called `cat` and `hat` with f
 
 [source, java]
 ----
+@ExtendWith(SpringExtension.class)
```
artembilan, line 124: Why `@SpringJUnitConfig` doesn't work for us here?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
```diff
@@ -350,6 +354,38 @@ They include:
 * xref:testing.adoc#kafka-testing-junit4-class-rule[JUnit4 Class Rule]
 * xref:testing.adoc#kafka-testing-embeddedkafka-annotation[`@EmbeddedKafka` Annotation or `EmbeddedKafkaBroker` Bean]
 
+[[kafka-testing-junit4-embedded-broker]]
+=== Junit4 Embedded Broker
```
artembilan, line 358: I think if we deprecate JUnit 4, we just should mention that we have done that and recommend to migrate to JUnit 5 without any further advises how to configure JUnit 4.

spring-kafka-test/src/main/java/org/springframework/kafka/test/context/EmbeddedKafka.java
```diff
@@ -43,7 +43,7 @@
  * <p>
  * The typical usage of this annotation is like:
  * <pre class="code">
- * &#064;RunWith(SpringRunner.class)
+ * &#064;ExtendWith(SpringExtension.class)
```
artembilan, line 46: DITTO about `@SpringJUnitConfig`.
It is very rare situation when we need to use `@ExtendWith(SpringExtension.class)`.

spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/Log4j2LevelAdjuster.java
```diff
@@ -37,8 +37,11 @@
  * @author Dave Syer
  * @author Artem Bilan
  * @author Gary Russell
- *
+ * @author Sanghyeok An
+ * @deprecated since Spring for Apache Kafka 4.0 in favor of the
+ * {@link org.springframework.kafka.test.condition.LogLevels} and JUnit Jupiter.
  */
+@Deprecated(since = "4.0")
```
artembilan, line 44: Missed `forRemoval = true`.
We really are going to remove it in the next `4.1` or `4.2`

spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/Log4j2LevelAdjuster.java
```diff
@@ -37,8 +37,11 @@
  * @author Dave Syer
  * @author Artem Bilan
  * @author Gary Russell
- *
+ * @author Sanghyeok An
+ * @deprecated since Spring for Apache Kafka 4.0 in favor of the
```
artembilan, line 41: This is not good style since the `@author` section becomes not visible and we lose a credit for those who contributed to the class.

You use too many words for the `@deprecated` tag. It is really obvious if we use version without project mentioning. If that is not a case, we would mention the target library name.
So, for me this sentence could be like this:
```
 @deprecated since 4.0 in favor of {@link org.springframework.kafka.test.condition.LogLevels}.
```

That is also obvious by `LogLevels` design that it is for JUnit 5.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
```diff
@@ -350,6 +354,38 @@ They include:
...
+            return new EmbeddedKafkaKraftBroker(1, 1, "someTopic");
+        }
+
+    }
+
+}
+----
+
 [[kafka-testing-junit4-class-rule]]
 === JUnit4 Class Rule
```
artembilan, line 390: I would prefer to have all the JUnit 4 text removed in favor of a single sentence that its support is deprecated for removal in the next minor version.

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
chickenchickenlove: I removed all contents about JUnit4 and left comment for asking migration to JUnit5. 
Thanks a lot 🙇‍♂️

spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/Log4j2LevelAdjuster.java
artembilan: Sure! Different projects may have different styles or just don't care about readability.
I try to be with my code as friendly as possible.
That style:

image

makes it hard for me to read the code, so I imaging that not only me, therefore in my code I try to separate those sections for readers of my code.

If you use `@` here in GH comments, please, wrap that into code snippet. 
Otherwise GH treats it as mentioning of that user here on GH.
Just try to follow that `@author` link. Since there that GH user you have just called him into our discussion 🤷 

I'm not sure what Javadoc link you show there, but the one we provide on our site has `@author` list: https://docs.spring.io/spring-kafka/docs/3.3.5/api/org/springframework/kafka/listener/AbstractMessageListenerContainer.html.

Either way, my point is not about Javadoc tool, but rather source code to be readable clearly.
In the end that's why it is an Open Source!

spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
```diff
@@ -121,15 +120,19 @@ The following example configuration creates topics called `cat` and `hat` with f
 
 [source, java]
 ----
+@SpringJUnitConfig
+@EmbeddedKafka(
+		partitions = 5,
```
artembilan, line 125: It is not clear from this view, but can we be sure that we don't use tabs for indents?

artembilan (changes requested):

spring-kafka-test/src/main/java/org/springframework/kafka/test/rule/Log4j2LevelAdjuster.java
chickenchickenlove: Thanks for your comments!!
I totally misunderstood your points. 😅 
I thought you were letting me know that the author would not be visible in the JavaDoc.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
```diff
@@ -56,8 +56,7 @@ If this is not possible for some reason, note that the `consumeFromEmbeddedTopic
 Since it does not have access to the consumer properties, you must use the overloaded method that takes a `seekToEnd` boolean parameter to seek to the end instead of the beginning.
 ====
 
-NOTE: The `EmbeddedKafkaRule` JUnit 4 rule has been removed in version 4.0.
-For JUnit 4, you should use the `EmbeddedKafkaKraftBroker` directly or migrate to JUnit 5 with the `@EmbeddedKafka` annotation.
+NOTE: `Spring Kafka` no longer supports JUnit4. Please consider migrating from JUnit4 to JUnit5.
```
sbrannen, line 59: I would not refer to JUnit Jupiter as "JUnit 5", especially since we (the JUnit team) will be releasing JUnit Jupiter 6 as part of "JUnit 6" later this year.

In other words, I think it is better to refer to JUnit Jupiter as "JUnit Jupiter" to avoid any confusion and unnecessary updates to documentation in the future.

I would also suggest that "JUnit Jupiter" be used consistently across the documentation.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
chickenchickenlove: I think I had used tabs for indents.
Now, I fixed it with 8 spaces fir indents.
I attached rendered image! Please refer it!

image

spring-kafka-docs/src/main/antora/modules/ROOT/pages/testing.adoc
```diff
@@ -56,8 +56,11 @@ If this is not possible for some reason, note that the `consumeFromEmbeddedTopic
 Since it does not have access to the consumer properties, you must use the overloaded method that takes a `seekToEnd` boolean parameter to seek to the end instead of the beginning.
 ====
 
-NOTE: The `EmbeddedKafkaRule` JUnit 4 rule has been removed in version 4.0.
-For JUnit 4, you should use the `EmbeddedKafkaKraftBroker` directly or migrate to JUnit 5 with the `@EmbeddedKafka` annotation.
+[NOTE]
+====
+`Spring Kafka` no longer supports `JUnit4`.
+Migration to `JUnit Jupiter` is recommended.
```
artembilan, line 62: Those words don’t justify to be a code.
Plus, why stick `4` to `JUnit` word?

Also, the official project name is `Spring for Apache Kafka` and since we don’t use code wrapping here why do that for JUnit projects?

artembilan (changes requested):

PR #3881, opened 2025-05-03
GH-3880: The DefaultHandler resolves an incorrect value for the parameter annotated with @Header.

### LINKED ISSUE #3880: The DefaultHandler resolves an incorrect value for the parameter annotated with @Header.
## In what version(s) of Spring for Apache Kafka are you seeing this issue?

Current, main branch.

## Describe the bug
```java

@KafkaListener(topics = "myTopic")
class MyListener {

  @KafkaHandler(isDefault = true)
  public void handleValue(@Header("kafka_receivedTopic") String topic, String message) { 
   // Anything.
  }
 }

public static void main() {
  String myMessage = "msg";
  template.send("myTopic", myMessage);
}
```

If we have listener with `DefaultHandler`, 
the `topic` value is always same with `message`. 
it means that equation below is true.

```java
topic.equals(message)
```

## To Reproduce

You can refer to code snippet `MyListener`.

## Expected behavior
- As-Is
  - Either `topic` or `message` are `msg`.
- To-Be
  - The value of `topic` should be 'myTopic'. 
  - The value of `message` should be 'msg'.

chickenchickenlove:
- Fixes #3880 

```java
// HandlerAdapter.java
Object[] args = new Object[providedArgs.length + 1];
args[0] = message.getPayload();
System.arraycopy(providedArgs, 0, args, 1, providedArgs.length);
return this.delegatingHandler.invoke(message, args);
```
If `defaultHandler` is set, the above code is executed.
During this call stack, the `args` array contains a payload of `KafkaMessage`, which is then passed to the invoked method.

[InvocableHandlerMethod.java (L212-L224)](https://github.com/spring-projects/spring-framework/blob/ecdb63371e99b4a285903106dcc06e90b355db53/spring-web/src/main/java/org/springframework/web/method/support/InvocableHandlerMethod.java#L212-L224)

```java

protected @Nullable Object[] getMethodArgumentValues(Message<?> message, @Nullable Object... providedArgs) throws Exception {
		MethodParameter[] parameters = getMethodParameters();
                 ...
 
		@Nullable Object[] args = new Object[parameters.length];
		for (int i = 0; i < parameters.length; i++) {
                 ...
			args[i] = findProvidedArgument(parameter, providedArgs); // Resolved by ProvidedArgs
			if (args[i] != null) {
				continue;
			}
                 ...
			try {
				args[i] = this.resolvers.resolveArgument(parameter, message); // Resolved By @Header
			}

```
each paramater's value is resolved here. 

```java
protected static @Nullable Object findProvidedArgument(MethodParameter parameter, @Nullable Object... providedArgs) {
		if (!ObjectUtils.isEmpty(providedArgs)) {
			for (Object providedArg : providedArgs) {
				if (parameter.getParameterType().isInstance(providedArg)) {
					return providedArg;
				}
			}
		}
		return null;
	}
```

However, the method `findProvidedArgument(...)` resolves parameter by only checking `parameterType`.
So, if either the payload type or the parameter type annotated with `@Header` is `String`, 
the payload and the annotated parameter will have the same value.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/class-level-kafkalistener.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/class-level-kafkalistener.adoc
@@ -62,3 +62,25 @@ void listen(Object in, @Header(KafkaHeaders.RECORD_METADATA) ConsumerRecordMetad
     ...
 }
 ----
+
+Also, this won't work as well.
+The `topic` is resolved to the `payload`.
+
+[source, java]
+----
+@KafkaHandler(isDefault = true)
+public void listenDefault(String payload, @Header(KafkaHeaders.RECEIVED_TOPIC) String topic) {
+    // payload.equals(topic) is True.
+    ...
+}
+----
+
+If there are use cases in which discrete custom headers are required in a default method, use this:
+[source, java]
+----
+@KafkaHandler(isDefault = true)
+void listenDefault(String payload, @Headers Map<String, Object> headers) {
+    Object myValue = headers.get("MyCustomHeader");
+    ...
+}
+----


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/HandlerAdapter.java
```diff
@@ -81,10 +82,7 @@ public Object invoke(Message<?> message, @Nullable Object... providedArgs) throw
 		}
 		else if (Objects.requireNonNull(this.delegatingHandler).hasDefaultHandler()) {
 			// Needed to avoid returning raw Message which matches Object
-			Object[] args = new Object[providedArgs.length + 1];
-			args[0] = message.getPayload();
-			System.arraycopy(providedArgs, 0, args, 1, providedArgs.length);
-			return this.delegatingHandler.invoke(message, args);
+			return this.delegatingHandler.invoke(message, providedArgs);
```
chickenchickenlove, line 85: inline comment; 

I think previous code in here cause the bug.  (https://github.com/spring-projects/spring-kafka/pull/577)
However, I don't fully understand why the payload needs to be resolved and included in the args when a defaultHandler is set. There may be reasons I haven't realized, so I would appreciate any advice from the maintainers on this point.
For now, only one of the existing test cases has been modified.

artembilan: See that comment:
```
// Needed to avoid returning raw Message which matches Object
```

If `@KafkaHandler(isDefault = true)` method has argument like `Object payload`, it would match the whole message instead of just its payload.

Usually, a `@KafkaHandler(isDefault = true)` method is used as a last resort when any other `@KafkaHandler` cannot match the provided payload.

The whole point of the `@KafkaHandler` is to handle different payload types with different methods.
So, I'm not sure if your fix is legit: https://docs.spring.io/spring-kafka/reference/kafka/receiving-messages/class-level-kafkalistener.html#page-title

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/HandlerAdapter.java
chickenchickenlove: @artembilan 
I have a question! 
```java
@KafkaListener
class MyListener { 

   @KafkaHandler(isDefault = true)
   public void (@Header("receivedTopic") String topic, Object payload) { 
        boolean isSame = topic == payload;
   }
}
```
There are two parameters.
One is annotated with `@Header`,  and the other one is the expected `payload`. 

I means that the `topic` and `payload` parameters are same in the current implementation.
May I ask if this behavior is intentional by design?
Thanks for your time 🙇‍♂️

artembilan: I have to look into those details deeper, but that looks suspicious that `payload` argument is resolved to header 🤷

chickenchickenlove: > _I have to look into those details deeper, but that looks suspicious that payload argument is resolved to header 🤷_

The code I mentioned PR description causes the problem. 
It is not part of `spring-kafka`. 😢 
InvocableHandlerMethod

artembilan:
Just wrote this test:
```
@SpringBootTest
@EmbeddedKafka
class SpringKafkaIssue3880ApplicationTests {

	static final String TEST_TOPIC = "TEST_TOPIC";

	@Autowired
	KafkaTemplate<String, String> kafkaTemplate;

	@Autowired
	KafkaListenerConfiguration kafkaListenerConfiguration;

	@Test
	void verifyKafkaHandlerDefault() throws InterruptedException {
		this.kafkaTemplate.send(TEST_TOPIC, "test data");

		assertThat(this.kafkaListenerConfiguration.dataReceivedLatch.await(10, TimeUnit.SECONDS)).isTrue();
		assertThat(this.kafkaListenerConfiguration.payload).isEqualTo("test data");
		assertThat(this.kafkaListenerConfiguration.topic).isEqualTo(TEST_TOPIC);
	}

	@TestConfiguration
	@KafkaListener(topics = SpringKafkaIssue3880ApplicationTests.TEST_TOPIC, groupId = "test_group")
	static class KafkaListenerConfiguration {

		CountDownLatch dataReceivedLatch = new CountDownLatch(1);

		volatile String topic;

		volatile Object payload;

		@KafkaHandler(isDefault = true)
		public void defaultHandler(@Header("receivedTopic") String topic, Object payload) {
			this.topic = topic;
			this.payload = payload;
			this.dataReceivedLatch.countDown();
		}

	}

}
```

It fails on the `assertThat(this.kafkaListenerConfiguration.topic).isEqualTo(TEST_TOPIC);`.

So, apparently the header is resolved to the payload somehow...

artembilan:
According to the doc, the existing behavior is correct: 

> Due to some limitations in the way Spring resolves method arguments, a default @KafkaHandler cannot receive discrete headers; it must use the ConsumerRecordMetadata as discussed in [Consumer Record Metadata](https://docs.spring.io/spring-kafka/reference/kafka/receiving-messages/listener-annotation.html#consumer-record-metadata).

So, that `if (parameter.getParameterType().isInstance(providedArg)) {` would apply for any first argument provided.
And if we have `Object payload`, but no real payload as a first provided arg, then this `payload` param would be resolved to something else, e.g. `ConsumerRecord`, or `Acknowledgment`.
Which really is not expected. It is less harmful to request a default method signature like `@Headers Map<String, Object> headers, Object payload`, instead of trying to mitigate all the possible use-cases when we fix the framework like you suggest.

Do I miss anything else?

chickenchickenlove:
Exactly! This is exactly what I'm saying.
Would it be okay to revert current commits and make a new commit to write description and limitation to `adoc` like `@Headers Map<String, Object> headers` instead of `@Header("blahblah") String header`.
WDYT?

artembilan:
> a new commit to write description and limitation to adoc

sure! That could be a good plan.
Thanks

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/class-level-kafkalistener.adoc
```diff
@@ -62,3 +62,25 @@ void listen(Object in, @Header(KafkaHeaders.RECORD_METADATA) ConsumerRecordMetad
     ...
 }
 ----
+
+Also, this won't work as well.
+The `payload` value will be injected as the `topic`.
```
artembilan, line 67: " will be injected into the `topic` parameter"?

Otherwise I ma getting confused as before.
Such a sentence makes me think that header value is injected into the `payload` parameter.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/class-level-kafkalistener.adoc
```diff
@@ -62,3 +62,25 @@ void listen(Object in, @Header(KafkaHeaders.RECORD_METADATA) ConsumerRecordMetad
...
+[source, java]
+----
+@KafkaHandler(isDefault = true)
+public void listenDefault(String payload, @Header(KafkaHeaders.RECEIVED_TOPIC) String topic) {
+    // payload.equals(topic) is True.
+    ...
+}
+----
+
+If you need discrete custom headers in a default method, use this:
```
artembilan, line 78: I thought we discussed before that no "you", "me", "we" or even "they" in the technical documentation.

artembilan (changes requested):

artembilan:
@chickenchickenlove ,

you contribute a lot and we are grateful for that.

Tried to merge this your PR and noticed this:

Signed-off-by: chickenchickenlove <ojt90902@naver.com>

from legal respective that doesn’t follow DCO requirements: https://docs.brigade.sh/topics/contributor-guide/signing/#:~:text=Verify%20the%20contribution%20in%20your,the%20terms%20of%20the%20DCO.&text=You%20MUST%20use%20your%20legal,Configure%20your%20git%20client%20appropriately.

Please , revise your Git client for your real name and real email.

chickenchickenlove:
@artembilan Thanks for your comment! 
I have a question! 
Should I revert to all commits I made in this PR and make new commits with legal names? 
Or would it be okay to apply it starting from the next PR?

artembilan:
You can just squash all the commits to one, add real Sign off by, and force push to PR.
Of course, you would need to remove those wrong Sign offs from a new commit message.

chickenchickenlove:
I squashed commits to update legal name, and force pushed.

artembilan (changes requested):
Please, re-push with the proper commit message.
Here are some tips how that supposed to be: https://cbea.ms/git-commit/

chickenchickenlove:
@artembilan 
Sorry for my poor commit message and to bother you again. 
I didn't catch your point clearly. 

I read the article you shared, and I assumed it might be one of the following requests.
- Should I include all commit messages in the rebased commit's description body?
- Are you recommending that I add an appropriate description since the commit message currently doesn’t have one?

What is your point? Please let me know 🙇‍♂️

artembilan:
I mean mostly “what?” and “why?”.
Turned out we had to change docs, and that had to be mentioned in the respective commit and with a reasoning .

Very often there is a situation when we look into the past change and we cannot determine why. 

I don’t insist on that, but if you’d like to avoid this kind of discussion in the future contributions I’d recommend to make commits meaningful. For example: https://github.com/spring-projects/spring-kafka/commit/b0db30c5ad954749bce7826ca8b4e14469e36bcf

chickenchickenlove:
Thanks for sharing proper examples 🙇‍♂️ 
I revised my commits to follow your guide and force push!
When you have time, please take another look. 🙇‍♂️

Issue #3712, opened 2025-01-17
Make `ContainerPausingBackOffHandler` working in batch mode.


user734879:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.3.1

**Describe the bug**

Intention was to process entire batch of messages as a single entity and retry it several times, then skip/recover entire batch if all retries exhausted. 

Throwing _BatchListenerFailedException_ is not applicable since 
- it requires to specify failed record, while all records are considered failed and in case of passing first record  - only it will be skipped, while all the rest from the batch are retried with the next poll operation to the Listener (as explained in https://github.com/spring-projects/spring-kafka/issues/3297). 
- passing the last record from the batch - gives that only last message is retried instead of entire batch

[Documentation](https://docs.spring.io/spring-kafka/reference/kafka/annotation-error-handling.html#retrying-batch-eh) suggests throwing non _BatchListenerFailedException_ and it works only if backoff time is less than max.poll.interval. When delays are longer it causes rebalancing since _FallbackBatchErrorHandler_ sleeps for backoff time between poll operations. Configuring _ContainerPausingBackOffHandler_ as a backoff error handler in order to handle long delays does not have impact.

**To Reproduce**

Use batch Kafka listener which throws non-BatchListenerFailedException to retry entire batch of events.

**Expected behavior**

_ContainerPausingBackOffHandler_ is applied when non-BatchListenerFailedException error is thrown.

**Sample**

Configuration of ErrorHandler:

    final ExponentialBackOff backOff = new ExponentialBackOff(A, B);
    backOff.setMaxInterval(C); // longer than max.poll.interval
    backOff.setMaxElapsedTime(X);
    return new DefaultErrorHandler(null, backOff, new ContainerPausingBackOffHandler(new ListenerContainerPauseService(null, taskScheduler));

sobychacko:
@user734879 Have you observed that when the `maxInterval` is less than the `max.poll.interval`, it works as expected? In that case, this is something that we might have to look into the `ContainerPausingBackOffHandler.` Do you have a small sample where we can reproduce your scenario? Thanks!

user734879:
@sobychacko yes, it works as expected when maxInterval is less than max.poll.interval (at the same time `ContainerPausingBackOffHandler` is not invoked at all as far as I can see)

added sample project here https://github.com/user734879/spring-kafka-retry-test

user734879:
[...]

PR #3885, opened 2025-05-05
GH-3712: Make ContainerPausingBackOffHandler working in batch mode.

chickenchickenlove:
- Fixes : #3712 

In [Retrying Complete Batches](https://docs.spring.io/spring-kafka/reference/kafka/annotation-error-handling.html#retrying-batch-eh) doc, document said that 

```
There is no guarantee that, when a batch is redelivered, the batch has the same number of records and/or the redelivered records are in the same order. 
It is impossible, therefore, to easily maintain retry state for a batch.
```

So, we cannot use `ContainerPausingBackOffHandler` in batch mode. 
The important things in #3712 is that `Consumer` don't call `poll(...)` during backoff sleep. 
It causes the consumer group rebalancing. 

If consumer calls the `poll(...)` during backoff sleep, 
The `FallbackBatchErrorHandler` can handle even `Exponential BackOff`. 

I think that adding `consumer.poll(Duration.ZERO)` during backoff is okay. 
Because, In `ErrorHandlingUtils.retryBatch(...)`, consumer pauses all `TopicPartition` assigned. 
So, `consumer.poll(...)` will never fetch records.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
@@ -51,6 +51,7 @@ import org.springframework.util.backoff.BackOffExecution;
  * @author Andrii Pelesh
  * @author Antonio Tomac
  * @author Wang Zhiyang
+ * @author Sanghyeok An
  *
  * @since 2.8
  *
@@ -113,12 +114,17 @@ public final class ErrorHandlingUtils {
 					throw new KafkaException("Woken up during retry", logLevel, we);
 				}
 				try {
-					ListenerUtils.conditionalSleep(
+					ListenerUtils.conditionalSleepWithPoll(
 							() -> container.isRunning() &&
 									!container.isPauseRequested() &&
 									records.partitions().stream().noneMatch(container::isPartitionPauseRequested),
-							nextBackOff
+							nextBackOff,
+							consumer
 					);
+				}
+				catch (WakeupException we) {
+					seeker.handleBatch(thrownException, records, consumer, container, NO_OP);
+					throw new KafkaException("Woken up during retry", logLevel, we);
 				}
 				catch (InterruptedException e1) {
 					Thread.currentThread().interrupt();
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
I am confused a bit.
First you say that this is a fix for `ContainerPausingBackOffHandler`, but then you state:

> So, we cannot use ContainerPausingBackOffHandler in batch mode.

So, what this fix really does?
How your fix is related to the `ContainerPausingBackOffHandler`?
Does your fix help the reported to follow his/her expectations?

chickenchickenlove:
@artembilan 
Sorry for my poor explanation.

> [Documentation](https://docs.spring.io/spring-kafka/reference/kafka/annotation-error-handling.html#retrying-batch-eh) suggests throwing non BatchListenerFailedException and it works only if backoff time is less than max.poll.interval. When delays are longer it causes rebalancing since FallbackBatchErrorHandler sleeps for backoff time between poll operations. Configuring ContainerPausingBackOffHandler as a backoff error handler in order to handle long delays does not have impact.

He/She attempted to use `ContainerPausingBackOffHandler` because we want to use `ExponentialBackOff`. 
However, for the reasons I mentioned earlier, the `BatchListener` don't support `ContainerPausingBackOffHandler`. 

If `FallbackBatchErrorHandler` supports condition `max.poll.interval > BackOffTime`, 
We don't need to use `ContainerPausingBackOffHandler` for condition `max.poll.interval > BackOffTime`. 
So, the problem issued in #3712 can be resolved by this PR. 

What do you think?

chickenchickenlove:
I see!
When you have time, Could you take a look? 
After this PR is submitted, I will talk to reporter about this and check it. 
Thanks in advance always 🙇‍♂️

artembilan (approved):
LGTM.

That `while` loop may indeed lead to the problem.
Adding `Consumer.poll()` in between iterations is great fix.

Thank you!

CC @sobychacko

chickenchickenlove:
Sorry to bother you.
After you approved PR, I added an inline comment to the Javadoc explaining the requirements. 😓 
Sorry again.

artembilan (changes requested):

artembilan (approved):
LGTM. Will merged and back-port tomorrow.

chickenchickenlove:
I squashed commit to update legal name!

Issue #3917, opened 2025-05-22
A constructor that fails to validate null value properly


ghgh415263:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

current main

**Describe the bug**

CommonMixedErrorHandler constructor check recordErrorHandler twice
```
public CommonMixedErrorHandler(CommonErrorHandler recordErrorHandler, CommonErrorHandler batchErrorHandler) {
		Assert.notNull(recordErrorHandler, "'recordErrorHandler' cannot be null");
		Assert.notNull(recordErrorHandler, "'batchErrorHandler' cannot be null");
		this.recordErrorHandler = recordErrorHandler;
		this.batchErrorHandler = batchErrorHandler;
	}
```

**To Reproduce**

Steps to reproduce the behavior.

**Expected behavior**

i think second assert have to check batchErrorHandler
```
public CommonMixedErrorHandler(CommonErrorHandler recordErrorHandler, CommonErrorHandler batchErrorHandler) {
		Assert.notNull(recordErrorHandler, "'recordErrorHandler' cannot be null");
		Assert.notNull(batchErrorHandler, "'batchErrorHandler' cannot be null");
		this.recordErrorHandler = recordErrorHandler;
		this.batchErrorHandler = batchErrorHandler;
	}
```

thanks for reading

ghgh415263:
thanks for review

i just sent PR

PR #3918, opened 2025-05-22
#3917 A constructor that fails to validate null value properly

ghgh415263:
#3917 A constructor that fails to validate null value properly


ghgh415263:
always thanks for quick response !

PR #3932, opened 2025-05-28
Add and refactor integration tests in DefaultShareConsumerFactoryTests

sobychacko:
- Added tests to check that multiple shared consumers each get records and all data is consumed.
- Refactored shared test logic into a helper method.
- Other cleanup in the test

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
@@ -21,7 +21,9 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.kafka.clients.admin.Admin;
 import org.apache.kafka.clients.admin.AdminClient;
@@ -49,7 +51,7 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
  * @since 4.0
  */
 @EmbeddedKafka(
-		topics = {"embedded-share-test"}, partitions = 1,
+		topics = {"embedded-share-test", "embedded-share-multi-test", "embedded-share-distribution-test"}, partitions = 1,
 		brokerProperties = {
 				"unstable.api.versions.enable=true",
 				"group.coordinator.rebalance.protocols=classic,share",
@@ -144,7 +146,6 @@ class DefaultShareConsumerFactoryTests {
 	}
 
 	@Test
-	@SuppressWarnings("try")
 	void integrationTestDefaultShareConsumerFactory(EmbeddedKafkaBroker broker) throws Exception {
 		final String topic = "embedded-share-test";
 		final String groupId = "testGroup";
@@ -165,16 +166,14 @@ class DefaultShareConsumerFactoryTests {
 		// For this test: force new share groups to start from the beginning of the topic.
 		// This is NOT the same as the usual consumer auto.offset.reset; it's a group config,
 		// so use AdminClient to set share.auto.offset.reset = earliest for our test group.
-		try (AdminClient ignored = AdminClient.create(adminProperties)) {
-			ConfigEntry entry = new ConfigEntry("share.auto.offset.reset", "earliest");
-			AlterConfigOp op = new AlterConfigOp(entry, AlterConfigOp.OpType.SET);
+		ConfigEntry entry = new ConfigEntry("share.auto.offset.reset", "earliest");
+		AlterConfigOp op = new AlterConfigOp(entry, AlterConfigOp.OpType.SET);
 
-			Map<ConfigResource, Collection<AlterConfigOp>> configs = Map.of(
-					new ConfigResource(ConfigResource.Type.GROUP, "testGroup"), Arrays.asList(op));
+		Map<ConfigResource, Collection<AlterConfigOp>> configs = Map.of(
+				new ConfigResource(ConfigResource.Type.GROUP, "testGroup"), Arrays.asList(op));
 
-			try (Admin admin = AdminClient.create(adminProperties)) {
-				admin.incrementalAlterConfigs(configs).all().get();
-			}
+		try (Admin admin = AdminClient.create(adminProperties)) {
+			admin.incrementalAlterConfigs(configs).all().get();
 		}
 
 		var consumerProps = new HashMap<String, Object>();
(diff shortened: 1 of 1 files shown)


spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
```diff
@@ -197,4 +185,112 @@ void integrationTestDefaultShareConsumerFactory(EmbeddedKafkaBroker broker) thro
...
+		final String topic = "embedded-share-distribution-test";
+		final String groupId = "distributionTestGroup";
+		int recordCount = 8;
+		List<String> consumerIds = List.of("client-dist-1", "client-dist-2");
+		Map<String, Set<String>> consumerRecords = runSharedConsumerTest(topic, groupId, consumerIds,
+				recordCount, broker);
+
+		// Assert all records were received (no loss)
+		Set<String> allReceived = new java.util.HashSet<>();
+		consumerRecords.values().forEach(allReceived::addAll);
```
artembilan: This still ha not been fixed.
When we put in set a value which has been already there, nothing happens with the set.
Therefore an assertion against set does not reflect a reality and does not prove that parallel consumers receive same data. Or did I misunderstood what is going on here?

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
```diff
@@ -197,4 +185,112 @@ void integrationTestDefaultShareConsumerFactory(EmbeddedKafkaBroker broker) thro
...
+				future.get();
+			}
+			catch (Exception e) {
+				throw new RuntimeException(e);
+			}
+		}
+		executor.shutdown();
+		assertThat(completed)
+			.as("All records should be received within timeout")
+			.isTrue();
```
artembilan, line 261: We don't need all of these extra variables and even don't need to deal with `Future`.
The `executor.shutdown();` and then assert on the `awaitTermination()` should do the trick.
See that `shutdown()` JavaDocs.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
```diff
@@ -197,4 +185,112 @@ void integrationTestDefaultShareConsumerFactory(EmbeddedKafkaBroker broker) thro
...
+			final int idx = i;
+			Future<?> future = executor.submit(() -> {
+				DefaultShareConsumerFactory<String, String> shareConsumerFactory = new DefaultShareConsumerFactory<>(
+						Map.of(
+							"bootstrap.servers", bootstrapServers,
+							"key.deserializer", org.apache.kafka.common.serialization.StringDeserializer.class,
+							"value.deserializer", org.apache.kafka.common.serialization.StringDeserializer.class
+						));
+				try (var consumer = shareConsumerFactory
+						.createShareConsumer(groupId, consumerIds.get(idx))) {
```
artembilan: Any reason to create consumers in those threads?
Why cannot we do that outside for the ` executor.submit()`?

artembilan (changes requested):

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
sobychacko: The parallel consumers should not receive the same data, so we are asserting that we got all the data, and don't care which consumers received it, as it is a non-deterministic process how kafka allocates which consumer receives the data when using shared consumers.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
sobychacko: @artembilan Don't you think that creating the shared consumer outside of the thread creates some race conditions? For ex, if we extract the consumer creation outside of the `submit`, then the consumers may show unpredictable behavior, since both threads have access to both instances of the shared consumers. Since kafka consumers are inherently not thread safe, we may want to create the consumer within the thread bounds. Or did I understand it wrong? I tried your suggestion and then ran into issues.

artembilan: OK. But that is not what I see in the `KafkaMessageListenerContainer`.
We create a `ListenerConsumer` from the `doStart()` and immediately there we call `consumerFactory.createConsumer()`.
However the `poll()` we indeed call from the other thread.
Therefore my confusion here why would we create this shared consumer instance exactly in a thread where we poll?

sobychacko: In the framework code you mentioned in `KafkaMessageListenerContainer`, although we hand off the created consumer to another thread, in the `doStart()` method, we ensure that the created consumer instance is exactly what is handed off to the polling thread and therefore it is thread safe. However, in our test scenario here, unless we do the same thing, (i.e., create a dedicated consumer instance and then pass that exactly to the thread), we may run into thread safety issues. IMO, it might be okay to to do what we do right now in the test. But if you see any issues with that, I will try to refactor the test.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
```diff
@@ -197,4 +185,102 @@ void integrationTestDefaultShareConsumerFactory(EmbeddedKafkaBroker broker) thro
 		consumer.close();
 	}
 
+	@Test
+	void integrationTestSharedConsumersDistribution(EmbeddedKafkaBroker broker) throws Exception {
+		final String topic = "embedded-share-distribution-test";
+		final String groupId = "distributionTestGroup";
+		int recordCount = 8;
+		List<String> consumerIds = List.of("client-dist-1", "client-dist-2");
+		Map<String, Set<String>> consumerRecords = runSharedConsumerTest(topic, groupId, consumerIds,
```
artembilan: Right. So, if we talk about all the data received (exclusively) independently of the number of consumers (queue semantics), then I wonder if just shared `List` to be fulfilled by those consumers would be enough for our test.
Then we just verify content of that list and check if there are no duplicates.
At the same time it might be hard to determine that those records indeed come from different consumers.
Not sure how to be, but `Set` is misleading here since it will eliminate duplicates.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
```diff
@@ -197,4 +185,102 @@ void integrationTestDefaultShareConsumerFactory(EmbeddedKafkaBroker broker) thro
...
+					"bootstrap.servers", bootstrapServers,
+					"key.deserializer", org.apache.kafka.common.serialization.StringDeserializer.class,
+					"value.deserializer", org.apache.kafka.common.serialization.StringDeserializer.class
+				)
+		);
+		for (int i = 0; i < consumerIds.size(); i++) {
+			final int idx = i;
+			executor.submit(() -> {
+				try (var consumer = shareConsumerFactory.createShareConsumer(groupId, consumerIds.get(idx))) {
+					consumer.subscribe(Collections.singletonList(topic));
```
artembilan, line 246: My idea was to move `create` and `subscribe` outside of the thread.
And only `poll` in that submitted task.
Doesn't that work?
Indeed it might be not related to our test case, but that how I feel consumers are supposed to be used.

spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
```diff
@@ -197,4 +185,102 @@ void integrationTestDefaultShareConsumerFactory(EmbeddedKafkaBroker broker) thro
...
+						}
+					}
+				}
+			});
+		}
+
+		assertThat(latch.await(10, TimeUnit.SECONDS))
+			.as("All records should be received within timeout")
+			.isTrue();
+		running.set(false);
```
artembilan: Can we instead of this flag really close those consumers after that latch is fulfilled?

Issue #3542, opened 2024-10-08
Adds the ability to add record interceptors instead of override them


Nyamiou:
**Expected Behavior**

I'm using a Spring Cloud Steam `ListenerContainerCustomizer` to customize a `AbstractMessageListenerContainer` in order to add a record interceptor. I would expect to be able to add a record interceptor at the last position, or to be able to get the current record interceptor, create a `CompositeRecordInterceptor` to do the same thing and then set it (either is fine).

**Current Behavior**

The issue I have is that only `setRecordInterceptor(...)` is public, not `getRecordInterceptor()` and there is no `addRecordInterceptor(...)` either. I would like to keep the existing record interceptors there and just add mine in last position but there seem to be no easy way to do this.

**Context**

I have a code similar to this:

```
@Bean
public ListenerContainerCustomizer<AbstractMessageListenerContainer<byte[], byte[]>> listenerContainerCustomizer() {
	return (container, destinationName, group) -> container.setRecordInterceptor(new MyRecordInterceptor());
}
```

That I want to use in a small company internal library, but I don't want to remove the other projects the ability to use their own record interceptors. 

I tried to use Spring Cloud Steam `ChannelInterceptor` (with `GlobalChannelInterceptor`) to achieve the thing I wanted, but they call `afterSendCompletion(...)` after each attempts, while the record interceptor `afterRecord(...)` is called after all attempts. Also they do not allow to modify headers.

artembilan:
Thank for the report!
We don't find this as an easy fix since there is going to be too many moving parts with those `add/remove([indext])` in the `CompositeRecordInterceptor` and `AbstractMessageListenerContainer`.
So, unless you see how to fix it quickly, we are going to look into that for the next version.

Nyamiou:
No problem, this is not urgent. I just wanted this to hopefully be fixed in the future. Thank you for answering so quickly.

chickenchickenlove:
@Nyamiou
I have an idea for you.
How about using this class for your case?

```java
public class TestRecordInterceptor<K,V> implements RecordInterceptor<K, V> {

    private final List<RecordInterceptor<K, V>> children = new ArrayList<>();

    public void addRecordInterceptor(RecordInterceptor<K, V> interceptor) {
        this.children.add(interceptor);
    }
    
    @Override
    public ConsumerRecord<K, V> intercept(ConsumerRecord<K, V> record, Consumer<K, V> consumer) {
[...]

PR #3937, opened 2025-06-01
spring-projectsGH-3542: Adds the ability to add record interceptors instead of override them

chickenchickenlove:
Fixes: #3542
Issue link: #3542

What?
Change `RecordInterceptor` to `List<RecordInterceptor>` in `MessageListenerContainer`.

Why?
To allow adding multiple `RecordInterceptor` instances instead of overriding the existing one. Currently, only a single `RecordInterceptor` is supported. Users may want to register multiple `RecordInterceptors`. There are some workarounds, but they are not clean or ideal solutions.

By supporting `List<RecordInterceptor`>, users can add their own interceptors via `setRecordInterceptor(...)`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -16,6 +16,7 @@
 
 package org.springframework.kafka.listener;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
@@ -114,7 +115,8 @@ public abstract class AbstractMessageListenerContainer<K, V>
 
 	private int topicCheckTimeout = DEFAULT_TOPIC_CHECK_TIMEOUT;
 
-	private @Nullable RecordInterceptor<K, V> recordInterceptor;
+	// TODO: HERE ASH
+	private List<RecordInterceptor<K, V>> recordInterceptors = new ArrayList<>();
 
 	private @Nullable BatchInterceptor<K, V> batchInterceptor;
 
@@ -460,8 +462,8 @@ public abstract class AbstractMessageListenerContainer<K, V>
 		this.kafkaAdmin = kafkaAdmin;
 	}
 
-	protected @Nullable RecordInterceptor<K, V> getRecordInterceptor() {
-		return this.recordInterceptor;
+	protected List<RecordInterceptor<K, V>> getRecordInterceptors() {
+		return this.recordInterceptors;
 	}
 
 	/**
@@ -472,7 +474,9 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	 * @see #setInterceptBeforeTx(boolean)
 	 */
 	public void setRecordInterceptor(@Nullable RecordInterceptor<K, V> recordInterceptor) {
-		this.recordInterceptor = recordInterceptor;
+		if (recordInterceptor != null) {
+			this.recordInterceptors.add(recordInterceptor);
+		}
 	}
 
 	protected @Nullable BatchInterceptor<K, V> getBatchInterceptor() {
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainer.java
@@ -60,6 +60,7 @@ import org.springframework.util.Assert;
  * @author Tomaz Fernandes
  * @author Wang Zhiyang
  * @author Lokesh Alamuri
+ * @author Sanghyeok An
  */
 public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageListenerContainer<K, V> {
 
@@ -282,7 +283,9 @@ public class ConcurrentMessageListenerContainer<K, V> extends AbstractMessageLis
 		container.setClientIdSuffix(this.concurrency > 1 || this.alwaysClientIdSuffix ? "-" + index : "");
 		container.setCommonErrorHandler(getCommonErrorHandler());
 		container.setAfterRollbackProcessor(getAfterRollbackProcessor());
-		container.setRecordInterceptor(getRecordInterceptor());
+		for (RecordInterceptor<K, V> recordInterceptor : getRecordInterceptors()) {
+			container.setRecordInterceptor(recordInterceptor);
+		}
 		container.setBatchInterceptor(getBatchInterceptor());
 		container.setInterceptBeforeTx(isInterceptBeforeTx());
 		container.setListenerInfo(getListenerInfo());
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -714,17 +714,17 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 
 		private final @Nullable Duration syncCommitTimeout;
 
-		private final @Nullable RecordInterceptor<K, V> recordInterceptor =
+		private final List<RecordInterceptor<K, V>> recordInterceptors =
 				!isInterceptBeforeTx() || this.transactionManager == null
-						? getRecordInterceptor()
-						: null;
+						? getRecordInterceptors()
+						: new ArrayList<>();
 
-		private final @Nullable RecordInterceptor<K, V> earlyRecordInterceptor =
+		private final List<RecordInterceptor<K, V>> earlyRecordInterceptors =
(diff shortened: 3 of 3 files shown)


artembilan (changes requested):
We already have this: https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeRecordInterceptor.java#L51.
So, I would expect done change over there . This is really not a container responsibility. Plus, you made too many braking changes. And setter contract is to override , not add. For add we would implement adder.
At a glance a cannot accept the change .
Thank you for understanding!

chickenchickenlove:
@artembilan 

Sorry to bother you. 
I misunderstood your point at all 🙇‍♂️.

I made new commits to use `CompositeRecordInterceptor`. 
When you have time, please take a look!

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -475,6 +475,28 @@ public void setRecordInterceptor(@Nullable RecordInterceptor<K, V> recordInterce
...
+	 * If the {@link AbstractMessageListenerContainer} is not configured with a
+	 * {@link CompositeRecordInterceptor}, {@link CompositeRecordInterceptor} will be created internally
+	 * and configured to hold the added interceptor.
+	 * Does not apply to batch listeners.
+	 * @param recordInterceptor the interceptor.
+	 * @since 4.0
+	 */
+	public void addRecordInterceptor(RecordInterceptor<K, V> recordInterceptor) {
+		if (this.recordInterceptor instanceof CompositeRecordInterceptor<K, V> compositeRecordInterceptor) {
+			compositeRecordInterceptor.addRecordInterceptor(recordInterceptor);
```
artembilan: So, if the composite is an external and shared, calling this from one container would mutate all others. I understand a desire to have everything container related in one place, but consequences might not be good.
Or we just leave responsibility only in the composite, or bite a bullet and always have a composite here internally as a value of interceptor property. This way any add on the container would not effect the rest of application . The awkward situation when composite is provided, but I don’t see any other way to protect from race condition .
Of course, I’d prefer to have add only in the composite, but I’m opened for any other arguments.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
chickenchickenlove: You are right. 
I missed that case. (Many users already create it as bean and use it) 

As you said, Adding a new API `addRecordInterceptor(...)` to only `CompositeRecordInterceptor` makes sense to me 👍 . Also, I don't like useless recursive usage (Composite -> Composite -> ... -> Composite) 😄.

spring-kafka/src/main/java/org/springframework/kafka/listener/CompositeRecordInterceptor.java
```diff
@@ -92,4 +92,12 @@ public void afterRecord(ConsumerRecord<K, V> record, Consumer<K, V> consumer) {
 		this.delegates.forEach(del -> del.afterRecord(record, consumer));
 	}
 
+	/**
+	 * Add an {@link RecordInterceptor} to delegates.
+	 * @param recordInterceptor the interceptor.
```
artembilan, line 98: `@since 4.0` ?

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -460,7 +460,11 @@ public void setKafkaAdmin(KafkaAdmin kafkaAdmin) {
 		this.kafkaAdmin = kafkaAdmin;
 	}
 
-	protected @Nullable RecordInterceptor<K, V> getRecordInterceptor() {
+	/**
+	 * Get the {@link RecordInterceptor} for modification, if configured.
+	 * @return the {@link RecordInterceptor}, or {@code null} if not configured
```
artembilan, line 465: `@since 4.0`

artembilan (changes requested):

sobychacko (changes requested):
@chickenchickenlove Changes look good to me. As @artembilan suggested, please make the getter in AMLC public (add since tag etc.). I would also suggest adding some type of unit testing for the changes involved. Apart from that, I think we may need some documentation updates as well as a small indicator for this in the `whats-new` docs. Thanks!

artembilan (approved):

chickenchickenlove:
I'm so sorry. I forgot it. 🙇‍♂️ 
I fixed lint error.

artembilan (approved):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -76,3 +76,9 @@ For details, see xref:kafka/receiving-messages/rebalance-listeners.adoc#new-reba
 
 The `DefaultKafkaHeaderMapper` and `SimpleKafkaHeaderMapper` support multi-value header mapping for Kafka records.
 More details are available in xref:kafka/headers.adoc#multi-value-header[Support multi-value header mapping].
+
+[[x40-add-record-interceptor]]
+=== Configure additional `RecordInterceptor`
+
+The `KafkaMessageListenerContainer` and `ConcurrentMessageListenerContainer` support `getRecordInterceptor()`.
```
sobychacko: I think we need to include a few sentences about it in the actual ref docs and just briefly mention it in the whats-new. Otherwise, what will happen is that this won't be noticed as more new versions are released.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -76,3 +76,10 @@ For details, see xref:kafka/receiving-messages/rebalance-listeners.adoc#new-reba
...
+If the returned interceptor is an instance of `CompositeRecordInterceptor`, additional `RecordInterceptor` instances can be added to it.
```
artembilan: This technical detail is not required to be present in `What's New`.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/message-listener-container.adoc
```diff
@@ -22,6 +22,10 @@ IMPORTANT: If the interceptor mutates the record (by creating a new one), the `t
 
 The `CompositeRecordInterceptor` and `CompositeBatchInterceptor` can be used to invoke multiple interceptors.
 
+Starting with version 4.0, `AbstractMessageListenerContainer` exposes `getRecordInterceptor()` as a public method.
+If the returned interceptor is an instance of `CompositeRecordInterceptor`, additional `RecordInterceptor` instances can be added to it even after the container instance extending `AbstractMessageListenerContainer` has been created and a `RecordInterceptor` has already been configured.
```
artembilan, line 26: I think some example how to get an interceptor, create composite if necessary and set a new one into container would be great.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/message-listener-container.adoc
```diff
@@ -22,6 +22,30 @@ IMPORTANT: If the interceptor mutates the record (by creating a new one), the `t
...
+
+[source, java]
+----
+public void configureRecordInterceptor(KafkaMessageListenerContainer<Integer, String> container) {
+    CompositeRecordInterceptor compositeInterceptor;
+
+    if (container.getRecordInterceptor() instanceof CompositeRecordInterceptor interceptor) {
+        compositeInterceptor = interceptor;
+    } else {
+        compositeInterceptor = new CompositeRecordInterceptor<>();
```
artembilan, line 38: Looks like just returned interceptor is not preserved in a new composite.

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/message-listener-container.adoc
chickenchickenlove: Sorry for that.
I added a new commit to fix that!

artembilan: Please, look into my comment again . 
that `interceptor` variable is out of use after this. I believe it has to be propagated to ctor of new composite .

artembilan (approved):
LGTM
Paging @sobychacko for final approval .
Thank you!

chickenchickenlove:
@sobychacko Hi!
When you have time, Please take a look 🙇‍♂️

sobychacko (approved):

PR #3938, opened 2025-06-01
KafkaTestUtils unified consumerPros call

mipo256:
Currently, the `KafkaTestUtils#consumerProps` overloaded methods I think have a couple of problems:

1. Their arguments order is not aligned with each other:

```
(String group, String autoCommit, EmbeddedKafkaBroker embeddedKafka)
(String brokers, String group)
(String brokers, String group, String autoCommit)
```

The group goes either the first one, or the second one, the same goes for autocommit setting

2. The `enable.auto.commit` is a `boolean` [according to doc](https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#enable-auto-commit), so I think accepting a `String` is not type safe, since the value. Making the API more typesafe I think is generally a good idea.

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
@@ -59,6 +59,7 @@ import org.springframework.util.Assert;
  * @author Hugo Wood
  * @author Artem Bilan
  * @author Sanghyeok An
+ * @author Mikhail Polivakha
  */
 public final class KafkaTestUtils {
 
@@ -81,9 +82,26 @@ public final class KafkaTestUtils {
 	 * @param autoCommit the auto commit.
 	 * @param embeddedKafka a {@link EmbeddedKafkaBroker} instance.
 	 * @return the properties.
+	 * @deprecated please, use {@link #consumerProps(EmbeddedKafkaBroker, String, boolean)} instead
 	 */
+	@Deprecated(forRemoval = true, since = "4.0.0")
 	public static Map<String, Object> consumerProps(String group, String autoCommit,
 			EmbeddedKafkaBroker embeddedKafka) {
+
+		return consumerProps(embeddedKafka.getBrokersAsString(), group, autoCommit);
+	}
+
+	/**
+	 * Set up test properties for an {@code <Integer, String>} consumer.
+	 *
+	 * @param group the group id.
+	 * @param autoCommit the auto commit.
+	 * @param embeddedKafka a {@link EmbeddedKafkaBroker} instance.
+	 * @return the properties.
+	 * @since 4.0
+	 */
+	public static Map<String, Object> consumerProps(EmbeddedKafkaBroker embeddedKafka, String group,
+			boolean autoCommit) {
 
 		return consumerProps(embeddedKafka.getBrokersAsString(), group, autoCommit);
 	}
@@ -96,7 +114,7 @@ public final class KafkaTestUtils {
 	 * @since 3.3
 	 */
 	public static Map<String, Object> consumerProps(String brokers, String group) {
-		return consumerProps(brokers, group, "false");
+		return consumerProps(brokers, group, false);
 	}
 
 	/**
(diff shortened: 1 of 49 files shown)


.editorconfig
```diff
@@ -1,11 +1,6 @@
 root=true
 
-[*.java]
-indent_style = tab
-indent_size = 4
-continuation_indent_size = 8
-
-[*.xml]
+[*.{java,xml}]
```
artembilan: This change is not related to the problem and I'm not sure if this config is used at all.
Any confirmation that this change does not cause any harm?
I cannot check because I use IntelliJ IDEA.
Thanks

spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/KafkaTestUtils.java
```diff
@@ -114,8 +129,21 @@ public static Map<String, Object> producerProps(EmbeddedKafkaBroker embeddedKafk
...
 	public static Map<String, Object> consumerProps(String brokers, String group, String autoCommit) {
+		return consumerProps(brokers, group, Boolean.parseBoolean(autoCommit));
+	}
+
+	/**
+	 * Set up test properties for an {@code <Integer, String>} consumer.
+	 * @param brokers the bootstrapServers property.
+	 * @param group the group id.
+	 * @param autoCommit the auto commit.
+	 * @return the properties.
```
artembilan, line 147: `@since 4.0` and for another new method.

.editorconfig
mipo256: This file in this particular case (I guess) [is meant to be used by Intelij](https://www.jetbrains.com/help/idea/editorconfig.html), and Intelij support EditorConfig for a pretty long time now, I think...

However, you are correct that this change does not relate to the main changes in the PR, that is why it comes in a separate commit :smile:

This is just a cleanup and removing unnecessary duplication. And old and this version are supposed to be equivalent. 

P.S: I'm one of the EditorConfig spec designers, so, from the [specification perspective](https://spec.editorconfig.org/#glob-expressions), this is correct. I'll leave this out to you, whether to retain or discard this change

artembilan: I see. Thanks.
Can you tell us then, please, if this one does exactly the same but with more options: https://github.com/spring-projects/spring-kafka/blob/main/src/idea/spring-framework.xml? 
I mean I really use that one and fully ignored this `.editorconfig` forever 😄 .

If we are good with `/idea/` dire content, then this `.editorconfig` might be fully removed?

mipo256: In this particular case, the `spring-framework.xml` is Intelij specific code style settings. They are applied in Intelij during formatting. 

Functionally, the Intelij settings are far more advanced, in the sense that EditorConfig is language agnostic, but Intelij assumes Java/Kotlin, and therefore is generally more powerful. 

However, as far as I know, Intelij takes into account config in the `.editorconfig` when it generated the code style settings (like `codeStyleConfig.xml`). But in our case, code style settings are committed to VCS, so the changes in them will be seen.

But  I'm not 100% sure, let me double-check this with Intelij team,

artembilan: As far as I know from my experience I have to apply that `spring-framework.xml` manually into my IDEA. 
Perhaps it hast to be moved into different place to be visible by IDEA when we open project?
Yeah... This is too much for this PR.
So, let me review your change one more time and we are good to do.
The rest of discussion about editor config should go somewhere else.
Like separate GH issue.
Thanks

artembilan (changes requested):
Please, build the project locally and look into logs.
There are a bunch of warnings like this:
```
/home/runner/work/spring-kafka/spring-kafka/spring-kafka-test/src/test/java/org/springframework/kafka/test/utils/KafkaTestUtilsTests.java:58: warning: [removal] consumerProps(String,String,EmbeddedKafkaBroker) in KafkaTestUtils has been deprecated and marked for removal
		Map<String, Object> consumerProps = KafkaTestUtils.consumerProps("ktuTests1", "false", broker);
```
I expect that new API would be used whenever it is still rely on just deprecated one.
Thanks

mipo256:
I agree, I'll migrate the usages fo the deprecated API. The `.editorconfig` commit really seems to be a bit off-topic, I'll roll it back.

sobychacko:
I wonder whether we can make these changes in `3.3.x` so that we can remove the deprecated methods from `4.0.0` altogether?

mipo256:
That is a good point, @sobychacko, I also thought about it. I think this is open to discussion. The question is do we want to provide the extended grace period, or we do not. The `KafkaTestUtil` is quite heavily used, and `spring-kafka` 4.0 is going to be released _relatively_ soon. So, I'll leave the resolution of this to you guys :smile:.

P.S: @artembilan I have cleaned-up all the usages of the deprecated API

artembilan:
Thank you for contribution, @mipo256 !

Please, consider to fix your Git client to add your real Mikhail Polivakha name to the `signed-off-by` sentence in the commit message.

mipo256:
Sure @artembilan. Thank you!

Issue #3872, opened 2025-04-27
# Issue: Tracing Lost When Using Batch Listeners in Kafka Consumer


patpatpat123:
## Description
When enabling batch listeners in the Kafka consumer configuration by setting `factory.setBatchListener(true)`.

## Current Configuration
In `KafkaConsumerConfig.java`, we have:
```java
factory.getContainerProperties().setObservationEnabled(true);
factory.setBatchListener(true);
```

```

@EnableKafka
@Configuration
public class KafkaConsumerConfig {

    @Bean
    public ConsumerFactory<String, String> consumerFactory() {
        Map<String, Object> props = new HashMap<>();
        props.put(
                ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG,
                "localhost:9093");
        props.put(
                ConsumerConfig.GROUP_ID_CONFIG,
                "birk");
        props.put(
                ConsumerConfig.MAX_POLL_RECORDS_CONFIG,
                "20");
        props.put(
                ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG,
                StringDeserializer.class);
        props.put(
                ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG,
                StringDeserializer.class);
        return new DefaultKafkaConsumerFactory<>(props);
    }

    @Bean
    public ConcurrentKafkaListenerContainerFactory<String, String>
    kafkaListenerContainerFactory() {

        ConcurrentKafkaListenerContainerFactory<String, String> factory =
                new ConcurrentKafkaListenerContainerFactory<>();
        factory.getContainerProperties().setObservationEnabled(true);
        factory.setBatchListener(true);
        factory.setConsumerFactory(consumerFactory());
        factory.getContainerProperties().setPollTimeout(5000);
        factory.setConcurrency(1);
        return factory;
    }
}
```

## Expected Behavior
- Traces should be properly propagated and visible in the tracing system (Tempo/Grafana)

## Actual Behavior
- No traces are being recorded when batch processing is enabled

## Workaround
Currently, the only workaround is to disable batch processing by removing `factory.setBatchListener(true)`, but this comes at the cost of reduced performance.

## Environment
- Spring Boot 3.4.4
- Spring Kafka
- OpenTelemetry with Tempo/Grafana backend
- Java 17

## Related Components
- OpenTelemetry configuration
```
@Bean
    public SpanExporter primaryOtlpGrpcSpanExporter() {
[...]

PR #3944, opened 2025-06-04
support per-record observations in batch listeners

igormq:
- support per-record observations in batch listeners
- document recordObservationsInBatch container property
- describe new option in the change history
- add integration test for per-record observations

Related to https://github.com/spring-projects/spring-kafka/issues/3872#issuecomment-2891259642


sobychacko:
@igormq Thank you for the PR. We will do a detailed review soon. But, please add your name as an author to the classes you modified. Also, make the modifications in the ref docs directly and then add a short sentence about it in the `whats-new` doc. The change history is only for previous versions. Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -1091,6 +1093,27 @@ public void setRestartAfterAuthExceptions(boolean restartAfterAuthExceptions) {
...
+	public boolean isRecordObservationsInBatch() {
+		return this.recordObservationsInBatch;
+	}
+
+	/**
+	 * Set whether to enable individual record observations in a batch.
+	 * When true, and a batch listener is configured with observation enabled, an observation
+	 * will be started for each record in the batch. Default false.
+	 * @param recordObservationsInBatch true to enable individual record observations.
+	 * @since 3.4
```
sobychacko: 4.0

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -1091,6 +1093,27 @@ public void setRestartAfterAuthExceptions(boolean restartAfterAuthExceptions) {
 		this.restartAfterAuthExceptions = restartAfterAuthExceptions;
 	}
 
+	/**
+	 * When true, and a batch listener is configured with observation enabled, an observation
+	 * will be started for each record in the batch.
+	 * @return recordObservationsInBatch.
+	 * @since 3.4
```
sobychacko: 4.0

sobychacko (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/appendix/change-history.adoc
```diff
@@ -1,6 +1,23 @@
...
+
+[[x34-record-observations-in-batch]]
+=== Individual Record Observations in Batch Processing
+
+A new `ContainerProperties` property `recordObservationsInBatch` has been introduced to enable individual record observations within batch listeners.
+When enabled, and a batch listener is configured with observation enabled, an observation will be started for each record in the batch.
+This allows for proper distributed tracing of individual records within a batch, even when each record originated from different producers with different trace contexts.
+The default value is `false` to maintain backward compatibility.
+
+See xref:kafka/micrometer.adoc#observation[Micrometer Observation] for more details.
```
artembilan: The place for this info is not correct.
Has to be in `What's New`.
The `Change History` is for older versions - former `What's New`.
And `What's New` is just a simple sentence to lead us to the target chapter with more details.
More over we are not sure in the implementation yet, so docs was a bit of overhead for your effort.
But thanks anyway!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -4005,6 +4012,69 @@ private Long computeBackwardWhereTo(long offset, boolean toCurrent, TopicPartiti
 
 		}
 
+		private void invokeBatchWithIndividualRecordTracing(final ConsumerRecords<K, V> records,
```
artembilan: The implementation deals with an `observation`, but you still call method `tracing`.
Shouldn't method name reflect reality, please?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -4005,6 +4012,69 @@ private Long computeBackwardWhereTo(long offset, boolean toCurrent, TopicPartiti
...
+				for (ConsumerRecord<K, V> record : recordList) {
+					Observation observation = KafkaListenerObservation.LISTENER_OBSERVATION.observation(
+							this.containerProperties.getObservationConvention(),
+								DefaultKafkaListenerObservationConvention.INSTANCE,
+								() -> new KafkaRecordReceiverContext(record, getListenerId(), getClientId(), this.consumerGroupId,
+										this::clusterId),
+								this.observationRegistry);
+					observation.start();
+					observations.add(observation);
+					scopes.add(observation.openScope());
```
artembilan: I don't think this approach is OK.
Every next scope would override whatever was there in the `ThreadLocal`.
Therefore there would be only one last observation available downstream for the service method.
So, that info would be misleading and dangerous.
My proposal was just `observe()` every record here to finish their traces.
Nothing more  suspicious and confusing.

The logic you propose is something what end-user would do in their service method for the batch.
And since we cannot predict what's the goal, therefore we don't provide any observation handling for batches.

artembilan (changes requested):
Thank you for looking into this!
Not easy task to tackle.
Hope you'll find my review as reasonable.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
igormq: > I don't think this approach is OK. Every next scope would override whatever was there in the `ThreadLocal`. Therefore there would be only one last observation available downstream for the service method. So, that info would be misleading and dangerous. My proposal was just `observe()` every record here to finish their traces. Nothing more suspicious and confusing.
> 
> The logic you propose is something what end-user would do in their service method for the batch. And since we cannot predict what's the goal, therefore we don't provide any observation handling for batches.

Don't know if i understood everything correctly, but did all the changes that i thought were pertinent!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -4005,6 +4012,62 @@ private Long computeBackwardWhereTo(long offset, boolean toCurrent, TopicPartiti
...
+					doInvokeBatchOnMessage(records, recordList);
+				}
+
+				batchInterceptAfter(records, null);
+				successTimer(sample, null);
+			}
+			catch (RuntimeException e) {
+				this.batchFailed = true;
+				failureTimer(sample, null, e);
+				batchInterceptAfter(records, e);
```
artembilan: I don't think that we need to handle errors and timer here.
That simply can be done in the calling method as it was before.

See `Observation.observe()` API to be called for each those records before we call the target listener.
We don't need to surround the target service call with any of those observations.
They are for each record, not the whole batch, so whatever you have left here still leads to confusion.

In the `Observation.observe()` callback we simply can log each of those record under debug logging level.
This way, the trace and span for those records would be present in logs and reported down to the observation collector.

But again: that's my feeling how that supposed to be for the batch if we still want to do this.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
sobychacko: I agree with @artembilan's comments above. Explicitly starting before the listener and stopping the observations after the listener method is invoked creates semantic confusion. It gives the impression that the observations cover the listener's processing. However, they are not for the whole batch; observations are made per record. Each observation will include the listener method invocation duration, which is misleading. 
In addition, managing `start()` and `stop()` manually, plus storing observations in a `List`, adds complexity and potential memory overhead for large batches. 
As Artem pointed out, we should use `Observation.observe()` to create and complete each observation before the listener runs. This method handles the observation lifecycle in one shot. It starts the observation, opens any scopes, executes the callback, handles errors by marking them on the observation, stops the observation, and closes any scopes. As you can see, it automatically does a lot of heavy lifting on your behalf. Manually calling start, stop, etc., makes things more error-prone, so I suggest going by the `observe()` approach with a callback for logging purposes, as Artem suggested above. This clarifies the semantics, as observations are completed before the listener, showing they only cover header extraction, avoiding confusion about their scope. Especially when tracing from a tool on the front-end, it will be very confusing if all the observations from the batch erroneously report the entire batch processing as part of their spans. 

Also, as Artem has already pointed out, please consolidate all the error handling, timers, etc., from the calling method.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -4005,6 +4012,62 @@ private Long computeBackwardWhereTo(long offset, boolean toCurrent, TopicPartiti
...
+
+				// Mark all observations with error
+				for (Observation observation : observations) {
+					if (!isListenerAdapterObservationAware()) {
+						observation.error(e);
+					}
+				}
+				throw e;
+			}
+			finally {
```
sobychacko: if we switch to `Observe.observe()`, then there is no need for this `finally` block as `observe()` will implicitly do that for you.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -4005,6 +4012,62 @@ private Long computeBackwardWhereTo(long offset, boolean toCurrent, TopicPartiti
...
+				if (!isListenerAdapterObservationAware()) {
+					// Stop observations
+					for (Observation observation : observations) {
+						observation.stop();
```
sobychacko: No need to stop when using `observe()`.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2443,25 +2443,32 @@ private void invokeBatchOnMessageWithRecordsOrList(final ConsumerRecords<K, V> r
...
-			try {
-				if (this.wantsFullRecords) {
-					Objects.requireNonNull(this.batchListener).onMessage(records, // NOSONAR
-							this.isAnyManualAck
-									? new ConsumerBatchAcknowledgment(records, recordList)
-									: null,
-							this.consumer);
+
+			// Handle individual record tracing for batch mode if enabled
+			if (this.containerProperties.isRecordObservationsInBatch() && this.observationEnabled) {
```
sobychacko: Isn't `this.observationEnabled` already covering `this.containerProperties.isRecordObservationsInBatch()`?

sobychacko (changes requested):

sobychacko:
@igormq Thank you for this PR. Implementing observability with batch listeners is not trivial, and I think you laid good foundations here. Providing this as an optional feature is very important, as we don't want to impose observability by default in all batch listeners. Please take a look at some of the feedback I provided, which largely echoes what @artembilan has already reviewed.  As you make further changes, we will need to add docs and other related components, but getting the implementation right is the critical step right now. Thanks again!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
sobychacko: I agree with @artembilan's comments above. Explicitly starting before the listener and stopping the observations after the listener method is invoked creates semantic confusion. It gives the impression that the observations cover the listener's processing. However, they are not for the whole batch; observations are made per record. Each observation will include the listener method invocation duration, which is misleading.
In addition, managing `start()` and `stop()` manually, plus storing observations in a `List`, adds complexity and potential memory overhead for large batches.
As Artem pointed out, we should use `Observation.observe()` to create and complete each observation before the listener runs. This method handles the observation lifecycle in one shot. It starts the observation, opens any scopes, executes the callback, handles errors by marking them on the observation, stops the observation, and closes any scopes. As you can see, it automatically does a lot of heavy lifting on your behalf. Manually calling `start`, `stop`, etc., makes things more error-prone, so I suggest going by the `observe()` approach with a callback for logging purposes, as Artem suggested above. This clarifies the semantics, as observations are completed before the listener, showing they only cover header extraction, avoiding confusion about their scope. Especially when tracing from a tool on the front-end, it will be very confusing if all the observations from the batch erroneously report the entire batch processing as part of their spans.

Also, as Artem has already pointed out, please consolidate all the error handling, timers, etc., from the calling method.

igormq: Done!

igormq:
> @igormq Thank you for this PR. Implementing observability with batch listeners is not trivial, and I think you laid good foundations here. Providing this as an optional feature is very important, as we don't want to impose observability by default in all batch listeners. Please take a look at some of the feedback I provided, which largely echoes what @artembilan has already reviewed. As you make further changes, we will need to add docs and other related components, but getting the implementation right is the critical step right now. Thanks again!

Hey @sobychacko and @artembilan thank you so much for all the feedback, really appreciate! I think that i addressed all of them. Let me know if I am missing something :)

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -4005,6 +4011,21 @@ private Long computeBackwardWhereTo(long offset, boolean toCurrent, TopicPartiti
...
+				// Create individual observations for each record without scopes
+				for (ConsumerRecord<K, V> record : recordList) {
+					Observation observation = KafkaListenerObservation.LISTENER_OBSERVATION.observation(
+							this.containerProperties.getObservationConvention(),
+								DefaultKafkaListenerObservationConvention.INSTANCE,
+								() -> new KafkaRecordReceiverContext(record, getListenerId(), getClientId(), this.consumerGroupId,
+										this::clusterId),
+								this.observationRegistry);
+					observation.observe(() -> {
+						logger.debug("processing record: " + KafkaUtils.format(record));
```
michaldo:
```suggestion
						logger.debug(() -> "processing record: " + KafkaUtils.format(record));
```

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2443,7 +2443,13 @@ private void invokeBatchOnMessageWithRecordsOrList(final ConsumerRecords<K, V> r
 				}
 			}
 			Object sample = startMicrometerSample();
+
+
 			try {
+				if (this.containerProperties.isObservationEnabled() ) {
```
artembilan: I think we can use here that `this.observationEnabled` already.
Yo have just calculated in the ctor if we can observer per-record in the batch.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -4005,6 +4011,21 @@ private Long computeBackwardWhereTo(long offset, boolean toCurrent, TopicPartiti
...
+				// Create individual observations for each record without scopes
+				for (ConsumerRecord<K, V> record : recordList) {
+					Observation observation = KafkaListenerObservation.LISTENER_OBSERVATION.observation(
+							this.containerProperties.getObservationConvention(),
+								DefaultKafkaListenerObservationConvention.INSTANCE,
+								() -> new KafkaRecordReceiverContext(record, getListenerId(), getClientId(), this.consumerGroupId,
+										this::clusterId),
+								this.observationRegistry);
+					observation.observe(() -> {
+						logger.debug(() -> "processing record: " + KafkaUtils.format(record));
```
artembilan: How about to be more specific?
`Observing record in batch: `

artembilan (changes requested):

sobychacko (changes requested):
@igormq Code changes look good. Now, we need to add reference docs with instructions on enabling observability in batch mode, maybe some examples etc. Then we need to mention this new feature in the `whats-new` doc.

igormq:
> Now, we need to add reference docs with instructions on enabling observability in batch mode, maybe some examples et

done that, and also fixed the check style issues

spring-kafka-docs/src/main/antora/modules/ROOT/pages/micrometer.adoc
```diff
@@ -0,0 +1,60 @@
+= Micrometer Support
+
+Spring for Apache Kafka provides support for Micrometer.
+
+[[listener-obs]]
+== Listener Observations
+
+When you have `micrometer-observation` on the class path, and you configure a
+`ObservationRegistry` in the application context, the container will automatically create
+`Observation`s for consumed records.
```
artembilan: Our preferences is `One sentence per line`: https://asciidoctor.org/docs/asciidoc-recommended-practices/

spring-kafka-docs/src/main/antora/modules/ROOT/pages/micrometer.adoc
```diff
@@ -0,0 +1,60 @@
+= Micrometer Support
+
+Spring for Apache Kafka provides support for Micrometer.
```
artembilan: Isn't this something what is addressed now via antora/modules/ROOT/pages/kafka/micrometer.adoc?
Can we consider to move your news into that file, please?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/micrometer.adoc
```diff
@@ -0,0 +1,60 @@
...
+Listener observation names:
+
+* `spring.kafka.listener`
+
+Key Values:
+
+* `spring.kafka.listener.id` - the listener id.
+* `spring.kafka.group.id` - the consumer group id.
+* `spring.kafka.client.id` - the consumer client id.
+* `spring.kafka.cluster.id` - the cluster id.
```
artembilan: These are generated in the docs: https://docs.spring.io/spring-kafka/reference/appendix/micrometer.html

spring-kafka-docs/src/main/antora/modules/ROOT/pages/micrometer.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+    ConcurrentKafkaListenerContainerFactory<Object, Object> factory = new ConcurrentKafkaListenerContainerFactory<>();
+    configurer.configure(factory, kafkaConsumerFactory);
+    factory.getContainerProperties().setRecordObservationsInBatch(true);
+    return factory;
+}
+----
+
+When this property is `true`, an observation will be created for each record in the
+batch, similar to a record listener.
```
artembilan: Well, this is not similar because we don't propagate those per-record observations down to the batch listener.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/micrometer.adoc
```diff
@@ -0,0 +1,60 @@
...
+This allows you to have visibility into the processing of each record, even within a
+batch context.
+No spans will be created for the batch itself, only for each record.
```
artembilan: I'm not sure if you broke something with your change, but we have to revise that an observation is really created for the whole batch around that listener method call.
If that was not supported at all, then it is OK.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/micrometer.adoc
michaldo: Right. Batch observation, contrary non-batch, is just tiny span confirming record reception. Worth to clearly say that in doc

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -76,3 +76,10 @@ For details, see xref:kafka/receiving-messages/rebalance-listeners.adoc#new-reba
 
 The `DefaultKafkaHeaderMapper` and `SimpleKafkaHeaderMapper` support multi-value header mapping for Kafka records.
 More details are available in xref:kafka/headers.adoc#multi-value-header[Support multi-value header mapping].
+
+[[x40-batch-observability]]
+=== Per-Record Observation in Batch Listeners
+
+It is now possible to get an observation for each record when using a batch listener.
+Previously, only a single observation was created for the entire batch.
```
sobychacko: I don't think we ever supported observability for batch listeners, right?

sobychacko (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
igormq: you're right, sorry

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -77,8 +77,16 @@ For details, see xref:kafka/receiving-messages/rebalance-listeners.adoc#new-reba
...
+
+It is now possible to get an observation for each record when using a batch listener.
+See xref:kafka/micrometer.adoc#batch-listener-obs[Observability for Batch Listeners] for more information.
+=======
 [[x40-add-record-interceptor]]
 === Configure additional `RecordInterceptor`
 
 Listener containers now support interceptor customization via `getRecordInterceptor()`.
 See the xref:kafka/receiving-messages/message-listener-container.adoc#message-listener-container[Message Listener Containers] section for details.
+>>>>>>> main
```
artembilan: I think merge conflict has not been fixed.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2423,6 +2423,21 @@ private void ackBatch(final ConsumerRecords<K, V> records) throws InterruptedExc
 			}
 		}
 
+		private void invokeBatchWithIndividualRecordObservation(List<ConsumerRecord<K, V>> recordList) {
```
artembilan: Our code style preference have a method declared after its calling method: https://github.com/spring-projects/spring-framework/wiki/Code-Style#java-source-file-organization

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/BatchIndividualRecordObservationTests.java
```diff
@@ -0,0 +1,378 @@
...
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for batch individual record observation functionality.
+ *
+ * @author Igor Quintanilha
+ * @since 3.4
```
artembilan: `4.0` ?

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/micrometer.adoc
```diff
@@ -119,3 +119,28 @@ The `record` property in both observation contexts contains the `ConsumerRecord`
...
+        ConsumerFactory<Object, Object> kafkaConsumerFactory) {
+
+    ConcurrentKafkaListenerContainerFactory<Object, Object> factory = new ConcurrentKafkaListenerContainerFactory<>();
+    configurer.configure(factory, kafkaConsumerFactory);
+    factory.getContainerProperties().setRecordObservationsInBatch(true);
+    return factory;
+}
+----
+
+When this property is `true`, an observation will be created for each record in the batch, but the observation is not propagated to the listener method.
```
artembilan, line 145: OK. It is not propagated to the listener, but can we have an explanation what happens instead then?

artembilan: Any update, please?
Or we can go ahead and clean up ourselves ?
Thanks

sobychacko:
Merged via https://github.com/spring-projects/spring-kafka/commit/85b3aee62bf40cac75790835fe05af10b751afb5.

PR #3945, opened 2025-06-05
Add Kafka shared consumer container support

sobychacko:
- New AbstractShareKafkaMessageListenerContainer base class with lifecycle management
- ShareKafkaMessageListenerContainer implementation for share consumer protocol
- Integration tests for end-to-end message delivery validation

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
@@ -1,0 +1,249 @@
+/*
+ * Copyright 2025-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.listener;
+
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.regex.Pattern;
+
+import org.apache.commons.logging.LogFactory;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.beans.BeanUtils;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.BeanNameAware;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.context.ApplicationEventPublisherAware;
+import org.springframework.core.log.LogAccessor;
+import org.springframework.kafka.core.ShareConsumerFactory;
+import org.springframework.kafka.support.TopicPartitionOffset;
+import org.springframework.util.Assert;
+
+/**
+ * Abstract base class for share consumer message listener containers.
+ * <p>
+ * Handles common lifecycle, configuration, and event publishing for containers using a
+ * {@link org.springframework.kafka.core.ShareConsumerFactory}.
+ * <p>
+ * Subclasses are responsible for implementing the actual consumer loop and message dispatch logic.
+ *
+ * @param <K> the key type
+ * @param <V> the value type
+ *
+ * @author Soby Chacko
+ */
+public abstract class AbstractShareKafkaMessageListenerContainer<K, V>
+		implements GenericMessageListenerContainer<K, V>, BeanNameAware, ApplicationEventPublisherAware,
+		ApplicationContextAware {
+
+	/**
+	 * The default {@link org.springframework.context.SmartLifecycle} phase for listener containers.
+	 */
+	public static final int DEFAULT_PHASE = Integer.MAX_VALUE - 100;
+
+	@NonNull
+	protected final ShareConsumerFactory<K, V> shareConsumerFactory;
+
+	protected final LogAccessor logger = new LogAccessor(LogFactory.getLog(this.getClass()));
+
+	private final ContainerProperties containerProperties;
+
+	protected final ReentrantLock lifecycleLock = new ReentrantLock();
+
+	@NonNull
+	private String beanName = "noBeanNameSet";
+
+	@Nullable
+	private ApplicationEventPublisher applicationEventPublisher;
+
+	private boolean autoStartup = true;
+
+	private int phase = DEFAULT_PHASE;
+
+	@Nullable
+	private ApplicationContext applicationContext;
+
+	private volatile boolean running = false;
+
+	/**
+	 * Construct an instance with the provided factory and properties.
+	 * @param shareConsumerFactory the factory.
+	 * @param containerProperties the properties.
+	 */
+	@SuppressWarnings("unchecked")
+	protected AbstractShareKafkaMessageListenerContainer(@Nullable ShareConsumerFactory<? super K, ? super V> shareConsumerFactory,
+														ContainerProperties containerProperties) {
+		Assert.notNull(containerProperties, "'containerProperties' cannot be null");
+		Assert.notNull(shareConsumerFactory, "'shareConsumerFactory' cannot be null");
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
```diff
@@ -0,0 +1,249 @@
...
+public abstract class AbstractShareKafkaMessageListenerContainer<K, V>
+		implements GenericMessageListenerContainer<K, V>, BeanNameAware, ApplicationEventPublisherAware,
+		ApplicationContextAware {
+
+	/**
+	 * The default {@link org.springframework.context.SmartLifecycle} phase for listener containers.
+	 */
+	public static final int DEFAULT_PHASE = Integer.MAX_VALUE - 100;
+
+	@NonNull
```
artembilan: I think you are missing a `package-info.java` file over here with that our `@org.jspecify.annotations.NullMarked` to avoid this annotation explicitly.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
```diff
@@ -0,0 +1,249 @@
...
+ * <p>
+ * Handles common lifecycle, configuration, and event publishing for containers using a
+ * {@link org.springframework.kafka.core.ShareConsumerFactory}.
+ * <p>
+ * Subclasses are responsible for implementing the actual consumer loop and message dispatch logic.
+ *
+ * @param <K> the key type
+ * @param <V> the value type
+ *
+ * @author Soby Chacko
```
artembilan, line 49: `@since 4.0`?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -0,0 +1,233 @@
...
+		this.listenerConsumer = new ShareListenerConsumer(listener);
+		setRunning(true);
+		this.listenerConsumerFuture = CompletableFuture.runAsync(this.listenerConsumer, consumerExecutor);
+	}
+
+	@Override
+	protected void doStop() {
+		setRunning(false);
+		// The consumer will exit its loop naturally when running becomes false.
+	}
```
artembilan, line 137: Doesn't look like we need this method at all.
The `stop()` in the super class can do everything for us.
Or do you have something else in mind to add over here eventually?
I understand that it was modeled after `AbstractMessageListenerContainer`, but still...

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -0,0 +1,233 @@
...
+
+		private final GenericMessageListener<?> genericListener;
+
+		private final @Nullable String consumerGroupId = ShareKafkaMessageListenerContainer.this.getGroupId();
+
+		private final @Nullable String clientId;
+
+		ShareListenerConsumer(GenericMessageListener<?> listener) {
+			this.consumer = ShareKafkaMessageListenerContainer.this.shareConsumerFactory.createShareConsumer(
+				ShareKafkaMessageListenerContainer.this.getGroupId(),
```
artembilan, line 171: Just nit-pick: I prefer to have a blank line after method subject if it is multi-line.
Otherwise it is hard to read the method body.
Well, just my humble opinion.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -0,0 +1,233 @@
...
+				ShareKafkaMessageListenerContainer.this.getClientId());
+			this.genericListener = listener;
+			this.clientId = ShareKafkaMessageListenerContainer.this.getClientId();
+			// Subscribe to topics, just like in the test
+			ContainerProperties containerProperties = getContainerProperties();
+			this.consumer.subscribe(java.util.Arrays.asList(containerProperties.getTopics()));
```
artembilan: Why fully-qualified class name?
Why not import for this `java.util.Arrays`?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -0,0 +1,233 @@
...
+			return this.clientId;
+		}
+
+		@Override
+		public void run() {
+			initialize();
+			Throwable exitThrowable = null;
+			while (isRunning()) {
+				try {
+					var records = this.consumer.poll(java.time.Duration.ofMillis(1000));
```
artembilan: I guess the poll duration could be a property of this class.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -0,0 +1,233 @@
...
+							// Temporarily auto-acknowledge and commit.
+							// We will refactor it later on to support more production-like scenarios.
+							this.consumer.acknowledge(record, AcknowledgeType.ACCEPT);
+							this.consumer.commitSync();
+						}
+					}
+				}
+				catch (Error e) {
+					this.logger.error(e, "Stopping share consumer due to an Error");
+					throw e;
```
artembilan, line 208: Don't we need to close consumer before re-throwing this?
You are mentioning stopping in the log message, but I don't see how that could happen if we throw here and don't let to pass down to the end of this method.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -0,0 +1,125 @@
...
+	private static void setShareAutoOffsetResetEarliest(String bootstrapServers, String groupId) throws Exception {
+		Map<String, Object> adminProperties = new HashMap<>();
+		adminProperties.put("bootstrap.servers", bootstrapServers);
+		ConfigEntry entry = new ConfigEntry("share.auto.offset.reset", "earliest");
+		AlterConfigOp op = new AlterConfigOp(entry, AlterConfigOp.OpType.SET);
+		Map<ConfigResource, Collection<AlterConfigOp>> configs = Map.of(
+				new ConfigResource(ConfigResource.Type.GROUP, groupId), List.of(op));
+		try (Admin admin = AdminClient.create(adminProperties)) {
+			admin.incrementalAlterConfigs(configs).all().get();
+		}
```
artembilan, line 122: This is a bit awkward approach to just set that `share.auto.offset.reset` property.
Cannot it be done via our `brokerProperties` on the `@EmbeddedKafka`?
How this suppose to happen in the target applications?
It feels like changing the broker state from the application is not OK.
Especially when we have several instances looking into the same Kafka cluster.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
sobychacko: Ya, I tried to follow the same patterns as the other container. In the subsequent iterations, we may have to do more decorations when we stop and consider it a placeholder.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
sobychacko: I think in the current release of Kafka (4.0.0), this is how it is done, i.e. we need to set it at the groupId. I don't think we can set it via broker properties in `EmbeddedKafka`. Maybe that will change once this feature is stabilized and production ready in 4.1. See this for details: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=255070434#KIP932:QueuesforKafka-Groupconfiguration

artembilan:
```
 > Could not resolve all files for configuration ':spring-kafka:compileClasspath'.
   > Could not find io.projectreactor.kafka:reactor-kafka:.
     Required by:
         project :spring-kafka
```
I guess time to deprecate those Reactor Kafka components in `3.3.x` and remove altogether here in `4.0`.
Looks like `reactor-kafka` is not a part of `reactor-bom` any more.

sobychacko:
Ok, i will send a PR soon.

artembilan (changes requested):
```
> Could not resolve all files for configuration ':spring-kafka:compileClasspath'.
   > Could not find io.projectreactor.kafka:reactor-kafka:.
     Required by:
         project :spring-kafka
```
I guess this PR has to be rebased to the latest `main`.

Thanks

PR #3950, opened 2025-06-07
propagate scope in async failures

igormq:
I still can’t reproduce the issue locally (possibly due to the use of embedded Kafka), but applying the fix in our codebase confirms that it resolves the underlying problem.

When async returns are enabled and a failure occurs, the Kafka sender loses the original consumer’s trace context. As a result, a new trace is started, causing messages that arrive in the retry or DLT topics to have a different trace ID from the original.

**Can you help me understand how to test it correctly?**

P.S.: I tested over the 3.3.6 branch so I could test the jar in my project.


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -734,12 +734,13 @@ protected void asyncFailure(Object request, @Nullable Acknowledgment acknowledgm
 									"Async Fail", source.getPayload()), cause));
 		}
 		catch (Throwable ex) {
-			this.logger.error(t, () -> "Future, Mono, or suspend function was completed with an exception for " + source);
 			acknowledge(acknowledgment);
 			if (canAsyncRetry(request, ex) && this.asyncRetryCallback != null) {
 				@SuppressWarnings("unchecked")
 				ConsumerRecord<K, V> record = (ConsumerRecord<K, V>) request;
 				this.asyncRetryCallback.accept(record, (RuntimeException) ex);
+			} else {
```
igormq: annoying receiving this error if we can recover from that

artembilan: Not sure what your comment means, but that is indeed not our code style.
See this formatting config to be provided for IntelliJ IDEA: https://github.com/spring-projects/spring-kafka/blob/main/src/idea/spring-framework.xml

igormq: Just that we are receiving these errors in the log, but we still can recover from that, so I moved this log here!

igormq:
> I'd like to see the fix issued against `main`. And please, follow a DCO requirements.

done!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1498,7 +1498,11 @@ protected void handleAsyncFailure() {
 			// We will give up on retrying with the remaining copied and failed Records.
 			for (FailedRecordTuple<K, V> copyFailedRecord : copyFailedRecords) {
 				try {
-					invokeErrorHandlerBySingleRecord(copyFailedRecord);
+//					if (copyFailedRecord.observation != null) {
+//						copyFailedRecord.observation.scoped(() -> invokeErrorHandlerBySingleRecord(copyFailedRecord));
+//					} else {
+						invokeErrorHandlerBySingleRecord(copyFailedRecord);
+//					}
```
artembilan: What happened to `scoped()` wrapper?
Why all of these commenting?
And what is the point of that `Observation` property in the `FailedRecordTuple` if it is out of use?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
artembilan: The `else` sentence has to be from a new line.

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/ObservationTests.java
```diff
@@ -801,4 +897,52 @@ Mono<Void> receive1(ConsumerRecord<Object, Object> record) {
...
+
+		@DltHandler
+		void handleDlt(ConsumerRecord<Integer, String> record, Exception exception) {
+			this.capturedSpanInDlt = this.tracer.currentSpan();
+			this.asyncFailureLatch.countDown();
+		}
+	}
+
+
+
```
artembilan: Only one blank line before closing `}` in the class, even if that class is inner and last in the outer class.

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/ObservationTests.java
```diff
@@ -72,6 +86,9 @@
 import org.springframework.kafka.KafkaException;
 import org.springframework.kafka.annotation.EnableKafka;
 import org.springframework.kafka.annotation.KafkaListener;
+
```
artembilan: I don't think all of these black lines in imports are correct.
Please, run `gradlew check` and fix all the Checkstyle violations before pushing to PR.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
igormq: sorry, submit code that i was trying to fix the test. SOrry @artembilan

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
igormq, line 1: when i put  a break point in line 551 the test start passing. without the breaking point , in the dlt, for some reason the current observation goes to null.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3433,7 +3438,7 @@ private Collection<ConsumerRecord<K, V>> getHighestOffsetRecords(ConsumerRecords
 		}
 
 		private void callbackForAsyncFailure(ConsumerRecord<K, V> cRecord, RuntimeException ex) {
-			this.failedRecords.addLast(new FailedRecordTuple<>(cRecord, ex));
+			this.failedRecords.addLast(new FailedRecordTuple<>(cRecord, ex, this.observationRegistry.getCurrentObservation()));
```
artembilan: This is not correct.
That `this.observationRegistry.getCurrentObservation()` is going to be called when this whole method is called from the `completableFutureResult.whenComplet`. And this, in turn, might be called from totally different thread.
So, not a fact that observation is there.

I'm not sure what is the real problem, so I cannot think of something how to help you to propagate that one down to that exception handler for async callbacks.

igormq: Hey @artembilan while doing the tests realised that. 

i refactored everything, if you can have another look i would really appreciate. I will update the description of the PR to clearly exemplify the problem. Also, the tests that i added is pretty clear of the expectations!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1498,7 +1498,13 @@ protected void handleAsyncFailure() {
 			// We will give up on retrying with the remaining copied and failed Records.
 			for (FailedRecordTuple<K, V> copyFailedRecord : copyFailedRecords) {
 				try {
-					invokeErrorHandlerBySingleRecord(copyFailedRecord);
+						KafkaListenerObservation.LISTENER_OBSERVATION.observation(
+										getContainerProperties().getObservationConvention(),
+										DefaultKafkaListenerObservationConvention.INSTANCE,
+										() -> new KafkaRecordReceiverContext(copyFailedRecord.record(), getListenerId(),
+												getClientId(), this.consumerGroupId, this::clusterId),
+										this.observationRegistry)
```
artembilan: I don't think that it is OK to start a new observation.
The error handling is really a part of already started one.
I will dig later a bit more how we do that in other places.

artembilan (changes requested):

artembilan:
OK. The logic is like this:
1. The `MessagingMessageListenerAdapter` does this in its `handleResult()` on the `completableFutureResult.whenComplete`:
```
				else {
					Throwable cause = t instanceof CompletionException ? t.getCause() : t;
					observation.error(cause);
					asyncFailure(request, acknowledgment, consumer, cause, source);
				}
```
2. That `asyncFailure()` calls the mentioned `asyncRetryCallback` with an implementation in the `KafkaMessageListenerContainer` like this:
```
		private void callbackForAsyncFailure(ConsumerRecord<K, V> cRecord, RuntimeException ex) {
			this.failedRecords.addLast(new FailedRecordTuple<>(cRecord, ex));
		}
```

3. We probably can propagate that `observation` from the `handleResult()` down to the `asyncFailure`. Open scope there.
4. And get access to the `currentObservation` from the mentioned `callbackForAsyncFailure` to be populated to that `FailedRecordTuple`.
5. This way `KafkaMessageListenerContainer.handleAsyncFailure()` would be able to restore an observation from the tuple to in that `invokeErrorHandlerBySingleRecord()`.

Not sure if that is a goal of your solution.
WDYT?

igormq:
@artembilan , does it make sense what i did?

> OK. The logic is like this:
> 
> 1. The `MessagingMessageListenerAdapter` does this in its `handleResult()` on the `completableFutureResult.whenComplete`:
> 
> ```
> 				else {
> 					Throwable cause = t instanceof CompletionException ? t.getCause() : t;
> 					observation.error(cause);
> 					asyncFailure(request, acknowledgment, consumer, cause, source);
> 				}
> ```
> 
> 2. That `asyncFailure()` calls the mentioned `asyncRetryCallback` with an implementation in the `KafkaMessageListenerContainer` like this:
> 
> ```
> 		private void callbackForAsyncFailure(ConsumerRecord<K, V> cRecord, RuntimeException ex) {
> 			this.failedRecords.addLast(new FailedRecordTuple<>(cRecord, ex));
> 		}
> ```
> 
> 3. We probably can propagate that `observation` from the `handleResult()` down to the `asyncFailure`. Open scope there.
> 4. And get access to the `currentObservation` from the mentioned `callbackForAsyncFailure` to be populated to that `FailedRecordTuple`.
> 5. This way `KafkaMessageListenerContainer.handleAsyncFailure()` would be able to restore an observation from the tuple to in that `invokeErrorHandlerBySingleRecord()`.
> 
> Not sure if that is a goal of your solution. WDYT?

@artembilan thank you a lot for the feedback. this is exactly what i was trying to achieve! made the changes accordingly!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -680,16 +680,16 @@ private final class ListenerConsumer implements SchedulingAwareRunnable, Consume
 
 		private final @Nullable CommonErrorHandler commonErrorHandler;
 
-		@Deprecated(since = "3.2", forRemoval = true)
```
artembilan: I don't think this was a correct change to do.
Please, bring it back.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -680,16 +680,16 @@ private final class ListenerConsumer implements SchedulingAwareRunnable, Consume
...
 				this.containerProperties.getKafkaAwareTransactionManager() != null ?
 						this.containerProperties.getKafkaAwareTransactionManager() :
 						this.containerProperties.getTransactionManager();
 
 		private final @Nullable KafkaAwareTransactionManager<?, ?> kafkaTxManager =
-				this.transactionManager instanceof KafkaAwareTransactionManager<?, ?> kafkaAwareTransactionManager ?
-						kafkaAwareTransactionManager : null;
+				this.transactionManager instanceof KafkaAwareTransactionManager
+						? (KafkaAwareTransactionManager<?, ?>) this.transactionManager
+						: null;
```
artembilan: This change also has nothing to do with the subject of the issue.
I'm think about back-porting your fix into `3.3.x`, so less distraction - better.

artembilan:
@igormq ,

Issue #3953, opened 2025-06-11
Random behavior in DelegatingByTypeSerializer based on type of delegates map


MaheshAravindV:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

For example:

3.3.6

**Describe the bug**

When creating a `DelegatingByTypeSerializer` with `assignable = true`, if one of the types in the input map is a child of another type, the serializer that gets picked is random.

**To Reproduce**

1. Create a `DelegatingByTypeSerializer` with `assignable = true`
2. Have one type in the input map as the child of another type
3. When a message of the child type is send, the actual serializer used is random if the map used to create the serializer does not enforce order. This is because the contract does not enforce people to use a SequencedMap. 

**Expected behavior**

The order in which the serializer is picked for a type if there are multiple available should be deterministic.

**Sample**

https://github.com/MaheshAravindV/spring-kafka-delegating-serializer-sample

artembilan:
I see.
Even if the logic there is like this:
```
private final Map<Class<?>, Serializer<?>> delegates = new LinkedHashMap<>();
...
this.delegates.putAll(delegates);
```
the provided map is copied to the internal `LinkedHashMap` and access order has to be deterministic, but between application restarts, the one run would have entries in one order, but another may have in different run.

I wonder if we can come up with some sorting algorithm internally instead of enforcing contract which still could fail if you do:
```
Map<Class<?>, Serializer<?>> typeMap = new LinkedHashMap<>();
typeMap.put(Object.class, new JsonSerializer<>());
typeMap.put(byte[].class, new ByteArraySerializer());
```
I mean this deterministic, but wrong by definition, since everything is going to be processed by the first entry.

MaheshAravindV:
I was thinking it might be a simpler fix to have the onus on the person creating the Serializer to define the order since this seems to be a pretty niche use case. But the suggested approach should work as well.

MaheshAravindV:
Either way, I can raise a PR for the same once we decide the approach.

artembilan:
I prefer to have a sort algorithm inside of this `DelegatingByTypeSerializer` with JavaDoc explanation what is going on when we use `assignable = true`.

MaheshAravindV:
I am done with the changes but I am unable to open the contributor's agreement page.

artembilan:
I think you talk about this page: https://github.com/spring-projects/spring-kafka?tab=readme-ov-file#contributing-to-spring-kafka.

[...]

PR #3956, opened 2025-06-13
Ensure correct ordering of delegates regardless of user input

MaheshAravindV:
Ensure correct ordering of delegates regardless of user input. 
Fixes: #3953

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
@@ -16,10 +16,7 @@
 
 package org.springframework.kafka.support.serializer;
 
-import java.util.Comparator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
 
 import org.apache.kafka.common.errors.SerializationException;
@@ -41,10 +38,8 @@ import org.springframework.util.Assert;
  */
 public class DelegatingByTypeSerializer implements Serializer<Object> {
 
-	private static final Comparator<Entry<Class<?>, Serializer<?>>> DELEGATES_ASSIGNABILITY_COMPARATOR =
-			(entry1, entry2) -> {
-				Class<?> type1 = entry1.getKey();
-				Class<?> type2 = entry2.getKey();
+	private static final Comparator<Class<?>> DELEGATES_ASSIGNABILITY_COMPARATOR =
+			(type1, type2) -> {
 
 				if (type1.isAssignableFrom(type2)) {
 					return 1;
@@ -56,7 +51,7 @@ public class DelegatingByTypeSerializer implements Serializer<Object> {
 				return 0;
 			};
 
-	private final Map<Class<?>, Serializer<?>> delegates = new LinkedHashMap<>();
+	private final Map<Class<?>, Serializer<?>> delegates = new TreeMap<>(DELEGATES_ASSIGNABILITY_COMPARATOR);
 
 	private final boolean assignable;
 
@@ -86,16 +81,7 @@ public class DelegatingByTypeSerializer implements Serializer<Object> {
 		Assert.notNull(delegates, "'delegates' cannot be null");
 		Assert.noNullElements(delegates.values(), "Serializers in delegates map cannot be null");
 
-		if (!assignable) {
-			this.delegates.putAll(delegates);
-		}
-		else {
-			List<Entry<Class<?>, Serializer<?>>> sortedDelegates = delegates.entrySet().stream()
-					.sorted(DELEGATES_ASSIGNABILITY_COMPARATOR).toList();
-
-			sortedDelegates.forEach(it -> this.delegates.put(it.getKey(), it.getValue()));
-		}
-
+		this.delegates.putAll(delegates);
 		this.assignable = assignable;
 	}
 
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
```diff
@@ -32,12 +34,28 @@
...
  * @author Artem Bilan
  * @author Wang Zhiyang
+ * @author Mahesh Aravind V
  *
  * @since 2.7.9
  *
  */
 public class DelegatingByTypeSerializer implements Serializer<Object> {
 
+	private static final Comparator<Entry<Class<?>, Serializer<?>>> DELEGATES_ASSIGNABILITY_COMPARATOR =
```
artembilan: Can we think about using `TreeMap(Comparator<? super K> comparator)` instead?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
```diff
@@ -51,18 +69,33 @@ public DelegatingByTypeSerializer(Map<Class<?>, Serializer<?>> delegates) {
...
-	 * such as a {@link LinkedHashMap} is recommended.
-	 * @param delegates the delegates.
-	 * @param assignable whether the target is assignable to the key.
+	 * Construct an instance with the map of delegates.
+	 * If {@code assignable} is {@code false}, only exact key matches are considered.
+	 * If {@code assignable} is {@code true}, a delegate is selected if its key class
+	 * is assignable from the target object's class. When multiple matches are possible,
+	 * the most specific matching class is selected — that is, the closest match in the
+	 * class hierarchy.
+	 *
```
artembilan, line 73: No blank lines in method JavaDocs, please.
You can use `spring-framework.xml` IntelliJ IDEA editor config from the project to realign with our code style.

spring-kafka/src/test/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializerTests.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+/**
+ * @author Mahesh Aravind V
+ *
+ */
+class DelegatingByTypeSerializerTests {
```
artembilan, line 33: Why do you create a new test class if we have one on the matter already: `DelegatingSerializationTests`?
Please, consider to add your test over there.

MaheshAravindV: There is a class `DelegatingSerializer` and the `DelegatingSerializationTests` is for tests of that class I assume.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
artembilan: Have not been addressed yet

spring-kafka/src/test/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializerTests.java
artembilan: It is true, but I see tests for `DelegatingByTypeSerializer` over there, so I would like to have your new one over there as well.
And what is the reason of `@Nested`?
Cannot we really have it as a top-level test method and have the code flow as simple as it could be?

sobychacko: @MaheshAravindV It's better to have all the tests in `DelegatingSerializationTests`. Could you move it there?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
```diff
@@ -32,13 +31,27 @@
...
+			(type1, type2) -> {
+
+				if (type1.isAssignableFrom(type2)) {
+					return 1;
+				}
+				if (type2.isAssignableFrom(type1)) {
+					return -1;
+				}
+
+				return 0;
```
artembilan, line 51: I wonder if more robust algorithm would be like this:
```
public class ClassHierarchyComparator implements Comparator<Class<?>> {

    @Override
    public int compare(Class<?> class1, Class<?> class2) {
        if (class1 == class2) {
            return 0; // Classes are the same
        }
        if (class1.isAssignableFrom(class2)) {
            return -1; // class1 is a superclass or superinterface of class2, so class1 should come first
        }
        if (class2.isAssignableFrom(class1)) {
            return 1; // class2 is a superclass or superinterface of class1, so class2 should come first
        }
        return class1.getName().compareTo(class2.getName()); // If no inheritance relation, compare by name
    }
}
```

artembilan (changes requested):

artembilan:
Has been merged as https://github.com/spring-projects/spring-kafka/commit/df5860d0c4cf37fe94d24684ffbb03f288715b37
With subsequent clean up and refinements: https://github.com/spring-projects/spring-kafka/commit/797b96eb361271d37a48aa0945f8a8b3440bc794

@MaheshAravindV ,

MaheshAravindV:
Was busy with something else. Apologies. Will continue contributing

PR #3988, opened 2025-07-01
Add KafkaListener support for shared consumer containers

sobychacko:
- Add AbstractShareKafkaListenerContainerFactory base class for share consumer factories
- Add ShareKafkaListenerContainerFactory concrete implementation
- Add ShareRecordMessagingMessageListenerAdapter for share consumer message handling
- Modify MethodKafkaListenerEndpoint to create appropriate listener adapters based on container type
- Add integration tests for ShareKafkaListener functionality

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
@@ -124,13 +124,9 @@ public class ShareKafkaListenerContainerFactory<K, V>
 	 */
 	protected void initializeContainer(ShareKafkaMessageListenerContainer<K, V> instance, KafkaListenerEndpoint endpoint) {
 		ContainerProperties properties = instance.getContainerProperties();
+		Boolean effectiveAutoStartup = endpoint.getAutoStartup() != null ? endpoint.getAutoStartup() : this.autoStartup;
 		JavaUtils.INSTANCE
-				.acceptIfNotNull(endpoint.getAutoStartup(), instance::setAutoStartup)
-				.acceptIfNotNull(this.autoStartup, autoStartup -> {
-					if (endpoint.getAutoStartup() == null) {
-						instance.setAutoStartup(autoStartup);
-					}
-				})
+				.acceptIfNotNull(effectiveAutoStartup, instance::setAutoStartup)
 				.acceptIfNotNull(this.phase, instance::setPhase)
 				.acceptIfNotNull(this.applicationContext, instance::setApplicationContext)
 				.acceptIfNotNull(this.applicationEventPublisher, instance::setApplicationEventPublisher)


spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,67 @@
...
+ * instances for Kafka's share consumer model.
+ *
+ * @param <K> the key type
+ * @param <V> the value type
+ *
+ * @author Soby Chacko
+ * @since 4.0
+ */
+public class ShareKafkaListenerContainerFactory<K, V>
+		extends AbstractShareKafkaListenerContainerFactory<ShareKafkaMessageListenerContainer<K, V>, K, V> {
```
artembilan: Any idea why do we need an abstract super class instead of combining all the logic in this class?

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
```diff
@@ -66,6 +70,7 @@ public abstract class AbstractShareKafkaMessageListenerContainer<K, V>
 
 	protected final ReentrantLock lifecycleLock = new ReentrantLock();
 
+	@NonNull
```
artembilan: No one ever use this annotation if we do `@org.jspecify.annotations.NullMarked` in the `package-info.java`.
That one assumes that all not marked with `@Nullable` properties are not null.
If you have a complain here that it might be null, then we need to revise the logic around this property.
However I believe it is good enough to mark it as `@SuppressWarnings("NullAway.Init")` because this property is usually initialized from the application context.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -73,7 +71,7 @@ public class ShareKafkaMessageListenerContainer<K, V>
 	 * @param shareConsumerFactory the share consumer factory
 	 * @param containerProperties the container properties
 	 */
-	public ShareKafkaMessageListenerContainer(ShareConsumerFactory<? super K, ? super V> shareConsumerFactory,
+	public ShareKafkaMessageListenerContainer(@Nullable ShareConsumerFactory<? super K, ? super V> shareConsumerFactory,
```
artembilan: Why have this happened?
How did that turned out that now a `ShareKafkaMessageListenerContainer` can be created without a factory?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ShareRecordMessagingMessageListenerAdapter.java
```diff
@@ -0,0 +1,63 @@
...
+	 * Kafka {@link MessageListener} entry point.
+	 * <p> Delegate the message to the target listener method,
+	 * with appropriate conversion of the message argument.
+	 * <p> This implementation passes null for both acknowledgment and consumer parameters
+	 * to the super implementation.
+	 * @param record the incoming Kafka {@link ConsumerRecord}.
+	 */
+	@Override
+	public void onMessage(ConsumerRecord<K, V> record) {
+		onMessage(record, null, null);
```
artembilan: I don't see a reason in this class at all.
What is wrong with using `RecordMessagingMessageListenerAdapter` as it is?

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaListenerIntegrationTests.java
```diff
@@ -0,0 +1,149 @@
...
+
+	static class TestListener {
+
+		@KafkaListener(topics = "share-listener-integration-test", containerFactory = "shareKafkaListenerContainerFactory")
+		public void listen(ConsumerRecord<String, String> record) {
+			received.set(record.value());
+			latch.countDown();
+		}
+	}
+}
```
artembilan, line 153: For better readability (e.g. review), I prefer to have every class member to be surrounded with blank lines, including last method of the last inner class.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaListenerIntegrationTests.java
```diff
@@ -0,0 +1,149 @@
...
+			configs.put(ConsumerConfig.GROUP_ID_CONFIG, "share-listener-test-group");
+			configs.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
+			configs.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
+			return new DefaultShareConsumerFactory<>(configs);
+		}
+
+		@Bean
+		public ShareKafkaListenerContainerFactory<String, String> shareKafkaListenerContainerFactory(
+				ShareConsumerFactory<String, String> shareConsumerFactory) {
+			return new ShareKafkaListenerContainerFactory<>(shareConsumerFactory);
```
artembilan, line 134: For better readability, it is recommended to add a blank line before method body after multi-line method header.

spring-kafka/src/main/java/org/springframework/kafka/config/MethodKafkaListenerEndpoint.java
```diff
@@ -175,7 +175,9 @@ protected MessagingMessageListenerAdapter<K, V> createMessageListener(MessageLis
 
 		Assert.state(this.messageHandlerMethodFactory != null,
 				"Could not create message listener - MessageHandlerMethodFactory not set");
-		MessagingMessageListenerAdapter<K, V> messageListener = createMessageListenerInstance(messageConverter);
+
+		final MessagingMessageListenerAdapter<K, V> messageListener;
+		messageListener = createMessageListenerInstance(messageConverter);
```
artembilan: Let's revise changes in this class!
What you have done on these lines makes no sense so far.
The previous code version was much  better to read, and it is identical to the logic you have so far.

spring-kafka/src/main/java/org/springframework/kafka/config/MethodKafkaListenerEndpoint.java
```diff
@@ -201,11 +203,6 @@ protected HandlerAdapter configureListenerAdapter(MessagingMessageListenerAdapte
 		return new HandlerAdapter(invocableHandlerMethod);
 	}
 
-	/**
-	 * Create an empty {@link MessagingMessageListenerAdapter} instance.
-	 * @param messageConverter the converter (may be null).
-	 * @return the {@link MessagingMessageListenerAdapter} instance.
-	 */
```
artembilan: Not sure if this is OK to remove this JavaDoc.
More explanations - better.

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,240 @@
...
+		if (endpoint.getClientIdPrefix() != null) {
+			instance.getContainerProperties().setClientId(endpoint.getClientIdPrefix());
+		}
+		if (endpoint.getConsumerProperties() != null) {
+			instance.getContainerProperties().setKafkaConsumerProperties(endpoint.getConsumerProperties());
+		}
+	}
+
+	@Override
+	public ShareKafkaMessageListenerContainer<K, V> createContainer(TopicPartitionOffset... topicPartitions) {
```
artembilan, line 139: Is this even correct for a queue abstraction?
I thought partitions are out of share consumer API:
```
public interface ShareConsumer<K, V> extends Closeable {
    Set<String> subscription();

    void subscribe(Collection<String> var1);

    void unsubscribe();

    ConsumerRecords<K, V> poll(Duration var1);
```

There is no `assign()` API, therefore the partition-based configuration is wrong.

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,240 @@
...
+
+	/**
+	 * Create a container instance for the provided endpoint.
+	 * @param endpoint the endpoint
+	 * @return the container instance
+	 */
+	protected ShareKafkaMessageListenerContainer<K, V> createContainerInstance(KafkaListenerEndpoint endpoint) {
+		TopicPartitionOffset[] topicPartitions = endpoint.getTopicPartitionsToAssign();
+		if (topicPartitions != null && topicPartitions.length > 0) {
+			return new ShareKafkaMessageListenerContainer<>(getShareConsumerFactory(), new ContainerProperties(topicPartitions));
```
artembilan: And probably this one is not covered by tests.
So far we have there in the `ShareKafkaMessageListenerContainer` only this:
```
			// Subscribe to topics, just like in the test
			ContainerProperties containerProperties = getContainerProperties();
			this.consumer.subscribe(Arrays.asList(containerProperties.getTopics()));
```
Not sure why you say `just like in the test` though. Probably just remove that comment?

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,240 @@
...
+		else {
+			Collection<String> topics = endpoint.getTopics();
+			Assert.state(topics != null, "'topics' must not be null");
+			if (!topics.isEmpty()) {
+				return new ShareKafkaMessageListenerContainer<>(getShareConsumerFactory(),
+						new ContainerProperties(topics.toArray(new String[0])));
+			}
+			else {
+				return new ShareKafkaMessageListenerContainer<>(getShareConsumerFactory(),
+						new ContainerProperties(endpoint.getTopicPattern()));
```
artembilan: Probably `KafkaShareConsumer` does not support pattern subscription as well...
Please, revise.

spring-kafka/src/main/java/org/springframework/kafka/config/MethodKafkaListenerEndpoint.java
sobychacko: ya, that was an oversight. will fix it.

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
sobychacko: You are right, partition assignments are all handled on the server side, from what I understand.

spring-kafka/src/main/java/org/springframework/kafka/config/MethodKafkaListenerEndpoint.java
```diff
@@ -175,7 +175,8 @@ protected MessagingMessageListenerAdapter<K, V> createMessageListener(MessageLis
 
 		Assert.state(this.messageHandlerMethodFactory != null,
 				"Could not create message listener - MessageHandlerMethodFactory not set");
-		MessagingMessageListenerAdapter<K, V> messageListener = createMessageListenerInstance(messageConverter);
+
+		final MessagingMessageListenerAdapter<K, V> messageListener = createMessageListenerInstance(messageConverter);
```
artembilan: Still some unnecessary change for me to review? 😉

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,218 @@
...
+
+	@Override
+	public ShareKafkaMessageListenerContainer<K, V> createListenerContainer(KafkaListenerEndpoint endpoint) {
+		ShareKafkaMessageListenerContainer<K, V> instance = createContainerInstance(endpoint);
+		JavaUtils.INSTANCE
+				.acceptIfNotNull(endpoint.getId(), instance::setBeanName);
+		if (endpoint instanceof AbstractKafkaListenerEndpoint) {
+			configureEndpoint((AbstractKafkaListenerEndpoint<K, V>) endpoint);
+		}
+		endpoint.setupListenerContainer(instance, null); // No message converter for MVP
```
artembilan: What is `MVP`, please?

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaListenerIntegrationTests.java
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	static class TestListener {
+
+		@KafkaListener(topics = "share-listener-integration-test", containerFactory = "shareKafkaListenerContainerFactory")
+		public void listen(ConsumerRecord<String, String> record) {
+			received.set(record.value());
+			latch.countDown();
+		}
+	}
```
artembilan, line 151: Consider to use `src/idea/spring-framework.xml` for your IntelliJ IDEA editor config.
Then reformat classes, like this and `ShareKafkaMessageListenerContainer`.
So, all those missed blank lines around class members are going to be added.
Therefore, next time when I do some changes in the class would be consistent with what you've done to it.

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,221 @@
...
+ * @param <K> the key type
+ * @param <V> the value type
+ *
+ * @author Soby Chacko
+ * @since 4.0
+ */
+public class ShareKafkaListenerContainerFactory<K, V>
+		implements KafkaListenerContainerFactory<ShareKafkaMessageListenerContainer<K, V>>, ApplicationEventPublisherAware, ApplicationContextAware {
+
+	protected final LogAccessor logger = new LogAccessor(LogFactory.getLog(getClass()));
```
artembilan: Doesn't look like it has to be `protected`.
Also nit-pick, there is this ctor:
```
	/**
	 * Create a new accessor for the specified Commons Log category.
	 * @see LogFactory#getLog(Class)
	 */
	public LogAccessor(Class<?> logCategory) {
```

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,221 @@
...
+	@Override
+	public void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher) {
+		this.applicationEventPublisher = applicationEventPublisher;
+	}
+
+	/**
+	 * Get the container properties.
+	 * @return the container properties
+	 */
+	public ContainerProperties getContainerProperties() {
```
artembilan: Why do we need getter for this?

I even wonder if we need `this.containerProperties` at all...

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,221 @@
...
+		return this.containerProperties;
+	}
+
+	@Override
+	public ShareKafkaMessageListenerContainer<K, V> createListenerContainer(KafkaListenerEndpoint endpoint) {
+		ShareKafkaMessageListenerContainer<K, V> instance = createContainerInstance(endpoint);
+		JavaUtils.INSTANCE
+				.acceptIfNotNull(endpoint.getId(), instance::setBeanName);
+		if (endpoint instanceof AbstractKafkaListenerEndpoint) {
```
artembilan: This could be pattern expression:
```
if (endpoint instanceof AbstractKafkaListenerEndpoint abstractKafkaListenerEndpoint) {
```
And not cast in the next line.

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,221 @@
...
+	@Override
+	public void setApplicationContext(ApplicationContext applicationContext) {
+		this.applicationContext = applicationContext;
+	}
+
+	/**
+	 * Get the share consumer factory.
+	 * @return the share consumer factory
+	 */
+	public ShareConsumerFactory<? super K, ? super V> getShareConsumerFactory() {
```
artembilan: Why do we need getter for this?

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,221 @@
...
+		return this.shareConsumerFactory;
+	}
+
+	/**
+	 * Set the share consumer factory to use for creating containers.
+	 * @param shareConsumerFactory the share consumer factory
+	 */
+	public void setShareConsumerFactory(ShareConsumerFactory<? super K, ? super V> shareConsumerFactory) {
```
artembilan: Why do we need setter for this if we request it from the ctor?

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,221 @@
...
+		}
+		if (endpoint.getGroupId() != null) {
+			instance.getContainerProperties().setGroupId(endpoint.getGroupId());
+		}
+		if (endpoint.getClientIdPrefix() != null) {
+			instance.getContainerProperties().setClientId(endpoint.getClientIdPrefix());
+		}
+		if (endpoint.getConsumerProperties() != null) {
+			instance.getContainerProperties().setKafkaConsumerProperties(endpoint.getConsumerProperties());
+		}
```
artembilan: Let's see if the logic of this method could be reworked to `JavaUtils` API.

artembilan (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -0,0 +1,175 @@
...
+	 */
+	protected void initializeContainer(ShareKafkaMessageListenerContainer<K, V> instance, KafkaListenerEndpoint endpoint) {
+		ContainerProperties properties = instance.getContainerProperties();
+		JavaUtils.INSTANCE
+				.acceptIfNotNull(endpoint.getAutoStartup(), instance::setAutoStartup)
+				.acceptIfNotNull(this.autoStartup, autoStartup -> {
+					if (endpoint.getAutoStartup() == null) {
+						instance.setAutoStartup(autoStartup);
+					}
+				})
```
artembilan: Some is off with these two calls.
Both of them are about calling `instance.setAutoStartup()` in the end.
Can we calculate a single `autoStartup` variable before propagating it down?

artembilan (changes requested):

PR #3994, opened 2025-07-14
GH-3989 Consider the custom name of the reply topic name in sendAndRec…

### LINKED ISSUE #3989: ReplyingKafkaTemplate checks for default reply header name instead of custom one

3.3.6 (and most likely earlier and later versions as well)

**Describe the bug**

The `sendAndReceive` method of the `ReplyingKafkaTemplate` does not take the name of a user-defined reply header into account when it checks whether the header already exists.

https://github.com/spring-projects/spring-kafka/blob/b1d8836b7ca619c80cff1100363f5665e934758a/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java#L425C1-L431C4
```
		boolean hasReplyTopic = headers.lastHeader(KafkaHeaders.REPLY_TOPIC) != null;
		if (!hasReplyTopic && this.replyTopic != null) {
			headers.add(new RecordHeader(this.replyTopicHeaderName, this.replyTopic));
			if (this.replyPartition != null) {
				headers.add(new RecordHeader(this.replyPartitionHeaderName, this.replyPartition));
			}
		}
```

The check uses `KafkaHeaders.REPLY_TOPIC` to get the header and then uses `this.replyTopicHeaderName` to set the header. If a custom header name was provided earlier via `setReplyTopicHeaderName`, this name is not taken into consideration during the initial check.

This leads to duplicate headers (reply topic and reply partition) being set. In my case, this is particularly problematic because I have to provide the reply partition header in a different form - but the ReplyingKafkaTemplate overwrites it with its own header.

**To Reproduce**

Create a ReplyingKafkaTemplate and set a custom reply header

```
ReplyingKafkaTemplate<String, Message, Message> replyingKafkaTemplate = new ReplyingKafkaTemplate<>(producerFactory, repliesContainer);
replyingKafkaTemplate.setReplyTopicHeaderName("REPLY_TOPIC");
...
```

Create your own custom header

```
List<Header> headers = new ArrayList<>();
headers.add(new RecordHeader("REPLY_TOPIC", replyTopic.getBytes()));
```

Use the headers in the `ProducerRecord` and send it via the replyingKafkaTemplate

```
ProducerRecord<String, Message> producerRecord = new ProducerRecord<>("my-topic", null, null, key, value, headers);
replyingKafkaTemplate .sendAndReceive(producerRecord)
```

**Expected behavior**

Since the reply topic header is already provided it will not be set by the ReplyingKafkaTemplate. The reply partition header is also not set.

mipo256:
Closes #3989 

WIP, writing tests

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/requestreply/CorrelationKey.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/requestreply/CorrelationKey.java
@@ -35,8 +35,16 @@ public final class CorrelationKey {
 
 	private final byte[] correlationId;
 
+	/**
+	 * Cached hex representation of the {@link #correlationId}.
+	 * TODO: Migrate to stable values JEP 502
+	 */
 	private @Nullable String asString;
 
+	/**
+	 * Cached hash code.
+	 * TODO: Migrate to stable values JEP 502
+	 */
 	private volatile @Nullable Integer hashCode;
 
 	public CorrelationKey(byte[] correlationId) { // NOSONAR array reference
--- a/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
@@ -73,6 +73,7 @@ import org.springframework.util.Assert;
  * @author Artem Bilan
  * @author Borahm Lee
  * @author Francois Rosiere
+ * @author Mikhail Polivakha
  *
  * @since 2.1.3
  *
@@ -422,12 +423,13 @@ public class ReplyingKafkaTemplate<K, V, R> extends KafkaTemplate<K, V> implemen
 		CorrelationKey correlationId = this.correlationStrategy.apply(record);
 		Assert.notNull(correlationId, "the created 'correlationId' cannot be null");
 		Headers headers = record.headers();
-		boolean hasReplyTopic = headers.lastHeader(KafkaHeaders.REPLY_TOPIC) != null;
+		boolean hasReplyTopic = headers.lastHeader(this.replyTopicHeaderName) != null;
 		if (!hasReplyTopic && this.replyTopic != null) {
 			headers.add(new RecordHeader(this.replyTopicHeaderName, this.replyTopic));
-			if (this.replyPartition != null) {
-				headers.add(new RecordHeader(this.replyPartitionHeaderName, this.replyPartition));
-			}
+		}
+		boolean hasReplyPartition = headers.lastHeader(this.replyPartitionHeaderName) != null;
+		if (!hasReplyPartition && this.replyPartition != null) {
+			headers.add(new RecordHeader(this.replyPartitionHeaderName, this.replyPartition));
 		}
 		Object correlation = this.binaryCorrelation ? correlationId : correlationId.toString();
 		byte[] correlationValue = this.binaryCorrelation
--- a/spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
@@ -102,12 +102,14 @@ import static org.mockito.Mockito.verify;
  * @author Gary Russell
  * @author Nathan Xu
  * @author Soby Chacko
+ * @author Mikhail Polivakha
  * @since 2.1.3
  *
  */
 @SpringJUnitConfig
 @DirtiesContext
-@EmbeddedKafka(partitions = 5, topics = { ReplyingKafkaTemplateTests.A_REPLY, ReplyingKafkaTemplateTests.A_REQUEST,
+@EmbeddedKafka(partitions = 5, topics = {
+		ReplyingKafkaTemplateTests.A_REPLY, ReplyingKafkaTemplateTests.A_REQUEST,
 		ReplyingKafkaTemplateTests.B_REPLY, ReplyingKafkaTemplateTests.B_REQUEST,
 		ReplyingKafkaTemplateTests.C_REPLY, ReplyingKafkaTemplateTests.C_REQUEST,
 		ReplyingKafkaTemplateTests.D_REPLY, ReplyingKafkaTemplateTests.D_REQUEST,
@@ -119,7 +121,10 @@ import static org.mockito.Mockito.verify;
 		ReplyingKafkaTemplateTests.J_REPLY, ReplyingKafkaTemplateTests.J_REQUEST,
 		ReplyingKafkaTemplateTests.K_REPLY, ReplyingKafkaTemplateTests.K_REQUEST,
 		ReplyingKafkaTemplateTests.L_REPLY, ReplyingKafkaTemplateTests.L_REQUEST,
-		ReplyingKafkaTemplateTests.M_REPLY, ReplyingKafkaTemplateTests.M_REQUEST })
+		ReplyingKafkaTemplateTests.M_REPLY, ReplyingKafkaTemplateTests.M_REQUEST,
+		ReplyingKafkaTemplateTests.CUSTOM_REPLY_HEADER_REPLY, ReplyingKafkaTemplateTests.CUSTOM_REPLY_HEADER_REQUEST,
+		ReplyingKafkaTemplateTests.CUSTOM_REPLY_HEADER_WITH_PARTITION_REPLY, ReplyingKafkaTemplateTests.CUSTOM_REPLY_HEADER_WITH_PARTITION_REQUEST
+})
 public class ReplyingKafkaTemplateTests {
 
 	public static final String A_REPLY = "aReply";
@@ -173,6 +178,14 @@ public class ReplyingKafkaTemplateTests {
 	public static final String M_REPLY = "mReply";
 
 	public static final String M_REQUEST = "mRequest";
+
+	public static final String CUSTOM_REPLY_HEADER_REPLY = "CUSTOM_REPLY_HEADER_REPLY";
+
+	public static final String CUSTOM_REPLY_HEADER_REQUEST = "CUSTOM_REPLY_HEADER_REQUEST";
+
+	public static final String CUSTOM_REPLY_HEADER_WITH_PARTITION_REPLY = "CUSTOM_REPLY_HEADER_WITH_PARTITION_REPLY";
+
+	public static final String CUSTOM_REPLY_HEADER_WITH_PARTITION_REQUEST = "CUSTOM_REPLY_HEADER_WITH_PARTITION_REQUEST";
 
 	@Autowired
 	private EmbeddedKafkaBroker embeddedKafka;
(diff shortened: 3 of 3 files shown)


sobychacko:
@mipo256 PR build is failing, just letting you know.

mipo256:
Hey @sobychacko ! I know, thanks! I'll fix the checkstyle violations shortly and then ping you :)

mipo256:
Hey guys @sobychacko, @artembilan.

I've added some test cases for this scenario and introduced minor polishing, the PR is ready for review

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
```diff
@@ -33,6 +33,7 @@
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
+import com.google.common.collect.Iterables;
```
sobychacko: I wonder whether we can avoid using this. I searched the repository, and we are not using this anywhere else. Maybe we can simply look for the header in the code?

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
```diff
@@ -174,6 +180,18 @@ public class ReplyingKafkaTemplateTests {
 
 	public static final String M_REQUEST = "mRequest";
 
+	// GH-3989
```
sobychacko: I think we should avoid commenting this way using the GH issue number. Either leave it out, or describe what you want to comment there.

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
```diff
@@ -174,6 +180,18 @@ public class ReplyingKafkaTemplateTests {
...
+	public static final String CUSTOM_REPLY_HEADER_REPLY = "CUSTOM_REPLY_HEADER_REPLY";
+
+	// GH-3989
```
sobychacko: same as above.

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
```diff
@@ -174,6 +180,18 @@ public class ReplyingKafkaTemplateTests {
...
+	public static final String CUSTOM_REPLY_HEADER_REQUEST = "CUSTOM_REPLY_HEADER_REQUEST";
+
+	// GH-3989
```
sobychacko: ditto

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
```diff
@@ -174,6 +180,18 @@ public class ReplyingKafkaTemplateTests {
...
+	public static final String CUSTOM_REPLY_HEADER_WITH_PARTITION_REPLY = "CUSTOM_REPLY_HEADER_WITH_PARTITION_REPLY";
+
+	// GH-3989
```
sobychacko: ditto

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
```diff
@@ -1046,6 +1112,47 @@ public List<Message<String>> handleM(String in) throws InterruptedException {
 			return Collections.singletonList(message);
 		}
 
+		// GH-3989
```
sobychacko: describe the comment or remove it.

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
```diff
@@ -1046,6 +1112,47 @@ public List<Message<String>> handleM(String in) throws InterruptedException {
...
+
+			if (Iterables.size(headers.headers("X-Custom-Reply-Header")) != 1) {
+				return "The X-Custom-Reply-Header header that signify the custom reply topic header name is duplicated. It is supposed to present only once";
+			}
+
+			if (Iterables.size(headers.headers(KafkaHeaders.REPLY_PARTITION)) != 0) {
+				return "It is expected that the user does NOT specify the reply partition in this test case";
+			}
+
+			if (!"foo".equals(inputMessage.value())) {
```
sobychacko: Try to avoid things like `foo` and use more descriptive words.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/CorrelationKey.java
```diff
@@ -35,8 +35,16 @@ public final class CorrelationKey {
 
 	private final byte[] correlationId;
 
+	/**
+	 * Cached hex representation of the {@link #correlationId}.
+	 * TODO: Migrate to stable values JEP 502
```
sobychacko, line 40: Could you elaborate what you mean by migrate to JEP 502 here?

sobychacko (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/requestreply/CorrelationKey.java
mipo256: I mean that in JEP 502 we would have stable values in Java, which are lazily initilized `final` fields. The lazily computed hashcode and hex representation of correlation id is exactly that use case.

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
mipo256: This particular issue is about the duplication of headers in the Kafka message. We cannot just check the existence of the header in the message - we need to count the occurrences of it.

The alternative would be to write our own analog of `Iterables.size`, but I do not think it makes much sense.

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
mipo256: I just was used to it [in the Spring Data](https://github.com/spring-projects/spring-data-relational/blob/main/spring-data-jdbc/src/test/java/org/springframework/data/jdbc/repository/JdbcRepositoryIntegrationTests.java#L381). In Spring Data when bug is getting fixed for instance, the tests that verify the actual fix typically contain `GH` comment as a reference to the issue.

I think this approach makes sense, since it just adds the context about what exactly use case is targeted by particular test method. 

Either way, of course I do not insist on this. The final decision is on core maintainers of the spring-kafka.

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
mipo256: Sure, no problem

mipo256:
Hey @sobychacko, can you please re-review? I've answered you questions and fixed the comments that I think we both agree upon.

Thank you!

artembilan (changes requested):
Can we have the title of PR and commit message fixed for the typo?
That `Consder` word is wrong.
Thanks

build.gradle
```diff
@@ -1,3 +1,6 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
```
artembilan: This is not our style.
Please, bring back the previous code.

spring-kafka/src/main/java/org/springframework/kafka/listener/DeliveryAttemptAware.java
```diff
@@ -23,7 +23,6 @@
  *
  * @author Gary Russell
  * @since 2.5
- *
```
artembilan: Does not look like necessary change in this file.

mipo256:
Hey @artembilan, sure, it's done

spring-kafka/src/main/java/org/springframework/kafka/requestreply/RequestReplyFuture.java
```diff
@@ -26,7 +26,7 @@
 /**
  * A {@link CompletableFuture} for requests/replies.
  *
- * @param <K> the key type.
+ * @param <K> the key type of inbound kafka message.
```
artembilan: I don't think this change has something to do with the request in the issue.
Can we just revert it to make review easier?

spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
```diff
@@ -269,11 +270,7 @@ private boolean doesMatchInternal(String header) {
 	 * @since 2.2.5
 	 */
 	protected Object headerValueToAddOut(String key, Object value) {
-		Object valueToAdd = mapRawOut(key, value);
-		if (valueToAdd == null) {
-			valueToAdd = value;
-		}
-		return valueToAdd;
+		return Optional.ofNullable((Object) mapRawOut(key, value)).orElse(value);
```
artembilan: This is change is not relevant to the request.
Plus it is a critical path to use extra objects like `Optional` and `Stream`.
So, please, revert.

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
artembilan: This has nothing to do with the request.
Please, revert.
And we prefer to not use Guava in our code.

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
artembilan: The reality is not that clean as we intend.
Some future changes could be applied to the test class where original issue would not fully apply.
Plus we have a Git blame feature to see what commit has done changes.
Yes, that also has its won flaw to be lost with future changes, but tracing history helps anyway.
So, please, remove all of these comments and make these constants as not public.

artembilan (changes requested):
I don't know issue in details to be sure why we didn't have this feature for years, but it still worth to look into `sending-messages.adoc` to determine the reason.
And mention this new feature over there if that is what we would want.

Thanks

artembilan:
OK. I see @sobychacko said in the issue:

> not using `this.replyTopicHeaderName` and directly checking against `KafkaHeaders.REPLY_TOPIC` in that conditional seems like a bug. 

So, feels like this fix is supposed to be a back-port candidate.
Therefore, indeed, all the not relevant changes have to go.

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
mipo256: I'll change this, okay. Still, it is used to check the length of a given `Iterable` in the test for this issue.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
```diff
@@ -415,19 +417,17 @@ public RequestReplyFuture<K, V, R> sendAndReceive(ProducerRecord<K, V> record) {
 	@Override
 	public RequestReplyFuture<K, V, R> sendAndReceive(ProducerRecord<K, V> record, @Nullable Duration replyTimeout) {
 		Assert.state(this.running, "Template has not been started"); // NOSONAR (sync)
-		Duration timeout = replyTimeout;
-		if (timeout == null) {
-			timeout = this.defaultReplyTimeout;
-		}
+		Duration timeout = Optional.ofNullable(replyTimeout).orElse(this.defaultReplyTimeout);
```
artembilan: As I said in other place: it’s is not OK to use extra objects in the critical execution path of our code.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
```diff
@@ -415,19 +417,17 @@ public RequestReplyFuture<K, V, R> sendAndReceive(ProducerRecord<K, V> record) {
...
 		CorrelationKey correlationId = this.correlationStrategy.apply(record);
 		Assert.notNull(correlationId, "the created 'correlationId' cannot be null");
 		Headers headers = record.headers();
-		boolean hasReplyTopic = headers.lastHeader(KafkaHeaders.REPLY_TOPIC) != null;
-		if (!hasReplyTopic && this.replyTopic != null) {
+		boolean replyTopicIsUnset = headers.lastHeader(this.replyTopicHeaderName) == null;
```
artembilan: I don’t think the change in conditions and renaming variables does some impact on the logic.
Can we concentrate on the problem in hands?
As I said: we are aiming to back -port the fix.
So, making the change as small as possible would be good.

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
artembilan: I’m not sure how that relates to the issue. The length are supposed to be the same.

artembilan (changes requested):

mipo256:
With all due respect Artem @artembilan, I do not think you're fully in the context of the problem.

> I’m not sure how that relates to the issue. The length are supposed to be the same.

No, it is not. And it is the core of the issue. The custom reply header is considered already, and it works fine. The problem arises when the the custom reply header is overridden by the user via public API, and in this case the custom reply header is sent in the kafka message two times. **It's get duplicated**, see from the original issue:

> This leads to duplicate headers (reply topic and reply partition) being set. In my case, this is particularly problematic because I have to provide the reply partition header in a different form - but the ReplyingKafkaTemplate overwrites it with its own header.

I'm reading your code review and I think we have a misunderstanding here. I'll fix all your comments regarding the code style etc, no problem, I'm just making sure we're on the same page.

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
mipo256: If the task is to leave the names of the variables as they are, then sure, I'll do that.

mipo256:
> As I said: we are aiming to back -port the fix.

My 2 cents. I'm not sure it is a good backport candidate. 

This is the bug, and when it get's fixed, then hypothetically, when previously producer produced 4 headers in the message (two for custom reply header, and two for custom reply partition), now it will produce 4. 

For me at least, it is highly questionable if this change is backward compatible, since the consumers now will receive de-facto different messages (4 vs 2 headers). 

CC : @sobychacko @artembilan

spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
mipo256: Okay, no problem

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
mipo256: I think no, it is not, please, check my explanation below :handshake:

spring-kafka/src/main/java/org/springframework/kafka/requestreply/RequestReplyFuture.java
mipo256: Sure, done

sobychacko (approved):

sobychacko:
Merged via https://github.com/spring-projects/spring-kafka/commit/97f482a866b5c0776742034bcb0c67228ccb490c.

Issue #3930, opened 2025-05-28
Fix deprecation for the `MappingJacksonParameterizedConverter`


artembilan:
The `MappingJackson2MessageConverter` is deprecated for removal in Spring Framework `7.0` in favor of `JacksonJsonMessageConverter`.

Related to: https://github.com/spring-projects/spring-framework/issues/33798

artembilan:
Not yet.
See related Spring Framework issue: https://github.com/spring-projects/spring-framework/issues/34963

artembilan:
FYI, the required issue in the Framework has been fixed.
Feel free to contribute this migration to a new API!
Thanks

artembilan:
There was misunderstanding from my side.
The new `JacksonJsonMessageConverter` in Spring Framework is for Jackson 3.
While it is recommended for end-users to migrate to Jackson 3 and that is how we got an encourage from that deprecated `MappingJackson2MessageConverter`, but that does not mean that we have a freedom in library like this ours Spring for Apache Kafka and make decision for consumers of it.
Therefore we have to go here slightly different direction:

1. Manage Jackson 3 dependency.
2. Implement a `MappingJacksonParameterizedConverter` variant for that Jackson 3 based `JacksonJsonMessageConverter`
3. And deprecated existing for Jackson 2.

This way, even if we phasing out Jackson 2 support, that does not mean that we drop it off altogether.
After that we would need to think about all the places where we use Jackson 2 directly, e.g. `DefaultKafkaHeaderMapper` and `Json(De)Serializer`.
Looks like Spring Framework has Jackson 3 variants for all the possible converters and encoders, e.g. new `AbstractJacksonHttpMessageConverter`.

sobychacko:
Ya, eventually, I think we should migrate all our Jackson support to the new one.

sobychacko:
@chickenchickenlove We can look at this issue since we need to tackle a bunch of external dependencies, deprecations, etc., and then coordinate with the framework changes. Were you already in the process of making any changes?

chickenchickenlove:
@sobychacko 
Sorry to late 🙇‍♂ 
I just took a look code about how to deal with this problem and I didn't start making any changes with code. 
So, If you want to take this task, please go ahead. 
Sorry for inconvenience. 🙇‍♂

anthologia:
@sobychacko
Hi! I'm contributing to Jackson 3 migration in Spring Integration.
[...]

PR #3996, opened 2025-07-17
GH-3930: Add Jackson 3 support; deprecate Jackson 2

sobychacko:
Fixes: https://github.com/spring-projects/spring-kafka/issues/3930

* Manage  dependencies, similar way we do for Jackson 2
* Add Jackson 3 counterparts for existing Jackson 2 based classes (mostly copy/paste)
* Deprectate Jackson 2 classes
* Update tests
* Initial round of updates in docs

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,6 +56,7 @@ ext {
 	hamcrestVersion = '3.0'
 	hibernateValidationVersion = '8.0.2.Final'
 	jacksonBomVersion = '2.19.1'
+	jackson3Version = '3.0.0-rc5'
 	jaywayJsonPathVersion = '2.9.0'
 	junit4Version = '4.13.2'
 	junitJupiterVersion = '5.13.3'
@@ -110,6 +111,7 @@ allprojects {
 
 		imports {
 			mavenBom "com.fasterxml.jackson:jackson-bom:$jacksonBomVersion"
+			mavenBom "tools.jackson:jackson-bom:$jackson3Version"
 			mavenBom "org.junit:junit-bom:$junitJupiterVersion"
 			mavenBom "io.micrometer:micrometer-bom:$micrometerVersion"
 			mavenBom "io.micrometer:micrometer-tracing-bom:$micrometerTracingVersion"
@@ -260,6 +262,13 @@ project ('spring-kafka') {
 		optionalApi 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
 		optionalApi 'com.fasterxml.jackson.datatype:jackson-datatype-joda'
 		optionalApi ('com.fasterxml.jackson.module:jackson-module-kotlin') {
+			exclude group: 'org.jetbrains.kotlin'
+		}
+
+		optionalApi 'tools.jackson.core:jackson-databind'
+		optionalApi 'tools.jackson.datatype:jackson-datatype-joda'
+		optionalApi 'tools.jackson.dataformat:jackson-dataformat-xml'
+		optionalApi('tools.jackson.module:jackson-module-kotlin') {
 			exclude group: 'org.jetbrains.kotlin'
 		}
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultJacksonKafkaHeaderMapper.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/DefaultJacksonKafkaHeaderMapper.java
@@ -27,58 +27,56 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.header.internals.RecordHeader;
+import tools.jackson.core.type.TypeReference;
+import tools.jackson.databind.DeserializationFeature;
+import tools.jackson.databind.MapperFeature;
+import tools.jackson.databind.ObjectMapper;
+import tools.jackson.databind.json.JsonMapper;
 
 import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.converter.JacksonJsonMessageConverter;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 
 /**
- * Default header mapper for Apache Kafka.
+ * Default header mapper for Apache Kafka. Based on Jackson 3.
  * Most headers in {@link KafkaHeaders} are not mapped on outbound messages.
  * The exceptions are correlation and reply headers for request/reply
  * messaging.
  * Header types are added to a special header {@link #JSON_TYPES}.
  *
- * @author Gary Russell
- * @author Artem Bilan
  * @author Soby Chacko
- * @author Sanghyoek An
  *
- * @since 1.3
+ * @since 4.0
  *
- * @deprecated since 4.0 in favor of {@link DefaultJacksonKafkaHeaderMapper} for Jackson 3.
  */
-@Deprecated(forRemoval = true, since = "4.0")
-public class DefaultKafkaHeaderMapper extends AbstractKafkaHeaderMapper {
+public class DefaultJacksonKafkaHeaderMapper extends AbstractKafkaHeaderMapper {
 
 	private static final String JAVA_LANG_STRING = "java.lang.String";
 
 	private static final Set<String> TRUSTED_ARRAY_TYPES = Set.of(
-					"[B",
-					"[I",
-					"[J",
-					"[F",
-					"[D",
-					"[C"
-			);
+			"[B",
+			"[I",
+			"[J",
+			"[F",
+			"[D",
+			"[C"
+	);
 
 	private static final List<String> DEFAULT_TRUSTED_PACKAGES = List.of(
-					"java.lang",
-					"java.net",
-					"java.util",
-					"org.springframework.util"
-			);
+			"java.lang",
+			"java.net",
+			"java.util",
+			"org.springframework.util"
+	);
 
 	private static final List<String> DEFAULT_TO_STRING_CLASSES = List.of(
-					"org.springframework.util.MimeType",
-					"org.springframework.http.MediaType"
-			);
+			"org.springframework.util.MimeType",
+			"org.springframework.http.MediaType"
+	);
 
 	/**
 	 * Header name for java types of other headers.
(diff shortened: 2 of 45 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/DefaultJacksonKafkaHeaderMapper.java
```diff
@@ -0,0 +1,503 @@
...
+	 * preceding them with "!". The patterns will replace the default patterns; you
+	 * generally should not map the {@code "id" and "timestamp"} headers. Note:
+	 * most of the headers in {@link KafkaHeaders} are ever mapped as headers since they
+	 * represent data in consumer/producer records.
+	 * @param patterns the patterns.
+	 * @see org.springframework.util.PatternMatchUtils#simpleMatch(String, String)
+	 */
+	public DefaultJacksonKafkaHeaderMapper(String... patterns) {
+		this(JsonMapper.builder()
+				.findAndAddModules(JacksonJsonMessageConverter.class.getClassLoader())
```
artembilan: The class loader from this class exactly, not something else to avoid class tangle and so.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultJacksonKafkaHeaderMapper.java
```diff
@@ -0,0 +1,503 @@
...
+		Assert.notNull(objectMapper, "'objectMapper' must not be null");
+		Assert.noNullElements(patterns, "'patterns' must not have null elements");
+		this.objectMapper = objectMapper;
+	}
+
+	/**
+	 * Create an instance for inbound mapping only with pattern matching.
+	 * @param patterns the patterns to match.
+	 * @return the header mapper.
+	 * @since 2.8.8
```
artembilan: No `@since` on methods in all new classes, please.

spring-kafka/src/main/java/org/springframework/kafka/support/DefaultJacksonKafkaHeaderMapper.java
```diff
@@ -0,0 +1,503 @@
...
+		if (String.class.equals(type) && (header.value().length == 0 || header.value()[0] != '"')) {
+			headers.put(header.key(), new String(header.value(), getCharset()));
+		}
+		else {
+			if (trusted) {
+				try {
+					Object value = decodeValue(header, type);
+					headers.put(header.key(), value);
+				}
+				catch (IOException e) {
```
artembilan: This should be revised since Jackson 3 does not throw `IOException` any more.
May just `catch (Exception e) {` instead?

spring-kafka/src/main/java/org/springframework/kafka/support/converter/BatchMessagingMessageConverter.java
```diff
@@ -101,6 +103,9 @@ public BatchMessagingMessageConverter(@Nullable RecordMessageConverter recordCon
 		if (JacksonPresent.isJackson2Present()) {
 			this.headerMapper = new DefaultKafkaHeaderMapper();
 		}
+		else if (JacksonPresent.isJackson3Present()) {
```
artembilan: Let's prefer Jackson 3!
If they have both in classpath, so Jackson 3 choice would win.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/JacksonJsonMessageConverter.java
```diff
@@ -0,0 +1,151 @@
...
+import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
+
+/**
+ * Base class for JSON message converters that use Jackson 3; on the consumer side, it can
+ * handle {@code byte[]}, {@link Bytes} and {@link String} record values.
+ * On the producer side, select a subclass that matches the corresponding
+ * Kafka Serializer.
+ *
+ * @author Soby Chacko
```
artembilan, line 44: Please, honor authors of class we are deprecating.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/JacksonJsonMessageConverter.java
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * @since 4.0
+ */
+public class JacksonJsonMessageConverter extends MessagingMessageConverter {
+
+	private final ObjectMapper objectMapper;
+
+	private JacksonJavaTypeMapper typeMapper = new DefaultJacksonJavaTypeMapper();
+
+	private final TypeFactory typeFactory = TypeFactory.createDefaultInstance();
```
artembilan: Let's use `ObjectMapper.getTypeFactory()` instead in the ctor!

spring-kafka/src/main/java/org/springframework/kafka/support/converter/JacksonProjectingMessageConverter.java
```diff
@@ -0,0 +1,200 @@
...
+import org.springframework.data.projection.MethodInterceptorFactory;
+import org.springframework.data.projection.ProjectionFactory;
+import org.springframework.data.projection.SpelAwareProxyProjectionFactory;
+import org.springframework.data.web.JsonProjectingMethodInterceptorFactory;
+import org.springframework.kafka.support.KafkaNull;
+import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
+
+/**
+ * A {@link MessageConverter} implementation based on Jacthat uses a Spring Data
```
artembilan: You probably meant `based on Jackson 3` 😄

build.gradle
```diff
@@ -263,6 +265,13 @@ project ('spring-kafka') {
 			exclude group: 'org.jetbrains.kotlin'
 		}
 
+		optionalApi 'tools.jackson.core:jackson-databind'
+		optionalApi 'tools.jackson.datatype:jackson-datatype-joda'
+		optionalApi 'tools.jackson.dataformat:jackson-dataformat-xml'
```
artembilan: We don't do XML here in Spring Kafka, so let's just remove this redundant dependency!

build.gradle
```diff
@@ -55,7 +55,8 @@ ext {
 	awaitilityVersion = '4.3.0'
 	hamcrestVersion = '3.0'
 	hibernateValidationVersion = '9.0.1.Final'
-	jacksonBomVersion = '2.19.2'
+	jacksonBomVersion = '2.19.1'
```
artembilan: Why downgrade?

spring-kafka/src/main/java/org/springframework/kafka/support/Jackson3Utils.java
```diff
@@ -0,0 +1,55 @@
...
+	 */
+	public static ObjectMapper enhancedObjectMapper() {
+		ObjectMapper objectMapper = JsonMapper.builder()
+				.findAndAddModules(JsonKafkaHeaderMapper.class.getClassLoader())
+				.disable(tools.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+				.disable(tools.jackson.databind.MapperFeature.DEFAULT_VIEW_INCLUSION)
+				.enable(SerializationFeature.FAIL_ON_EMPTY_BEANS)
+				.addModule(new Jackson3MimeTypeModule())
+				.build();
+		return objectMapper;
```
artembilan: Why do we need local variable?
Why cannot we just use `return` forthe builder result?

spring-kafka/src/main/java/org/springframework/kafka/support/Jackson3Utils.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public final class Jackson3Utils {
+
+	/**
+	 * Factory for {@link ObjectMapper} instances with registered well-known modules
+	 * and disabled {@link MapperFeature#DEFAULT_VIEW_INCLUSION} and
+	 * {@link DeserializationFeature#FAIL_ON_UNKNOWN_PROPERTIES} features.
+	 * @return the {@link ObjectMapper} instance.
+	 */
+	public static ObjectMapper enhancedObjectMapper() {
```
artembilan: See my general comment.
Apparently we have to be as specific as possible.
And make this as `JsonMapper` instead.

spring-kafka/src/main/java/org/springframework/kafka/support/JacksonUtils.java
```diff
@@ -29,7 +29,10 @@
  * @author Artem Bilan
  *
  * @since 2.3
+ *
+ * @deprecated since 4.0 in favor of native Jackson 3 {@link tools.jackson.databind.json.JsonMapper#builder()} API.
```
artembilan: And now this is not correct since you have just introduced `Jackson3Utils`.

spring-kafka/src/main/java/org/springframework/kafka/support/Jackson3Utils.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * The utilities for Jackson {@link ObjectMapper} instances.
+ *
+ * @author Artem Bilan
+ * @author Soby Chacko
+ *
+ * @since 4.0
+ */
+public final class Jackson3Utils {
```
artembilan: The name is not what is required from the framework perspective.
The goal is to avoid numbers as much as possible.
How about to name this class as `JacksonMapperUtils` instead?

spring-kafka/src/main/java/org/springframework/kafka/support/JacksonMimeTypeModule.java
```diff
@@ -31,6 +31,8 @@
  * @author Artem Bilan
  *
  * @since 2.3
+ *
+ * @deprecated since 4.0 in favor of {@link Jackson3Utils}.
```
artembilan: Not correct link.
And class is not marked with `@Deprecated`.

spring-kafka/src/main/java/org/springframework/kafka/support/Jackson3MimeTypeModule.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * A {@link SimpleModule} extension for {@link MimeType} serialization.
+ *
+ * @author Artem Bilan
+ * @author Soby Chacko
+ *
+ * @since 4.0
+ */
+public final class Jackson3MimeTypeModule extends SimpleModule {
```
artembilan: How about `MimeTypeJacksonModule` instead?
Just to avoid that number!

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
```diff
@@ -94,6 +96,9 @@ public MessagingMessageConverter(Function<Message<?>, @Nullable Integer> partiti
 		if (JacksonPresent.isJackson2Present()) {
 			this.headerMapper = new DefaultKafkaHeaderMapper();
 		}
+		else if (JacksonPresent.isJackson3Present()) {
```
artembilan: I think Jackson 3 has to be in priority if both are on classpath.

spring-kafka/src/main/java/org/springframework/kafka/support/JsonKafkaHeaderMapper.java
```diff
@@ -0,0 +1,485 @@
+/*
+ * Copyright 2017-present the original author or authors.
```
artembilan: And no this is not right.
This class was not started back in that year.
Just nit-picking thought, since I know that Spring Boot uses a year of the project beginning for everything 🤷

artembilan:
FYI, the refinement has already happened in Spring AMQP: https://github.com/spring-projects/spring-amqp/issues/3151.

The fix might be useful to follow here as well.

spring-kafka/src/main/java/org/springframework/kafka/support/JacksonMapperUtils.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+public final class JacksonMapperUtils {
+
+	/**
+	 * Factory for {@link ObjectMapper} instances with registered well-known modules
+	 * and disabled {@link MapperFeature#DEFAULT_VIEW_INCLUSION} and
+	 * {@link DeserializationFeature#FAIL_ON_UNKNOWN_PROPERTIES} features.
+	 * @return the {@link JsonMapper} instance.
+	 */
+	public static JsonMapper jsonMapper() {
```
artembilan: Why cannot we call `enhanced` any more?

artembilan (changes requested):

PR #4016, opened 2025-07-26
Fix SpEL/placeholder resolution in `@RetryableTopic.includeNames` and `@RetryableTopic.excludeNames`

NipponSunrise:
Previously, the `@RetryableTopic.includeNames` and `@RetryableTopic.excludeNames` attributes were not resolved from property placeholders `(${...})` or SpEL expressions `(#{...})`, despite the documentation stating that all String properties support such resolution.

This commit ensures proper resolution via `BeanExpressionResolver` and `BeanFactory`, consistent with the behavior of other annotations like `@KafkaListener`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
@@ -19,6 +19,7 @@ package org.springframework.kafka.annotation;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


artembilan:
>  despite the documentation stating that all String properties support such resolution.

Can you point us, please, where Spring for Apache Kafka documentation says that?

NipponSunrise:
> Can you point us, please, where Spring for Apache Kafka documentation says that?

Certainly — the Javadoc of [`@RetryableTopic`](https://github.com/spring-projects/spring-kafka/blob/1bdfc4b3e3e0d5709f8d16cbf9c56e5131e34add/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopic.java#L36) includes the statement:

> All String properties can be resolved from property placeholders ${...} or SpEL expressions #{...}.

While `includeNames` and `excludeNames` are technically `String[]` rather than `String`, it’s a reasonable and common expectation that this applies to all string-valued attributes — including arrays. This interpretation is also consistent with how similar annotations behave, such as `@KafkaListener`, where array properties like `topics` support both property placeholders and SpEL expressions on a per-element basis.

Enabling resolution here improves consistency with `@KafkaListener`, reduces potential surprises, and allows for cleaner externalization of retryable or non-retryable exception lists.

artembilan:
Thank you for explanation and contribution!

will be merged once PR build is green.

Looking forward for more!

NipponSunrise:
The pull request build failed due to a formatting issue — specifically, a `} else if {` on the same line.

I've corrected the formatting accordingly.

artembilan:
Thank you, @NipponSunrise , for contribution; looking forward for more!

Issue #4025, opened 2025-07-26
Refactor newTopics() to Reduce Redundant Iteration and Improve Filtering Logic


cwangg897:
### Expected Behavior
The newTopics() method should efficiently build and filter the topic collection by minimizing redundant iterations and clearly separating lookup, filtering, and removal steps. Duplicate topic removal logic should be efficient and readable, leveraging auxiliary data structures like maps for O(1) lookups.

### Current Behavior
The current implementation uses nested loops with O(n²) time complexity, performing multiple iterations over topic collections and in-place removals within loops. This leads to reduced readability, potential performance overhead for larger topic sets, and increased maintenance complexity.

### Context
This performance bottleneck affects application startup time when dealing with large topic configurations. The goal is to optimize the topic deduplication logic from O(n²) to O(n) complexity while maintaining code clarity.
Alternative approaches like end-to-end Stream API were considered, but the stepwise map-based approach better expresses the intent of each filtering stage.

PR #4026, opened 2025-07-26
Refactor newTopics to remove nested loops

cwangg897:
Replace O(n²) nested iteration with O(n) map-based lookups for filtering topics Use topicNameToMapKey to efficiently resolve duplicate topic names This improves application startup time and scalability when managing a large number of topics, while preserving the original filtering logic

Resolves: #4025

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
@@ -22,7 +22,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


artembilan:
Thank you for contribution !
We will review this next week.
Just for future convenience: if you have some implementation or fix in hands, no need to create an issue upfront. GitHub and we treat both issues and pull requests equally. No need in redundant entity when we simply can discuss directly on the PR. The commit message, though, is more important as I said in some of our previous discussion.

artembilan (changes requested):
Do you have proof how bad is a performance?
How many topics should we have to suffer from the logic we have so far?
I see that you have extracted a new local variable and you do have a new extra loop.
I'm not sure how that help with what we have so far.
I do see an improvement in the end of method though:
```
		newTopicsMap.entrySet().removeIf(next -> !this.createOrModifyTopic.test(next.getValue()));
		return new ArrayList<>(newTopicsMap.values());
```
And it feels like those duplicated `TopicForRetryable` instances also could be removed over here with that `removeIf()` callback.

Please, elaborate.

cwangg897:
> Do you have proof how bad is a performance? How many topics should we have to suffer from the logic we have so far? I see that you have extracted a new local variable and you do have a new extra loop. I'm not sure how that help with what we have so far. I do see an improvement in the end of method though:
> 
> ```
> 		newTopicsMap.entrySet().removeIf(next -> !this.createOrModifyTopic.test(next.getValue()));
> 		return new ArrayList<>(newTopicsMap.values());
> ```
> 
> And it feels like those duplicated `TopicForRetryable` instances also could be removed over here with that `removeIf()` callback.
> 
> Please, elaborate.

I am attaching the results of the benchmark performance test.
There does not appear to be a significant change in execution time.
However, the time complexity of the code has been reduced. 
 
The above is before, and the below is after the code change
 
image

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -333,30 +332,22 @@ protected Collection<NewTopic> newTopics() {
 		Map<String, NewTopic> topicsForRetry = newTopicsMap.entrySet().stream()
 				.filter(entry -> entry.getValue() instanceof TopicForRetryable)
 				.collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+		Map<String, String> topicNameToMapKey = new HashMap<>();
```
artembilan: I still don't believe in this extra map.
More over its name does not reflect reality.
That `this.applicationContext.getBeansOfType()` returns for us a map of bean names to beans.
Doesn't look like we deal with bean names over here in this method at all.
So, we could simplify the logic just to deal with a `List<NewTopic>` in this method.
Then those `topicsForRetry` could be just topic names, since according to its logic we deliberately aware that those names are only for `TopicForRetryable`.
And so on from here, down to a single `removeIf()`  in the end.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -322,33 +322,28 @@ private void updateClusterId(Admin adminClient) throws InterruptedException, Exe
...
+		// Deal with List<NewTopic> directly instead of Map (no need for bean names)
+		List<NewTopic> newTopicsList = new ArrayList<>(
+				this.applicationContext.getBeansOfType(NewTopic.class, false, false).values());
+
+		// Add topics from NewTopics wrappers (no need for bean names either)
+		this.applicationContext.getBeansOfType(NewTopics.class, false, false).values()
+				.forEach(wrapper -> newTopicsList.addAll(wrapper.getNewTopics()));
+
+		// Collect normal topic names to check against TopicForRetryable
+		Set<String> normalNames = newTopicsList.stream()
```
artembilan: Just nit-pick: I don't think "normal" is a proper term to use.
And not only here.
It raises a question like what is the "abnormal"?
Is a `TopicForRetryable` abnormal?

I would suggest to go with a "regular" instead.
This way a `TopicForRetryable` is really "unregular" and that does not cause extra questions.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -322,33 +322,28 @@ private void updateClusterId(Admin adminClient) throws InterruptedException, Exe
...
+				.filter(nt -> !(nt instanceof TopicForRetryable))
+				.map(NewTopic::name)
+				.collect(Collectors.toSet());
+
+		// Remove TopicForRetryable if there's a regular NewTopic with same name
+		newTopicsList.removeIf(nt -> nt instanceof TopicForRetryable && normalNames.contains(nt.name()));
+
+		// Apply predicate filter
+		newTopicsList.removeIf(nt -> !this.createOrModifyTopic.test(nt));
```
artembilan: Cannot we combine both conditions in the same `removeIf()`?
This way we would perform only one loop for removal logic.

artembilan (changes requested):
Thank you for an update!
Let me know what do you think about the latest review!

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
cwangg897: Thank you very much!
I hadn't thought about it that way, but your point makes a lot of sense.
I'll update the code to reflect your feedback.

spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
cwangg897: By the way, I'm curious—should these kinds of cases always be combined into one? I separated them intentionally for readability. In what situations is it better to merge them, and what are your thoughts on this?

artembilan: Right. I wouldn't merge then if you would not raise a performance concern 😄

PR #4027, opened 2025-07-28
Fix updateClusterId logic

cwangg897:
Change condition from != null to == null to only fetch clusterId 
when it's not already set. 
This avoids unnecessary broker calls and improves 
performance while maintaining consistency with clusterId() method

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
@@ -81,6 +81,7 @@ import org.springframework.util.Assert;
  * @author Valentina Armenise
  * @author Anders Swanson
  * @author Omer Celik
+ * @author Choi Wang Gyu
  *
  * @since 1.3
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
```diff
@@ -300,7 +301,7 @@ public final boolean initialize() {
 	private void updateClusterId(Admin adminClient) throws InterruptedException, ExecutionException, TimeoutException {
 		try {
 			this.clusterIdLock.lock();
-			if (this.clusterId != null) {
+			if (this.clusterId == null) {
```
artembilan, line 304: The original change is this: https://github.com/spring-projects/spring-kafka/commit/c84d0f7cdcf3c9dde35facfcaf90254fca3007c8.
According to its logic this indeed has to be `== null`.
So, i treat this as a simple omission and schedule for back-port.
Thank you for spotting and fixing!

PR #4031, opened 2025-07-30
GH-4008: Improve ReplyingKafkaTemplateTests with modular validation

### LINKED ISSUE #4008: Test improvement in ReplyingKafkaTemplateTests
Improve these tests, so that we don't return the actual headers, but maybe `Message<?>` and then iterate on the headers externally in the tests. 

https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java#L1120

https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java#L1140

These tests were originally added via this commit: https://github.com/spring-projects/spring-kafka/commit/97f482a866b5c0776742034bcb0c67228ccb490c

snowykte0426:
## Summary
  This PR addresses issue #4008 by refactoring `ReplyingKafkaTemplateTests`
   to improve test maintainability and readability through better
  separation of concerns.

  - Refactored listener methods to return `Message<?>` objects instead of
  handling validation internally
  - Extracted header validation logic into separate utility methods
  (`validateHeaders`, `validateSuccessfulResponse`)
  - Simplified test methods by moving validation concerns outside of
  listener logic

  ## Changes Made
  - **Listener Methods**: Simplified
  `handleCustomReplyHeaderNoReplyPartition` and
  `handleCustomReplyHeaderDefaultPartitionHeader` to focus only on business
   logic
  - **Validation Utilities**: Added `validateHeaders()` and
  `validateSuccessfulResponse()` methods for external header verification
  - **Test Methods**: Updated `testCustomReplyTopicHeaderIsNotDuplicated`
  and `testCustomReplyHeadersAreNotDuplicated` to use the new validation
  approach

  Fixes #4008

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
@@ -392,7 +392,9 @@ public class ReplyingKafkaTemplateTests {
 			RequestReplyMessageFuture<Integer, String> future = template.sendAndReceive(message, Duration.ofSeconds(30));
 			future.getSendFuture().get(10, TimeUnit.SECONDS); // send ok
 			Message<?> resultingMessage = future.get(30, TimeUnit.SECONDS);
+
 			assertThat(resultingMessage.getPayload()).isEqualTo("OK");
+			assertThat(resultingMessage.getHeaders()).containsEntry("originalPayload", "expected_message");
 		}
 		finally {
 			template.stop();
@@ -418,7 +420,9 @@ public class ReplyingKafkaTemplateTests {
 			RequestReplyMessageFuture<Integer, String> future = template.sendAndReceive(message, Duration.ofSeconds(30));
 			future.getSendFuture().get(10, TimeUnit.SECONDS); // send ok
 			Message<?> resultingMessage = future.get(30, TimeUnit.SECONDS);
+
 			assertThat(resultingMessage.getPayload()).isEqualTo("OK");
+			assertThat(resultingMessage.getHeaders()).containsEntry("originalPayload", "expected_message");
 		}
 		finally {
 			template.stop();
@@ -932,14 +936,6 @@ public class ReplyingKafkaTemplateTests {
 		}
 	}
 
-	private static int length(Iterable<?> iterable) {
-		int counter = 0;
-		for (Object o : iterable) {
-			counter++;
-		}
-		return counter;
-	}
-
 	@Configuration
 	@EnableKafka
 	public static class Config {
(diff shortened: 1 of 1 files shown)


spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
```diff
@@ -932,12 +936,15 @@ void testMessageIterableReturn() throws Exception {
...
-		return counter;
+	private void validateHeaders(Message<?> message, String expectedPayload) {
+		assertThat(message.getPayload()).isEqualTo(expectedPayload);
+		assertThat(message.getHeaders()).containsKey("originalPayload");
+		assertThat(message.getHeaders().get("originalPayload")).isEqualTo("expected_message");
+	}
+
+	private void validateSuccessfulResponse(Message<?> message, String expectedPayload) {
+		assertThat(message.getPayload()).isEqualTo(expectedPayload);
+		assertThat(message.getHeaders()).containsKey("originalPayload");
```
artembilan: If we do assert this header in the `validateHeaders()`, then no need to duplicate it here.
And in the end we would not need this extra method then.
To be honest, it is hard for me to understand the test logic and maintain it afterwards when there are some custom methods. I really prefer to stick with just `assertThat()` in place.
And now imaging the situation that one of those tests would be modified eventually and this custom method (read `validateHeaders()`) would not work any more.
What I mean, that tests are essentially independent units of work, so trying to avoid duplication does not pay off with a maintenance time.

But that's my opinion.

artembilan (changes requested):
Thank you for refactoring this hard-to-read logic!

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
snowykte0426: > If we do assert this header in the `validateHeaders()`, then no need to duplicate it here. And in the end we would not need this extra method then. To be honest, it is hard for me to understand the test logic and maintain it afterwards when there are some custom methods. I really prefer to stick with just `assertThat()` in place. And now imaging the situation that one of those tests would be modified eventually and this custom method (read `validateHeaders()`) would not work any more. What I mean, that tests are essentially independent units of work, so trying to avoid duplication does not pay off with a maintenance time.
> 
> But that's my opinion.

Thank you for the feedback, @artembilan — you're absolutely right about keeping test assertions direct and clear.

I've updated the implementation in commit b87aa26 to remove the custom validation methods (`validateHeaders()` and `validateSuccessfulResponse()`) and replaced them with inline `assertThat()` statements. After reviewing the existing codebase and seeing similar tests use just a few repeated assertions, I realized the inline approach is much more intuitive.

This change makes each test’s intent immediately obvious without needing to jump to separate methods. Thanks again for pointing me in the right direction!

spring-kafka/src/test/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplateTests.java
```diff
@@ -392,7 +392,10 @@ public void testCustomReplyTopicHeaderIsNotDuplicated() throws Exception {
 			RequestReplyMessageFuture<Integer, String> future = template.sendAndReceive(message, Duration.ofSeconds(30));
 			future.getSendFuture().get(10, TimeUnit.SECONDS); // send ok
 			Message<?> resultingMessage = future.get(30, TimeUnit.SECONDS);
+			
 			assertThat(resultingMessage.getPayload()).isEqualTo("OK");
+			assertThat(resultingMessage.getHeaders()).containsKey("originalPayload");
+			assertThat(resultingMessage.getHeaders().get("originalPayload")).isEqualTo("expected_message");
```
artembilan: There is an API in AssertJ to do both at the same time:
```
assertThat(resultingMessage.getHeaders()).containsEntry("originalPayload", "expected_message");
```

snowykte0426: > There is an API in AssertJ to do both at the same time:
> 
> ```
> assertThat(resultingMessage.getHeaders()).containsEntry("originalPayload", "expected_message");
> ```

@artembilan Thank you for the suggestion! I’ve implemented your feedback in commit e8baed9 by using AssertJ’s containsEntry() method. This replaces the separate assertions with the more concise approach you recommended.

artembilan (changes requested):
Thank you for the update!
It is really very close to merge.
However I’ve noticed the problem:

> Signed-off-by: snowykte0426 <snowykte0426@naver.com>

this one has to come with your official name to satisfy all the legal requirements.
Please, revise your local Git client config.
Squash and push a commit with with right sign-off.

snowykte0426:
> Thank you for the update! It is really very close to merge. However I’ve noticed the problem:
> 
> > Signed-off-by: snowykte0426 [snowykte0426@naver.com](mailto:snowykte0426@naver.com)
> 
> this one has to come with your official name to satisfy all the legal requirements. Please, revise your local Git client config. Squash and push a commit with with right sign-off.

Thank you for the review feedback. I've squashed the commits and updated
  the Signed-off-by with my official name as requested (commit b965e86).

PR #4050, opened 2025-08-16
Fix Javadoc error

snicoll:
`main` is failing at the moment, this PR should hopefully restore the build


artembilan:
Thanks, Stephane !
Had something similar a couple days ago. Not sure why this one didn’t fail for me. Probably Gradle cache is not always friendly 🤪

PR #4059, opened 2025-08-22
Replace Spring Retry usage to core retry

snicoll:
This commit replaces Spring Retry by the core retry support introduced in Spring Framework 7. This is a breaking change mostly in configuration that is detailed below.

The main feature in Spring Kafka is BackOffValuesGenerator that generates the required BackOff values upfront. These are then managed by the listener infrastructure and Spring Retry is no longer involved.

Moving this code from BackOffPolicy to BackOff dramatically simplifies that class as Spring Framework's core API naturally provides this information without the need of an extra infrastructure.

From a configuration standpoint, Spring Kafka relies quite heavily on Spring Retry's `@Backoff`. As there is no equivalent, the annotation has been moved to Spring Kafka proper with the following improvements:

* Harmonized name (`@BackOff` instead of `@Backoff`).
* Revisited Javadoc.
* Support for expression evaluation and `java.util.Duration` format.

The creation of a `BackOff` instance from the annotation is now isolated in `BackOffFactory` and the relevant tests have been added. `RetryTopicConfigurationBuilder` is mostly backward-compatible but `uniformRandomBackoff` has been deprecated as we feel that its name does not convey what it actually does.

`RetryingDeserializer` no longer offer a `RecoveryCallback` but an equivalent function that takes `RetryException` as an input. This contains the exceptions thrown as well as the number of retry attempts.

The use of BinaryExceptionClassifier has been replaced by the newly introduced `ExceptionMatcher` that is a copy of the original algorithm with a polished API.

With the migration done, we believe that further improvements can be made here: `@BackOff` oddly looks like Spring Framework's `@Retryable`. As a matter of a fact, the `maxAttempts` and `includes`/`excludes` from `@RetryableTopic` are touching the same concepts. One option would be to open up `@Retryable` so that it can be used in more case.

Another area of improvement is that harmonization of BackOff as a term. It is named "Backoff" in several places, including in `@RetryableTopic`, and it would be nice if the concept had the same syntax everywhere.

With Spring Retry being completely removed, this commit also removes the dependency and any further references to it.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/serdes.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/serdes.adoc
@@ -400,9 +400,10 @@ ConsumerFactory cf = new DefaultKafkaConsumerFactory(myConsumerConfigs,
     new RetryingDeserializer(myUnreliableValueDeserializer, retryTemplate));
 ----
 
-Starting with version `3.1.2`, a `RecoveryCallback` can be set on the `RetryingDeserializer` optionally.
+A recovery callback be set on the `RetryingDeserializer`, to return a fallback object
+if all retries are exhausted.
 
-Refer to the https://github.com/spring-projects/spring-retry[spring-retry] project for configuration of the `RetryTemplate` with a retry policy, back off policy, etc.
+Refer to the https://github.com/spring-projects/spring-framework[Spring Framework] project for configuration of the `RetryTemplate` with a retry policy, back off, etc.
 
 
 [[messaging-message-conversion]]
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/BackOffFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/BackOffFactory.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2018-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.annotation;
+
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.core.retry.RetryPolicy;
+import org.springframework.format.annotation.DurationFormat;
+import org.springframework.format.datetime.standard.DurationFormatterUtils;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+import org.springframework.util.StringValueResolver;
+import org.springframework.util.backoff.FixedBackOff;
+
+/**
+ * Create a {@link org.springframework.util.backoff.BackOff} from the state of a
+ * {@link BackOff @BackOff} annotation.
+ *
+ * @author Stephane Nicoll
+ */
+final class BackOffFactory {
+
+	private static final long DEFAULT_DELAY = 1000;
+
+	private final @Nullable StringValueResolver embeddedValueResolver;
+
+	BackOffFactory(@Nullable StringValueResolver embeddedValueResolver) {
+		this.embeddedValueResolver = embeddedValueResolver;
+	}
+
+	/**
+	 * Create a {@link org.springframework.util.backoff.BackOff} instance based on the
+	 * state of the given {@link BackOff @Backff}. The returned backoff instance has
+	 * unlimited number of attempts as these are controlled externally.
+	 *
+	 * @param annotation the annotation to source the parameters from
+	 * @return a {@link org.springframework.util.backoff.BackOff}
+	 */
+	public org.springframework.util.backoff.BackOff createFromAnnotation(BackOff annotation) { // NOSONAR
+		Duration delay = resolveDuration("delay", () -> annotation.delay() == DEFAULT_DELAY
+				? annotation.value() : annotation.delay(), annotation::delayString);
+		Duration maxDelay = resolveDuration("maxDelay", annotation::maxDelay, annotation::maxDelayString);
+		double multiplier = resolveMultiplier(annotation);
+		Duration jitter = resolveDuration("jitter", annotation::jitter, annotation::jitterString);
+		if (maxDelay == Duration.ZERO && multiplier == 0 && jitter == Duration.ZERO) {
+			Assert.isTrue(!delay.isNegative(),
+					() -> "Invalid delay (%dms): must be >= 0.".formatted(delay.toMillis()));
+			return new FixedBackOff(delay.toMillis());
+		}
+		RetryPolicy.Builder retryPolicyBuilder = RetryPolicy.builder().maxAttempts(Long.MAX_VALUE);
+		retryPolicyBuilder.delay(delay);
+		if (maxDelay != Duration.ZERO) {
+			retryPolicyBuilder.maxDelay(maxDelay);
+		}
+		if (multiplier != 0) {
+			retryPolicyBuilder.multiplier(multiplier);
+		}
+		if (jitter != Duration.ZERO) {
+			retryPolicyBuilder.jitter(jitter);
+		}
+		return retryPolicyBuilder.build().getBackOff();
+	}
+
+	private Duration resolveDuration(String attributeName, Supplier<@Nullable Long> valueRaw,
+			Supplier<String> valueString) {
+		String resolvedValue = resolve(valueString.get());
+		if (StringUtils.hasLength(resolvedValue)) {
+			try {
+				return toDuration(resolvedValue, TimeUnit.MILLISECONDS);
+			}
+			catch (RuntimeException ex) {
+				throw new IllegalArgumentException(
+						"Invalid duration value for '%s': '%s'; %s".formatted(attributeName, resolvedValue, ex));
+			}
+		}
+		Long raw = valueRaw.get();
+		return (raw != null && raw != 0) ? Duration.ofMillis(raw) : Duration.ZERO;
+	}
+
+	private Double resolveMultiplier(BackOff annotation) {
+		String resolvedMultiplier = resolve(annotation.multiplierString());
+		if (StringUtils.hasLength(resolvedMultiplier)) {
+			try {
+				return Double.valueOf(resolvedMultiplier);
+			}
+			catch (NumberFormatException ex) {
+				throw new IllegalArgumentException(
+						"Invalid multiplier: '%s'; %s".formatted(resolvedMultiplier, ex));
+			}
+		}
+		return annotation.multiplier();
+	}
+
+	private @Nullable String resolve(String valueString) {
+		if (StringUtils.hasLength(valueString) && this.embeddedValueResolver != null) {
+			return this.embeddedValueResolver.resolveStringValue(valueString);
+		}
+		return valueString;
+	}
+
+	private static Duration toDuration(String valueToResolve, TimeUnit timeUnit) {
+		DurationFormat.Unit unit = DurationFormat.Unit.fromChronoUnit(timeUnit.toChronoUnit());
+		return DurationFormatterUtils.detectAndParse(valueToResolve, unit);
+	}
+
+}
--- a/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
@@ -18,26 +18,23 @@ package org.springframework.kafka.retrytopic;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 import org.jspecify.annotations.Nullable;
 
-import org.springframework.classify.BinaryExceptionClassifier;
-import org.springframework.classify.BinaryExceptionClassifierBuilder;
 import org.springframework.kafka.config.ConcurrentKafkaListenerContainerFactory;
 import org.springframework.kafka.core.KafkaOperations;
 import org.springframework.kafka.support.AllowDenyCollectionManager;
 import org.springframework.kafka.support.EndpointHandlerMethod;
-import org.springframework.retry.backoff.BackOffPolicy;
-import org.springframework.retry.backoff.ExponentialBackOffPolicy;
-import org.springframework.retry.backoff.ExponentialRandomBackOffPolicy;
-import org.springframework.retry.backoff.FixedBackOffPolicy;
-import org.springframework.retry.backoff.NoBackOffPolicy;
-import org.springframework.retry.backoff.SleepingBackOffPolicy;
-import org.springframework.retry.backoff.UniformRandomBackOffPolicy;
+import org.springframework.kafka.support.ExceptionMatcher;
 import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.backoff.BackOff;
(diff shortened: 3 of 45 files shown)


snicoll:
I've just noticed the samples are broken, `sample-04` in particular. Not sure how to deal with this given that it doesn't use the snapshot atm.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/serdes.adoc
```diff
@@ -400,9 +400,10 @@ ConsumerFactory cf = new DefaultKafkaConsumerFactory(myConsumerConfigs,
     new RetryingDeserializer(myUnreliableValueDeserializer, retryTemplate));
 ----
 
-Starting with version `3.1.2`, a `RecoveryCallback` can be set on the `RetryingDeserializer` optionally.
+A recovery callback be set on the `RetryingDeserializer`, to return a fallback object
+if all retries are exhausted.
```
artembilan: We prefer follow `One sentence per line` rule: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

spring-kafka/src/main/java/org/springframework/kafka/annotation/BackOffFactory.java
```diff
@@ -0,0 +1,124 @@
...
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+import org.springframework.util.StringValueResolver;
+import org.springframework.util.backoff.FixedBackOff;
+
+/**
+ * Create a {@link org.springframework.util.backoff.BackOff} from the state of a
+ * {@link BackOff @BackOff} annotation.
+ *
+ * @author Stephane Nicoll
```
artembilan, line 37: `@since 4.0`

spring-kafka/src/main/java/org/springframework/kafka/annotation/BackOffFactory.java
```diff
@@ -0,0 +1,124 @@
...
+
+	/**
+	 * Create a {@link org.springframework.util.backoff.BackOff} instance based on the
+	 * state of the given {@link BackOff @Backff}. The returned backoff instance has
+	 * unlimited number of attempts as these are controlled externally.
+	 *
+	 * @param annotation the annotation to source the parameters from
+	 * @return a {@link org.springframework.util.backoff.BackOff}
+	 */
+	public org.springframework.util.backoff.BackOff createFromAnnotation(BackOff annotation) { // NOSONAR
```
artembilan: The `public` is redundant here.
And I agree that no reason to make this class as `public`.

spring-kafka/src/main/java/org/springframework/kafka/support/ExceptionMatcher.java
```diff
@@ -0,0 +1,257 @@
...
+ * it isn't found amongst the registered exceptions. As for the allow list, any
+ * exception that extends or implements one of the registered exception will lead it
+ * to not match.</li>
+ * </ol>
+ * By default, only the given {@link Throwable} is inspected. To also search nested
+ * causes, {@link #traverseCauses} should be enabled.
+ *
+ * @author Stephane Nicoll
+ * @author Dave Syer
+ * @author Gary Russell
```
artembilan, line 48: `@since 4.0`: https://github.com/spring-projects/spring-framework/wiki/Code-Style#use-of-since

spring-kafka/src/main/java/org/springframework/kafka/support/ExceptionMatcher.java
```diff
@@ -0,0 +1,257 @@
...
+		Map<Class<? extends Throwable>, Boolean> cache = new HashMap<>();
+		for (Class<? extends Throwable> type : exceptionType) {
+			cache.put(type, value);
+		}
+		return cache;
+	}
+
+	/**
+	 * Create a matcher that matches any {@link Exception}, but not errors.
+	 *
```
artembilan: No blank lines in method Javadoc: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilderTests.java
```diff
@@ -107,6 +107,8 @@ void shouldSetNoBackoffPolicy() {
 	}
 
 	@Test
+	@Deprecated
+	@SuppressWarnings("removal")
```
artembilan, line 111: I think we should remove the test for deprecated API even now.

spring-kafka/src/test/java/org/springframework/kafka/streams/KafkaStreamsInteractiveQueryServiceTests.java
```diff
@@ -275,9 +274,7 @@ public KafkaStreamsInteractiveQueryService kafkaStreamsInteractiveQueryService(S
 			final KafkaStreamsInteractiveQueryService kafkaStreamsInteractiveQueryService =
 					new KafkaStreamsInteractiveQueryService(streamsBuilderFactoryBean);
 			RetryTemplate retryTemplate = new RetryTemplate();
-			retryTemplate.setBackOffPolicy(new FixedBackOffPolicy());
-			RetryPolicy retryPolicy = new SimpleRetryPolicy(3);
-			retryTemplate.setRetryPolicy(retryPolicy);
+			retryTemplate.setRetryPolicy(RetryPolicy.builder().maxAttempts(2).build());
```
artembilan: I found in other places that default `RetryPolicy` comes with one second delay between attempts.
It might be OK for real applications, but that is a bit painful for tests execution time.
So, I use this instead:
```
RetryPolicy.builder().maxAttempts(2).delay(Duration.ZERO).build())
```

artembilan (approved):
Thanks, Stephane!

I'm approving this change since all my feedback is minor, and we can address it on merge as subsequent commit.

We also would need to add your name to all the affected classes.
But a `whats-new.adoc` deserves respective paragraph to explain such a breaking change.

Let's hear from @sobychacko yet!

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/serdes.adoc
snicoll: Sorry that's an oversight on my end.

spring-kafka/src/main/java/org/springframework/kafka/annotation/BackOffFactory.java
snicoll: OK. In Spring Boot we don't add `@since` to non-public classes.

spring-kafka/src/main/java/org/springframework/kafka/annotation/BackOffFactory.java
snicoll: Good catch. There is a rule in Spring Boot if you want to enforce it. I think it might be ArchUnit. LMK if you're interested.

spring-kafka/src/main/java/org/springframework/kafka/support/ExceptionMatcher.java
snicoll: 👍

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilderTests.java
snicoll: I kept it as the deprecated code would be dead code and deprecating it doesn't mean nobody would call it. The test makes sure something else does not break it. Or do you mean removing the code and the test?

spring-kafka/src/test/java/org/springframework/kafka/streams/KafkaStreamsInteractiveQueryServiceTests.java
snicoll: 👍

snicoll:
Thanks for the review Artem. I've addressed some of that in https://github.com/spring-projects/spring-kafka/pull/4059/commits/ec94299cc5847d8f0fd6a0ae63ff1b5cd502145d - The only thing I am not sure is about the test testing the deprecated method, see above for more details.

spring-kafka/src/test/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilderTests.java
artembilan: Sure! That is just personal preference.

When I see a deprecation warning I assume that code is not supported anymore.
So, I migrate to a new API as soon as possible.

On the other hand, we don't do changes into deprecated code, therefore we may assume that behavior is the same as it was before deprecation.
Hence, removing tests for deprecated API does not change the status quo of that API: even if someone still calls, it really was covered by test before.

I may change my mind, though, if that is really found useful and helpful in other projects to keep those tests just for sake of upstream API heal check.

snicoll: It's of course ultimately up to you. I think my point is that not modifying the code and calling it as part of a test are two different things. 

The code may not change but what it instruments might. And without the test you won't spot the difference in behaviour. 

For something like this one, it doesn't seem very important though.

artembilan (approved):
LGTM.

I'm deferring to @sobychacko for further steps on this PR.

Thanks

sobychacko:
@snicoll The PR looks good - this is a great effort. As we previously discussed, replacing Spring Retry with the built-in retry efforts in SF 7 seems like the right approach here. The migration feels solid overall from what I can gather. 

I like the new isolated approach of `BackOff` annotation and further isolating its creation via the factory. You alluded to the fact that `BackOff` looks very similar to SF's `@Retryable`. Since the current `BackOff` in Spring Kafka does not contain any Kafka-specific features, are we considering moving it into the Spring Framework proper? Or is that not in the scope? I think this might also benefit other projects, as we would then have a unified `BackOff` story. Would you have any thoughts on that?

As expected, this is a significant breaking change for the downstream users, especially on configs, annotation switches, or `RetryingDeserializer` shifting from `RecoveryCallback` to function with `RetryException`. Existing projects might have non-trivial work to adjust to these changes. I know @artembilan  mentioned adding a section to the `whats-new` docs, but are we also considering providing migration guidelines with clear examples on how users can adapt to these breaking changes in their applications? It may be possible that people might overlook the new `BackOff` annotation and wonder what happened to the old `Backoff` they were used to. On that note, +1 for using `BackOff` instead of using it interchangeably with `Backoff`. 

Regarding the deprecation of `uniformRandomBackoff`, I see that you added a note stating there is no alternative in the javadocs. Since it is a public method, we may need to justify the deprecation without providing an alternative. Are there any similar functionalities available in the framework efforts?

Overall, this is cool stuff. The main thing I can think of is the migration docs with quick before/after examples, which will certainly help with a smooth user experience. Once things are merged and settled down, we certainly need to spend some cycles ensuring that there is no regression or gotchas from dropping Spring Retry altogether. 

Thank you!

snicoll:
> You alluded to the fact that BackOff looks very similar to SF's @Retryable. Since the current BackOff in Spring Kafka does not contain any Kafka-specific features, are we considering moving it into the Spring Framework proper?

No. Doing so would reintroduce the problems we're trying to solve here. Trying to accommodate to slight variations in projects in the core would recreate Spring Retry in the long run.

But here, that's not what I meant. Have you looked at `@Retryable`? It has [all the fields](https://github.com/spring-projects/spring-framework/blob/0c024439654e64939b5d1a9ecff145ed9eaf6c1f/spring-context/src/main/java/org/springframework/resilience/annotation/Retryable.java#L109-L228) of `BackOff` plus support for exceptions includex/excludes and max attempts that are in `@RetryableTopic`. 

So here, I am rather advocating whether there's an interest to streamline things in Spring Kafka to adapt to the new annotation in framework, rather than keeping a custom one. That's more efforts that I didn't want to engage in this PR as it could be rather significant. It would require some javadoc change in framework as now it's quite specific to method invocation.

> are we also considering providing migration guidelines with clear examples on how users can adapt to these breaking changes in their applications? 

Spring Boot has migration guides for new majors. IMO, the only thing that needs explaining is:

* `random` being replaced by `jitter` and how to use it.
* The semantic of `maxAttempts` on a retry policy, and the fact that the total invocations in case of failure will be 4 and not 3 by default.

The rest isn't so bad IMO. The replacement of the calback by a Function is actually adding values. Same for replacing from Spring Retry's callback to a dedicated callback in Spring Kafka.

> On that note, +1 for using BackOff instead of using it interchangeably with Backoff.

Can we please consider renaming the field on the annotation? It's still `backoff` and that looks a bit odd at the moment. I didn't want to take the decision of changing it.

> Regarding the deprecation of uniformRandomBackoff, I see that you added a note stating there is no alternative in the javadocs. Since it is a public method, we may need to justify the deprecation without providing an alternative. Are there any similar functionalities available in the framework efforts?

The justification isn't in the javadoc but the name doesn't make sense based on what it does. As for an alternative, just look at the body of the method, it's configuring the jitter which is only using Spring Framework's API and a bit of math :)

We didn't know if this feature is often used. If it is, then I suggest to reintroduce it and use the opportunity to fix the name. You can then have a new method with a more meaningful name being the replacement of the deprecated method.

Thanks for the review!

sobychacko:
>So here, I am rather advocating whether there's an interest to streamline things in Spring Kafka to adapt to the new annotation in framework, rather than keeping a custom one. That's more efforts that I didn't want to engage in this PR as it could be rather significant. It would require some javadoc change in framework as now it's quite specific to method invocation.

Indeed, this sounds like a good idea. This is something we can evaluate to see the impact once we merge this PR. 

Good with the migration guidelines in Boot. We can monitor on the Spring Kafka side to see if there is any feedback from the community side, but that is certainly down the road.

>Can we please consider renaming the field on the annotation? It's still backoff and that looks a bit odd at the moment. I didn't want to take the decision of changing it.

Sure, I think we can look into it post merge?

>We didn't know if this feature is often used. If it is, then I suggest to reintroduce it and use the opportunity to fix the name. You can then have a new method with a more meaningful name being the replacement of the deprecated method.

I don't know how widely this method is used, but the feature of retryable topics via non-blocking retries is pretty popular and may be utilized by many people, so I was just commenting from a public API perspective. I suggest we keep it deprecated and monitor for any potential impacts down the line.

Once again, thanks for this effort!

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
```diff
@@ -315,105 +316,103 @@ public RetryTopicConfigurationBuilder timeoutAfter(long timeout) {
...
 	}
 
 	/**
-	 * Configure a {@link UniformRandomBackOffPolicy}.
+	 * Configure a {@link BackOff} that applies  random delay between the specified
+	 * minimum interval and maximum interval.
 	 * @param minInterval the minimum interval.
 	 * @param maxInterval the maximum interval.
 	 * @return the builder.
+	 * @deprecated as of 4.0 with no replacement
```
artembilan, line 369: Well, looking into the `UniformRandomBackOffPolicy` in Spring Retry, it feels similar to the `ExponentialRandomBackOffPolicy`, but without a `multiplier`.
I would say it should be safe to state this one is deprecated in favor of `exponentialBackoff()` since we don't carry out `UniformRandomBackOffPolicy` abstraction anymore.

snicoll:
> Good with the migration guidelines in Boot. 

There might be a misunderstanding here. I mentioned Spring Boot has migration guides in major releases. However, we do not intend to document the changes in Spring Boot as it belongs here. We could, however, link to your migration guide from ours. Does that make sense?

spring-kafka/src/main/java/org/springframework/kafka/retrytopic/RetryTopicConfigurationBuilder.java
sobychacko: sounds good then.

snicoll:
I can also amend my commit to rename the backoff field if you like.

sobychacko:
> I can also amend my commit to rename the backoff field if you like.

@snicoll, please go ahead with that if you have time.

snicoll:
I've done that in https://github.com/spring-projects/spring-kafka/pull/4059/commits/ba8e05346eb3fbc3e2ef9e53e5db1a20b7cc8c07 - There are more places where we use the term "backoff" in the doc but those are largely subjective so I didn't change them.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/fde7d310d72e56e7fe2bf2cc1e489751e6efd09c.

Thank you @snicoll again!

Issue #3954, opened 2025-06-12
automatic topic creation fails with @PostConstruct


TrentonAdams:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.1.0 - though probably all as the main branch appears to be using SmartInitializingSingleton

**Describe the bug**

If I use @PostConstruct, and my bean depends on a NewTopic, and I've also created a KafkaAdmin bean, the topic is not automatically created.

I suspect the problem is the use of the SmartInitializingSingleton.  It only runs after all beans have finished being initialized, and @PostConstruct methods have ran.  

The workaround was to use a event listern on a context refreshed event, but I'm uncertain if that guarantees it's done before any business logic runs.  I have to verify.  But, the key point is that I've never encountered a problem like this with spring, it's very unexpected.  At the very least, in the documentation for how topics are automatically created, it should state that it doesn't work with @PostConstruct.

**To Reproduce**

Use @PostConstruct, and depend on your NewTopic bean.  You'll get a failure about the topic not existing.

**Expected behavior**

I was expecting, that because I depended on the NewTopic, it would be created, according to the documentation.  This is true, that works, but not if my bean uses @PostConstruct

**Sample**

I don't have one, but I suspect it's easy enough for devs to replicate.  If not, I can try and find the time to create one.

artembilan:
Show, please, at least, some config, so we will know how to model the test to reproduce.
Thanks

TrentonAdams:
I decided to just make a project...

https://github.com/TrentonAdams/spring-kafka-post-construct

docker compose up # in one terminal

In another terminal...
./mvnw clean install spring-boot:run -Dspring-boot.run.profiles=post-construct
./mvnw clean install spring-boot:run -Dspring-boot.run.profiles=application-ready

http://localhost:8080 for the kafka-ui

artembilan:
```
    @PostConstruct
    public void init() {
        // This will fail because the topic hasn't been created yet
        // The @PostConstruct runs before SmartInitializingSingleton
        log.info("Attempting to send message to topic: {}", postConstructTopic.name());
        kafkaTemplate.send(postConstructTopic.name(), "test-message-from-post-construct");
    }
```
You cannot do heavy network communication work from the `@PostConstruct`.
It has to be really done when an application context is ready.
Or via the mentioned event, or implementing `SmartLifecycle`.

> I've never encountered a problem like this with spring, it's very unexpected

[...]

PR #4069, opened 2025-08-30
GH-3954: Add warning about @PostConstruct timing with NewTopic beans

Chaedie:
Related Issue: #3954

Added a warning note to the topic configuration docs about `@PostConstruct` timing issues.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/sending-messages.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/sending-messages.adoc
@@ -74,15 +74,14 @@ The `execute` method provides direct access to the underlying javadoc:org.apache
 
 [NOTE]
 ====
-When sending messages from Spring components, avoid using `@PostConstruct` methods
-if you rely on automatic topic creation via `NewTopic` beans. `@PostConstruct` runs
-before the application context is fully ready, which may cause
-`UnknownTopicOrPartitionException` on clean brokers.
+When sending messages from Spring components, avoid using `@PostConstruct` methods if relying on automatic topic creation via `NewTopic` beans.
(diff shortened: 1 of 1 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/sending-messages.adoc
```diff
@@ -72,6 +72,19 @@ If the topic is configured to use `LOG_APPEND_TIME`, the user-specified timestam
 The `metrics` and `partitionsFor` methods delegate to the same methods on the underlying javadoc:org.apache.kafka.clients.producer.Producer[].
 The `execute` method provides direct access to the underlying javadoc:org.apache.kafka.clients.producer.Producer[].
 
+[NOTE]
+====
+When sending messages from Spring components, avoid using `@PostConstruct` methods
+if you rely on automatic topic creation via `NewTopic` beans. `@PostConstruct` runs
+before the application context is fully ready, which may cause
+`UnknownTopicOrPartitionException` on clean brokers.
```
artembilan: All good, but, please, consider to fix your change according to best practice rules we try to follow: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line.

It also would be great to make the text as "personless".
Use the language which would not be like pointing to someone.
In other words try to avoid "you", "we", "they" etc.

artembilan (changes requested):

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/sending-messages.adoc
Chaedie: Thanks for advice. 
I revised documentation.

PR #4072, opened 2025-09-04
Upgrade to Kafka 4.1.0 and update API usage

sobychacko:
- Bump kafkaVersion from 4.0.0 to 4.1.0 in build.gradle
- Remove broker properties no longer needed in embedded Kafka tests:

  * unstable.api.versions.enable=true
  * group.coordinator.rebalance.protocols=classic,share

- Replace AcknowledgeType.ACCEPT with commitSync() in share consumer tests
- Update SerializationIntegrationTests to use Plugin-based deserializer access

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
@@ -31,7 +31,6 @@ import org.apache.kafka.clients.admin.Admin;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AlterConfigOp;
 import org.apache.kafka.clients.admin.ConfigEntry;
-import org.apache.kafka.clients.consumer.AcknowledgeType;
 import org.apache.kafka.clients.consumer.ShareConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerConfig;
@@ -56,8 +55,6 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 @EmbeddedKafka(
 		topics = {"embedded-share-test", "embedded-share-distribution-test"}, partitions = 1,
 		brokerProperties = {
-				"unstable.api.versions.enable=true",
-				"group.coordinator.rebalance.protocols=classic,share",
 				"share.coordinator.state.topic.replication.factor=1",
 				"share.coordinator.state.topic.min.isr=1"
 		})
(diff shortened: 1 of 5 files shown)


spring-kafka/src/test/java/org/springframework/kafka/core/DefaultShareConsumerFactoryTests.java
```diff
@@ -248,7 +245,7 @@ private static List<String> runSharedConsumerTest(String topic, String groupId,
 						var records = consumer.poll(Duration.ofMillis(200));
 						for (var r : records) {
 							allReceived.add(r.value());
-							consumer.acknowledge(r, AcknowledgeType.ACCEPT);
+							consumer.commitSync(Duration.ofMillis(10000));
```
artembilan, line 250: Did you just generate commit message using AI? 😉 

I see "what" is done here in the code, but totally unclear "why".

Please, elaborate.

spring-kafka/src/test/java/org/springframework/kafka/support/serializer/SerializationIntegrationTests.java
```diff
@@ -63,8 +66,10 @@ void configurePreLoadedDelegates() {
 		props.setMessageListener(mock(MessageListener.class));
 		KafkaMessageListenerContainer<String, Object> container = new KafkaMessageListenerContainer<>(cFact, props);
 		container.start();
-		assertThat(KafkaTestUtils.getPropertyValue(container,
-				"listenerConsumer.consumer.delegate.deserializers.valueDeserializer"))
+		Plugin<Deserializer<?>> valueDeserializerPlugin = (Plugin<Deserializer<?>>) KafkaTestUtils.getPropertyValue(container,
+				"listenerConsumer.consumer.delegate.deserializers.valueDeserializerPlugin");
```
artembilan, line 71: Same here: "why".

Now imaging you would look into this commit in a month.
Would you understand what is going on?

artembilan (changes requested):
The change looks OK, but totally unclear what is going on.
Can we have answers, please, why do we have to do this changes?
And I then place that into the final commit message.

Thanks

artembilan:
BTW, I don't see any changes in the `main` code, so feels like it still would be compatible with Kafka `4.0`.
You made changes only in the `test`, which might be expected due to some internal changes in the broker or client.

sobychacko:
@artembilan, It came down to the fact that in 4.1.0, they introduced an implicit/explicit mode of ack (or changed the semantics of it). In the implicit mode (the default one), you are not allowed to explicitly call `acknowledge`, rather only commit (which by default is done on the next poll). Since in the test, we are not polling in a loop, the explicit call to commit was necessary. I updated the commit message and added some comments.

In the serializer test, they removed the deserializer field in preference to a plugin mechanism.

I will update the commit message and the comments. Thanks!

PR #4080, opened 2025-09-12
Configure interceptors on listener container factory

cfredri4:
This changes allows configuring interceptors on AbstractKafkaListenerContainerFactory in same way as on KafkaMessageListenerContainer, and changes the examle docs to show this instead.
I can revert the update to the docs if desired but I thought it to be more useful as an example since you normally have access to the listener container factory as a bean, but not to the listener container since it's not a bean.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/message-listener-container.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/message-listener-container.adoc
@@ -22,25 +22,24 @@ IMPORTANT: If the interceptor mutates the record (by creating a new one), the `t
 
 The `CompositeRecordInterceptor` and `CompositeBatchInterceptor` can be used to invoke multiple interceptors.
 
-Starting with version 4.0, `AbstractMessageListenerContainer` exposes `getRecordInterceptor()` and `getBatchInterceptor()` as public methods.
+Starting with version 4.0, `AbstractKafkaListenerContainerFactory` and `AbstractMessageListenerContainer` exposes `getRecordInterceptor()` and `getBatchInterceptor()` as public methods.
 If the returned interceptor is an instance of `CompositeRecordInterceptor` or `CompositeBatchInterceptor`, additional `RecordInterceptor` or `BatchInterceptor` instances can be added to it even after the container instance extending `AbstractMessageListenerContainer` has been created and a `RecordInterceptor` or `BatchInterceptor` has already been configured.
 The following example shows how to do so:
 
 [source, java]
 ----
-public void configureRecordInterceptor(KafkaMessageListenerContainer<Integer, String> container) {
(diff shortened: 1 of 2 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/message-listener-container.adoc
```diff
@@ -22,32 +22,39 @@ IMPORTANT: If the interceptor mutates the record (by creating a new one), the `t
 
 The `CompositeRecordInterceptor` and `CompositeBatchInterceptor` can be used to invoke multiple interceptors.
 
-Starting with version 4.0, `AbstractMessageListenerContainer` exposes `getRecordInterceptor()` and `getBatchInterceptor()` as public methods.
```
sobychacko, line 25: Why did you remove this? This is still available on the `AbstractMessageListenerContainer`, correct? In that case, we might still want to preserver this statement.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/message-listener-container.adoc
```diff
@@ -22,32 +22,39 @@ IMPORTANT: If the interceptor mutates the record (by creating a new one), the `t
...
+Starting with version 4.0, `AbstractKafkaListenerContainerFactory` exposes `getRecordInterceptor()` and `getBatchInterceptor()` as public methods.
 If the returned interceptor is an instance of `CompositeRecordInterceptor` or `CompositeBatchInterceptor`, additional `RecordInterceptor` or `BatchInterceptor` instances can be added to it even after the container instance extending `AbstractMessageListenerContainer` has been created and a `RecordInterceptor` or `BatchInterceptor` has already been configured.
-The following example shows how to do so:
+The following example shows how to do so in a `BeanPostProcessor`:
 
 [source, java]
 ----
-public void configureRecordInterceptor(KafkaMessageListenerContainer<Integer, String> container) {
```
sobychacko, line 31: I guess this example is still valid. Why are we removing it?

cfredri4:
@sobychacko I thought it was enough with a single example to illustrate the concept of adding interceptors afterwards, and so I changed the example to instead show how it's done with the listener container factory since that should be more universally applicable.

The reasoning being that the listener container factory is normally a bean, but only the few who manually create their own listener container would have access to it and could follow the previous example.

Should I add back the old example and show both?

sobychacko:
Since that old functionality is sill available (via the previous changes you made as public API's), I think its probably better to show both examples?

cfredri4:
@sobychacko the two examples became almost identical except for two lines with container vs containerFactory so I mostly reverted back to the original example (but there was an issue with the original example; if `previousInterceptor` already was a `CompositeRecordInterceptor` it would add itself to itself 🙃).
Please check again.

cfredri4:
@sobychacko gentle reminder.

PR #4087, opened 2025-09-15
Add comprehensive acknowledgment support for Kafka share consumers

sobychacko:
Implement explicit and implicit acknowledgment modes for share consumer containers, enabling fine-grained control over record processing outcomes with ACCEPT, RELEASE, and REJECT acknowledgment types.

- Add ShareAcknowledgment interface with ACCEPT/RELEASE/REJECT support
- Add ShareAcknowledgmentException for acknowledgment failures
- Implement ShareAcknowledgmentMode enum (IMPLICIT/EXPLICIT) in ContainerProperties
- Add poll-level acknowledgment constraints in explicit mode

- Add ShareConsumerAwareMessageListener for ShareConsumer access
- Add AcknowledgingShareConsumerAwareMessageListener for explicit acknowledgment
- Add ShareRecordMessagingMessageListenerAdapter for KafkaListener integration
- Use non-polymorphic onShareRecord method names to avoid regression issues with existing listener infrastructure and maintain clear API separation

- Enhanced ShareKafkaMessageListenerContainer with acknowledgment tracking
- Automatic error handling with REJECT acknowledgment on exceptions
- Poll blocking in explicit mode until all records acknowledged
- Support for mixed acknowledgment patterns within single poll

- Auto-detection of ShareKafkaListenerContainerFactory for share consumer endpoints
- Validation preventing batch listeners with share consumers
- Factory-level and container-level acknowledgment mode configuration
- Message converter extensions for ShareAcknowledgment parameter injection

- Comprehensive integration tests covering all acknowledgment scenarios
- Constraint tests validating poll-level acknowledgment requirements
- Unit tests for container behavior and listener dispatching
- Updated documentation with acknowledgment examples

- Implicit: Records auto-acknowledged as ACCEPT on success, REJECT on error
- Explicit: Application must manually acknowledge each record before next poll

- Explicit mode blocks subsequent polls until all records acknowledged
- Prevents message loss and ensures proper acknowledgment ordering
- Concurrent acknowledgment attempts properly handled with IllegalStateException

- Processing exceptions trigger automatic REJECT acknowledgment
- Acknowledgment failures reset state and throw ShareAcknowledgmentException
- Container continues processing after individual record failures

- ShareAcknowledgment parameter injection follows existing Spring Kafka patterns
- Non-polymorphic listener method names (onShareRecord vs onMessage) prevent potential conflicts with existing listener infrastructure and ensure clear separation between regular and share consumer listener contracts
- Factory and container level configuration options provide flexibility

This implementation provides acknowledgment semantics for Kafka share groups while maintaining backward compatibility with existing implicit acknowledgment behavior.


spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,257 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
-More sophisticated acknowledgment patterns will be added in future versions.
+Share consumers support two acknowledgment modes that control how records are acknowledged after processing.
+
+[[share-acknowledgment-modes]]
+=== Acknowledgment Modes
+
+Share containers support two acknowledgment modes:
+[[share-implicit-acknowledgment]]
+
+==== Implicit Acknowledgment (Default)
```
artembilan: I think the blank line above has to be before the section id.
Kinda such an id has to be stuck to the section title.
Now sure, though, if we need such a fine-grained indexing with ids...

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,257 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
+
+[source,java]
+----
+@Bean
+public ShareConsumerFactory<String, String> explicitShareConsumerFactory() {
+    Map<String, Object> props = new HashMap<>();
+    props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092");
+    props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
+    props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
+    props.put("share.acknowledgement.mode", "explicit");
```
artembilan: This is something new.
I mean an approach to configure such a an option on the target `ContainerProperties`.
We never followed this pattern before and just tried to stay in the `configs` with only Apache Kafka Client properties as much as possible.
So, why now introducing this extra level of complexity and paradox of choice?

Or... Just make it as a setter on this `DefaultShareConsumerFactory` to be type-safe!

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,257 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
+----
+
+[[share-acknowledgment-types]]
+=== Acknowledgment Types
+
+Share consumers support three acknowledgment types:
+
+`ACCEPT`: Record processed successfully, mark as completed
+`RELEASE`: Temporary failure, make record available for redelivery
+`REJECT`: Permanent failure, do not retry
```
artembilan: Don't we need to add dash in the beginning of every entry to make them as a list in the rendered doc?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,257 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
+
+[source,java]
+----
+public interface ShareAcknowledgment {
+    void acknowledge(AcknowledgeType type);
+    void acknowledge();  // Convenience method for ACCEPT
+    void release();      // Convenience method for RELEASE
+    void reject();       // Convenience method for REJECT
+    boolean isAcknowledged();
+    AcknowledgeType getAcknowledgmentType();
```
artembilan: I'm not sure in the getter.
And what is useful for end-user with an `isAcknowledged()`?
If we still need such an accessor, why not make it internal at the place where we implement this `ShareAcknowledgment`
and make end-user API as straightforward as possible?
I also think that API like `acknowledge(AcknowledgeType type)` is not needed since we have all those other choices.
In other words: let's avoid paradox of choice: or one, or another!
But again: this is my preference. If you have some argument to keep this, I'm OK.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,257 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
+public void listen(ConsumerRecord<String, String> record,
+                  ShareAcknowledgment acknowledgment,
+                  ShareConsumer<?, ?> consumer) {
+
+    try {
+        processRecord(record);
+        acknowledgment.acknowledge(); // ACCEPT
+    } catch (RetryableException e) {
+        acknowledgment.release(); // Will be redelivered
+    } catch (Exception e) {
```
artembilan: Can we have a `catch` on a new like even in docs?
Just let's don't encourage for bad habit copy/pasting samples from our docs!

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,257 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
+----
+
+[[share-acknowledgment-constraints]]
+=== Acknowledgment Constraints
+
+In explicit acknowledgment mode, the container enforces important constraints:
+
+Poll Blocking: Subsequent polls are blocked until all records from the previous poll are acknowledged.
+One-time Acknowledgment: Each record can only be acknowledged once.
+Error Handling: If processing throws an exception, the record is automatically acknowledged as `REJECT`.
```
artembilan: Just curious that this is really a proper style these days.
Reading now `The Three-Body Problem`, and indeed they capitalize after colon. 
Let me know, and I'll follow same pattern from now on! 😄

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,257 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
+
+[WARNING]
+In explicit mode, failing to acknowledge records will block further message processing.
+Always ensure records are acknowledged in all code paths.
```
artembilan, line 418: Do we have any trigger to notify target application that ack was missed and we are blocking?
I mean this might be a typical misuse and respective reports like "Spring Kafka does not consume new records any more" 🤷

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,257 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
+        String orderData = record.value();
+        try {
+            if (isValidOrder(orderData)) {
+                if (processOrder(orderData)) {
+                    acknowledgment.acknowledge(); // Success - ACCEPT
+                }
+                else {
+                    acknowledgment.release(); // Temporary failure - retry later
+                }
+                }
```
artembilan: Something off with indents in this code snippet.
Perhaps not everything is indented with 4-spaces instead of tabs.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
```diff
@@ -651,6 +652,10 @@ protected void processListener(MethodKafkaListenerEndpoint<?, ?> endpoint, Kafka
 		KafkaListenerContainerFactory<?> listenerContainerFactory = resolveContainerFactory(kafkaListener,
 				containerFactory, beanName);
 
+		if (listenerContainerFactory instanceof ShareKafkaListenerContainerFactory<?, ?>) {
+			endpoint.setShareConsumer(Boolean.TRUE);
```
artembilan: Why just simple `boolean` is not enough?

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
```diff
@@ -98,6 +98,8 @@ public abstract class AbstractKafkaListenerEndpoint<K, V>
 
 	private @Nullable Boolean batchListener;
 
+	private @Nullable  Boolean shareConsumer;
```
artembilan: I guess if we go with simple `boolean`, then even these two spaces after `@Nullable` would go away 😄 .

Not sure, though, why `batchListener` is not a primitive...

spring-kafka/src/main/java/org/springframework/kafka/config/MethodKafkaListenerEndpoint.java
```diff
@@ -210,7 +211,15 @@ protected MessagingMessageListenerAdapter<K, V> createMessageListenerInstance(
 			@Nullable MessageConverter messageConverter) {
 
 		MessagingMessageListenerAdapter<K, V> listener;
-		if (isBatchListener()) {
+		if (getShareConsumer() != null && getShareConsumer()) {
```
artembilan: So, weird. 
I guess with a primitive this would not be that convoluted.
Either way something like `Boolean.TRUE.equals(getShareConsumer())` looks much better and does exactly the same, but better because of only one method call.

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -125,14 +125,43 @@ private void configureEndpoint(AbstractKafkaListenerEndpoint<K, V> endpoint) {
...
 				.acceptIfNotNull(endpoint.getClientIdPrefix(), properties::setClientId)
-				.acceptIfNotNull(endpoint.getConsumerProperties(), properties::setKafkaConsumerProperties);
+				.acceptIfCondition(explicitAck != null && explicitAck.equals("explicit"),
+						ContainerProperties.ShareAcknowledgmentMode.EXPLICIT,
+						properties::setShareAcknowledgmentMode);
+	}
+
+	private void validateShareConfiguration(KafkaListenerEndpoint endpoint) {
+		// Validate that batch listeners aren't used with share consumers
+		if (endpoint.getBatchListener() != null && endpoint.getBatchListener()) {
```
artembilan: DITTO. 
First of all see my suggestion about `Boolean.TRUE`.
And second: let's double check if we can make it a primitive instead!

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -125,14 +125,43 @@ private void configureEndpoint(AbstractKafkaListenerEndpoint<K, V> endpoint) {
...
+			throw new IllegalArgumentException(
+					"Batch listeners are not supported with share consumers. " +
+							"Share groups operate at the record level.");
+		}
+
+		// Validate acknowledgment mode consistency
+		Object ackMode = this.shareConsumerFactory.getConfigurationProperties()
+				.get("share.acknowledgement.mode");
+		if (ackMode != null && !Arrays.asList("implicit", "explicit").contains(ackMode)) {
```
artembilan: Cannot we rely on that `ContainerProperties.ShareAcknowledgmentMode` enum instead?
Of course if we go such a `share.acknowledgement.mode` property configuration style...

spring-kafka/src/main/java/org/springframework/kafka/listener/AcknowledgingShareConsumerAwareMessageListener.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ShareConsumer;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.kafka.support.ShareAcknowledgment;
+
+/**
+ * A message listener for share consumer containers that supports explicit acknowledgment.
+ * <p>
+ * This interface provides both access to the ShareConsumer and explicit acknowledgment
```
artembilan: Any chances to make Javadocs more reader-friendly?
I mean use of `{@link}`, `{@code}` etc.

spring-kafka/src/main/java/org/springframework/kafka/listener/AcknowledgingShareConsumerAwareMessageListener.java
```diff
@@ -0,0 +1,59 @@
...
+ * capabilities. When used with explicit acknowledgment mode, the acknowledgment parameter
+ * will be non-null and must be used to acknowledge each record. In implicit mode,
+ * the acknowledgment parameter will be null and records are auto-acknowledged.
+ *
+ * @param <K> the key type
+ * @param <V> the value type
+ * @author Soby Chacko
+ * @since 4.0
+ * @see ShareAcknowledgment
```
artembilan, line 48: I told before that my preference is it divide sections in Javadocs, so then it would give more respect to `@author` making it more visible.
But that's my preference: If you are not agreed, just let me know and I won't make such a comment anymore.

spring-kafka/src/main/java/org/springframework/kafka/listener/AcknowledgingShareConsumerAwareMessageListener.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+@FunctionalInterface
+public interface AcknowledgingShareConsumerAwareMessageListener<K, V> extends GenericMessageListener<ConsumerRecord<K, V>> {
+
+	/**
+	 * Invoked with data from kafka, an acknowledgment, and provides access to the consumer.
+	 * When explicit acknowledgment mode is used, the acknowledgment parameter will be non-null
+	 * and must be used to acknowledge the record. When implicit acknowledgment mode is used,
+	 * the acknowledgment parameter will be null.
+	 *
```
artembilan: And here is no: The style in Spring projects is to no blank lines in the method Javadoc.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -160,6 +161,31 @@ public enum EOSMode {
...
+	public enum ShareAcknowledgmentMode {
+		/**
+		 * Records are automatically acknowledged as ACCEPT on next poll, commitSync, or commitAsync.
+		 */
+		IMPLICIT("implicit"),
+
+		/**
+		 * Application must explicitly acknowledge all records before next poll.
+		 */
+		EXPLICIT("explicit");
```
artembilan: What is interesting if you would use lower case for enum names, then you would not need extra string property on it.
On the other hand, if we get rid off of property-based configuration, we may not need even think about lower case, since the type-safe setter would always accepts only this enum.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareConsumerAwareMessageListener.java
```diff
@@ -0,0 +1,51 @@
...
+@FunctionalInterface
+public interface ShareConsumerAwareMessageListener<K, V> extends GenericMessageListener<ConsumerRecord<K, V>> {
+
+	/**
+	 * Invoked with data from kafka and provides access to the share consumer.
+	 *
+	 * @param data the data to be processed.
+	 * @param consumer the share consumer.
+	 */
+	void onShareRecord(ConsumerRecord<K, V> data, ShareConsumer<?, ?> consumer);
```
artembilan: I wonder why do we need this extra abstraction if `AcknowledgingShareConsumerAwareMessageListener` can deal with nullable ack argument?

artembilan (changes requested):
The review so far.
Might be the case that some of it is void according to your new vision, but still might be helpful.

Thanks

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareConsumerAwareMessageListener.java
sobychacko: I tried to follow the same pattern as the regular message listener - `AcknowledgingConsumerAwareMessageListener` and `ConsumerAwareMessageListener`. But I see your concern. We can consider merging these new ones into a single interface.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,321 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
 [[share-record-acknowledgment]]
 == Record Acknowledgment
 
-Currently, share consumers automatically acknowledge records with `AcknowledgeType.ACCEPT` after successful processing.
-More sophisticated acknowledgment patterns will be added in future versions.
+Share consumers support two acknowledgment modes that control how records are acknowledged after processing.
+
+[[share-implicit-acknowledgment]]
+=== Implicit Acknowledgment (Default)
+In implicit mode, records are automatically acknowledged based on processing outcome:
```
artembilan, line 237: I thought I mentioned before that blank line has to be after section title.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,321 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
+public void listen(ConsumerRecord<String, String> record) {
+    System.out.println("Received: " + record.value());
+    // Automatically acknowledged in implicit mode
+}
+----
+
+[[share-acknowledging-listener]]
+==== AcknowledgingShareConsumerAwareMessageListener
+
+This interface provides access to the ShareConsumer instance with optional acknowledgment support.
```
artembilan: Why `ShareConsumer` is not a piece of code?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
artembilan: Any comments, please?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
artembilan: Any comments, please?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,321 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
+    // Set acknowledgment timeout (default is 60 seconds)
+    factory.getContainerProperties().setShareAcknowledgmentTimeout(Duration.ofSeconds(30));
+
+    return factory;
+}
+----
+
+When a record exceeds the timeout, you'll see a warning like:
+----
+WARN: Record not acknowledged within timeout (30 seconds).
```
artembilan, line 443: WARN?
Shouldn't this be ERROR and full listener container stop to avoid blocking?
In my feeling that would be a good trigger to notify that something is broken in the application itself.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,321 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
+            acknowledgment.acknowledge(AcknowledgeType.REJECT);
+            break;
+    }
+}
+----
+
+[[share-acknowledgment-configuration]]
+=== Acknowledgment Mode Configuration
+
+You can configure the acknowledgment mode at both the consumer factory and container levels:
```
artembilan: I think this is a duplication of what was discussed above.

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
```diff
@@ -291,6 +293,19 @@ public void setBatchListener(boolean batchListener) {
 		this.batchListener = batchListener;
 	}
 
+	public void setShareConsumer(Boolean shareConsumer) {
```
artembilan: Why cannot we use a primitive instead?

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -125,14 +126,90 @@ private void configureEndpoint(AbstractKafkaListenerEndpoint<K, V> endpoint) {
...
+	/**
+	 * Determine the acknowledgment mode following Spring Kafka's configuration precedence patterns.
+	 * <p>
+	 * Configuration precedence (highest to lowest):
+	 * <ol>
+	 * <li>Container Properties: {@code containerProperties.getShareAcknowledgmentMode()} (if explicitly set)</li>
+	 * <li>Consumer Config: {@code ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG}</li>
+	 * <li>Default: {@code ShareAcknowledgmentMode.IMPLICIT}</li>
+	 * </ol>
+	 *
```
artembilan: No black lines in method Javadocs

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -125,14 +126,90 @@ private void configureEndpoint(AbstractKafkaListenerEndpoint<K, V> endpoint) {
...
+		if (clientAckMode != null) {
+			String mode = clientAckMode.toString();
+			if ("explicit".equals(mode)) {
+				return ContainerProperties.ShareAcknowledgmentMode.EXPLICIT;
+			}
+			else if ("implicit".equals(mode)) {
+				return ContainerProperties.ShareAcknowledgmentMode.IMPLICIT;
+			}
+			else {
+				throw new IllegalArgumentException(
```
artembilan: Take a look into a `valueOf(mode.toUpperCase())` instead of this manual `if..else`

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -125,14 +126,90 @@ private void configureEndpoint(AbstractKafkaListenerEndpoint<K, V> endpoint) {
...
+
+		// Validate acknowledgment mode consistency using official Kafka client configuration
+		Object ackMode = this.shareConsumerFactory.getConfigurationProperties()
+				.get(ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG);
+		if (ackMode != null) {
+			String ackModeStr = ackMode.toString().toLowerCase();
+			boolean isValid = Arrays.stream(ContainerProperties.ShareAcknowledgmentMode.values())
+					.anyMatch(mode -> mode.name().toLowerCase().equals(ackModeStr));
+			if (!isValid) {
+				throw new IllegalArgumentException(
```
artembilan: Looks like this logic is redundant since the next `determineAcknowledgmentMode()` would do something similar.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -160,6 +161,21 @@ public enum EOSMode {
...
+	public enum ShareAcknowledgmentMode {
+		/**
+		 * Records are automatically acknowledged as ACCEPT on next poll, commitSync, or commitAsync.
+		 */
+		IMPLICIT,
+
+		/**
+		 * Application must explicitly acknowledge all records before next poll.
+		 */
+		EXPLICIT
```
artembilan: Now, seeing this enum, I wonder if we need it at all.
It is just `boolean`: `manualShareAck`, or `autoShareAck`.

I know that you have modeled it after respective Apache Kafka config property, but is there really any plans to extend it to other mode rather than this boolean switch?

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -1115,6 +1135,63 @@ public void setRecordObservationsInBatch(boolean recordObservationsInBatch) {
...
+	 * by regular consumer containers. The acknowledgment mode determines
+	 * how records are acknowledged:
+	 * <ul>
+	 * <li>{@link ShareAcknowledgmentMode#IMPLICIT} - Records are automatically
+	 * acknowledged as ACCEPT when the next poll occurs or when commitSync/commitAsync
+	 * is called</li>
+	 * <li>{@link ShareAcknowledgmentMode#EXPLICIT} - Application must explicitly
+	 * acknowledge each record using the provided {@link ShareAcknowledgment}</li>
+	 * </ul>
+	 *
```
artembilan: No blank lines in Javadocs.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -1115,6 +1135,63 @@ public void setRecordObservationsInBatch(boolean recordObservationsInBatch) {
...
+	 * @param shareAcknowledgmentTimeout the timeout duration
+	 * @since 4.0
+	 */
+	public void setShareAcknowledgmentTimeout(Duration shareAcknowledgmentTimeout) {
+		this.shareAcknowledgmentTimeout = shareAcknowledgmentTimeout;
+	}
+
+	/**
+	 * Get the timeout for share acknowledgments in explicit mode.
+	 *
```
artembilan: This looks like a new update of your IDE: almost every single method Javadoc has a blank line now.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -36,20 +40,40 @@
...
+ * <li><strong>Explicit</strong>: Application must manually acknowledge each record;
+ * subsequent polls are blocked until all records from the previous poll are acknowledged</li>
+ * </ul>
  *
  * @param <K> the key type
  * @param <V> the value type
- *
  * @author Soby Chacko
  * @since 4.0
+ * @see ShareConsumer
```
artembilan, line 78: Same Javadocs sections concern.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -226,6 +310,138 @@ public void run() {
...
+						AcknowledgingShareConsumerAwareMessageListener<K, V> typedAckListener =
+								(AcknowledgingShareConsumerAwareMessageListener<K, V>) ackListener;
+						typedAckListener.onShareRecord(record, acknowledgment, this.consumer);
+					}
+					else {
+						// Basic listener remains the same
+						@SuppressWarnings("unchecked")
+						GenericMessageListener<ConsumerRecord<K, V>> listener =
+								(GenericMessageListener<ConsumerRecord<K, V>>) this.genericListener;
+						listener.onMessage(record);
```
artembilan, line 350: Do we have some protection or way to ack manually in case of just a `GenericMessageListener`?
I see that we provide `acknowledgment` to the `AcknowledgingShareConsumerAwareMessageListener`, but there is no way here.
Shouldn't we fail somewhere during configuration phase?

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
```diff
@@ -209,6 +211,35 @@ public Message<?> toMessage(ConsumerRecord<?, ?> record, @Nullable Acknowledgmen
 		return message;
 	}
 
+	@Override
+	public Message<?> toShareMessage(ConsumerRecord<?, ?> record, @Nullable ShareAcknowledgment acknowledgment, @Nullable ShareConsumer<?, ?> consumer,
```
artembilan: I don't think we need this dedicated API.
How about just make those args of the existing `toMessage` as ` @Nullable Object acknowledgment, @Nullable Object consumer` ?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
```diff
@@ -442,6 +450,33 @@ protected void invoke(Object records, @Nullable Acknowledgment acknowledgment, @
 		}
 	}
 
+	protected void invoke(Object records, @Nullable ShareAcknowledgment acknowledgment, @Nullable ShareConsumer<?, ?> consumer,
```
artembilan: Take a look, please, if existing `invoke()` could be reworked to have `@Nullable Object acknowledgment, @Nullable Object consumer` instead.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ShareRecordMessagingMessageListenerAdapter.java
```diff
@@ -0,0 +1,92 @@
...
+ * <p>The original {@link ConsumerRecord} and
+ * the {@link ShareAcknowledgment} are provided as additional arguments so that these can
+ * be injected as method arguments if necessary.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ * @author Soby Chacko
+ * @since 4.0
+ */
+public class ShareRecordMessagingMessageListenerAdapter<K, V> extends MessagingMessageListenerAdapter<K, V>
```
artembilan, line 49: If we can make ` @Nullable Object acknowledgment, @Nullable Object consumer` in the `MessagingMessageListenerAdapter`, then we might not need this class at all.
And might be not need even `AcknowledgingShareConsumerAwareMessageListener`.
I don't see how is that important to such a contract fine-grained for expected types.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessageConverter.java
```diff
@@ -74,4 +76,22 @@ default void commonHeaders(@Nullable Acknowledgment acknowledgment, @Nullable Co
 			.acceptIfNotNull(KafkaHeaders.CONSUMER, consumer, (key, val) -> rawHeaders.put(key, val));
 	}
 
+	@SuppressWarnings("NullAway") // Dataflow analysis limitation
+	default void commonHeaders(@Nullable ShareAcknowledgment acknowledgment, @Nullable ShareConsumer<?, ?> consumer, Map<String, Object> rawHeaders,
```
artembilan: DITTO: `@Nullable Object acknowledgment, @Nullable Object consumer`.
Or might be the case that we don't need this method at all.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerConstraintTests.java
```diff
@@ -0,0 +1,347 @@
...
+			assertThat(isAcknowledgedInternal(ack)).isTrue();
+
+		}
+		finally {
+			container.stop();
+		}
+	}
+
+	// Utility methods
+	private Map<String, Object> createConsumerProps(String bootstrapServers, String groupId, boolean explicit) {
```
artembilan: `static`?
As well as all those supporting methods below.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -0,0 +1,402 @@
...
+				.willReturn(ConsumerRecords.empty());
+
+		ContainerProperties containerProperties = new ContainerProperties("test-topic");
+		containerProperties.setMessageListener(messageListener);
+
+		ShareKafkaMessageListenerContainer<String, String> container =
+				new ShareKafkaMessageListenerContainer<>(shareConsumerFactory, containerProperties);
+
+		container.start();
+		Thread.sleep(1000);
```
artembilan: I'm not a fun of delays like this in tests.
Isn't there a way to assert on some thread barrier?
More over: do we still need this test class at all if you have all those comprehensive integration tests?
Isn't the functionality verified here covered over there as well?

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
sobychacko: Yep, that sounds like a great idea. Keep in mind though that this is going to be a big breaking change from an API standpoint, no? But then again, we are in 4.0 and probably ok to make those types of changes.

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ShareRecordMessagingMessageListenerAdapter.java
sobychacko: I see your point. However, I think, making this type of change might be too risky from a regression standpoint, as we have to change the API in `GenericMessageListener` to accept `Object` types. Don't you think that we are compromising on the type safety that the frameworks provides today? Plus, `MessagingMessageListenerAdapter` is an abstract class and we need a concrete adapter to deal with share consumers (as we have record and batch adapters). In any case, I think, we may have to look at this particular change after this PR is merged, but let's chat next week.

artembilan: OK. We can live with that, but `MessagingMessageConverter` is not very justified.

sobychacko: Yes, `MessagingMessageConverter` is now fixed.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
artembilan: Any comments on this one, please?

spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
```diff
@@ -291,6 +293,19 @@ public void setBatchListener(boolean batchListener) {
...
+		this.shareConsumer = shareConsumer;
+	}
+
+	/**
+	 * Return true if this endpoint is for a share consumer.
+	 * @return true for a share consumer endpoint.
+	 * @since 4.0
+	 */
+	public boolean isShareConsumer() {
+		return this.shareConsumer != null && this.shareConsumer;
```
artembilan: Still not primitive?
I mean internally.
Why do we need that ` @Nullable Boolean` if simple `boolean` would be enough.
What do I miss, please?

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
artembilan: Any comments here, please?

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -125,14 +126,67 @@ private void configureEndpoint(AbstractKafkaListenerEndpoint<K, V> endpoint) {
...
+						"Invalid " + ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG + ": " + mode +
+						". Must be 'implicit' or 'explicit'");
+			}
+		}
+		// Default to implicit acknowledgment (false)
+		return containerProperties.isExplicitShareAcknowledgment();
+	}
+
+	private void validateShareConfiguration(KafkaListenerEndpoint endpoint) {
```
artembilan: Looks like could be `static`

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerConstraintTests.java
```diff
@@ -0,0 +1,347 @@
...
+		try {
+			// Wait for first batch
+			assertThat(firstBatchLatch.await(15, TimeUnit.SECONDS)).isTrue();
+			assertThat(firstBatchAcks).hasSize(3);
+
+			// Produce second batch while first is pending
+			produceTestRecords(bootstrapServers, topic, 2);
+
+			// Wait and verify second batch is NOT processed yet
+			Thread.sleep(3000);
```
artembilan: I don't think this is reliable way to delay.
Please, consider to use some other async barrier instead.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerConstraintTests.java
```diff
@@ -0,0 +1,347 @@
...
+			// Acknowledge only first 3 records
+			for (int i = 0; i < 3; i++) {
+				batchAcks.get(i).acknowledge();
+			}
+
+			// Produce more records
+			produceTestRecords(bootstrapServers, topic, 1);
+
+			// Should not process new records while one is still pending
+			Thread.sleep(3000);
```
artembilan: DITTO

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
artembilan: Have not been addressed yet.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -173,6 +238,18 @@ private class ShareListenerConsumer implements Runnable {
...
+
+			// Configure acknowledgment mode
+			this.isExplicitMode = containerProperties.isExplicitShareAcknowledgment();
+			this.ackTimeoutMs = containerProperties.getShareAcknowledgmentTimeout().toMillis();
+
+			// Configure consumer properties based on acknowledgment mode
+			if (this.isExplicitMode) {
+				// Apply explicit mode configuration to consumer
+				// Note: This should ideally be done during consumer creation in the factory
+				this.logger.info(() -> "Share consumer configured for explicit acknowledgment mode");
```
artembilan: The `Supplier` is redundant here: you just use a string constant.
Therefore, it is much better to avoid a `Supplier` style.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
artembilan: Any comments here, please?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -226,6 +319,138 @@ public void run() {
...
+					this.consumer.acknowledge(pendingAck.record, pendingAck.type);
+					// Find and notify the acknowledgment object
+					ShareConsumerAcknowledgment ack = this.pendingAcknowledgments.get(pendingAck.record);
+					if (ack != null) {
+						ack.notifyAcknowledged(pendingAck.type);
+						onRecordAcknowledged(pendingAck.record);
+					}
+				}
+				catch (Exception e) {
+					this.logger.error(e, "Failed to process queued acknowledgment for record: " + pendingAck.record);
```
artembilan: Supplier-based, please.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -226,6 +319,138 @@ public void run() {
...
+			if (!this.isExplicitMode || this.acknowledgmentTimestamps.isEmpty()) {
+				return;
+			}
+
+			long currentTime = System.currentTimeMillis();
+			for (Map.Entry<ConsumerRecord<K, V>, Long> entry : this.acknowledgmentTimestamps.entrySet()) {
+				long recordAge = currentTime - entry.getValue();
+				if (recordAge > this.ackTimeoutMs) {
+					ConsumerRecord<K, V> record = entry.getKey();
+					this.logger.warn(() -> String.format(
```
artembilan: Use `LogMessage.format()` instead of these two `Supplier` and `String.format()`

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -96,28 +111,562 @@ void integrationTestShareKafkaMessageListenerContainer(EmbeddedKafkaBroker broke
...
+		container.setBeanName("explicitAckTestContainer");
+		container.start();
+
+		try {
+			assertThat(latch.await(30, TimeUnit.SECONDS)).isTrue();
+			assertThat(received).hasSize(3);
+			assertThat(acknowledgments).hasSize(3);
+			assertThat(acknowledgments).allMatch(Objects::nonNull);
+			assertThat(acknowledgments).allMatch(this::isAcknowledgedInternal);
+			assertThat(acknowledgments).allMatch(ack -> getAcknowledgmentTypeInternal(ack) == AcknowledgeType.ACCEPT);
```
artembilan: You can use chain style instead:
```
		assertThat(acknowledgments).hasSize(3)
                        .allMatch(Objects::nonNull)
                        .allMatch(this::isAcknowledgedInternal)
                        .allMatch(ack -> AcknowledgeType.ACCEPT.equals(getAcknowledgmentTypeInternal(ack)));
```

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -96,28 +111,562 @@ void integrationTestShareKafkaMessageListenerContainer(EmbeddedKafkaBroker broke
...
+		container.setBeanName("constraintTestContainer");
+		container.start();
+
+		try {
+			// Wait for first batch to be processed
+			assertThat(firstBatchLatch.await(15, TimeUnit.SECONDS)).isTrue();
+			assertThat(pendingAcks).hasSize(3);
+
+			// Wait a bit to ensure no more records are processed while acknowledgments are pending
+			Thread.sleep(2000);
```
artembilan: Cannot we ensure ordering instead of vague timeout?

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -96,28 +111,562 @@ void integrationTestShareKafkaMessageListenerContainer(EmbeddedKafkaBroker broke
...
+		// Test initial state
+		assertThat(container.isRunning()).isFalse();
+
+		// Test start
+		container.start();
+		assertThat(container.isRunning()).isTrue();
+
+		// Test processing
+		produceTestRecords(bootstrapServers, topic, 1);
+		Thread.sleep(3000); // Give time for processing
```
artembilan: Cannot we use a `CountDownLatch` instead?
Might be the case that even these 3 seconds are not enough.
Or opposite: the processing is so fast that we block here for nothing.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -96,28 +111,562 @@ void integrationTestShareKafkaMessageListenerContainer(EmbeddedKafkaBroker broke
...
+		container.stop();
+		assertThat(container.isRunning()).isFalse();
+
+		// Test restart
+		listenerCalled.set(false);
+		container.start();
+		assertThat(container.isRunning()).isTrue();
+
+		produceTestRecords(bootstrapServers, topic, 1);
+		Thread.sleep(3000);
```
artembilan: One more blocking.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -96,28 +111,562 @@ void integrationTestShareKafkaMessageListenerContainer(EmbeddedKafkaBroker broke
...
+		assertThat(listenerCalled.get()).isTrue();
+
+		container.stop();
+	}
+
+	private void testBasicMessageListener(DefaultShareConsumerFactory<String, String> factory,
```
artembilan: Looks like `static` methods below starting from this one.

artembilan (changes requested):
Another review round.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
sobychacko: Are you suggesting to exit the app completely at that point after throwing an error? If we log things at ERROR, I think it is probably more appropriate to throw an exception and then completely bail from the application. Is that what you are implying here?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
sobychacko: Do you prefer a blank like after the `@since` line?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
sobychacko: Isn't this validation enough? https://github.com/spring-projects/spring-kafka/pull/4087/files#diff-9adfeabe4d9d037f9e183ae8646ad37c08bfb62b27900b676515056c40a2fbb7R153

This happens even before the container is started.

spring-kafka/src/main/java/org/springframework/kafka/support/converter/MessagingMessageConverter.java
sobychacko: This is resolved, right?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
artembilan: Yes. That's my thought.
I don't know details, but feels like when we are blocked without ack, that might mean that app is broken at this point.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
artembilan: Yes. And after `@author`.
It is more readable and really honor authors of the class.

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
sobychacko: Could you elaborate more? I think the line numbers are updated now.

artembilan: Yeah... I see you have turned that to `boolean` in the props.
Not sure why this comment is not marked as `Outdated`.

Sorry about that.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerConstraintTests.java
```diff
@@ -122,7 +122,8 @@ public void onShareRecord(ConsumerRecord<String, String> record,
 			produceTestRecords(bootstrapServers, topic, 2);
 
 			// Wait and verify second batch is NOT processed yet
-			Thread.sleep(3000);
+			// Using a latch that should NOT count down to verify blocking behavior
+			assertThat(secondBatchLatch.await(3, TimeUnit.SECONDS)).isFalse();
```
artembilan: This does not make difference from the `Thread.sleep`.
Since you expect false here, it indeed is going to block for those 3 seconds waiting for the barrier which is not going to be fulfilled.

If you cannot figure out right now, please, bring `Thread.sleep()` back.
We may look into these tests together later on.
And with `Thread.sleep()` it would be easier to spot problematic tests.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
artembilan: Did you try method chain here?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
sobychacko: I understand your concern, but I don't think stopping the container is the right approach here. Let me explain why.

The timeout warning we have is a Spring Kafka level diagnostic tool to help developers catch bugs in their code (like forgetting to call `ack.acknowledge()`). It's not a Kafka-mandated timeout that requires fatal action.

At the Kafka level (per KIP-932), if a record isn't acknowledged within `share.record.lock.duration.ms`, Kafka automatically releases it back to the pool for redelivery - it doesn't fail or stop. The system is designed to be self-healing through redelivery.

  If we stop the container on this timeout:

  1. Those unacknowledged records would be stuck in "acquired" state until Kafka's timeout releases them
  2. We'd be breaking the cooperative consumption model that KIP-932 provides
  3. We'd be treating a potentially transient issue (slow processing, temporary DB issues) as fatal

  I think `WARN `is appropriate here because:
  - The container is still functional, just blocked waiting for acks (which is correct KIP-932 behavior)
  - The situation might resolve itself (app eventually acknowledges)
  - It alerts developers to potential bugs without crashing the application

Also, I'm planning to add proper retry delivery support in a subsequent PR, which will handle this scenario more gracefully following the KIP-932 delivery tracking model. Does that make sense?

artembilan: Sounds good!
Thank you for explanation!

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -248,8 +229,276 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
+@KafkaListener(topics = "order-processing", containerFactory = "explicitShareKafkaListenerContainerFactory")
+    public void processOrder(ConsumerRecord<String, String> record, ShareAcknowledgment acknowledgment) {
+        String orderId = record.key();
+        String orderData = record.value();
+        try {
+            if (isValidOrder(orderData)) {
+                if (processOrder(orderData)) {
+                    acknowledgment.acknowledge(); // Success - ACCEPT
+                }
+                else {
```
artembilan, line 467: Please, revise this code snippet.
It is not valid at the moment.

spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
artembilan: Please, address.

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -49,6 +50,7 @@
  * @param <V> the value type
  *
  * @author Soby Chacko
+ *
  * @since 4.0
  */
 public class ShareKafkaListenerContainerFactory<K, V>
```
artembilan, line 56: The `implements` line is too long.
Please, ensure not more than `120` symbols per line.

I don't see a reason in `autoStartup` and `phase` being wrappers.
The `applicationEventPublisher` and `applicationContext` must be marked with `@SuppressWarnings("NullAway.Init")`.
Might simplify the logic in the class letting to avoid null checks.

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -125,14 +127,67 @@ private void configureEndpoint(AbstractKafkaListenerEndpoint<K, V> endpoint) {
...
+	 * @throws IllegalArgumentException if an invalid acknowledgment mode is configured
+	 */
+	private boolean determineExplicitAcknowledgment(ContainerProperties containerProperties) {
+		// Check Kafka client configuration
+		Object clientAckMode = this.shareConsumerFactory.getConfigurationProperties()
+				.get(ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG);
+
+		if (clientAckMode != null) {
+			String mode = clientAckMode.toString().toLowerCase();
+			if ("explicit".equals(mode)) {
```
artembilan: Can we rely on the `ShareAcknowledgementMode.fromString()` API instead of the manual logic in this method?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ShareRecordMessagingMessageListenerAdapter.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ShareConsumer;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.kafka.listener.AcknowledgingConsumerAwareMessageListener;
+import org.springframework.kafka.listener.AcknowledgingShareConsumerAwareMessageListener;
+import org.springframework.kafka.listener.KafkaListenerErrorHandler;
+import org.springframework.kafka.support.ShareAcknowledgment;
+import org.springframework.kafka.support.converter.JacksonProjectingMessageConverter;
+import org.springframework.kafka.support.converter.ProjectingMessageConverter;
```
artembilan: I prefer to avoid deprecated imports and use FQCN instead of the place of use where you already have that `@SuppressWarnings("removal")`

spring-kafka/src/main/java/org/springframework/kafka/support/ShareAcknowledgmentException.java
```diff
@@ -0,0 +1,57 @@
...
+ * <li>Network issues communicating with the broker</li>
+ * <li>Record state conflicts (e.g., record already acknowledged by another consumer)</li>
+ * <li>Share session errors</li>
+ * <li>Broker-side errors processing the acknowledgment</li>
+ * </ul>
+ *
+ * @author Soby Chacko
+ * @since 4.0
+ */
+public class ShareAcknowledgmentException extends KafkaException {
```
artembilan: Looks like this class is out of use.
Perhaps just remove?

spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/ShareRecordMessagingMessageListenerAdapter.java
```diff
@@ -81,8 +75,8 @@ public void onShareRecord(ConsumerRecord<K, V> record, @Nullable ShareAcknowledg
 		}
 		if (logger.isDebugEnabled()) {
 			RecordMessageConverter messageConverter = getMessageConverter();
-			if (!(messageConverter instanceof JacksonProjectingMessageConverter
-					|| messageConverter instanceof ProjectingMessageConverter)) {
+			if (!(messageConverter instanceof org.springframework.kafka.support.converter.JacksonProjectingMessageConverter
```
artembilan: This is an overreacting.
I have talked only about a deprecated `ProjectingMessageConverter` class.

PR #4094, opened 2025-09-29
Update listener observation operation tag

cfredri4:
This updates the listener observation operation tag according to latest standard.

E.g. [Open Telemetry states](https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/messaging/messaging-spans.md#consumer):
> The consumer receives, processes, and settles a message. "Receiving" is the process of obtaining a message from the intermediary, **"processing" is the process of acting on the information a message contains**, "settling" is the process of notifying an intermediary that a message was processed successfully.

E.g. [Micrometer JMS follows this](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-jakarta9/src/main/java/io/micrometer/jakarta9/instrument/jms/DefaultJmsProcessObservationConvention.java#L34).
> `private static final KeyValue OPERATION_PROCESS = KeyValue.of(LowCardinalityKeyNames.OPERATION, "process");`

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
@@ -231,7 +231,7 @@ public enum KafkaListenerObservation implements ObservationDocumentation {
 			KeyValues keyValues = KeyValues.of(
 					ListenerLowCardinalityTags.LISTENER_ID.withValue(context.getListenerId()),
 					ListenerLowCardinalityTags.MESSAGING_SYSTEM.withValue("kafka"),
-					ListenerLowCardinalityTags.MESSAGING_OPERATION.withValue("receive"),
+					ListenerLowCardinalityTags.MESSAGING_OPERATION.withValue("process"),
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
I would like to call out a few things:
- The OTel doc you quoted from says this: "Status: Experimental". It's unstable, it's not really a standard that you want to use in prod since it's likely it will have breaking changes in the future.
- Doing this is a breaking change, it can break dashboards/alerts/etc.

I think it is fine to change this (in the next major version) if we think it's worth to introduce this (breaking) change to make it more similar to the instrumentation of JMS and/or have a better name (if we think it's better).

artembilan:
Thanks, @jonatan-ivanov , for confirmation.

I think we can bite a bullet and updated the Kafka listener observation into a `process` in our current `4.0.0`.
We may introduce a `receive` operation for the `KafkaConsumer.poll()` in the future.

artembilan (approved):
LGTM.

@sobychacko ,

let me know if you'd like to take a look before I merge this.

Thanks

sobychacko:
Go ahead @artembilan with the merge.

PR #4096, opened 2025-10-03
GH-4088 : Bug Fix (send offset to tx, when no active tx)

### LINKED ISSUE #4088: KafkaMessageListenerContainer tries to send offsets to tx when record is filtered out by early record interceptor

3.3.7

**Describe the bug**

KafkaMessageListenerContainer tries to send offsets to a transaction when a record is filtered out by an early record interceptor and fails because there is no tx in progress (the interceptor is called before tx).
```
o.s.k.l.KafkaMessageListenerContainer 149 - Consumer exception
java.lang.IllegalStateException: Cannot send offsets if a transaction is not in progress (currentState= READY)
	at org.apache.kafka.clients.producer.internals.TransactionManager.sendOffsetsToTransaction(TransactionManager.java:367)
	at org.apache.kafka.clients.producer.KafkaProducer.sendOffsetsToTransaction(KafkaProducer.java:785)
	at org.springframework.kafka.core.DefaultKafkaProducerFactory$CloseSafeProducer.sendOffsetsToTransaction(DefaultKafkaProducerFactory.java:1169)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.doSendOffsets(KafkaMessageListenerContainer.java:3080)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.sendOffsetsToTransaction(KafkaMessageListenerContainer.java:3073)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.ackCurrent(KafkaMessageListenerContainer.java:3062)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.ackCurrent(KafkaMessageListenerContainer.java:3051)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.checkEarlyIntercept(KafkaMessageListenerContainer.java:2673)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.invokeRecordListenerInTx(KafkaMessageListenerContainer.java:2517)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.invokeRecordListener(KafkaMessageListenerContainer.java:2500)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.invokeListener(KafkaMessageListenerContainer.java:2152)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.invokeIfHaveRecords(KafkaMessageListenerContainer.java:1528)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.pollAndInvoke(KafkaMessageListenerContainer.java:1466)
	at org.springframework.kafka.listener.KafkaMessageListenerContainer$ListenerConsumer.run(KafkaMessageListenerContainer.java:1335)
```

**To Reproduce**

1. Set up a transactional kafka listener
2. Add an early (interceptBeforeTx is `true`) filtering RecordInterceptor
3. Send a record that is _not_ filtered out by the interceptor (this step sets KafkaMessageListenerContainer.ListenerConsumer#producer)
4. Send a record that is filtered out by the interceptor

**Expected behavior**

No offsets sending to a transaction.

moonyoungCHAE:
Fixes #4088

# Description
When a message is filtered out by the Kafka interceptor while the previous message passes through the interceptor and the producer is created, a commit occurs while the Kafka transaction is not active.

structure: Interceptor → Listener (a transactional listener that consumes a message and produces another message)

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/TransactionalContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/TransactionalContainerTests.java
@@ -76,6 +76,8 @@ import org.springframework.kafka.test.utils.KafkaTestUtils;
 import org.springframework.kafka.transaction.KafkaTransactionManager;
 import org.springframework.messaging.MessageHeaders;
 import org.springframework.transaction.TransactionDefinition;
+import org.springframework.transaction.TransactionExecution;
+import org.springframework.transaction.TransactionExecutionListener;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.transaction.support.DefaultTransactionDefinition;
 import org.springframework.util.backoff.FixedBackOff;
@@ -1167,7 +1169,6 @@ public class TransactionalContainerTests {
 		DefaultKafkaConsumerFactory<Integer, String> cf = new DefaultKafkaConsumerFactory<>(consumerProperties);
 		ContainerProperties containerProps = new ContainerProperties(topic11);
 		containerProps.setPollTimeout(10_000);
-		final var successLatch = new AtomicReference<>(new CountDownLatch(2));
 		containerProps.setMessageListener(new MessageListener<Integer, String>() {
 			@Transactional("testSendOffsetOnlyOnActiveTransaction")
 			@Override
@@ -1177,9 +1178,19 @@ public class TransactionalContainerTests {
 
 		// init container
 		KafkaTransactionManager<Object, Object> tm = new KafkaTransactionManager<>(pf);
+		AtomicInteger txCount = new AtomicInteger(0);
+		tm.addListener(new TransactionExecutionListener() {
+			@Override
+			public void afterCommit(TransactionExecution transaction, @Nullable Throwable commitFailure) {
+				txCount.incrementAndGet();
+				TransactionExecutionListener.super.afterCommit(transaction, commitFailure);
+			}
+		});
 		containerProps.setKafkaAwareTransactionManager(tm);
+
 		KafkaMessageListenerContainer<Integer, String> container = new KafkaMessageListenerContainer<>(cf, containerProps);
 		container.setBeanName("testSendOffsetOnlyOnActiveTransaction");
+		final var interceptorLatch = new AtomicReference<>(new CountDownLatch(1));
 		container.setRecordInterceptor(new RecordInterceptor<Integer, String>() {
 			boolean isFirst = true;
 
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3080,6 +3080,9 @@ else if (!this.autoCommit && (!this.isAnyManualAck || commitRecovered)) {
 		}
 
 		private void sendOffsetsToTransaction() {
+			if (this.kafkaTxManager != null && TransactionSynchronizationManager.getResource(this.kafkaTxManager.getProducerFactory()) == null) {
```
sobychacko: What if the `kafkaTxManager` is null? Don't we need to address that also? If it is null, we can return as there is no point in sending the offsets to transaction. Do I miss anything on that?

moonyoungCHAE: @sobychacko You're right. I fixed it. Thanks for the feedback.

spring-kafka/src/test/java/org/springframework/kafka/listener/TransactionalContainerTests.java
```diff
@@ -1148,4 +1152,65 @@ public void onMessage(List<ConsumerRecord<Integer, String>> data) {
...
+			}
+		});
+		container.start();
+
+		template.executeInTransaction(t -> {
+			template.send(new ProducerRecord<>(topic11, 0, 0, "bar1"));
+			template.send(new ProducerRecord<>(topic11, 0, 0, "bar2"));
+			return null;
+		});
+		assertThat(successLatch.get().await(30, TimeUnit.SECONDS)).isTrue();
```
sobychacko: How is this test ensuring that offsets are sent to only active transactions? I mean, i see that you are asserting that two records are consumed transactionally, but how do we ensure the case where there is no active txn? Thanks!

moonyoungCHAE: @sobychacko  Thanks for the feedback. I modified the test to verify whether the commit occurred within a transaction by using `TransactionExecutionListener`.

PR #4102, opened 2025-10-07
Add concurrency support to `ShareKafkaMessageListenerContainer`

sobychacko:
Share consumers (KIP-932) enable record-level load balancing where
  multiple consumers can cooperatively process from the same partitions.
  Unlike traditional consumer groups with exclusive partition ownership,
  share groups distribute work at the broker level via the share group
  coordinator.

  This commit adds native concurrency support to the existing
  `ShareKafkaMessageListenerContainer` rather than creating a separate
  `ConcurrentShareKafkaMessageListenerContainer`. This design choice
  avoids the parent/child container complexity that exists in the regular
  consumer model, since share consumers fundamentally operate differently:

  - Work distribution happens at the broker level, not at the Spring layer
  - Multiple threads simply provide more capacity for the broker to distribute records across
  - No partition ownership model to coordinate between child containers

  This approach provides:

  - Simpler architecture with a single container managing multiple threads
  - No parent/child context propagation concerns
  - Better alignment with share consumer semantics (record-level vs partition-level distribution)
  - Increased throughput for high-volume workloads
  - Better resource utilization across consumer threads

  Users can configure concurrency at three levels:
  1. Per-listener via `@KafkaListener(concurrency = N)`
  2. Factory-level default via `factory.setConcurrency(N)`
  3. Programmatically via `container.setConcurrency(N)`

  The feature works seamlessly with both implicit (auto-acknowledge) and
  explicit (manual acknowledge/release/reject) acknowledgment modes, with
  each consumer thread independently managing its own acknowledgments.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
@@ -107,7 +107,7 @@ factory.addListener(new ShareConsumerFactory.Listener<String, String>() {
 [[share-kafka-message-listener-container]]
 === ShareKafkaMessageListenerContainer
 
-The `ShareKafkaMessageListenerContainer` provides a simple, single-threaded container for share consumers:
+The `ShareKafkaMessageListenerContainer` provides a container for share consumers with support for concurrent processing:
 
 [source,java]
 ----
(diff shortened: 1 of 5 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -151,6 +151,145 @@ Share consumers do not support:
...
 ====
 
+[[share-container-concurrency]]
+=== Concurrency
+
+The `ShareKafkaMessageListenerContainer` supports concurrent processing by creating multiple consumer threads within a single container.
+Each thread runs its own `ShareConsumer` instance that participates in the same share group.
+
+Unlike traditional consumer groups where concurrency involves partition distribution, share consumers leverage Kafka's record-level distribution at the broker.
+This means multiple consumer threads in the same container work together as part of the share group, with the Kafka broker distributing records across all consumer instances.
```
artembilan, line 161: Can we mention somehow that concurrency on a single container is exactly the same as several instance of the application. In other words, if several application instances declare concurrency on their container, all the concurrency is additive.

Did I understand this right?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -151,6 +151,145 @@ Share consumers do not support:
...
+**Work Distribution Behavior:**
+
+With share consumers, record distribution is controlled by Kafka's share group coordinator at the broker level, not by Spring for Apache Kafka.
+The broker may assign all records to a single consumer thread at any given time, especially when:
+
+* The topic has a single partition
+* There's low message volume
+* The broker's distribution algorithm favors certain consumers
+
+This is normal behavior. The key benefit of concurrency is having multiple consumer threads *available* to the share group coordinator for distribution.
```
artembilan: One sentence per line.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -151,6 +151,145 @@ Share consumers do not support:
...
+        acknowledgment.reject(); // Permanent failure
+    }
+}
+----
+
+[NOTE]
+====
+**Work Distribution Behavior:**
+
+With share consumers, record distribution is controlled by Kafka's share group coordinator at the broker level, not by Spring for Apache Kafka.
```
artembilan: Does this mean that it works same way as in many other message broker implementations for queues: the next record is pushed down to the next available consumer?
Or is that done in batches?
Where do we control that batch size then? `min.bytes`? `max.record`, `max.wait`?

Thanks

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -98,6 +101,22 @@ public void setPhase(int phase) {
...
+	 * This specifies the number of consumer threads to create within each container.
+	 * Each thread creates its own {@link org.apache.kafka.clients.consumer.ShareConsumer}
+	 * instance and participates in the same share group. The Kafka broker distributes
+	 * records across all consumer instances, providing record-level load balancing.
+	 * <p>
+	 * This can be overridden per listener endpoint using the {@code concurrency}
+	 * attribute on {@code @KafkaListener}.
+	 * @param concurrency the number of consumer threads (must be greater than 0)
+	 */
+	public void setConcurrency(Integer concurrency) {
```
artembilan: I think this could be a primitive.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -122,19 +134,54 @@ public void setClientId(String clientId) {
 		this.clientId = clientId;
 	}
 
+	/**
+	 * Get the concurrency level (number of consumer threads).
+	 * @return the concurrency level
+	 */
+	public int getConcurrency() {
```
artembilan: Why do we need getter?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -122,19 +134,54 @@ public void setClientId(String clientId) {
...
-		ShareListenerConsumer listenerConsumerForMetrics = this.listenerConsumer;
-		if (listenerConsumerForMetrics != null) {
-			Map<MetricName, ? extends Metric> metrics = listenerConsumerForMetrics.consumer.metrics();
-			return Collections.singletonMap(listenerConsumerForMetrics.getClientId(), metrics);
+		this.lifecycleLock.lock();
+		try {
+			if (this.consumers.isEmpty()) {
+				return Collections.emptyMap();
+			}
+			Map<String, Map<MetricName, ? extends Metric>> allMetrics = new ConcurrentHashMap<>();
```
artembilan: Why concurrent?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -161,15 +208,51 @@ protected void doStart() {
...
 	}
 
 	@Override
 	protected void doStop() {
 		setRunning(false);
-		// The consumer will exit its loop naturally when running becomes false.
+		// Wait for all consumer threads to complete
+		this.lifecycleLock.lock();
+		try {
+			for (CompletableFuture<Void> future : this.consumerFutures) {
```
artembilan: See `CompletableFuture.allOf()` instead.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -83,6 +84,8 @@
 )
 class ShareKafkaMessageListenerContainerIntegrationTests {
 
+	private static final LogAccessor logger = new LogAccessor(LogFactory.getLog(ShareKafkaMessageListenerContainerIntegrationTests.class));
```
artembilan: The `LogAccessor` has a ctor based on the class:
```
	/**
	 * Create a new accessor for the specified Commons Log category.
	 * @see LogFactory#getLog(Class)
	 */
	public LogAccessor(Class<?> logCategory) {
```

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -828,4 +833,290 @@ private static AcknowledgeType getAcknowledgmentTypeInternal(ShareAcknowledgment
...
+			receivedValues.add(record.value());
+			latch.countDown();
+		});
+
+		ShareKafkaMessageListenerContainer<String, String> container =
+				new ShareKafkaMessageListenerContainer<>(factory, containerProps);
+		container.setBeanName("concurrencyBasicTest");
+		container.setConcurrency(concurrency);
+
+		assertThat(container.getConcurrency()).isEqualTo(concurrency);
```
artembilan: This is something very obvious and totally redundant.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -828,4 +833,290 @@ private static AcknowledgeType getAcknowledgmentTypeInternal(ShareAcknowledgment
...
+		container.start();
+
+		try {
+			// Verify all records are processed
+			assertThat(latch.await(30, TimeUnit.SECONDS))
+					.as("All %d records should be processed", numRecords)
+					.isTrue();
+
+			// Log thread distribution for debugging
+			logger.info(() -> "Thread distribution: " + threadCounts);
```
artembilan: Why do we need logger in this test class just for these infos?
How is this useful for the whole test suite, esspecialy running on CI/CD?

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -828,4 +833,290 @@ private static AcknowledgeType getAcknowledgmentTypeInternal(ShareAcknowledgment
...
+			assertThat(threadCounts.size())
+					.as("At least one consumer thread should process records")
+					.isGreaterThanOrEqualTo(1)
+					.isLessThanOrEqualTo(concurrency);
+
+			// Verify work was done
+			int totalProcessed = threadCounts.values().stream()
+					.mapToInt(AtomicInteger::get)
+					.sum();
+			assertThat(totalProcessed).isEqualTo(numRecords);
```
artembilan: This looks totally equal to the `assertThat(receivedValues).hasSize(numRecords);`.
Since we really don't care what thread has processed those records, there is just no need to track them.
Why over-complicate tests?

PR #4103, opened 2025-10-08
Document share consumer poison message protection and delivery count

sobychacko:
Add comprehensive documentation on KIP-932's broker-side poison message
  protection mechanism. Clarifies that delivery count is broker-internal
  and not exposed to applications.

  Includes:
  - How delivery count works and configuration via group.share.delivery.attempt.limit
  - Retry strategy recommendations using RELEASE/REJECT/ACCEPT acknowledgment types
  - Example showing proper error handling patterns for transient vs permanent failures

  Addresses common questions about redelivery semantics in share consumers.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
@@ -667,6 +667,96 @@ public void validateData(ConsumerRecord<String, String> record, ShareAcknowledgm
 }
 ----
 
+[[share-poison-message-protection]]
+== Poison Message Protection and Delivery Count
+
+KIP-932 includes broker-side poison message protection to prevent unprocessable records from being endlessly redelivered.
+
+=== How It Works
+
+Every time a record is acquired by a consumer in a share group, the broker increments an internal delivery count.
+The first acquisition sets the delivery count to 1, and each subsequent acquisition increments it.
+When the delivery count reaches the configured limit (default: 5), the record moves to **Archived** state and is not eligible for additional delivery attempts.
+
+=== Configuration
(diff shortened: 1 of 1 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -667,6 +667,96 @@ public void validateData(ConsumerRecord<String, String> record, ShareAcknowledgm
...
+        // Attempt to process the order
+        orderService.process(record.value());
+        ack.acknowledge(); // ACCEPT - successfully processed
+    }
+    catch (TransientException e) {
+        // Temporary failure (network issue, service unavailable, etc.)
+        // Release the record for redelivery
+        // Broker will retry up to group.share.delivery.attempt.limit times
+        logger.warn("Transient error processing order, will retry: {}", e.getMessage());
+        ack.release(); // RELEASE - make available for retry
```
artembilan, line 740: Just a question.
When we do `release` for the intermittent errors, does broker ensure an order of records?
Or when we use share group consumer the order of records is of the scope and we should treat every record as independent logical entity which does not effect any others?
Thanks

sobychacko: When using Share Groups, there's no ordering guarantee - records are distributed at the record level across any available consumer, so they can be processed out of order. When `release()` is called, that record goes back into the pool and could be picked up by any consumer at any time, with no guarantee about when it gets redelivered relative to other records. Each record should be treated as an independent entity that doesn't depend on the order of others. If ordering matters for a use case, traditional consumer groups are the right choice since they provide ordering guarantees within each partition.

Issue #3562, opened 2024-10-18
Add an option to only commit consumer offsets for records that are not filtered out using RecordFilterStrategy


geertvb:
**Expected Behavior**

When using a RecordFilterStrategy in combination with an AckMode like RECORD, only the consumer offsets of the records that have been handled, should be committed.

Suggestion: Add a new AckMode = RECORD_FILTERED that commits the offset only when a record is not filtered out.

**Current Behavior**

When using `ack-mode: RECORD`, the offset of every record is committed to Kafka even if it was filtered out.

**Context**

The synchronous commit of consumer record offset is an expensive operation.
It slows down record consumption unnecessarily in case the consumer is only interested in a small percentage of the records and still wants every record that is processed to be committed.

geertvb:
Hi @sobychacko, thanks for the quick response and considering it as a feature for one of the next releases.
For the time being we work around the issue by using a pure kafka-client ConsumerInterceptor.

Sauhardstark:
Hi @sobychacko , just wanted to check if this is ready for pickup now or if its already been done ?

sobychacko:
@Sauhardstark We haven't gotten to this yet, but if you want to contribute a PR, that would be appreciated. Thanks!

Chaedie:
Hi @sobychacko, I'd like to work on this issue if it's available.
Could you please assign this issue to me? Thanks!

(I have already looked into the source code)

sobychacko:
Hi @Chaedie, please go ahead and submit a PR.

sobychacko:
@Chaedie Any updates on this issue? No problem if you are not ready to send a PR, but we have the RC release happening next week, so if you are in the middle of this PR, this might be the time to submit a PR.

Chaedie:
@sobychacko Thanks for the reminder. I already implemented the code. 
But I've been too shy to open the PR... 😅
I'll open a PR within the next 24 hours.

PR #4108, opened 2025-10-16
GH-3562: Add AckMode.RECORD_FILTERED

Chaedie:
**Issue:** #3562

### Summary
Add a new AckMode `AckMode.RECORD_FILTERED`.
It will commits offsets only for records that are not filtered by `RecordFilterStrategy`.

### Problem

When RecordFilterStrategy is used with `AckMode.RECORD`, the container commits the offset for every record, including those filtered out (never delivered to the listener).
This causes unnecessary synchronous commits and can advance offsets for records that were effectively skipped.

### Implementation
- `ContainerProperties.AckMode.RECORD_FILTERED`
- `KafkaMessageListenerContainer`: per-record commit path skips filtered records
- `FilteringAware` + `FilteringMessageListenerAdapter` to expose filter results

### Tests
- RECORD vs RECORD_FILTERED behavior with/without filtering
- Mixed partitions and all-filtered cases
- Thread isolation test (recordFilteredModeShouldBeThreadIsolated) verifying independent adapter state per thread

---

I wasn’t fully confident about my code, so I hesitated to open this PR for a few weeks. 😅
Please feel free to review it — thank you!

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/AckModeRecordFilteredTest.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/AckModeRecordFilteredTest.java
@@ -345,27 +345,25 @@ class AckModeRecordFilteredTest {
 		RecordFilterStrategy<String, String> filter = rec -> rec.offset() == 0;
 
 		FilteringMessageListenerAdapter<String, String> adapter =
-				new FilteringMessageListenerAdapter<>(
-						(MessageListener<String, String>) r -> {
-						},
-						filter
-				) {
+				new FilteringMessageListenerAdapter<>((MessageListener<String, String>) r -> {
+				}, filter) {
 					@Override
-					public void onMessage(ConsumerRecord<String, String> rec,
-					                      Acknowledgment ack,
-					                      Consumer<?, ?> consumer) {
+					public void onMessage(ConsumerRecord<String, String> rec, Acknowledgment ack, Consumer<?, ?> consumer) {
 						super.onMessage(rec, ack, consumer);
 						if (rec.offset() == 0) {
 							aHasSetState.countDown();
 							try {
 								bHasProcessed.await(500, TimeUnit.MILLISECONDS);
-							} catch (InterruptedException e) {
+							}
+							catch (InterruptedException e) {
 								Thread.currentThread().interrupt();
 							}
-						} else if (rec.offset() == 1) {
+						}
+						else if (rec.offset() == 1) {
 							try {
 								aHasSetState.await(200, TimeUnit.MILLISECONDS);
-							} catch (InterruptedException e) {
+							}
+							catch (InterruptedException e) {
 								Thread.currentThread().interrupt();
 							}
 							bHasProcessed.countDown();


spring-kafka/src/test/java/org/springframework/kafka/listener/AckModeRecordFilteredTest.java
```diff
@@ -0,0 +1,329 @@
...
+
+/**
+ * Tests for the RECORD_FILTERED acknowledge mode.
+ *
+ * Related to GitHub issue #3562
+ *
+ * @author Chaedong Im
+ * @see AckModeRecordWithFilteringTest
+ */
+public class AckModeRecordFilteredTest {
```
sobychacko: You can drop public modifier from tests.

spring-kafka/src/test/java/org/springframework/kafka/listener/AckModeRecordFilteredTest.java
```diff
@@ -0,0 +1,329 @@
...
+
+	@SuppressWarnings({"unchecked", "deprecation"})
+	@Test
+	public void testRecordFilteredModeOnlyCommitsProcessedRecords() throws InterruptedException {
```
sobychacko: No public modifier needed.

spring-kafka/src/test/java/org/springframework/kafka/listener/AckModeRecordFilteredTest.java
```diff
@@ -0,0 +1,329 @@
...
+
+		// Then: Verify that only odd offset records were processed
+		assertThat(processedValues).containsExactly("value1", "value3");
+
+		verify(consumer, times(2)).commitSync(any(), any(Duration.class));
+	}
+
+	@SuppressWarnings({"unchecked", "deprecation"})
+	@Test
+	public void testRecordFilteredModeWithAllRecordsFiltered() throws InterruptedException {
```
sobychacko: no public modifier needed.

spring-kafka/src/test/java/org/springframework/kafka/listener/AckModeRecordWithFilteringTest.java
```diff
@@ -0,0 +1,321 @@
...
+
+/**
+ * Tests to verify the behavior of RECORD acknowledge mode when used with filtering strategies.
+ *
+ * Related to GitHub issue #3562
+ *
+ * @author Chaedong Im
+ * @see AckModeRecordFilteredTest
+ */
+public class AckModeRecordWithFilteringTest {
```
sobychacko: No public modifier needed.

sobychacko (changes requested):
You don't need to use `public` modifier on test classes or methods any longer.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -69,6 +69,16 @@ public enum AckMode {
 		 */
 		RECORD,
 
+		/**
```
sobychacko, line 73: need to add author tag to the class.

spring-kafka/src/test/java/org/springframework/kafka/listener/AckModeRecordFilteredTest.java
Chaedie: I removed it. Thanks 😊

spring-kafka/src/test/java/org/springframework/kafka/listener/AckModeRecordFilteredTest.java
Chaedie: I removed it. Thanks 😊

spring-kafka/src/test/java/org/springframework/kafka/listener/AckModeRecordFilteredTest.java
Chaedie: I removed it. Thanks 😊

spring-kafka/src/test/java/org/springframework/kafka/listener/AckModeRecordWithFilteringTest.java
Chaedie: I removed it. Thanks 😊

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
Chaedie: Oh, I missed this class. I added author tag. Thanks 😊

PR #4126, opened 2025-10-29
Fixes classloading for sasl callback handler, when the handler needs to be loaded via the spring-boot fat-jar classloader

alpapad:


sobychacko:
Merged via https://github.com/spring-projects/spring-kafka/commit/15b3032ec694d9b0d654f22832f2a89386556fdb. Polished the code a little bit and converted the integration tests into few unit tests. Your tests were good, but it was a bit complex due to the runtime compilation, jars required etc.

PR #4149, opened 2025-11-08
GH-1587: Refactor fixTxOffsets to use ConsumerRecords#nextOffsets

bandalgomsu:
Refactor fixTxOffsets to use nextOffsets API to correctly skip transactional commit markers
issue : #1587 

thanks 🙂

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -173,6 +173,7 @@ import org.springframework.util.StringUtils;
  * @author Christian Fredriksson
  * @author Timofey Barabanov
  * @author Janek Lasocki-Biczysko
+ * @author Su Ko
  */
 public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 		extends AbstractMessageListenerContainer<K, V> implements ConsumerPauseResumeEventPublisher {
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/TransactionalContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/TransactionalContainerTests.java
@@ -108,6 +108,7 @@ import static org.mockito.Mockito.verifyNoMoreInteractions;
  * @author Wang Zhiyang
  * @author Soby Chacko
  * @author Raphael Rösch
+ * @author Su Ko
  *
  * @since 1.3
  *


sobychacko:
@bandalgomsu Thanks for the PR! Given our proximity to GA release, we'd prefer to defer this merge to avoid last-minute regressions. We'll revisit after next week's release.

Meanwhile, could you add either a detailed commit message explaining the motivation or inline code comments? We'll also need tests to validate these changes.

We need to make sure that the feedback that was given here is addressed adequately: https://github.com/spring-projects/spring-kafka/issues/1587#issuecomment-3433517159

bandalgomsu:
Thanks for the comments !! 
i added it 🙂

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1604,50 +1609,69 @@ private void notIdle() {
 			}
 		}
 
-		private void savePositionsIfNeeded(ConsumerRecords<K, V> records) {
-			if (this.fixTxOffsets) {
-				this.savedPositions.clear();
-				records.partitions().forEach(tp -> this.savedPositions.put(tp, this.consumer.position(tp)));
```
mjsax, line 1613: I am not familiar with the code base, but is `savedPositions` still used? Or could it be removed entirely? It seems to get replaced by newly added `lastRecords`?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1604,50 +1609,69 @@ private void notIdle() {
...
+		private void saveLastRecordsIfNeeded(@Nullable ConsumerRecords<K, V> records) {
+			if (this.fixTxOffsets && records != null && !records.nextOffsets().isEmpty()) {
+				this.lastRecords = records;
```
mjsax: Not sure if this logic is correct? I believe it would be required to store `nextOffsets` per partition -- assume we get an result from `poll()` which give us `nextOffsets` for partition A and B (and no data for A and B). The next call to `poll` might be empty for partitions C and D, providing `nextOffsets` for partition C and D though. The current code would lose the information for A and B, replacing it with the information from C and D.

If we use a `Map<TopicPartition, OffsetAndMetadata>` however, we can accumulate the information correctly across multiple `poll()` calls. (This is also how we do it inside Kafka Streams.)

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1604,50 +1609,69 @@ private void notIdle() {
...
-				finally {
+
+				Map<TopicPartition, OffsetAndMetadata> nextOffsets = this.lastRecords.nextOffsets();
+				Map<TopicPartition, OffsetAndMetadata> toFix = new HashMap<>();
+
+				// Fix offset only if records were actually processed
+				nextOffsets.forEach((tp, nextOffset) -> {
+					OffsetAndMetadata committed = this.lastCommits.get(tp);
+
+					if (committed == null) {
```
mjsax: Not sure if this is correct? Assume the application commits offset X, which is "endOffset - 1", and "endOffset - 1" is a TX marker. Next a rebalance happens and the partition is re-assigned to a different consumer in the group. The new owner would never process any data, and not commit anything on the regular commit path, however, it is actually supposed to step over the TX marker and to advance the committed offset to "endOffset".

I think we should only skip a commit, if the last committed offset is the same as "nextOffsets" for a specific partition?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1604,50 +1609,69 @@ private void notIdle() {
...
+		/**
+		 * Fix transactional offsets using ConsumerRecords#nextOffsets() API.
+		 * This method addresses the issue where Kafka transaction markers can cause
+		 * incorrect offset tracking. By using nextOffsets() instead of position(),
+		 * we ensure:
+		 * - Transaction markers are automatically filtered out
+		 * - Partition-leader epoch information is correctly captured
+		 * - Empty poll() cases are properly handled
+		 */
 		private void fixTxOffsetsIfNeeded() {
```
mjsax, line 1624: For my own education -- when is this method called exactly?

My impression is (I could be wrong of course, as I do not know this code base), that this method is called to do some additional(?) offset commits, outside of the regular commit code path? For this case, I am wondering if it would be better to change the regular commit code path to commit the correct offset to begin with, to avoid doing a second "commit offset" RPC? Also wondering how this works with TX -- would there be a new TX just to advance offsets? -- If yes, sounds unnecessary expensive.

sobychacko:
@bandalgomsu Could you take a look at the comments and feedback from @mjsax above and see what you think? Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
bandalgomsu: It runs before each poll (right after doProcessCommits() in the polling loop) and after revoke (right after commitPendingAcks() in onPartitionsRevoked()), so it’s a correction step separate from the normal commit path.

I agree that committing the correct offset in the regular path would be better, but the scope/impact of that change looks too large at the moment,,,

sobychacko:
@bandalgomsu Where do we stand on the PR? It looks like you addressed most of the concerns that Matthias raised above. The use of `lastPollNextOffsets` Map with `putAll()` correctly accumulates across polls per the suggestion.

 One clarification I want to check. When `nextOffsets` is empty, the code clears `lastPollNextOffsets`. I traced through the tests and see this works because `fixTxOffsetsIfNeeded()` runs before the next poll. But should we instead only clear entries that have been successfully fixed?

 Here's my concern: if `max.poll.records` exactly equals the transaction size, Poll 1 returns records with `nextOffsets` (including leader epoch), but `position()` hasn't stepped over the marker yet so no fix happens. Then Poll 2 is empty, clears `lastPollNextOffsets`, and now position has advanced past the marker. The fix runs, but `nextOffset` is null so we lose the leader epoch - which was the whole point of using `nextOffsets()` API per [KAFKA-17600.](https://issues.apache.org/jira/browse/KAFKA-17600)

Maybe clear entries only after they've been used in a successful fix, something like `toFix.keySet().forEach(this.lastPollNextOffsets::remove)` after the commit? What do you think?

  Could you add a brief comment in `savePositionsIfNeeded()` explaining that:
  - `savedPositions` is for detecting user seeks
  - `lastPollNextOffsets` is for TX offset correction with leader epoch

sobychacko:
@bandalgomsu Any chance you can take a look into my comments? We have a release coming up next week and it would be nice to include these changes in the `4.1.0-M1` release. Thanks!

bandalgomsu:
Make sense 
thanks for feedback 👍

sobychacko:
@bandalgomsu Could you add your name as an author to the classes you modified? Thanks!

sobychacko:
Manually merged via https://github.com/spring-projects/spring-kafka/commit/0ae324a738c8350d3ea4394b60e2ea75102c7272.

sobychacko:
@bandalgomsu Thanks for the PR!

Issue #4174, opened 2025-11-20
Add ackMode attribute to @KafkaListener for per-listener acknowledgment configuration


K-jun98:
**Expected Behavior**

  Add ackMode attribute to @KafkaListener, similar to existing concurrency and autoStartup:

```
@Service
  public class MyListener {
      @KafkaListener(topics = "critical", ackMode = "MANUAL")
      public void handleCritical(String msg, Acknowledgment ack) {
          ack.acknowledge();
      }

      @KafkaListener(topics = "normal", ackMode = "BATCH")
      public void handleNormal(String msg) { }
  }

  Configuration simplified to:
  @Bean
  public ConcurrentKafkaListenerContainerFactory<String, String> kafkaListenerContainerFactory(...) {
      // Single factory with default ack mode
      factory.getContainerProperties().setAckMode(AckMode.BATCH);
      return factory;
  }
```

  API Design

```
  public @interface KafkaListener {
      /**
       * Override the container factory's ackMode for this listener.
       * Valid values: RECORD, BATCH, TIME, COUNT, COUNT_TIME, MANUAL, MANUAL_IMMEDIATE
       * Supports SpEL #{...} and property placeholders ${...}
       */
      String ackMode() default "";
  }

```

**Current Behavior**
  Currently, using different acknowledgment modes for different listeners requires creating multiple container factory beans:

```
@Configuration
  public class KafkaConfig {
      @Bean
      public ConcurrentKafkaListenerContainerFactory<String, String> manualAckFactory(...) {
          factory.getContainerProperties().setAckMode(AckMode.MANUAL);
          return factory;
      }

      @Bean
      public ConcurrentKafkaListenerContainerFactory<String, String> batchAckFactory(...) {
          factory.getContainerProperties().setAckMode(AckMode.BATCH);
          return factory;
      }
  }
```

 ```
 @Service
  public class MyListener {
      @KafkaListener(topics = "critical", containerFactory = "manualAckFactory")
      public void handleCritical(String msg, Acknowledgment ack) {
          ack.acknowledge();
      }

      @KafkaListener(topics = "normal", containerFactory = "batchAckFactory")
      public void handleNormal(String msg) { }
  }
```

  This creates unnecessary boilerplate and maintenance overhead.

**Context**

  Our application handles different message types with different reliability requirements:
  - Critical messages (orders, payments): MANUAL ack for precise control
  - Normal messages (notifications): BATCH ack for performance
[...]

PR #4175, opened 2025-11-22
Add per-listener ackMode attribute to @KafkaListener

K-jun98:
Allows individual Kafka listeners to specify different acknowledgment modes without creating multiple container factory beans. This addresses the need to handle different reliability requirements across listeners in the same application (e.g., critical transactions with manual acknowledgment, notifications with batch acknowledgment, and analytics with record acknowledgment).

Key changes:
- Added ackMode() attribute to @KafkaListener annotation supporting all ContainerProperties.AckMode values (RECORD, BATCH, TIME, COUNT, COUNT_TIME, MANUAL, MANUAL_IMMEDIATE)
- Supports SpEL expressions and property placeholders
- Endpoint-level ackMode overrides factory default when specified
- Added resolveAckMode() to KafkaListenerAnnotationBeanPostProcessor to process the annotation attribute
- Updated endpoint infrastructure (KafkaListenerEndpoint, AbstractKafkaListenerEndpoint, AbstractKafkaListenerContainerFactory) to store and apply ackMode
- Added comprehensive tests in KafkaListenerAckModeTests

Example usage:
  @KafkaListener(topics = "critical", ackMode = "MANUAL") @KafkaListener(topics = "notifications", ackMode = "BATCH")

resolve: GH-4174

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/listener-annotation.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/receiving-messages/listener-annotation.adoc
@@ -189,6 +189,20 @@ public void listen(String data, Acknowledgment ack) {
 }
 ----
 
+Starting with version 4.0.1, you can override the container factory's default `AckMode` directly on the `@KafkaListener` annotation using the `ackMode` attribute:
+
+[source, java]
+----
+@KafkaListener(id = "manual", topics = "myTopic", ackMode = "MANUAL")
+public void listen(String data, Acknowledgment ack) {
+    ...
+    ack.acknowledge();
+}
+----
+
+The `ackMode` attribute accepts string values corresponding to `ContainerProperties.AckMode` enum values: `RECORD`, `BATCH`, `TIME`, `COUNT`, `COUNT_TIME`, `MANUAL`, or `MANUAL_IMMEDIATE`.
+This eliminates the need to create separate container factories solely for different acknowledgment modes.
+
 [[consumer-record-metadata]]
 == Consumer Record Metadata
 


spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerContainerFactory.java
```diff
@@ -457,6 +457,12 @@ else if (this.autoStartup != null) {
 				.acceptIfNotNull(endpoint.getConsumerProperties(),
 						instance.getContainerProperties()::setKafkaConsumerProperties)
 				.acceptIfNotNull(endpoint.getListenerInfo(), instance::setListenerInfo);
+
+		// Set ackMode if specified in endpoint (overrides factory default)
+		String ackMode = endpoint.getAckMode();
+		if (ackMode != null) {
```
artembilan: The `acceptIfNotNull()` API is good to go instead of this `if`:
```
.acceptIfNotNull(endpoint.getAckMode(), (ackMode) ->  properties.setAckMode(ContainerProperties.AckMode.valueOf(ackMode.toUpperCase())))
```

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerAckModeTests.java
```diff
@@ -0,0 +1,163 @@
...
+import org.springframework.kafka.test.utils.KafkaTestUtils;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link KafkaListener} ackMode attribute.
+ *
+ * @author GO BEOMJUN
```
artembilan: I don't think legal names go fully in upper case.

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerAckModeTests.java
```diff
@@ -0,0 +1,163 @@
...
+	@Autowired
+	private Config config;
+
+	@Autowired
+	private KafkaListenerEndpointRegistry registry;
+
+	@Test
+	void testAckModeRecordOverride() throws Exception {
+		this.template.send("ackModeRecord", "test-record");
+		assertThat(this.config.recordLatch.await(10, TimeUnit.SECONDS)).isTrue();
```
artembilan: I don't think we need to test an interaction with a consumer to just verify `AckMode` propagation.
That is just a configuration phase feature.
Therefore we might fully don't need an `@EmbeddedKafka`  for this test suite.
Feels like just a mock for `ConsumerFactory` is enough.
And we won't need producer part at all here.

sobychacko:
@K-jun98 These are great changes. I wonder if we want to add this into a point release (`4.0.1`) or wait until `4.1.0`. Since these changes are not very intrusive, it seems ok to me to add this to the point release. cc @artembilan.

K-jun98:
@artembilan Thanks for the review! I’ve applied the changes.

K-jun98:
@sobychacko I’ve applied the changes to 4.0.1 for now.
If you think 4.1.0 is more appropriate, I’m happy to move it there.

artembilan:
Merged as https://github.com/spring-projects/spring-kafka/commit/a3f99f4b82a80072b18616ff4fdda4a306ac1c9c.

@K-jun98 ,

Issue #3618, opened 2024-11-08
DefaultErrorHandler doesn't work with suspend methods


igloo12:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**
3.2.4

**Describe the bug**

Suspend methods don't trigger the default error handler on exception 

**To Reproduce**
I have a bean defined like so

```
    @Bean
    fun defaultErrorHandler(): DefaultErrorHandler {
        val defaultErrorHandler = DefaultErrorHandler()
        defaultErrorHandler.setBackOffFunction { _, exception ->
            if (exception is HttpRequestTimeoutException) {
                FixedBackOff(2000, FixedBackOff.UNLIMITED_ATTEMPTS)
            } else {
                FixedBackOff(2000, 10)
            }
        }
        return defaultErrorHandler
    }
```

If I listen like 

```
    @KafkaListener(
        id = "item_updates",
        groupId = "groupid",
        topics = [TOPIC_ITEM_UPDATED],
    )
   suspend fun syncItemListener(message: String) {
                   throw HttpRequestTimeoutException("google.com", 100)
    }
```
The defaultErrorHandler is not triggered. If I remove the suspend it is triggered

```
    @KafkaListener(
        id = "item_updates",
        groupId = "groupid",
        topics = [TOPIC_ITEM_UPDATED],
    )
   fun syncItemListener(message: String) {
                   throw HttpRequestTimeoutException("google.com", 100)
    }
```

**Expected behavior**

The documentation seems to imply this should work

sobychacko:
@igloo12 We will take a look at this issue soon. In the meantime, could you put your code in a runnable sample app and share it with us?

igloo12:
@sobychacko I have push the sample code here. I created two listeners one working one not 

https://github.com/igloo12/kafka-suspend/blob/main/src/main/kotlin/com/chasecorp/kafkasuspend/KafkaSuspendApplication.kt

sobychacko:
When using suspend functions, the exceptions thrown from those might be wrapped inside a different exception type by the Kotlin coroutine machinery. You don't have that problem in the non-suspend function since it goes through the standard Java exception propagation mechanism, which Spring Kafka detects correctly. Before we dig into this issue, can you try to change your `DefaultErorrHandler` bean so that it checks for the exception cause? Here is the rough idea.

```
@Bean
fun defaultErrorHandler(): DefaultErrorHandler {
    val defaultErrorHandler = DefaultErrorHandler()
[...]

PR #4180, opened 2025-11-23
GH-3618: Consider CommonErrorHandler for suspended functions

mipo256:
I think that the general and the most precise fix for the #3618 would be to just account for the `RecordMessagingMessageListenerAdapter` when applying the `callbackForAsyncFailure` in the listener container. This callback just applies the `CommonErrorHandler` that supposedly implements the retry policy. So, the same thing I guess is also applicable in the case of just `RecordMessagingMessageListenerAdapter`.


artembilan:
Superseded with: https://github.com/spring-projects/spring-kafka/pull/4254.

Thank you!

Issue #4194, opened 2025-12-02
`group.share.delivery.attempt.limit` is not a valid configuration param in Kafka Admin API


sandonjacobs:
You reference a `group.id` level configuration as something that can [alter the delivery count](https://github.com/spring-projects/spring-kafka/blob/9c3718daf1e2c412ad4368c81ef95cc9126e54da/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc?plain=1#L691) for a specific group id for ShareConsumer instances.

There is no such configuration with the name `group.share.delivery.attempt.limit` in the Kafka Admin API - as evidenced in this code example:

```
@SpringBootApplication
public class SpringKafkaQueuesApplication {

	public static void main(String[] args) {
		SpringApplication.run(SpringKafkaQueuesApplication.class, args);
	}

    @Bean
    public ShareConsumerFactory<String, String> shareConsumerFactory() {

        final String bootstrapServers = "localhost:9092";
        final String groupId = "my-share-group";

        Map<String, Object> adminProps = new HashMap<>();
        adminProps.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);

        try (Admin admin = Admin.create(adminProps)) {
            ConfigResource configResource = new ConfigResource(ConfigResource.Type.GROUP, groupId);

            // Default is 5, adjust based on your retry tolerance
            ConfigEntry maxAttempts = new ConfigEntry("group.share.delivery.attempt.limit", "10");

            Map<ConfigResource, Collection<AlterConfigOp>> configs = Map.of(
                    configResource, List.of(new AlterConfigOp(maxAttempts, AlterConfigOp.OpType.SET))
            );

            admin.incrementalAlterConfigs(configs).all().get();
        } catch (ExecutionException e) {
            throw new RuntimeException(e);
        } catch (InterruptedException e) {
            throw new RuntimeException(e);
        }

        Map<String, Object> props = new HashMap<>();
        props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
        props.put(ConsumerConfig.GROUP_ID_CONFIG, groupId);
        props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
        props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
[...]

PR #4195, opened 2025-12-02
GH-4194 Correcting errors in the docs for Share Consumer delivery counts.

sandonjacobs:
Previous documentation referenced a configuration parameter as being exposed via the Kafka Admin API. This is NOT the case, as the `share.group.delivery.attempt.limit` is NOT a valid share group configuration. As such, it is not available via the Admin API. Also, the correct value of `share.group.delivery.count.limit` is a broker-level configuraion. This cannot be overriden at the consumer group level.


sobychacko:
@sandonjacobs Thanks for the PR. The original docs probably got confused with that property that is described in https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka. But, I checked in the [official docs](https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka) and confirmed that property is non-existent. The correct one is what you mentioned - `share.group.delivery.count.limit` which is a broker level property. Thanks for this clarification. We will merge the PR soon.

Issue #4205, opened 2025-12-09
Unable to properly serialize/deserialize null values with JackJsonSerializer/Deserializer in Kotlin


trondgzi:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

4.0.0

**Describe the bug**

When trying to serialize a `null` value with Kotlin a runtime error occurs as the serializer can't return a `null` value, and while trying to deserialize a `null` value the code won't compile.

I will provide a pull request with failing tests and a fix

PR #4206, opened 2025-12-09
GH-4205: Allow serializing/deserializing null values in Kotlin

trondgzi:
Fixes GH-4205 (https://github.com/spring-projects/spring-kafka/issues/4205)

- Replaced byte[] with byte @Nullable [] to be able to return and receive null values

### DIFF
--- a/spring-kafka/src/test/kotlin/org/springframework/kafka/support/serializer/JacksonJsonDeserializerTests.kt
+++ b/spring-kafka/src/test/kotlin/org/springframework/kafka/support/serializer/JacksonJsonDeserializerTests.kt
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2016-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


spring-kafka/src/test/kotlin/org/springframework/kafka/support/serializer/JacksonJsonDeserializerTests.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    @Test
+    fun `Expect deserializing a null value to work`() {
+        val deserializedData = deserializer.deserialize("topic", null)
+        assertNull(deserializedData)
+    }
+
+    private data class DeserializedData(val value: String)
+
+}
```
artembilan: New line in the end of file, please.
`@since 4.0.1`.
Your name to the `@author` list of all the affected classes.
Copy a Copyright from other similar Kotlin classes.

spring-kafka/src/test/kotlin/org/springframework/kafka/support/serializer/JacksonJsonDeserializerTests.kt
```diff
@@ -0,0 +1,29 @@
...
+        val dataToDeserialize = """{"value":"test-data"}""".toByteArray(UTF_8)
+        val deserializedData = deserializer.deserialize("topic", dataToDeserialize)
+
+        assertEquals(DeserializedData("test-data"), deserializedData)
+    }
+
+    @Test
+    fun `Expect deserializing a null value to work`() {
+        val deserializedData = deserializer.deserialize("topic", null)
+        assertNull(deserializedData)
```
artembilan: We prefer to use AssertJ instead of JUnit native API.

Issue #4232, opened 2026-01-06
@BackOff does not recognize the __listener placeholder


Krejtcha:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

Spring Boot 4.0.0, 4.0.1

**Describe the bug**

Having annotations like this on a listener method:

```
@RetryableTopic(
    backOff = @BackOff(delayString = "#{__listener.config.backOffDelay.toMillis()}"),
    attempts = "#{__listener.config.attempts}",
    dltTopicSuffix = "#{__listener.config.dltTopicSuffix}")
@KafkaListener(topics = "#{__listener.config.topics}")
```

The `__listener` placeholder is recognized for `attempts `, `dltTopicSuffix ` and `topics`, but when used also in `@BackOff` `delayString` the following error occurs while starting the application:

```
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'classified': Expression parsing failed
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:610)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:525)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:333)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:371)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:331)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:196)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.instantiateSingleton(DefaultListableBeanFactory.java:1218)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingleton(DefaultListableBeanFactory.java:1184)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:1121)
	at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:993)
	at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:620)
[...]

PR #4234, opened 2026-01-07
GH-4232: Resolve placeholders and SpEL for delayString property

ngocnhan-tran1996:
Fixes: GH-4232

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/BackOff.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/BackOff.java
@@ -125,7 +125,7 @@ public @interface BackOff {
 	 * <li>any of the known {@link DurationFormat.Style}: the {@link DurationFormat.Style#ISO8601 ISO8601}
 	 * style or the {@link DurationFormat.Style#SIMPLE SIMPLE} style &mdash; using
 	 * milliseconds as fallback if the string doesn't contain an explicit unit</li>
-	 * <li>Regular expressions, such as {@code ${example.property}} to use the
+	 * <li>Regular expressions, such as {@code ${example.property}} or {@code #{example.property}} to use the
 	 * {@code example.property} from the environment</li>
 	 * </ul>
 	 *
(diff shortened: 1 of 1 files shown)


spring-kafka/src/main/java/org/springframework/kafka/annotation/BackOff.java
```diff
@@ -95,7 +95,7 @@
 	 * <li>any of the known {@link DurationFormat.Style}: the {@link DurationFormat.Style#ISO8601 ISO8601}
 	 * style or the {@link DurationFormat.Style#SIMPLE SIMPLE} style &mdash; using
 	 * milliseconds as fallback if the string doesn't contain an explicit unit</li>
-	 * <li>Regular expressions, such as {@code ${example.property}} to use the
+	 * <li>Regular expressions, such as {@code ${example.property}} or {@code #{example.property}} to use the
```
artembilan, line 98: I think such an improvement touches not only `delay`, but the rest of attributes of this annotation.
Makes sense?

ngocnhan-tran1996: Yes, because I noticed this test in `BackOffFactoryTests.java`, so the rest of attributes of this annotation should be resolved via SpEL as well.

https://github.com/spring-projects/spring-kafka/blob/7c2f894c53935b6287f74a120bb9cdc97de5712e/spring-kafka/src/test/java/org/springframework/kafka/annotation/BackOffFactoryTests.java#L272-L274

artembilan: Yeah. I see. So, an 'EmbeddedValueResolver` is able to evaluate some basic SpEL. What we have missed is an access to listener from scope. Now propagating evaluation context from the annotation processor we solved a missing part.
What I only ask here is to mention SpEL support in Javadocs for the rest of string attributes in this annotation.
Thanks

ngocnhan-tran1996: Sorry, I forgot to update the rest of string attributes

artembilan:
Thank you for contribution @ngocnhan-tran1996 !

Please, consider in the future add more info into the commit message: https://cbea.ms/git-commit/

Issue #4230, opened 2026-01-02
Duplicated metric "spring.kafka.listener" increment


victorpasqualino:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

4.0.1, 4.0.0, 3.5.6

**Describe the bug**

When configuring @KafkaListener with @RetryableTopic the metrics is being incremented twice.

The problem seems to be on https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2822

When the retryableTopic is disabled the listener class is [RecordMessagingMessageListenerAdapter](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordMessagingMessageListenerAdapter.java) and the metric register on [line](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2822) will be ignored 

but when I have retryableTopic enabled the listener class is [KafkaBackoffAwareMessageListenerAdapter](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/KafkaBackoffAwareMessageListenerAdapter.java#L78) and this one delegates to [RecordMessagingMessageListenerAdapter](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/RecordMessagingMessageListenerAdapter.java)

The problem is when the method [isListenerAdapterObservationAware](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L1244-L1246) check returns false because, as I said, listener is of class  KafkaBackoffAwareMessageListenerAdapter.

In my mind in this case should also be ignored because KafkaBackoffAwareMessageListenerAdapter delegates to RecordMessagingMessageListenerAdapter.

The metric is registered twice because the observation is the same in both instances and both are calling observation.stop()

**To Reproduce**

https://github.com/victorpasqualino/spring-kafka-duplicated-metric

**Expected behavior**

I was expecting to the metric be incremented only once

**Sample**

https://github.com/victorpasqualino/spring-kafka-duplicated-metric

[...]

PR #4238, opened 2026-01-10
GH-4230: Enhance listener observation handling with additional adapter support

Flamme1004K:
I'm submitting a PR for issue #4230. Although the KafkaBackoffAwareMessageListenerAdapter class delegates to RecordMessagingMessageListenerAdapter, it was observed that metrics were being recorded incorrectly because it wasn't adding metrics despite RecordMessagingMessageListenerAdapter. 

To resolve this, I modified the code to set `isListenerAdapterObservationAware` to `true` even when it's a KafkaBackoffAwareMessageListenerAdapter. Additionally, when adding observations, I ensured that if the KafkaBackoffAwareMessageListenerAdapter class delegates to RecordMessagingMessageListenerAdapter, it also adds the metrics.

Thank you for your attention.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -986,8 +986,8 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 		private void setupObservationRegistry(ObservationRegistry observationRegistry) {
 			if (this.listener != null) {
 				MessageListener<?, ?> target = unwrapDelegateIfAny(this.listener);
-				if (target instanceof RecordMessagingMessageListenerAdapter<?, ?> adapter) {
-					adapter.setObservationRegistry(observationRegistry);
+				if (RecordMessagingMessageListenerAdapter.class.equals(target.getClass())) {
+					((RecordMessagingMessageListenerAdapter<?, ?>) target).setObservationRegistry(observationRegistry);
 					this.isListenerAdapterObservationAware = true;
 				}
 			}


victorpasqualino:
I did something very similar, with a few differences: https://github.com/spring-projects/spring-kafka/pull/4240

Flamme1004K:
@victorpasqualino good Idea! 

I got inspired by your code and tried refactoring it. Can you take a look?

- [211c631](https://github.com/spring-projects/spring-kafka/pull/4238/commits/211c6313eee43311ec5cf21e3f1287b915f1fc48)

victorpasqualino:
> @victorpasqualino good Idea!
> 
> I got inspired by your code and tried refactoring it. Can you take a look?
> 
>     * [211c631](https://github.com/spring-projects/spring-kafka/pull/4238/commits/211c6313eee43311ec5cf21e3f1287b915f1fc48)

Looks good to me. Thank you

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
```diff
@@ -46,10 +47,12 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 /**
  * @author Soby Chacko
- * @since 3.2.7
+ * @author Hyoungjune Kim
+ * @since 4.0.2
```
artembilan: This is not correct.
The class was created really in that version, so you cannot modify it.
Plus, the version you target is still not correct since I fill like we need to back-port the fix to `3.3.x` as well.
Therefore, if we do `@since`, we should aim for the lowest version we are going to back-port to.
Luckily,  we don't need to use `@since` anywhere in your change.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -977,15 +979,27 @@ else if (listener instanceof MessageListener) {
...
+			}
+			MessageListener<?, ?> target = unwrapDelegateIfAny(this.listener);
+			if (target instanceof RecordMessagingMessageListenerAdapter<?, ?> adapter) {
+				adapter.setObservationRegistry(observationRegistry);
+				this.isListenerAdapterObservationAware = true;
 			}
 		}
 
+		private MessageListener<?, ?> unwrapDelegateIfAny(MessageListener<?, ?> listener) {
+			if (listener instanceof KafkaBackoffAwareMessageListenerAdapter<?, ?> backoffAware) {
```
artembilan: The `KafkaBackoffAwareMessageListenerAdapter` is an `AbstractDelegatingMessageListenerAdapter`, so I think any of those should be covered here.
If we go this method, it has to be `static`.

artembilan (changes requested):
Please, sure you run `/gradlew check` locally before pushing.

Thanks

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
Flamme1004K: Thank you! I rollbacked `@since` 

- [214d1c1](https://github.com/spring-projects/spring-kafka/pull/4238/commits/214d1c1a4d8a4f302a265ba6cef3d9e9608f23f5)

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
Flamme1004K: Good! I added `static` to this method 

- [71fadb5](https://github.com/spring-projects/spring-kafka/pull/4238/commits/71fadb5dfa93a159aa3901f1dfba03d7a4d0d3f6)

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
```diff
@@ -195,5 +232,22 @@ void listen(ConsumerRecord<Integer, String> in) {
 		}
 	}
 
+	static class RetryBackOffObservationListener {
+		final CountDownLatch latch = new CountDownLatch(1);
```
artembilan: Every member in the class (including before the last `}` of the last inner class) has to be surrounded with blank lines.
That makes the code much easier to read.
And we do here an Open Source, so our code should be as clean as possible and as a good example for community.

Flamme1004K:
@artembilan Thank you, Code Review! I run `/gradlew check`. result I modifed source code.

- [f39cf62](https://github.com/spring-projects/spring-kafka/pull/4238/commits/f39cf6269dcc1052f862606135367c7444937611)
- [d8d4a66](https://github.com/spring-projects/spring-kafka/pull/4238/commits/d8d4a6656d6239c9faa9a6b75bfdefbffcde5bee)

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -977,15 +980,27 @@ else if (listener instanceof MessageListener) {
...
-						(RecordMessagingMessageListenerAdapter<?, ?>) this.listener;
-				if (recordMessagingMessageListenerAdapter != null) {
-					recordMessagingMessageListenerAdapter.setObservationRegistry(observationRegistry);
-				}
+			setupObservationRegistry(observationRegistry);
+		}
+
+		private void setupObservationRegistry(ObservationRegistry observationRegistry) {
+			if (this.listener == null) {
+				return;
```
artembilan: I think for better `void` digesting it would be better to make it opposite: use `if (this.listener != null) {` instead around the rest of the code in this method.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: Good. No need to add a comment with a confirmation.
I'll see an agreement in the next commit you push.

How about an `AbstractDelegatingMessageListenerAdapter` support?
You didn't comment anything to contradict my suggestion and didn't implement it.
With the 👍 it give a false impression that all the request have been addressed.
Any comments, please?

And another our code style suggestion: the `static` methods should go in the end of class.
Not a big deal: just an info if you'd like to know more about our developing process.

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
```diff
@@ -195,5 +231,22 @@ void listen(ConsumerRecord<Integer, String> in) {
...
+		@RetryableTopic(attempts = "1")
+		@KafkaListener(id = "retryBackOffObservationTest",
+				topics = METRICS_TEST_TOPIC,
+				containerFactory = "retryBackOffObservationListenerContainerFactory")
+		void listen(ConsumerRecord<String, String> record) {
+			try {
+				// NOTHING
+			}
+			finally {
+				latch.countDown();
```
artembilan: I think we don't need any logic here since in the end you still fall back to the `await().untilAsserted()`.
Therefore, such a `latch` is just redundant barrier which really does not contribute to the async flow.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
Flamme1004K: That was such a good suggestion I ended up hitting the emoji button without thinking. Next time I'll make sure to use them appropriately.

I've been thinking about it, and instead of using AbstractDelegatingMessageListenerAdapter, how about checking if it's an instance of DelegatingMessageListener and extracting it if it is? What do you think? 

- [b75ae23](https://github.com/spring-projects/spring-kafka/pull/4238/commits/b75ae23fee7799ca5cda3c4d2096c78794c31563)

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
Flamme1004K: I'm learning about open source. Thank you.

Flamme1004K:
@artembilan  Thank you for your prompt feedback on the code review.

I've addressed the feedback you provided. Please review it.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -977,11 +980,15 @@ else if (listener instanceof MessageListener) {
...
-						(RecordMessagingMessageListenerAdapter<?, ?>) this.listener;
-				if (recordMessagingMessageListenerAdapter != null) {
-					recordMessagingMessageListenerAdapter.setObservationRegistry(observationRegistry);
+			setupObservationRegistry(observationRegistry);
+		}
+
+		private void setupObservationRegistry(ObservationRegistry observationRegistry) {
+			if (this.listener != null) {
+				MessageListener<?, ?> target = unwrapDelegateIfAny(this.listener);
+				if (target instanceof RecordMessagingMessageListenerAdapter<?, ?> adapter) {
```
artembilan: I think we need to think about this change twice.
You see, our existing logic is ` RecordMessagingMessageListenerAdapter.class.equals(this.listener.getClass())`.
I believe even if we extract the `delegate`, we still have to be sure that we talk about exactly the same object.
So, if we do this `instanceof`, then any `RecordMessagingMessageListenerAdapter` extension is a valid candidate for `isListenerAdapterObservationAware`.
But that is really not what we want here.
We want to check exactly only for the `RecordMessagingMessageListenerAdapter` and nothing more.
In some other places we do have some custom `RecordMessagingMessageListenerAdapter` , which must not be treated as an observation aware.
Makes sense?

Flamme1004K: @artembilan I agree. It's correct to change the code to verify it's specifically `RecordMessagingMessageListenerAdapter.class`.

I will modify the code to `RecordMessagingMessageListenerAdapter.class.equals(this.listener.getClass())`.

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MicrometerMetricsTests.java
Flamme1004K: During local testing, intermittent failures in the test code were observed, so the source code was modified to call `latch.countDown()`.

The reason is that the count for `spring.kafka.listener.id `can only be obtained after messages are consumed by the listener. If verification occurs before consumption, a null pointer exception occurs regardless of untilAsserted.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
Flamme1004K: I changed the code!

[878dc39](https://github.com/spring-projects/spring-kafka/pull/4238/commits/878dc397c32e0d61ce6e2c3d8dfa64e9e619ab2e)

artembilan:
@Flamme1004K ,

Flamme1004K:
@artembilan 
Thank you! I will strive to contribute more in the future.

PR #4241, opened 2026-01-11
GH-4236: Add `@Serial` to `serialVersionUID` properties

### LINKED ISSUE #4236: Standardize usage of @Serial annotation for serialVersionUID
Please note that English is not my first language, so I apologize in advance for any awkward wording.

**Expected Behavior**

Code style regarding `serialVersionUID` declarations should be consistent across the project.

Since the project targets Java 17+, it would be reasonable to either:
- consistently apply the `@Serial` annotation (introduced in Java 14) to all `serialVersionUID` fields to explicitly mark them as part of the serialization mechanism, or
- consistently omit the annotation and follow the legacy style throughout the codebase.

**Current Behavior**

There is currently an inconsistency in how `serialVersionUID` is declared.

Most classes declare it without the annotation:

```
public class ConsumerPartitionPausedEvent extends KafkaEvent {

	private static final long serialVersionUID = 1L;
```

However, ConcurrentContainerStoppedEvent explicitly uses the `@Serial` annotation:

```
public class ConcurrentContainerStoppedEvent extends KafkaEvent {

	@Serial
	private static final long serialVersionUID = 1L;
```

**Context**

I recently started exploring the Spring Kafka codebase to better understand its structure and design. During this process, I noticed the difference in how serialVersionUID is declared.

I wanted to ask:

Is there a specific reason why `@Serial` is applied only to ConcurrentContainerStoppedEvent?

Would it be acceptable to standardize this for consistency?

Given that the current Spring boot baseline is Java 17, my assumption is that consistently adding `@Serial` would be a reasonable direction. However, I would be happy to follow whichever approach is preferred by the maintainers (either adding it consistently or removing it from this class).

I raised this while exploring the codebase for learning purposes. If the current style is intended, feel free to close this. If not,  I hope to provide a PR.

programofktw:
As mentioned in GH-4326, while reviewing the Spring Kafka codebase, I noticed that the `@Serial` annotation was used only in the ConcurrentContainerStoppedEvent class.
After reviewing the related code, I could not find a specific reason for this class to differ from other similar Serializable classes, and removing the annotation does not appear to introduce any issues.
For the sake of consistency with the existing serialization style, this PR removes the `@Serial` annotation.
This is my first contribution to an open-source project, so please feel free to point out anything that could be improved.

Thank you for taking the time to review this PR.

**Related Issue**
Closes #4236

### DIFF
--- a/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/ContainerTestUtils.java
+++ b/spring-kafka-test/src/main/java/org/springframework/kafka/test/utils/ContainerTestUtils.java
@@ -16,6 +16,7 @@
 
 package org.springframework.kafka.test.utils;
 
+import java.io.Serial;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Collection;
@@ -132,6 +133,7 @@ public final class ContainerTestUtils {
 
 	private static class ContainerTestUtilsException extends RuntimeException {
 
+		@Serial
 		private static final long serialVersionUID = 1L;
 
 		ContainerTestUtilsException(String message, Throwable cause) {
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/NoProducerAvailableException.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/NoProducerAvailableException.java
@@ -16,6 +16,8 @@
 
 package org.springframework.kafka.core;
 
+import java.io.Serial;
+
 import org.springframework.kafka.KafkaException;
 
 /**
@@ -28,6 +30,7 @@ import org.springframework.kafka.KafkaException;
  */
 public class NoProducerAvailableException extends KafkaException {
 
+	@Serial
 	private static final long serialVersionUID = 1L;
 
 	private final String txIdPrefix;
--- a/spring-kafka/src/main/java/org/springframework/kafka/event/ConcurrentContainerStoppedEvent.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/event/ConcurrentContainerStoppedEvent.java
@@ -16,6 +16,8 @@
 
 package org.springframework.kafka.event;
 
+import java.io.Serial;
+
 /**
  * An event published when a concurrent container is stopped.
  *
@@ -25,6 +27,7 @@ package org.springframework.kafka.event;
  */
 public class ConcurrentContainerStoppedEvent extends KafkaEvent {
 
+	@Serial
 	private static final long serialVersionUID = 1L;
 
 	private final ConsumerStoppedEvent.Reason reason;
--- a/spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerFailedToStartEvent.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/event/ConsumerFailedToStartEvent.java
@@ -16,6 +16,8 @@
 
 package org.springframework.kafka.event;
 
+import java.io.Serial;
+
 /**
  * An event published when a consumer fails to start.
  *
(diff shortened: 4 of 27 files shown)


artembilan:
Thank you for contribution! 
however I believe we need to do opposite to align with modern Java.
Those classes you’ve mentioned come without annotation because they were added to the project way before Java 14.

sorry for pushing your first PR away, but you didn’t give us a chance to answer you on Monday , after weekend 😅

programofktw:
@artembilan 

> sorry for pushing your first PR away, but you didn’t give us a chance to answer you on Monday , after weekend 😅

Not at all! I really appreciated the opportunity to receive feedback from the maintainer.

> Thank you for contribution! however I believe we need to do opposite to align with modern Java. Those classes you’ve mentioned come without annotation because they were added to the project way before Java 14.

While working on this, I assumed that `@Serial` was intentionally not used in order to support users who are still on older Java versions. I apologize for that misunderstanding.

If I understand your comment correctly, it sounds like adding `@Serial` would be the preferred direction. 
Would it be okay if I take care of that change? 
I don’t plan to make any functional changes 
I would only add the `@Serial` annotation to places where 
```
private static final long serialVersionUID = 1L; 
```
is already defined.

If you’re happy for me to take this on, I can complete the work and push an additional commit within 48 hours.

artembilan:
Sounds like a plan ! Thank you!
There is no problem with older versions since this annotation is stripped by the compiler . It is only a source level annotation . More over Java doesn’t fail even at runtime if there is no runtime annotation in classpath. And another point: the project currently minimum is Java 17, so you cannot use even with Java 16 where that annotation is definitelly present.

Please, dedicate some time to learn how Java works.
And looking forward for your contribution !

programofktw:
> Sounds like a plan ! Thank you! There is no problem with older versions since this annotation is stripped by the compiler . It is only a source level annotation . More over Java doesn’t fail even at runtime if there is no runtime annotation in classpath. And another point: the project currently minimum is Java 17, so you cannot use even with Java 16 where that annotation is definitelly present.

Thank you for taking the time to explain this I really appreciate learning these details about how Java works.

I thought I had a solid understanding of Java, but this was a good reminder that there is still a lot for me to learn. I’ll definitely spend more time deepening my understanding.


> Please, dedicate some time to learn how Java works. And looking forward for your contribution !

I’ll proceed with the task you entrusted to me and aim to complete it within the next 48 hours. Once the additional commit is pushed, I’ll mention you, @artembilan.

Thanks again, and I appreciate the guidance!

programofktw:
@artembilan 

I’ve completed the updates. As discussed, I’ve added the `@Serial` annotation to all classes that declare serialVersionUID.

I also verified that all tests pass successfully with ./gradlew test.

Since the scope of this PR has changed, should I update the title to something like "Add `@Serial` annotation to Serializable classes"? Let me know if you prefer a different naming convention.

Thanks again for the guidance!

programofktw:
@artembilan 

>Please, consider to run /gradlew check locally.

Thank you for pointing this out, and I apologize for the inconvenience.

I didn’t have much experience working with project-specific code conventions, so I wasn’t aware of this.
Thanks to your guidance, I’ve learned something new. Thank you for bringing this to my attention.

>And fix all those Checkstyle violations before pushing to PR.

I’ve updated the code accordingly, run ./gradlew check locally, and confirmed that it completes with BUILD SUCCESSFUL. The fixes have been pushed to the PR.

Please let me know if there’s anything else I should address.
Thank you for your continued kind guidance and review.

artembilan:
@programofktw ,

programofktw:
@artembilan 
Without your help, this PR would not have been successful. Thank you very much!

PR #4244, opened 2026-01-14
Fix reference of deprecated serde APIs in docs

jjank:
Hi there,

just some minor doc cleanup: I noticed that the deprecated `JsonDeserializer` and `JsonDeserializer` were still mentioned in a couple of places in the docs.

I replaced them with  `JacksonJsonSerializer` and `JacksonJsonDeserializer` accordingly.

Thanks!


jjank:
Update: I noticed that `JsonMessageConverter` was also still referenced.

I pushed another commit fixing this as well.

jjank:
...as it happens I found that the [custom json serializer example](https://docs.spring.io/spring-kafka/reference/tips.html#tip-json) also references deprecated apis.

I took the liberty and tweaked the example a bit because the specific _disabled_ `WRITE_DATES_AS_TIMESTAMPS` feature in this example is [disabled by default in Jackson 3](https://github.com/FasterXML/jackson/blob/main/jackson3/MIGRATING_TO_JACKSON_3.md#changes-serializationfeature). Instead of searching for a _particular_ feature to `enable(..)` or `disable(..)` I took the liberty of leaving a more general comment in the example code. If that is not specific enough for the docs: Please let me know and I'll add a concrete example.

Thanks!

PR #4253, opened 2026-01-18
Remove TODO comment and unnecessary null checks for getBeanName()

david-parkk:
## Summary

  This PR attempts to resolve the TODO comment in `AbstractMessageListenerContainer.getBeanName()` by adding `@NonNull` annotation and removing what appear to be unnecessary null checks in related classes.

  ## Analysis

  Based on my understanding of the codebase:

  - The `beanName` field is declared as `@NonNull` with default value `"noBeanNameSet"`
  - Therefore, `getBeanName()` should never return `null`
  - The original `@Nullable` annotation seems to have been incorrect
  - The null checks `(getBeanName() == null ? "" : getBeanName())` appear to be unnecessary

  Please let me know if I've misunderstood any aspect of this.

  ## Changes

  - `AbstractMessageListenerContainer.java`: Remove TODO comment, add `@NonNull` to `getBeanName()`
  - `KafkaMessageListenerContainer.java`: Remove unnecessary null check
  - `ShareKafkaMessageListenerContainer.java`: Remove unnecessary null check

  ## Test

  Added `shouldSupportDefaultBeanNameSetting()` test to verify that `getBeanName()` returns the default value `"noBeanNameSet"` instead of `null` when `setBeanName()` is not called.

  ---

I'm happy to make any adjustments based on your feedback. Thank you to the maintainers for taking the time to review this contribution!

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -214,7 +214,6 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	 * Return the bean name.
 	 * @return the bean name.
 	 */
-	//	TODO: work on this @Nullable
 	public String getBeanName() {
 		return this.beanName;
 	}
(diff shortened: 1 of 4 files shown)


david-parkk:
I removed the `@NonNull` from the method since the field already has `@NonNull` annotation - adding it to the getter seemed redundant. Happy to add it back if you prefer explicit annotation on the method.

artembilan:
Thank you!
Will be merged after successful build.

PR #4254, opened 2026-01-18
GH-3618: Support CommonErrorHandler for suspend functions

zooo-code:
Fixes GH-3618(https://github.com/spring-projects/spring-kafka/issues/3618)

## Summary
Support `CommonErrorHandler` (e.g., `DefaultErrorHandler`) for Kotlin suspend functions.

Previously, exceptions from suspend functions were only logged and not propagated to `CommonErrorHandler`. This change sets up the async failure callback for `RecordMessagingMessageListenerAdapter` when `CommonErrorHandler` is explicitly configured.

## Changes
* Setup async failure callback when `CommonErrorHandler` is explicitly configured
* Add test for suspend function with `CommonErrorHandler` in `EnableKafkaKotlinCoroutinesTests.kt`

## Related
* Related to #4180 (similar approach, but that PR has conflicts)

Thanks!! 

Please let me know if any changes are needed. I'm happy to address any feedback promptly.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -928,14 +928,14 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 				this.pollThreadStateProcessor = setUpPollProcessor(false);
 				this.observationEnabled = this.containerProperties.isObservationEnabled();
 
-				MessageListener<K, V> target = this.listener;
-				while (target instanceof DelegatingMessageListener<?>) {
-					@SuppressWarnings("unchecked")
-					DelegatingMessageListener<MessageListener<K, V>> dml =
-							(DelegatingMessageListener<MessageListener<K, V>>) target;
+				Object target = this.listener;
+				while (target instanceof DelegatingMessageListener<?> dml) {
 					target = dml.getDelegate();
 				}
-				if (target instanceof RecordMessagingMessageListenerAdapter<K, V> adapter) {
+				if (target instanceof RecordMessagingMessageListenerAdapter<?, ?>) {
+					@SuppressWarnings("unchecked")
+					RecordMessagingMessageListenerAdapter<K, V> adapter =
+							(RecordMessagingMessageListenerAdapter<K, V>) target;
 					adapter.setCallbackForAsyncFailure(this::callbackForAsyncFailure);
 				}
 			}


spring-kafka/src/test/kotlin/org/springframework/kafka/listener/EnableKafkaKotlinCoroutinesTests.kt
```diff
@@ -217,6 +231,23 @@ class EnableKafkaKotlinCoroutinesTests {
 			return factory
 		}
 
+		@Bean
+		fun commonErrorHandler(): DefaultErrorHandler {
+			return DefaultErrorHandler { record, exception ->
+				commonHandlerInvoked = true
+				commonHandlerLatch.countDown()
```
artembilan, line 234: I don't think we need that `commonHandlerInvoked: boolean`.
Isn't condition `this.config.commonHandlerLatch.await(10, TimeUnit.SECONDS)` not enough to be sure that this method was called?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -946,6 +946,16 @@ else if (listener instanceof MessageListener) {
 			this.isConsumerAwareListener = listenerType.equals(ListenerType.ACKNOWLEDGING_CONSUMER_AWARE)
 					|| listenerType.equals(ListenerType.CONSUMER_AWARE);
 			this.commonErrorHandler = determineCommonErrorHandler();
+			// Setup async failure callback for suspend functions when CommonErrorHandler is explicitly configured
```
artembilan: The async failure callback is not only for Kotlin suspend functions.
So, we need to be a bit generic here and don't mention a possible limitation confusion just for Kotlin.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -946,6 +946,16 @@ else if (listener instanceof MessageListener) {
...
+			if (getCommonErrorHandler() != null && this.listener != null) {
```
artembilan: Why do we do this condition?
Do you see that we have a `setCallbackForAsyncFailure()` call in the `KafkaBackoffAwareMessageListenerAdapter` above in this method?
With your change, it looks like we don't need any conditions at all and just use  `setCallbackForAsyncFailure()` on the `RecordMessagingMessageListenerAdapter`.

Isn't that what we want?
Just don't care what is the specific listener, but always propagate `this::callbackForAsyncFailure` when we have a  `RecordMessagingMessageListenerAdapter`?

spring-kafka/src/test/kotlin/org/springframework/kafka/listener/EnableKafkaKotlinCoroutinesTests.kt
zooo-code: You're right. 
The commonHandlerLatch.await() returning true is sufficient to verify the method was called. 
I'll remove the commonHandlerInvoked boolean.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
zooo-code: I'll remove the comment since the code is self-explanatory!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
zooo-code: I tried removing the getCommonErrorHandler() != null condition so that callbackForAsyncFailure is always set on RecordMessagingMessageListenerAdapter. 

However, this causes 3 existing ObservationTests to fail:
- observationErrorExceptionWhenCompletableFutureReturned — expects Error but gets RecordInRetryException
- observationErrorExceptionWhenMonoReturned — same issue
- endToEnd — expected 4 spans but got 6

When the callback is always set, CompletableFuture/Mono listeners also propagate exceptions to DefaultErrorHandler, which triggers retries and changes the error type.

Should I update these tests to match the new behavior, or would you prefer a different approach?

artembilan: OK. So, this is somehow related to the https://github.com/spring-projects/spring-kafka/pull/4320.
So, we probably need to have that merged first, before we can proceed here.

zooo-code: > OK. So, this is somehow related to the #4320. So, we probably need to have that merged first, before we can proceed here.

Thanks for the context! 
I'll wait for #4320 to be merged first, then rebase and update this PR accordingly.

zooo-code: Hi @artembilan

I've rebased on top of the latest main (which includes #4320 ).

As you suggested, I removed the getCommonErrorHandler() != null condition and now always set setCallbackForAsyncFailure() on RecordMessagingMessageListenerAdapter.

However, the same 3 ObservationTests still fail even after #4320 is merged:
- observationErrorExceptionWhenCompletableFutureReturned — expects Error but gets RecordInRetryException
- observationErrorExceptionWhenMonoReturned — same issue
- endToEnd — expected 4 spans but got 6

This happens because CompletableFuture/Mono listeners now also propagate exceptions through DefaultErrorHandler, which triggers retries.

Should I update these ObservationTests to match the new behavior, or would you prefer a different approach?

artembilan: Well, it depends how we decide to move on.
Do we always want not handled listener error be propagated up to the container?
I think yes, since that is exactly behavior of a sync listener.
Therefore, I believe, for consistency an async one should lead to similar logic.

So, it looks like those test have to be fixed this or other way.
It might not be like a handling of the propagated error, but rather not propagating.
Although I need to look into those tests in details.

For now, if we are agreed that propagation is the way to go, then yes tests has to be fixed whatever way you see.

Would you mind also taking a look into this issue: https://github.com/spring-projects/spring-kafka/issues/4333 and see if that is just a duplication of whatever we are pursuing here?

Thanks

zooo-code: @artembilan Thanks for the guidance! 
I'll update the ObservationTests to accommodate the new behavior and push the changes.
I'll also take a look at #4333.

zooo-code:
@artembilan  I've updated the PR based on your guidance. Please review when you have time!

### Changes
- Removed the getCommonErrorHandler() != null condition — now always set setCallbackForAsyncFailure() on RecordMessagingMessageListenerAdapter
- Fixed 3 failing ObservationTests (obs6/obs7) by configuring a no-retry error handler (FixedBackOff(0, 0))
- All tests pass

Regarding #4333: this PR addresses the same root cause.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -935,10 +936,12 @@ else if (listener instanceof MessageListener) {
 					KafkaBackoffAwareMessageListenerAdapter<K, V> genListener =
 							(KafkaBackoffAwareMessageListenerAdapter<K, V>) this.genericListener;
 					if (genListener.getDelegate() instanceof RecordMessagingMessageListenerAdapter<K, V> adapterListener) {
-						// This means that the async retry feature is supported only for SingleRecordListener with @RetryableTopic.
 						adapterListener.setCallbackForAsyncFailure(this::callbackForAsyncFailure);
 					}
 				}
+				else if (this.listener instanceof RecordMessagingMessageListenerAdapter<K, V> adapter) {
```
artembilan: I think I saw a logic somewhere to extract ultimate delegate recursively calling `getDelegate()`.
I mean what you have so far is not enough, since the provided listener could be packed into several delegation layers.
With that it looks like we don't need `this.genericListener instanceof KafkaBackoffAwareMessageListenerAdapter<?, ?>` condition anymore since the new logic covers that one as well.
Please, revise this `if..else` in favor of more generic delegation extraction.

zooo-code: @artembilan  Updated to use generic delegation extraction as suggested.

Now using the while (target instanceof DelegatingMessageListener) pattern 
(same as the observation setup logic) to recursively unwrap delegation layers. 

This replaces the KafkaBackoffAwareMessageListenerAdapter-specific check and covers all delegation wrappers.

Also removed unused imports (AopUtils, KafkaBackoffAwareMessageListenerAdapter).

All tests pass.

Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -929,15 +928,20 @@ else if (listener instanceof MessageListener) {
...
+						target = delegate;
 					}
+					else {
+						break;
+					}
+				}
+				if (target instanceof RecordMessagingMessageListenerAdapter<?, ?>) {
+					@SuppressWarnings("unchecked")
+					RecordMessagingMessageListenerAdapter<K, V> adapter =
+							(RecordMessagingMessageListenerAdapter<K, V>) target;
```
artembilan, line 938: Why do we need this extra cast?
Cannot we just do `if (target instanceof RecordMessagingMessageListenerAdapter<K, V> adapterListener) {`?
See the code you've just removed for that `KafkaBackoffAwareMessageListenerAdapter` logic.

zooo-code: @artembilan Updated to use instance of RecordMessagingMessageListenerAdapter<K, V> adapter pattern matching directly.

To achieve this, changed target type from MessageListener<?, ?> to MessageListener<K, V> and cast DelegatingMessageListener to DelegatingMessageListener<MessageListener<K, V>> in the loop — the `@SuppressWarnings("unchecked")` is scoped to this delegation cast only, which is unavoidable due to type erasure on DelegatingMessageListener<?>.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -929,15 +928,15 @@ else if (listener instanceof MessageListener) {
...
-							(KafkaBackoffAwareMessageListenerAdapter<K, V>) this.genericListener;
-					if (genListener.getDelegate() instanceof RecordMessagingMessageListenerAdapter<K, V> adapterListener) {
-						// This means that the async retry feature is supported only for SingleRecordListener with @RetryableTopic.
-						adapterListener.setCallbackForAsyncFailure(this::callbackForAsyncFailure);
-					}
+				MessageListener<K, V> target = this.listener;
+				while (target instanceof DelegatingMessageListener<?>) {
+					@SuppressWarnings("unchecked")
+					DelegatingMessageListener<MessageListener<K, V>> dml =
+							(DelegatingMessageListener<MessageListener<K, V>>) target;
```
artembilan: I think this is not correct. The `target` in the `DelegatingMessageListener` could be not a `MessageListener`.
So, we are going to fail with class cast.
How about to do:
```
Object target = this.listener;
```
instead?
We really need those generics only in the end when we cast via pattern expression to the `adapter` variable

zooo-code: @artembilan Updated to use Object target as suggested, following the same pattern as determineListenerType(). 
Since Java does not allow pattern matching from Object to RecordMessagingMessageListenerAdapter<K, V>, the `@SuppressWarnings("unchecked")` is needed for the final cast.

artembilan:
@zooo-code ,

artembilan:
Sorry, closed too early.
It will be closed automatically when PR build is green.

Thanks again!

zooo-code:
@artembilan Thank you so much! 
I am truly thrilled to be able to contribute to Spring Kafka. 
I learned a great deal from your thoughtful feedback and guidance throughout this process. 
Thank you for your hard work in maintaining such a fantastic open-source project. 
I look forward to the opportunity to contribute again in the future. 
Thanks again!

sobychacko:
@zooo-code I want to echo what Artem said above. This is a long awaited feature we wanted to support. Thank you for your contribution implementing it. Much appreciated!

zooo-code:
@sobychacko Thank you so much for your kind words!
Hearing that this was a long-awaited feature makes me even more thrilled to have worked on it.
I'm really glad I could help out the Spring Kafka community.
It was a great experience, and I'll definitely look for more opportunities to contribute!

PR #4257, opened 2026-01-21
Extend AcknowledgingMessageListener to inherit from MessageListener

Lokrip:


sobychacko:
@Lokrip Could you give us some context for why this change was necessary? Thanks!

sobychacko:
Ah, ok. Completely missed that this was only in docs. Thanks for taking care of it.

Issue #4258, opened 2026-01-22
Consumer was paused infinitely since difference in async ack offsets is larger than 1


manhnt27:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

4.0.1

**Describe the bug**

I use producer with transaction mode and async listener. Although the async logic had finished, the consumer was paused by then. It was caused by this line of code in `ackInOrder` function (`KafkaMessageListenerContainer.java`)

`while (!ObjectUtils.isEmpty(deferred) && deferred.get(0).offset() == recordToAck.offset() + 1)`

We are assuming that the next offset to be acked is incremented by one. But in case of transaction and async reply , it is not guaranteed.

**To Reproduce**
It could be caused by any cases that the gap between offsets to be acked is larger than 1. I reproduce it in transaction context, Kafka adds control record offset so the gap is often larger than 1. I send messages and Kafka sends back these offsets: 0,2,4,6. But the async ack handles offsets in the following order: 0,2,6,4.
When the ack flow processes offset 4, deferred offset list is [6]. Sine deferred.offset = 6 <> recordToAck=4, offsetsInBatch will not be cleared and cause consumer paused again and again.
You can easily reproduce if the processing offsets in async ack flow  is not monotonically increasing.

It may link to the [bug](https://github.com/spring-projects/spring-kafka/issues/3295). In this case, we aborted the transaction which makes the gap between offsets larger.

**Expected behavior**

Consumer is not paused and able to receive new messages.

**Sample**

A link to a GitHub repository with a [minimal, reproducible,  sample](https://stackoverflow.com/help/minimal-reproducible-example).

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

manhnt27:
I have submitted my PR. Please have a look at: https://github.com/spring-projects/spring-kafka/pull/4264

PR #4264, opened 2026-01-26
Fix offset gap larger than 1 in case async ack

manhnt27:
-  Implement fix logic for issue #4258 

    - Remove condition `deferred.get(0).getOffset() == recordToAck.offset() + 1`, sine there may be gaps which is larger than 1 for these offsets

   - Implement testcase for this change

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
@@ -725,9 +725,9 @@ public class KafkaMessageListenerContainerTests {
 	}
 
 	@ParameterizedTest(name = "{index} AckMode.{0}")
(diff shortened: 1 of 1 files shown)


manhnt27:
@sobychacko It seems that we have conflicts with branch 3.3.x.
The constructor of ConsumerRecords in branch 3.3.x accepts only 1 argument:

```
public ConsumerRecords(Map<TopicPartition, List<ConsumerRecord<K, V>>> records) {
         this.records = records;
     }

```
While the 4.x version has an extra argument
```
public ConsumerRecords(Map<TopicPartition, List<ConsumerRecord<K, V>>> records, final Map<TopicPartition, OffsetAndMetadata> nextOffsets) {
        this.records = records;
        this.nextOffsets = Map.copyOf(nextOffsets);
    }
```

sobychacko:
@manhnt27 It has been fixed on `3.3.x`.

PR #4265, opened 2026-01-27
Fix infinite pause with async acks on transactional topics

### LINKED ISSUE #4258: Consumer was paused infinitely since difference in async ack offsets is larger than 1

RAJESHREDDY0508:
The ackInOrder method in KafkaMessageListenerContainer assumed contiguous offsets when processing deferred acknowledgments. However, with transactional producers and read_committed isolation, Kafka inserts transaction control records that create gaps in the offset sequence (e.g., 0, 2, 4, 6 instead of 0, 1, 2, 3).

When async acknowledgments arrived out of order, the buggy logic would check if the next deferred offset equals current offset + 1. This check failed with non-contiguous offsets, causing the consumer to pause indefinitely waiting for acknowledgments that would never be processed.

Changed the logic to check if the next deferred offset exists in the batch's offset list, rather than checking for sequential offsets. This correctly handles transaction control records and any other scenarios where Kafka creates offset gaps.

Fixes #4258

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -2114,12 +2114,16 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 					offs.remove(0);
 					ConsumerRecord<K, V> recordToAck = cRecord;
 					if (!CollectionUtils.isEmpty(deferred)) {
-						deferred.sort(Comparator.comparingLong(ConsumerRecord::offset));
-						while (!ObjectUtils.isEmpty(deferred) && deferred.get(0).offset() == recordToAck.offset() + 1) {
-							recordToAck = deferred.remove(0);
-							offs.remove(0);
-						}
-					}
+              deferred.sort(Comparator.comparingLong(ConsumerRecord::offset));
+              // Fix for non-contiguous offsets in transactional/read_committed mode
+              // Check if next offset exists in batch, not if it's exactly +1
+              while (!ObjectUtils.isEmpty(deferred) && !offs.isEmpty()
+                      && offs.get(0).equals(deferred.get(0).offset())) {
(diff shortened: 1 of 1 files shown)


sobychacko:
@RAJESHREDDY0508 Looks like this PR is a duplicate of https://github.com/spring-projects/spring-kafka/pull/4264.

RAJESHREDDY0508:
Thanks for pointing that out, Soby.

I see that #4264 already addresses the same issue. 
I’ll close this PR and follow the other approach instead. 
Appreciate the quick review!

Issue #4303, opened 2026-02-19
assertTopic validation breaks meta-annotated @KafkaListener with programmatic topic resolution


mcebanupgrade:
## Expected Behavior

`@KafkaListener` used as a meta-annotation on a custom annotation should allow topics to be resolved programmatically by a custom `ConcurrentKafkaListenerContainerFactory` — the same pattern that worked in Spring Kafka 3.3.11 and earlier.

## Actual Behavior

Since Spring Kafka 3.3.12 (GH-4170 / PR #4172), `KafkaListenerAnnotationBeanPostProcessor.assertTopic()` requires **exactly one** of `topics`/`topicPartitions`/`topicPattern` to be set at annotation-processing time:

```java
Assert.state(count == 1, "Only one of @Topic or @TopicPartition or @TopicPattern must be provided");
```

This fails with `count == 0` when a custom annotation is meta-annotated with `@KafkaListener` that only specifies `containerFactory` (no topics), because topics are resolved programmatically by the container factory's `createListenerContainer()` method — which runs after the validation.

```
BeanCreationException: Error creating bean with name 'myEventHandler'
  Caused by: IllegalStateException: Only one of @Topic or @TopicPartition or @TopicPattern must be provided
    at o.s.kafka.annotation.KafkaListenerAnnotationBeanPostProcessor.assertTopic(KafkaListenerAnnotationBeanPostProcessor.java:876)
```

## Steps to Reproduce

1. Define a custom annotation meta-annotated with `@KafkaListener`:

```java
@Target(ElementType.METHOD)
@Retention(RetentionPolicy.RUNTIME)
@KafkaListener(containerFactory = "myContainerFactory")
public @interface MyEventListener {
    String id();
}
```

2. Create a custom `ConcurrentKafkaListenerContainerFactory` that resolves topics programmatically:

```java
public class MyContainerFactory extends ConcurrentKafkaListenerContainerFactory<String, Object> {
    @Override
    public ConcurrentMessageListenerContainer<String, Object> createListenerContainer(KafkaListenerEndpoint endpoint) {
        if (endpoint instanceof AbstractKafkaListenerEndpoint akle) {
            // Resolve topic from method parameter type, external config, etc.
            akle.setTopics(resolveTopicFromEndpoint(endpoint));
        }
        return super.createListenerContainer(endpoint);
    }
}
```

3. Use the annotation on a handler method:

```java
@Service
public class MyHandler {
    @MyEventListener(id = "my.handler")
[...]

PR #4172, opened 2025-11-19
GH-4170 : Add KafkaListener Validation (Allow @Topic or @TopicPartition)

moonyoungCHAE:
* As-is
  * When a @KafkaListener is created with both @Topic and @TopicPartition, the listener consumes messages based on @TopicPartition.
  * When @RetryableTopic is added, retry topics are managed based on @Topic.
* To-be
  * Creating a @KafkaListener with both @Topic and @TopicPartition is not allowed.
 
fixes #4170

Issue #4170, opened 2025-11-17
@KafkaListener with Both topics and topicPartitions Breaks @RetryableTopic


moonyoungCHAE:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.3.7

**Describe the bug**

1. When I declare a consumer using `@KafkaListener` and mix `topics` and `topicPartitions`, no error occurs. The consumer ends up consuming messages from the topics specified in `topicPartitions`. I think it would be more user-friendly if a validation exception were thrown in this case. Would it make sense to add such validation?

2. When both `topics` and `topicPartitions` are declared in `@KafkaListener` and the `@RetryableTopic` annotation is used, the retry topic is managed based on `topics`. However, as explained in point 1, the consumer actually receives messages from the topics in `topicPartitions`. As a result, if an error occurs while processing messages from topicPartitions, the retry mechanism does not work correctly.

**To Reproduce**
1. Create a consumer with @KafkaListener that uses both topics and topicPartitions.
2. Add the @RetryableTopic annotation.
3. Send a message to a topicPartition and trigger an exception in the consume

**Expected behavior**

**Sample**

```java
public class MyListener {
    @RetryableTopic(attempts = "3", backoff = @Backoff(delay = 1000, multiplier = 2), topicSuffixingStrategy = TopicSuffixingStrategy.SUFFIX_WITH_INDEX_VALUE)
    @KafkaListener(id = "foo",
            topics = "myTopic1",
            clientIdPrefix = "myClientId", topicPartitions =
            {
                    @TopicPartition(topic = "topic1", partitions = {"0"}),
                    @TopicPartition(topic = "topic2", partitions = {"0"})
            })
    public void listen(String data) {
        throw new RuntimeException("a");
    }
}
```

```java
2025-11-17T18:32:05.694+09:00 ERROR 40508 --- [      foo-0-C-1] o.s.kafka.listener.DefaultErrorHandler   : Failed to determine if this record (topic1-0@9) should be recovered, including in seeks

java.lang.NullPointerException: No destination found for topic: topic1
	at java.base/java.util.Objects.requireNonNull(Objects.java:340) ~[na:na]
	at org.springframework.kafka.retrytopic.DefaultDestinationTopicResolver.doGetDestinationFor(DefaultDestinationTopicResolver.java:230) ~[spring-kafka-3.3.7.jar:3.3.7]
[...]

PR #4304, opened 2026-02-19
GH-4303: Allow @KafkaListener with no topic specification for programmatic resolution

mcebanupgrade:
## Summary

The `assertTopic` validation introduced in GH-4170 / PR #4172 requires exactly one of `topics`/`topicPartitions`/`topicPattern` to be set (`count == 1`). This breaks a valid use case where a meta-annotated `@KafkaListener` intentionally omits all topic specifications because the custom `KafkaListenerContainerFactory` resolves topics programmatically in `createListenerContainer()`.

This PR changes the validation from `count == 1` to `count <= 1` so that:
- **Multiple topic specifications** (`count > 1`) are still rejected with `IllegalStateException`
- **No topic specification** (`count == 0`) is permitted, allowing container factories to resolve topics at runtime

## Use Case

A common pattern for custom annotation-driven listeners:

```java
@Target(ElementType.METHOD)
@KafkaListener(containerFactory = "myFactory")
@interface MyCustomKafkaListener {
}
```

With a custom container factory that resolves topics programmatically:

```java
public class TopicResolvingContainerFactory extends ConcurrentKafkaListenerContainerFactory<String, Object> {

    @Override
    public ConcurrentMessageListenerContainer<String, Object> createListenerContainer(
            KafkaListenerEndpoint endpoint) {
        if (endpoint instanceof AbstractKafkaListenerEndpoint<?, ?> akle) {
            akle.setTopics("programmatically-resolved-topic");
        }
        return super.createListenerContainer(endpoint);
    }
}
```

Before GH-4170, this worked because no validation existed. After GH-4170, the `count == 1` check rejects `count == 0`, throwing:
```
IllegalStateException: Only one of @Topic or @TopicPartition or @TopicPattern must be provided
```

## Changes

- **`KafkaListenerAnnotationBeanPostProcessor.java`**: Changed `Assert.state(count == 1, ...)` to `Assert.state(count <= 1, ...)`
- **`KafkaListenerAnnotationBeanPostProcessorTests.java`**: Added two tests:
  - `shouldRejectMultipleTopicSpecifications` — verifies that specifying both `topics` and `topicPattern` still fails
  - `shouldAllowNoTopicSpecificationForProgrammaticResolution` — verifies that a meta-annotated `@KafkaListener` with no topics works when a custom container factory resolves topics programmatically

Fixes: #4303

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessorTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessorTests.java
@@ -41,6 +41,7 @@ import static org.mockito.Mockito.mock;
 
 /**
  * @author Sanghyeok An
+ * @author Maxim Ceban
  *
  * @since 4.0.0
  */
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessor.java
@@ -150,6 +150,7 @@ import org.springframework.validation.Validator;
  * @author Soby Chacko
  * @author Omer Celik
  * @author Go BeomJun
+ * @author Maxim Ceban
  *
  * @see KafkaListener
  * @see KafkaListenerErrorHandler


artembilan:
Please, ensure DCO requirements.

mcebanupgrade:
@artembilan thanks! Ready for review now

sobychacko:
@mcebanupgrade Can you add your name as an author to the classes changed?

spring-kafka/src/test/java/org/springframework/kafka/annotation/KafkaListenerAnnotationBeanPostProcessorTests.java
```diff
@@ -47,6 +60,26 @@ void ctx_should_be_fail_to_register_bean_when_no_listener_methods_exist() {
...
+	}
+
+	@Test
+	void shouldAllowNoTopicSpecificationForProgrammaticResolution() {
+		AnnotationConfigApplicationContext ctx = new AnnotationConfigApplicationContext();
+		ctx.register(NoTopicSpecConfig.class);
+
+		assertThatNoException().isThrownBy(ctx::refresh);
+		ctx.close();
+	}
```
sobychacko, line 89: The test verifies that no exception is thrown, but doesn't assert that the topic was actually resolved programmatically. Consider verifying via `KafkaListenerEndpointRegistry` that the container's topics contain "programmatically-resolved-topic".

mcebanupgrade: @sobychacko thanks, test updated via https://github.com/spring-projects/spring-kafka/pull/4304/changes/b1fd2b047526c10dd11112de013ccf7a8146bb2e

sobychacko:
@mcebanupgrade Thanks for the PR! It has been merged upstream and back-ported to the relevant branches (See above for details).

Issue #4198, opened 2025-12-04
With a @KafkaListener that returns a Mono, the error handler does not retry


moonyoungCHAE:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

3.3.7

**Describe the bug**

When creating an @KafkaListener that returns a Mono, the error handler does not retry when an exception occurs. I got this bug from [stack overflow report](https://stackoverflow.com/questions/79785344/spring-kafka-3-3-reactive-kafkalistener-commits-offset-immediately-on-error-d).

**To Reproduce**

```java
@Component
public class MyListener {
    @KafkaListener(
            clientIdPrefix = "myNewClient",
            groupId = "aa",
            topics = "monoTest"
    )
    public void listen(ConsumerRecord<?, ?> record) throws MyException {
        System.out.println("get data:  offset" + record.offset());
        throw new MyException();
    }
}
```

```java
@Configuration
public class Config {
    private static final Logger logger = LoggerFactory.getLogger(Config.class);

    @Bean
    public CommonErrorHandler handler() {
        DefaultErrorHandler errorHandler = new DefaultErrorHandler(new FixedBackOff(1L, 20L));
        errorHandler.addRetryableExceptions(MyException.class, TimeoutException.class, ListenerExecutionFailedException.class);
        errorHandler.setRetryListeners((record, ex, attempt) ->
                logger.info("[CUSTOM-RETRY] attempt={} topic={} partition={} offset={} ex={}",
                        attempt, record.topic(), record.partition(), record.offset(), ex.toString())
        );
        return errorHandler;
    }
}
```

**Expected behavior**

When an exception occurs, the message is retried. For non-Mono listeners, retries behave as shown below.
```
2025-12-04T16:07:54.241+09:00  INFO 9045 --- [| adminclient-2] o.apache.kafka.common.metrics.Metrics    : Metrics reporters closed
get data:  offset0
2025-12-04T16:07:54.246+09:00  INFO 9045 --- [ntainer#0-0-C-1] org.example.consumer.Config              : [CUSTOM-RETRY] attempt=1 topic=monoTest partition=0 offset=0 ex=org.springframework.kafka.listener.ListenerExecutionFailedException: Listener method 'public void org.example.consumer.MyListener.listen(org.apache.kafka.clients.consumer.ConsumerRecord<?, ?>) throws org.example.consumer.MyException' threw exception
[...]

PR #4320, opened 2026-03-04
GH-4198: Fix async retry bookkeeping for Mono listener failures

MinChul-Son:
Fixes https://github.com/spring-projects/spring-kafka/issues/4198

## Summary

  This PR fixes retry behavior when a @KafkaListener method has a Mono return type and throws an exception directly (before creating the async result).

  Previously, with async acks enabled (default behavior for async return types), retry could stop after the first attempt because async-ack bookkeeping state could become inconsistent during RecordInRetryException handling.

  This change keeps retry progressing while preserving unrelated async-ack state.

  ## What was happening

  In async failure paths, RecordInRetryException indicates the record should remain in retry flow.
  However, container state handling around that path could lead to retry/pause issues.

  ## Changes

  - Keep async failed records queued when retry is still in progress.
  - On RecordInRetryException, clean async-ack bookkeeping only for retry-target records (instead of clearing all tracked async offsets).
  - Apply this behavior consistently for:
      - single-record async failure handling
      - record listener error handling path
      - batch error handling path
  - Add regression test:
      - AsyncMonoDefaultErrorHandlerRetryTests
      - verifies Mono listener direct-throw + DefaultErrorHandler retry behavior

  ## Why this approach

  A full async-ack state reset can affect unrelated in-flight records.
  This PR removes only retry-target offsets to avoid collateral side effects while fixing the retry stall.

  ## Tests

  Validated with:

  - AsyncMonoDefaultErrorHandlerRetryTests
  - AsyncListenerTests
  - ObservationTests.asyncRetryScopePropagation
  - AsyncAckAfterHandleTests
  - PauseContainerWhileErrorHandlerIsRetryingTests

  ## Scope

  This PR addresses the issue path where the listener directly throws in a Mono return method.
  It does not change semantics for Mono.error(...) completion-path handling.

  Thanks for reviewing. I’m happy to adjust anything based on feedback.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -2581,9 +2581,9 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 				List<ConsumerRecord<K, V>> list, RuntimeException rte) {
 
 			try {
-
 				if (Objects.requireNonNull(this.commonErrorHandler).seeksAfterHandling()
 						|| this.transactionManager != null || rte instanceof CommitFailedException) {
+
 					this.commonErrorHandler.handleBatch(rte, records, this.consumer,
 							KafkaMessageListenerContainer.this.thisOrParentContainer,
 							() -> invokeBatchOnMessageWithRecordsOrList(records, list));


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3068,6 +3091,28 @@ private void invokeErrorHandler(final ConsumerRecord<K, V> cRecord,
 			}
 		}
 
+		private synchronized void removeOffsetsInBatch(Iterable<ConsumerRecord<K, V>> records) {
```
artembilan: Why do we need `synchronized` if this method is called from a single consumer thread?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2994,10 +3005,16 @@ private void invokeErrorHandlerBySingleRecord(FailedRecordTuple<K, V> failedReco
 				catch (Exception ex) { // NO SONAR
 					this.logger.error(ex, "Failed to commit before handling error");
 				}
-				List<ConsumerRecord<?, ?>> records = new ArrayList<>();
-				records.add(cRecord);
-				this.commonErrorHandler.handleRemaining(rte, records, this.consumer,
-						KafkaMessageListenerContainer.this.thisOrParentContainer);
+				try {
+					List<ConsumerRecord<?, ?>> records = new ArrayList<>();
+					records.add(cRecord);
```
artembilan: I wonder if we can have a `List.of(cRecord)` instead and before `try` to be reused in the `catch` afterwards.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3034,13 +3051,19 @@ private void invokeErrorHandler(final ConsumerRecord<K, V> cRecord,
...
-				List<ConsumerRecord<?, ?>> records = new ArrayList<>();
-				records.add(cRecord);
+				List<ConsumerRecord<K, V>> retryRecords = new ArrayList<>();
+				retryRecords.add(cRecord);
 				while (iterator.hasNext()) {
-					records.add(iterator.next());
+					retryRecords.add(iterator.next());
+				}
+				try {
+					this.commonErrorHandler.handleRemaining(rte, new ArrayList<>(retryRecords), this.consumer,
```
artembilan: Why do we need a new copy of the list we have just created before the loop?

spring-kafka/src/test/java/org/springframework/kafka/annotation/AsyncMonoDefaultErrorHandlerRetryTests.java
```diff
@@ -0,0 +1,133 @@
...
+			Map<String, Object> consumerProps = KafkaTestUtils.consumerProps(embeddedKafka,
+					"async-mono-default-eh-group", false);
+			return new DefaultKafkaConsumerFactory<>(consumerProps);
+		}
+
+		@Bean
+		CommonErrorHandler commonErrorHandler(Listener listener) {
+			DefaultErrorHandler errorHandler = new DefaultErrorHandler(new FixedBackOff(1L, 2L));
+			errorHandler.setRetryListeners((consumerRecord, ex, deliveryAttempt) -> {
+				if (TOPIC.equals(consumerRecord.topic())) {
```
artembilan: Why do we need this condition if according to the test environment there is definitely cannot be any other topics?

MinChul-Son: Makes sense. In this test setup there is only one relevant topic, so the condition was redundant. I removed it and simplified the retry-listener assertion.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
MinChul-Son: Agreed. I now create the single-record list once before `try` and reuse it

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
MinChul-Son: You are right, the extra copy was unnecessary.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
MinChul-Son: Good point, thanks. This helper is only invoked on the ListenerConsumer flow (single consumer thread), so I removed `synchronized`.

MinChul-Son:
@artembilan Thanks for the review 😊
I’ve applied the requested changes and updated the implementation.
Could you please re-review when you have time?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2569,20 +2569,32 @@ private void invokeBatchErrorHandler(final ConsumerRecords<K, V> records,
...
+					try {
+						ConsumerRecords<K, V> afterHandling = this.commonErrorHandler.handleBatchAndReturnRemaining(rte,
+								records, this.consumer, KafkaMessageListenerContainer.this.thisOrParentContainer,
+								() -> invokeBatchOnMessageWithRecordsOrList(records, list));
+						if (afterHandling != null && !afterHandling.isEmpty()) {
+							this.remainingRecords = afterHandling;
+							this.pauseForPending = true;
+						}
+					}
+					catch (RecordInRetryException e) {
```
artembilan: Why do we need the same `try...catch` logic in both `if..else` branches?
Cannot we just have one around the whole `try..catch`?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2994,10 +3006,16 @@ private void invokeErrorHandlerBySingleRecord(FailedRecordTuple<K, V> failedReco
 				catch (Exception ex) { // NO SONAR
 					this.logger.error(ex, "Failed to commit before handling error");
 				}
-				List<ConsumerRecord<?, ?>> records = new ArrayList<>();
-				records.add(cRecord);
-				this.commonErrorHandler.handleRemaining(rte, records, this.consumer,
-						KafkaMessageListenerContainer.this.thisOrParentContainer);
+				List<ConsumerRecord<?, ?>> retryRecords = new ArrayList<>();
+				retryRecords.add(cRecord);
```
artembilan: Why don't we use `List.of()` instead?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3034,14 +3052,20 @@ private void invokeErrorHandler(final ConsumerRecord<K, V> cRecord,
...
+					List<ConsumerRecord<?, ?>> retryRecords = new ArrayList<>();
+					retryRecords.add(cRecord);
+					while (iterator.hasNext()) {
+						retryRecords.add(iterator.next());
+					}
+					try {
+						this.commonErrorHandler.handleRemaining(rte, retryRecords, this.consumer,
+								KafkaMessageListenerContainer.this.thisOrParentContainer);
+					}
+					catch (RecordInRetryException e) {
```
artembilan: I feel like the same common `try..catch` around the whole `if..else` could be done here as well.

spring-kafka/src/test/java/org/springframework/kafka/annotation/AsyncMonoDefaultErrorHandlerRetryTests.java
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		@Bean
+		@SuppressWarnings("SpringJavaInjectionPointsAutowiringInspection")
+		KafkaTemplate<String, String> kafkaTemplate(EmbeddedKafkaBroker embeddedKafka) {
+			return new KafkaTemplate<>(new DefaultKafkaProducerFactory<>(KafkaTestUtils.producerProps(embeddedKafka)));
+		}
+
+		@Bean
+		@SuppressWarnings("SpringJavaInjectionPointsAutowiringInspection")
```
artembilan: We really don't pursue IDE-specific warnings mitigations.
If this does not cause a compilation warnings, then we don't need such a suppression.

spring-kafka/src/test/java/org/springframework/kafka/annotation/AsyncMonoDefaultErrorHandlerRetryTests.java
```diff
@@ -0,0 +1,129 @@
...
+
+	@Autowired
+	private Listener listener;
+
+	@Test
+	void retriesMonoErrorWithDefaultErrorHandler() throws Exception {
+		this.kafkaTemplate.send(TOPIC, "data").get(10, TimeUnit.SECONDS);
+
+		assertThat(this.listener.invocations.await(30, TimeUnit.SECONDS)).isTrue();
+		assertThat(this.listener.retryAttempts.await(30, TimeUnit.SECONDS)).isTrue();
```
artembilan, line 67: OK. So, we are checking that `DefaultErrorHandler` with its retry logic works.
How about that new logic in the `removeOffsetsInBatch()`?
Do we need to verify it somehow in our new test?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
MinChul-Son: Yes! I agreed, that's more immutable and safe way.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
MinChul-Son: I refactored invokeBatchErrorHandler to use a single common try/catch around the whole if/else 😁

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
MinChul-Son: I refactored invokeErrorHandler to use a single common try/catch around the whole if/else 👍

spring-kafka/src/test/java/org/springframework/kafka/annotation/AsyncMonoDefaultErrorHandlerRetryTests.java
MinChul-Son: `AsyncMonoDefaultErrorHandlerRetryTests` only verifies that `DefaultErrorHandler` retry is triggered for async/Mono failures.

  To verify the new `removeOffsetsInBatch()` logic itself, I added
  `ConcurrentMessageListenerContainerMockTests.removeOffsetsInBatchForRetryRecords()`.

spring-kafka/src/test/java/org/springframework/kafka/annotation/AsyncMonoDefaultErrorHandlerRetryTests.java
MinChul-Son: Agreed! removed IDE-specific suppressions. Compilation/test are clean without them.

MinChul-Son:
@artembilan Please take another look when you have a moment.
I applied the requested updates and would appreciate a re-review. Many Thanks 👍

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerMockTests.java
```diff
@@ -1285,6 +1286,86 @@ void pruneRevokedPartitionsFromPendingOutOfOrderCommitsCoopAssignor() throws Int
...
+			assertThat(subscribeLatch.await(10, TimeUnit.SECONDS)).isTrue();
+			assertThat(handleRemainingLatch.await(10, TimeUnit.SECONDS)).isTrue();
+			KafkaMessageListenerContainer child = (KafkaMessageListenerContainer) KafkaTestUtils
+					.getPropertyValue(container, "containers", List.class).get(0);
+			Map offsets = null;
+			Map deferred = null;
+			for (int i = 0; i < 20; i++) {
+				offsets = KafkaTestUtils.getPropertyValue(child, "listenerConsumer.offsetsInThisBatch", Map.class);
+				deferred = KafkaTestUtils.getPropertyValue(child, "listenerConsumer.deferredOffsets", Map.class);
+				if ((offsets == null || offsets.isEmpty()) && (deferred == null || deferred.isEmpty())) {
```
artembilan: Consider to use a convenient `CollectionUtils.isEmpty()` instead.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerMockTests.java
```diff
@@ -1285,6 +1286,86 @@ void pruneRevokedPartitionsFromPendingOutOfOrderCommitsCoopAssignor() throws Int
...
+	void removeOffsetsInBatchForRetryRecords() throws InterruptedException {
+		TopicPartition topicPartition = new TopicPartition("foo", 0);
+		Map<TopicPartition, List<ConsumerRecord<String, String>>> recordMap = new LinkedHashMap<>();
+		recordMap.put(topicPartition,
+				List.of(new ConsumerRecord("foo", 0, 0, null, "bar-0"),
+						new ConsumerRecord("foo", 0, 1, null, "bar-1")));
+		ConsumerRecords polledRecords = new ConsumerRecords<>(recordMap, Map.of());
+		AtomicInteger pollCount = new AtomicInteger();
+
+		Consumer consumer = mock(Consumer.class);
```
artembilan: We don't need to specify a class for the `mock()` API.

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerMockTests.java
```diff
@@ -1285,6 +1286,86 @@ void pruneRevokedPartitionsFromPendingOutOfOrderCommitsCoopAssignor() throws Int
...
 	}
 
+	@SuppressWarnings({ "unchecked", "rawtypes" })
+	@Test
+	void removeOffsetsInBatchForRetryRecords() throws InterruptedException {
+		TopicPartition topicPartition = new TopicPartition("foo", 0);
+		Map<TopicPartition, List<ConsumerRecord<String, String>>> recordMap = new LinkedHashMap<>();
+		recordMap.put(topicPartition,
+				List.of(new ConsumerRecord("foo", 0, 0, null, "bar-0"),
+						new ConsumerRecord("foo", 0, 1, null, "bar-1")));
```
artembilan: May we don't use not `foo` and `bar` in our code?

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerMockTests.java
```diff
@@ -1285,6 +1286,86 @@ void pruneRevokedPartitionsFromPendingOutOfOrderCommitsCoopAssignor() throws Int
...
+			rebal.set(invocation.getArgument(1));
+			subscribeLatch.countDown();
+			return null;
+		}).given(consumer).subscribe(any(Collection.class), any());
+		willAnswer(invocation -> {
+			if (pollCount.getAndIncrement() == 0) {
+				rebal.get().onPartitionsAssigned(List.of(topicPartition));
+				return polledRecords;
+			}
+			Thread.sleep(50);
```
artembilan: Why do we need to sleep here?

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerMockTests.java
MinChul-Son: Applied. I replaced the manual null/empty checks with `CollectionUtils.isEmpty` 🤗

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerMockTests.java
MinChul-Son: Updated to use typed `mock()` without an explicit class argument!

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerMockTests.java
MinChul-Son: Actually `foo` and `bar` seem like meaningless name, So I replaced placeholder names with intent-revealing name

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerMockTests.java
MinChul-Son: You're right. I removed `Thread.sleep(50)` from the poll stub. I initially added it to throttle rapid empty polls and give async cleanup time, but synchronization is already covered by `handleRemainingLatch` and the existing state-check loop, so the sleep was unnecessary.

MinChul-Son:
@artembilan I updated and applied request changes, Please re-review when you have time!
Thanks

MinChul-Son:
Sry, I miss clicked `close with comment` instead `comment` button..

spring-kafka/src/test/java/org/springframework/kafka/listener/ConcurrentMessageListenerContainerMockTests.java
artembilan: All good.
But let's consider this for productivity.

I don't need verbal agreements on my comments.
If you are OK with my request, you just do it and push to the PR.
I definitely will see that you have agreed in the change set.
Otherwise we got a duplication of the info and some extra spam in my mail box.
Plus, on your side that takes some time for you to type all those agreement comments.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1511,26 +1512,25 @@ protected void pollAndInvoke() {
...
+			// are handled in the next loop iteration.
+			int capturedRecordsCount = this.failedRecords.size();
 			for (int i = 0; i < capturedRecordsCount; i++) {
-				this.failedRecords.pollFirst();
-			}
-
-			// If any copied and failed record fails to complete due to an unexpected error,
-			// We will give up on retrying with the remaining copied and failed Records.
-			for (FailedRecordTuple<K, V> copyFailedRecord : copyFailedRecords) {
+				FailedRecordTuple<K, V> failedRecord = this.failedRecords.pollFirst();
```
artembilan: Looks like dramatic change is happening here.
The previous logic was to make a safe copy of `this.failedRecords`.
Right now it is not, and more over you do mutation with this `pollFirst()` on the shared collection which could be modified from the other thread.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2581,22 +2581,27 @@ private void doInvokeBatchOnMessage(final ConsumerRecords<K, V> records,
...
-				ConsumerRecords<K, V> afterHandling = this.commonErrorHandler.handleBatchAndReturnRemaining(rte,
-						records, this.consumer, KafkaMessageListenerContainer.this.thisOrParentContainer,
-						() -> invokeBatchOnMessageWithRecordsOrList(records, list));
-				if (afterHandling != null && !afterHandling.isEmpty()) {
-					this.remainingRecords = afterHandling;
-					this.pauseForPending = true;
+			try {
+				if (Objects.requireNonNull(this.commonErrorHandler).seeksAfterHandling()
+						|| this.transactionManager != null || rte instanceof CommitFailedException) {
+					this.commonErrorHandler.handleBatch(rte, records, this.consumer,
```
artembilan, line 2587: You see, removing that blank line before code block body you have made code not readable.
Please, revise.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2581,22 +2581,27 @@ private void doInvokeBatchOnMessage(final ConsumerRecords<K, V> records,
...
+							KafkaMessageListenerContainer.this.thisOrParentContainer,
+							() -> invokeBatchOnMessageWithRecordsOrList(records, list));
+				}
+				else {
+					ConsumerRecords<K, V> afterHandling = this.commonErrorHandler.handleBatchAndReturnRemaining(rte,
+							records, this.consumer, KafkaMessageListenerContainer.this.thisOrParentContainer,
+							() -> invokeBatchOnMessageWithRecordsOrList(records, list));
+					if (afterHandling != null && !afterHandling.isEmpty()) {
```
artembilan, line 2596: I think `CollectionUtils.isEmpty()` is the way to go here.
Probably blank line before this `if` would make it more readable.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3038,47 +3048,75 @@ private void invokeErrorHandlerBySingleRecord(FailedRecordTuple<K, V> failedReco
...
-					|| rte instanceof CommitFailedException) {
-
-				try {
-					if (this.producer == null) {
-						processCommits();
+			List<ConsumerRecord<?, ?>> retryRecords = List.of(cRecord);
+			try {
+				if (Objects.requireNonNull(this.commonErrorHandler).seeksAfterHandling()
+						|| rte instanceof CommitFailedException) {
+					try {
```
artembilan, line 3058: DITTO: blank line for readability.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3038,47 +3048,75 @@ private void invokeErrorHandlerBySingleRecord(FailedRecordTuple<K, V> failedReco
...
+						if (this.producer == null) {
+							processCommits();
+						}
 					}
+					catch (Exception ex) { // NO SONAR
+						this.logger.error(ex, "Failed to commit before handling error");
+					}
+					retryRecords = new ArrayList<>();
+					retryRecords.add(cRecord);
```
artembilan: It looks like this initialization has to go before the `try` instead of `List.of()`.
Otherwise feels like duplication.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
MinChul-Son: I changed it to snapshot-based processing and replaced `pollFirst()` with `removeFirstOccurrence()` so we remove exactly the snapshot entry we are handling, rather than mutating by current head order. How does this approach look to you?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
MinChul-Son: Thanks for the comment. `afterHandling` is `Iterable`, so we can’t apply `CollectionUtils.isEmpty()` to it directly. Could you clarify the intent of your suggestion?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: OK. You didn't get me.

The `if` expression is multi-line.
So, we need a blank line exactly after that, just before its body.
The `try` is OK and it does not need any readability improvements.
Same multi-line rule applies to any code block: method header, loops ifs and even `try-with-resource`.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
artembilan: I see. You are right.
Let's leave it then!
Sorry for the noise.

artembilan:
Merged and back-ported.

PR #4330, opened 2026-03-05
GH-4306: Stop listener observations for filtered records

### LINKED ISSUE #4306: Observations for filtered messages are leaked, filling memory

4.0.2+, 3.3.12+

**Describe the bug**

When configuring a `@KafkaListener` with a `RecordFilterStrategy` via `filter` attribute, the observation is leaked in the metric `spring.kafka.listener.active`. This quickly fills available memory. Additionally, the message is not counted, but this may be correct though.

Image

This happened after updating spring-boot 3.5.9 to spring-boot 3.5.10. So we need a backport to 3.3.x for any fix.

The issue seems to be related to https://github.com/spring-projects/spring-kafka/issues/4230. The `FilteringMessageListenerAdapter` instance in the `ListenerConsumer.listener` field is unwrapped to see the `RecordMessagingMessageListenerAdapter` delegate. Since the filtering adapter does not always call `onMessage()` on the delegate (it is filtering), no `invoke()` call errors or stops the current observation.

**To Reproduce**

Get https://github.com/andi5/spring-kafka-observation-leak and run tests with different versions of spring-boot (build.gradle). They all fail.

**Expected behavior**

No memory leak, the active listeners should get back to zero without load.

**Sample**

https://github.com/andi5/spring-kafka-observation-leak

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

BDARJH:
Fixes #4306

## Problem

When observation is enabled for record listeners, the container starts an observation per record.
For adapter-aware listeners, the container expects the delegate adapter to stop it.

With `FilteringMessageListenerAdapter`, filtered records do not call delegate `onMessage()`.
In that path, the observation could remain active and leak.

## Changes

- Added `ObservationRegistry` support to `FilteringMessageListenerAdapter`.
- Explicitly stop the current observation when a record is filtered out.
- Updated `KafkaMessageListenerContainer` observation setup to propagate `ObservationRegistry`
  through delegating filtering listener chains.

## Tests

- Added filtered-record observation test:
  `verifyFilteredRecordStopsObservation`
- Added retry + filter + observation combination test:
  `verifyFilteredRetryableRecordStopsObservation`
- Verified `MicrometerMetricsTests` passes.

## Notes

The fix is intentionally scoped to the filtered-record short-circuit path to avoid changing
the existing async/reply observation lifecycle behavior.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringMessageListenerAdapter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringMessageListenerAdapter.java
@@ -16,6 +16,8 @@
 
 package org.springframework.kafka.listener.adapter;
 
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationRegistry;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.jspecify.annotations.Nullable;
@@ -23,6 +25,7 @@ import org.jspecify.annotations.Nullable;
 import org.springframework.kafka.listener.AcknowledgingConsumerAwareMessageListener;
 import org.springframework.kafka.listener.MessageListener;
 import org.springframework.kafka.support.Acknowledgment;
+import org.springframework.util.Assert;
 
 /**
  * A {@link MessageListener} adapter that implements filter logic
@@ -32,6 +35,7 @@ import org.springframework.kafka.support.Acknowledgment;
  * @param <V> the value type.
  *
  * @author Gary Russell
+ * @author Jinhui Kim
  *
  */
 public class FilteringMessageListenerAdapter<K, V>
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/FilteringMessageListenerAdapter.java
```diff
@@ -64,6 +69,16 @@ public FilteringMessageListenerAdapter(MessageListener<K, V> delegate,
 		this.ackDiscarded = ackDiscarded;
 	}
 
+	/**
+	 * Set the {@link ObservationRegistry} to stop observations for filtered records.
+	 * @param observationRegistry the observation registry.
+	 * @since 4.0.3
```
artembilan: Please, refine the version for real upcoming release - `4.0.4`: https://github.com/spring-projects/spring-kafka/milestone/289

BDARJH: I’ve added my name to the author list in all affected classes and updated the since tag to 4.0.4.
Thanks for the review. 😄 👍

artembilan:
@BDARJH ,

thank you for the contribution; looking forward for more!

One note: consider to do a good commit messages instead of posting everything to the PR description.
When we open PR, the commit message becomes exactly that PR description automatically. 
Then on merge, I would not need to copy/paste your text to the final commit message.

And another note: no need in squashing commits on contribution.
It is much easier to review and follow progress.
We do squash them in the end on merge.

Just couple remarks how to make your contribution process much easier.

Issue #4332, opened 2026-03-05
Listener with async acks pauses indefinitely


asardaes:
I have a Spring Boot 4.0.3 application. I am using a `@KafkaListener` method with a `RecordFilterStrategy`, `setAckDiscarded(true)` used in its container factory, and a `CompletableFuture` as returned value. I see that the listener pauses pretty much after consuming the first batch even though all messages appear to be either filtered or processed correctly. I looked through the code and, unless I'm missing something, this function has a problem:

https://github.com/spring-projects/spring-kafka/blob/e9ff63ebd3ab282149597ac5d84a69bb7da964bc/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L2174

Let's say my batch has 10 records from a single partition with offsets 0 through 9:

1. Records with offsets 0 through 4 are processed in order, so `offsetsInThisBatch` should contain `[5, 6, 7, 8, 9]`
2. Records with offsets 8 and 9 finish next and are added to `deferredOffsets`
3. Record with offset 5 completes next and it's removed from `offsetsInThisBatch` and this loop runs:

```java
if (!CollectionUtils.isEmpty(deferred)) {
	deferred.sort(Comparator.comparingLong(ConsumerRecord::offset));
	while (!ObjectUtils.isEmpty(deferred)) {
		recordToAck = deferred.remove(0);
		offs.remove(0);
	}
}
```

4. The loop removes 6 and 7 from `offsetsInThisBatch` as it empties `deferredOffsets`, so we end up with `[8, 9]` as pending offsets in `offsetsInThisBatch`

After this, regardless of whether record with offset 6 or 7 finishes next, their offset will never fulfill `if (offs.get(0) == cRecord.offset())`. In fact an `IllegalStateException` should be thrown, but since this happens in [the lambda passed to `completableFutureResult.whenComplete`](https://github.com/spring-projects/spring-kafka/blob/e9ff63ebd3ab282149597ac5d84a69bb7da964bc/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java#L584) and that future isn't tracked anywhere, maybe it's swallowed? And since `offsetsInThisBatch` isn't empty, the consumer is paused.

manhnt27:
Hi @asardaes,
[...]

PR #4334, opened 2026-03-08
Fix consumer pausing infinitely in async ack flow

manhnt27:
Fixes issue #4332
### Problem
PR https://github.com/spring-projects/spring-kafka/pull/4264 fixed offset gap in transactional context by removing condition `deferred.get(0).offset() == recordToAck.offset() + 1`.  However, it misses the case when there are still unacked offset between current acked offset and deferred offset. In this case, it will remove unacked offset instead of deferred offset, so deferred offset still present in offsetInThisBatch and cause consumer pause forever.
This PR's approach is only removing deferred offset when the gap between current acked offset and deferred offset does not contain any unacked offset.

### Changes
* Add function `containsUnackedOffset` to check whether there are any unacked offsets between recordToAck offset and deferred offset
* Add condition `containsUnackedOffset = false` in the while loop to remove offset when there are no such unacked offsets
* Add testcase to verify that change

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -2196,7 +2196,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 					ConsumerRecord<K, V> recordToAck = cRecord;
 					if (!CollectionUtils.isEmpty(deferred)) {
 						deferred.sort(Comparator.comparingLong(ConsumerRecord::offset));
-						// Only remove deferred records if there are not any unacked offsets between the current record's offset and the deferred offset
+						// Only remove deferred records if there are not any unacked offsets between the current recordToAck offset and the deferred offset
 						// Otherwise, we will remove the unacked offset instead of the targeted deferred offset
 						// Next time, we receive the already removed offsets to do ack and couldn't do anything further
 						// This may lead to pausing consumer infinitely.


artembilan:
@sobychacko ,

I have assigned you as reviewer to this PR since it looks like you were involved in the original fix.

Thanks

sobychacko:
ok, will take a look today.

rhanton:
I think we're perhaps encountering this issue after the #4264 was backported to the 3.x versions in 3.3.13. If this fixes the bug, can it be backported to 3.x as well? Thanks for the PR and code @manhnt27 !

artembilan:
The issue this PR is based on is already scheduled for back-port to `4.0.x` and `3.0.x`.
All branches are scheduled to be released next Monday: https://spring.io/projects#release-calendar

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2223,6 +2223,15 @@ else if (cRecord.offset() < offs.get(0)) {
 			}
 		}
 
+		private boolean containsUnackedOffsets(final List<Long> offs, long startingOffest, long endingOffset) {
```
sobychacko: Typo in `startingOffest`.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
```diff
@@ -865,6 +865,77 @@ void testInOrderAckOffsetGapLargerThan1ManualImmediateAck(AckMode ackMode) throw
...
+		containerProps.setGroupId("grp");
+		containerProps.setAckMode(ackMode);
+		containerProps.setAsyncAcks(true);
+		final CountDownLatch latch = new CountDownLatch(5);
+		final List<Acknowledgment> acks = new ArrayList<>();
+		final AtomicReference<IllegalStateException> illegal = new AtomicReference<>();
+		AcknowledgingMessageListener<Integer, String> messageListener = (data, ack) -> {
+			if (latch.getCount() == 5) {
+				try {
+					ack.nack(Duration.ofSeconds(1));
```
sobychacko, line 898: You might want to explain as a comment what the significance of `nack` here.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
sobychacko: A small inline javadoc comment might help here by explaining the purpose of this method.

manhnt27: I fixed the typo.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
manhnt27: As far as I know, it is used to verify we are doing async ack flow here. Any other reasons behind it? I would appreciate to learn more about it. Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
manhnt27: I added java doc comment. Btw, I wonder there are any `Utils`, `Helper` class in the project. We can write common function to check whether there are any elements that lie between lower and upper bound. It would be more suitable to put this function to Utils, wouldn't it? Thanks!

manhnt27:
It seems that we have failed tests in the pipeline. Not sure whether it relates to this PR.

artembilan:
@manhnt27 ,

would you mind taking a look into this new issue: https://github.com/spring-projects/spring-kafka/issues/4342, and confirm my feeling that it is a duplication of what you are working here on?
Thanks

sobychacko:
@artembilan I think this PR fixes #4342 as well — I am pretty sure that it’s the same bug. The cause is the same: in `ackInOrder()`, draining deferred without checking for unacked offsets in between leads to the wrong pairing (the head of offs no longer matches the deferred record being committed).  So is the symptom: the consumer with async acks and out-of-order acks gets stuck paused. And the scenario: for a batch like [10–14], 12, 13, 14 are acked first and go to deferred, then 10 is acked; the buggy loop drains deferred and removes 11, 12, 13 from offs while leaving 14 in offs; when 11 is acked later it doesn’t match `offs.get(0) == 14`, so the batch never clears. I’d suggest closing #4342 as a duplicate of #4332 once this is merged.

spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
sobychacko: I think thats the main reason. I think we can leave it as is.

Issue #4321, opened 2026-03-05
Add error handling strategy (poll-level and listener-level) to Share consumer container


sobychacko:
The share consumer container (`ShareKafkaMessageListenerContainer`) has two error handling problems.

### 1. Poll-level exceptions terminate the consumer thread

The poll loop catches exceptions generically, logs it, sets `exitThrowable`, and breaks out of the loop:

```
catch (Exception e) {
    if (e.getCause() instanceof InterruptedException) {
        Thread.currentThread().interrupt();
    }
    this.logger.error(e, "Error in share consumer poll loop");
    exitThrowable = e;
    break;
}
```

`RecordDeserializationException` is thrown by `poll()`, not by the listener. A single record that fails deserialization terminates the entire consumer thread.

Catch `RecordDeserializationException` before the generic `Exception` catch. `REJECT` the affected record via `consumer.acknowledge(record, AcknowledgeType.REJECT)` using the topic-partition and offset from the exception, log a warning, and continue the poll loop. The important fix here is not killing the thread — the record itself is undeserializable and should be rejected.

### 2. Listener-level exceptions always `REJECT`

`handleProcessingError` unconditionally `REJECT`s the record in both modes:

- Explicit mode: calls `acknowledgment.reject()`
- Implicit mode: calls `consumer.acknowledge(record, AcknowledgeType.REJECT)`

There is no equivalent of the traditional container's `CommonErrorHandler` chain. All exceptions result in permanent rejection regardless of type.

### Proposed changes

**Poll-level fix:**

Add a `RecordDeserializationException` catch in the poll loop before the generic `Exception` catch. Extract the topic-partition and offset from the exception, call `consumer.acknowledge()` with `AcknowledgeType.REJECT`, log a warning, and continue polling.

**Listener-level pluggable error handling:**

Introduce a `ShareConsumerRecordRecoverer` functional interface:

```java
@FunctionalInterface
public interface ShareConsumerRecordRecoverer {
    AcknowledgeType recover(ConsumerRecord<?, ?> record, Exception exception);
}
```

The container calls this from `handleProcessingError` instead of hardcoding `REJECT`. The return value determines how the record is acknowledged.

Provide a `DefaultShareConsumerRecordRecoverer` with defaults:
- `CorruptRecordException` → `AcknowledgeType.REJECT` (permanent, record is bad)
- Everything else → `AcknowledgeType.REJECT`

[...]

PR #4338, opened 2026-03-10
GH-4321: Add error handling to share consumer container

sobychacko:
Fixes: https://github.com/spring-projects/spring-kafka/issues/4321

- A single record that failed deserialization or had a corrupt batch was stopping the whole consumer thread. Catch `RecordDeserializationException` and `CorruptRecordException` from `poll()`, `REJECT` or log and continue, so the thread keeps running and later records are still processed.

- Every listener exception was always REJECTing the record, with no way to RELEASE for transient failures. Introduce `ShareConsumerRecordRecoverer` (default: `DefaultShareConsumerRecordRecoverer`, `REJECT`) so callers can plug in `ACCEPT/RELEASE/REJECT` per failure; wire it on `AbstractShareKafkaMessageListenerContainer` and `ShareKafkaListenerContainerFactory`.

- In explicit mode, failed records were only removed from pending acks and not from `acknowledgmentTimestamps`, which could leak. Clear both on error.

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareConsumerRecordRecovererTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareConsumerRecordRecovererTests.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2025-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.listener;
+
+import org.apache.kafka.clients.consumer.AcknowledgeType;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Unit tests for {@link ShareConsumerRecordRecoverer} built-in implementations.
+ *
+ * @author Soby Chacko
+ * @since 4.1
+ */
+class ShareConsumerRecordRecovererTests {
+
+	private static final ConsumerRecord<String, String> RECORD =
+			new ConsumerRecord<>("test-topic", 1, 42L, "key", "value");
+
+	@Test
+	void rejectingReturnsReject() {
+		assertThat(ShareConsumerRecordRecoverer.REJECTING.recover(RECORD, new RuntimeException("test")))
+				.isEqualTo(AcknowledgeType.REJECT);
+	}
+
+	@Test
+	void releasingReturnsRelease() {
+		assertThat(ShareConsumerRecordRecoverer.RELEASING.recover(RECORD, new RuntimeException("test")))
+				.isEqualTo(AcknowledgeType.RELEASE);
+	}
+
+}


spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -13,3 +13,12 @@ For changes in earlier versions, see xref:appendix/change-history.adoc[Change Hi
...
 The attribute also supports SpEL expressions and property placeholders.
 See xref:kafka/receiving-messages/listener-annotation.adoc[`@KafkaListener` Annotation] for more information.
+
+[[x41-share-consumer-error-handling]]
+=== Share Consumer Error Handling
+
+Share consumer containers now provide configurable error handling:
+
+* **Poll-level**: `RecordDeserializationException` and `CorruptRecordException` from `poll()` are caught so the consumer thread continues; undeserializable records are REJECTed and the next poll proceeds.
+* **Listener-level**: A `ShareConsumerRecordRecoverer` interface decides whether to ACCEPT, RELEASE, or REJECT when the listener throws. The default implementation REJECTs all failures; you can set a custom recoverer on the factory or container to RELEASE transient errors for redelivery.
```
artembilan: One sentence per line.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareConsumerRecordRecoverer.java
```diff
@@ -0,0 +1,49 @@
...
+ * or {@link AcknowledgeType#REJECT REJECT} the failed record.
+ *
+ * <p>The default implementation ({@link DefaultShareConsumerRecordRecoverer}) REJECTs
+ * all exceptions. Users can provide custom implementations to RELEASE records for
+ * transient exceptions (e.g., downstream service timeouts), allowing the broker to
+ * redeliver them to another consumer.
+ *
+ * @author Soby Chacko
+ * @since 4.1
+ * @see DefaultShareConsumerRecordRecoverer
```
artembilan: Why all of these are stuck together?
Cannot it be made more readable?
Like each block over blank line?

spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultShareConsumerRecordRecoverer.java
```diff
@@ -0,0 +1,60 @@
...
+ * }</pre>
+ *
+ * @author Soby Chacko
+ * @since 4.1
+ * @see ShareConsumerRecordRecoverer
+ */
+public class DefaultShareConsumerRecordRecoverer implements ShareConsumerRecordRecoverer {
+
+	private static final LogAccessor logger = new LogAccessor(
+			LogFactory.getLog(DefaultShareConsumerRecordRecoverer.class));
```
artembilan: We don't need `LogFactory` to use here.
See `LogAccessor(Class<?>)` ctor.

Do you foresee any evolution of this class (`DefaultShareConsumerRecordRecoverer`) at all?
I wonder if it could just make it into a property in the `ShareConsumerRecordRecoverer` interface.
Something like:
```
@FunctionalInterface
public interface ShareConsumerRecordRecoverer {

private LogAccessor logger = new LogAccessor(ShareConsumerRecordRecoverer .class);

ShareConsumerRecordRecoverer REJECTING = (record, ex) -> {
        logger.error(exception, () -> "Share consumer record processing failed; rejecting record from "
				+ record.topic() + "-" + record.partition() + "@" + record.offset());
        return AcknowledgeType.REJECT;
    }
```
And with that you may consider to add a `RELEASING` one, too...

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
```diff
@@ -239,6 +241,28 @@ public void setupMessageListener(Object messageListener) {
...
+	/**
+	 * Set the {@link ShareConsumerRecordRecoverer} to use when a listener throws
+	 * an exception. The recoverer determines whether to ACCEPT, RELEASE, or
+	 * REJECT the failed record.
+	 * @param recoverer the recoverer (must not be null)
+	 * @since 4.1
+	 * @see ShareConsumerRecordRecoverer
+	 */
+	public void setShareConsumerRecordRecoverer(ShareConsumerRecordRecoverer recoverer) {
+		Assert.notNull(recoverer, "'recoverer' must not be null");
```
artembilan: I think we can just stop from now checking for null on non-null arguments.
And even don't mention that in Javadocs.
It becomes obvious from signature and from Nullability tools.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
```diff
@@ -239,6 +241,28 @@ public void setupMessageListener(Object messageListener) {
...
+		this.recordRecoverer = recoverer;
+	}
+
+	/**
+	 * Return the configured {@link ShareConsumerRecordRecoverer}.
+	 * @return the recoverer
+	 * @since 4.1
+	 */
+	public ShareConsumerRecordRecoverer getShareConsumerRecordRecoverer() {
```
artembilan: Why do we need `public` getter?
The more methods are visible from the class for me in IDE while typing, the more I get frustrated.

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -68,6 +70,9 @@ public class ShareKafkaListenerContainerFactory<K, V>
 
 	private int concurrency = 1;
 
+	@Nullable
+	private ShareConsumerRecordRecoverer recordRecoverer;
```
artembilan: Put annotation next to type.
Not saying this doesn't work, but for visual reasoning behind the annotation.

spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
```diff
@@ -187,6 +203,10 @@ protected void initializeContainer(ShareKafkaMessageListenerContainer<K, V> inst
 		instance.setApplicationContext(this.applicationContext);
 		instance.setApplicationEventPublisher(this.applicationEventPublisher);
 
+		if (this.recordRecoverer != null) {
```
artembilan: Why does this go to the separate `if`, but not as a part of the `JavaUtils.INSTANCE` chain below?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -352,8 +355,31 @@ public void run() {
...
+						catch (Exception ackEx) {
+							this.logger.error(ackEx, () -> "Failed to reject undeserializable record at "
+									+ tp + " offset " + offset);
+						}
+						continue;
+					}
+					catch (CorruptRecordException e) {
+						// CRC check failure. The client automatically rejects the corrupt batch.
+						this.logger.error(e, () -> "CorruptRecordException during poll; "
+								+ "Kafka client has auto-rejected the corrupt batch");
```
artembilan, line 378: Doesn't look like we need a supplier for log message: it is just single static string, even if you put `+` in between, the compiler makes it a single one in the end.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -445,6 +449,101 @@ void shouldHandleProcessingErrorsInExplicitMode(EmbeddedKafkaBroker broker) thro
...
+		Map<String, Object> consumerProps = createConsumerProps(bootstrapServers, groupId, false);
+		DefaultShareConsumerFactory<String, String> factory = new DefaultShareConsumerFactory<>(consumerProps);
+
+		AtomicInteger goodRecordsReceived = new AtomicInteger();
+		CountDownLatch twoGoodRecordsLatch = new CountDownLatch(1);
+		ContainerProperties containerProps = new ContainerProperties(topic);
+		containerProps.setMessageListener((MessageListener<String, String>) record -> {
+			if (goodRecordValue.equals(record.value())) {
+				if (goodRecordsReceived.incrementAndGet() == 2) {
+					twoGoodRecordsLatch.countDown();
```
artembilan: Looks like you don't verify anything else afterwards, so we probably don't need to counters.
You just can make `CountDownLatch twoGoodRecordsLatch = new CountDownLatch(2);`.
Wait exactly for two counts. Kinda the same what you do with the increment, but more straightforward and without extra code.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -171,4 +176,61 @@ void shouldSupportContainerProperties() {
...
+	@Test
+	void shouldUseDefaultShareConsumerRecordRecovererByDefault() {
+		ContainerProperties containerProperties = new ContainerProperties("test-topic");
+		containerProperties.setMessageListener(messageListener);
+
+		ShareKafkaMessageListenerContainer<String, String> container =
+				new ShareKafkaMessageListenerContainer<>(shareConsumerFactory, containerProperties);
+
+		ShareConsumerRecordRecoverer recoverer = container.getShareConsumerRecordRecoverer();
+		assertThat(recoverer).isNotNull();
```
artembilan: Redundant. The `isInstanceOf()` covers us.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -20,5 +20,6 @@ See xref:kafka/receiving-messages/listener-annotation.adoc[`@KafkaListener` Anno
 Share consumer containers now provide configurable error handling:
 
 * **Poll-level**: `RecordDeserializationException` and `CorruptRecordException` from `poll()` are caught so the consumer thread continues; undeserializable records are REJECTed and the next poll proceeds.
-* **Listener-level**: A `ShareConsumerRecordRecoverer` interface decides whether to ACCEPT, RELEASE, or REJECT when the listener throws. The default implementation REJECTs all failures; you can set a custom recoverer on the factory or container to RELEASE transient errors for redelivery.
+* **Listener-level**: A `ShareConsumerRecordRecoverer` interface decides whether to ACCEPT, RELEASE, or REJECT when the listener throws.
+The default is `ShareConsumerRecordRecoverer.REJECTING`; `RELEASING` is also available. You can set a custom recoverer on the factory or container.
```
artembilan: Still: one sentence per line.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareConsumerRecordRecoverer.java
```diff
@@ -19,25 +19,51 @@
...
+ *
  * @see ShareKafkaMessageListenerContainer
  */
 @FunctionalInterface
 public interface ShareConsumerRecordRecoverer {
 
+	/**
+	 * Logger used by built-in implementations.
+	 */
+	LogAccessor LOGGER = new LogAccessor(ShareConsumerRecordRecoverer.class);
```
artembilan, line 46: Cannot this be `private`?

sobychacko: its an interface and i don't think we can have private instance fields in interfaces. All fields declared in an interface implicitly become `public`, `static` and `final`. No?

artembilan: Right. I see now.
So, if you are OK having this to be available for target implementations of this interface, then we can live it as is.
Otherwise we would need to push this property down to every lambda.
WDYT?

sobychacko: Keeping it as is. I think its ok to have a logger living on the interface; hiding it would mean adding extra components for little benefit.

Issue #4337, opened 2026-03-09
Fix compilation warnings due to dependencies upgrades


artembilan:
Currently we have these set of warning in build logs:
```
/home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/test/java/org/springframework/kafka/listener/CommitOnAssignmentTests.java:199: warning: [removal] ConsumerGroupMetadata(String) in ConsumerGroupMetadata has been deprecated and marked for removal
			willReturn(new ConsumerGroupMetadata("")).given(consumer).groupMetadata();
> Task :spring-kafka:compileTestJava
			           ^
/home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/test/java/org/springframework/kafka/listener/TransactionalContainerTests.java:247: warning: [removal] ConsumerGroupMetadata(String) in ConsumerGroupMetadata has been deprecated and marked for removal
		ConsumerGroupMetadata consumerGroupMetadata = new ConsumerGroupMetadata("group");
		                                              ^
/home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/test/java/org/springframework/kafka/listener/SeekToCurrentOnErrorRecordModeTXTests.java:240: warning: [removal] ConsumerGroupMetadata(String) in ConsumerGroupMetadata has been deprecated and marked for removal
			willReturn(new ConsumerGroupMetadata(CONTAINER_ID)).given(consumer).groupMetadata();
			           ^
/home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/test/java/org/springframework/kafka/listener/DefaultErrorHandlerSeekAfterCommitExceptionBatchListenerTests.java:203: warning: [removal] ConsumerGroupMetadata(String) in ConsumerGroupMetadata has been deprecated and marked for removal
			willReturn(new ConsumerGroupMetadata(CONTAINER_ID)).given(consumer).groupMetadata();
			           ^
/home/runner/work/spring-kafka/spring-kafka/spring-kafka/src/test/java/org/springframework/kafka/listener/DefaultErrorHandlerNoSeeksBatchListenerTests.java:221: warning: [removal] ConsumerGroupMetadata(String) in ConsumerGroupMetadata has been deprecated and marked for removal
			willReturn(new ConsumerGroupMetadata(CONTAINER_ID)).given(consumer).groupMetadata();
			           ^
[...]

PR #4339, opened 2026-03-11
Replace ConsumerGroupMetadata initialization with mock

ngocnhan-tran1996:
ConsumerGroupMetadata will become an interface in Kafka 5.0, so replace direct initialization with a mock implementation.

Fixes gh-4337

Javadoc: [ConsumerGroupMetadata.html#<init>(java.lang.String,int,java.lang.String,java.util.Optional)](https://kafka.apache.org/42/javadoc/org/apache/kafka/clients/consumer/ConsumerGroupMetadata.html#%3Cinit%3E(java.lang.String,int,java.lang.String,java.util.Optional)
)

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/CommitOnAssignmentTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/CommitOnAssignmentTests.java
@@ -195,9 +195,7 @@ public class CommitOnAssignmentTests {
 				this.closeLatch.countDown();
 				return null;
 			}).given(consumer).close();
-			final ConsumerGroupMetadata metadata = mock(ConsumerGroupMetadata.class);
-			given(consumer.groupMetadata()).willReturn(metadata);
-			given(metadata.groupId()).willReturn("");
+			given(consumer.groupMetadata()).willReturn(mock(ConsumerGroupMetadata.class));
 			return consumer;
 		}
 
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessorTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessorTests.java
@@ -79,9 +79,7 @@ public class DefaultAfterRollbackProcessorTests {
 		IllegalStateException illegalState = new IllegalStateException();
 		@SuppressWarnings("unchecked")
 		Consumer<String, String> consumer = mock(Consumer.class);
-		ConsumerGroupMetadata metadata = mock(ConsumerGroupMetadata.class);
-		given(consumer.groupMetadata()).willReturn(metadata);
-		given(metadata.groupId()).willReturn("foo");
+		given(consumer.groupMetadata()).willReturn(mock(ConsumerGroupMetadata.class));
 		MessageListenerContainer container = mock(MessageListenerContainer.class);
 		given(container.getContainerProperties()).willReturn(new ContainerProperties("foo"));
 		processor.process(records, consumer, container, illegalState, true, EOSMode.V2);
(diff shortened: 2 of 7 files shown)


spring-kafka/src/test/java/org/springframework/kafka/listener/CommitOnAssignmentTests.java
```diff
@@ -196,7 +195,9 @@ public Consumer consumer() {
 				this.closeLatch.countDown();
 				return null;
 			}).given(consumer).close();
-			willReturn(new ConsumerGroupMetadata("")).given(consumer).groupMetadata();
+			final ConsumerGroupMetadata metadata = mock(ConsumerGroupMetadata.class);
```
artembilan: We don't need the specific class for the `mock()` API.

spring-kafka/src/test/java/org/springframework/kafka/listener/CommitOnAssignmentTests.java
```diff
@@ -196,7 +195,9 @@ public Consumer consumer() {
...
+			given(consumer.groupMetadata()).willReturn(metadata);
+			given(metadata.groupId()).willReturn("");
```
artembilan: Do we really need to stub this method?
Plus it doesn't look straightforward.
You give a mock for another mock, and stub it only after that.
I mean we have to fully prepare mock before give it to something.

Looks like you do something similar in other tests, too.

spring-kafka/src/test/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessorTests.java
```diff
@@ -79,7 +79,9 @@ void testClassifier() {
 		IllegalStateException illegalState = new IllegalStateException();
 		@SuppressWarnings("unchecked")
 		Consumer<String, String> consumer = mock(Consumer.class);
-		given(consumer.groupMetadata()).willReturn(new ConsumerGroupMetadata("foo"));
+		ConsumerGroupMetadata metadata = mock(ConsumerGroupMetadata.class);
+		given(consumer.groupMetadata()).willReturn(metadata);
+		given(metadata.groupId()).willReturn("foo");
```
artembilan: May we consider to not use `foo/bar`?
More over I don't think we need to stub this method at all.

ngocnhan-tran1996:
In summary, the compiler warnings only reported two issues:

1/ `warning: [removal] ConsumerGroupMetadata(String) in ConsumerGroupMetadata has been deprecated and marked for removal.`
This was the most frequent warning.

2/ `[serial] serializable class TestAbortedException has no definition of serialVersionUID.`
I have fixed this [here.
](https://github.com/spring-projects/spring-kafka/pull/4339/changes#diff-7c58ea33657ee5385cd171dae0b8cff5779a9dcc4251b50a0a343d39df87d0a6R146-R147)


Therefore, there are no compiler warnings anymore.

https://github.com/spring-projects/spring-kafka/actions/runs/22943432322/job/66589938572

spring-kafka/src/test/java/org/springframework/kafka/listener/CommitOnAssignmentTests.java
ngocnhan-tran1996: I have tried commenting out the code and the test fails.

artembilan: That's probably wrong spot for this comment, but that is OK to keep those stubs.
Then let's just see if we can get rid off of class in the `mock()` calls!

spring-kafka/src/test/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessorTests.java
artembilan: Was not addressed

spring-kafka/src/test/java/org/springframework/kafka/listener/TransactionalContainerTests.java
```diff
@@ -244,8 +244,9 @@ private void testConsumeAndProduceTransactionGuts(boolean handleError, AckMode a
 		props.setAssignmentCommitOption(AssignmentCommitOption.ALWAYS);
 		props.setEosMode(eosMode);
 		props.setStopContainerWhenFenced(stopWhenFenced);
-		ConsumerGroupMetadata consumerGroupMetadata = new ConsumerGroupMetadata("group");
+		final ConsumerGroupMetadata consumerGroupMetadata = mock(ConsumerGroupMetadata.class);
```
artembilan: DITTO.
And why is it `final`?
Do we really need to add more code just for sake of more code?
What is the logical purpose in this test to have this variable as `final`?

ngocnhan-tran1996: I just copied this from here, and I think using final doesn't seem very reasonable.

https://github.com/spring-projects/spring-kafka/blob/8ee8d0f70054119b5214c27588c67011b1f2d663/spring-kafka/src/test/java/org/springframework/kafka/listener/CommitOnAssignmentTests.java#L173

Issue #4322, opened 2026-03-05
Add `RENEW` acknowledgment type to `ShareAcknowledgment`


sobychacko:
KIP-1222 (Apache Kafka 4.2) added `AcknowledgeType.RENEW` — a fourth ack type that extends the acquisition lock on a record when processing takes longer than `group.share.record.lock.duration.ms` (default 30s). Without it, long-running records time out and are redelivered to another consumer.

`ShareAcknowledgment` exposes `acknowledge()` (ACCEPT), `release()` (RELEASE), and `reject()` (REJECT). `AcknowledgeType.RENEW` is not wired through currently.

### Changes needed

1. Add `renew()` to `ShareAcknowledgment`
2. Add the corresponding override in `ShareConsumerAcknowledgment` mapping to `AcknowledgeType.RENEW`
3. Update `ShareAcknowledgment` Javadoc to list `RENEW`

`RENEW` is non-terminal — it resets the lock timer but a terminal ack (`ACCEPT`, `RELEASE`, or `REJECT`) is still required afterward. `acknowledgeInternal()` currently uses `compareAndSet(null, type)` which prevents any second call on the same record. This needs to allow `RENEW` followed by a terminal acknowledgment.

PR #4340, opened 2026-03-12
GH-4322: Add `RENEW` acknowledgment type to `ShareAcknowledgment`

sobychacko:
Fixes: https://github.com/spring-projects/spring-kafka/issues/4322

- Add `renew()` to `ShareAcknowledgment` and document RENEW (KIP-1222) in Javadoc; implement in `ShareConsumerAcknowledgment` with `AcknowledgeType.RENEW`.

- Allow RENEW then terminal ack: change `acknowledgeInternal` so RENEW is allowed when state is null or RENEW, and terminal (`ACCEPT/RELEASE/REJECT`) is allowed when state is null or RENEW; only one terminal ack per record.

- Call `onRecordAcknowledged` only for terminal acks in `processQueuedAcknowledgments` so RENEW does not remove the record from pending/timestamp tracking.

- Treat `isAcknowledged()` as terminally acknowledged (false after RENEW until acknowledge/release/reject). Update timeout warning to mention `renew()` and that a terminal ack is still required.

- Add no-op `renew()` to `NoOpShareAck` in `MessagingMessageListenerAdapter`.

- Add reference docs and integration tests.

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
@@ -806,10 +806,8 @@ class ShareKafkaMessageListenerContainerIntegrationTests {
 			ack.acknowledge();
 
 			// Wait for consumer thread to process queued acks (RENEW, RENEW, ACCEPT)
-			Awaitility.await().atMost(10, TimeUnit.SECONDS).untilAsserted(() -> {
-				assertThat(isAcknowledgedInternal(ack)).isTrue();
-				assertThat(getAcknowledgmentTypeInternal(ack)).isEqualTo(AcknowledgeType.ACCEPT);
-			});
+			Awaitility.await().atMost(10, TimeUnit.SECONDS).untilAsserted(() -> assertThat(isAcknowledgedInternal(ack)).isTrue());
+			assertThat(getAcknowledgmentTypeInternal(ack)).isEqualTo(AcknowledgeType.ACCEPT);
 		}
 		finally {
 			container.stop();


spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -24,3 +24,11 @@ Share consumer containers now provide configurable error handling:
 The default is `ShareConsumerRecordRecoverer.REJECTING`; `RELEASING` is also available.
 You can set a custom recoverer on the factory or container.
 * See xref:kafka/kafka-queues.adoc#share-error-handling[Share consumer error handling] in the Kafka Queues documentation.
+
+[[x41-share-acknowledgment-renew]]
+=== ShareAcknowledgment.renew()
+
+`ShareAcknowledgment` now supports `renew()` to extend the acquisition lock on a record when processing may exceed the broker's lock duration (`group.share.record.lock.duration.ms`, default 30 seconds).
+This follows KIP-1222 (Kafka 4.2): RENEW is non-terminal.
+You may call `renew()` multiple times, but you must still call exactly one of `acknowledge()`, `release()`, or `reject()` when processing completes.
```
artembilan: I think all of these details would be better to have in the target chapter.
The `whats-new` is updated for every single major/minor version.
So, it is easy to lose it.

spring-kafka/src/main/java/org/springframework/kafka/support/ShareAcknowledgment.java
```diff
@@ -89,4 +90,29 @@ public interface ShareAcknowledgment {
...
+	 * The acknowledgment will be committed when:
+	 * <ul>
+	 * <li>The next {@code poll()} is called (batched with fetch)</li>
+	 * <li>{@code commitSync()} or {@code commitAsync()} is explicitly called</li>
+	 * <li>The consumer is closed</li>
+	 * </ul>
+	 *
+	 * @throws IllegalStateException if the record has already been terminally
+	 * acknowledged (via {@link #acknowledge()}, {@link #release()}, or {@link #reject()})
+	 * @since 4.0
```
artembilan: `4.1` ?

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -765,6 +766,101 @@ void shouldHandleConcurrentAcknowledgmentAttempts(EmbeddedKafkaBroker broker) th
...
+			ShareAcknowledgment ack = ackRef.get();
+			assertThat(ack).isNotNull();
+
+			// RENEW is non-terminal: call renew() then acknowledge(); record should be terminally acknowledged
+			ack.renew();
+			ack.renew(); // multiple RENEWs allowed
+			ack.acknowledge();
+
+			// Allow consumer thread to process queued acks (RENEW, RENEW, ACCEPT)
+			Thread.sleep(500);
```
artembilan: i think this is not robust solution.
Let's consider to use `await()` instead polling until that `assertThat(isAcknowledgedInternal(ack)).isTrue()` is OK.
See Awaitility feature `untilAsserted()`.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -765,6 +766,101 @@ void shouldHandleConcurrentAcknowledgmentAttempts(EmbeddedKafkaBroker broker) th
...
+		container.start();
+
+		try {
+			assertThat(processedLatch.await(10, TimeUnit.SECONDS)).isTrue();
+			ShareAcknowledgment ack = ackRef.get();
+			assertThat(ack).isNotNull();
+
+			ack.acknowledge();
+			// Second terminal ack must throw
+			assertThatExceptionOfType(IllegalStateException.class)
```
artembilan: There is a dedicated, convenient `assertThatIllegalStateException()`

artembilan (changes requested):
Very cool!

Just couple nit-picks.

One more note, not related to the change.
Looks like that one could be as a `record`:
```
private record PendingAcknowledgment<K, V>(ConsumerRecord<K, V> record, AcknowledgeType type) {

	}
```

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -765,6 +766,101 @@ void shouldHandleConcurrentAcknowledgmentAttempts(EmbeddedKafkaBroker broker) th
...
+
+			// RENEW is non-terminal: call renew() then acknowledge(); record should be terminally acknowledged
+			ack.renew();
+			ack.renew(); // multiple RENEWs allowed
+			ack.acknowledge();
+
+			// Wait for consumer thread to process queued acks (RENEW, RENEW, ACCEPT)
+			Awaitility.await().atMost(10, TimeUnit.SECONDS).untilAsserted(() -> {
+				assertThat(isAcknowledgedInternal(ack)).isTrue();
+				assertThat(getAcknowledgmentTypeInternal(ack)).isEqualTo(AcknowledgeType.ACCEPT);
```
artembilan: I don't think we need to put both assertion into `await()`.
We just need to wait for the first one, and second won't be called until this `await()` has successed.

Issue #4329, opened 2026-03-05
Expose `group.protocol` configuration for Kafka Streams server-side rebalance (KIP-1071)


sobychacko:
KIP-1071 (server-side rebalance for Streams) is GA in Kafka 4.2. Enabled by default on new clusters; existing clusters opt in via `group.protocol=streams`.

`DefaultKafkaConsumerFactory` already warns when `group.protocol=consumer` conflicts with custom partition assignors, but `StreamsBuilderFactoryBean` has no configuration for this. Users must set it via raw properties.

Expose `group.protocol` configuration through `StreamsBuilderFactoryBean` so Streams users can opt in directly.

jad837:
Hi @sobychacko  I would like to contribute to this issue. 

afaik I can see that the current implementation of StreamsBuilderFactoryBean is using constructor 
`streamsBuilderFactoryBean(KafkaStreamsConfiguration streamsConfig, CleanupConfig cleanupConfig)`

So my task would be 
1. Create instance variable for group protocol & expose setter for this property.
2. Then pass the property correctly to kafka.client.StreamsConfig class which currently handlers group.protocol as `classic` (default) or `streams`

Do we also want to add warning of sorts if someone sends partition.assigment.strategy configuration along with group protocol ? Similar to `DefaultKafkaConsumerFactory`
And I believe if a group protocol is passed as raw property like today then as well we should parse this config and warn user if its applicable?

sobychacko:
Hi @jad837 That sounds like a great idea, please contribute a PR. The approach you described sounds correct; we can iron out any details during the PR review process. Thank you!

PR #4373, opened 2026-03-27
GH-4329 Added new property group protocol in StreamsBuilderFactoryBean

jad837:
This PR addresses the issue https://github.com/spring-projects/spring-kafka/issues/4329. It exposes group protocol configuration for streams directly through `StreamsBuilderFactoryBean`

**This PR does the following**

- Expose configuration for Group Protocols in `StreamsBuilderFactoryBean`.

- Adds test for `StreamsBuilderFactoryBean` to test the new configuration.

**Questions**

- Do we want to keep raw property exposure for group protocol? This means that configuration inside factory bean can be null & if both are set then instance variable will override the raw properties configuration. We can add warning in such conditions.

- Currently lowercase/uppercase values can be provided as "classic"/"streams", Do we want to add an assertion to check if values fall inside GroupProtocol enum? Currently values outside of enum will throw exception.

Closes https://github.com/spring-projects/spring-kafka/issues/4329

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/streams.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/streams.adoc
@@ -235,7 +235,7 @@ Starting with version 2.7, the default is to never clean up local state.
 [[streams-group-protocol]]
 == Group Protocol Configuration
 
-Starting with version 4.1.0, you can configure the consumer group protocol used by the underlying Kafka Streams consumers by setting the `groupProtocol` property.
+Starting with version 4.1, you can configure the consumer group protocol used by the underlying Kafka Streams consumers by setting the `groupProtocol` property.
 This allows you to explicitly manage whether the consumers rely on the `classic` protocol or opt into the newer `streams` group protocol introduced in KIP-1071 (Server-side rebalance for Streams, GA in Kafka 4.2).
 While this is enabled by default on new clusters, existing clusters can opt-in using this property.
 It can be configured using `StreamsBuilderFactoryBean.setGroupProtocol()`.


jad837:
Hi @sobychacko can you please provide feedback on the PR?

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -146,6 +151,7 @@ public StreamsBuilderFactoryBean(KafkaStreamsConfiguration streamsConfig, Cleanu
 		Assert.notNull(streamsConfig, STREAMS_CONFIG_MUST_NOT_BE_NULL);
 		Assert.notNull(cleanupConfig, CLEANUP_CONFIG_MUST_NOT_BE_NULL);
 		this.properties = streamsConfig.asProperties();
+		this.applyGroupProtocol();
```
sobychacko: The call to `applyGroupProtocol()` in the constructor is always a no-op — `this.groupProtocol` is always null at construction time. See my review on that method.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -282,6 +288,33 @@ public void setCleanupConfig(CleanupConfig cleanupConfig) {
...
+	 * @return groupProtocol returns {@link GroupProtocol} value defined for this {@link StreamsBuilderFactoryBean}
+	 */
+	public @Nullable GroupProtocol getGroupProtocol() {
+		return this.groupProtocol;
+	}
+
+	/**
+	 * Retrieves and sets group protocol for properties of {@link StreamsBuilderFactoryBean}.
+	 */
+	public void applyGroupProtocol() {
```
sobychacko: if `(this.groupProtocol != null)` guard never passes since it is called from the ctor. The setter sets the field later, but there is no path from the setter back to applying it to `this.properties`. This means the feature silently has no effect. Also, `applyGroupProtocol() `should not be a public method — it is an internal implementation detail and should not be part of the API surface. However, I think that the right fix is to remove `applyGroupProtocol()` entirely and apply the value inline in `start()`, just before initializing KafkaStreams:
```
  if (this.groupProtocol != null) {
      this.properties.setProperty(ConsumerConfig.GROUP_PROTOCOL_CONFIG,
          this.groupProtocol.name().toLowerCase(Locale.ROOT));
  }
  this.kafkaStreams = this.kafkaStreamsCustomizer.initKafkaStreams(
      this.topology, this.properties, this.clientSupplier
  );
```

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
```diff
@@ -110,6 +114,26 @@ protected StreamsBuilder createInstance() {
...
+		};
+		streamsBuilderFactoryBean.setGroupProtocol(testGroupProtocol);
+		streamsBuilderFactoryBean.afterPropertiesSet();
+		StreamsBuilder builder = streamsBuilderFactoryBean.getObject();
+		builder.stream(Pattern.compile("foo"));
+		streamsBuilderFactoryBean.afterSingletonsInstantiated();
+		streamsBuilderFactoryBean.start();
+		StreamsBuilder streamsBuilder = streamsBuilderFactoryBean.getObject();
+		verify(streamsBuilder).build(kafkaStreamsConfiguration.asProperties());
+		assertThat(streamsBuilderFactoryBean.getGroupProtocol()).isEqualTo(GroupProtocol.valueOf(testGroupProtocol.toUpperCase(Locale.ROOT)));
```
sobychacko: The test only verifies that `getGroupProtocol()` returns the correct enum value (the setter/getter round-trip), but never asserts that `group.protocol` actually ends up in the properties passed to `KafkaStreams` at `start()` time. A meaningful assertion would be:
```
  assertThat(streamsBuilderFactoryBean.getStreamsConfiguration())
      .containsEntry(ConsumerConfig.GROUP_PROTOCOL_CONFIG,
          testGroupProtocol.toLowerCase(Locale.ROOT));
```

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -282,6 +288,33 @@ public void setCleanupConfig(CleanupConfig cleanupConfig) {
 		this.cleanupConfig = cleanupConfig; // NOSONAR (sync)
 	}
 
+	/**
+	 * Set group protocol to be used by {@link StreamsBuilderFactoryBean}.
+	 * Only allowed values are from {@link GroupProtocol}
+	 * @param groupProtocol groupProtocol value as given in {@link org.apache.kafka.clients.consumer.GroupProtocol}
+	 */
+	public void setGroupProtocol(String groupProtocol) {
```
sobychacko: The setter should accept `GroupProtocol` directly rather than a `String`:
```
public void setGroupProtocol(GroupProtocol groupProtocol) {
    Assert.notNull(groupProtocol, "'groupProtocol' must not be null");
    this.groupProtocol = groupProtocol;
}
```
  A typo like `fb.setGroupProtocol("strems")` compiles fine today but blows up at runtime with an `IllegalArgumentException`. With the typed setter, `fb.setGroupProtocol(GroupProtocol.STREMS)`
   simply does not compile. The valid options are also immediately discoverable via IDE auto-complete. 
The typical usage would then be:
```
@Bean
public StreamsBuilderFactoryBeanConfigurer groupProtocolConfigurer() {
    return fb -> fb.setGroupProtocol(GroupProtocol.STREAMS);
}
```

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -282,6 +288,33 @@ public void setCleanupConfig(CleanupConfig cleanupConfig) {
 		this.cleanupConfig = cleanupConfig; // NOSONAR (sync)
 	}
 
+	/**
+	 * Set group protocol to be used by {@link StreamsBuilderFactoryBean}.
+	 * Only allowed values are from {@link GroupProtocol}
+	 * @param groupProtocol groupProtocol value as given in {@link org.apache.kafka.clients.consumer.GroupProtocol}
```
sobychacko: wrong package for `GroupProtocol`.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/streams.adoc
```diff
@@ -232,6 +232,23 @@ By default, when the factory bean is stopped, the `KafkaStreams.cleanUp()` metho
 Starting with version 2.1.2, the factory bean has additional constructors, taking a `CleanupConfig` object that has properties to let you control whether the `cleanUp()` method is called during `start()` or `stop()` or neither.
 Starting with version 2.7, the default is to never clean up local state.
 
+=== Group Protocol Configuration
```
sobychacko: Might want to consider adding an extra anchor - something like `[[streams-group-protocol]` and then use it to link from `whats-new` entry.

sobychacko:
@jad837 Thanks for the PR. I added some comments, please take a look. Add your name as an `@author` to all the classes you changed. Also, follow these commit guidelines: https://cbea.ms/git-commit/

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -365,6 +388,10 @@ public void start() {
 				try {
 					Assert.state(this.properties != null,
 							"streams configuration properties must not be null");
+					if (this.getGroupProtocol() != null) {
+						this.properties.setProperty(ConsumerConfig.GROUP_PROTOCOL_CONFIG,
```
artembilan: I think we need to step back and revise requirements for this property to be exposed as a top-level option.
It does not look like too hard to configure it from existing Kafka API.
It does not justify how this property has a priority over many other config properties not exposed.
Therefore it raises a question: why don't we expose any others then?
Plus there is a conflict: user have to provide `KafkaStreamsConfiguration` into this class instance, and then there is a `setStreamsConfiguration(Properties streamsConfig)`.
Having this `GROUP_PROTOCOL_CONFIG` exposed as a top-level option leads to very hard paradox of choice.

My vote is do nothing (may some docs) and don't make over-complication for the API we expose and use.
Having different places to set the same thing in the end, more over allowing to override and mutate, would lead to more questions and maintenance.

I'm opened for any other arguments.

Thanks

sobychacko: @artembilan  Thanks for raising this — worth discussing.
 
I think `group.protocol` is not just another config knob though. KIP-1071 introduces server-side rebalance for Kafka Streams and is GA in Kafka 4.2. It's a fundamental protocol change that directly impacts every Streams user upgrading -- existing clusters need to explicitly opt in via `group.protocol=streams`, and users on new clusters may want to lock in `classic` during a transition period. Without a typed API here, this critical decision is invisible unless users happen to stumble across the KIP docs. That's a real usability gap worth closing.

artembilan: After off-line discussion with @sobychacko we decide to go with this top-level property for visibility.
Even if that property will be deprecated and removed eventually in Apache Kafka itself and we would need to deal with that resolution agian, it is still better to make our API as friendly and transparent as possible for easier migration.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -282,6 +287,24 @@ public void setCleanupConfig(CleanupConfig cleanupConfig) {
...
+	public void setGroupProtocol(GroupProtocol groupProtocol) {
+		Assert.notNull(groupProtocol, "'groupProtocol' must not be null");
+		this.groupProtocol = groupProtocol;
+	}
+
+	/**
+	 * Get groupProtocol defined for this {@link StreamsBuilderFactoryBean}.
+	 * @return groupProtocol returns {@link GroupProtocol} value defined for this {@link StreamsBuilderFactoryBean}
+	 */
+	public @Nullable GroupProtocol getGroupProtocol() {
```
artembilan: There is real value to get access to this property.
Please, consider to remove the getter, unless you have a strong argument why do we need it.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
artembilan: I think we need to emit the WARN log message in the case when property is provided already in the `properties` and we are going to override it with this our `groupProtocol`.
And probably only when they are not equal 😄

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -365,6 +388,10 @@ public void start() {
 				try {
 					Assert.state(this.properties != null,
 							"streams configuration properties must not be null");
+					if (this.getGroupProtocol() != null) {
+						this.properties.setProperty(ConsumerConfig.GROUP_PROTOCOL_CONFIG,
+								this.getGroupProtocol().name().toLowerCase(Locale.ROOT));
```
artembilan: There is no need in the `Locale` since we turn pre-compiled enum value to string, not opposite to cause some vulnerability.

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
```diff
@@ -110,6 +116,30 @@ protected StreamsBuilder createInstance() {
...
+		builder.stream("foo");
+		streamsBuilderFactoryBean.afterSingletonsInstantiated();
+		streamsBuilderFactoryBean.start();
+		StreamsBuilder streamsBuilder = streamsBuilderFactoryBean.getObject();
+		verify(streamsBuilder).build(kafkaStreamsConfiguration.asProperties());
+		assertThat(streamsBuilderFactoryBean.getStreamsConfiguration())
+				.containsEntry(ConsumerConfig.GROUP_PROTOCOL_CONFIG, testGroupProtocol.name().toLowerCase(Locale.ROOT));
+		// Need to remove group protocol config to ensure other tests not get affected
+		// due to pattern usage with streams group protocol
+		kafkaStreamsConfiguration.asProperties().remove(ConsumerConfig.GROUP_PROTOCOL_CONFIG);
```
artembilan: Well, this is a bug.
We must not mutate shared instance.
And looks like that is exactly what happens here: we provide shared `kafkaStreamsConfiguration` and then we mutate its `properties` adding a new one in the `StreamsBuilderFactoryBean`.
I think `KafkaStreamsConfiguration` must return a fresh copy of its internal state.
We also have to make a copy in the `setStreamsConfiguration(Properties streamsConfig)`.
And must return a copy from the `getStreamsConfiguration()`.
All of that to avoid bug on mutation of shared objects.

However I believe that has to be done in the separate PR and back-ported.
And then when you rebase this one, all should be good in this test.

Although I still have concerns for the purpose of this test.
I believe we should aim for asserting whatever `ConsumerConfig.GROUP_PROTOCOL_CONFIG` is in the end for the `StreamsBuilder`, not `streamsBuilderFactoryBean`.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
jad837: Thanks for raising this. I have removed the method and added the property check in start() method. 
I see that there is also a createInstance() method which returns a streamsBuilder, Do we need to add this in that method as well?

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
jad837: Updated this as per the recommendation.

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
jad837: Right now we are adding group.protocol to `KafkaStreams` and not to `StreamsBuilder`. Might have to add group protocol property through `createInstance`/`createStreamsBuilder` in case we need to add it for streamsbuilder. I am not sure though if streamsbuilder can use group.protocol as it seems to mainly handle topology configurations?

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
jad837: Removed the Locale usage.

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
artembilan: We just must not modify shared instance.
That's the point.
An internal `properties` in the `StreamsBuilderFactoryBean` has to be isolated and modified before it is propagated down to the Kafka Streams API.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
jad837: I have added a warning, can you please check if wording is correct?

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
jad837: Removed getter.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/streams.adoc
jad837: Updated whats-new entry with an anchor to stream doc

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
jad837: I have updated the comment.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
jad837: I have updated implementation to use GroupProtcol enum for Streams. For all the methods.

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
jad837: Test case is updated, Can you please check comment on same ? https://github.com/spring-projects/spring-kafka/pull/4373#discussion_r3011957331 
Please suggest if we need to update implementation for createInstance/createStreamsBuilder.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
artembilan: Thanks.
First of all that is too many `this.properties.get(ConsumerConfig.GROUP_PROTOCOL_CONFIG)` call.
Consider to extract it into a local variable.

I think the wording in the warning message should be like:

> The 'group.protocol=%s' property is overridden with '%s'."

Nothing more.
Also, consider to use `LogMessage.format()` instead of string.
That would be unfortunate to calculate string for nothing when warn is not enabled.

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
jad837: `StreamsBuilderFactoryBean` creates kafkaStreams in `start()` so Ideally the test case is just ensuring that the passed down properties when creating `KafkaStreams` are correct. 
As we are not able to directly check KafkaStreams configuration properties, better option would be to drop the test case assuming that KafkaStreams will use properties correctly. 
Please share if you have an idea on how can we handle this better.

As for mutable shared instance of properties I will create a separate PR/issue it.

artembilan: Well, what I see so far:
1. `StreamsBuilderFactoryBean` manages a `KafkaStreams` object after its start.
2. The `KafkaStreams` is based on the provided properties (including our custom `ConsumerConfig.GROUP_PROTOCOL_CONFIG`)
3. Those props are stored in the `KafkaStreams.applicationConfigs` .
4. So, in our test we should check that provided `ConsumerConfig.GROUP_PROTOCOL_CONFIG` is really propagated down there:
```
KafkaStreams kafkaStreams = streamsBuilderFactoryBean.getKafkaStreams();
StreamsConfig streamsConfig = KafkaTestUtils.getPropertyValue(kafkaStreams, "applicationConfigs", StreamsConfig.class);
streamsConfig.get(ConsumerConfig.GROUP_PROTOCOL_CONFIG);
```

What do I miss?

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -302,6 +306,16 @@ public void setCleanupConfig(CleanupConfig cleanupConfig) {
 		this.cleanupConfig = cleanupConfig; // NOSONAR (sync)
 	}
 
+	/**
+	 * Set group protocol to be used by {@link StreamsBuilderFactoryBean}.
+	 * Only allowed values are from {@link GroupProtocol}
+	 * @param groupProtocol groupProtocol value as given in {@link GroupProtocol}
```
artembilan, line 311: `@since 4.1`

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
artembilan: > too many `this.properties.get(ConsumerConfig.GROUP_PROTOCOL_CONFIG)`

was not addressed.

artembilan (changes requested):
> 17 commits

I think I was clear when I asked to **rebase** your branch to the latest `upstream main`: https://stackoverflow.com/questions/804115/when-do-you-use-git-rebase-instead-of-git-merge.

Now this is very hard to review.

jad837:
> > 17 commits
> 
> I think I was clear when I asked to **rebase** your branch to the latest `upstream main`: https://stackoverflow.com/questions/804115/when-do-you-use-git-rebase-instead-of-git-merge.
> 
> Now this is very hard to review.

Hi give me sometime I will rebase and push. 

I am trying to use this property via test project once done I will rebase and ask for review. 
Sorry for Inconvenience, some tests are flaky when running locally so I have to push to ensure things are working correctly.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -67,4 +67,19 @@ The provided exception handlers support multiple Kafka Streams DLQ enabling stra
...
+[[x41-kafka-streams-group-protocol]]
+=== Kafka Streams Group Protocol Configuration
+
+The `StreamsBuilderFactoryBean` now supports configuring the `groupProtocol` property.
+This allows you to explicitly manage whether the underlying Kafka Streams consumers rely on the `CLASSIC` protocol or opt into the newer `STREAMS` group protocol (KIP-1071: Server-side rebalance for Streams).
+
+See xref:streams.adoc#streams-group-protocol[Group Protocol Configuration] for more details.
+
+[[x41-kafka-streams-group-protocol]]
+=== Kafka Streams Group Protocol Configuration
```
artembilan, line 73: I think there is some duplication in this doc now.
Please, revise your two new sections.

spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -302,6 +305,17 @@ public void setCleanupConfig(CleanupConfig cleanupConfig) {
...
+	 * @since 4.1.1
```
artembilan: I'm not why have you decided to use this version if we asked explicitly for `4.1`

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
```diff
@@ -110,6 +116,29 @@ protected StreamsBuilder createInstance() {
 		assertThat(streamsBuilderFactoryBean.getTopology()).isNotNull();
 	}
 
+	@ParameterizedTest
+	@EnumSource(GroupProtocol.class)
+	public void testBuildWithGroupProtocolProperty(GroupProtocol testGroupProtocol) throws Exception {
+		streamsBuilderFactoryBean = new StreamsBuilderFactoryBean(kafkaStreamsConfiguration) {
+			@Override
+			protected StreamsBuilder createInstance() {
+				return spy(super.createInstance());
```
artembilan: Why do you `spy` if you don't verify eventually?

spring-kafka/src/test/java/org/springframework/kafka/config/StreamsBuilderFactoryBeanTests.java
```diff
@@ -110,6 +116,29 @@ protected StreamsBuilder createInstance() {
...
+			}
+		};
+		streamsBuilderFactoryBean.setGroupProtocol(testGroupProtocol);
+		streamsBuilderFactoryBean.afterPropertiesSet();
+		StreamsBuilder builder = streamsBuilderFactoryBean.getObject();
+		builder.stream("foo");
```
artembilan: Please, no unprofessional  `foo/bar` language in new code.
I know there is a number of them around to be fixed one day, but that doesn't mean we should continue following such a bad habit.

artembilan:
@jad837 ,

PR #4374, opened 2026-03-27
GH-4369: Pass explicit acknowledgment mode as consumer override property

maxwell-balla:
Fixes #4369
  When "setExplicitShareAcknowledgment(true)" is set on the container properties, the "share.acknowledgement.mode=explicit" property is now passed to the consumer factory via a new "createShareConsumer" overload that accepts override properties, instead of relying on the user to set it manually in the factory configuration.

  Signed-off-by: Maxwell Balla <ballamaxwell7@gmail.com>

Issue #4369, opened 2026-03-26
Setting explicit acknowledgment via ShareKafkaListenerContainerFactory container properties does not enable explicit acknowledgment mode for Share Groups


j-tim:
**In what version(s) of Spring for Apache Kafka are you seeing this issue?**

4.0.4

**Describe the bug**

The documentation states there are two ways to enable explicit acknowledgment in shared groups in Spring Kafka:

* [Option 1](https://docs.spring.io/spring-kafka/reference/4.1/kafka/kafka-queues.html#_option_1_using_kafka_client_configuration) is working `props.put(ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG, "explicit");`.
* [Option 2](https://docs.spring.io/spring-kafka/reference/4.1/kafka/kafka-queues.html#_option_2_using_spring_container_configuration): is not working `factory.getContainerProperties().setExplicitShareAcknowledgment(true);`

**To Reproduce**

Steps to reproduce the behavior.

1. Start the application

```bash
./mvnw spring-boot:run -Dspring-boot.run.profiles=explicit
```

2. Publish a valid message to the topic

```bash
curl -X POST "http://localhost:8080/produce?message=Hi"
```
3. Observe the logs, the setting on the `ShareConsumerConfig` show we are still using `implicit` acknowledgment mode

```log
2026-03-26T11:05:46.702+01:00  INFO 55104 --- [           main] o.a.kafka.common.config.AbstractConfig   : ShareConsumerConfig values: 
        share.acknowledgement.mode = implicit
```
4. Observe the logs, we can't acknowledge the message we cause we are in `implicit` acknowledgment mode. Resulting in this exception in our Spring Kafka consumer:

```log
2026-03-26T11:06:02.761+01:00 ERROR 55104 --- [Container#0-C-1] s.k.l.ShareKafkaMessageListenerContainer : Failed to process queued acknowledgment for record: ConsumerRecord(topic = high-throughput-topic, partition = 0, leaderEpoch = 0, offset = 2, CreateTime = 1774519562719, deliveryCount = 1, serialized key size = -1, serialized value size = 2, headers = RecordHeaders(headers = [], isReadOnly = false), key = null, value = Hi)

java.lang.IllegalStateException: Implicit acknowledgement of delivery is being used.
[...]

PR #4376, opened 2026-03-29
Introduce `ShareAckMode` enum for share consumers

sobychacko:
The 4.0 boolean `explicitShareAcknowledgment` only controlled whether the listener received a non-null acknowledgment object. It had no effect on Kafka's `share.acknowledgement.mode`, so `acknowledge()` would throw `IllegalStateException` when the factory used the Kafka default (implicit mode).

`ContainerProperties.ShareAckMode` (EXPLICIT, MANUAL, IMPLICIT) maps each use case to a clear name, following the same pattern as regular containers disabling auto.commit. For EXPLICIT and MANUAL the container now enforces `share.acknowledgement.mode=explicit` via consumer override properties, regardless of factory configuration.

Other related changes:

  - `RecordDeserializationException` handler now guards `acknowledge()` against IMPLICIT mode
  - `createShareConsumer` variant added to `ShareConsumerFactory` for consumer-level overrides
  - 4.0 dead code and naming issues cleaned up

Deprecated methods keep the 4.0 API intact.
Reference docs and What's New include a migration guide. Test changes.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/ShareConsumerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/ShareConsumerFactory.java
@@ -59,6 +59,7 @@ public interface ShareConsumerFactory<K, V> {
 	 */
 	default ShareConsumer<K, V> createShareConsumer(@Nullable String groupId, @Nullable String clientId,
 			Map<String, Object> overrideProperties) {
+
 		return createShareConsumer(groupId, clientId);
 	}
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
@@ -33,6 +33,7 @@ import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.ShareConsumer;
+import org.apache.kafka.clients.consumer.internals.ShareAcknowledgementMode;
 import org.apache.kafka.common.Metric;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.TopicPartition;
@@ -201,7 +202,7 @@ public class ShareKafkaMessageListenerContainer<K, V>
 		GenericMessageListener<?> listener = (GenericMessageListener<?>) messageListener;
 		Assert.state(listener != null, "'messageListener' cannot be null");
 
-		if (containerProperties.getShareAckMode() == ContainerProperties.ShareAckMode.MANUAL) {
+		if (ContainerProperties.ShareAckMode.MANUAL.equals(containerProperties.getShareAckMode())) {
 			boolean isAcknowledgingListener = listener instanceof AcknowledgingShareConsumerAwareMessageListener;
 			Assert.state(isAcknowledgingListener,
 					"ShareAckMode.MANUAL requires an AcknowledgingShareConsumerAwareMessageListener. " +
@@ -209,7 +210,7 @@ public class ShareKafkaMessageListenerContainer<K, V>
 					"Either use ShareAckMode.EXPLICIT or provide a listener that accepts a ShareAcknowledgment parameter.");
 		}
 
-		if (containerProperties.getShareAckMode() == ContainerProperties.ShareAckMode.EXPLICIT
+		if (ContainerProperties.ShareAckMode.EXPLICIT.equals(containerProperties.getShareAckMode())
 				&& listener instanceof AcknowledgingShareConsumerAwareMessageListener) {
 			this.logger.warn("Listener is an AcknowledgingShareConsumerAwareMessageListener but "
 					+ "ShareAckMode.EXPLICIT is active; the acknowledgment argument will be null. "
@@ -304,22 +305,22 @@ public class ShareKafkaMessageListenerContainer<K, V>
 			this.genericListener = listener;
 			this.clientId = consumerClientId;
 			ContainerProperties containerProperties = getContainerProperties();
+			String groupId = ShareKafkaMessageListenerContainer.this.getGroupId();
 
 			ContainerProperties.ShareAckMode shareAckMode = containerProperties.getShareAckMode();
-			this.isManualAckMode = shareAckMode == ContainerProperties.ShareAckMode.MANUAL;
+			this.isManualAckMode = ContainerProperties.ShareAckMode.MANUAL.equals(shareAckMode);
 			this.ackTimeoutMs = containerProperties.getShareAcknowledgmentTimeout().toMillis();
 
-			if (shareAckMode == ContainerProperties.ShareAckMode.IMPLICIT) {
+			if (ContainerProperties.ShareAckMode.IMPLICIT.equals(shareAckMode)) {
 				ShareConsumerRecordRecoverer recoverer =
 						ShareKafkaMessageListenerContainer.this.getShareConsumerRecordRecoverer();
-				if (recoverer != ShareConsumerRecordRecoverer.REJECTING) {
+				if (ShareConsumerRecordRecoverer.REJECTING != recoverer) {
 					this.logger.warn("A custom ShareConsumerRecordRecoverer is configured but its acknowledgment "
 							+ "decision will be ignored in ShareAckMode.IMPLICIT — all records are always "
 							+ "ACCEPTed by the broker regardless of processing outcome.");
 				}
 				this.consumer = ShareKafkaMessageListenerContainer.this.shareConsumerFactory.createShareConsumer(
-						ShareKafkaMessageListenerContainer.this.getGroupId(),
-						consumerClientId);
+						groupId, consumerClientId);
 			}
 			else {
 				// EXPLICIT and MANUAL both use Kafka client explicit mode so the container
@@ -327,16 +328,15 @@ public class ShareKafkaMessageListenerContainer<K, V>
 				Object configured = ShareKafkaMessageListenerContainer.this.shareConsumerFactory
 						.getConfigurationProperties()
 						.get(ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG);
-				if (configured != null && "implicit".equalsIgnoreCase(configured.toString())) {
+				if (configured != null && ShareAcknowledgementMode.IMPLICIT.name().equalsIgnoreCase(configured.toString())) {
 					this.logger.warn("Factory configuration has share.acknowledgement.mode=implicit "
 							+ "but ShareAckMode." + shareAckMode + " is active; the container will "
 							+ "override it with explicit mode. To use implicit mode, set "
 							+ "ShareAckMode.IMPLICIT on the container properties instead.");
 				}
 				this.consumer = ShareKafkaMessageListenerContainer.this.shareConsumerFactory.createShareConsumer(
-						ShareKafkaMessageListenerContainer.this.getGroupId(),
-						consumerClientId,
-						Map.of(ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG, "explicit"));
+						groupId, consumerClientId,
+						Map.of(ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG, ShareAcknowledgementMode.EXPLICIT.name()));
 			}
 
 			this.consumer.subscribe(Arrays.asList(containerProperties.getTopics()));
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
@@ -78,9 +78,7 @@ import static org.mockito.Mockito.verify;
 				"share-container-mixed-ack-test",
 				"share-container-lifecycle-test",
 				"share-container-recoverer-release-test",
-				"share-container-deser-error-test",
-				"share-container-explicit-reject-test",
-				"share-container-implicit-error-test"
+				"share-container-deser-error-test"
 		},
 		partitions = 1,
 		brokerProperties = {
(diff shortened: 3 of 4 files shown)


sobychacko:
The 4.0 `explicitShareAcknowledgment` boolean never touched Kafka's `share.acknowledgement.mode`. Kafka defaults that to implicit, which means `consumer.acknowledge()` throws `IllegalStateException`. The container called it unconditionally, so the feature only worked if the user had manually configured share.acknowledgement.mode=explicit in the factory.

`ShareAckMode (EXPLICIT, MANUAL, IMPLICIT)` replaces the boolean, following the same pattern as regular containers disabling `auto.commit`. For EXPLICIT and MANUAL, the container now enforces `share.acknowledgement.mode=explicit` via a new createShareConsumer override, so the correct Kafka protocol mode is always active regardless of factory config.

`share.acknowledgement.mode=implicit` in factory config was non-functional in 4.0. It's now detected, warned about, and overridden. Users who want true implicit mode should set `ShareAckMode.IMPLICIT` on ContainerProperties.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -14,6 +14,63 @@ The `@KafkaListener` annotation now supports an `ackMode` attribute, allowing in
 The attribute also supports SpEL expressions and property placeholders.
 See xref:kafka/receiving-messages/listener-annotation.adoc[`@KafkaListener` Annotation] for more information.
 
+[[x41-share-ack-mode]]
+=== Share Consumer Acknowledgment Modes
+
+The boolean `setExplicitShareAcknowledgment(boolean)` property on `ContainerProperties` has been replaced by the `ShareAckMode` enum, which clearly names the three distinct acknowledgment modes:
```
artembilan: That is too much info for `whats-new.adoc`.
Why change pattern and provide all of that here, but not in the target chapter?

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -117,6 +116,40 @@ public enum AckMode {
 
 	}
 
+	/**
+	 * The acknowledgment mode for share consumer containers.
+	 * @since 4.1
+	 * @see #setShareAckMode(ShareAckMode)
+	 */
+	public enum ShareAckMode {
```
artembilan, line 124: Feels like a decision to reuse `ContainerProperties` for `ShareConsumer` was wrong direction.
Now we give end-user confusion with existing `AckMode`, which apparently is not used `ShareKafkaMessageListenerContainer`.
And there are probably many other conflicting options not used here or there.
Not saying that it should be revised now, but something to keep in mind if we'd like to still keep our API end-user friendly.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -69,19 +70,24 @@
...
  * <strong>Acknowledgment Modes:</strong>
  * <ul>
- * <li><strong>Implicit</strong>: Records are automatically acknowledged as ACCEPT
- * after successful processing or REJECT on errors</li>
- * <li><strong>Explicit</strong>: Application must manually acknowledge each record;
- * subsequent polls are blocked until all records from the previous poll are acknowledged</li>
+ * <li><strong>EXPLICIT</strong> (default): Container-managed. The container sends ACCEPT after
+ * successful processing and delegates error handling to the {@link ShareConsumerRecordRecoverer}
+ * (default: REJECT). Equivalent to disabling {@code auto.commit} on a regular consumer.</li>
+ * <li><strong>MANUAL</strong>: Listener-managed. The listener must acknowledge each record via
```
artembilan, line 87: I don't say that there is something wrong with the logic described here, but I wonder if that has crossed your mind that technically there is no difference between `EXPLICIT` and `MANUAL`.
Let's imaging the environment if we change the logic only for two `EXPLICIT` and `IMPLICIT`!
The `EXPLICIT` is `MANUAL` at the same time.
So, we always provide ack callback to end-user listener in `EXPLICIT` mode.
And if that callback was not called from end-user code, the container does that, respectively, for both success and error cases.

What do I miss, please?
What is the real driver for the `MANUAL` if it still per record?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -283,40 +288,62 @@ private class ShareListenerConsumer implements Runnable {
...
+							.getConfigurationProperties()
+							.get(ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG);
+					if (configured != null && "implicit".equalsIgnoreCase(configured.toString())) {
+						this.logger.warn("Factory configuration has share.acknowledgement.mode=implicit "
+								+ "but ShareAckMode." + shareAckMode + " is active; the container will "
+								+ "override it with explicit mode. To use implicit mode, set "
+								+ "ShareAckMode.IMPLICIT on the container properties instead.");
+					}
+				}
+				catch (UnsupportedOperationException ex) {
```
artembilan: Why do we need to catch it here?
If that is not implemented in the provided `ShareConsumerFactory`, then better to fail for that reason.
This adds extra unnecessary noise into our code and even make it worse: this catch-n-log hides the problem with invoked method.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
sobychacko: Ya, we have time to refine it before RC1.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
sobychacko: Hey Artem, thanks for the thoughtful feedback on this particular issue. I hear your arguments there, but I'd like to make the case for keeping both modes. The distinction is actually the same one we already draw for regular consumers — `AckMode.BATCH` (container owns everything, listener is unaware) vs `AckMode.MANUAL` (listener owns it, must call `acknowledge()`). There's no auto-fallback in regular consumer `MANUAL` mode either, and I think that strictness is intentional and valuable.
  
For share consumers:
  • EXPLICIT--  listener is completely unaware of acknowledgment. The acknowledgment parameter is null, no API in scope, no decision to make. Container handles success and error outcomes.
  • MANUAL--  listener owns every outcome. Non-null acknowledgment, and if you don't call it, the next `poll()` throws — the Kafka client enforces the contract, not just the docs.

The auto-fallback model silently auto-accepts a record when a listener forgets to call `acknowledge()`. In the strict `MANUAL` model that's a fast-fail -- the blocked poll makes the bug immediately obvious. Collapsing the two for share consumers while keeping them separate for regular consumers would be an inconsistency in the framework, no?

I think keeping both gives users a clear, unambiguous contract that aligns with how Spring Kafka already thinks about acknowledgment.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
sobychacko: good idea. will fix.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
sobychacko: Ok, will move to the chapter.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
sobychacko: The problem is that we need to distinguish two kinds of containers if we introduce a properties class just for share consumers.

sobychacko: Artem - as we discussed yesterday, we will try to tackle this as a follow up item as this involves a lot of moving parts.

spring-kafka/src/main/java/org/springframework/kafka/core/ShareConsumerFactory.java
```diff
@@ -43,6 +43,25 @@ public interface ShareConsumerFactory<K, V> {
...
+	 * will be ignored.
+	 * @param groupId the group id (maybe null).
+	 * @param clientId the client id (maybe null).
+	 * @param overrideProperties properties to apply on top of the factory configuration.
+	 * @return the share consumer.
+	 * @since 4.1
+	 */
+	default ShareConsumer<K, V> createShareConsumer(@Nullable String groupId, @Nullable String clientId,
+			Map<String, Object> overrideProperties) {
+		return createShareConsumer(groupId, clientId);
```
artembilan, line 63: Just minor and no need to change anything now.

For better clarity of the code, I prefer to add blank line before method body when method header is multi-line.
Otherwise the body has stuck to the header and it is hard to read the whole code and easy to miss something.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -483,8 +481,116 @@ void shouldUseCustomRecovererReturningReleaseWhenListenerThrows(EmbeddedKafkaBro
...
+		container.setBeanName("explicitRejectTestContainer");
+		container.start();
+
+		try {
+			assertThat(firstDeliveryLatch.await(15, TimeUnit.SECONDS)).isTrue();
+			// REJECT archives the record; it must NOT be redelivered
+			assertThat(redeliveryLatch.await(5, TimeUnit.SECONDS))
+					.as("Record should be archived (REJECT) and not redelivered")
+					.isFalse();
+			assertThat(deliveryCount.get()).isEqualTo(1);
```
artembilan: The last two conditions are suspicious: the redelivery may not happen not only fort the REJECT reason.
It can be accepted in the error handler or some other way.
Or... Record could be delivered to other consumer on the same share topic.
Therefore we are blocked here for 5 seconds without any clarity.
If there is no way to track rejection somehow on Kafka side, then I don't see any value in this test at all.
Feel like we are trying to test here Kafka client itself: nothing is under our container control.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -483,8 +481,116 @@ void shouldUseCustomRecovererReturningReleaseWhenListenerThrows(EmbeddedKafkaBro
...
+
+		ShareKafkaMessageListenerContainer<String, String> container =
+				new ShareKafkaMessageListenerContainer<>(factory, containerProps);
+		container.setBeanName("implicitErrorTestContainer");
+		container.start();
+
+		try {
+			assertThat(bothSeenLatch.await(15, TimeUnit.SECONDS)).isTrue();
+			// No redelivery: if a third invocation happened the latch would have reached 0
+			assertThat(redeliveryLatch.await(5, TimeUnit.SECONDS))
```
artembilan: I don't think using timing barriers to ensure that something has not happened is the right way to model test behavior.
The event might not happened for may reasons, or our barrier is not long enough to catch that event.

It looks again that you are trying to prove the Kafka client (or even broker) promises with unreliable tools.
I don't know Kafka internal details to give a good advice how to verify that record has been rejected because we had thrown error on it.
But again probably out of our project scope...

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -395,206 +396,289 @@ private void configureShareGroup(String bootstrapServers, String groupId) throws
...
 Share consumers support four acknowledgment types:
 
- ACCEPT: Record processed successfully, mark as completed
- RELEASE: Temporary failure, make record available for redelivery
- REJECT: Permanent failure, do not retry
- RENEW: Extend the acquisition lock (non-terminal; a terminal ack is still required when processing completes)
+* `ACCEPT` — Record processed successfully; mark as completed.
+* `RELEASE` — Temporary failure; make the record available for redelivery to this or another consumer.
+* `REJECT` — Permanent failure; archive the record and do not redeliver.
+* `RENEW` — Extend the acquisition lock (non-terminal).
```
artembilan, line 566: Not related to the change in this PR (which is great), but raises the question as natural evolution afterwards.
How to deal with `RENEW` if we chose `EXPLICIT`?
Is there a possibility that we have to extend the lock somehow from the container perspective?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -195,18 +201,23 @@ protected void doStart() {
 		GenericMessageListener<?> listener = (GenericMessageListener<?>) messageListener;
 		Assert.state(listener != null, "'messageListener' cannot be null");
 
-		// Validate listener type for explicit acknowledgment mode
-		if (containerProperties.isExplicitShareAcknowledgment()) {
+		if (containerProperties.getShareAckMode() == ContainerProperties.ShareAckMode.MANUAL) {
```
artembilan: I think we need to use `equals()` and have it flipped instead of `==` operator which does not compare objects, but rather their memory pointers:
```
if (ContainerProperties.ShareAckMode.MANUAL.equals(containerProperties.getShareAckMode())) {
```

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -195,18 +201,23 @@ protected void doStart() {
...
 			boolean isAcknowledgingListener = listener instanceof AcknowledgingShareConsumerAwareMessageListener;
 			Assert.state(isAcknowledgingListener,
-					"Explicit acknowledgment mode requires an AcknowledgingShareConsumerAwareMessageListener. " +
+					"ShareAckMode.MANUAL requires an AcknowledgingShareConsumerAwareMessageListener. " +
 					"Current listener type: " + listener.getClass().getName() + ". " +
-					"Either use implicit acknowledgment mode or provide a listener that can handle acknowledgments.");
+					"Either use ShareAckMode.EXPLICIT or provide a listener that accepts a ShareAcknowledgment parameter.");
+		}
+
+		if (containerProperties.getShareAckMode() == ContainerProperties.ShareAckMode.EXPLICIT
```
artembilan: DITTO

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -283,40 +288,55 @@ private class ShareListenerConsumer implements Runnable {
...
+			ContainerProperties.ShareAckMode shareAckMode = containerProperties.getShareAckMode();
+			this.isManualAckMode = shareAckMode == ContainerProperties.ShareAckMode.MANUAL;
 			this.ackTimeoutMs = containerProperties.getShareAcknowledgmentTimeout().toMillis();
 
-			// Configure consumer properties based on acknowledgment mode
-			if (this.isExplicitMode) {
-				// Apply explicit mode configuration to consumer
-				// Note: This should ideally be done during consumer creation in the factory
-				this.logger.info("Share consumer configured for explicit acknowledgment mode");
+			if (shareAckMode == ContainerProperties.ShareAckMode.IMPLICIT) {
```
artembilan: DITTO

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -283,40 +288,55 @@ private class ShareListenerConsumer implements Runnable {
...
+				ShareConsumerRecordRecoverer recoverer =
+						ShareKafkaMessageListenerContainer.this.getShareConsumerRecordRecoverer();
+				if (recoverer != ShareConsumerRecordRecoverer.REJECTING) {
```
artembilan: Here identity is good, but has to be flipped to avoid NPE reports.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -283,40 +288,55 @@ private class ShareListenerConsumer implements Runnable {
...
+						ShareKafkaMessageListenerContainer.this.getGroupId(),
+						consumerClientId);
+			}
+			else {
+				// EXPLICIT and MANUAL both use Kafka client explicit mode so the container
+				// has full per-record acknowledgment control.
+				Object configured = ShareKafkaMessageListenerContainer.this.shareConsumerFactory
+						.getConfigurationProperties()
+						.get(ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG);
+				if (configured != null && "implicit".equalsIgnoreCase(configured.toString())) {
```
artembilan: Why cannot we use `ShareAcknowledgementMode` instead of hard-coded string?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -283,40 +288,55 @@ private class ShareListenerConsumer implements Runnable {
...
+					this.logger.warn("Factory configuration has share.acknowledgement.mode=implicit "
+							+ "but ShareAckMode." + shareAckMode + " is active; the container will "
+							+ "override it with explicit mode. To use implicit mode, set "
+							+ "ShareAckMode.IMPLICIT on the container properties instead.");
+				}
+				this.consumer = ShareKafkaMessageListenerContainer.this.shareConsumerFactory.createShareConsumer(
+						ShareKafkaMessageListenerContainer.this.getGroupId(),
+						consumerClientId,
+						Map.of(ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG, "explicit"));
```
artembilan: DITTO.
See `ConsumerConfig` defaults:
```
                .define(ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG,
                                        Type.STRING,
                                        ShareAcknowledgementMode.IMPLICIT.name(),
                                        new ShareAcknowledgementMode.Validator(),
                                        Importance.MEDIUM,
                                        ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_DOC)
```

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -283,40 +288,55 @@ private class ShareListenerConsumer implements Runnable {
...
+						.getConfigurationProperties()
+						.get(ConsumerConfig.SHARE_ACKNOWLEDGEMENT_MODE_CONFIG);
+				if (configured != null && "implicit".equalsIgnoreCase(configured.toString())) {
+					this.logger.warn("Factory configuration has share.acknowledgement.mode=implicit "
+							+ "but ShareAckMode." + shareAckMode + " is active; the container will "
+							+ "override it with explicit mode. To use implicit mode, set "
+							+ "ShareAckMode.IMPLICIT on the container properties instead.");
+				}
+				this.consumer = ShareKafkaMessageListenerContainer.this.shareConsumerFactory.createShareConsumer(
+						ShareKafkaMessageListenerContainer.this.getGroupId(),
```
artembilan: I think we use too many `ShareKafkaMessageListenerContainer.this.getGroupId()` in this method.
Cannot it be extracted to a local variable?

spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
sobychacko: hmm - thats a very interesting scenario that I didn't think about. Before this PR, we still had this problem with the old non-"explict" mode where we were not handing an Acknowledgment object to the listener. There we were not allowing listeners to do any renew, which was a true gap and that gap is still present with the new `EXPLICIT` mode. I guess, we have couple of options - 1. expose a renew-only acknowledgment in explicit mode for listeners to use 2. or document strictly that if they have a renew use case, then they have to use manual mode. We can also auto-renew periodically from the container based on some property maybe. I think we can probably tackle this as a follow-up item maybe?

sobychacko: Issue created to track: https://github.com/spring-projects/spring-kafka/issues/4381

artembilan:
Now that this is merged, @sobychacko , please, take a look if PR https://github.com/spring-projects/spring-kafka/pull/4374 is not a duplication someway, and if this merged PR is fixing respective issue: https://github.com/spring-projects/spring-kafka/issues/4369

Thanks

sobychacko:
Thanks, @artembilan. Will do.

Issue #4324, opened 2026-03-05
Add async commit option to share consumer container


sobychacko:
The container calls `commitSync()` unconditionally in `commitAcknowledgments()`:

```
private void commitAcknowledgments() {
    try {
        this.consumer.commitSync();
    }
    catch (Exception e) {
        this.logger.error(e, "Failed to commit acknowledgments");
    }
}
```

`ShareConsumer` also has `commitAsync()` but it is not used. Add a `ContainerProperties.setSyncShareCommits(boolean)` flag (default `true`) so users can opt into async commits when slight ack-durability lag is acceptable.

maxwell-balla:
@sobychacko 

I'd like to work on this one.

maxwell-balla:
@sobychacko 

Here's my plan : 

1. Add "syncShareCommits" flag (default true) with setter/getter in ContainerProperties
2. Read the flag in "ShareListenerConsumer" constructor
3. Use it in "commitAcknowledgments()" to switch between "commitSync()" and "commitAsync()"

PR #4379, opened 2026-03-31
GH-4324: Add async commit option to share consumer container

maxwell-balla:
Add syncShareCommits flag (default true) to ContainerProperties to allow users to opt into async commits for share consumer acknowledgments.

Previously, ShareKafkaMessageListenerContainer called commitSync() unconditionally in commitAcknowledgments(). 
This change reads the new flag and switches between commitSync() and commitAsync() accordingly.

Changes:
- ContainerProperties: added syncShareCommits field, setter and getter
- ShareKafkaMessageListenerContainer: read the flag in ShareListenerConsumer constructor and use it in commitAcknowledgments()

Fixes gh-4324

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
@@ -1281,6 +1281,7 @@ public class ContainerProperties extends ConsumerProperties {
 		appendProperty(sb, "stopImmediate", this.stopImmediate);
 		appendProperty(sb, "stopContainerWhenFenced", this.stopContainerWhenFenced);
 		appendProperty(sb, "asyncAcks", this.asyncAcks);
+		appendProperty(sb, "syncShareCommits", this.syncShareCommits);
 
 		// Polling and partition configuration
 		appendProperty(sb, "pollTimeoutWhilePaused", this.pollTimeoutWhilePaused);
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
@@ -95,6 +95,7 @@ import org.springframework.util.Assert;
  * @param <V> the value type
  *
  * @author Soby Chacko
+ * @author Maxwell Balla
  *
  * @since 4.0
  *
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
@@ -1312,11 +1312,12 @@ class ShareKafkaMessageListenerContainerIntegrationTests {
 		restartContainer.setBeanName("asyncCommitRestartContainer");
 		restartContainer.start();
 
-		// Allow enough time for any redelivery to occur
-		Thread.sleep(3000);
+		// Verify no redelivery occurs over a sustained period
+		Awaitility.await()
+				.during(3, TimeUnit.SECONDS)
+				.atMost(5, TimeUnit.SECONDS)
+				.untilAsserted(() -> assertThat(redeliveryCount.get()).isZero());
 
 		restartContainer.stop();
-
-		assertThat(redeliveryCount.get()).isZero();
 	}
 }
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
@@ -142,6 +142,7 @@ Share containers support a subset of the container properties available for regu
 * `clientId`: The client ID for the consumer
 * `shareAckMode`: The acknowledgment mode (`EXPLICIT`, `MANUAL`, or `IMPLICIT`; see <<share-record-acknowledgment>>)
 * `shareAcknowledgmentTimeout`: Timeout for detecting unacknowledged records in `MANUAL` mode
+* `syncShareCommits`: Whether to use `commitSync()` (default) or `commitAsync()` for share consumer acknowledgments (see <<share-sync-async-commits>>)
 * `kafkaConsumerProperties`: Additional consumer properties
 
 [IMPORTANT]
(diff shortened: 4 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -1147,6 +1149,26 @@ public boolean isExplicitShareAcknowledgment() {
...
+	 */
+	public void setSyncShareCommits(boolean syncShareCommits) {
+		this.syncShareCommits = syncShareCommits;
+	}
+
+	/**
+	 * Check whether share consumer commits are synchronous.
+	 * @return true if using commitSync(), false if using commitAsync().
+	 */
+	public boolean isSyncShareCommits() {
```
sobychacko, line 1198: Need the `@since` tag.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -502,7 +505,12 @@ private void handleProcessingError(ConsumerRecord<K, V> record,
 
 		private void commitAcknowledgments() {
 			try {
-				this.consumer.commitSync();
+				if (this.syncShareCommits) {
+					this.consumer.commitSync();
+				}
+				else {
+					this.consumer.commitAsync();
```
sobychacko, line 535: I created a related issue that needs to be tackled as a follow-up to this PR: https://github.com/spring-projects/spring-kafka/issues/4380

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -218,4 +219,28 @@ void factoryShouldApplyShareConsumerRecordRecovererToContainer() {
...
+		ContainerProperties containerProperties = new ContainerProperties("test-topic");
+		containerProperties.setSyncShareCommits(false);
+		containerProperties.setMessageListener(messageListener);
+
+		ShareKafkaMessageListenerContainer<String, String> container =
+				new ShareKafkaMessageListenerContainer<>(shareConsumerFactory, containerProperties);
+
+		assertThat(container.getContainerProperties().isSyncShareCommits())
+				.isFalse();
+	}
```
sobychacko, line 352: The two new unit tests (`shouldConfigureSyncShareCommitsByDefault` / `shouldConfigureAsyncShareCommitsWhenSet`) are really just asserting that the getter returns what you set. They don't actually prove the container calls `commitSync()` or `commitAsync()` based on the flag. So right now, you could delete the `if/else` in `commitAcknowledgments()` entirely and both tests
would still go green. 
The existing test `shouldPassExplicitOverrideToConsumerEvenWhenFactoryConfigIsImplicit` shows the pattern to follow — mock a `ShareConsumer`, start the container, produce a record, and
  then `verify(mockConsumer).commitAsync()`. That way the test is actually exercising the behavior rather than just the properties.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -502,7 +505,12 @@ private void handleProcessingError(ConsumerRecord<K, V> record,
...
+				}
 			}
 			catch (Exception e) {
```
sobychacko, line 538: The catch `(Exception e)` won't catch async commit failures since `commitAsync()` returns immediately. That's tracked in #4380 though, so no need to address it here.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -1268,4 +1269,45 @@ void shouldStopAllConsumerThreadsGracefully(EmbeddedKafkaBroker broker) throws E
 		assertThat(container.isRunning()).isFalse();
 	}
 
+	@Test
+	void shouldSupportAsyncCommits(EmbeddedKafkaBroker broker) throws Exception {
```
sobychacko, line 1270: The test passes as long as 5 records are consumed, but that would be true even if the `commitAsync()` call were removed entirely. To actually verify async commits work, the test should confirm durability --consume the records, stop the container, restart it with the same groupId, and assert that the records are not redelivered. That's the only way to know the async commits actually worked as expected.

sobychacko (changes requested):
@maxwell-balla I added some more feedback. Please take a look. Also, make sure that you add the `@author` tag to all the classes you changed. You may want to start thinking about adding docs and whats-new changes. They need to go in with this PR.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -1265,4 +1266,57 @@ void shouldStopAllConsumerThreadsGracefully(EmbeddedKafkaBroker broker) throws E
...
+			redeliveryCount.incrementAndGet();
+		});
+
+		ShareKafkaMessageListenerContainer<String, String> restartContainer =
+				new ShareKafkaMessageListenerContainer<>(consumerFactory, restartProps);
+		restartContainer.setBeanName("asyncCommitRestartContainer");
+		restartContainer.start();
+
+		// Allow enough time for any redelivery to occur
+		Thread.sleep(3000);
```
sobychacko: maybe use Awaitility until redelivery count is zero instead of `Thread.sleep()`?

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
sobychacko, line 97: Need to add your name here as an author.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
sobychacko, line 1248: `toString()` needs to take into account the new property.

sobychacko:
@maxwell-balla The PR has been merged upstream via https://github.com/spring-projects/spring-kafka/commit/12286b1044ebc6bb6c8c23bdfd6e71f9a3d9642b. Many thanks for your contribution! More PR's are welcomed!!

maxwell-balla:
Thanks @sobychacko ! Great experience contributing to the project. 
I just pinged you on #4380.
Ready for the next one.

PR #4382, opened 2026-04-01
Fix immutability StreamsBuilderFactory properties

jad837:
Fix immuable object sharing across `StreamsBuilderFactoryBean`

As discussed https://github.com/spring-projects/spring-kafka/pull/4373#discussion_r3011957331, This PR ensures that `StreamsBuilderFactoryBean`  & `KafkaStreamsConfiguration` always sends a new copy of properties, ensuring immutablility.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
@@ -190,12 +190,12 @@ public class StreamsBuilderFactoryBean extends AbstractFactoryBean<StreamsBuilde
 	 */
 	public void setStreamsConfiguration(Properties streamsConfig) {
 		Assert.notNull(streamsConfig, STREAMS_CONFIG_MUST_NOT_BE_NULL);
-		this.properties = streamsConfig;
+		this.properties = (Properties) streamsConfig.clone();
 	}
 
 	@Nullable
 	public Properties getStreamsConfiguration() {
-		return this.properties; // NOSONAR - inconsistent synchronization
+		return (Properties) this.properties.clone(); // NOSONAR - inconsistent synchronization
 	}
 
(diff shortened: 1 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/config/StreamsBuilderFactoryBean.java
```diff
@@ -190,12 +190,12 @@ public void setDeadLetterTopicName(String deadLetterTopicName) {
...
 	public void setStreamsConfiguration(Properties streamsConfig) {
 		Assert.notNull(streamsConfig, STREAMS_CONFIG_MUST_NOT_BE_NULL);
-		this.properties = streamsConfig;
+		this.properties = (Properties) streamsConfig.clone();
 	}
 
 	@Nullable
 	public Properties getStreamsConfiguration() {
-		return this.properties; // NOSONAR - inconsistent synchronization
+		return (Properties) this.properties.clone(); // NOSONAR - inconsistent synchronization
```
artembilan: Let's remove that `NOSONAR` comment!
Not only saying that we don't use it anymore, but also it won't make sense after cloning anyway.

artembilan (changes requested):
Any comments why this is in the draft?
Thanks

artembilan:
@jad837 ,

Thank you!

pdevalck:
The changes cause NPE when the properties are not initialized.

jad837:
Hey @pdevalck  is this in case when StreamsBuilderFactoryBean.getProperties() is called? Can you share a bit more on how Can I create this example? I believe this could be when StreamsBuilderFactoryBean is created without any constructor argument... Let me know if I am correct, I can check on this..

pdevalck:
Yes, you are right. In our integration tests, the constructor without arguments is being called to instantiate te factory. When this happens and we happen to get the streams properties, the NPE happens

PR #4383, opened 2026-04-01
Add start/stop count metrics to listener container

### LINKED ISSUE #4277: Add Container start, stop count metric
**Expected Behavior**

Provide metrics to track the lifecycle events of Spring Kafka message listener containers:

spring_kafka_container_start_count

spring_kafka_container_stop_count

These metrics should increment each time a listener container is started or stopped.

**Current Behavior**

There are currently no built-in metrics that expose start and stop events for Spring Kafka message listener containers. As a result, users cannot directly monitor how often containers are started or stopped.

**Context**

In some use cases, message listener containers are started and stopped dynamically based on business logic. Without dedicated metrics, it is difficult to observe these lifecycle transitions or detect unintended start/stop behavior.

Introducing spring_kafka_container_start_count and spring_kafka_container_stop_count would provide better operational visibility. This would help users monitor container lifecycle activity, diagnose unexpected behavior, and improve observability in production environments.

Yelagandula:
Adds two Micrometer counters to AbstractMessageListenerContainer to track listener container lifecycle events:

- spring.kafka.container.start.count: incremented each time a container is successfully started via doStart()
- spring.kafka.container.stop.count: incremented each time a container is stopped, covering both wait=true and wait=false paths

Closes #4277

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
@@ -73,6 +73,7 @@ import org.springframework.util.StringUtils;
  * @author Sanghyeok An
  * @author Lokesh Alamuri
  * @author Christian Fredriksson
+ * @author Vineeth Yelagandula
  */
 public abstract class AbstractMessageListenerContainer<K, V>
 		implements GenericMessageListenerContainer<K, V>, BeanNameAware, ApplicationEventPublisherAware,
@@ -530,6 +531,7 @@ public abstract class AbstractMessageListenerContainer<K, V>
 						() -> "A " + GenericMessageListener.class.getName() + " implementation must be provided");
 				Assert.state(!this.fenced, "Container Fenced. It is not allowed to start.");
 				doStart();
+				recordContainerStarted();
 			}
 		}
 		finally {
@@ -609,6 +611,20 @@ public abstract class AbstractMessageListenerContainer<K, V>
 
 	protected abstract void doStart();
 
+	/**
+	 * Called when the container is started. Subclasses may override to record metrics.
+	 * @since 4.1.0
+	 */
+	protected void recordContainerStarted() {
+	}
+
+	/**
+	 * Called when the container is stopped. Subclasses may override to record metrics.
+	 * @since 4.1.0
+	 */
+	protected void recordContainerStopped() {
+	}
+
 	@Override
 	public final void stop() {
 		stop(true);
@@ -680,6 +696,7 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	public void stopAbnormally(Runnable callback) {
 		this.lifecycleLock.lock();
 		try {
+			recordContainerStopped();
 			doStop(callback, false);
 			publishContainerStoppedEvent();
 		}
@@ -689,6 +706,7 @@ public abstract class AbstractMessageListenerContainer<K, V>
 	}
 
 	protected void doStop(Runnable callback) {
+		recordContainerStopped();
 		doStop(callback, true);
 		publishContainerStoppedEvent();
 	}
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -51,6 +51,7 @@ import java.util.function.Function;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.observation.Observation;
 import io.micrometer.observation.ObservationRegistry;
 import org.apache.kafka.clients.admin.AdminClientConfig;
@@ -122,6 +123,7 @@ import org.springframework.kafka.support.TopicPartitionOffset.SeekPosition;
 import org.springframework.kafka.support.micrometer.KafkaListenerObservation;
 import org.springframework.kafka.support.micrometer.KafkaListenerObservation.DefaultKafkaListenerObservationConvention;
 import org.springframework.kafka.support.micrometer.KafkaRecordReceiverContext;
+import org.springframework.kafka.support.micrometer.ContainerLifecycleMicrometerHolder;
 import org.springframework.kafka.support.micrometer.MicrometerHolder;
 import org.springframework.kafka.support.serializer.DeserializationException;
 import org.springframework.kafka.support.serializer.SerializationUtils;
@@ -177,6 +179,7 @@ import org.springframework.util.StringUtils;
  * @author Jinhui Kim
  * @author Minchul Son
  * @author Youngjoo Kim
+ * @author Vineeth Yelagandula
  */
 public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 		extends AbstractMessageListenerContainer<K, V> implements ConsumerPauseResumeEventPublisher {
@@ -194,6 +197,9 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 	private final @Nullable TopicPartitionOffset @Nullable [] topicPartitions;
 
 	private @Nullable String clientIdSuffix;
+
+	@Nullable
+	private ContainerLifecycleMicrometerHolder lifecycleMicrometerHolder;
 
 	private Runnable emergencyStop = () -> stopAbnormally(() -> {
 	});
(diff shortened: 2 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -32,6 +32,9 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Metrics;
```
sobychacko: I think it is better to avoid tight Micrometer coupling in this abstract base class, since it is shared by all container types. Putting Micrometer types here means anything that loads `AbstractMessageListenerContainer` pulls in those classes even when the application does not use metrics. I would rather follow the same pattern the listener consumer and KafkaTemplate use: obtain the Micrometer context only when it is needed (take a look at `obtainMicrometerHolder()`). A cleaner approach might be to introduce two protected lifecycle methods to record start and stop. Subclasses such as `KafkaMessageListenerContainer` can override them, and a dedicated class similar to `MicrometerHolder` can handle the core Micrometer work. That will take a bit more design work. Let us know what you think about this idea.

sobychacko (changes requested):

Yelagandula:
Thank you for the detailed feedback @sobychacko! I've refactored the 
implementation following your suggestion:

- Added protected recordContainerStarted() and recordContainerStopped() 
  lifecycle hooks in AbstractMessageListenerContainer (no-op by default)
- Created ContainerLifecycleMicrometerHolder to handle counter 
  registration and increment logic
- Overridden the lifecycle methods in KafkaMessageListenerContainer 
  following the obtainMicrometerHolder() pattern to avoid tight 
  Micrometer coupling in the abstract base class

Please let me know if this looks good or if anything needs adjustment.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/ContainerLifecycleMicrometerHolder.java
```diff
@@ -0,0 +1,59 @@
...
+package org.springframework.kafka.support.micrometer;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+
+/**
+ * Micrometer holder for container lifecycle metrics (start/stop counts).
+ *
+ * @author Vineeth Yelagandula
+ * @since 3.4
```
sobychacko, line 26: This needs to be `4.1.0`.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/ContainerLifecycleMicrometerHolder.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public class ContainerLifecycleMicrometerHolder {
```
sobychacko, line 28: This class needs to be final.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -365,6 +369,43 @@ private void consumerWakeIfNecessary() {
...
+	protected void recordContainerStopped() {
+		if (this.lifecycleMicrometerHolder != null) {
+			this.lifecycleMicrometerHolder.recordStop();
+		}
+	}
+
+	@Nullable
+	private ContainerLifecycleMicrometerHolder obtainLifecycleMicrometerHolder() {
+		try {
+			if (KafkaUtils.MICROMETER_PRESENT) {
```
sobychacko, line 392: We also should add the `this.containerProperties.isMicrometerEnabled()` check (similar to the other obtain method for the listener). That way, if the user explicitly disabled micrometer via the property, then this method will (correctly so) return null.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/ContainerLifecycleMicrometerHolder.java
```diff
@@ -0,0 +1,59 @@
...
+
+	private final Counter startCounter;
+
+	private final Counter stopCounter;
+
+	public ContainerLifecycleMicrometerHolder(MeterRegistry registry, String containerName) {
```
sobychacko, line 34: Need javadoc for the constructor.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -689,6 +702,7 @@ public void stopAbnormally(Runnable callback) {
 	}
 
 	protected void doStop(Runnable callback) {
+		recordContainerStopped();
```
sobychacko, line 705: This call should also go into the `stopAbnormally` method above.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -609,6 +610,18 @@ public void checkGroupId() {
 
 	protected abstract void doStart();
 
+	/**
+	 * Called when the container is started. Subclasses may override to record metrics.
+	 */
+	protected void recordContainerStarted() {
```
sobychacko, line 616: Need `since` tag.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractMessageListenerContainer.java
```diff
@@ -609,6 +610,18 @@ public void checkGroupId() {
...
+	}
+
+	/**
+	 * Called when the container is stopped. Subclasses may override to record metrics.
+	 */
+	protected void recordContainerStopped() {
```
sobychacko, line 622: Need `since` tag.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -365,6 +369,43 @@ private void consumerWakeIfNecessary() {
...
+				org.springframework.context.ApplicationContext ctx = getApplicationContext();
```
sobychacko, line 393: Why are we using FQCN here? Can we import it normally?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -365,6 +369,43 @@ private void consumerWakeIfNecessary() {
...
+				if (ctx != null) {
+					io.micrometer.core.instrument.MeterRegistry registry =
+							ctx.getBeanProvider(io.micrometer.core.instrument.MeterRegistry.class).getIfUnique();
```
sobychacko, line 396: Why are we using FQCN here? Can we import it normally?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -365,6 +369,43 @@ private void consumerWakeIfNecessary() {
...
+					if (registry != null) {
+						return new ContainerLifecycleMicrometerHolder(registry, getBeanName());
+					}
+				}
+			}
+		}
+		catch (Exception ex) {
```
sobychacko, line 403: Maybe exception could be more specific? Also, add a logging statement?

sobychacko (changes requested):
Latest changes in the PR are in the right direction. Thanks for the updates. I added some review comments. Please take a look. 

In addition, please address these concerns also.

1. Add tests to verify the expected behavior.
2. Reference docs - since these are new metrics, we should add reference docs with detailed information.
3. Then add an entry in `whats-new` doc and link it to the ref docs section.
4. You need to add your name as an author to all the classes you added or modified.
5. Need to sign your commit using DCO - https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring

Thanks.

sobychacko:
@Yelagandula Could you take a look at these review comments? https://github.com/spring-projects/spring-kafka/pull/4383#pullrequestreview-4134802049

Yelagandula:
Hi @sobychacko, addressed all the review comments:

- Fixed @since to 4.1.0
- Made ContainerLifecycleMicrometerHolder final
- Added javadoc for the constructor
- Added isMicrometerEnabled() check in obtainLifecycleMicrometerHolder()
- Replaced FQCN with proper imports for ApplicationContext and MeterRegistry
- Changed catch to IllegalStateException with a debug log statement
- Added recordContainerStopped() call in stopAbnormally() as well
- Added @since 4.1.0 to both recordContainerStarted() and recordContainerStopped()

Still working on tests, reference docs, whats-new entry, and author 
tags will push those shortly.

Yelagandula:
Hi @sobychacko, addressed all remaining review comments:

- Added tests verifying start/stop counter behavior
- Added reference docs in micrometer.adoc under "Monitoring Listener Container Lifecycle"
- Added whats-new entry linking to the ref docs section
- Added @author Vineeth Yelagandula to all modified classes

Please take a look when you get a chance!

Yelagandula:
Also squashed all commits into one clean commit with DCO sign-off. 
Should be good now!

sobychacko:
@Yelagandula Build failed on the CI. Can you take a look? Also, there are some test failures. You need to use a proper email address when signing the commit, the DCO check failed on the PR because of that.

sobychacko:
@Yelagandula Any updates on this PR? Just a reminder that we have a release coming up in a few days and if you want to include these changes as part of the `4.1.0` release, the review concerns I raised above must be addressed before then. Otherwise, it has to wait until the next release cycle. Thanks!

sobychacko:
@Yelagandula Any updates on this?

sobychacko:
@Yelagandula Merged the PR via this commit: https://github.com/spring-projects/spring-kafka/commit/f1f69f46aaef30a53ce25cc2bf336bc6bdb5d07b. Updated the PR to reflect the `4.2.0` release line, among other cleanups. Your author name is preserved in the classes modified. Thanks for the PR!

Issue #4327, opened 2026-03-05
Add stop and failure lifecycle events to share consumer container


sobychacko:
The share consumer container publishes `ConsumerStartingEvent` and `ConsumerStartedEvent` but no stop or failure events.

`wrapUp()` only calls `consumer.close()` — no event publishing.

`ConsumerStoppedEvent` and `ConsumerFailedToStartEvent` take generic `Object` parameters and can be reused directly. `ConsumerStoppingEvent` takes `Consumer<?, ?>` and `Collection<TopicPartition>`, which won't work — `ShareConsumer` does not extend `Consumer` and share consumers have no partition assignments. This needs either a new `ShareConsumerStoppingEvent` or a refactor of the existing event.

zooo-code:
Hi @sobychacko !
I'd like to work on this issue if it's open for contribution.

From looking at the code, I understand:
- ConsumerStoppedEvent and ConsumerFailedToStartEvent take generic Object parameters, so they can be reused directly
- ConsumerStoppingEvent takes Consumer<?, ?> and Collection<TopicPartition>, which won't work since ShareConsumer doesn't extend Consumer and has no partition assignments

Would you prefer creating a new ShareConsumerStoppingEvent, or refactoring the existing ConsumerStoppingEvent to be more generic?

Thanks!

sobychacko:
@zooo-code Please go ahead. I like the idea of a separate component for share consumers. Thanks!

PR #4385, opened 2026-04-03
GH-4327: Add stop and failure lifecycle events to share consumer container

zooo-code:
Resolves GH-4327(#4327)

## Summary

Add stop and failure lifecycle events to ShareKafkaMessageListenerContainer.               
Previously, the share consumer container only published ConsumerStartingEvent and ConsumerStartedEvent.
This change adds ShareConsumerStoppingEvent, ConsumerStoppedEvent, and ConsumerFailedToStartEvent to provide full lifecycle event coverage.

## Changes
- Introduce ShareConsumerStoppingEvent (new class, since ShareConsumer does not extend Consumer<?, ?>)                                                                       
- Publish ShareConsumerStoppingEvent before share consumer close
- Publish ConsumerStoppedEvent after share consumer close with reason (NORMAL, ABNORMAL, ERROR)
- Publish ConsumerFailedToStartEvent when share consumer fails to start
- Add unit tests for lifecycle event publishing scenarios

Please let me know if any changes are needed.
I'm happy to address any feedback promptly!

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
@@ -233,12 +233,7 @@ public class ShareKafkaMessageListenerContainer<K, V>
 		catch (Exception e) {
 			this.logger.error(e, "Failed to start share consumer");
 			for (ShareListenerConsumer built : builtConsumers) {
-				try {
-					built.getConsumer().close();
-				}
-				catch (Exception closeEx) {
-					this.logger.error(closeEx, "Failed to close share consumer after startup failure");
-				}
+				built.closeAfterFailedStartup();
 			}
 			publishConsumerFailedToStartEvent();
 			throw e;
@@ -406,8 +401,17 @@ public class ShareKafkaMessageListenerContainer<K, V>
 			this.consumer.subscribe(Arrays.asList(containerProperties.getTopics()));
 		}
 
-		ShareConsumer<K, V> getConsumer() {
-			return this.consumer;
+		/**
+		 * Quietly close the underlying {@link ShareConsumer} after a failed startup.
+		 * No lifecycle events are published because the consumer never reached {@code run()}.
+		 */
+		private void closeAfterFailedStartup() {
+			try {
+				this.consumer.close();
+			}
+			catch (Exception closeEx) {
+				this.logger.error(closeEx, "Failed to close share consumer after startup failure");
+			}
 		}
 
 		@Nullable
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
@@ -154,6 +154,24 @@ Share consumers do not support:
 * Manual offset management
 ====
 
+[[share-container-lifecycle-events]]
+=== Lifecycle Events
+
+The `ShareKafkaMessageListenerContainer` publishes Spring application events across the consumer lifecycle:
+
+* `ConsumerStartingEvent`: published when a share consumer thread is first started, before it starts polling.
+* `ConsumerStartedEvent`: published when a share consumer is about to start polling.
+* `ConsumerFailedToStartEvent`: published when a share consumer fails to start (for example, the underlying `ShareConsumer` could not be created).
+Any already-constructed `ShareConsumer` instances from the same startup attempt are closed quietly and do not publish stopping or stopped events because they never reached the poll loop.
+* `ShareConsumerStoppingEvent`: published by each share consumer just before its underlying `ShareConsumer` is closed.
+Unlike `ConsumerStoppingEvent`, this event carries a `ShareConsumer` reference (not a `Consumer`) because `ShareConsumer` is a separate client API and is not assigned to specific partitions.
+* `ConsumerStoppedEvent`: published after the share consumer is closed, with a `reason`:
+** `NORMAL` - the container was stopped.
+** `ABNORMAL` - the consumer thread exited because of an `Exception`.
+** `ERROR` - a `java.lang.Error` was thrown.
+
+These events can be consumed with `@EventListener` (see xref:kafka/events.adoc#event-consumption[Event Consumption]) to integrate share containers with the same monitoring and restart patterns available for regular containers.
+
 [[share-container-concurrency]]
 === Concurrency
 
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
@@ -40,6 +40,19 @@ The default is `ShareConsumerRecordRecoverer.REJECTING`; `RELEASING` is also ava
 You can set a custom recoverer on the factory or container.
 * See xref:kafka/kafka-queues.adoc#share-error-handling[Share consumer error handling] in the Kafka Queues documentation.
 
+[[x41-share-consumer-lifecycle-events]]
+=== Share Consumer Lifecycle Events
+
+`ShareKafkaMessageListenerContainer` now publishes stop and failure lifecycle events in addition to `ConsumerStartingEvent` and `ConsumerStartedEvent`:
+
+* `ShareConsumerStoppingEvent` (new): published before the underlying `ShareConsumer` is closed.
+A new event class is introduced because `ShareConsumer` is a separate client API from `Consumer`, so the existing `ConsumerStoppingEvent` is not reusable.
+* `ConsumerStoppedEvent`: published after the share consumer is closed, with a `reason` of `NORMAL`, `ABNORMAL`, or `ERROR`.
+* `ConsumerFailedToStartEvent`: published when a share consumer fails to start.
+Any already-constructed consumers from the same startup attempt are closed quietly.
+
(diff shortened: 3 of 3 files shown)


spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -351,4 +361,172 @@ void shouldUseCommitAsyncWhenSyncShareCommitsIsFalse() throws Exception {
...
+				stoppingConsumer.set(stoppingEvent.getConsumer());
+			}
+			else if (event instanceof ConsumerStoppedEvent stoppedEvent) {
+				stoppedReason.set(stoppedEvent.getReason());
+				stoppedLatch.countDown();
+			}
+		});
+
+		container.start();
+		Thread.sleep(500);
```
sobychacko: maybe use a combination of `CountdownLatch` and `Awaitility` utililty instead of `Thread.sleep`?

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -351,4 +361,172 @@ void shouldUseCommitAsyncWhenSyncShareCommitsIsFalse() throws Exception {
...
+		assertThat(stoppedReason.get()).isEqualTo(Reason.NORMAL);
+		assertThat(stoppingConsumer.get()).isSameAs(consumer);
+	}
+
+	@SuppressWarnings("unchecked")
+	@Test
+	void shouldPublishLifecycleEventsInOrder() throws InterruptedException {
+		ShareConsumer<String, String> consumer = mock(ShareConsumer.class);
+		willAnswer(invocation -> {
+			Thread.sleep(100);
```
sobychacko: Try to avoid `Thread.sleep`. See my suggestion above.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
zooo-code: Thank you for the review! I've replaced Thread.sleep with CountDownLatch — the test now waits for ConsumerStartedEvent before proceeding to stop, and uses a pollLatch to control the poll loop timing.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
zooo-code: Fixed as well. 
The Thread.sleep inside poll() has been replaced with pollLatch.await(), which blocks until the test is ready to proceed.

Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -220,12 +225,20 @@ protected void doStart() {
...
+		try {
+			for (int i = 0; i < this.concurrency; i++) {
+				String consumerClientId = determineClientId(i);
+				ShareListenerConsumer consumer = new ShareListenerConsumer(listener, consumerClientId);
+				this.consumers.add(consumer);
+				CompletableFuture<Void> future = CompletableFuture.runAsync(consumer, consumerExecutor);
+				this.consumerFutures.add(future);
+			}
+		}
+		catch (Exception e) {
```
sobychacko, line 233: We need to move `setRunning(true)` to after all consumers are built. With `concurrency > 1`, you can get a case where a later consumer fails to start while an earlier one is already running under `runAsync`, and we still treat the whole thing as “failed to start", which is confusing. So I’d split startup in two: first build every `ShareListenerConsumer` in a `try`/`catch` and move your current failure handling into that `catch`. `setRunning(true)` and `CompletableFuture.runAsync` should happen only after that build step completes without an exception. Then call `runAsync` on each of the prepared consumers (the full set, since we only get there if all of them were constructed).

Once you move the `catch`, then we should close any `ShareConsumer` instances that were constructed before the failure. They never reached `run()`, so no stopping/stopped events, just a quiet `close()` so we don’t leave clients open.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -351,4 +361,185 @@ void shouldUseCommitAsyncWhenSyncShareCommitsIsFalse() throws Exception {
...
+			}
+			else if (event instanceof ConsumerStoppedEvent stoppedEvent) {
+				stoppedReason.set(stoppedEvent.getReason());
+				stoppedLatch.countDown();
+			}
+		});
+
+		container.start();
+		assertThat(pollEnteredLatch.await(10, TimeUnit.SECONDS)).isTrue();
+		container.stop();
```
sobychacko: Small nit here:  the mock `poll()` above blocks on `pollReleaseLatch.await(10, …)` on the consumer thread, while `stop()` here is on the test thread that joins until that consumer task finishes. If `countDown()` runs after `stop()` as it is now , the test thread can’t reach `countDown()` until `join()` returns. So you get a self-lock until the latch times out (10s), which makes these tests slow and can bite CI on timeouts. I think if we call `pollReleaseLatch.countDown()` before `container.stop()`, then `poll()` returns right away and shutdown is clean. What do you think?

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -351,4 +361,185 @@ void shouldUseCommitAsyncWhenSyncShareCommitsIsFalse() throws Exception {
...
+			}
+			else if (event instanceof ConsumerStoppedEvent) {
+				eventOrder.add("stopped");
+				stoppedLatch.countDown();
+			}
+		});
+
+		container.start();
+		assertThat(pollEnteredLatch.await(10, TimeUnit.SECONDS)).isTrue();
+		container.stop();
```
sobychacko: Same concern as above.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
zooo-code: Thank you for the detailed feedback! I've split doStart() into two phases:
1. Build phase: construct all ShareListenerConsumer instances in a try/catch
2. Run phase: setRunning(true) and runAsync only after all consumers are built successfully

On failure, any already-constructed consumers are quietly closed before publishing `ConsumerFailedToStartEvent`

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
zooo-code: Moved pollReleaseLatch.countDown() before container.stop() to avoid the self-lock on join().
Thanks!

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
zooo-code: Fixed in the same change.
Thanks!

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -218,11 +223,30 @@ protected void doStart() {
...
+			this.logger.error(e, "Failed to start share consumer");
+			for (ShareListenerConsumer built : builtConsumers) {
+				try {
+					built.getConsumer().close();
+				}
+				catch (Exception closeEx) {
+					this.logger.error(closeEx, "Failed to close share consumer after startup failure");
+				}
```
sobychacko: What do you think about moving this entire `try/catch` block as a private method on `ShareListenerConsumer`? That way, we can avoid having that getter method (`getConsumer()`) whose sole job is to provide access to the consumer only for closing. If we provide a method like `closeAfterFailedStartup()` (or something similar) in the consumer itself, then we don't need to expose a getter like that and close it all within the consumer boundaries. Not a big deal though, but worth taking a look.

zooo-code: Moved the close logic into a private `closeAfterFailedStartup()` method on `ShareListenerConsumer`, so the `getConsumer()` accessor is no longer needed and the consumer stays fully encapsulated within the ShareListenerConsumer boundary.

zooo-code:
> We are getting close, I have a small comment added. Please take a look. Also, we need to update reference docs and whats-new sections as part of this PR. Thanks!

Thanks for the review!
Also added a new "Lifecycle Events" section under Share Message Listener Containers in the reference docs and a corresponding entry in whats-new.

sobychacko:
@zooo-code Thanks for the PR. It has been merged upstream via https://github.com/spring-projects/spring-kafka/commit/cecc2f9a85d39ad2d0ab5ae91d04103fb9befd26.

zooo-code:
Thanks for merging this, @sobychacko ! 
I really appreciate the detailed reviews across each iteration.
Thanks!

PR #4390, opened 2026-04-07
GH-4380: Add AcknowledgementCommitCallback support for async share co…

### LINKED ISSUE #4380: Add AcknowledgementCommitCallback support for async share consumer commits
Following up on #4324 which adds the `syncShareCommits` option for share consumer containers, we should consider adding support for `AcknowledgementCommitCallback` when async commits are used.

Currently when `syncShareCommits=false`, the container calls `commitAsync()` (PR #4379) but there's no visibility into whether those commits succeed or fail. The Kafka `ShareConsumer` API provides `setAcknowledgementCommitCallback()` for this purpose.

  Proposal:

  • When async commits are enabled, register a default callback that logs failures
  • Optionally expose a way for users to provide their own callback via `ContainerProperties` for custom error handling

  This would give users better observability when opting into async commits for higher throughput.

  Related: #4324, PR #4379

maxwell-balla:
Fixes #4380

Following up on #4379 (async commit option), this PR adds
AcknowledgementCommitCallback support to the share consumer container.

When `syncShareCommits=false`, a default callback is registered that logs
commit failures. Users can provide a custom callback via
`ContainerProperties.setAcknowledgementCommitCallback()` for custom error
handling.

Changes:
- `ContainerProperties`: added `acknowledgementCommitCallback` field with setter/getter
- `ShareKafkaMessageListenerContainer`: register the callback on the `ShareConsumer`
  before `subscribe()`, with a default logging callback for async mode

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
@@ -28,6 +28,7 @@ import java.util.regex.Pattern;
 
 import io.micrometer.observation.ObservationRegistry;
 import org.aopalliance.aop.Advice;
+import org.apache.kafka.clients.consumer.AcknowledgementCommitCallback;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.jspecify.annotations.Nullable;
 
@@ -351,6 +352,8 @@ public class ContainerProperties extends ConsumerProperties {
 	private ShareAckMode shareAckMode = ShareAckMode.EXPLICIT; // default: container-managed explicit mode
 
 	private boolean syncShareCommits = true;
+
+	private @Nullable AcknowledgementCommitCallback acknowledgementCommitCallback;
 
 	private Duration shareAcknowledgmentTimeout = Duration.ofSeconds(30); // Align with Kafka's share.record.lock.duration.ms default
 
@@ -1197,6 +1200,28 @@ public class ContainerProperties extends ConsumerProperties {
 	 */
 	public boolean isSyncShareCommits() {
 		return this.syncShareCommits;
+	}
+
+	/**
+	 * Set the callback to be invoked when acknowledgement commits complete.
+	 * <p>This is particularly useful when {@link #setSyncShareCommits(boolean) syncShareCommits}
+	 * is {@code false} (async mode) to get visibility into commit success or failure.
+	 * <p>If not set and async commits are enabled, a default callback that logs failures will be used.
+	 * @param acknowledgementCommitCallback the callback, or null to use the default.
+	 * @since 4.1
+	 * @see AcknowledgementCommitCallback
+	 */
+	public void setAcknowledgementCommitCallback(AcknowledgementCommitCallback acknowledgementCommitCallback) {
+		this.acknowledgementCommitCallback = acknowledgementCommitCallback;
+	}
+
+	/**
+	 * Get the acknowledgement commit callback.
+	 * @return the callback, or null if not set.
+	 * @since 4.1
+	 */
+	public @Nullable AcknowledgementCommitCallback getAcknowledgementCommitCallback() {
+		return this.acknowledgementCommitCallback;
 	}
 
 	/**
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
@@ -79,7 +79,8 @@ import static org.mockito.Mockito.verify;
 				"share-container-lifecycle-test",
 				"share-container-recoverer-release-test",
 				"share-container-deser-error-test",
-				"share-container-async-commit-test"
+				"share-container-async-commit-test",
+				"share-container-async-callback-test"
 		},
 		partitions = 1,
 		brokerProperties = {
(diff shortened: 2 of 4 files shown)


artembilan (changes requested):
@maxwell-balla ,

> Change order of import in ShareKafkaMessageListenerContainer… 

please, consider to run `./gradlew check` and fix all the failures locally before pushing to the remote branch.
All of those Checkstyle violations feels like your PR is not ready for review.

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -1199,6 +1202,29 @@ public boolean isSyncShareCommits() {
...
+	/**
+	 * Set the callback to be invoked when acknowledgement commits complete.
+	 * <p>This is particularly useful when {@link #setSyncShareCommits(boolean) syncShareCommits}
+	 * is {@code false} (async mode) to get visibility into commit success or failure.
+	 * <p>If not set and async commits are enabled, a default callback that logs failures will be used.
+	 * @param acknowledgementCommitCallback the callback, or null to use the default.
+	 * @since 4.1
+	 * @see AcknowledgementCommitCallback
+	 */
+	public void setAcknowledgementCommitCallback(@Nullable AcknowledgementCommitCallback acknowledgementCommitCallback) {
```
artembilan: Why do we need `@Nullable` in the setter?

spring-kafka/src/main/java/org/springframework/kafka/listener/ContainerProperties.java
```diff
@@ -1199,6 +1202,29 @@ public boolean isSyncShareCommits() {
...
+		this.acknowledgementCommitCallback = acknowledgementCommitCallback;
+	}
+
+	/**
+	 * Get the acknowledgement commit callback.
+	 * @return the callback, or null if not set.
+	 * @since 4.1
+	 */
+	@Nullable
+	public AcknowledgementCommitCallback getAcknowledgementCommitCallback() {
```
artembilan: For consistency the `@Nullable` has to be next to the type.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -1320,4 +1321,53 @@ void shouldSupportAsyncCommits(EmbeddedKafkaBroker broker) throws Exception {
...
+		containerProps.setMessageListener((MessageListener<String, String>) record -> messageLatch.countDown());
+
+		ShareKafkaMessageListenerContainer<String, String> container =
+				new ShareKafkaMessageListenerContainer<>(consumerFactory, containerProps);
+		container.setBeanName("asyncCallbackTestContainer");
+		container.start();
+
+		Awaitility.await()
+				.atMost(10, TimeUnit.SECONDS)
+				.untilAsserted(() -> assertThat(messageLatch.getCount()).isZero());
```
artembilan: That's is interesting approach, but I guess regular `messageLatch.await(10, TimeUnit.SECONDS)` should be enough to assert instead of polling by the Awaitility.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -351,4 +352,123 @@ void shouldUseCommitAsyncWhenSyncShareCommitsIsFalse() throws Exception {
...
+		containerProperties.setMessageListener(messageListener);
+
+		ShareKafkaMessageListenerContainer<String, String> container =
+				new ShareKafkaMessageListenerContainer<>(mockFactory, containerProperties);
+		container.setBeanName("customCallbackContainer");
+		container.start();
+		Thread.sleep(200);
+		container.stop();
+
+		verify(mockConsumer).setAcknowledgementCommitCallback(customCallback);
```
artembilan, line 403: So, in the end we just like to verify that our callback is propagated down to the `ShareConsumer`.
Therefore, why do we need that extra hustle about producing records?
And why do we need that exta sleep between start and stop?
Probably stub with latch for the `setAcknowledgementCommitCallback()` would do the trick instead of sleeping for nothing?

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -351,4 +352,123 @@ void shouldUseCommitAsyncWhenSyncShareCommitsIsFalse() throws Exception {
...
+		ContainerProperties containerProperties = new ContainerProperties("test-topic");
+		containerProperties.setSyncShareCommits(false);
+		// No custom callback set
+		containerProperties.setMessageListener(messageListener);
+
+		ShareKafkaMessageListenerContainer<String, String> container =
+				new ShareKafkaMessageListenerContainer<>(mockFactory, containerProperties);
+		container.setBeanName("defaultCallbackContainer");
+		container.start();
+		Thread.sleep(200);
```
artembilan: Same here: feels like an overhead for the logic of this test.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -351,4 +352,123 @@ void shouldUseCommitAsyncWhenSyncShareCommitsIsFalse() throws Exception {
...
+
+		ContainerProperties containerProperties = new ContainerProperties("test-topic");
+		// syncShareCommits=true (default), no callback
+		containerProperties.setMessageListener(messageListener);
+
+		ShareKafkaMessageListenerContainer<String, String> container =
+				new ShareKafkaMessageListenerContainer<>(mockFactory, containerProperties);
+		container.setBeanName("noCallbackSyncContainer");
+		container.start();
+		Thread.sleep(200);
```
artembilan: DITTO

artembilan (changes requested):
Thanks for the update!
Let's see if my review makes sense!

PR #4393, opened 2026-04-11
GH-4326: Verify share.acquire.mode works through Spring's config model

### LINKED ISSUE #4326: Verify `share.acquire.mode` (KIP-1206) works through Spring's config model
Kafka 4.2 adds `share.acquire.mode` (`ConsumerConfig.SHARE_ACQUIRE_MODE_CONFIG`) with two values:
- `batch_optimized` (default) — soft limit on `max.poll.records`
- `record_limit` — strict enforcement

This is a pass-through consumer property. Verify it works through Spring's config model (`ConsumerProperties` / `ShareConsumerFactory`) and document it.

maxwell-balla:
Verify that the share.acquire.mode property passes through Spring's config model correctly.

The property is a pass-through consumer configuration that requires no changes to ContainerProperties or ShareKafkaMessageListenerContainer. DefaultShareConsumerFactory already propagates all consumer properties to the underlying KafkaShareConsumer.

Changes:
- Added unit test verifying the property is retained in factory configuration
- Added integration test verifying record_limit mode works end-to-end
- Added documentation for share.acquire.mode in the Share Consumer reference

Fixes gh-4326

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
@@ -314,7 +314,35 @@ When a consumer polls, the broker's share-partition leader:
 * Selects records in "Available" state
 * Moves them to "Acquired" state with a time-limited acquisition lock (default 30 seconds, configurable via `group.share.record.lock.duration.ms`)
 * Prefers to return complete record batches for efficiency
-* Applies `max.poll.records` as a soft limit, meaning complete record batches will be acquired even if it exceeds this value
+* Applies `max.poll.records` as a soft limit by default, meaning complete record batches will be acquired even if it exceeds this value
+
+[[share-acquire-mode]]
+===== Acquire Mode (KIP-1206)
+
+Apache Kafka 4.2 introduces the `share.acquire.mode` consumer property that controls how strictly `max.poll.records` is enforced:
+
+* `batch_optimized` (default) — soft limit.
+The broker may return more records than `max.poll.records` to align with batch boundaries for better throughput.
+* `record_limit` — strict limit.
+The broker will never return more records than `max.poll.records` per `poll()` call.
+Useful when you need precise control over memory usage or processing rate.
+
+This is a pass-through consumer property.
+Configure it directly in the consumer properties:
+
+[source,java]
+----
+Map<String, Object> props = new HashMap<>();
+props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092");
+props.put("share.acquire.mode", "record_limit");
+props.put("max.poll.records", 5);
+// ... other properties
+
+DefaultShareConsumerFactory<String, String> factory = new DefaultShareConsumerFactory<>(props);
+----
+
+NOTE: No additional Spring Kafka configuration is needed.
+The property is passed directly to the underlying `KafkaShareConsumer`.
 
 While records are acquired by one consumer, they are not available to other consumers.
 When the acquisition lock expires, unacknowledged records automatically return to "Available" state and can be delivered to another consumer.


spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/kafka-queues.adoc
```diff
@@ -314,7 +314,30 @@ When a consumer polls, the broker's share-partition leader:
...
-* Applies `max.poll.records` as a soft limit, meaning complete record batches will be acquired even if it exceeds this value
+* Applies `max.poll.records` as a soft limit by default, meaning complete record batches will be acquired even if it exceeds this value
+
+[[share-acquire-mode]]
+===== Acquire Mode (KIP-1206)
+
+Apache Kafka 4.2 introduces the `share.acquire.mode` consumer property that controls how strictly `max.poll.records` is enforced:
+
+* `batch_optimized` (default) — soft limit. The broker may return more records than `max.poll.records` to align with batch boundaries for better throughput.
+* `record_limit` — strict limit. The broker will never return more records than `max.poll.records` per `poll()` call. Useful when you need precise control over memory usage or processing rate.
```
sobychacko: Please use one sentence per line in asciidoc.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -654,4 +657,18 @@ void shouldPublishFailedToStartEventOnInitializationFailure() {
...
+		configs.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092");
+		configs.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
+		configs.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
+		configs.put("share.acquire.mode", "record_limit");
+
+		DefaultShareConsumerFactory<String, String> factory = new DefaultShareConsumerFactory<>(configs);
+
+		assertThat(factory.getConfigurationProperties())
+				.containsEntry("share.acquire.mode", "record_limit");
+	}
```
sobychacko: This test only checks that `share.acquire.mode` is still present in the factory’s configuration map, which the DefaultShareConsumerFactory already guarantees. It doesn’t validate anything about the underlying `ShareConsumer` or the Kafka client’s handling of the property. To keep the change focused and avoid duplicating concerns that belong to the Kafka client, I’d prefer we not add unit tests for this pass-through config and rely on the documentation instead. This is mainly because it is outside the boundary of the framework (spring kafka that is).

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerIntegrationTests.java
```diff
@@ -1364,4 +1365,40 @@ void shouldInvokeAcknowledgementCommitCallbackOnAsyncCommit(EmbeddedKafkaBroker
...
+
+		ShareKafkaMessageListenerContainer<String, String> container =
+				new ShareKafkaMessageListenerContainer<>(consumerFactory, containerProps);
+		container.setBeanName("acquireModeTestContainer");
+		container.start();
+
+		assertThat(allReceived.await(15, TimeUnit.SECONDS)).isTrue();
+
+		container.stop();
+	}
```
sobychacko: The test waits until all 20 records are processed, even with `max.poll.records` set to 5. That outcome would still be expected if `share.acquire.mode` were ignored or set to the default soft limit. The assertion never checks that each `poll()` returns at most 5 records, which is what record_limit is about. That kind of behavior is really owned by the Kafka broker/client, not something Spring Kafka should try to re-verify here. I suggest we drop this integration test from this PR.

sobychacko (changes requested):
Thanks for putting this together. I left a few inline review comments on the tests in the PR. If you get a chance, take a look at those first. I’d love to keep this PR mostly about documentation. The original issue (#4326)  was really about making sure the property behaves as expected when you actually run share consumers in a real app (i.e. you try `batch_optimized` vs `record_limit`, see that the broker/client does what you expect, and you’re not fighting Spring’s config along the way). That kind of verification is hard to capture meaningfully in our test suite without basically re-testing the kafka client which is beyond the scope of the issue. 

So my ask would be to lean only on the docs in this PR, and if you learn anything new and interesting while trying it on a real app (gotchas, what you observed, broker version, etc.), drop a short summary on the GitHub issue too so the community who run into any issues can learn from it. We can also aggregate that information into out docs too. What do you think? Happy to merge once it’s trimmed down along those lines.

sobychacko:
Thanks, @maxwell-balla as always. PR has been merged upstream now.

PR #4398, opened 2026-04-16
Fix/backoff recovery multi partition

### LINKED ISSUE #4371: Backoff breaks recovery on batching listener container in multi partition environment
spring-boot = "3.4.4"

**Describe the bug**

```
		@Nullable
			try {
				invokeBatchOnMessage(records, recordList);
				if (this.batchFailed) {
					this.batchFailed = false;
						this.commonErrorHandler.clearThreadState();
					}
					getAfterRollbackProcessor().clearThreadState();
				}
				if (!this.autoCommit && !this.isRecordAck) {
					processCommits();
				}
			}
			catch (RuntimeException e) {
					throw e;
				}
				try {
				}
```
doInvokeBatchListener is invoked with indeterministic record list.  
Therefore this situation quite often happens:

*Given:*   
we have 2 partitions: A(contains broken record), B(contains correct records).  

*When:*   
doInvokeBatchListener is executed with record list from A:   
batchFailed is set to true in invokeBatchErrorHandler and backoff is started  

doInvokeBatchListener is executed with record list from B:   
batchFailed is set to false and state of backoff is cleared (by `this.commonErrorHandler.clearThreadState();`)  

*Then:*  
Record from A has never a chance to go through full backoff cycle and be sent to dlq.  
Resulting in being stuck on this record and never being able to pass through.  

**To Reproduce**
Create a topic with >1 partition, add backoff to dlq recovery, and use non-concurrent container.

**Expected behavior**
Record from A should be sent to DLQ and processing should continue.

**Workaround**
Use Concurrent Container and set concurrency to the number of partition so each container can consume only from one partition (making therefore poll deterministic).

**Sample**
No sample.

**Proposed solution**
Change `this.commonErrorHandler.clearThreadState();` method to:  
`this.commonErrorHandler.clearTopicPartitionState(TopicPartition tp);`  
and change the code accordingly.

Arbaz4Sayyad:
Fixes #4371

**Problem:**
Backoff breaks recovery on batching listener container in multi-partition environments. When a batch from partition A fails and starts backoff, a subsequent successful batch from partition B would clear the backoff state for ALL partitions, preventing partition A from completing its backoff cycle.

**Solution:**
- Added [clearTopicPartitionState(TopicPartition)](cci:1://file:///d:/Projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java:167:1-175:2) method to [CommonErrorHandler](cci:2://file:///d:/Projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java:38:0-206:1) and [AfterRollbackProcessor](cci:2://file:///d:/Projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/AfterRollbackProcessor.java:42:0-120:1) interfaces
- Implemented topic-partition-specific state clearing in [FailedRecordTracker](cci:2://file:///d:/Projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java:48:0-322:1) and [DefaultAfterRollbackProcessor](cci:2://file:///d:/Projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/DefaultAfterRollbackProcessor.java:59:0-278:1)
- Updated [doInvokeBatchListener](cci:1://file:///d:/Projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java:2348:2-2399:3) to clear state per partition instead of thread-wide clearing

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
@@ -169,6 +169,9 @@ public class RetryableTopicAnnotationProcessor {
 
 		Integer attempts = resolveExpressionAsInteger(annotation.attempts(), "attempts", true);
 		if (attempts != null) {
+			if (attempts <= 0) {
+				throw new IllegalArgumentException("Retry attempts must be greater than 0, but got: " + attempts);
+			}
 			builder.maxAttempts(attempts);
 		}
 		Integer concurrency = resolveExpressionAsInteger(annotation.concurrency(), "concurrency", false);
--- a/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java
@@ -169,9 +169,6 @@ public class RetryableTopicAnnotationProcessor {
 
 		Integer attempts = resolveExpressionAsInteger(annotation.attempts(), "attempts", true);
 		if (attempts != null) {
-			if (attempts <= 0) {
-				throw new IllegalArgumentException("Retry attempts must be greater than 0, but got: " + attempts);
-			}
 			builder.maxAttempts(attempts);
 		}
 		Integer concurrency = resolveExpressionAsInteger(annotation.concurrency(), "concurrency", false);
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AfterRollbackProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AfterRollbackProcessor.java
@@ -94,6 +94,16 @@ public interface AfterRollbackProcessor<K, V> {
 	}
 
 	/**
+	 * Optional method to clear state for a specific topic partition.
+	 * This can be used to clear state for a single partition without affecting
+	 * other partitions, which is useful in multi-partition environments.
+	 * @param topicPartition the topic partition.
+	 * @since 3.2
+	 */
+	default void clearTopicPartitionState(TopicPartition topicPartition) {
+	}
+
+	/**
 	 * Return true to invoke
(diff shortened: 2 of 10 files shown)


Arbaz4Sayyad:
Hi @mentor,

I've opened a pull request to fix issue #4371 - "Backoff breaks recovery on batching listener container in multi partition environment."

Summary of changes:

Added clearTopicPartitionState(TopicPartition) method to CommonErrorHandler and AfterRollbackProcessor interfaces
Implemented topic-partition-specific state clearing in FailedRecordTracker and DefaultAfterRollbackProcessor
Updated doInvokeBatchListener to clear state per partition instead of thread-wide clearing
Problem solved: Previously, when a batch from partition A failed and started backoff, a subsequent successful batch from partition B would clear the backoff state for ALL partitions. This prevented partition A from completing its backoff cycle and being sent to DLQ, causing the consumer to get stuck.

Solution: The fix uses topic-partition-specific state clearing, so each partition maintains its own backoff state independently. This allows proper backoff recovery in multi-partition environments.

Could you please review the changes and provide feedback?

PR: https://github.com/Arbaz4Sayyad/spring-kafka/pull/new/fix/backoff-recovery-multi-partition

artembilan:
> `Hi @mentor,`

How do you determine who to call on the PR?

That GH user is not Spring team member.
Please, respect time and boundaries other community members calling them out for discussion which doesn't belong to them.
You used `maintainers` GH handle in other place, and I've already pointed you out that it was not OK.

> Summary of changes:

Why do you do duplicate the text in the PR description and in the additional comment?

> Added clearTopicPartitionState(TopicPartition)

But I see that there are much more commits in this PR then you claim.
Having an experience with you on other PRs, I can say that you don't start from fresh branch based on the `main`, but somehow have all your changes spread everywhere.

Can you please, close this PR and revise the local state of code base?
And then come back to us with clean PR if that would be still necessary though.

Thanks for understanding!

PR #4416, opened 2026-04-19
Improve description of messageListener in container properties documentation

Gautam-aman:
Improve clarity of the `messageListener` property description by specifying that it processes records consumed from Kafka topics.


artembilan (changes requested):
Looks like this PR has commits from the https://github.com/spring-projects/spring-kafka/pull/4391, which is not related to the subject.

please, consider to raise PR from fresh branch created agains the latest `upstream/main`. We call top-level branch.

Thanks

PR #4417, opened 2026-04-19
Improve description of messageListener in container properties documentation

Gautam-aman:
Improve clarity of the messageListener property description by specifying that it processes records consumed from Kafka topics.


artembilan:
Thanks for the update, @Gautam-aman !
We will review your PRs quickly tomorrow before release.

PR #4419, opened 2026-04-22
Fix inconsistent handling of non-string values in Properties argument

### LINKED ISSUE #4384: Inconsistent handling of non-string values in Properties argument passed to DefaultConsumerFactory::createKafkaConsumer

Arbaz4Sayyad:
Fixes #4384
 
## Problem
The [createKafkaConsumer](cci:1://file:///d:/Projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java:422:1-440:2) method in [DefaultKafkaConsumerFactory](cci:2://file:///d:/Projects/spring-kafka/spring-kafka/src/main/java/org/springframework/kafka/core/DefaultKafkaConsumerFactory.java:80:0-532:1) inconsistently handles non-string values in the `Properties` argument. When properties contain only non-string values (e.g., `Integer`), they are not passed to the Kafka consumer, while the same non-string values are passed when accompanied by string properties.
 
**Example of the bug:**
```java
// Case 1: Integer value alone - NOT passed to consumer
Properties overrides = new Properties();
overrides.put(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, 2); // Integer
factory.createConsumer(null, null, null, overrides);
// Result: max.poll.records is NOT passed to KafkaConsumer
 
// Case 2: Integer with string property - IS passed to consumer
Properties overrides = new Properties();
overrides.put(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, 2); // Integer
overrides.put("additional", "property"); // String
factory.createConsumer(null, null, null, overrides);
// Result: max.poll.records IS passed to KafkaConsumer
This occurs because the method has an if/else that checks properties.stringPropertyNames().isEmpty(). When properties contain only non-string values, it bypasses createConsumerWithAdjustedProperties and ignores the override properties.

Solution
Removed the if/else condition that bypassed createConsumerWithAdjustedProperties when properties contained non-string values. Now the method always calls createConsumerWithAdjustedProperties, which properly handles both string and non-string property values.

The createConsumerWithAdjustedProperties method already handles all the conditions that were in the if statement, so removing the if/else is safe and eliminates the inconsistency.

Intentions
Consistency: Ensure property overrides work consistently regardless of value type (String, Integer, Long, etc.)
Simplicity: Remove unnecessary conditional logic causing the inconsistency
User Experience: Users should be able to pass property overrides with any value type without workarounds
Maintainability: Simpler code flow, easier to understand and maintain
Test Coverage
Added two new test cases:

testOverriddenMaxPollRecordsOnlyAsInteger() - Verifies Integer property alone is passed to consumer
testOverriddenMaxPollRecordsAsIntegerWithAdditionalStringProperty() - Verifies Integer with string property is passed to consumer
All existing tests continue to pass.

Impact
Risk: Low - Simplifies code and removes inconsistent behavior
Breaking changes: None - Bug fix that aligns with user expectations
Performance: No impact


artembilan:
@Arbaz4Sayyad ,

Why do you do this?
We do have a PR on the matter: https://github.com/spring-projects/spring-kafka/pull/4391.

So, this is a duplication of the effort.
And wasting of time: yours and ours.
Plus your PR contains stuff which does not belong to the subject.

PR #4431, opened 2026-04-27
GH-4430: Wrong record content passed to `KafkaStreamsDeadLetterDestinationResolver`

loicgreffier:


sobychacko:
@loicgreffier There are test failures on CI. Could you make sure if they pass on your end? Thanks!

loicgreffier:
@sobychacko Sorry for the delay. I've updated the PR. List of updates:

1. The `KafkaStreamsDeadLetterDestinationResolver` now correctly receives the record as it is at the input of the processor where it fails. The record used to build the dead letter payload is still the raw source payload.

2. The tests have been strenghten to validate the dead letter record content and the `DeadLetterDestinationResolver` input record. Mocked values have been updated to be more human understandable.

Maybe we should clarify the documentation about the DLQ record content:
- The source raw key
- The source raw value
- The source headers + additional Spring Kafka headers (already described in the documentation)

Also, for `processing` and `production` handlers, the `KafkaHeaders.DLT_ORIGINAL_TIMESTAMP_TYPE` is always `NO_TIMESTAMP_TYPE` since we cannot identify the source timestamp type (I assume we cannot, but this is not too critical I guess...): https://github.com/spring-projects/spring-kafka/blob/84ed275f8571e686d85c71fa9fa1e799243b178f/spring-kafka/src/main/java/org/springframework/kafka/streams/RecoveringProcessingExceptionHandler.java#L63

Maybe we should specify all of this somewhere (in a dedicated dead letter record section ?). Please let me know your opinion

spring-kafka/src/test/java/org/springframework/kafka/streams/RecoveringProductionExceptionHandlerTests.java
```diff
@@ -46,13 +50,39 @@ protected RecoveringProductionExceptionHandler createHandler(Map<String, Object>
...
+				context.topic(),
+				context.partition(),
+				context.offset(),
+				12345L,
+				TimestampType.NO_TIMESTAMP_TYPE,
+				ConsumerRecord.NULL_SIZE,
+				ConsumerRecord.NULL_SIZE,
+				"key".getBytes(),
+				"value".getBytes(),
+				context.headers(),
```
sobychacko: Aren't we silently dropping the `headers` var above? Why are we using `context.headers()` here? did you mean the `headers`? My understanding is that you need to pass the headers from the processor stage rather than the source record headers, no? Also, is there anything wrong with the assertions then? because the test is apparently passing.

spring-kafka/src/main/java/org/springframework/kafka/streams/AbstractRecoveringExceptionHandler.java
```diff
@@ -71,27 +71,26 @@ protected AbstractRecoveringExceptionHandler(ConsumerRecordRecoverer recoverer)
 	 * @param context the error handler context
 	 * @param record the consumer record that caused the error
 	 * @param exception the exception that occurred
+	 * @param sourceRecord the source record as it was at the input of the sub-topology
```
sobychacko: maybe mention that these are raw bytes?

spring-kafka/src/main/java/org/springframework/kafka/streams/AbstractRecoveringExceptionHandler.java
```diff
@@ -71,27 +71,26 @@ protected AbstractRecoveringExceptionHandler(ConsumerRecordRecoverer recoverer)
 	 * @param context the error handler context
 	 * @param record the consumer record that caused the error
```
sobychacko: Maybe call it out as typed processor-stage midstream record? (as a javadoc comment)

sobychacko (changes requested):
Added some comments. Do you think we should also need any doc changes? Thanks!

spring-kafka/src/main/java/org/springframework/kafka/streams/AbstractRecoveringExceptionHandler.java
loicgreffier: Updated, I've added `unserialized source raw record`

spring-kafka/src/main/java/org/springframework/kafka/streams/AbstractRecoveringExceptionHandler.java
loicgreffier: Updated

spring-kafka/src/test/java/org/springframework/kafka/streams/RecoveringProductionExceptionHandlerTests.java
loicgreffier: The var was unused and has been removed.

`createDestinationResolverConsumerRecord` creates the `ConsumerRecord` expected by the `KafkaStreamsDeadLetterDestinationResolver`. The deserialization, processing, and production test classes each have their own implementation of this function because the expected record differs depending on the exception handler that invokes the `DestinationResolver` (e.g., for deserialization, the timestamp type is expected to be `TimestampType.CREATE_TIME`, the key and value are of type `byte[]`, etc.).

I've updated the tests so that `createDestinationResolverConsumerRecord` builds the expected `KafkaStreamsDeadLetterDestinationResolver` record from both the context and the record passed to the `handleError` function (and not the context only, which was confusing and not exact), as implemented for processing and production exception handlers, or from the record passed to the `handleError` function only, for the deserialization exception handler.

`handleError` (in the unit tests) no longer relies on `createDestinationResolverConsumerRecord`, which was also confusing.

loicgreffier:
@sobychacko I’ve updated the PR with the feedback. I also updated the documentation to mention the dead-letter record content, without changing it too much.

spring-kafka/src/test/java/org/springframework/kafka/streams/RecoveringProductionExceptionHandlerTests.java
sobychacko: Ok, thanks for the explanation. Are you sure that the `context.headers()` issue i raised above is addressed? From what I can see, both `this.record.headers()` and `context.headers()` end up with the same "original-header", so `KafkaStreamsDestinationResolverWithAsserts` would pass even if the bug were still present and the handler was passing the source record to the resolver. To make the header assertion meaningful, `this.record` should carry a distinct header (e.g. "processor-header") that differs from `context.headers()`, so there's no ambiguity about which record the resolver is actually seeing. Or do I miss something in my understanding? Anyways, not a big deal, something to think about, but i will wait for you to comment before merging.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/streams.adoc
```diff
@@ -331,7 +331,10 @@ The recovery strategies are as follows, by priority order:
 - If a `ConsumerRecordRecoverer` implementation is defined, invoke it and resume the stream without producing dead-letter records, as handling is delegated to the `ConsumerRecordRecoverer`. For example, the provided xref:streams.adoc#dead-letter-publishing-recoverer[`DeadLetterPublishingRecoverer`] implementation can be used.
 - Fail the stream without producing dead-letter records.
 
-When a dead-letter record is published to a dead-letter topic, whether through the native Kafka Streams DLQ or via a `ConsumerRecordRecoverer` implementation, the record is enriched with the xref:kafka/annotation-error-handling.adoc#dead-letters[Spring for Apache Kafka DLT headers].
+When a dead-letter record is published to a dead-letter topic, whether through the native Kafka Streams DLQ or via a `ConsumerRecordRecoverer` implementation:
+
+- The unserialized source raw key, source raw value and source headers are used to build the record.
```
sobychacko: maybe we don't need to say `unserialized` since it comes from Kafka and in serialized form - just simply say the `raw source key`??

loicgreffier: Fixed

spring-kafka/src/test/java/org/springframework/kafka/streams/RecoveringProductionExceptionHandlerTests.java
loicgreffier: @sobychacko I've renamed `record.headers()` to `processor-header` in the processing and production exception handler tests. 

For the deserialization exception handler:
- The `record.headers()` and `context.headers()` should be the same, since the record is the source record itself.
- The dead letter record key, value and headers come from the `handleError` input record, not from the context (both should be equal). I did not rebuild an intermediate `ConsumerRecord` for that. This differs a little bit from the processing and production handlers, where a `ConsumerRecord` is built with the `context` data: https://github.com/loicgreffier/spring-kafka/blob/70d4c81d909b06fa306248a265d92e07de1fb791/spring-kafka/src/main/java/org/springframework/kafka/streams/RecoveringProcessingExceptionHandler.java#L58. 

‼️ Maybe it's worth building a new `ConsumerRecord` using the source information from the `context` for the deserialization exception handler as well, just to align the behavior across all exception handlers and ensure that all the information used to build the dead-letter record comes from the `context` object.

sobychacko: Thanks for the updates and the explanation. I think we can leave it like that for the deser handler.

sobychacko:
@loicgreffier Thanks for the PR and for being patient with the reviews back and forth! It is now merged upstream. Great job and thanks again!

PR #4447, opened 2026-05-16
make the method setBackOffFunction work for the batch processing as well (backport to the branch 3.3.x)

LeosBitto:
The method setBackOffFunction of FailedRecordProcessor and its descendants (FailedBatchProcessor and DefaultErrorHandler) is useful for classifying the exceptions thrown from the application code (listeners). However, I find it very surprising that this method does not work when processing the messages in batches (the supplied backOffFunction is never called there). Please review my attempt to make it work for the batch processing as well.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
@@ -78,17 +78,51 @@ public final class ErrorHandlingUtils {
 	 * @param logLevel the log level.
 	 * @param retryListeners the retry listeners.
 	 * @param classifier the exception classifier.
-	 * @param reClassifyOnExceptionChange true to reset the state if a different exception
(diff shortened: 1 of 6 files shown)


sobychacko:
@LeosBitto Thanks for the PR! This is targeting `3.3.x` but it should go against `main`. Then we can think about back-porting. Keep in mind that the API changed on the latest versions`: `BinaryExceptionClassifier ` is now `ExceptionMatcher` and `getClassifier()` is `getExceptionMatcher()`. Can you retarget the PR on `main`? One other thing, we have releases coming up in a week or so, so if you want to make this into the `4.1.0` release (and back-ported to the other branches), we only have a week's time. Let us know how you want to proceed. Thanks!

LeosBitto:
> @LeosBitto Thanks for the PR! This is targeting `3.3.x` but it should go against `main`. Then we can think about back-porting.

I thought that a forward merge would be used, but if you prefer backporting instead then it is fine for me too. I have opened #4455 against main.

sobychacko:
Ok, thanks. Closing this in favor of #4455.

PR #4450, opened 2026-05-26
GH-4444: Fix COUNT_TIME ack mode ignoring the time condition

### LINKED ISSUE #4444: KafkaMessageListenerContainer for ackMode=COUNT_TIME doesn't check ackTime

Between 3.2.x and 4.0.x

**Describe the bug**

In `KafkaMessageListenerContainer` there is a bug for `ackMode=COUNT_TIME`.
https://github.com/spring-projects/spring-kafka/blob/5be5015a67952ab88891c05676b58489bac12ee2/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java#L3249-L3253
For `ackMode=COUNT_TIME`, both `isCountAck` and `isTimeAck` are true. However, only `countAcks()` will be executed because of the if-else if structure – `timedAcks()` is never called.

**To Reproduce**

Check logic in `KafkaMessageListenerContainer.processCommits()`.

**Expected behavior**

`KafkaMessageListenerContainer` should consider both `ackTime` and `ackCount` for `ackMode=COUNT_TIME`.

serelli:
## Summary

Fixes #4444.

In `KafkaMessageListenerContainer.processCommits()`, the `if-else if` structure caused `timedAcks()` to be unreachable for `COUNT_TIME` ack mode.

For `COUNT_TIME`, both `isCountAck` and `isTimeAck` are `true`. Because of the `else if`, `timedAcks()` was dead code — only the count threshold was ever checked, meaning the time component of `COUNT_TIME` was silently ignored.

### Root cause

```java
// BEFORE — timedAcks() unreachable for COUNT_TIME
if (this.isCountAck) {
    countAcks();
}
else if (this.isTimeAck) {   // never reached when isCountAck == true
    timedAcks();
}
```

### Fix

Split into two independent `if` checks so both conditions are evaluated for `COUNT_TIME`:

```java
if (this.isCountAck) {
    countAcks();
}
if (this.isTimeAck) {
    timedAcks();
}
else if (!this.isCountAck && !this.isManualImmediateAck) {
    commitIfNecessary();
    this.count = 0;
}
```

**No double-commit risk**: when `countAcks()` fires it resets `last = now`, so the subsequent `timedAcks()` call sees `elapsed ≈ 0` and skips committing again. All other ack modes (`COUNT`, `TIME`, `RECORD`, `BATCH`, `MANUAL`, `MANUAL_IMMEDIATE`) are unaffected.

## Tests

Added `testCountTimeAckModeCommitsOnTimeExpiry` to `KafkaMessageListenerContainerTests` — a mock-based test that configures `COUNT_TIME` with a large `ackCount` (1000, never reached by 2 records/poll) and a small `ackTime` (300 ms), then asserts that `commitSync` is called within 10 s, exercising the previously broken time-based commit path.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -3249,10 +3249,10 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			if (this.isCountAck) {
 				countAcks();
 			}
-			else if (this.isTimeAck) {
+			if (this.isTimeAck) {
 				timedAcks();
 			}
-			else if (!this.isManualImmediateAck) {
+			else if (!this.isCountAck && !this.isManualImmediateAck) {
 				commitIfNecessary();
 				this.count = 0;
 			}
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/KafkaMessageListenerContainerTests.java
@@ -523,6 +523,52 @@ public class KafkaMessageListenerContainerTests {
 		verify(consumer.get(), times(2)).commitSync(anyMap(), any());
 	}
 
+	@SuppressWarnings("unchecked")
+	@Test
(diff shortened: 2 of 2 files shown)


sobychacko:
Thanks for the PR, but closing it since this is a duplicate of this one: https://github.com/spring-projects/spring-kafka/pull/4449.

PR #4460, opened 2026-05-30
GH-4442: Fix value serializer mappings for different classloaders

### LINKED ISSUE #4442: Different behavior for value serializer mappings for different classloaders

spring-kafka 4.0.5

**Describe the bug**

Value serializer mappings are not honored (have a non-deterministic behavior) when multiple classloaders are involved.
Consider a property-based (i.e. via `spring.kafka.*` properties) kafka template initialization. Property-based seems to be important here because beans are initialized in a lazy manner.
Mappings (i.e. `spring.kafka.producer.properties.spring.json.type.mapping` entries) are created and cached on template initialization. Also, mappings are stored/indexed based on the class bytes (not the class string).
So, if the template happens to be initialized under classloader A, then a mapping is created for the class in question with the class bytes bound to classloader A. If at a later time the the template is accessed using a message class that happens to be bound to another classloader, looking up the mapping fails and the resulting message is not the expected one.

One case of different classloaders is when using spring dev tools (good thing it's local only...). In the sample project below, the force-init in a different classloader happens when a kafka write is triggered inside a reactive chain which in turn involves a custom-built mongo client. Using the default mongo client as built by spring, resolves the issue. Using custom kafka beans (i.e. not based on properties but programmatically configuring them) also resolves the issue.

The problem here is the non-deterministic behavior. Because kafka template is initialized lazily, whoever initializes first, also "owns" the class mappings.
Not necessarily expecting a fix but i'd like to hear your thoughts ☔

On a sidenote, this might look like a weird edge case but it has caused real pain in a real project in real life 🚑

**To Reproduce**

- use properties to create kafka configuration
- configure a mapping for produced messages, e.g. `NotificationEvent:com.example.demo.NotificationEvent`
- force the initialization of kafka beans under a different classloader

**Expected behavior**

- messages should have a header of: `[__TypeId__ :: NotificationEvent]`

**Actual behavior**

- messages have the (fallback) FQN header: `[__TypeId__ :: com.example.demo.NotificationEvent]`

**Sample**

https://github.com/diamondT/demo-kafka-mappings

`./mvnw clean spring-boot:run`

sobychacko:
Fixes: https://github.com/spring-projects/spring-kafka/issues/4442

When the type mapper is initialized under one classloader and messages are produced under another (e.g. Spring DevTools restart classloader), the reverse lookup map keyed on `Class<?>` identity misses the entry because two `Class<?>` objects representing the same class loaded by different classloaders are not equal in a `HashMap` lookup. The fallback writes the FQCN header instead of the configured alias, silently breaking consumers that depend on the alias.

Fix by changing the reverse map key from `Class<?>` to `String` (using `clazz.getName()`), which is classloader-agnostic.

The same change is applied to the deprecated `AbstractJavaTypeMapper` for Jackson 2.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/mapping/AbstractJavaTypeMapper.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/mapping/AbstractJavaTypeMapper.java
@@ -79,7 +79,7 @@ public abstract class AbstractJavaTypeMapper implements BeanClassLoaderAware {
 
 	private final Map<String, Class<?>> idClassMapping = new ConcurrentHashMap<String, Class<?>>();
 
-	private final Map<Class<?>, byte[]> classIdMapping = new ConcurrentHashMap<Class<?>, byte[]>();
+	private final Map<String, byte[]> classIdMapping = new ConcurrentHashMap<String, byte[]>();
 
 	private String classIdFieldName = DEFAULT_CLASSID_FIELD_NAME;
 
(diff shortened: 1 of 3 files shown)


sobychacko:
When back-porting to `3.3.x` only jackson2 specific changes need to be cherry-picked as Jackson3 was only introduced in `4.0.x`.

PR #4463, opened 2026-06-02
GH-4436: Improve Javadoc on BatchListenerFailedException and FailedBatchProcessor

### LINKED ISSUE #4436: `BatchListenerFailedException` silently commits offsets for unprocessed records in `BatchMessageListener`
## Summary

When a `BatchMessageListener` throws a `BatchListenerFailedException`, Spring for Apache Kafka's `DefaultErrorHandler` commits the offsets of all records **before** the indicated failed record. This behavior is dangerous because it assumes that all preceding records have been fully and successfully processed — an assumption that does not hold in many legitimate `BatchMessageListener` use cases, leading to **silent data loss**.

### Current Behavior

When a `BatchMessageListener` throws a `BatchListenerFailedException(message, cause, failedRecord)`, the `DefaultErrorHandler`:

1. Identifies the index of the failed record in the batch.
2. **Commits offsets for all records before that index** (treating them as successfully processed).
3. Seeks back to the failed record for retry.

### Why This Is Dangerous

The `BatchMessageListener` interface exists specifically for use cases where batch processing does **not** follow a simple sequential, one-record-at-a-time model. Common patterns include:

1. **Parallel processing**: Records in the batch are processed concurrently (e.g., using virtual threads). If record N fails, records N+1, N+2, … may have succeeded, while records before N may still be in-flight or may have produced side effects that need to be rolled back.

2. **Multi-step pipelines**: A batch listener may process records in multiple stages — e.g., transform, then produce to an output topic. If an error occurs during the *produce* step, the *transform* step for earlier records may have completed, but their output was never actually sent. Committing their offsets means those events are permanently lost.

3. **Reordered processing**: The listener may deliberately reorder records (e.g., grouping by key for deduplication). The positional index of the failed `ConsumerRecord` in the original batch has no meaningful relationship to which records have actually been processed.

4. **Transactional processing**: The listener may be doing work inside a transaction. If any record fails, the entire transaction should be rolled back — but Spring for Apache Kafka has already committed offsets for the "successful" records, making rollback impossible.

### Concrete Example

Consider a `BatchMessageListener` that:

1. Transforms each input record (Input → Stage 1 → Stage 2 → Output).
2. Produces the transformed records to an output Kafka topic.

```java
public void onMessage(List<ConsumerRecord<byte[], GenericRecord>> data) {
    // Step 1: Transform all records (parallel, using virtual threads)
    var results = transformAll(data);

    // Step 2: Produce all results to the output topic
    var futures = results.stream()
            .map(kafkaTemplate::send)
            .toArray(CompletableFuture[]::new);
    CompletableFuture.allOf(futures).join();
}
```

If `transformAll()` throws a `BatchListenerFailedException` pointing to record index 5:

- **Expected**: No offsets are committed. The entire batch is retried.
- **Actual**: Offsets for records 0–4 are committed, even though `kafkaTemplate::send` was **never called** for any of them. Those 5 events are silently and permanently lost.

This is not a theoretical concern — we encountered this exact bug in a production system in our organization, where a batch listener was processing records in parallel and threw `BatchListenerFailedException` on a failure, resulting in data loss for all preceding records.

### Why This Should Not Be the Default

While this behavior is [documented in the reference guide](https://docs.spring.io/spring-kafka/reference/kafka/annotation-error-handling.html#batch-listener-error-flow), it remains highly unexpected and error-prone in practice. The documentation describes the mechanism, but does not warn about the data-loss risk. A developer encountering a batch processing failure will naturally reach for `BatchListenerFailedException` — it is, after all, the only exception specifically designed for `BatchMessageListener` — without realizing that it triggers partial offset commits that silently discard unprocessed events.

Critically, the Javadoc of `BatchListenerFailedException` itself does not mention the offset-commit side effect at all. Developers who discover the exception via IDE auto-completion or Javadoc have no indication of this behavior.

The `BatchMessageListener` API makes no guarantees about processing order or completion semantics. The error handler should not make assumptions that contradict the flexibility the API provides.

### Option A: Change the Default Behavior (Preferred)

`BatchListenerFailedException` should **not** cause the `DefaultErrorHandler` to commit offsets for records before the failed index. Instead, the entire batch should be retried — the same behavior as throwing any other `RuntimeException`.

If the current "commit preceding offsets" behavior is desired by some users, it should be opt-in via:

- A flag on `DefaultErrorHandler` (e.g., `setCommitRecoveredOffsets(true)`), or
- A separate, explicitly named exception (e.g., `PartialBatchFailedException`).

### Option B: Document the Behavior and Clarify Intent

At minimum, the Javadoc of `BatchListenerFailedException` must explicitly describe the offset-commit side effect. Currently, the Javadoc does not mention this behavior at all.

**Proposed Javadoc for `BatchListenerFailedException`:**

```java
/**
 * Exception to be thrown by a {@link BatchMessageListener} to indicate that a
 * specific record within the batch has failed processing.
 *
 * <p><strong>Important:</strong> When this exception is handled by the
 * {@link org.springframework.kafka.listener.DefaultErrorHandler}, it will
 * <strong>commit the offsets of all records preceding the failed record</strong>
 * in the batch, treating them as successfully processed. The failed record
 * (and all subsequent records) will be seeked back for retry or recovery.
 *
 * <p>This means that throwing this exception <strong>implicitly asserts</strong>
 * that all records before the failed record have been <strong>fully and
 * irreversibly processed</strong>, including any side effects such as producing
 * to output topics, writing to databases, or calling external services.
 *
 * <p>If your {@link BatchMessageListener} processes records in parallel,
 * in a non-sequential order, in multiple steps, or within a transaction,
 * do <strong>not</strong> use this exception — throw a standard
 * {@link RuntimeException} instead to ensure the entire batch is retried
 * without committing any offsets.
 *
 * @see org.springframework.kafka.listener.DefaultErrorHandler
 * @see BatchMessageListener
 */
```

Adityapal67:
Fixes GH-4436 (https://github.com/spring-projects/spring-kafka/issues/4436)

* Clarify that throwing BatchListenerFailedException asserts all preceding records were fully and irreversibly processed before the failed index
* Warn against use in parallel, multi-step, or transactional batch listeners
* Add matching warning to FailedBatchProcessor class-level Javadoc

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchListenerFailedException.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchListenerFailedException.java
@@ -24,13 +24,36 @@ import org.jspecify.annotations.Nullable;
 import org.springframework.kafka.KafkaException;
 
 /**
- * An exception thrown by user code to inform the framework which record in a batch has
- * failed.
+ * An exception thrown by a batch listener to indicate which record in the batch
+ * failed. The framework will commit the offsets of all records <em>before</em>
+ * the failed record and seek the remaining records (starting from the failed one)
+ * for redelivery.
  *
- * @author Gary Russell
+ * <p><strong>Important contract:</strong> throwing this exception carries an
+ * implicit assertion that every record before the indicated index (or before
+ * the provided {@link org.apache.kafka.clients.consumer.ConsumerRecord}) has
+ * been <em>fully and irreversibly</em> processed. The framework commits those
+ * preceding offsets immediately; they will not be redelivered.
+ *
+ * <p>This exception is <strong>not</strong> appropriate when:
+ * <ul>
+ *   <li>records are processed in parallel (e.g. using virtual threads or an
+ *       executor), because preceding records may still be in-flight when the
+ *       exception is thrown;</li>
+ *   <li>processing involves multiple steps (e.g. transform then produce to an
+ *       output topic), because earlier records may have completed only some
+ *       steps;</li>
+ *   <li>the listener is transactional and a full rollback is required on any
+ *       failure.</li>
+ * </ul>
+ * In those cases, throw a plain {@link RuntimeException} instead so that the
+ * entire batch is redelivered.
+ *
+ * @author Aditya Pal
  * @author Wang Zhiyang
  * @since 2.5
- *
+ * @see org.springframework.kafka.listener.DefaultErrorHandler
+ * @see org.springframework.kafka.listener.FailedBatchProcessor
  */
 public class BatchListenerFailedException extends KafkaException {
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedBatchProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedBatchProcessor.java
@@ -40,18 +40,29 @@ import org.springframework.kafka.support.KafkaUtils;
 import org.springframework.util.backoff.BackOff;
 
 /**
- * Subclass of {@link FailedRecordProcessor} that can process (and recover) a batch. If
- * the listener throws a {@link BatchListenerFailedException}, the offsets prior to the
- * failed record are committed and the remaining records have seeks performed. When the
- * retries are exhausted, the failed record is sent to the recoverer instead of being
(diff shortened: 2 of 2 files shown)


sobychacko:
Merged via https://github.com/spring-projects/spring-kafka/commit/01755d34f300f173669c1d2f69a391ef311054cb. Thanks for the PR!

PR #4467, opened 2026-06-04
make the method setBackOffFunction work for the batch processing as well (backport to the branch 3.3.x)

LeosBitto:
This is a backport of the commit 7c9f5c0b0dcd5f027708112fdb703e5410f705fd to the branch 3.3.x which does not modify any existing public API. The only addition to the public API is a new overload of the method retryBatch in ErrorHandlingUtils which supports the implementation of a custom BackOffFunction and resetStateOnExceptionChange for the batch processing.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
@@ -51,6 +51,7 @@ import org.springframework.util.backoff.BackOffExecution;
  * @author Antonio Tomac
  * @author Wang Zhiyang
  * @author Sanghyeok An
+ * @author Leos Bitto
  *
  * @since 2.8
  *
@@ -78,7 +79,7 @@ public final class ErrorHandlingUtils {
 	 * @param logLevel the log level.
 	 * @param retryListeners the retry listeners.
 	 * @param classifier the exception classifier.
-	 * @param reClassifyOnExceptionChange true to reset the state if a different exception
+	 * @param reClassifyOnExceptionChange true to reclassify the exception if a different exception
 	 * is thrown during retry.
 	 * @since 2.9.7
 	 */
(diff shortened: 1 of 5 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
```diff
@@ -87,8 +88,40 @@ public static void retryBatch(Exception thrownException, ConsumerRecords<?, ?> r
...
+	 * @param logger the logger.
+	 * @param logLevel the log level.
+	 * @param retryListeners the retry listeners.
+	 * @param classifier the exception classifier.
+	 * @param reClassifyOnExceptionChange true to reclassify the exception if a different exception
+	 * is thrown during retry.
+	 * @param resetStateOnExceptionChange true if a different exception thrown during retry should end this method.
+	 * @return a new exception if resetStateOnExceptionChange was true and a different exception has occurred
+	 * (retry with a possibly different BackOffExecution is expected if non-null is returned)
+	 * @since 4.0.7
```
sobychacko: The problem with this is that we have the same method on 4.1.x and there we use the since tag as `4.1.0`, which is kind of moot since we are introducing this method all way back to 3.3.x. What do you think?

sobychacko (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/ErrorHandlingUtils.java
LeosBitto: I have changed it to include the versions with backport. If you agree, the same change should be made to main as well - there should be no need to explicitly mention 4.1.0, because all versions after 4.0.6 contain the change.

sobychacko:
@LeosBitto I didn't get a chance to fully review the changes on the other branches. I'm still not sure about the new API methods added, especially given that 3.3.x is going to be out of OSS soon. Anyways, we can certainly pick these up in next month's maintenance release if we don't get a chance to merge these PRs before this month's release. Paging @artembilan also for his opinion. Thanks!

PR #4470, opened 2026-06-05
GH-4468: Fix no-op ack detection for non-null Acknowledgment

### LINKED ISSUE #4468: @RetryableTopic: built-in DLT logging handler fails with "No Acknowledgment available" on 4.0 (works on 3.3.x)
### What happens

With a `@RetryableTopic` listener and **no** `@DltHandler`, once a record exhausts its retries and reaches the DLT, the framework's built-in DLT logging consumer fails on every record with:

```
org.springframework.kafka.listener.ListenerExecutionFailedException: invokeHandler Failed
Caused by: java.lang.IllegalStateException: No Acknowledgment available as an argument, the listener container must have a MANUAL AckMode to populate the Acknowledgment.
```

Because the DLT consume fails, the record is re-published to the same DLT (`... won't be retried. Sending to DLT ...`), so it loops and the DLT topic keeps growing.

This works fine on spring-kafka 3.3.x and starts failing after upgrading to 4.0.x. I didn't change any acknowledgment configuration — only the version.

### Affected versions

- Broken: Spring Boot 4.0.5 / spring-kafka 4.0.4
- Works: Spring Boot 3.5.5 / spring-kafka 3.3.13

### Minimal reproducer

A plain Spring Boot app using the default ack mode (no `ack-mode` set, `enable-auto-commit: false`) and `String` key/value (so there is no deserialization/serialization involved — the value is just a `String`).

```java
@Component
class DemoListener {

    @KafkaListener(topics = "demo.topic", groupId = "demo")
    @RetryableTopic(attempts = "2", backOff = @BackOff(value = 1000))
    public void handle(ConsumerRecord<String, String> record) {
        throw new RuntimeException("force retry -> DLT");
    }
}
```

```yaml
spring:
  kafka:
    bootstrap-servers: localhost:9092
    consumer:
      enable-auto-commit: false
      key-deserializer: org.apache.kafka.common.serialization.StringDeserializer
      value-deserializer: org.apache.kafka.common.serialization.StringDeserializer
    producer:
      key-serializer: org.apache.kafka.common.serialization.StringSerializer
      value-serializer: org.apache.kafka.common.serialization.StringSerializer
```

Produce a single record to `demo.topic`.

**On spring-kafka 3.3.13** — the record flows main → retry → DLT, and the DLT is consumed cleanly:

```
RetryTopicConfigurer : Received message in dlt listener: demo.topic-dlt-0@0
```

**On spring-kafka 4.0.4** — same flow, but the DLT consumer fails and loops:

```
DeadLetterPublishingRecovererFactory : Record: topic = demo.topic-dlt, partition = 0, offset = 0, main topic = demo.topic threw an error at topic demo.topic-dlt and won't be retried. Sending to DLT with name demo.topic-dlt.
...
Caused by: java.lang.IllegalStateException: No Acknowledgment available as an argument, the listener container must have a MANUAL AckMode to populate the Acknowledgment.
```

The only thing changed between the two runs is the spring-kafka version.

### Expected

The built-in DLT logging consumer should consume and log the record under the default ack mode, the same as on 3.3.x — without the user having to switch the container to a MANUAL ack mode or add a custom `@DltHandler`.

### Possible cause (a guess — I haven't confirmed it in a debugger)

The built-in DLT handler is `RetryTopicConfigurer.LoggingDltListenerHandlerMethod.logMessage(Object, @NonNull Acknowledgment)`.

`MessagingMessageListenerAdapter.determineInferredType` decides whether to substitute a no-op `Acknowledgment` when the container doesn't provide one, based on:

```java
this.noOpAck |= methodParameter.getParameterAnnotation(NonNull.class) != null;
```

In 4.0 that `@NonNull` is `org.jspecify.annotations.NonNull`, which is `@Target(TYPE_USE)`. `MethodParameter.getParameterAnnotation(...)` returns only *declaration* annotations, not `TYPE_USE` ones, so it returns `null`, `noOpAck` stays `false`, and no no-op ack is substituted — the null ack then reaches the handler and the check throws. In 3.3.x the same parameter used `org.springframework.lang.NonNull` (a declaration annotation), which `getParameterAnnotation` did see. That lines up with the version where it started failing, but I haven't verified it's the real trigger.

abhimoondra:
Fixes #4468.

The built-in DLT logging handler `LoggingDltListenerHandlerMethod.logMessage(Object, @NonNull Acknowledgment)` stopped working under a non-manual ack mode on 4.0. Its `@NonNull` is now the JSpecify `@NonNull` (a `TYPE_USE` annotation), which `MethodParameter.getParameterAnnotation()` doesn't return, so `noOpAck` stayed `false` and the DLT consumer failed with `No Acknowledgment available ... MANUAL AckMode` and re-published the record to the same DLT in a loop.

Switched the check in `MessagingMessageListenerAdapter.determineInferredType` to `Nullness.forMethodParameter()` (per the suggestion on the issue), which understands `TYPE_USE` and `@NullMarked` nullness.

Also added a test covering the non-null `Acknowledgment` parameter case, which wasn't covered before.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java
@@ -40,11 +40,11 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.ShareConsumer;
 import org.apache.kafka.common.TopicPartition;
-import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 import reactor.core.publisher.Mono;
 
 import org.springframework.core.MethodParameter;
+import org.springframework.core.Nullness;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.expression.BeanResolver;
 import org.springframework.expression.Expression;
@@ -98,6 +98,7 @@ import org.springframework.util.TypeUtils;
  * @author Huijin Hong
  * @author Soby Chacko
  * @author Sanghyeok An
+ * @author Abhishek Moondra
  */
 public abstract class MessagingMessageListenerAdapter<K, V> implements ConsumerSeekAware, AsyncRepliesAware {
 
@@ -883,7 +884,7 @@ public abstract class MessagingMessageListenerAdapter<K, V> implements ConsumerS
 			}
 			this.hasAckParameter |= isAck;
 			if (isAck) {
-				this.noOpAck |= methodParameter.getParameterAnnotation(NonNull.class) != null;
+				this.noOpAck |= Nullness.forMethodParameter(methodParameter) == Nullness.NON_NULL;
 			}
 			isNotConvertible |= isAck;
 			boolean isConsumer = parameterIsType(parameterType, Consumer.class);
--- a/spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/annotation/EnableKafkaIntegrationTests.java
@@ -184,6 +184,7 @@ import static org.mockito.Mockito.spy;
  * @author Wang Zhiyang
  * @author Borahm Lee
  * @author Sean Sullivan
+ * @author Abhishek Moondra
  */
 @SpringJUnitConfig
 @DirtiesContext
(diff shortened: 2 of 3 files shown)


artembilan (changes requested):
LGTM!

Thank you!

abhimoondra:
Done. Added my name to the @author list of the affected classes.

One thing — `testBadAckConfig` started failing with the change. Its listener's `Acknowledgment` param is unannotated in a `@NullMarked` package, so Nullness now sees it as non-null and substitutes the no-op ack instead of failing. Marked that param (and the one in `testMissingAck`) `@Nullable` so they still hit the missing-ack path. Lmk if you'd rather handle it differently.

PR #4501, opened 2026-06-11
Allow extending Context outside package 

ca-stefan-cordes:
When implementing the KafkaConsumerBackoffManager and need to extend the Context (e.g. to add Headers to it), the "extends" needs to be done in package org.springframework.kafka.listener 
due to "package visibility" of class Context.

Please change to "protected" or better to "public" as interface KafkaConsumerBackoffManager is public, too.


sobychacko:
Thanks for the report. The use case makes sense, but the fix doesn't quite address the root cause. Context is already implicitly public since it's a member type inside a public interface, and making it explicitly public is a no-op.

The actual barrier to subclassing from outside the package is the package-private constructor. If you want to extend `Context` externally, the constructor would need to be protected. That said, `createContext()` is already a default method you can override to return a custom subtype — can you share more about your specific use case? Depending on what you're trying to do, there may already be a way to accomplish it without changes to the framework. 

Closing the PR for the time being. Feel free to re-open if need be with more information about your use case. Thanks!

PR #4505, opened 2026-06-13
GH-4504: Fix silent record loss with two failing async records on one partition

### LINKED ISSUE #4504: Suspend `@KafkaListener` with two failing records on the same partition skips one of them entirely (never retried, never recovered)

`4.1.0` (with Spring Boot `4.1.0`) — i.e. **including** the fix for [#4465](https://github.com/spring-projects/spring-kafka/issues/4465) from [PR #4469](https://github.com/spring-projects/spring-kafka/pull/4469).

**Describe the bug**

[PR #4469](https://github.com/spring-projects/spring-kafka/pull/4469) fixed the unbounded re-delivery for a **single** failing record, but the retry behaviour is still wrong when **two records on the same partition are failing at the same time**.

With a `suspend` `@KafkaListener`, a `DefaultErrorHandler(FixedBackOff(100L, 2L))`, and **two** always-failing records `r1` (offset 0) and `r2` (offset 1) sent back-to-back to the same single-partition topic:

- `r1` is delivered **once**, is never retried, and is **never handed to the recoverer** — with a `DeadLetterPublishingRecoverer` it would never reach the DLT. Once `r2` is recovered, the committed offset moves past `r1`, so the record is **silently lost**.
- `r2` is delivered 3 times and recovered once (as expected).

The identical scenario with a **blocking** listener behaves correctly: each record is delivered 3 times and recovered exactly once, in offset order.

The result is deterministic across runs.

**To Reproduce**

Minimal project: **https://github.com/lobodpav/spring-kafka-retry-issue** (the same one as for [#4465](https://github.com/spring-projects/spring-kafka/issues/4465), extended with the two-record scenarios).

Run `./gradlew test`:

- `Suspend listener with two failing records on the same partition retries and recovers each like the blocking listener` — **fails** as described above.
- `Blocking listener with two failing records on the same partition retries and recovers each in offset order` — passes (control case).
- `Single failing record is delivered initial + retries times for both listener kinds` — passes (confirms the #4465 fix).

**Expected behavior**

The `suspend` listener should behave like the blocking one: every failing record is delivered `n + 1` times and then recovered exactly once; no record is skipped without recovery.

**Actual behavior**

```
r1 deliveries ==> expected: <3> but was: <1>
r1 recoveries ==> expected: <1> but was: <null>
r2 deliveries: 3 (OK)
r2 recoveries: 1 (OK)
```

The test also reads the group's committed offset straight from the broker (`AdminClient.listConsumerGroupOffsets`) after the dust settles: it is `2`, past **both** records. So the skipped `r1` is not even re-delivered after a consumer restart — the record is permanently lost.

**Sample**

The same [repository](https://github.com/lobodpav/spring-kafka-retry-issue) as for https://github.com/spring-projects/spring-kafka/issues/4465, updated to Spring Boot 4.1.0 and extended with the two-record tests.

BK202503:
Fixes #4504.

> This PR now contains the **data-loss fix only**, per the maintainer's request to split (https://github.com/spring-projects/spring-kafka/pull/4505#issuecomment-4774232096). The bounded-amplification follow-up has been moved to a separate PR: #4512.

## Summary
- `KafkaMessageListenerContainer$ListenerConsumer#handleAsyncFailure` drained `failedRecords` in arrival order and handed each tuple to `commonErrorHandler.handleRemaining(rte, List.of(cRecord), ...)` individually. With a seek-after-handling handler (e.g. `DefaultErrorHandler`) and two failing records on the same partition in the same poll, each call registered its own seek — and the later call's seek to the higher offset clobbered the earlier call's seek to the lower offset before the next poll. The earlier-offset record was never re-delivered, never reached the recoverer, and the committed offset advanced past it once the later record was recovered, so the record was silently lost.
- Sort the claimed failures by `(topic, partition, offset)` and, once a `RecordInRetryException` has been thrown for the lowest offset on a partition, drop the remaining claimed tuples on that partition from this iteration. The seek that the error handler already registered will re-poll those records on the next loop iteration, the async failure callback will re-add them, and the next handleAsyncFailure call will process them again — there is no re-queueing, so the GH-4465 fix is preserved, and there is no per-record seek clobbering, so the earlier-offset record is no longer silently skipped.

## Problem
With a suspend `@KafkaListener`, `DefaultErrorHandler(FixedBackOff(100L, 2L))`, and two always-failing records `r1` (offset 0) and `r2` (offset 1) on the same single-partition topic:

- `r1` is delivered once, never retried, never handed to the recoverer — with a `DeadLetterPublishingRecoverer` it would never reach the DLT. Once `r2` is recovered, the committed offset moves past `r1`, so the record is silently lost.
- `r2` is delivered three times and recovered once, as expected.

The blocking listener with the same configuration behaves correctly (each record delivered three times, each recovered once), because the sync throw aborts the batch iteration and only `r1`'s failure reaches `handleRemaining` in the first cycle, while the sync path already drains the iterator's remaining records into the seek list.

Reporter's repro: https://github.com/lobodpav/spring-kafka-retry-issue

## Fix
In `handleAsyncFailure`:

1. Snapshot `failedRecords` and sort the snapshot by `(topic, partition, offset)` so that the lowest offset on a partition is handed to the error handler first.
2. Maintain a per-iteration `Set<TopicPartition> partitionsInRetry` and, before calling the error handler for a record, check whether its partition already has an in-flight retry seek registered from this iteration. If so, claim the tuple from the deque and `continue` without invoking the handler — the seek will re-poll the record, the async failure callback will re-add it, and a future handleAsyncFailure call will process it.
3. On `RecordInRetryException`, add the partition to `partitionsInRetry`. The `FailedRecordTracker` entry keyed by `(topic, partition, offset)` continues to advance on subsequent loop iterations via the natural seek-induced re-delivery, so the attempt counter is correct without re-queueing the tuple (which would re-introduce the unbounded re-delivery loop fixed by GH-4465).

The non-seeking handler path (e.g. `@RetryableTopic`) is unaffected: those calls never throw `RecordInRetryException`, `partitionsInRetry` stays empty, and every record is processed individually as before.

## Test plan
- [x] New regression test `EnableKafkaKotlinCoroutinesTests#test suspend function bounded retries for two records on the same partition`. Sends two always-failing records to a suspend `@KafkaListener` backed by `DefaultErrorHandler(FixedBackOff(100L, 2L))` and asserts both records are handed to the recoverer exactly once. Without the fix the earlier record is delivered once and the recovery latch never reaches zero.
- [x] Existing GH-4465 regression test (`test suspend function bounded retries with CommonErrorHandler`) still passes — the single-record case sees an empty `partitionsInRetry` and dispatches identically to the previous behavior.
- [x] `./gradlew :spring-kafka:test --tests "*listener*" --tests "*Async*RetryTopic*"` is green, including the `Async{CompletableFuture,Mono}RetryTopicScenarioTests` scenarios that the first version of this PR regressed.
- [x] Verified against the reporter's reproducer (https://github.com/lobodpav/spring-kafka-retry-issue) with this snapshot built locally and Testcontainers Kafka — recovery / committed-offset assertions pass (`r1 deliveries = 3`, `r1 recoveries = 1`, `r2 recoveries = 1`, committed offset `= 2`).

## Scope and trade-off
This change prevents data loss by ensuring that failed async records from the same partition are handled in offset order.

With suspend listeners and seek-based `DefaultErrorHandler` semantics, later records may still be re-delivered more than `n + 1` times when an earlier record in the same partition fails. This preserves recovery and committed-offset correctness, but it does not make suspend listener retry counts identical to blocking listeners.

For strict per-record retry counts with asynchronous listener execution, retry-topic based non-blocking retry (`@RetryableTopic` / `RetryTopicConfiguration`) is the better semantic model. The `async-returns.adoc` IMPORTANT block in this PR calls this out and points readers at `@RetryableTopic` (non-blocking retry) or an idempotent listener body when the per-delivery work is non-idempotent. The same block also notes that the choice between a seek-based handler and `@RetryableTopic` can be made per topic rather than application-wide.

### Why `r2` is delivered five times in the repro

The async dispatch model invokes the listener on every polled record before any async failure callback fires. The blocking listener's sync throw aborts the batch iteration so a later record is never invoked while an earlier record is being retried; the async path cannot replicate that without giving up the parallelism that is the point of suspend / `Mono` / `CompletableFuture` listeners.

`r2`'s five listener invocations break down as follows (tracker = `FailedRecordTracker` state for `r2`):

| Iteration | Event | `r2` invocation count | `r2` tracker state |
|---|---|---|---|
| 0 | Initial poll | 1 | no entry (incidental redelivery during r1's retry window) |
| 1 | r1 retry #1 (r2 dropped from queue while r1's partition is in retry) | 2 | no entry  re-polled by seek, not processed |
| 2 | r1 retry #2 | 3 | no entry — re-polled by seek, not processed |
| 3 | r1 recovered, r2 polled at its own offset | 4 | new entry created (attempt = 0) |
| 4 | r2 retry #1 | 5 | attempt = 1 → STOP on iter 5's `handleAsyncFailure`, recover |

`FailedRecordTracker` counts exactly three retry attempts for `r2` (iter 3, 4, 5), which is the `n + 1` budget. The two additional listener invocations in iter 0, 1, 2 are at-least-once re-deliveries from polls that re-fetched `r2` alongside `r1` during `r1`'s retry cycle; the tracker does not see them. Listeners that need exact-once-per-attempt semantics should be idempotent (the Kafka default expectation) or use `@RetryableTopic` so failed records do not block the rest of the partition. The follow-up PR bounds this further , see #4512.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -1549,7 +1549,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 					continue;
 				}
 				try {
-					failedRecord.observation.scoped(() -> invokeErrorHandlerForFailedRecords(List.of(failedRecord)));
+					failedRecord.observation.scoped(() -> invokeErrorHandlerBySingleRecord(failedRecord));
 				}
 				catch (RecordInRetryException e) {
 					partitionsInRetry.add(tp);
@@ -3061,9 +3061,9 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			}
 		}
 
-		private void invokeErrorHandlerForFailedRecords(List<FailedRecordTuple<K, V>> partitionFailures) {
-			FailedRecordTuple<K, V> head = partitionFailures.get(0);
-			RuntimeException rte = head.ex;
+		private void invokeErrorHandlerBySingleRecord(FailedRecordTuple<K, V> failedRecordTuple) {
+			final ConsumerRecord<K, V> cRecord = failedRecordTuple.record;
+			RuntimeException rte = failedRecordTuple.ex;
 			if (Objects.requireNonNull(this.commonErrorHandler).seeksAfterHandling() || rte instanceof CommitFailedException) {
 				try {
 					if (this.producer == null) {
@@ -3073,10 +3073,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 				catch (Exception ex) { // NO SONAR
 					this.logger.error(ex, "Failed to commit before handling error");
 				}
-				List<ConsumerRecord<?, ?>> retryRecords = new ArrayList<>(partitionFailures.size());
-				for (FailedRecordTuple<K, V> tuple : partitionFailures) {
-					retryRecords.add(tuple.record);
-				}
+				List<ConsumerRecord<?, ?>> retryRecords = List.of(cRecord);
 				try {
 					this.commonErrorHandler.handleRemaining(rte, retryRecords, this.consumer,
 							KafkaMessageListenerContainer.this.thisOrParentContainer);
@@ -3087,24 +3084,18 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 				}
 			}
 			else {
-				// Non-seeking handler: handleOne does not seek, so there is no risk of a
-				// later record's call clobbering an earlier one. Process each tuple
-				// independently and aggregate any unhandled records into remainingRecords.
+				boolean handled = false;
+				try {
+					handled = this.commonErrorHandler.handleOne(rte, cRecord, this.consumer,
+							KafkaMessageListenerContainer.this.thisOrParentContainer);
+				}
+				catch (Exception ex) {
+					this.logger.error(ex, "ErrorHandler threw unexpected exception");
+				}
 				Map<TopicPartition, List<ConsumerRecord<K, V>>> records = new LinkedHashMap<>();
-				for (FailedRecordTuple<K, V> tuple : partitionFailures) {
-					ConsumerRecord<K, V> cRecord = tuple.record;
-					boolean handled = false;
-					try {
-						handled = this.commonErrorHandler.handleOne(tuple.ex, cRecord, this.consumer,
-								KafkaMessageListenerContainer.this.thisOrParentContainer);
-					}
-					catch (Exception ex) {
-						this.logger.error(ex, "ErrorHandler threw unexpected exception");
-					}
-					if (!handled) {
-						records.computeIfAbsent(new TopicPartition(cRecord.topic(), cRecord.partition()),
-								tp -> new ArrayList<>()).add(cRecord);
-					}
+				if (!handled) {
+					records.computeIfAbsent(new TopicPartition(cRecord.topic(), cRecord.partition()),
+							tp -> new ArrayList<>()).add(cRecord);
 				}
 				if (!records.isEmpty()) {
 					this.remainingRecords = new ConsumerRecords<>(records, Map.of());
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -868,6 +868,16 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 
 		private final ConcurrentLinkedDeque<FailedRecordTuple<K, V>> failedRecords = new ConcurrentLinkedDeque<>();
 
+		// Lowest in-flight retry offset per partition for the async path. Populated when
+		// a seek-after-handling error handler seeks back without recovering (or when an
+		// async listener's failure callback fires synchronously within the dispatch
+		// loop). Consulted by doInvokeWithRecords to skip listener invocation for
+		// records on the same partition at higher offsets — those records will be
+		// re-fetched by the next poll after the retry resolves, so re-invoking the
+		// listener on them in the meantime only inflates the delivery count (GH-4504,
+		// head-of-line amplification). Accessed only on the consumer thread.
+		private final Map<TopicPartition, Long> asyncRetryOffsets = new HashMap<>();
+
 		private boolean isListenerAdapterObservationAware = false;
 
 		@SuppressWarnings(UNCHECKED)
@@ -1517,14 +1527,11 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			if (failedRecordsSnapshot.isEmpty()) {
 				return;
 			}
-			// Sort by (topic, partition, offset) so that for a seek-after-handling error
-			// handler we always hand the lowest offset on a partition over first; this
-			// way, when the handler registers a seek-back, any subsequent records on the
-			// same partition can be dropped from this iteration and rely on the seek to
-			// re-fetch them, avoiding the per-record seek clobbering that silently
-			// skipped the earlier-offset record (GH-4504). For non-seeking handlers
-			// (e.g. RetryTopic) the sort is harmless and each record is processed
-			// individually to keep the existing dispatch semantics.
+			// Sort by (topic, partition, offset) so per-partition processing handles the
+			// lowest offset first. For a seek-after-handling error handler this is what
+			// avoids the per-record seek clobbering that silently skipped the earlier-
+			// offset record (GH-4504). For non-seeking handlers (e.g. RetryTopic) the
+			// sort is harmless.
 			failedRecordsSnapshot.sort(Comparator
 					.comparing((FailedRecordTuple<K, V> t) -> t.record.topic())
 					.thenComparingInt(t -> t.record.partition())
@@ -1548,11 +1555,13 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 					// already been performed by the error handler).
 					continue;
 				}
+				long retryOffsetBefore = this.asyncRetryOffsets.getOrDefault(tp, -1L);
 				try {
 					failedRecord.observation.scoped(() -> invokeErrorHandlerForFailedRecords(List.of(failedRecord)));
 				}
 				catch (RecordInRetryException e) {
 					partitionsInRetry.add(tp);
+					this.asyncRetryOffsets.put(tp, failedRecord.record.offset());
 					// Retry is in progress: the seek-after-handling error handler has
 					// repositioned the consumer to this offset and the
 					// FailedRecordTracker entry keyed by topic-partition-offset persists
@@ -1561,6 +1570,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 					// processed twice per loop (once from the queue, once from the
 					// seek-induced re-delivery) — the unbounded re-delivery reported in
 					// GH-4465.
+					continue;
 				}
 				catch (Exception e) {
 					this.logger.warn(() ->
@@ -1568,7 +1578,41 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 									+ failedRecord
 									+ ", Exception : "
 									+ e.getMessage());
+					this.asyncRetryOffsets.remove(tp);
+					continue;
 				}
+				// handleRemaining returned normally (the record was recovered or the
+				// non-seeking handler routed it elsewhere). If we were in a multi-
+				// attempt retry on this partition, the recovered head's doSeeks call
+				// registered no seek (only the head was in the list and it was skipped
+				// as recovered) — manually advance the consumer past it so the next
+				// poll fetches the next still-failing record on the partition instead
+				// of re-fetching the recovered head and starting a fresh retry cycle on
+				// it (GH-4504, head-of-line amplification — without this the suspend
+				// listener invocation count grows quadratically with the burst size).
+				if (retryOffsetBefore == failedRecord.record.offset()) {
+					try {
+						this.consumer.seek(tp, failedRecord.record.offset() + 1);
+					}
+					catch (Exception ex) {
+						this.logger.error(ex, () -> "Failed to advance past recovered offset for " + tp);
+					}
+					// The polled records on this partition that were skipped while it was
+					// in retry still have entries in offsetsInThisBatch (each polled
+					// record is registered for an out-of-commit container regardless of
+					// whether the listener was invoked). Clear them so
+					// pauseConsumerIfNecessary does not keep the consumer paused waiting
+					// for async acks that will never arrive — those records have been
+					// re-fetched-and-skipped, never acknowledged, and the next poll
+					// (after the manual seek above) will re-capture them.
+					if (this.offsetsInThisBatch != null) {
+						this.offsetsInThisBatch.remove(tp);
+					}
+					if (this.deferredOffsets != null) {
+						this.deferredOffsets.remove(tp);
+					}
+				}
+				this.asyncRetryOffsets.remove(tp);
 			}
 		}
 
@@ -2773,8 +2817,36 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 				if (cRecord == null) {
 					continue;
 				}
+				if (shouldSkipForAsyncRetry(cRecord)) {
+					// A lower-offset record on this partition is currently being retried
+					// by a seek-after-handling error handler. Skip listener invocation for
+					// this record — the seek for the in-flight retry will re-fetch it on
+					// the next poll, and re-invoking the listener on it now only inflates
+					// the delivery count (GH-4504, head-of-line amplification). Also drop
+					// the record from offsetsInThisBatch: for an out-of-commit container
+					// (manual ack + async replies) the poll registered the offset there
+					// to wait for an async ack that will never arrive (we are not
+					// invoking the listener), and a non-empty offsetsInThisBatch would
+					// keep the consumer paused.
+					removeOffsetsInBatch(List.of(cRecord));
+					continue;
+				}
+				int failedBefore = this.failedRecords.size();
 				this.logger.trace(() -> "Processing " + KafkaUtils.format(cRecord));
 				doInvokeRecordListener(cRecord, iterator);
+				if (this.commonErrorHandler != null && this.commonErrorHandler.seeksAfterHandling()
+						&& this.failedRecords.size() > failedBefore) {
+					// The async listener's failure callback fired synchronously while
+					// invoking this record (typical for suspend functions that throw or
+					// Mono/CompletableFuture pre-completed with an error). Mark the
+					// partition as in-retry so the subsequent records in this batch are
+					// skipped by the check above instead of being invoked once each
+					// before the next handleAsyncFailure registers the retry. putIfAbsent
+					// preserves any lower-offset retry that is already in flight on this
+					// partition (GH-4504).
+					this.asyncRetryOffsets.putIfAbsent(
+							new TopicPartition(cRecord.topic(), cRecord.partition()), cRecord.offset());
+				}
 				if (this.commonRecordInterceptor != null) {
 					this.commonRecordInterceptor.afterRecord(cRecord, this.consumer);
 				}
@@ -2786,6 +2858,14 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 					break;
 				}
 			}
+		}
+
+		private boolean shouldSkipForAsyncRetry(ConsumerRecord<K, V> cRecord) {
+			if (this.asyncRetryOffsets.isEmpty()) {
+				return false;
+			}
+			Long retryOffset = this.asyncRetryOffsets.get(new TopicPartition(cRecord.topic(), cRecord.partition()));
+			return retryOffset != null && cRecord.offset() > retryOffset;
 		}
 
 		private boolean checkImmediatePause(Iterator<ConsumerRecord<K, V>> iterator) {
@@ -3080,6 +3160,13 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 				try {
 					this.commonErrorHandler.handleRemaining(rte, retryRecords, this.consumer,
 							KafkaMessageListenerContainer.this.thisOrParentContainer);
+					// Recovery succeeded. Clear the offsets from offsetsInThisBatch so
+					// pauseConsumerIfNecessary does not keep the consumer paused waiting
+					// for async acks on the recovered record. Without this, an out-of-
+					// commit container (manual ack + async replies, which Kotlin suspend
+					// listeners fall into) stays paused forever after recovery on the
+					// partition.
+					removeOffsetsInBatch(retryRecords);
 				}
 				catch (RecordInRetryException e) {
 					removeOffsetsInBatch(retryRecords);
@@ -3103,7 +3190,7 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 					}
 					if (!handled) {
 						records.computeIfAbsent(new TopicPartition(cRecord.topic(), cRecord.partition()),
-								tp -> new ArrayList<>()).add(cRecord);
+								k -> new ArrayList<>()).add(cRecord);
 					}
 				}
 				if (!records.isEmpty()) {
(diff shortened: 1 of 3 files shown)


lobodpav:
Thanks for tackling this. I built this PR locally and ran my reproducer against it: it does fix the silent data loss. With `N` always-failing records on one partition, every record now reaches the recoverer, and the committed offset is correct.

I'd like to surface one residual issue before it merges, because this changes the failure *mode* rather than fully resolving it: later records on a partition are now re-executed many times while an earlier record is still being retried, and the count isn't constant — it grows quadratically with the number of records queued behind the failing one.

Since retries are seek-based, each retry of the in-flight record re-fetches and re-invokes every not-yet-recovered record behind it on the same partition. Building this PR and running my reproducer against it — `DefaultErrorHandler(FixedBackOff(100ms, 2))`, `N` always-failing records on a single partition, counting total listener invocations:

| N | blocking listener | suspend listener |
|---|---|---|
| 5 | 15 | 35 |
| 10 | 30 | 120 |
| 15 | 45 | 255 |
| 20 | 60 | 440 |

The blocking listener is linear: `N·(maxAttempts+1)`. The suspend listener fits `N·(N+2)` exactly in this configuration — quadratic — so the amplification `(N+2)/(maxAttempts+1)` grows without bound as the backlog deepens (4× at N=10, 7× at N=20).

For a listener whose body does real work per delivery (a DB write, an outbound call, etc.), a burst of poison pills therefore becomes a quadratic storm of that work, even though every record is ultimately recovered. That reads less like inefficiency and more like an availability concern under bad input.

My reproducer is updated with new tests for this — a poison-pill burst test (asserting the blocking listener's bounded `N·(maxAttempts+1)`) plus the original single-record and two-record cases. It targets released 4.1.0, where the suspend scenarios fail on the silent loss this PR fixes; I pointed the same burst test at a local build of this PR to produce the numbers above, where it instead exceeds the bounded budget (120 vs 30 invocations at N=10): https://github.com/lobodpav/spring-kafka-retry-issue

BK202503:
Thanks for the data  the `N * (N + 2)` measurement was very useful. Pushed an update (`48d486bc`) that brings the suspend listener down to `N * (n + 1)` to match the blocking listener.

The previous commit sorted failures by `(topic, partition, offset)` and dropped subsequent failures on a partition once the head entered retry, which was enough to repair the data loss but left the consumer re-polling the whole backlog on every retry cycle of the in-flight head. The update adds an `asyncRetryOffsets` map keyed by partition and uses it in three places:

- `doInvokeWithRecords` skips listener invocation for records whose offset is greater than the in-flight retry offset on their partition; the skipped records are also removed from `offsetsInThisBatch` so an out-of-commit container (manual ack + async replies, which Kotlin suspend listeners fall into) doesn't stay paused on async acks that will never arrive.
- A synchronous failure observed during dispatch records the partition's offset via `putIfAbsent`, so the rest of the polled batch on the same partition is skipped in the same loop iteration. Without this the first poll always invokes every failing record once before any retry state is registered.
- When `handleRemaining` returns normally for a previously-in-retry partition, the consumer is manually seeked to `recordOffset + 1` and that partition's entries are cleared from `offsetsInThisBatch` / `deferredOffsets`. Without this the recovered head's `doSeeks` call (single record, recovered = skipped) registers no seek and the consumer would re-fetch the recovered head and start a fresh retry cycle on it.

Built the snapshot locally and ran your updated reproducer against it on Testcontainers Kafka: the `Poison-pill burst - suspend listener` test now passes with `N * (n + 1) = 30` invocations for `BURST_SIZE = 10`. The single-record and two-record tests pass as well.

The non-seeking `@RetryableTopic` path is unchanged (calls don't throw `RecordInRetryException`, so `asyncRetryOffsets` is only ever populated by the dispatch-loop sync detection on a seek handler), and the existing `Async*RetryTopicScenarioTests` continue to pass on this branch.

BK202503:
Update: the linear-bound attempt (`48d486bc`) was reverted in `669a66e4` because the dispatch-loop skip + manual-seek-on-recovery interacted with the existing `Async{CompletableFuture,Mono}RetryTopicScenarioTests` `allFailCaseTest` and `oneLongSuccessMsgBetween49ShortFailMsg` started timing out on the recovery latch. The shape of the conflict: `DeadLetterPublishingRecoverer` is an immediately-recovering recoverer used under a seek-after-handling `DefaultErrorHandler`, so the post-recovery manual seek that the multi-attempt path needs to advance past a recovered head fires there too, and re-seeking the consumer back to a record whose listener had already routed it to a retry topic produces re-deliveries or in-flight ack tracking that the RetryTopic scenarios do not expect.

This branch is now back at the previous version: the data loss is fixed, the burst case is still bounded (the head-of-line amplification you measured remains, the suspend listener still over-delivers later records by `(N+2)/(n+1)`), and the existing RetryTopic scenarios continue to pass. The PR body has been restored to the previous wording with a note on the reverted attempt.

I think a clean fix for the head-of-line amplification is doable, but it likely needs a different signal than "this handler seeks after handling" to decide whether to skip subsequent records on the partition something that distinguishes "the recoverer is the immediate publish-and-done kind" from "the tracker is going to use its backoff budget". Happy to pursue that as a follow-up once the data-loss fix is in.

lobodpav:
Thank you for the explanation - I understand the amplification needs to be addressed separately, and landing the data-loss fix first is the right call.

It does mean we can't adopt `suspend` listeners yet, though: our listeners make a REST call per record, so under a poison-pill burst, the quadratic re-execution would be effectively a self-inflicted DoS on our own downstream services. So the bounded follow-up is what would unblock `suspend` for us - happy to validate it against the reproducer's burst test whenever you pick it up.

In the meantime, it would be great to call this out in the reference docs: a short caveat that, with a seek-based error handler (e.g. `DefaultErrorHandler`), a `suspend/async` listener can re-execute the records queued behind a failing one on the same partition — and that this grows with backlog depth — so listeners should be idempotent (or use `@RetryableTopic`) where that matters. That would save others from hitting it unknowingly until the follow-up lands.

BK202503:
Added , `1be0fc83`. `IMPORTANT` block in `async-returns.adoc` covers the seek-based re-execution and points at `@RetryableTopic` or idempotent listener for non-idempotent per-record work.

> happy to validate it against the reproducer's burst test whenever you pick it up

Will run it on the next pass at the amplification fix and ping you before pushing.

BK202503:
Small docs follow-up on the `IMPORTANT` block: `1a3a310e` makes explicit that the choice between a seek-based handler and `@RetryableTopic` is per topic rather than application-wide.

The framing: most non-trivial Kafka workloads carry a mix of ordering requirements.

- State transitions keyed by an entity id (settlement, order, inventory) need strict per-partition ordering and stay on a seek-based handler. Throughput there comes from partition count and key distribution, not from parallelizing inside one partition.
- Events that do not require ordering (search index updates, notifications, downstream enrichment) can run on `@RetryableTopic`, where a failing record does not re-execute its partition peers.

This does not change behavior — the bounded-amplification follow-up is still what unblocks `suspend` for the first group, where the listener body cannot be made idempotent. The framing only narrows the population that depends on it.

BK202503:
Pushed `94af7042`  bounded amplification follow-up. The previous attempt at this (`48d486bc`, reverted in `669a66e4`) interacted with the `Async{CompletableFuture,Mono}RetryTopicScenarioTests` because it wiped `offsetsInThisBatch[tp]` wholesale on recovery, which collided with `DeadLetterPublishingRecoverer`'s ack tracking on those tests. This version drops the wholesale clear and uses `seenMultiAttemptRetry` (first `RecordInRetryException` flips the gate) to keep publish-and-done recoverers out of the dispatch-loop sync-failure path entirely. Both `allFailCaseTest` and `oneLongSuccessMsgBetween49ShortFailMsg` stay green, and the full `:spring-kafka:test` sweep passes locally.

Also pushed `fbc057c6`  a 10-record poison-pill burst regression test (`EnableKafkaKotlinCoroutinesTests#test suspend function bounded burst stays linear with N`) that asserts the total listener invocations equal `N * (n + 2) - 1` (= 39 for N = 10, n = 2). The same formula was verified locally at N = 5 (= 19) and N = 20 (= 79).

Burst numbers on the reproducer (`DefaultErrorHandler(FixedBackOff(100ms, 2))`, `N` always-failing records, one partition; ✓ = locally measured, ✱ = predicted by formula at adjacent measured N):

| N | blocking (`N * (n + 1)`) | suspend, this PR | suspend, previous commit on this branch (lobodpav's measurements) |
|---|---|---|---|
| 5 | 15 | 19 ✓ | 35 |
| 10 | 30 | 39 ✓ | 120 |
| 15 | 45 | 59 ✱ | 255 |
| 20 | 60 | 79 ✓ | 440 |

Linear in `N` at `N * (n + 2) - 1`. The `+ 1` extra delivery per non-head record is the iter-0 dispatch invocation that fires before `seenMultiAttemptRetry` flips on the first `RecordInRetryException` - after that, subsequent retry iterations of the head skip the records queued behind it via `asyncRetryOffsets`. Eliding the iter-0 invocation would require either configuration introspection on the `BackOff` or giving up per-poll parallelism, so I held the line at "one extra delivery per record" rather than chase strict parity with the blocking listener.

@lobodpav happy for you to validate against the burst test in the reproducer when you have a chance — the PR body has the updated test plan and the docs caveat in `async-returns.adoc` was updated to describe the bounded behavior and the `N * (n + 2) - 1` formula.

sobychacko:
@BK202503 Thanks for digging into this one - the PR and the analysis above are solid, and the root cause is clear: the per-record seeks stepping on each other on the same partition is definitely an issue.

The diff isn't huge, but it's really two changes bundled together, and they have pretty different risk profiles. Let me try to paraphrase them based on the descriptions you provided. 

  1. The actual data-loss fix -- sort by offset and drop the same-partition records once the head is in retry. This one stands on its own, and since silent record loss is the kind of thing we want out the door quickly (and backported), I'd like to keep it small and focused.
  2. The follow-up that keeps bursts from blowing up. Once the data-loss fix is in, there's a side effect: while the first record is stuck retrying, the consumer keeps re-reading the records behind it on the same partition every retry cycle, and re-runs the listener on each of them every time. With a handful of records, that's no big deal, but with a burst of failing records, it snowballs and becomes quadratic, as you have shown above. The fix you added stops the listener from running on those queued-up records while the one ahead of them is still retrying, and then seeks past the first record once it's recovered, so the rest get picked up. That's the trickier piece, since it deals with offset-tracking, manual seek, etc. I'd rather review it on its own, especially since you had to revert your original commit and rework. 

How about splitting this into two PRs? The first one carries the data-loss fix plus the two-record regression test. The second one builds the amplification bound plus the burst test on top of it. Easier to reason about and review each in isolation, and it keeps the history clean.

To be clear, this isn't about deprioritizing the amplification work. I know from the thread that's the part that actually unblocks suspend listeners for the per-record case that @lobodpav mentioned. Just want to decouple them so each gets the right level of review and scrutiny, and so we can land the correctness piece sooner.

Please let us know what you think. Thanks!

lobodpav:
@sobychacko Agree on the split.

@BK202503 I think (N·(n+2)−1) is the right place to stop. The concern I raised was the quadratic blow-up under a burst — that's gone.

One more thing worth flagging for anyone who lands here with a similar use case, because it caught me out:

Even with both fixes in, a `suspend`/async listener doesn't preserve per-partition (and therefore per-key) processing order once records on the same partition have mixed outcomes. Concretely:
1. r1 fails and enters its retry/backoff cycle
2. r2 (same partition) succeeds on first delivery
3. r2's processing has completed before r1's

Kafka still delivers in order, and the committed offset stays consistent, but the listener bodies run concurrently, so the side effects can land out of order. A blocking listener avoids this because it processes the partition strictly one record at a time — r1 to completion (success or recovery) before r2 is ever invoked.

So, for consumers that need ordered per-key processing, the blocking listener is the correct model regardless of this PR: [#4505](https://github.com/valka-ai/livedealer-transaction-service/issues/4505) fixes the data loss and bounds the amplification, but ordering loss is inherent to processing a partition concurrently, not something a retry fix can address. 

It might be worth a short note next to the `@RetryableTopic` routing guidance in `async-returns.adoc`, since it's easy to assume "same key → ordered" carries over to async listeners when it doesn't.

I'd missed this myself until I worked through the fail/succeed case.

BK202503:
Split done.

@sobychacko #4505 (this PR) now carries the data-loss fix + two-record regression test + the docs caveat. The bounded amplification follow-up is at #4512, opened against `main` and built on top of this branch; once this PR merges, the diff there will be just the two amplification commits (`f1d2b4b8` + `febfda8d`) plus the docs revision in `f1d2b4b8`.

@lobodpav Good catch on the mixed-outcome ordering — I'd missed that the per-key ordering guarantee doesn't carry over to async listeners even with both fixes in. Added a paragraph (`1cf5250d`) next to the `@RetryableTopic` routing guidance in `async-returns.adoc` using your framing, calling out that:

- the listener bodies on the same partition run concurrently, so a successful later-offset record can complete its side effects before an earlier-offset record that is still in its retry / backoff cycle finishes;
- Kafka still delivers in order and the committed offset stays consistent, but the side effects can land out of partition order;
- consumers that rely on per-key processing order (idempotent updates keyed by entity id, ordered state transitions) should stay on the blocking listener regardless of error handler.

The amplification fix in #4512 narrows the per-record dispatch-loop overhead but does not change this.  that's an inherent property of running a partition concurrently, which is what async listeners are.

The docs commit is on both branches (PR #4512 has been rebased on top of the updated PR #4505 tip), so the caveat ships with whichever PR lands first.

spring-kafka/src/test/kotlin/org/springframework/kafka/listener/EnableKafkaKotlinCoroutinesTests.kt
```diff
@@ -140,6 +142,45 @@ class EnableKafkaKotlinCoroutinesTests {
...
+				// The earlier-offset record is retried exactly n + 1 times.
+				assertThat(this.config.twoRecordRetryDeliveries["r1"]).isEqualTo(3)
+				// The later-offset record is delivered for each poll cycle that
+				// re-fetches it alongside r1 during r1's retry, plus its own
+				// n + 1 retry cycle once r1 is recovered. The async dispatch
+				// model invokes the listener on every polled record before any
+				// failure callback fires, so this is the bounded minimum given
+				// the blocking listener cannot be replicated without giving up
+				// per-poll parallelism.
+				assertThat(this.config.twoRecordRetryDeliveries["r2"]).isEqualTo(5)
```
sobychacko: I think this is flaky since `r2 == 5` assumes r1 and r2 land in the same first poll. If r1 gets polled alone first, this comes out to 4. Could we assert a bound instead (e.g. >= 3 and <= 5)? We just need to make sure that the records are recovered once, as that is the real proof of the fix anyway. What do you think?

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -3033,9 +3061,9 @@ private void doInvokeOnMessage(final ConsumerRecord<K, V> recordArg) {
 			}
 		}
 
-		private void invokeErrorHandlerBySingleRecord(FailedRecordTuple<K, V> failedRecordTuple) {
-			final ConsumerRecord<K, V> cRecord = failedRecordTuple.record;
-			RuntimeException rte = failedRecordTuple.ex;
+		private void invokeErrorHandlerForFailedRecords(List<FailedRecordTuple<K, V>> partitionFailures) {
```
sobychacko: `invokeErrorHandlerForFailedRecords` takes a `List<FailedRecordTuple>`, but the caller always passes a single-element `List.of(failedRecord)`. Since the sort-and-drop above means only the head ever reaches here, the list is always size 1, so the `get(0)` and the loops never really loop. It works, but the plural signature makes it look like it batches per-partition failures when it doesn't, and it touches the `RetryableTopic` branch for no real change here. How about keeping the original single-record method and just leaving the fix in handleAsyncFailure? Keeps the diff small and that path untouched. Thanks!

sobychacko (changes requested):

lobodpav:
One more note on the mixed-ordering case. Intuitively, I'd have expected a suspend listener to process records concurrently across partitions but sequentially within a partition — i.e. keep Kafka's per-partition ordering guarantee, just parallelised across partitions. Parallelising within a partition was the surprising part to me, since it quietly gives up Kafka's ordering guarantee.

A doc note helps, but it's still an easy trap to fall back into — I suspect I'd make the same assumption again a year from now once I've forgotten this thread 😅. So if "sequential within a partition, concurrent across partitions" were ever feasible as the default — or even an opt-in mode — it would match expectations far better than a caveat alone.

BK202503:
Done in `58f3f5c5`  restored `invokeErrorHandlerBySingleRecord(FailedRecordTuple)` so the RetryableTopic branch is untouched, and relaxed r2 to `isBetween(3, 5)`.
Full `:spring-kafka:test` sweep green; #4512 rebased on top.

BK202503:
@lobodpav Agreed, I think most users would reasonably expect that, and the current "concurrent within a partition" model is the surprising part rather than the docs caveat being missed.

For this PR I'd like to keep the scope on the silent data loss fix; partition-ordered async dispatch feels right as a separate issue / follow-up where it can be designed properly (opt-in mode, `AckMode` interaction, ordering semantics).

On the docs side I'll keep it explicit that a suspend / `Mono` / `CompletableFuture` listener can dispatch same-partition records concurrently today, and that callers need to plan for ordering / idempotency.

lobodpav:
> For this PR I'd like to keep the scope on the silent data loss fix; partition-ordered async dispatch feels right as a separate issue / follow-up where it can be designed properly (opt-in mode, `AckMode` interaction, ordering semantics).
> 
> On the docs side I'll keep it explicit that a suspend / `Mono` / `CompletableFuture` listener can dispatch same-partition records concurrently today, and that callers need to plan for ordering / idempotency.

Sounds great. Thanks!
Will you create the issue? Or shall I do that?

BK202503:
Opened #4513 to track it. Happy to take a first cut once this PR / #4512 land.

spring-kafka/src/test/kotlin/org/springframework/kafka/listener/EnableKafkaKotlinCoroutinesTests.kt
BK202503: @sobychacko please check updated commits.

sobychacko:
@BK202503 @lobodpav Thanks for the issue/PR and the great discussions here. I learned a lot myself during the review of the PR. I squashed the individual commits and merged the changes upstream (i will also backport to `4.0.x`). I made a slight mistake on the final merge by forgetting to add the commit signoff (I don't want to force-push it again so that we won't disrupt the commit history for anyone already fetched). In any case, the commit signatures are on your individual commits on the PR itself, which we can retrieve if there ever a need for that. Thanks again! I will get to the other PR's shortly. 

Update: Backported to `4.0.x` now. There were some conflicts which I had to manually address. Preserved the commit signature in `4.0.x` during the cherry-pick.

sobychacko:
@BK202503 Tests are failing on the `4.0.x` branch due to these changes. Could you take a look? If need be, we can back out these changes from `4.0.x`.

BK202503:
@sobychacko 
Reproduced locally on `4.0.x` (3727de52). All three failures are in `EnableKafkaKotlinCoroutinesTests` (`test suspend function with CommonErrorHandler`, `test suspend function bounded retries with CommonErrorHandler`, `test suspend function bounded retries for two records on the same partition`).

Root cause: the backport's three new suspend-listener tests depend on a `main`-only commit, `813f6891` (Support `CommonErrorHandler` for suspend functions), that was never carried to `4.0.x`. On `main`, `setCallbackForAsyncFailure(...)` is wired for any `RecordMessagingMessageListenerAdapter` via `DelegatingMessageListener` extraction. On `4.0.x`, it's only wired when the listener is a `KafkaBackoffAwareMessageListenerAdapter` (i.e., `@RetryableTopic` only). A plain `suspend @KafkaListener` + `DefaultErrorHandler` therefore never routes its async failure into `failedRecords`, so `handleAsyncFailure` never sees it, so the recoverer is never called.

Two ways to resolve this  (1) fix forward by cherry-picking `813f6891` to `4.0.x` so the backport works, or (2) back out the backport from `4.0.x`. Going with (1) since silent record loss is the kind of thing worth keeping fixed on a maintained branch.

Opened #4517 with the cherry-pick. Verified locally  `EnableKafkaKotlinCoroutinesTests`, `ObservationTests` (which `813f6891` also touches), and the full `:spring-kafka:test` sweep are all green on the cherry-pick. The other potential precursor, `dfe52787`, is effectively already on `4.0.x` because the backport (`ed31b093`) bundled both its container change (removal of `failedRecords.addLast(...)`) and its regression test, so it's not needed here.

Happy to defer if you'd rather back out the backport instead.

PR #4512, opened 2026-06-23
GH-4504: Bound async retry to linear delivery count per record

BK202503:
Builds on #4505 (data-loss fix). Per the maintainer's request to split #4505 into two PRs.

> This PR is the amplification follow-up to #4505 (merged). Now rebased onto current main — the diff is the two amplification commits (`d3dc946d` + `c8862adc`) plus a one-line `@author` attribution commit (`420cd00e`).

## Summary

After the data-loss fix in #4505 lands, both records reach the recoverer and the committed offset is correct — but @lobodpav measured a separate head-of-line amplification: with a burst of `N` always-failing records on one partition, the suspend listener total invocation count grows as `N * (N + 2)` (quadratic) vs the blocking listener's `N * (n + 1)`. The consumer keeps re-polling the whole backlog on every retry cycle of the in-flight head and the dispatch loop re-invokes each subsequent record on the partition every time.

This PR bounds that at `N * (n + 2) - 1` (linear) without re-introducing the regression that took the earlier attempt (`48d486bc`, reverted in `669a66e4`) out:

- Add `asyncRetryOffsets: Map<TopicPartition, Long>` populated only when the error handler throws `RecordInRetryException` (i.e., the tracker is going to use its backoff budget).
- Add a `seenMultiAttemptRetry` flag, set on the first `RecordInRetryException` and consulted before the dispatch-loop sync-failure detection populates `asyncRetryOffsets` for subsequent records in the same batch. The gate keeps publish-and-done recoverers (e.g. `DeadLetterPublishingRecoverer` wired into a `@RetryableTopic` flow) out of this path entirely — those calls never throw `RecordInRetryException`, so the gate stays false and subsequent records continue to be invoked rather than being silently skipped past the recoverer.
- `doInvokeWithRecords` skips listener invocation for records whose offset is greater than the in-flight retry offset on their partition and drops the offset from `offsetsInThisBatch` so an out-of-commit container (manual ack + async replies) does not stay paused waiting for an async ack on a record the listener was never called on.
- `handleAsyncFailure`, when `invokeErrorHandlerForFailedRecords` returns normally for a record whose partition has an `asyncRetryOffsets` entry at that offset, seeks the consumer past the recovered offset so the next poll re-fetches the still-failing records, and removes the entry. The wholesale clear of `offsetsInThisBatch[tp]` / `deferredOffsets[tp]` that `48d486bc` tried is left out — the skipped records have already been removed individually by `removeOffsetsInBatch` in the dispatch loop, and the wholesale clear is what wiped ack tracking for unrelated in-flight records on the partition and broke `Async{CompletableFuture,Mono}RetryTopicScenarioTests` (specifically `allFailCaseTest` and `oneLongSuccessMsgBetween49ShortFailMsg`).

## Test plan
- [x] New regression test `EnableKafkaKotlinCoroutinesTests#test suspend function bounded burst stays linear with N` for `N = 10`. Asserts total listener invocations equal `N * (n + 2) - 1 = 39`. Without the `asyncRetryOffsets` dispatch-loop skip the count would be `N * (N + 2) = 120` (quadratic) and the assertion would fail.
- [x] Existing GH-4504 two-record regression test (`test suspend function bounded retries for two records on the same partition`) — assertion for `r2` deliveries updated from 5 to 4, the four being iter-0 dispatch + own `n + 1` retry cycle.
- [x] Existing GH-4465 regression test (`test suspend function bounded retries with CommonErrorHandler`) — single-record case sees an empty `asyncRetryOffsets` and dispatches identically.
- [x] Full `./gradlew :spring-kafka:test` sweep passes locally, including `Async{CompletableFuture,Mono}RetryTopicScenarioTests` (`allFailCaseTest`, `oneLongSuccessMsgBetween49ShortFailMsg`), `RetryTopicIntegrationTests`, `KafkaMessageListenerContainerTests`, and the rest of `*Async*RetryTopic*` / `*KafkaMessageListenerContainer*`.

## Burst measurements

For `N` always-failing records on one partition with `maxAttempts = n = 2`. `✓` = locally measured against this branch; `✱` = predicted by the `N * (n + 2) - 1` formula at adjacent measured N. The "after #4505" column reproduces the original measurement from #4504 ([lobodpav's reproducer](https://github.com/lobodpav/spring-kafka-retry-issue)) on top of the data-loss-only state.

| N | blocking (`N * (n + 1)`) | this PR (suspend) | after #4505 alone (suspend) |
|---|---|---|---|
| 5 | 15 | 19 ✓ | 35 |
| 10 | 30 | 39 ✓ | 120 |
| 15 | 45 | 59 ✱ | 255 |
| 20 | 60 | 79 ✓ | 440 |

Linear in `N` at `N * (n + 2) - 1` vs the previous quadratic `N * (N + 2)`. Each non-head record gets one extra dispatch-loop delivery (the iter-0 invocation before `seenMultiAttemptRetry` flips on the first `RecordInRetryException`) on top of the blocking listener's `n + 1` retry budget.

## Why one extra delivery per non-head record

The async dispatch model invokes the listener on every polled record before any async failure callback fires. The blocking listener's sync throw aborts the batch iteration so a later record is never invoked while an earlier record is being retried; the async path cannot replicate that without giving up the parallelism that is the point of suspend / `Mono` / `CompletableFuture` listeners.

After the first `RecordInRetryException` on a consumer, the `seenMultiAttemptRetry` gate flips and the dispatch loop's sync-failure detection populates `asyncRetryOffsets` for subsequent records in the same batch — so once a partition is in retry, the rest of that batch on the partition is skipped. The gate cannot be on for the very first iteration on which a partition enters retry (the signal only exists after `RecordInRetryException` is observed), so each non-head record receives one iter-0 dispatch invocation before being skipped on later iterations. For non-idempotent per-record work, `@RetryableTopic` removes this entirely — `async-returns.adoc` covers the per-topic routing choice between a seek-based handler and `@RetryableTopic`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -869,24 +869,14 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 
 		private final ConcurrentLinkedDeque<FailedRecordTuple<K, V>> failedRecords = new ConcurrentLinkedDeque<>();
 
-		// Lowest in-flight retry offset per partition for the async path. Populated only
-		// when a seek-after-handling error handler throws RecordInRetryException (i.e.,
-		// the tracker is going to use its backoff budget). Consulted by
-		// doInvokeWithRecords to skip listener invocation for records on the same
-		// partition at higher offsets — those records will be re-fetched by the next
-		// poll after the retry resolves, so re-invoking the listener on them in the
-		// meantime only inflates the delivery count (GH-4504, head-of-line
-		// amplification). Accessed only on the consumer thread.
+		// Per-partition in-flight retry offset for the async seek-after-handling path
+		// (GH-4504 amplification bound). Consumer thread only.
 		private final Map<TopicPartition, Long> asyncRetryOffsets = new HashMap<>();
 
-		// True once any record on this consumer has caused the error handler to throw
-		// RecordInRetryException — i.e., the handler is a multi-attempt one. Gates the
-		// dispatch-loop sync-failure detection that populates asyncRetryOffsets within
-		// the same poll iteration: for publish-and-done recoverers (e.g.
-		// DeadLetterPublishingRecoverer wired into a @RetryableTopic flow) the failure
-		// callback fires synchronously but no RecordInRetryException is thrown, so the
-		// gate stays false and subsequent records on the partition continue to be
-		// invoked rather than being silently skipped past the recoverer.
+		// Gates the dispatch-loop sync-failure detection so publish-and-done recoverers
+		// (DeadLetterPublishingRecoverer into a retry topic) — which fire the callback
+		// synchronously without throwing RecordInRetryException — do not cause the
+		// remaining records on the partition to be silently skipped past the recoverer.
 		private boolean seenMultiAttemptRetry;
 
 		private boolean isListenerAdapterObservationAware = false;
@@ -1532,20 +1522,13 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 		}
 
 		protected void handleAsyncFailure() {
-			// Process only the records present at loop start; failures added concurrently
-			// are handled in the next loop iteration.
 			List<FailedRecordTuple<K, V>> failedRecordsSnapshot = new ArrayList<>(this.failedRecords);
 			if (failedRecordsSnapshot.isEmpty()) {
 				return;
 			}
-			// Sort by (topic, partition, offset) so that for a seek-after-handling error
-			// handler we always hand the lowest offset on a partition over first; this
-			// way, when the handler registers a seek-back, any subsequent records on the
-			// same partition can be dropped from this iteration and rely on the seek to
-			// re-fetch them, avoiding the per-record seek clobbering that silently
-			// skipped the earlier-offset record (GH-4504). For non-seeking handlers
-			// (e.g. RetryTopic) the sort is harmless and each record is processed
-			// individually to keep the existing dispatch semantics.
+			// Hand the lowest offset per partition to the handler first, so a
+			// seek-back covers the higher offsets and per-record seek clobbering
+			// does not silently skip the earlier-offset record (GH-4504).
 			failedRecordsSnapshot.sort(Comparator
 					.comparing((FailedRecordTuple<K, V> t) -> t.record.topic())
 					.thenComparingInt(t -> t.record.partition())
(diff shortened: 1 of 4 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2773,8 +2826,40 @@ private void doInvokeWithRecords(final ConsumerRecords<K, V> records) {
...
+					removeOffsetsInBatch(List.of(cRecord));
+					continue;
+				}
+				int failedBefore = this.failedRecords.size();
 				this.logger.trace(() -> "Processing " + KafkaUtils.format(cRecord));
 				doInvokeRecordListener(cRecord, iterator);
+				if (this.seenMultiAttemptRetry
+						&& this.commonErrorHandler != null
+						&& this.commonErrorHandler.seeksAfterHandling()
+						&& this.failedRecords.size() > failedBefore) {
```
sobychacko: This check assumes the listener that just ran is the one that added to failedRecords, but this deque is shared across all partitions, and async completions add to it from other threads. So `size()` here can be bigger just because some other partition's record failed in the meantime, and we'd wrongly tag `cRecord's` partition. If `cRecord` actually succeeded, that wrong tag never gets cleared (the records behind it get skipped, so they never fail to clear it), and the partition quietly stalls until a rebalance.

Could we check identity instead of the count - something like 

```
FailedRecordTuple<K, V> lastFailure = this.failedRecords.peekLast();
  if (... && lastFailure != null && lastFailure.record == cRecord) {
```

That way an unrelated failure just means "not this record" and we skip the optimization for a cycle (`handleAsyncFailure` catches it next poll) instead of risking a stuck partition.  What do you think?

spring-kafka/src/test/kotlin/org/springframework/kafka/listener/EnableKafkaKotlinCoroutinesTests.kt
```diff
@@ -142,6 +142,34 @@ class EnableKafkaKotlinCoroutinesTests {
...
+		this.config.burstRetryDeliveries.set(0)
+		repeat(burstSize) { idx ->
+			this.template.send("kotlinAsyncTestTopicBurstRetry", "burst-$idx")
+		}
+		assertThat(this.config.burstRetryRecoveredLatch.await(60, TimeUnit.SECONDS)).isTrue()
+		await()
+			.pollDelay(Duration.ofSeconds(2))
+			.atMost(Duration.ofSeconds(4))
+			.untilAsserted {
+				assertThat(this.config.burstRetryDeliveries.get()).isEqualTo(expected)
```
sobychacko: Same thing we ran into on #4505 with the r2 delivery count — an exact assertion here is going to be flaky, since 39 only holds if all 10 records land in the same first poll. If the burst splits across polls the number shifts.

Can we do a bound as we did there? Something like `isBetween(burstSize * (maxRetries + 1), burstSize * (maxRetries + 2))`. It proves it stayed linear (quadratic would be 120, way past it) without depending on poll timing.

sobychacko (changes requested):

spring-kafka/src/test/kotlin/org/springframework/kafka/listener/EnableKafkaKotlinCoroutinesTests.kt
BK202503: Switched to `isBetween(burstSize * (maxRetries + 1), burstSize * (maxRetries + 2))` in `69f07d32`.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
BK202503: Applied in `6fff256c` (helper `failedSyncForRecord`, mock test covering the race + the acknowledged miss).

PR #4514, opened 2026-06-24
Add consumer exception handler callback

### LINKED ISSUE #4490: handleConsumerException without CommonErrorHandler
**Expected Behavior**

When a fatal consumer exception is thrown, there should be a way to handle it without having to also use a CommonErrorHandler for regular listener error handling.

**Current Behavior**

Currently, you either have a CommonErrorHandler and participate in its regular listener error handling path, or there is no way to configure the error handling in `KafkaMessageListenerContainer.ListenerConsumer.handleConsumerException`

**Context**

Spring Kafka 4.0.5

We are running our Kafka listeners in an exactly-once setup with consumer-initiated transactions. 

In order to cleanly roll back transactions and handle errors outside the transaction boundary, we configure an `AfterRollbackProcessor`. We explicitly set the `CommonErrorHandler` to null, since it completely changes the error handling path, and there is no way to "opt out" of it handling listener errors that we want to leave to the AfterRollbackProcessor.

This is mostly fine, but in the rare occasion where something happens outside the message listener, and we hit `handleConsumerException(Exception)`, there is no error handler present to deal with this, and the error just gets logged. Instead, we would usually like to configure it to fail hard, stop the container and start fresh to make sure no messages get skipped.

So we are currently caught between a rock and a hard place - either configure a CommonErrorHandler, completely changing our error handling story, or don't, and have no ability to react to fatal consumer errors.

We looked into overriding `handleConsumerException(Exception)`, but that would require subclassing `KafkaMessageListenerContainer` which in turn is a problem because it is hard-wired into `ConcurrentMessageListenerContainer.constructContainer()` which is private.

Gautam-aman:
Fixes #4490

Adds a dedicated ConsumerExceptionHandler callback that can be configured through ContainerProperties.

The callback is invoked from KafkaMessageListenerContainer.ListenerConsumer#handleConsumerException, allowing applications to react to consumer-level exceptions without configuring a CommonErrorHandler.

This is useful for applications that rely on AfterRollbackProcessor-based listener error handling and do not want to participate in the CommonErrorHandler flow, while still being able to handle fatal consumer exceptions.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/annotation-error-handling.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/annotation-error-handling.adoc
@@ -113,6 +113,23 @@ The container commits any pending offset commits before calling the error handle
 
 If you are using Spring Boot, you simply need to add the error handler as a `@Bean` and Boot will add it to the auto-configured factory.
 
+[[consumer-exception-handler]]
+=== Consumer Exception Handler
+
+Starting with version 4.2, `ContainerProperties` provides a `consumerExceptionHandler` callback for exceptions thrown by the Kafka `Consumer` outside listener invocation, such as commit exceptions.
+Configure the callback on the container properties:
+
+[source, java]
+----
+factory.getContainerProperties().setConsumerExceptionHandler((exception, consumer, container) -> {
+    ...
+});
+----
+
+When a `consumerExceptionHandler` is configured, it takes precedence over the container's `CommonErrorHandler` for these consumer exceptions.
+In that case, `CommonErrorHandler.handleOtherException()` is not invoked.
+If no `consumerExceptionHandler` is configured, the container falls back to `CommonErrorHandler.handleOtherException()`, when a `CommonErrorHandler` is present.
+
 [[backoff-handlers]]
 == Back Off Handlers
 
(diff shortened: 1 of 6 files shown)


sobychacko:
@Gautam-aman Thanks for the PR! The approach looks reasonable overall, but a few things need to be addressed before this can be merged:

We need tests. Take a look at KafkaMessageListenerContainerTests around existing patterns testing handleOtherException via a mock error handler. We'd need at least: (a) ConsumerExceptionHandler gets called when set, (b) falls back to CommonErrorHandler.handleOtherException when not set, (c) RetriableCommitFailedException still bypasses both.

`@since` is wrong — the project is at 4.2.0-SNAPSHOT, so this should be `@since` 4.2, not 4.1.1. New API doesn't go into patch releases.

The new public methods need Javadoc.

Precedence needs to be documented — the else if in KMLC means if someone has both a ConsumerExceptionHandler and a CommonErrorHandler configured, handleOtherException on the handleOtherException on the
CommonErrorHandler silently won't fire for consumer exceptions. That should be mentioned in the Javadoc for setConsu merExceptionHandler in ContainerProperties.

Reference docs — new public API needs a mention in the docs module (error handling section). Take a look at how ContainerProperties options are documented there.

Thanks!

sobychacko:
@Gautam-aman Any updates? thanks!

Gautam-aman:
Sorry for delayed response. Will push the updates shortly. Thanks for showing patience.

sobychacko:
@Gautam-aman Thanks for the updates. The design looks good. One thing before we merge: the interface name. `ConsumerExceptionHandler` is a bit broad next to `CommonErrorHandler` / `KafkaListenerErrorHandler` and could read as "handles listener errors," when it's specifically for exceptions on the consumer thread *outside* listener invocation (e.g. commit failures). Could we rename to `ConsumerThreadExceptionHandler` or something similar? I'd avoid a `ConsumerAware*` name since that prefix already means "receives the Consumer" here. Keeping `ConsumerExceptionHandler` is ok if there is no better candidates, especially since it mirrors the internal `handleConsumerException` method, but let's settle the name now since it's public API in 4.2 and once it is out, then changing it would be really difficult.

sobychacko:
Thanks, @Gautam-aman. I am also seeking the opinion of @artembilan here for the new naming and a quick pass through this PR.

PR #4517, opened 2026-06-25
4.0.x: Cherry-pick `813f6891` to fix test regressions from the #4505 backport

BK202503:
Follow-up to the backport of #4505 on `4.0.x` (`ed31b093`). The cherry-pick depends on a precursor that exists on `main` but was never backported to `4.0.x`:

- `813f6891` — Support `CommonErrorHandler` for suspend functions. On `main`, `setCallbackForAsyncFailure(...)` is wired for any `RecordMessagingMessageListenerAdapter` (via `DelegatingMessageListener` extraction). On `4.0.x` it is only wired when the listener is a `KafkaBackoffAwareMessageListenerAdapter` (i.e., `@RetryableTopic` only), so a plain `suspend @KafkaListener` + `DefaultErrorHandler` never routes its async failure into `failedRecords`. With `ed31b093` in place but without `813f6891`, the three suspend-listener tests bundled into the backport never receive their recoverer callback and time out on the latch (`test suspend function with CommonErrorHandler`, `test suspend function bounded retries with CommonErrorHandler`, `test suspend function bounded retries for two records on the same partition`).

This PR cherry-picks `813f6891` onto `4.0.x` so the backport works as intended. Re-ran the full `./gradlew :spring-kafka:test` sweep locally on `4.0.x` with this commit applied — green, including the three previously failing tests and `ObservationTests` (which `813f6891` also touches).

The other potential precursor — `dfe52787` — is effectively already on `4.0.x` because `ed31b093` bundled both its container change (removal of `failedRecords.addLast(failedRecord)`) and its regression test. Cherry-picking `dfe52787` here produces only redundant conflicts, so it is not included.

The second commit adds `Bill Kim` to the class-level `@author` list of `KafkaMessageListenerContainer` — attribution for the GH-4504 fix that landed via #4505 but was omitted from the `@author` list on merge. Happy to drop this commit if you'd prefer to keep the PR scope strictly to the cherry-pick.

If you'd prefer to back out the backport from `4.0.x` instead, feel free to close this — happy to defer.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -78,7 +78,6 @@ import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.internals.RecordHeader;
 import org.jspecify.annotations.Nullable;
 
-import org.springframework.aop.support.AopUtils;
 import org.springframework.beans.BeanUtils;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationEventPublisher;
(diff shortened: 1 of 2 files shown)


sobychacko:
Thanks! Merged upstream to `4.0.x`. To merge this PR is the right thing to do since we are dealing with a data-loss situation.

PR #4518, opened 2026-06-26
Fix batch listener backoff state reset in multi-partition containers

### LINKED ISSUE #4371: Backoff breaks recovery on batching listener container in multi partition environment

29shivam:
## Problem

In a non-concurrent `KafkaMessageListenerContainer` with multiple partitions, when a successful batch follows a prior failure, `doInvokeBatchListener` calls `commonErrorHandler.clearThreadState()`. This wipes the entire `FailedRecordTracker` map for the current thread, resetting the failure count for **all** partitions — including partitions that are still inside their backoff/retry cycle.

The result: a broken record that should be retried N times (or sent to the DLQ after N attempts) restarts its counter every time any **other** partition successfully processes a batch, so it never exhausts its retries and never reaches the DLQ.

Reproduction scenario (from #4371):
1. Two-partition topic, non-concurrent container, backoff+DLQ configured
2. Partition A has a broken record, partition B has healthy records
3. Partition A fails → `batchFailed = true`, `FailedRecordTracker` records attempt 1 for A
4. Error handler seeks A back, commits B
5. Next poll only returns B's records → success → `clearThreadState()` → A's failure count wiped
6. A is un-paused; failure count is 1 again → cycle repeats, DLQ never reached

## Fix

Introduce `clearThreadStateFor(Collection<TopicPartition> partitions)` throughout the error-handler stack:

- `FailedRecordTracker` — removes only the specified partitions from the per-thread failure map
- `FailedRecordProcessor` — public delegate to tracker
- `CommonErrorHandler` — new default (no-op) so existing custom handlers need no changes
- `CommonMixedErrorHandler` / `CommonDelegatingErrorHandler` — delegate to sub-handlers
- `KafkaMessageListenerContainer.doInvokeBatchListener` — calls `clearThreadStateFor(records.partitions())` instead of `clearThreadState()`

This scopes the reset to only the partitions that were actually part of the successful batch, leaving backoff state intact for partitions still retrying.

The existing `clearThreadState()` call (on thread termination) is unchanged.

## Checklist
- [ ] Tests added/updated
- [ ] `@since` tag added to new API (`CommonErrorHandler.clearThreadStateFor`)

Closes #4371

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonDelegatingErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonDelegatingErrorHandler.java
@@ -16,11 +16,14 @@
 
 package org.springframework.kafka.listener;
 
+import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+
+import org.apache.kafka.common.TopicPartition;
 
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
@@ -95,6 +98,12 @@ public class CommonDelegatingErrorHandler implements CommonErrorHandler {
 	public void clearThreadState() {
 		this.defaultErrorHandler.clearThreadState();
 		this.delegates.values().forEach(CommonErrorHandler::clearThreadState);
+	}
+
+	@Override
+	public void clearThreadStateFor(Collection<TopicPartition> partitions) {
+		this.defaultErrorHandler.clearThreadStateFor(partitions);
+		this.delegates.values().forEach(h -> h.clearThreadStateFor(partitions));
 	}
 
 	@Override
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonErrorHandler.java
@@ -166,6 +166,15 @@ public interface CommonErrorHandler extends DeliveryAttemptAware {
 	}
 
 	/**
+	 * Clear thread state only for the given partitions. Called after a successful batch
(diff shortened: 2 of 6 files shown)


sobychacko:
@29shivam Thanks for the PR. Merged upstream via https://github.com/spring-projects/spring-kafka/commit/0c8ab73ec90e7037dda06857d0b8e972824c2d94.

PR #4538, opened 2026-07-27
GH-4537: Negative cache failed cluster id lookups

### LINKED ISSUE #4537: No negative caching in KafkaAdmin.clusterId() can block the listener and cause log spam
Note: This issue was written with AI assistance, but the bug occurred in a live
  application and is reproducible there for us. No minimal example (yet), though.

4.1.0 (via Spring Boot 4.1.0, kafka-clients 4.2.1)

**Describe the bug**

With `ContainerProperties.observationEnabled = true`, the listener container resolves the Kafka cluster id for every consumed record. If that lookup fails, `KafkaAdmin.clusterId()` does not remember the failure, so the next record starts the whole attempt again with a brand new `AdminClient`. This repeats for as long as records keep arriving.

In our case this produced thousands of INFO Log Entries within seconds.

The chain in 4.1.0:

1. `KafkaMessageListenerContainer.ListenerConsumer.doInvokeRecordListener()` creates the observation with `this::clusterId` as the context supplier:

   ```java
   Observation observation = KafkaListenerObservation.LISTENER_OBSERVATION.observation(
           this.containerProperties.getObservationConvention(),
           () -> new KafkaRecordReceiverContext(cRecord, getListenerId(), getClientId(), this.consumerGroupId,
                   this::clusterId),
           this.observationRegistry);
   ```

2. `KafkaRecordReceiverContext` invokes that supplier eagerly in its constructor:

   ```java
   String cluster = clusterId.get();
   setRemoteServiceName("Apache Kafka" + (cluster != null ? ": " + cluster : ""));
   ```

3. `ListenerConsumer.clusterId()` delegates to `KafkaAdmin.clusterId()` while its own field is `null`:

   ```java
   private @Nullable String clusterId() {
       if (this.kafkaAdmin != null && this.clusterId == null) {
           obtainClusterId();
       }
       return this.clusterId;
   }
   ```

4. `KafkaAdmin.clusterId()` only caches the success path. On failure the field stays `null` and nothing records that the attempt failed:

  ```java
  public String clusterId() {
      if (this.clusterId == null) {
          try (Admin client = createAdmin()) {
              this.clusterId = client.describeCluster().clusterId().get(this.operationTimeout, TimeUnit.SECONDS);
              if (this.clusterId == null) {
                  this.clusterId = "null";
              }
          }
          catch (InterruptedException ex) {
              Thread.currentThread().interrupt();
          }
          catch (Exception ex) {
              LOGGER.error(ex, "Could not obtain cluster info");   // no negative caching, no backoff
          }
      }
      return this.clusterId;
  }
  ```

So steps 1 to 4 run again for the next record, and for every record after that.

The cost per record is high. `get(...)` waits for `operationTimeout` (30 s by default), then the `try`-with-resources `close()` waits for the pending request as well, which the client reports right before the error is swallowed:

```
2026-07-27T16:00:23.376  INFO [| adminclient-1] o.a.k.clients.admin.KafkaAdminClient : [AdminClient clientId=adminclient-1] Timed out 1 remaining operation(s) during close.
2026-07-27T16:00:23.377 ERROR [ntainer#0-0-C-1] o.springframework.kafka.core.KafkaAdmin : Could not obtain cluster info
2026-07-27T16:01:23.391  INFO [| adminclient-2] o.a.k.clients.admin.KafkaAdminClient : [AdminClient clientId=adminclient-2] Timed out 1 remaining operation(s) during close.
```

Note the exactly 60 second interval and the increasing `adminclient-N` counter: a new client per record, for as long as traffic continues.

During each of those 60 second windows the client's reconnect loop logs two INFO lines per attempt, which is where the log volume comes from:

```
[AdminClient clientId=adminclient-1] Node -1 disconnected.
[AdminClient clientId=adminclient-1] Cancelled in-flight METADATA request with correlation id 5 due to node -1 being disconnected (...)
```

It can get considerably worse. `KafkaAdminClient.MetadataUpdateNodeIdProvider.provide()` logs `Rebootstrapping with Cluster(...)` at INFO whenever `LeastLoadedNode.hasNodeAvailableOrConnectionReady()` is `false`, and the runnable loop calls `provide()` for every entry in `pendingCalls` on each iteration while appending a new metadata call to that list. We have seen single 60 second windows end with `Timed out 148 remaining operation(s) during close.` and several thousand `Rebootstrapping` lines.

**Why the lookup fails in our case**

`ListenerConsumer.obtainAdmin()` picks up Spring Boot's auto-configured `KafkaAdmin` bean and only replaces `BOOTSTRAP_SERVERS_CONFIG` with the consumer's value:

```java
Map<String, Object> props = new HashMap<>(admin.getConfigurationProperties());
Object bootstrapServer = props.get(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG);
if (bootstrapServer != null && !bootstrapServer.equals(this.bootstrapServers)) {
    props.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, this.bootstrapServers);
    ...
}
```

No security properties are carried over. Our application configures its consumer itself and has no `spring.kafka.*` configuration, so the auto-configured `KafkaAdmin` has no SASL settings and cannot authenticate against the broker.

That is only one possible reason for the lookup to fail. The issue we would like to raise is the unbounded per-record retry, independent of the cause.

**To Reproduce**

Any setup in which the cluster id lookup fails while `observationEnabled` is `true` and records keep arriving:

1. Enable observation on the listener container (`containerProperties.isObservationEnabled = true`).
2. Make sure an `ObservationRegistry` with at least one `ObservationHandler` is present, otherwise `SimpleObservationRegistry.isNoop()` is `true` and the container skips the observation entirely.
3. Ensure the `KafkaAdmin` that `obtainAdmin()` resolves cannot reach the cluster - for example an auto-configured bean without the security settings the consumer uses.
4. Consume records.

Setting `isObservationEnabled = false` makes it stop immediately.

**Expected behavior**

`KafkaAdmin.clusterId()` should remember a failed attempt and apply a backoff, so that a failing cluster id lookup does not create a new `AdminClient` and a new blocking call for every single record.

Alternatives we can see would be to not resolve the cluster id eagerly in `KafkaRecordReceiverContext`, or to have `ListenerConsumer` stop asking after the first failure for its own lifetime.

**Sample**

No sample repository at this point - the analysis above is based on reading the 4.1.0 sources and on the logs of an affected application, with the configuration stripped down to the parts we believe are relevant.

renechoi:
Fixes #4537

`KafkaAdmin.clusterId()` only caches the success path. When the lookup fails the
field stays `null`, so the next call creates another `Admin` instance and blocks
for up to `operationTimeout` again. A listener container with observation enabled
resolves the cluster id for every record (`KafkaRecordReceiverContext` invokes the
supplier eagerly), so an admin that cannot reach the broker turns into a per record
stall plus a continuous stream of admin client reconnect logging, for as long as
records keep arriving.

This change remembers the time of a failed lookup and skips further attempts until
`clusterIdRetryInterval` has elapsed, five minutes by default. Behavior otherwise
is unchanged:

- a successful lookup is still cached for the lifetime of the bean
- an explicitly set cluster id still short circuits the lookup
- `setClusterIdRetryInterval(Duration.ZERO)` restores the previous behavior of
  attempting a fetch on every call

## Verification

`clusterIdFailureIsNotRetriedWithinRetryInterval` is the regression gate. Dropping
the `clusterIdFetchDue()` guard and keeping the test makes it fail with
`expected: 1 but was: 3`. `clusterIdIsFetchedAgainAfterRetryInterval` covers the new
setter and the fetch that happens once the interval has elapsed.

Run locally on JDK 25, based on `5d44034`:

- `./gradlew :spring-kafka:test --tests 'org.springframework.kafka.core.*'`, 118 tests,
  0 failures, 1 pre-existing skip
- `./gradlew :spring-kafka:test --tests 'org.springframework.kafka.support.micrometer.*'`,
  29 tests, 0 failures, 4 skips (the observation tests that consume the cluster id)
- `./gradlew :spring-kafka:check -x test`, includes `checkstyleMain` and `checkstyleTest`
- `git diff --check`

I did not reproduce the reporter's full SASL scenario against a live broker; the
tests drive the failure path through a mocked `Admin` instead.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/micrometer.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/micrometer.adoc
@@ -120,6 +120,10 @@ The sender and receiver contexts `remoteServiceName` properties are set to the K
 If, for some reason - perhaps lack of admin permissions, you cannot retrieve the cluster id, starting with version 3.1, you can set a manual `clusterId` on the `KafkaAdmin` and inject it into ``KafkaTemplate``s and listener containers.
 When it is `null` (default), the admin will invoke the `describeCluster` admin operation to retrieve it from the broker.
 
+Starting with version 4.1.1, a failed `describeCluster` attempt is remembered and is not retried until the `clusterIdRetryInterval` of the `KafkaAdmin` has elapsed; it defaults to 5 minutes.
+Each attempt creates an `Admin` instance and blocks for up to the `operationTimeout`, so, without this interval, every observed operation pays that cost while the cluster id is unavailable.
+Set `clusterIdRetryInterval` to `Duration.ZERO` to attempt the fetch on every call, as in earlier versions.
+
 [[batch-listener-obs]]
 === Batch Listener Observations
 
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
@@ -98,3 +98,11 @@ This enables consistent exception classification across both single and batch pr
 
 Note that to preserve backward compatibility, the default value for `resetStateOnExceptionChange` is `false` in `FailedBatchProcessor`, unlike `FailedRecordProcessor` where it defaults to `true`.
 If you want consistent behavior across both, call `setResetStateOnExceptionChange(true)` on your `DefaultErrorHandler` instance.
+
+[[x41-kafka-admin-cluster-id-retry]]
+=== `KafkaAdmin` Cluster Id Retry Interval
+
+When `KafkaAdmin.clusterId()` cannot retrieve the cluster id from the broker, the failure is now remembered and the fetch is not attempted again until the new `clusterIdRetryInterval` has elapsed; it defaults to 5 minutes.
+Previously every call retried, so an observation-enabled listener container created an `Admin` instance and blocked for up to the operation timeout on each record while the cluster id was unavailable.
+This changes the default behavior: set `clusterIdRetryInterval` to `Duration.ZERO` on the `KafkaAdmin` to opt out and attempt the fetch on every call, as before.
+See xref:kafka/micrometer.adoc#observation[Micrometer Observation] for more information.
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
@@ -373,9 +373,9 @@ public class KafkaAdmin extends KafkaResourceFactory
 
 	/**
 	 * Return the cluster id, fetching it from the broker on the first call. A failed
-	 * fetch is remembered and not attempted again until
-	 * {@link #setClusterIdRetryInterval(Duration)} has elapsed; each attempt creates an
-	 * {@link Admin} instance and blocks for up to
+	 * fetch is remembered and not attempted again until the
+	 * {@link #setClusterIdRetryInterval(Duration) clusterIdRetryInterval} has elapsed;
+	 * each attempt creates an {@link Admin} instance and blocks for up to
 	 * {@link #setOperationTimeout(int) operationTimeout}, so retrying on every call can
 	 * stall callers such as an observation-enabled listener container.
 	 * @return the cluster id, or null if it has not been fetched successfully.
--- a/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java
@@ -93,6 +93,13 @@ public class KafkaAdmin extends KafkaResourceFactory
 	 */
 	public static final Duration DEFAULT_CLOSE_TIMEOUT = Duration.ofSeconds(10);
 
+	/**
+	 * The default interval between attempts to obtain the cluster id after a failed
+	 * attempt, as 5 minutes.
+	 * @since 4.1.1
+	 */
+	public static final Duration DEFAULT_CLUSTER_ID_RETRY_INTERVAL = Duration.ofMinutes(5);
+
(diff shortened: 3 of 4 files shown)


sobychacko:
Thanks @renechoi, for the PR.  Caching the failure at `clusterId()` seems to be the right solution to tackle this issue. 

A couple of things before I merge:

  - Since this changes default behavior in a point release (a failed lookup now stays quiet for 5 minutes instead of retrying every call), let's add a note in `what's-new` mentioning the change and the `setClusterIdRetryInterval(Duration.ZERO) `opt-out.
  - Please tweak the `clusterId()` Javadoc - it currently says the retry waits until `setClusterIdRetryInterval(Duration)` "has elapsed", which should reference the interval value, not the setter.

Once those are in we can get it merged. Thanks again!

renechoi:
Thanks for the review @sobychacko. Done in f9e2ff0.

- `whats-new.adoc` now has a `KafkaAdmin` Cluster Id Retry Interval entry that calls out the default behavior change, the five minute default, and the `Duration.ZERO` opt-out.
- The `clusterId()` Javadoc reads "not attempted again until the `{@link #setClusterIdRetryInterval(Duration) clusterIdRetryInterval}` has elapsed", so it names the interval value while keeping the link for navigation, which matches the `{@link #setOperationTimeout(int) operationTimeout}` reference two lines below it.

One thing to call out: I also added three lines to `kafka/micrometer.adoc`, right where cluster id retrieval and the manual `clusterId` fallback are already described, so that the what's new `xref` lands on a section that actually documents the interval. Happy to drop that hunk if you would rather keep this change to the what's new entry alone.

Gates re-run on the pushed tree with JDK 25: `:spring-kafka:checkstyleMain` 0 violations, `:spring-kafka:javadoc` 0 warnings, and `org.springframework.kafka.core` 118 tests with 0 failures and 1 pre-existing skip.

sobychacko:
@renechoi Thanks again for the PR. It is now merged upstream to `main`.

renechoi:
Thanks for the review and for merging, @sobychacko.

PR #4551, opened 2026-08-04
GH-4545: Delegate onPartitionsAssigned in CommonDelegatingErrorHandler

### LINKED ISSUE #4545: CommonDelegatingErrorHandler does not delegate onPartitionsAssigned(), causing silent record loss during retry
4.1.0 (presumably since #2340, at least)

**Describe the bug**
`CommonDelegatingErrorHandler` does not override `onPartitionsAssigned()`, so the call hits the interface's no-op default and is never forwarded to the default handler or any delegate. A previously related bug (#2340) introduced this interface function as part of a safe retry, but failed to implement it in the CommonDelegatingErrorHandler.

Consequence: if a rebalance completes while the wrapped error handler is inside `retryBatch()`'s back-off, the newly assigned partitions are not re-paused. The keep-alive calls in the retry loop then fetch records from those partitions and discard them, advancing the positions. The records are never delivered to the listener; subsequent commits make the loss permanent. No error is raised.

**To Reproduce**
This is the reproduction flow:
- Setup uses a batch listener with `DefaultErrorHandler` (or any handler relying on `onPartitionsAssigned`) wrapped in `CommonDelegatingErrorHandler`, e.g.
```java
BackOff backOff = new FixedBackOff(3000, FixedBackOff.UNLIMITED_ATTEMPTS);
CommonErrorHandler defaultHandler = new DefaultErrorHandler(backOff);
CommonDelegatingErrorHandler delegatingErrorHandler = new CommonDelegatingErrorHandler(defaultHandler);
```
- A batch fails and the handler enters `retryBatch()`.
- A rebalance assigns partitions to the consumer while the retry back-off is in progress.
- Records available on the newly assigned partitions are then lost.

Manually overriding onPartitionsAssigned has been confirmed to solve the issue for our particular setup:
```java
CommonDelegatingErrorHandler delegatingErrorHandler = new CommonDelegatingErrorHandler(defaultHandler) {
    @Override
    public void onPartitionsAssigned(Consumer<?, ?> consumer, Collection<TopicPartition> partitions,
                                     Runnable publishPause) {
        defaultHandler.onPartitionsAssigned(consumer, partitions, publishPause);
    }
};
```
Test class available under Sample.

**Expected behavior**
`CommonDelegatingErrorHandler` should propagate `onPartitionsAssigned()` (it is not exception-scoped, so presumably to the default handler and all delegates).

**Sample**
Test project to reproduce the problem: [Kafka-CommonDelegatingErrorHandler-Issue.zip](https://github.com/user-attachments/files/30662698/Kafka-CommonDelegatingErrorHandler-Issue.zip)

kalayciburak:
Fixes #4545

## Summary
`CommonDelegatingErrorHandler` did not override `onPartitionsAssigned()`, so the call hit the no-op default on `CommonErrorHandler` and was never forwarded to the default handler or any delegate.

When a rebalance completes while a wrapped handler is inside `retryBatch()` back-off, newly assigned partitions are not re-paused. Keep-alive polls in the retry loop can then fetch and discard records from those partitions, advancing offsets without delivery.

This change forwards `onPartitionsAssigned` to the default handler and all delegates, mirroring `clearThreadState()`. The callback is not exception-scoped; handlers that are not actively retrying no-op, so only the retrying handler acts.

## Test plan
- [x] `./gradlew :spring-kafka:test --tests "org.springframework.kafka.listener.CommonDelegatingErrorHandlerTests"` — 12/12 GREEN (includes new `onPartitionsAssignedIsForwardedToDefaultAndDelegates`)
- [x] RED observed before the production change (`WantedButNotInvoked` on default/delegate `onPartitionsAssigned`)

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonDelegatingErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonDelegatingErrorHandler.java
@@ -41,6 +41,7 @@ import org.springframework.util.Assert;
  * @author Adrian Chlebosz
  * @author Antonin Arquey
  * @author Dan Blackney
+ * @author Burak Kalayci
  * @since 2.8
  *
  */
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonDelegatingErrorHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/CommonDelegatingErrorHandler.java
@@ -16,6 +16,7 @@
 
 package org.springframework.kafka.listener;
 
+import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -25,6 +26,7 @@ import java.util.Map.Entry;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.common.TopicPartition;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.kafka.support.ExceptionMatcher;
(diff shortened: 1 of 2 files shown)


sobychacko:
@kalayciburak Could you add your name as an author to the classes you changed?

kalayciburak:
done — added `@author` on both classes

PR #4552, opened 2026-08-05
Use Gradle version catalog for dependencies

cppwfs:
- Introduce gradle/libs.versions.toml to centralize dependency and plugin versions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
 	ext.isCI = System.getenv('GITHUB_ACTION')
-	ext.kotlinVersion = '2.3.21'
+	ext.kotlinVersion = libs.versions.kotlinVersion.get()
 	repositories {
 		gradlePluginPortal()
 		mavenCentral()
@@ -14,8 +14,8 @@ buildscript {
 plugins {
 	id 'base'
 	id 'idea'
-	id 'io.freefair.aggregate-javadoc' version '9.2.0'
-	id 'io.spring.nullability' version '0.0.14' apply false
+	alias(libs.plugins.aggregate.javadoc)
+	alias(libs.plugins.nullability) apply false
 }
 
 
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -1,6 +1,6 @@
 buildscript {
 	ext.isCI = System.getenv('GITHUB_ACTION')
-	ext.kotlinVersion = '2.3.21'
+	ext.kotlinVersion = libs.versions.kotlinVersion.get()
```
artembilan: I'd like to not have version variables without library.
Therefore, please, bring it back like we have in Spring AMQP & Spring Integration.

build.gradle
```diff
@@ -32,26 +32,26 @@ ext {
...
+	log4jVersion = libs.versions.log4jVersion.get()
+	micrometerDocsVersion = libs.versions.micrometerDocsVersion.get()
+	micrometerVersion = libs.versions.micrometerVersion.get()
+	micrometerTracingVersion = libs.versions.micrometerTracingVersion.get()
+	mockitoVersion = libs.versions.mockitoVersion.get()
+	reactorVersion = libs.versions.reactorVersion.get()
+	slf4jVersion = libs.versions.slf4jVersion.get()
+	springBootVersion = libs.versions.springBootVersion.get() // docs module
+	springDataVersion = libs.versions.springDataVersion.get()
+	springVersion = libs.versions.springVersion.get()
```
artembilan: Why do we need these variables yet?
I remember that we had similar discussion in other projects.
We ended up that we definitely don't need them. Do we?

build.gradle
```diff
@@ -163,9 +164,9 @@ configure(javaProjects) { subproject ->
 		testImplementation 'org.junit.jupiter:junit-jupiter-params'
 		testImplementation 'org.jetbrains.kotlin:kotlin-reflect'
 		testImplementation 'org.jetbrains.kotlin:kotlin-stdlib-jdk8'
-		testImplementation "org.awaitility:awaitility:$awaitilityVersion"
-		testImplementation "org.hamcrest:hamcrest-core:$hamcrestVersion"
-		testImplementation "org.assertj:assertj-core:$assertjVersion"
+		testImplementation libs.org.awaitility
+		testImplementation libs.org.hamcrest.core
+		testImplementation libs.org.assertj.core
```
artembilan, line 147: So, we use libs, but not variables here in Gradle.

artembilan (changes requested):

build.gradle
```diff
@@ -156,14 +135,13 @@ configure(javaProjects) { subproject ->
...
-		compileOnly spotBugs
-		testRuntimeOnly spotBugs
+		compileOnly libs.com.github.spotbugs.annotations
+		testRuntimeOnly libs.com.github.spotbugs.annotations
 
 		testImplementation 'org.junit.jupiter:junit-jupiter-api'
 		testImplementation 'org.junit.jupiter:junit-jupiter-params'
 		testImplementation 'org.jetbrains.kotlin:kotlin-reflect'
-		testImplementation 'org.jetbrains.kotlin:kotlin-stdlib-jdk8'
+		testImplementation libs.kotlin.stdlib.jdk8
```
artembilan: I don't think we need this as we didn't need it before.
Since you have brought Kotlin version variable back, we don't need any Kotlin-related entry in the TOML.
Note: that doesn't count with Coroutines.

artembilan (changes requested):

PR #4554, opened 2026-08-08
GH-3947: Align batch observation meter tags

### LINKED ISSUE #3947: Conflict when observability enabled and mixed batch and non-batch listeners

3.3.6

**Describe the bug**

Batch `@KafkaListener` is not observable, non-batch is observable - it causes Prometheus metric conflict.

When observability is enabled and there is one non-batch `@KafkaListener`, metric
http://localhost:8080/actuator/prometheus?includedNames=spring_kafka_listener_seconds is ok:

>  spring_kafka_listener_seconds_sum{error="none",messaging_kafka_consumer_group="aaa",messaging_operation="receive",messaging_source_kind="topic",messaging_source_name="bbb",messaging_system="kafka",spring_kafka_listener_id="org.springframework.kafka.KafkaListenerEndpointContainer#0-0"} 0.0016962

When second, batch `@KafkaListener` is added, it breaks non-batch metrics
> spring_kafka_listener_seconds_sum{exception="none",name="org.springframework.kafka.KafkaListenerEndpointContainer#0-0",result="success"} 0.0

Logs contains warning with explanation:
> The meter (MeterId{name='spring.kafka.listener', tags=[tag(error=none),tag(messaging.kafka.consumer.group=aaa),tag(messaging.operation=receive),tag(messaging.source.kind=topic),tag(messaging.source.name=bbb),tag(messaging.system=kafka),tag(spring.kafka.listener.id=org.springframework.kafka.KafkaListenerEndpointContainer#1-0)]}) registration has failed

**To Reproduce**

1. Build application with 
- org.springframework.kafka:spring-kafka
- org.springframework.boot:spring-boot-starter-actuator
- io.micrometer:micrometer-registry-prometheus
- org.springframework.boot:spring-boot-starter-web (for easy metric inspection)

Make sure: `spring.kafka.listener.observation-enabled=true`

Experiment with different combination of batch and non-batch `@KafkaListener`

**Expected behavior**

It cannot be accepted that adding batch listener breaks already working obsevations

When observability is enabled, metric for each listener should be registered with same set of tags, regardless listener is batch or non-batch). When some tag has no value, use placeholder "none"

Similar issue for MongoDB: https://github.com/spring-projects/spring-data-mongodb/pull/4994

arimu1:
## Description

Fixes #3947

When `observationEnabled` is true, non-batch listeners register
`spring.kafka.listener` meters via Micrometer Observation
(`messaging.*` + `spring.kafka.listener.id` + `error` tags), while
batch listeners previously fell back to legacy `MicrometerHolder`
timers (`name` / `result` / `exception` tags). Prometheus rejects
registering the same meter name with different tag key sets, so
adding a batch listener could break existing non-batch metrics.

### Changes

- When observation is enabled and `recordObservationsInBatch` is
  `false`, create a **batch-level** Observation wrapping the batch
  listener invocation (first record supplies source tags).
- Do not create legacy `MicrometerHolder` timers whenever
  `ContainerProperties.observationEnabled` is true (including batch).
- Always emit optional observation tags, using Micrometer's `none`
  placeholder when a value is unavailable (same approach as
  spring-data-mongodb #4994).
- Docs / what's-new updates for batch listener observations.

### Tests

```
./gradlew :spring-kafka:test \
  --tests org.springframework.kafka.support.micrometer.KafkaListenerObservationTests \
  --tests org.springframework.kafka.support.micrometer.MixedBatchRecordObservationMetersTests \
  --tests org.springframework.kafka.support.micrometer.BatchIndividualRecordObservationTests \
  --tests org.springframework.kafka.support.micrometer.MicrometerMetricsTests \
  --tests org.springframework.kafka.support.micrometer.ObservationTests \
  --tests org.springframework.kafka.support.micrometer.MicrometerHolderTests
```

**25/25 green** (Temurin 21). Also `:spring-kafka:checkstyleMain` +
`checkstyleTest` clean.

New regression: `MixedBatchRecordObservationMetersTests` asserts
mixed batch + record listeners share one `spring.kafka.listener` tag
key set.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
@@ -6,3 +6,26 @@
 
 This section covers the changes made from version 4.1 to version 4.2.
 For changes in earlier versions, see xref:appendix/change-history.adoc[Change History].
+
+[[x42-listener-observation-tags]]
+=== Listener Observation Meter Tags
+
+When `observationEnabled` is `true`, batch listeners create a batch-level observation (unless `recordObservationsInBatch` is enabled for per-record observations) using the same low-cardinality meter tags as non-batch listeners.
+Optional tag values use the Micrometer placeholder `none` when unavailable.
+This prevents Prometheus registration failures when an application mixes batch and non-batch listeners under the `spring.kafka.listener` meter name.
+See xref:kafka/micrometer.adoc#batch-listener-obs[Batch Listener Observations].
+
+==== Backward Compatibility
+
+These changes affect existing meters and dashboards:
+
+* **Batch listeners with `observationEnabled=true`:** Previously these used legacy `MicrometerHolder` timers with `name` / `result` / `exception` tags.
+They now emit observation-based meters with `messaging.*` tags (and Micrometer's `error` tag).
+Dashboards or recording rules that matched the old tag structure will stop matching and must be updated to the observation tag keys.
+
+* **All listeners (batch and non-batch) without a consumer group:** `messaging.kafka.consumer.group` was previously omitted from the low-cardinality tag set when no group was configured.
+It is now always present with value `none` (and the same applies to high-cardinality optional tags such as `messaging.kafka.client_id` and `messaging.consumer.id` when unavailable).
+Exact label-set matching queries that assumed the tag was absent need to be revised.
+
+These behavioral changes are intentional so that all `spring.kafka.listener` meters share one consistent tag key set for Prometheus.
+They ship in **4.2** only (they cannot be backported to 4.1).
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
@@ -178,6 +178,7 @@ import org.springframework.util.StringUtils;
  * @author Minchul Son
  * @author Youngjoo Kim
  * @author Bill Kim
+ * @author Hakaze Arimu
  */
 public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 		extends AbstractMessageListenerContainer<K, V> implements ConsumerPauseResumeEventPublisher {
@@ -1321,8 +1322,9 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 		private @Nullable MicrometerHolder obtainMicrometerHolder() {
 			MicrometerHolder holder = null;
 			try {
+				// Skip legacy timers when observation is active as mixed tag key sets break Prometheus registration.
 				if (KafkaUtils.MICROMETER_PRESENT && this.containerProperties.isMicrometerEnabled()
-						&& !this.observationEnabled) {
+						&& !this.containerProperties.isObservationEnabled()) {
 
 					Function<@Nullable Object, Map<String, String>> mergedProvider =
 							cr -> this.containerProperties.getMicrometerTags();
@@ -2570,8 +2572,14 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 			}
 		}
 
+		/**
+		 * Emit a short-lived observation per record when {@code recordObservationsInBatch} is true.
+		 * Does not invoke the listener or open a scope around listener execution; the batch
+		 * listener still runs via {@link #doInvokeBatchListenerBody} outside these observations
+		 * (see docs: per-record batch observations are not propagated to the listener method).
+		 * @param recordList the records in the batch.
+		 */
 		private void invokeBatchWithIndividualRecordObservation(List<ConsumerRecord<K, V>> recordList) {
-			// Create individual observations for each record without scopes
 			for (ConsumerRecord<K, V> record : recordList) {
 				Observation observation = KafkaListenerObservation.LISTENER_OBSERVATION.observation(
 						this.containerProperties.getObservationConvention(),
@@ -2582,6 +2590,51 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 				observation.observe(() -> {
 					this.logger.debug(() -> "Observing record in batch: " + KafkaUtils.format(record));
 				});
+			}
+		}
+
+		/**
+		 * Observe the entire batch listener invocation when observation is enabled but
+		 * per-record observations are not. Uses the first record for context tags so meters
+		 * share the same low-cardinality keys as non-batch listeners.
+		 * @param records the consumer records for the batch.
+		 * @param recordList the list of records passed to the listener.
+		 */
+		private void invokeBatchOnMessageWithObservation(final ConsumerRecords<K, V> records,
+				List<ConsumerRecord<K, V>> recordList) {
+
+			if (recordList.isEmpty()) {
+				// No record to derive messaging source tags; still invoke the listener.
+				doInvokeBatchListenerBody(records, recordList);
+				return;
+			}
+			ConsumerRecord<K, V> contextRecord = recordList.get(0);
+			Observation observation = KafkaListenerObservation.LISTENER_OBSERVATION.observation(
+					this.containerProperties.getObservationConvention(),
+					DefaultKafkaListenerObservationConvention.INSTANCE,
+					() -> new KafkaRecordReceiverContext(contextRecord, getListenerId(), getClientId(),
+							this.consumerGroupId, this::clusterId),
+					this.observationRegistry);
+			observation.observe(() -> doInvokeBatchListenerBody(records, recordList));
+		}
+
+		/**
+		 * Invoke the batch listener (full poll result or list).
+		 * @param records the consumer records for the batch.
+		 * @param recordList the list of records passed to the listener.
+		 */
+		private void doInvokeBatchListenerBody(final ConsumerRecords<K, V> records,
+				List<ConsumerRecord<K, V>> recordList) {
+
+			if (this.wantsFullRecords) {
+				Objects.requireNonNull(this.batchListener).onMessage(records, // NOSONAR
+						this.isAnyManualAck
+								? new ConsumerBatchAcknowledgment(records, recordList)
+								: null,
+						this.consumer);
+			}
+			else {
+				doInvokeBatchOnMessage(records, recordList); // NOSONAR
 			}
 		}
 
@@ -2608,18 +2661,17 @@ public class KafkaMessageListenerContainer<K, V> // NOSONAR line count
 
 			try {
 				if (this.observationEnabled) {
+					// recordObservationsInBatch: emit per-record spans only (listener runs below, unobserved)
 					invokeBatchWithIndividualRecordObservation(recordList);
 				}
 
-				if (this.wantsFullRecords) {
-					Objects.requireNonNull(this.batchListener).onMessage(records, // NOSONAR
-							this.isAnyManualAck
-									? new ConsumerBatchAcknowledgment(records, recordList)
-									: null,
-							this.consumer);
+				if (this.containerProperties.isObservationEnabled() && !this.observationEnabled) {
+					// Batch-level observation wraps the listener (same meter tags as non-batch)
+					invokeBatchOnMessageWithObservation(records, recordList);
 				}
 				else {
-					doInvokeBatchOnMessage(records, recordList); // NOSONAR
+					// Listener body: unobserved when recordObservationsInBatch, or when observation is off
+					doInvokeBatchListenerBody(records, recordList);
 				}
 				batchInterceptAfter(records, null);
 				successTimer(sample, null);
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
@@ -16,6 +16,7 @@
 
 package org.springframework.kafka.support.micrometer;
 
+import io.micrometer.common.KeyValue;
 import io.micrometer.common.KeyValues;
 import io.micrometer.common.docs.KeyName;
 import io.micrometer.observation.Observation.Context;
@@ -33,6 +34,7 @@ import org.springframework.util.StringUtils;
  * @author Christian Mergenthaler
  * @author Wang Zhiyang
  * @author Christian Fredriksson
+ * @author Hakaze Arimu
  *
  * @since 3.0
  *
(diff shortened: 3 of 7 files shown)


spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -7,6 +7,14 @@
...
 For changes in earlier versions, see xref:appendix/change-history.adoc[Change History].
 
+[[x41-batch-listener-observation-tags]]
+=== Batch Listener Observation Tags
+
+When `observationEnabled` is `true`, batch listeners create a batch-level observation (unless `recordObservationsInBatch` is enabled for per-record observations) using the same low-cardinality meter tags as non-batch listeners.
+Optional tag values use the Micrometer placeholder `none` when unavailable.
+This prevents Prometheus registration failures when an application mixes batch and non-batch listeners under the `spring.kafka.listener` meter name.
+See xref:kafka/micrometer.adoc#batch-listener-obs[Batch Listener Observations].
+
```
sobychacko, line 17: Needs a backward compatibility callout:

1. . Existing batch listeners with `observationEnabled=true` previously got legacy MicrometerHolder timers (name/result/exception tags). They now get observation-based meters (messaging.* tags). Any existing dashboards using the old tag structure will silently stop working.
2. All non-batch listener users who had no consumer group configured: `messaging.kafka.consumer.group` was previously omitted from the tag set; it is now always emitted as
 "none". This changes the cardinality of existing dashboard queries that don't expect this tag.

Both of these are behavioral breaking changes. Please add explicit migration guidance here.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -1321,8 +1323,12 @@ protected void checkConsumer() {
 		private @Nullable MicrometerHolder obtainMicrometerHolder() {
 			MicrometerHolder holder = null;
 			try {
+				// Do not create legacy Micrometer timers when observation is enabled on the
+				// container: batch listeners fall back to timers only when observation is off.
+				// Mixing observation meters (messaging.* tags) with timer meters (name/result/
+				// exception tags) under the same "spring.kafka.listener" name breaks Prometheus.
```
sobychacko: Not sure if we need that level of in-line java comment details. Maybe make it concise in a single line? something like - `// Skip legacy timers when observation is active as mixed tag key sets break Prometheus registration.`

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2585,6 +2591,48 @@ private void invokeBatchWithIndividualRecordObservation(List<ConsumerRecord<K, V
...
+		 * share the same low-cardinality keys as non-batch listeners.
+		 * @param records the consumer records for the batch.
+		 * @param recordList the list of records passed to the listener.
+		 */
+		private void invokeBatchOnMessageWithObservation(final ConsumerRecords<K, V> records,
+				List<ConsumerRecord<K, V>> recordList) {
+
+			ConsumerRecord<K, V> contextRecord = !recordList.isEmpty()
+					? recordList.get(0)
+					: new ConsumerRecord<>(KeyValue.NONE_VALUE, 0, 0L, null, null);
```
sobychacko: `KeyValue.NONE_VALUE` is a Micrometer constant that means "this tag has no value" — it's not meant to be used as a Kafka topic name. It just happens to be the string "none", which is why this compiles, but the two have nothing to do with each other. If Micrometer ever changes what NONE_VALUE is, the topic name in this record silently changes too. Better to just use a plain string like "unknown" here.

 Also worth considering: if recordList is empty, should we even be starting an observation? An early return before creating the observation would make it clearer that there's nothing to observe.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
```diff
@@ -2608,18 +2656,16 @@ private void invokeBatchOnMessageWithRecordsOrList(final ConsumerRecords<K, V> r
...
-									? new ConsumerBatchAcknowledgment(records, recordList)
-									: null,
-							this.consumer);
+				if (this.containerProperties.isObservationEnabled() && !this.observationEnabled) {
+					// Batch-level observation: same meter tags as non-batch listeners
+					invokeBatchOnMessageWithObservation(records, recordList);
 				}
 				else {
-					doInvokeBatchOnMessage(records, recordList); // NOSONAR
+					doInvokeBatchListenerBody(records, recordList);
```
sobychacko, line 2674: Question on this logic: when `recordObservationsInBatch=true` (so `this.observationEnabled=true`), the first if above runs `invokeBatchWithIndividualRecordObservation` which creates per-record Observation spans with empty lambdas (it does not invoke the listener). Then this `else` branch fires `doInvokeBatchListenerBody` which actually invokes the listener — outside any observation scope. The listener runs unobserved in per-record mode. Is that the intended semantics? If so, a comment on `invokeBatchWithIndividualRecordObservation` clarifying that it only emits spans and does not drive listener invocation would prevent future regressions here.

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MixedBatchRecordObservationMetersTests.java
```diff
@@ -0,0 +1,212 @@
...
+					.as("all spring.kafka.listener timers must share one tag key set")
+					.hasSize(1);
+
+			// Both listener ids present
+			Set<String> listenerIds = timers.stream()
+					.map(t -> t.getId().getTag("spring.kafka.listener.id"))
+					.collect(Collectors.toSet());
+			assertThat(listenerIds).anyMatch(id -> id != null && id.startsWith("mixedRecord-"));
+			assertThat(listenerIds).anyMatch(id -> id != null && id.startsWith("mixedBatch-"));
+		});
```
sobychacko, line 120: Looks like this test only covers the success path. Could you add a test where the batch listener throws a `RuntimeException` and assert that: (a) the Observation records the error tag correctly, and (b) the legacy failureTimer (via `MicrometerHolder`) does not fire — i.e., there's no `spring.kafka.listener` meter with exception as a tag key.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
```diff
@@ -227,44 +229,33 @@ public static class DefaultKafkaListenerObservationConvention implements KafkaLi
...
-			KeyValues keyValues = KeyValues.of(
+			return KeyValues.of(
 					ListenerLowCardinalityTags.LISTENER_ID.withValue(context.getListenerId()),
 					ListenerLowCardinalityTags.MESSAGING_SYSTEM.withValue("kafka"),
 					ListenerLowCardinalityTags.MESSAGING_OPERATION.withValue("process"),
 					ListenerLowCardinalityTags.MESSAGING_SOURCE_NAME.withValue(context.getSource()),
-					ListenerLowCardinalityTags.MESSAGING_SOURCE_KIND.withValue("topic")
+					ListenerLowCardinalityTags.MESSAGING_SOURCE_KIND.withValue("topic"),
+					ListenerLowCardinalityTags.MESSAGING_CONSUMER_GROUP.withValue(
+							StringUtils.hasText(groupId) ? groupId : KeyValue.NONE_VALUE)
```
sobychacko, line 242: This is a behavioral change for all existing non-batch listener users: `messaging.kafka.consumer.group` was previously omitted when no group was configured; now it's always present as "none". This is the right fix for the tag-consistency problem, but it is a breaking change for users who have Prometheus recording rules or Grafana queries that do exact label-set matching. This should be explicitly called out in the what's-new and migration docs — not just in the batch listener section, since it affects all listeners.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
```diff
@@ -227,44 +229,33 @@ public static class DefaultKafkaListenerObservationConvention implements KafkaLi
...
 		@NonNull
 		public KeyValues getHighCardinalityKeyValues(KafkaRecordReceiverContext context) {
 			String clientId = context.getClientId();
 			String consumerId = getConsumerId(context.getGroupId(), clientId);
-			KeyValues keyValues = KeyValues.of(
+			return KeyValues.of(
 					ListenerHighCardinalityTags.MESSAGING_PARTITION.withValue(context.getPartition()),
-					ListenerHighCardinalityTags.MESSAGING_OFFSET.withValue(context.getOffset())
+					ListenerHighCardinalityTags.MESSAGING_OFFSET.withValue(context.getOffset()),
+					ListenerHighCardinalityTags.MESSAGING_CLIENT_ID.withValue(
```
sobychacko, line 254: Same comment as above.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
```diff
@@ -227,44 +229,33 @@ public static class DefaultKafkaListenerObservationConvention implements KafkaLi
...
+							StringUtils.hasText(clientId) ? clientId : KeyValue.NONE_VALUE),
+					ListenerHighCardinalityTags.MESSAGING_CONSUMER_ID.withValue(
```
sobychacko, line 256: ditto.

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MixedBatchRecordObservationMetersTests.java
```diff
@@ -0,0 +1,212 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+
+/**
+ * Regression for GH-3947: mixed batch and non-batch listeners with observation enabled
+ * must register {@code spring.kafka.listener} meters with a consistent tag key set.
+ *
+ * @author Hakaze Arimu
+ * @since 4.1
```
sobychacko: 4.2 as we cannot backport the changes to 4.1 (main has already been switched to `4.2`).

sobychacko (changes requested):

arimu1:
Thanks @sobychacko — addressed all review feedback on tip `1d609b61b3722fb7e14de2bef8bed6b26d90d8b9` (rebased onto current `main` / 4.2).

### Point-by-point

1. **whats-new backward compatibility** — Section is now under **4.2** (`[[x42-listener-observation-tags]]`) with an explicit **Backward Compatibility** subsection covering:
   - Batch listeners with `observationEnabled=true`: legacy `MicrometerHolder` timers (`name`/`result`/`exception`) → observation meters (`messaging.*` + `error`); dashboards using the old tag structure will stop matching.
   - **All** listeners (not only batch): optional tags such as `messaging.kafka.consumer.group` are always present (`none` when unavailable); exact label-set queries that assumed omission break.
   - Note that this ships in **4.2 only** (cannot backport to 4.1).

2. **`obtainMicrometerHolder` comment** — Condensed to a single line:  
   `// Skip legacy timers when observation is active as mixed tag key sets break Prometheus registration.`

3. **`KeyValue.NONE_VALUE` as topic** — Removed. Empty `recordList` no longer fabricates a synthetic `ConsumerRecord`; it skips starting an observation and still invokes the listener body (nothing useful to tag).

4. **`recordObservationsInBatch=true` path** — Confirmed intentional (matches docs: per-record observations are **not** propagated to the listener method). Added javadoc on `invokeBatchWithIndividualRecordObservation` and call-site comments clarifying that per-record mode only emits short-lived spans; the listener still runs via `doInvokeBatchListenerBody` outside those observations.

5. **Failure-path test** — `MixedBatchRecordObservationMetersTests.batchListenerFailureRecordsObservationErrorNotLegacyExceptionTag`: batch listener throws `RuntimeException`; asserts observation meter has a non-`none` `error` tag and that no `spring.kafka.listener` meter uses legacy `exception` / `name` / `result` tag keys. `@since` updated to **4.2**.

6. **Always-present group/topic tags** — Consistency fix kept; migration guidance is in the whats-new callout (item 1), covering all listeners.

### Verification
- `MixedBatchRecordObservationMetersTests` (2), `KafkaListenerObservationTests` (3), `BatchIndividualRecordObservationTests` (2) — all passed (JDK 21)
- `:spring-kafka:checkstyleMain` + `checkstyleTest` — passed

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
arimu1: Done — expanded into a **Backward Compatibility** subsection under the 4.2 whats-new entry (`[[x42-listener-observation-tags]]`), covering both behavioral breaks (batch MicrometerHolder → observation meters, and always-present optional tags for all listeners) plus an explicit note that this ships in **4.2 only**.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
arimu1: Agreed — condensed to a single line as suggested.

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
arimu1: Fixed: no longer uses `KeyValue.NONE_VALUE` as a topic. For an empty `recordList` we skip starting an observation and still invoke the listener body (nothing useful to derive messaging source tags from).

spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java
arimu1: Intended semantics (matches existing docs): when `recordObservationsInBatch=true`, we only emit short-lived per-record observations — they are **not** scopes around the listener method. The batch listener still runs via `doInvokeBatchListenerBody` outside those observations. Documented with javadoc on `invokeBatchWithIndividualRecordObservation` and call-site comments so this is harder to regress.

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MixedBatchRecordObservationMetersTests.java
arimu1: Added `batchListenerFailureRecordsObservationErrorNotLegacyExceptionTag`: batch listener throws `RuntimeException`, asserts observation records a non-`none` `error` tag, and asserts no `spring.kafka.listener` meter uses legacy `exception`/`name`/`result` tag keys.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
arimu1: Kept the always-present optional tags (needed for Prometheus tag-key consistency). Explicit migration guidance is in the 4.2 whats-new **Backward Compatibility** subsection and covers **all** listeners, not only batch.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
arimu1: Same as above — documented in the 4.2 whats-new backward-compatibility callout for all listeners.

spring-kafka/src/main/java/org/springframework/kafka/support/micrometer/KafkaListenerObservation.java
arimu1: Same as above — documented in the 4.2 whats-new backward-compatibility callout for all listeners.

spring-kafka/src/test/java/org/springframework/kafka/support/micrometer/MixedBatchRecordObservationMetersTests.java
arimu1: Updated `@since` to **4.2** (and moved the whats-new entry under the 4.2 section after rebasing onto current `main`).

sobychacko:
Thanks, @arimu1 for the PR and addressing all review feedback. It is now merged upstream.

PR #4557, opened 2026-08-10
Migrate project to use TOML

cppwfs:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,8 +14,8 @@ buildscript {
 plugins {
 	id 'base'
 	id 'idea'
-	id 'io.freefair.aggregate-javadoc' version '9.2.0'
-	id 'io.spring.nullability' version '0.0.14' apply false
+	alias(libs.plugins.aggregate.javadoc)
+	alias(libs.plugins.nullability) apply false
 }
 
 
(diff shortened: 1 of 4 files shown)


cppwfs:
Investigating build failure

artembilan (approved):
LGTM — this backport follows the same resolution we settled on in #4552:

* `kotlinVersion` stays as a plain `ext` variable (no TOML entry), matching Spring AMQP/Spring Integration, per the earlier feedback.
* No leftover duplicate version variables in `build.gradle`'s `ext` block — dependencies reference `libs.*` accessors directly instead of re-declaring versions.
* Versions carried into `gradle/libs.versions.toml` are unchanged from what was already in the `4.1.x` branch (kafka 4.2.1, spring 7.0.8, spring-data 2026.0.0, reactor 2025.0.6, micrometer 1.17.0/1.7.0, junit 6.0.3, jackson3 3.1.5) — no accidental bumps snuck in.

One thing this PR actually gets right that #4552 missed on `main`: `gradle/docs.gradle` here is updated to use `libs.versions.kafkaVersion.get()` / `springVersion.get()` / `springBootVersion.get()` / `micrometerTracingVersion.get()` instead of the bare Groovy variables. On `main` today, `gradle/docs.gradle` still references those bare variables even though `build.gradle`'s `ext` block no longer defines them, so `:spring-kafka-docs:generateAntoraYml` currently fails with `Could not get unknown property 'springVersion'`. I reproduced that failure on `main` and confirmed `:spring-kafka-docs:generateAntoraYml --offline` passes cleanly on this branch, so `4.1.x` won't inherit that regression.

CI is green. Approving.

PR #4568, opened 2026-08-16
GH-4567: Honor autoStartup after registry stop

### LINKED ISSUE #4567: Kafka listener with autoStartup=false is unconditionally started on ApplicationContext.restart() (alwaysStartAfterRefresh=true ignores autoStartup after first refresh)

4.1.0 (via `spring-boot-starter-parent:4.1.0`), in combination with Spring Framework 7.0.3+.

**Describe the bug**

`KafkaListenerEndpointRegistry.startIfNecessary(MessageListenerContainer)` ignores a listener's
`autoStartup` value on every registry start **after** the first `ContextRefreshedEvent`, because
`alwaysStartAfterRefresh` defaults to `true`:

```java
private void startIfNecessary(MessageListenerContainer listenerContainer) {
    if ((this.contextRefreshed && this.alwaysStartAfterRefresh) || listenerContainer.isAutoStartup()) {
        listenerContainer.start();
    }
}
```

`contextRefreshed` is set to `true` once and never reset. So any later call to the registry's
`start()` — e.g. triggered by `ConfigurableApplicationContext.restart()` — unconditionally starts
**all** registered listener containers, regardless of their configured `autoStartup` (even
`autoStartup = "${some.property}"` resolving to `false`).

We hit this via a new trigger for that restart: Spring Framework 7.0.3 introduced test
`ApplicationContext` pause/resume in `DefaultContextCache` (default `PauseMode.ON_CONTEXT_SWITCH`),
which calls `ConfigurableApplicationContext.restart()` when switching back to a previously paused,
cached context. In our test suite, a `@KafkaListener` with
`autoStartup="${application.kafka.prognose.active}"` (resolving to `false`) is meant to stay
stopped. After such a context restart, the registry silently starts it anyway, and it begins
consuming real Kafka messages in the background across many unrelated subsequent test classes —
without any error or warning, corrupting shared test infrastructure (in our case a shared MongoDB
test database).

We consider `autoStartup=false` being honored only for the *first* startup, and silently
overridden on every later lifecycle restart, surprising and effectively undocumented. Either
`alwaysStartAfterRefresh` should default to `false` (so `autoStartup` is respected for the whole
context lifecycle), or the interaction with `ConfigurableApplicationContext.restart()` should be
documented clearly, since it silently reactivates listeners explicitly configured to stay stopped.

**To Reproduce**

1. Spring Boot 4.1.0 / Spring Framework 7.0.3+ with Spring Kafka 4.1.0.
2. Define a `@KafkaListener` with `autoStartup="${some.flag}"` resolving to `false`.
3. Write ≥2 `@SpringBootTest` classes with different contexts (e.g. different
   profiles/properties), so `DefaultContextCache` pauses one context, plus a third test class that
   switches back to the first context afterwards — this triggers `restart()` on it.
4. Assert whether `registry.getListenerContainer(<id>).isRunning()` is `true` after the switch
   back (or observe unexpected message consumption / consumer-group rebalances in the logs).
5. Observe: the listener container is running after `restart()`, even though `autoStartup` still
   resolves to `false`.

**Expected behavior**

A `@KafkaListener`/`MessageListenerContainer` whose `autoStartup` resolves to `false` should stay
stopped for the entire `ApplicationContext` lifecycle, including after `restart()`, unless
`alwaysStartAfterRefresh` was explicitly set to `true`. At minimum, this behavior should be clearly
documented.

**Sample**

No standalone public repo yet, but the root cause is fully reproducible from source:

- `org.springframework.kafka.config.KafkaListenerEndpointRegistry#startIfNecessary`
  (`spring-kafka-4.1.0-sources.jar`, see snippet above) — `alwaysStartAfterRefresh` defaults to
  `true`, `contextRefreshed` is set once and never reset.
- `org.springframework.test.context.cache.DefaultContextCache` (Spring Framework 7.0.3+,
  `PauseMode.ON_CONTEXT_SWITCH`) — calls `ConfigurableApplicationContext.restart()` when resuming a
  paused, cached context.

Workaround: calling `registry.setAlwaysStartAfterRefresh(false)` (e.g. in a `@PostConstruct` hook)
so the registry always respects each container's own `autoStartup`, including after `restart()`.

Note: altough this report is generated we have observed the described behavior in our project and fixed it by applying the mentioned workaround.

arimu1:
Fixes #4567

## Summary

`KafkaListenerEndpointRegistry` sets `contextRefreshed` on the first `ContextRefreshedEvent` and never clears it. After that, every registry `start()` (including those triggered by `ConfigurableApplicationContext.restart()` in Spring Framework 7.0.3+ test context pause/resume) unconditionally starts all listener containers when `alwaysStartAfterRefresh` is `true` (the default), ignoring `autoStartup=false`.

This change resets `contextRefreshed` when the registry stops (`stop()` and `stop(Runnable)`), so a subsequent `start()`/`restart()` re-applies each container's `autoStartup` setting. The default for `alwaysStartAfterRefresh` is unchanged, and late endpoint registration after refresh still starts containers immediately when `alwaysStartAfterRefresh` is `true`.

## Test plan

- [x] `./gradlew :spring-kafka:test --tests "org.springframework.kafka.config.KafkaListenerEndpointRegistryTests.stopThenStartHonorsAutoStartupWhenAlwaysStartAfterRefreshTrue"`
- [x] `./gradlew :spring-kafka:test --tests "org.springframework.kafka.config.KafkaListenerEndpointRegistryTests.startAfterRefreshWithoutStopStartsContainersWhenAlwaysStartAfterRefreshTrue"`
- [x] Prove-it: `stopThenStartHonorsAutoStartupWhenAlwaysStartAfterRefreshTrue` fails without the production change and passes with it
- Platform: macOS (darwin 25.6.0), JDK 21 (Eclipse Temurin 21.0.11+10)

## AI disclosure

Implementation and tests were written with AI assistance (Cursor) and reviewed by the contributor before submission.

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/config/KafkaListenerEndpointRegistryTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/config/KafkaListenerEndpointRegistryTests.java
@@ -29,6 +29,7 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
+import org.springframework.context.event.ContextRefreshedEvent;
 import org.springframework.context.support.GenericApplicationContext;
 import org.springframework.kafka.listener.ConcurrentMessageListenerContainer;
 import org.springframework.kafka.listener.MessageListenerContainer;
@@ -38,11 +39,15 @@ import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
 import static org.mockito.BDDMockito.given;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 
 /**
  * @author Gary Russell
  * @author Joo Hyuk Kim
  * @author Artem Bilan
+ * @author arimu1
  *
  * @since 2.8.9
  */
@@ -144,6 +149,33 @@ public class KafkaListenerEndpointRegistryTests {
 	}
 
 	@Test
+	void stopThenStartHonorsAutoStartupWhenAlwaysStartAfterRefreshTrue() {
+		KafkaListenerEndpointRegistry registry = new KafkaListenerEndpointRegistry();
+		GenericApplicationContext context = new GenericApplicationContext();
+		context.refresh();
+		registry.setApplicationContext(context);
+		MessageListenerContainer container = registerAutoStartupFalseContainer(registry, "listener");
+		registry.start();
+		verify(container, never()).start();
+		registry.onApplicationEvent(new ContextRefreshedEvent(context));
+		registry.stop();
+		registry.start();
+		verify(container, never()).start();
+	}
+
+	@Test
+	void startAfterRefreshWithoutStopStartsContainersWhenAlwaysStartAfterRefreshTrue() {
+		KafkaListenerEndpointRegistry registry = new KafkaListenerEndpointRegistry();
+		GenericApplicationContext context = new GenericApplicationContext();
+		context.refresh();
+		registry.setApplicationContext(context);
+		MessageListenerContainer container = registerAutoStartupFalseContainer(registry, "listener");
+		registry.onApplicationEvent(new ContextRefreshedEvent(context));
+		registry.start();
+		verify(container, times(1)).start();
+	}
+
+	@Test
 	void verifyUnregisteredListenerContainer() {
 		KafkaListenerEndpointRegistry registry = new KafkaListenerEndpointRegistry();
 		GenericApplicationContext applicationContext = new GenericApplicationContext();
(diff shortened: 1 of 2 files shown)


artembilan (changes requested):
Have just fixed the same problem in Spring AMQP: https://github.com/spring-projects/spring-amqp/pull/3601.

I see that this was generated by AI.
So, asked it, please, to follow code guidance from there.

Thanks

spring-kafka/src/test/java/org/springframework/kafka/config/KafkaListenerEndpointRegistryTests.java
```diff
@@ -187,6 +217,20 @@ private static void registerWithListenerIds(KafkaListenerEndpointRegistry regist
 		names.forEach(name -> registerListenerWithId(registry, name));
 	}
 
+	private static MessageListenerContainer registerAutoStartupFalseContainer(KafkaListenerEndpointRegistry registry,
+			String id) {
+
+		KafkaListenerEndpoint endpoint = mock(KafkaListenerEndpoint.class);
+		@SuppressWarnings("unchecked")
+		KafkaListenerContainerFactory<MessageListenerContainer> factory = mock(KafkaListenerContainerFactory.class);
```
artembilan: And why your AI didn’t follow `mock()` changes in respective Spring AMQP PR?
Or better question: why didn’t you review yourself ?
AI is the tool which doesn’t see the whole context until you point there. And in many cases doing review and follow up is much faster and easier yourself .

In my experience I don’t see that it is really intelligent. It does not learn and repeats mistakes again and again 🤷.

Please , involve yourself in contributions you do everywhere

artembilan (changes requested):

artembilan (approved):
Follow-up commits align this with the merged spring-amqp#3601 fix (record refactor, matching test names/structure). LGTM.

PR #4594, opened 2026-08-28
GH-4593: Apply factory configuration in createContainer(String...)

### LINKED ISSUE #4593: ShareKafkaListenerContainerFactory.createContainer(String...) does not apply factory configuration

4.0.x, and current `main` (f086f9ec).

**Describe the bug**

`ShareKafkaListenerContainerFactory.createListenerContainer(KafkaListenerEndpoint)` calls `initializeContainer(...)` (line 160), but `createContainer(String... topics)` (line 220) returns `createContainerInstance(...)` directly and never calls it.

As a result, containers created programmatically by topic silently ignore every factory-level setting: the copied `ContainerProperties`, concurrency, `autoStartup`, phase, `applicationContext`, `applicationEventPublisher`, the `ShareConsumerRecordRecoverer`, `groupId` and `clientIdPrefix`.

The `applicationEventPublisher` case is the easiest to miss — every publish site is null-guarded, so container lifecycle events are simply never emitted, with no error.

```java
var factory = new ShareKafkaListenerContainerFactory<>(consumerFactory);
factory.setShareConsumerRecordRecoverer(myRecoverer);
factory.setAutoStartup(false);

var container = factory.createContainer("my-topic");

// container.getShareConsumerRecordRecoverer() is the default REJECTING, not myRecoverer
// container.isAutoStartup() is true
```

**Expected behavior**

`createContainer(String...)` should configure the container the same way `createListenerContainer(KafkaListenerEndpoint)` does.

**Possible cause**

`ShareKafkaListenerContainerFactory` implements `KafkaListenerContainerFactory` directly rather than extending `AbstractKafkaListenerContainerFactory`, so there is no shared funnel guaranteeing `initializeContainer` runs on every creation path.

---

Happy to open a PR — it's a small change to call `initializeContainer(...)` on this path too. Wanted to check first whether bypassing factory configuration here was intentional.

Gaurav1112:
**Problem**

`ShareKafkaListenerContainerFactory.createContainer(String... topics)` returned `createContainerInstance(endpoint)` without calling `initializeContainer(...)`, while the sibling `createListenerContainer(KafkaListenerEndpoint)` calls it (line 160).

Factory-level settings were therefore silently dropped on the programmatic path: the copied `ContainerProperties`, concurrency, `autoStartup`, phase, `applicationContext`, `applicationEventPublisher`, the `ShareConsumerRecordRecoverer`, `groupId` and `clientIdPrefix`.

**Root cause**

The class implements `KafkaListenerContainerFactory` directly rather than extending `AbstractKafkaListenerContainerFactory`, so there is no inherited path guaranteeing `initializeContainer` runs on every creation route.

**Solution**

Hoist the endpoint to a local and call `initializeContainer(instance, endpoint)`, matching `createListenerContainer`. +5/-2 in one file.

**Tests**

Added `createContainerByTopicsShouldApplyFactoryConfiguration` to the existing `ShareKafkaMessageListenerContainerUnitTests`, asserting the recoverer, event publisher, `autoStartup` and phase are applied via `createContainer(String...)`.

It fails on current `main` with an `AssertionError` (the container holds the default `REJECTING` recoverer instead of the configured one) and passes with this change. I also reverted only the production hunk while keeping the test, and confirmed it fails again with an `AssertionError` rather than a compile error, so the test genuinely exercises the change.

Locally: the full `*Share*` suite is 66 tests / 0 failures (including the 27 integration tests), and `checkstyleMain` is clean, on JDK 25.

**Risk**

Low. Additive only — no API added, removed or renamed. The behaviour change is confined to the previously-unconfigured path.

Fixes GH-4593

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
@@ -218,13 +218,16 @@ public class ShareKafkaListenerContainerFactory<K, V>
 
 	@Override
 	public ShareKafkaMessageListenerContainer<K, V> createContainer(String... topics) {
-		return createContainerInstance(new KafkaListenerEndpointAdapter() {
+		KafkaListenerEndpoint endpoint = new KafkaListenerEndpointAdapter() {
 
 			@Override
 			public Collection<String> getTopics() {
 				return Arrays.asList(topics);
 			}
-		});
+		};
+		ShareKafkaMessageListenerContainer<K, V> instance = createContainerInstance(endpoint);
+		initializeContainer(instance, endpoint);
+		return instance;
 	}
 
 	@Override
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
@@ -39,6 +39,7 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
+import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.kafka.config.KafkaListenerEndpoint;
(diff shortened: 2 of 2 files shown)


sobychacko:
Thanks for the PR. Merged via https://github.com/spring-projects/spring-kafka/commit/e40a97976ebf2bf1c0d7a2997b081ee9f4bafcb7.

PR #4595, opened 2026-08-29
GH-4325: Expose client instance IDs for share consumers

### LINKED ISSUE #4325: Expose `ShareConsumer.clientInstanceId(Duration)` for KIP-714 telemetry
`ShareConsumer.clientInstanceId(Duration)` returns a broker-assigned `UUID` for correlating client and broker metrics in telemetry systems. This is distinct from the client ID string (which the container already sets with `-N` suffixes for concurrency) — `clientInstanceId` is a KIP-714 telemetry identifier.

Expose this through the container so users can retrieve the telemetry instance ID for a running share consumer.

OmarMorales71:
GH-4325: Expose client instance IDs for share consumers

Fixes https://github.com/spring-projects/spring-kafka/issues/4325

Adds `getClientInstanceIds(Duration)` to `ShareKafkaMessageListenerContainer`,
delegating to `ShareConsumer.clientInstanceId(Duration)` so the broker-assigned
client instance IDs (KIP-714 telemetry) can be correlated with client metrics.

The method returns the IDs keyed by client id, or an empty map when the
container is not running. The value is `null` for a consumer whose id was not
assigned within the timeout.

Tests passed.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
@@ -16,10 +16,12 @@
 
 package org.springframework.kafka.listener;
 
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


sobychacko:
@OmarMorales71 Thanks for the PR! Merged upstream via https://github.com/spring-projects/spring-kafka/commit/59711acaa19a591af34d77f71b0366722dcc0933. Applied another polish [commit](https://github.com/spring-projects/spring-kafka/commit/091474953f52a594af70a4f0835c03abc9954719) on top with docs.

OmarMorales71:
Thanks @sobychacko I was unsure If I needed to add docs for that PR, appreciate your commit with docs!

PR #4604, opened 2026-08-29
GH-4603: Handle a missing source record in the recovering Streams handlers

### LINKED ISSUE #4603: Recovering Streams exception handlers throw when there is no source record

4.2.0-SNAPSHOT (`main`), and 4.1.x.

**Describe the bug**

`RecoveringProcessingExceptionHandler.handleError` reads `context.sourceRawKey().length` and
`record.timestamp()`. Kafka calls it from `StreamTask.punctuate` with a null record and a context
whose topic and raw key and value are null, as `ErrorHandlerContext.sourceRawKey` documents, so the
handler throws `NullPointerException` and Streams kills the stream thread with "Fatal user code
error" instead of recovering. `RecoveringProductionExceptionHandler` has the same dereferences and
also passes the null topic into `ConsumerRecord`, which rejects it.

**Expected behavior**

The handler decides what to do when there is no source record, instead of throwing.

renechoi:
Fixes GH-4603

The recovering Streams handlers dereferenced `context.sourceRawKey()`, `context.sourceRawValue()`
and the record, which Kafka documents as null when the failure comes from a punctuation callback or
is not tied to an input record. The processing handler threw `NullPointerException` and the
production handler passed a null topic into `ConsumerRecord`, so a handler installed to recover
errors killed the stream thread instead.

Both now return `fail()` when there is no source record to recover, and the raw key and value sizes
fall back to `ConsumerRecord.NULL_SIZE` when Kafka does not provide them, which is what the existing
code already does for the forwarded record.

Four unit tests cover the punctuation and missing-raw-value paths; they throw on `main`.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/streams/RecoveringProcessingExceptionHandler.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/streams/RecoveringProcessingExceptionHandler.java
@@ -26,6 +26,7 @@ import org.apache.kafka.common.record.TimestampType;
 import org.apache.kafka.streams.errors.ErrorHandlerContext;
 import org.apache.kafka.streams.errors.ProcessingExceptionHandler;
 import org.apache.kafka.streams.processor.api.Record;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.kafka.listener.ConsumerRecordRecoverer;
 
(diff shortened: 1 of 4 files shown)


sobychacko:
@renechoi Merged via https://github.com/spring-projects/spring-kafka/commit/02be4e0d6b5dc5ec0c06f8579f135fce30630a63. Thanks for the PR! Going forward, please use your real name when signing the commit and also add your name as an author on the classes you are changing. Thanks again.

Issue #1244, opened 2019-09-23
Support request based reply timeout in ReplyingKafkaTemplate


keesbroenink:
Hi,

ReplyingKafkaTemplate (2.2.7-RELEASE) should be setup as a singleton. This means I can only define my reply-timeout once and all my interactions with th etemplate will use this setting. I want to set the timeout per request (replyFuture) so I can still re-use the template but have a different timeout per request.

Thx,

Kees Broenink 
keesbroenink@gmail.com

keesbroenink:
Thanks a lot!

keesbroenink:
@garyrussell 
Is it possible to use the change using a pre-release?

garyrussell:
You can get spring-kafka-2.3.0.BUID-SNAPSHOT from https://repo.spring.io/snapshot

It's scheduled to be released next week.

PR #4606, opened 2026-08-30
GH-4605: Name the validated property in Assert messages

### LINKED ISSUE #4605: Assert messages in setTopicPartitions and setDefaultReplyTimeout name the wrong property
`4.2.0-SNAPSHOT` (`main` at `f086f9ec`). Neither is a recent regression: the `setTopicPartitions` message 
goes back to `c0fe2c2f`, the original `@KafkaListener` commit, and the `setDefaultReplyTimeout` one to `220a5aae` in 2.3.

I found these by scanning `Assert.*` calls whose quoted property name does not match the value being validated, 
rather than from a production incident, so there is no field report behind this and the reproductions below are unit level.

**Describe the bug**

Two public setters reject invalid input with a message that names a *different* property, sending the caller to the wrong place.

**1. `AbstractKafkaListenerEndpoint#setTopicPartitions`**

`setTopics` and `setTopicPartitions` are two distinct, 
mutually exclusive properties on the same class, but they produce byte-identical messages:

```java
// AbstractKafkaListenerEndpoint.java:196
public void setTopics(String... topics) {
	Assert.notNull(topics, "'topics' must not be null");

// AbstractKafkaListenerEndpoint.java:220
public void setTopicPartitions(TopicPartitionOffset... topicPartitions) {
	Assert.notNull(topicPartitions, "'topics' must not be null");
```

So `setTopicPartitions(null)` reports a failure on `topics`, a property the caller may never have touched, 
and one the class javadoc explicitly contrasts with `topicPartitions` ("Either this or 'topic' or 'topicPattern' should be provided, 
but not a mixture"). The message has been copied from `setTopics` since `c0fe2c2f`, the original `@KafkaListener` commit.

**2. `ReplyingKafkaTemplate#setDefaultReplyTimeout`**

The two asserts in this one method disagree with each other:

```java
// ReplyingKafkaTemplate.java:194
public void setDefaultReplyTimeout(Duration defaultReplyTimeout) {
	Assert.notNull(defaultReplyTimeout, "'defaultReplyTimeout' cannot be null");
	Assert.isTrue(defaultReplyTimeout.toMillis() >= 0, "'replyTimeout' must be >= 0");
```

`replyTimeout` is not another name for this property. In this same class it is the **per-send** timeout, 
the parameter of `sendAndReceive(record, replyTimeout)` (lines 372, 386, 417), 
so the message points at a different API than the one that failed.

This is a leftover from `220a5aae` (GH-1244, 2.3), which split the single `replyTimeout` property into a template-level `defaultReplyTimeout` and a per-send `replyTimeout`. The old setter read:
```java
public void setReplyTimeout(long replyTimeout) {
	Assert.isTrue(replyTimeout >= 0, "'replyTimeout' must be >= 0");
```

When `setDefaultReplyTimeout(Duration)` replaced it, the new `notNull` message used the new name but the `isTrue` 
message was carried over unchanged. The old `setReplyTimeout(long)` has since been removed, 
so the name it refers to no longer exists as a property at all.

**To Reproduce**

Both assertions below describe the intended message and fail on `main` today

```java
@Test
void assertMessagesNameTheValidatedProperty() {
	assertThatIllegalArgumentException()
			.isThrownBy(() -> new MethodKafkaListenerEndpoint<String, String>()
					.setTopicPartitions((TopicPartitionOffset[]) null))
			.withMessage("'topicPartitions' must not be null");
	// actual: 'topics' must not be null

	GenericMessageListenerContainer<String, String> container = mock(GenericMessageListenerContainer.class);
	given(container.getContainerProperties()).willReturn(new ContainerProperties("reply"));
	ReplyingKafkaTemplate<String, String, String> template =
			new ReplyingKafkaTemplate<>(mock(ProducerFactory.class), container);

	assertThatIllegalArgumentException()
			.isThrownBy(() -> template.setDefaultReplyTimeout(Duration.ofMillis(-1)))
			.withMessage("'defaultReplyTimeout' must be >= 0");
	// actual: 'replyTimeout' must be >= 0
}
```

The test is self-contained against the library's own classes, so there is nothing an external sample repository would add.

**Expected behavior**

Each message names the property it actually validated

* `setTopicPartitions` → `'topicPartitions' must not be null`
* `setDefaultReplyTimeout` → `'defaultReplyTimeout' must be >= 0`

I am happy to submit a PR with both fixes and tests asserting the messages, if this is worth changing.

ohchanKyu:
Fixes: https://github.com/spring-projects/spring-kafka/issues/4605

`AbstractKafkaListenerEndpoint#setTopics` and `#setTopicPartitions` are two distinct, mutually exclusive properties, but both rejected null with `'topics' must not be null`, so a failure on `topicPartitions` pointed the caller at a property they may never have set. The message has been copied from `setTopics` since `c0fe2c2f`, the original `@KafkaListener` commit.

`ReplyingKafkaTemplate#setDefaultReplyTimeout` disagreed with itself: the `notNull` check named `defaultReplyTimeout` while the `isTrue` check named `replyTimeout`, which in that class is the per-send timeout parameter of `sendAndReceive(record, replyTimeout)`, not this property. The `isTrue` message was carried over unchanged from the old `setReplyTimeout(long)` when `220a5aae` (GH-1244) split the property in 2.3; that setter has since been removed, so the name no longer refers to a property at all.

* Name the property each check actually validates

Both new tests fail before the change with the old messages and pass after. A third test pins the already correct `setTopics` message so the two cannot converge again.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/AbstractKafkaListenerEndpoint.java
@@ -64,6 +64,7 @@ import org.springframework.util.ObjectUtils;
  * @author Wang Zhiyang
  * @author Sanghyeok An
  * @author Borahm Lee
+ * @author OhKyu Chan
  *
  * @see MethodKafkaListenerEndpoint
  */
@@ -218,7 +219,7 @@ public abstract class AbstractKafkaListenerEndpoint<K, V>
 	 * @see #setTopicPattern(Pattern)
 	 */
 	public void setTopicPartitions(TopicPartitionOffset... topicPartitions) {
-		Assert.notNull(topicPartitions, "'topics' must not be null");
+		Assert.notNull(topicPartitions, "'topicPartitions' must not be null");
 		this.topicPartitions.clear();
 		this.topicPartitions.addAll(Arrays.asList(topicPartitions));
 	}
--- a/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/requestreply/ReplyingKafkaTemplate.java
@@ -74,6 +74,7 @@ import org.springframework.util.Assert;
  * @author Borahm Lee
  * @author Francois Rosiere
  * @author Mikhail Polivakha
+ * @author OhKyu Chan
  *
  * @since 2.1.3
  *
@@ -193,7 +194,7 @@ public class ReplyingKafkaTemplate<K, V, R> extends KafkaTemplate<K, V> implemen
 	 */
 	public void setDefaultReplyTimeout(Duration defaultReplyTimeout) {
 		Assert.notNull(defaultReplyTimeout, "'defaultReplyTimeout' cannot be null");
-		Assert.isTrue(defaultReplyTimeout.toMillis() >= 0, "'replyTimeout' must be >= 0");
+		Assert.isTrue(defaultReplyTimeout.toMillis() >= 0, "'defaultReplyTimeout' must be >= 0");
 		this.defaultReplyTimeout = defaultReplyTimeout;
 	}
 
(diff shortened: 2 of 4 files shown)


sobychacko:
Merged via https://github.com/spring-projects/spring-kafka/commit/5a30fe8fa13ca7c9cfc174cf969823ac8830c2fa. Thanks for the PR!

PR #4619, opened 2026-09-01
GH-4519: Await async results on container stop

### LINKED ISSUE #4519: Async @KafkaListener (CompletableFuture / Mono / suspend): in-flight work outlives container stop and races downstream resources shutdown
# Expected Behavior

When a listener container is stopped (e.g. on application shutdown), in-flight async `@KafkaListener` results should be accounted for as part of stopping the container: either awaited for a bounded grace period or cancelled, and sequenced so this completes *before* the downstream beans the listener depends on (a reactive MongoDB client, a `DataSource`, an HTTP client, etc) are destroyed.

At minimum there should be a supported, opt-in way to get this behavior, so async listener work does not keep running against a half-destroyed application context during shutdown.

# Current Behavior

In-flight async listener results are not bound to the container's lifecycle. On shutdown the container is stopped early (it's a `SmartLifecycle` bean with a high phase), but the async work returned by the listener keeps running on its own threads and can outlive the container stop. In particular, it can take long enough to overlap the destruction of ordinary downstream beans, which happens later during normal bean destruction.

When the still-running listener work then uses an already-closed downstream resource it fails. With reactive MongoDB this surfaces as `IllegalStateException: state should be: open` (and `"Attempted to decrement the reference count below 0"`); separately, the eventual acknowledgement runs against an already-closed consumer.

To visualize this:

Image

*The container stops first (high `SmartLifecycle` phase). The async work is not bound to it, so it outlives the stop and overlaps the downstream resource's `close()` (ordinary bean destruction). Any downstream access in that overlap fails.*

From what I understand, the reason why this happens is in `MessagingMessageListenerAdapter#handleResult` ([source @ 87710ea](https://github.com/spring-projects/spring-kafka/blob/87710eacc15cb02c86aaed5593764f775af5a84d/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java#L585-L603)):
`mono.toFuture()` subscribes internally and returns a `CompletableFuture`; the Reactor `Disposable` is never surfaced, and the `CompletableFuture` is only used to attach the ack callback. So the container retains no handle to the in-flight work, meaning that there's nothing to await or cancel at stop time.
The container's `shutdownTimeout` only bounds waiting for the consumer thread to leave the poll loop (which returns immediately for async listeners), so increasing it does not help.

## Minimal reproducer

I created a self-contained reproducer that contains two tests showing the described issue, including the fact that this is an issue of async listeners in general, not coroutine-specific: https://github.com/L7R7/spring-kafka-async-shutdown-repro

# Context

In production we run Kotlin `suspend` `@KafkaListener`s that write to a reactive MongoDB. On every rolling deploy / scale-down we saw shutdown error logs (`state should be: open`, reference-count errors) because listener coroutines were still running while the Mongo client was being closed. I first raised this in discussion [#4506](https://github.com/spring-projects/spring-kafka/discussions/4506), having investigated it further, I've come to the opinion that this should be addressed in spring-kafka itself, and opening an issue seems like the right next step.

What I'm trying to accomplish: a clean shutdown where in-flight async listener work finishes (or is cancelled) before the resources it depends on are torn down. In my opinion, spring-kafka should take care of this without each of my applications having to re-implement that itself.

A rough idea for how this could be implemented is the following:

* retain a handle to each in-flight async result per container instead of fire-and-forget
* on container stop, await outstanding results up to a configurable grace period and then cancel/dispose the rest, sequnced before downstream beans are destroyed
* suppress shutdown-induced acks/error-handling for cancelled work so offsets stay uncommitted and the records are redelivered on next start (at-least-once semantics, which is already the implicit contract for `MANUAL`/out-of-order async listeners)

I'm not really familiar with the codebase of spring-kafka, so I'm eager to hear maintainers' opinions on that idea. If that's something worth adding, the question would be whether or not that should be the default or not. I kind of think it should be, because this currently is more like a bug than a missing feature.

# Workaround

We implemented a `SmartLifecycle` bean (with a phase below the phase of the Kafka container, and above the downstream resource beans) that runs the listener body on an application-managed `CoroutineScope`, and drain/cancel in-flight work in it's `stop()` function. This does work, but it's per-application boilerplate for something that the framework should take care of.

<details>
<summary>KafkaListenerLifecycle bean</summary>

```kotlin
import kotlinx.coroutines.CoroutineScope
import kotlinx.coroutines.Dispatchers
import kotlinx.coroutines.SupervisorJob
import kotlinx.coroutines.async
import kotlinx.coroutines.cancel
import kotlinx.coroutines.isActive
import kotlinx.coroutines.job
import kotlinx.coroutines.joinAll
import kotlinx.coroutines.runBlocking
import kotlinx.coroutines.withTimeoutOrNull
import org.slf4j.LoggerFactory
import org.springframework.beans.factory.annotation.Value
import org.springframework.context.SmartLifecycle
import org.springframework.stereotype.Component
import java.time.Duration
import java.util.concurrent.atomic.AtomicBoolean
import kotlin.time.toKotlinDuration

@Component
class KafkaListenerLifecycle(
  @Value("\${app.kafka.listener.drain-grace}") drainGrace: Duration
) : SmartLifecycle {
  private val logger = LoggerFactory.getLogger(javaClass)
  private val drainGrace = drainGrace.toKotlinDuration()
  private val running = AtomicBoolean(false)

  @Volatile
  private var scope = newScope()

  private fun newScope() = CoroutineScope(SupervisorJob() + Dispatchers.Default)

  @Synchronized
  private fun activeScope(): CoroutineScope {
    if (!scope.isActive) {
      scope = newScope()
    }
    return scope
  }

  /** Run a listener body on the lifecycle-owned scope so it can be drained on shutdown. */
  suspend fun <T> runDrainable(block: suspend () -> T): T = activeScope().async { block() }.await()

  override fun start() {
    activeScope()
    running.set(true)
  }

  override fun isRunning() = running.get()

  override fun stop() {
    running.set(false)
    val current = scope
    val inFlight =
      current.coroutineContext.job.children
        .toList()
    if (inFlight.isEmpty()) {
      current.cancel()
      return
    }

    logger.info("Shutdown: draining ${inFlight.size} in-flight Kafka listener batch(es)")
    runBlocking {
      val finished = withTimeoutOrNull(drainGrace) { inFlight.joinAll() }
      if (finished == null) {
        logger.warn("In-flight batches did not finish within $drainGrace; cancelling.")
      } else {
        logger.info("All in-flight Kafka listener batches drained cleanly")
      }
      current.cancel()
    }
  }

  /** Stop this bean after Kafka containers, but before the mongo client */
  override fun getPhase() = Int.MAX_VALUE - 200
}

```
</details>

nikita-kibitkin:
Fixes #4519

Implements the plan approved in the issue (https://github.com/spring-projects/spring-kafka/issues/4519#issuecomment-5431323992): in-flight async listener results are awaited on container stop within `shutdownTimeout`, then cancelled. Opt-in through a new container property, no new timeout knob, default behaviour unchanged. Main only (new API), `@since 4.2`.

### Changes

- `ContainerProperties.awaitAsyncResultsOnStop` (default `false`).
- `MessagingMessageListenerAdapter`: `addCallbackForAsyncResult(Consumer, callback)` / `removeCallbackForAsyncResult(Consumer)`, the same wiring shape as `setCallbackForAsyncFailure`. For each `CompletableFuture` or `Mono` result (suspend functions arrive as a `Mono`) the adapter hands the callback a `CompletableFuture<Void>` that completes after the result was handled, i.e. after the reply was sent and the record acked, or the failure handled. Cancelling that future cancels the source: `MonoToCompletableFuture.cancel()` disposes the subscription, which cancels the coroutine; for a plain `CompletableFuture` it is `cancel(true)`, best effort. A result cancelled this way skips the ack, the `KafkaListenerErrorHandler` and the async retry callback, so the offset stays uncommitted and the record is redelivered after reassignment. A future cancelled by user code (for example Kotlin `withTimeout`, whose `TimeoutCancellationException` is a `CancellationException`) still goes through the normal failure path; only container-initiated cancellation is skipped.
- `KafkaMessageListenerContainer.ListenerConsumer`: registers the callback when the property is on, keeps the in-flight set, and in `wrapUp()` waits for it before `commitPendingAcks()` and `consumer.close()`. The deadline is `shutdownTimeout` measured from the stop request, so `stop()` returns within the same window as today. The drain loop wakes on each completion and commits the queued acks right away, so offsets acked during the wait are committed before a blocking `stop()` returns; the rest are cancelled at the deadline with one INFO line naming the count.
- On a fatal consumer error the outstanding results are cancelled (not awaited) before `emergencyStop` and `consumer.close()`, so no result keeps running against a closed consumer; the timeout and the fatal path share `cancelAsyncResults(reason)`.
- The blocking stop test relies on the `WakeupException` commit retry from #4682 (merged), which was found while writing it.
- The callback is keyed by the child container's `Consumer` instead of a single setter because the children of a `ConcurrentMessageListenerContainer` share one adapter instance (the `messageListener` reference is copied by `BeanUtils.copyProperties`). With one slot the last started child would collect its siblings' results, and their acks would arrive on consumers that are already closed. `eachChildContainerAwaitsItsOwnResults` covers this with concurrency 2.
- Docs: `container-props.adoc`, a new section in `async-returns.adoc`, `whats-new.adoc`.

### Testing

- New tests: `AwaitAsyncResultsOnStopTests` (CompletableFuture and Mono, each completed within the timeout and cancelled after it with redelivery on restart, concurrency 2, a blocking `stop()` with a `ConsumerInterceptor` recording commit times, property off), two suspend tests in `EnableKafkaKotlinCoroutinesTests`, five unit tests in `MessagingMessageListenerAdapterTests`, `inFlightAsyncResultsCancelledOnFatalError` in `ConcurrentMessageListenerContainerMockTests` (fatal path, mock consumer). Verified to fail without the container change: 6 of 7 in the new class and both Kotlin tests (the seventh pins the property-off behaviour).

```
./gradlew :spring-kafka:test \
  --tests org.springframework.kafka.listener.AwaitAsyncResultsOnStopTests \
  --tests org.springframework.kafka.listener.EnableKafkaKotlinCoroutinesTests \
  --tests org.springframework.kafka.listener.adapter.MessagingMessageListenerAdapterTests
```

28/28 green (Temurin 21).
- `./gradlew :spring-kafka:checkstyleMain :spring-kafka:checkstyleTest` clean.
- Regression run, all green: `AsyncListenerTests`, `AsyncMonoDefaultErrorHandlerRetryTests`, `AsyncAckAfterHandleTests`, `EnableKafkaKotlinTests`, the five `retrytopic.Async*` classes, `KafkaMessageListenerContainerTests`, `ConcurrentMessageListenerContainerMockTests`, `ConcurrentMessageListenerContainerTests`, both adapter test classes, `KafkaListenerEndpointRegistryLifecycleTests`, `ObservationTests`, `EnableKafkaIntegrationTests`, `TransactionalContainerTests`, `RetryTopicIntegrationTests`, `ManualAckPartialBatchTests`, `ManualNackRecordTests`, `ManualNackBatchTests`, plus the new classes (24 classes, 273 tests, rerun after the review round).
- Reporter's reproducer (https://github.com/L7R7/spring-kafka-async-shutdown-repro) against the first revision of this change applied to `4.1.x` (`4.1.2-SNAPSHOT`, Boot 4.1.0), property enabled through a `ContainerCustomizer`, `shutdownTimeout` 3s: the suspend listener parked at shutdown is cancelled after 3s and never reaches `resource.use()`; the CompletableFuture listener released 1s into the shutdown completes before the resource is destroyed. The unmodified reproducer tests (property off) still show the original race.


HDPark95 (commented):
awaitAsyncResults deadlines at stopRequestedAt+shutdownTimeout, the same window the blocking stop() waits on (AbstractMessageListenerContainer L634, used by KafkaListenerEndpointRegistry.stop()), so on timeout stop() expires as the wait gives up and cancel/commitPendingAcks/close run after it returns. Measured, shutdownTimeout 2000, work parked: stop() returned at 2000ms, consumer thread at 2015ms; property off, 407ms and already done. An offset acked during the wait committed only after the return. The 8 new tests all use stop(Runnable), so the blocking path is uncovered.

L7R7:
I might have missed this in the diff, but is it possible to set the new configuration via properties? Or do I have to set it in code?

nikita-kibitkin:
@sobychacko while testing the blocking stop path I found a pre-existing bug: a pending `consumer.wakeup()` aborts the next `commitSync`, and `commitIfNecessary` drops the commit because `buildCommits()` has already cleared the offsets. Filed #4622 with the details. The third commit here fixes it with a single retry, since the new blocking stop test hit it in about one run out of six; happy to split it into its own PR if you prefer, in which case that PR needs to land before this one, the drain commits here sit on the same wakeup.

nikita-kibitkin:
@HDPark95 confirmed, thanks for measuring it. Two changes pushed:

- The drain now commits as results complete instead of once after the wait: the loop wakes on each completion, processes the queued acks and commits them, so an offset acked during the wait is committed well before the blocking `stop()` returns. What is left after the deadline is the cancellation of the outstanding results and `consumer.close()`, the same tail a blocking listener that runs past `shutdownTimeout` has today.
- While adding the test for this I found a pre-existing bug: a pending `consumer.wakeup()` (an ack from another thread while paused for out of order acks, or a stop request landing after the poll returned) aborts the next `commitSync`, and `commitIfNecessary` dropped the commit after `buildCommits()` had cleared the offsets; the blocking stop test hit it in about one run out of six. Details in #4622. Fixed here in the third commit: `commitIfNecessary` retries the commit once after a `WakeupException`, and `processAck` / `processAcks` skip the wakeup once the container is stopping. `commitRetriedAfterWakeup` in `ConcurrentMessageListenerContainerMockTests` covers the retry.

New test `blockingStopCommitsCompletedResultsBeforeReturning`: two partitions, one result released 500ms into a blocking `stop()`, the other never; a `ConsumerInterceptor` records the commit time. Asserts the commit happened within 1.5s of the stop request, before `stop()` returned, and the other partition was not committed. Fails without the in-loop commit.

nikita-kibitkin:
@L7R7 in code for now. Boot's `spring.kafka.listener.*` does not know this property (it does not expose `shutdownTimeout` either), so it is either `factory.getContainerProperties().setAwaitAsyncResultsOnStop(true)` on the factory bean, or a `ContainerCustomizer` bean, which Boot applies to every container it builds; that is what I used to run your reproducer. A `spring.kafka.listener.await-async-results-on-stop` property would be a Spring Boot change once this is released, I can open that issue there.

L7R7:
@nikita-kibitkin yes, it would be great if you could open the issue to make sure that the property is added to Spring Boot. I think setting the property in code is fine as a start, but having the property would help in making it easy for users to adopt the new behavior

nikita-kibitkin:
@L7R7 filed spring-projects/spring-boot#51561. Agreed, code works as a start and the property is the right end state; I'll pick it up once a spring-kafka release with the property is out.

sobychacko:
Thanks for the PR, @nikita-kibitkin. This looks really good, and thanks for validating against the reporter's reproducer. The approach is exactly what we talked through on the issue.

Before this goes further, one thing to untangle: we can't put API changes on the maintenance branches, so this one's main-only. Let's bump `@since` to `4.2` and trim the docs to just 4.2 (drop the 4.0.8/4.1.2 mentions).

The `WakeupException` commit-retry piece for #4622 is separate though -- that's an internal fix, no new API, and the bug's on 4.1.x and 4.0.x too. Can you split that into its own PR so we can backport it properly? The `isRunning()` guard in processAck/processAcks should ride along with it.

I'll go through the container and adapter details next, but wanted to get the split and versioning sorted first since it decides how you'll want the commits.

sobychacko:
@nikita-kibitkin Any updates on the PR?

nikita-kibitkin:
Done, both. `@since` is 4.2 and the docs mention 4.2 only. The #4622 fix (the retry in `commitIfNecessary` plus the `isRunning()` guard in `processAck` / `processAcks`, with its mock test) is now #4682 against main for backporting. This branch is rebased on top of that fix until it lands, since the blocking stop test here needs the retry to stay stable; I'll rebase on `main` right after. Commits here are the feature and the "commit as results complete" change from the first round, history rewritten as you suggested.

sobychacko:
Thanks @nikita-kibitkin! Now that #4682 is merged, please rebase this branch on main. The GH-4622 commit is identical to what's now in main, so it'll drop out automatically and leave just the two feature commits.

One thing worth addressing while you're at it: in `wrapUp()`, `awaitAsyncResults()` is inside the `if (!this.fatalError) ` block, so on a fatal error the in-flight results are neither awaited nor cancelled. They keep running against the consumer we're about to `close()`, which is the race this feature is meant to prevent. Could you cancel (not await, as there is no point waiting on the fatal path) the outstanding results before `consumer.close()` so no handles are left dangling?

Also a small style note for future PRs: we've been leaning toward the argument-less `mock()` API where the type is inferable from the variable -  e.g. `Consumer<?, ?> consumer = mock();` rather than `mock(Consumer.class)`. Not a blocker; feel free to apply it to the new tests here. Everything else looks good. Thanks, again.

nikita-kibitkin:
Thanks for the review, @sobychacko! Rebased on main, the #4682 commit dropped out; the two first-round commits are left and the two below are on top.

Fatal path: `wrapUp()` now cancels the outstanding results on the fatal branch too, before `emergencyStop` and `consumer.close()`, without waiting. The cancellation moved into `cancelAsyncResults(reason)`, shared with the timeout path. `inFlightAsyncResultsCancelledOnFatalError` in `ConcurrentMessageListenerContainerMockTests` covers it: a pending `CompletableFuture` from the listener, then `GroupAuthorizationException` from `poll()` with no retry interval; asserts the result is cancelled and that this happens before `close()`. Fails without the change.

Applied the argument-less `mock()` in the new tests where the type comes from the variable; the inline `mock(X.class)` arguments stayed, `onMessage` has overloads there. Ready for the next pass.

sobychacko:
@nikita-kibitkin Looks good to me. Could you do one final rebase against the latest `main`? There are some merge conflicts with the latest changes on `main`. Once that is resolved, we could hopefully merge this PR tomorrow. Thanks!

nikita-kibitkin:
@sobychacko done, rebased on the latest main. The conflict was in `handleResult` with #4677: the container-cancelled check now comes before the try/catch #4677 added around the success path. Same four commits, nothing else changed. Thanks!

sobychacko:
@nikita-kibitkin Great enhancement changes. The PR has been merged to main. Thanks for all your efforts!

PR #4625, opened 2026-09-03
GH-4623: Fix multi-value headers with JSON types

### LINKED ISSUE #4623: Multi-value header mapping loses values on inbound when a JSON type header is present

`main` (4.2.0-SNAPSHOT); the defect is also present in 4.x releases with multi-value header support.

**Describe the bug**

The multi-value header documentation says that on the consumer side, values of headers matching a configured multi-value pattern are "collected into a `List`". That works only for raw `byte[]` headers.

`JsonKafkaHeaderMapper.toHeaders` routes headers *without* a `spring_json_header_types` entry through `AbstractKafkaHeaderMapper.fromUserHeader`, which contains the multi-value collection logic. But any non-`byte[]` outbound value (String, Integer, POJO, ...) always produces a JSON-types entry, and `populateJsonValueHeader` uses `headers.put(...)` in every branch — so each successive header with the same key overwrites the previous one. The same applies to the deprecated `DefaultKafkaHeaderMapper`.

**To Reproduce**

Configure a mapper with a multi-value pattern (e.g. `test-multi-*`), map outbound `MessageHeaders` containing `List.of("value1", "value2")` for `test-multi-value-string`, then map the resulting record headers back inbound.

**Expected behavior**

The inbound header value is `List.of("value1", "value2")`.

Actual: the inbound value is `"value2"` (the last value only).

I will submit a PR with a failing test and fix.

harrisleesh:
Fixes https://github.com/spring-projects/spring-kafka/issues/4623

`JsonKafkaHeaderMapper.toHeaders` only routed headers **without** a `spring_json_header_types` entry through `AbstractKafkaHeaderMapper.fromUserHeader`, which holds the multi-value collection logic. Any non-`byte[]` outbound value always produces a JSON types entry, and `populateJsonValueHeader` used `headers.put(...)` in every branch, so successive values of a multi-value header overwrote each other — only the last value survived the inbound mapping, contradicting the documented behavior ("collected into a `List`"). The deprecated `DefaultKafkaHeaderMapper` had the same defect and is fixed for parity.

A `fromUserHeader` overload accepting the already-converted value is extracted in `AbstractKafkaHeaderMapper` (`@since 4.2`), and all `populateJsonValueHeader` branches route through it in both mappers.

Added a round-trip regression test (String and Integer multi-value headers) that fails on `main` and passes with the fix; `JsonKafkaHeaderMapperTests` 39/39, `SimpleKafkaHeaderMapperTests` 10/10, checkstyle passes.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/AbstractKafkaHeaderMapper.java
@@ -307,14 +307,26 @@ public abstract class AbstractKafkaHeaderMapper implements KafkaHeaderMapper {
 	 * @since 4.0
 	 */
 	protected void fromUserHeader(String headerName, Header header, final Map<String, Object> headers) {
+		fromUserHeader(headerName, headerValueToAddIn(header), headers);
+	}
+
+	/**
+	 * Add the already converted header value to the target headers, collecting values
+	 * for headers matching a multi-value pattern into a {@link List}.
+	 * @param headerName the header name.
+	 * @param headerValue the converted header value.
+	 * @param headers the target headers.
+	 * @since 4.2
+	 */
+	protected void fromUserHeader(String headerName, @Nullable Object headerValue, final Map<String, Object> headers) {
 		if (!doesMatchMultiValueHeader(headerName)) {
-			headers.put(headerName, headerValueToAddIn(header));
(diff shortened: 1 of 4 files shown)


sobychacko:
Thanks for the PR! It has been merged via https://github.com/spring-projects/spring-kafka/commit/f2ccb74e78863d7841019058882550141947be57.

PR #4626, opened 2026-09-03
GH-4624: Fix most-specific delegate selection

### LINKED ISSUE #4624: DelegatingByTypeSerializer assignable mode can select a more general delegate due to a non-transitive comparator

`main` (4.2.0-SNAPSHOT), since the fix for GH-3953.

**Describe the bug**

The GH-3953 fix ordered the delegate map with `DELEGATES_ASSIGNABILITY_COMPARATOR` (a `TreeMap`) so that "the most specific matching class is selected" in assignable mode. That comparator is not transitive: with an interface `A`, an unrelated class `M`, and `Z implements A` where the class names order `A < M < Z`, we get the cycle `Z < A` (assignability), `A < M` and `M < Z` (name order). `TreeMap` behavior is undefined with a non-transitive comparator; concretely, inserting in the order `M, A, Z` yields the in-order iteration `A, M, Z`, so `findDelegate` returns the **interface** delegate for a `Z` instance even though an exact `Z` delegate is registered — violating the constructor javadoc guarantee.

**To Reproduce**

Register delegates in a `LinkedHashMap` in the order `MiddleUnrelated`, `AwareIface`, `ZebraImpl` (with `ZebraImpl implements AwareIface` and names ordered `AwareIface < MiddleUnrelated < ZebraImpl`), construct `DelegatingByTypeSerializer(map, true)`, and serialize a `ZebraImpl` instance.

**Expected behavior**

The exact `ZebraImpl` delegate is selected. Actual: the `AwareIface` delegate is selected.

Because no total order exists for assignability, sorting cannot fix this in general. I will submit a PR that replaces the sort with an exact-match lookup plus a most-specific-assignable scan, preserving the GH-3953 guarantee (its regression test still passes).

harrisleesh:
Fixes https://github.com/spring-projects/spring-kafka/issues/4624

The delegate map in `DelegatingByTypeSerializer` was ordered with an assignability comparator that is not transitive (assignability with a class-name fallback), so `TreeMap` iteration order was undefined. In assignable mode the lookup could return a more general delegate — even when an exact delegate for the value's class was registered (interface `A`, unrelated `M`, `Z implements A` with names `A < M < Z` inserted as `M, A, Z` iterates `A, M, Z`).

Since no total order exists for assignability, the sorted map is replaced with a `LinkedHashMap` and `findDelegate` now does an exact-class lookup first, then a scan keeping the most specific assignable key. This preserves the GH-3953 guarantee — its regression test (`shouldOrderDelegatesSoChildComesBeforeParent`) still passes.

Added a regression test with the cycle-producing fixture that fails on `main` and passes with the fix; `DelegatingSerializationTests` 12/12, checkstyle passes.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
@@ -16,10 +16,9 @@
 
 package org.springframework.kafka.support.serializer;
 
-import java.util.Comparator;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.TreeMap;
 
 import org.apache.kafka.common.errors.SerializationException;
 import org.apache.kafka.common.header.Headers;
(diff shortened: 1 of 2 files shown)


sobychacko:
Thanks for the PR! Merged via https://github.com/spring-projects/spring-kafka/commit/6c80b79505f362c69d7e400c9905332c5f6f87fc.

PR #4639, opened 2026-09-06
GH-2233: Add batch-to-record fallback advice

### LINKED ISSUE #2233: Switch from batch to record-by-record processing in case an exception occured.
**Current Behavior**

When processing serveral records at the same time, it is sometimes impossible to say which record caused an exception and therefore you cannot rise a BatchListenerFailedException. In consequence the listener will reprocess the whole batch again according to https://docs.spring.io/spring-kafka/reference/html/#retrying-batch-eh

**Expected Behavior**

It would be helpful if one could switch to a record-by-record processing in case an exception occured during batch processing. Then one could detect which record caused the exception. 

**Context**
I opened also a question on stackoverflow: https://stackoverflow.com/questions/71856309/spring-kafka-errorhandling-for-batchlisteners

goutamadwant:
Fixes #2233.

Add `BatchToRecordFallbackAdvice` for batch listeners that cannot identify which record caused a failure. A successful batch is invoked once. After a failure, the advice invokes the same listener with singleton lists in order and reports the first failed record through `BatchListenerFailedException`. `DefaultErrorHandler` retains responsibility for offsets, retries, recovery, and redelivery.

The advice is opt-in through the container advice chain and reruns downstream advice for each singleton invocation. It preserves existing identified failures and bypasses manual acknowledgments, asynchronous adapters, and invocations with active transactions or thread-bound Spring resources. Documentation explains the synchronous, idempotent processing requirement and the extra poll-interval budget.

Tests cover converted listeners, cross-partition offset recovery, filtering, advice ordering, manual acknowledgments, nested asynchronous adapters, and Kafka transaction rollback with synchronization enabled and disabled.

Validation: `./gradlew test :spring-kafka:checkstyleMain :spring-kafka:checkstyleTest api --max-workers=2 --no-build-cache --rerun-tasks`. 1,195 tests passed, 9 skipped, and no failures. Checkstyle and API documentation passed.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchToRecordFallbackAdvice.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/BatchToRecordFallbackAdvice.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.listener;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.aopalliance.intercept.MethodInterceptor;
+import org.aopalliance.intercept.MethodInvocation;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.aop.ProxyMethodInvocation;
+import org.springframework.kafka.listener.adapter.AsyncRepliesAware;
+import org.springframework.kafka.support.Acknowledgment;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+
+/**
+ * Advice that invokes a batch listener with singleton lists after a batch fails.
+ * The first singleton failure is reported as a {@link BatchListenerFailedException},
+ * allowing the container's error handler to recover that record and redeliver the
+ * remainder. An existing {@code BatchListenerFailedException} is propagated unchanged.
+ *
+ * <p>Add this advice to {@link ContainerProperties#setAdviceChain} before any advice
+ * that should also apply to each singleton invocation. Listeners must process records
+ * synchronously and be idempotent: records processed before the original batch failure
+ * can be processed again. Successful singleton processing must be complete before the
+ * listener returns. Account for the additional processing time when configuring
+ * {@code max.poll.interval.ms} and {@code max.poll.records}.
+ *
+ * <p>Fallback is not applied to record listeners, {@code ConsumerRecords} listeners,
+ * asynchronous listener adapters, invocations with a manual {@link Acknowledgment},
+ * or invocations with an active Spring transaction or thread-bound Spring resources.
+ * Such invocations retain their
+ * normal behavior. In particular, this advice cannot recover within a failed transaction
+ * or safely reuse a whole-batch acknowledgment for individual records.
+ *
+ * @author Goutam Adwant
+ *
+ * @since 4.2
+ *
+ * @see DefaultErrorHandler
+ */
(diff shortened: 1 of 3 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/BatchToRecordFallbackAdvice.java
```diff
@@ -0,0 +1,136 @@
...
+					|| hasCause(ex, BatchListenerFailedException.class) || isInterruptedOrError(ex)) {
+				throw ex;
+			}
+			if (records.size() == 1) {
+				throw new BatchListenerFailedException("Failed to process record", ex,
+						(ConsumerRecord<?, ?>) records.get(0));
+			}
+			for (Object record : records) {
+				@Nullable Object[] singletonArguments = arguments.clone();
+				singletonArguments[0] = new ArrayList<>(List.of(record));
```
artembilan, line 93: Why do we need wrap one (immutable) `List` into another?

spring-kafka/src/main/java/org/springframework/kafka/listener/BatchToRecordFallbackAdvice.java
```diff
@@ -0,0 +1,136 @@
...
+		return TransactionSynchronizationManager.isActualTransactionActive()
+				|| !TransactionSynchronizationManager.getResourceMap().isEmpty();
+	}
+
+	private boolean isInterruptedOrError(Exception ex) {
+		return Thread.currentThread().isInterrupted() || hasCause(ex, InterruptedException.class)
+				|| hasCause(ex, Error.class);
+	}
+
+	private boolean hasCause(Throwable throwable, Class<? extends Throwable> type) {
```
artembilan, line 120: Looks like all these support methods could be `static`.

spring-kafka/src/test/java/org/springframework/kafka/listener/BatchToRecordFallbackAdviceTests.java
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	@SuppressWarnings("unchecked")
+	private BatchMessageListener<String, String> advised(BatchMessageListener<String, String> listener) {
+		ProxyFactory factory = new ProxyFactory(listener);
+		factory.addAdvice(new BatchToRecordFallbackAdvice());
+		return (BatchMessageListener<String, String>) factory.getProxy();
+	}
+
+	private ConsumerRecord<String, String> record(long offset) {
```
artembilan, line 344: Looks like all these `private` methods could be `static`.

spring-kafka/src/main/java/org/springframework/kafka/listener/BatchToRecordFallbackAdvice.java
```diff
@@ -0,0 +1,136 @@
...
+			}
+			if (records.size() == 1) {
+				throw new BatchListenerFailedException("Failed to process record", ex,
+						(ConsumerRecord<?, ?>) records.get(0));
+			}
+			for (Object record : records) {
+				@Nullable Object[] singletonArguments = arguments.clone();
+				singletonArguments[0] = new ArrayList<>(List.of(record));
+				try {
+					proxyInvocation.invocableClone(singletonArguments).proceed();
```
artembilan, line 95: So, when we fail in the listener method, we just swallow an exception and call same method again but as many times as we have records over there?
How can my code distinguish from one state or another?
I don't see how such a "corruption" in the batch listener logic could be helpful...

In my opinion the batch is a unit of work itself, so all or nothing. Like in the case of a single record: or we accept it or retry in case of error.
Not sure why batch should be treated anyhow differently. 🤷

spring-kafka/src/main/java/org/springframework/kafka/listener/BatchToRecordFallbackAdvice.java
```diff
@@ -0,0 +1,136 @@
...
+		catch (Exception ex) {
+			if (records.isEmpty() || isTransactionActive()
+					|| hasCause(ex, BatchListenerFailedException.class) || isInterruptedOrError(ex)) {
+				throw ex;
+			}
+			if (records.size() == 1) {
+				throw new BatchListenerFailedException("Failed to process record", ex,
+						(ConsumerRecord<?, ?>) records.get(0));
+			}
+			for (Object record : records) {
```
artembilan, line 91: I think some INFO log message could be useful here to give a clue that we are switching to per-record processing.
I also wonder if we should set some `ThreadLocal` variable here which could be consulted from the user listener code to determine that current list of records we are being fed is a fallback, not a new data from Kafka topic.
(I took an idea from the `TransactionSynchronizationManager.isActualTransactionActive()` you are mentioning in your code already)

artembilan (changes requested):
Thank you!
Let's see if my review makes sense!

PR #4640, opened 2026-09-06
GH-3448: Guard parent from stale children

### LINKED ISSUE #3448: childContainer of previous run is stopping ConcurrentContainer after a new start

sudhanshubliz:
Fixes https://github.com/spring-projects/spring-kafka/issues/3448

## Problem

A child `KafkaMessageListenerContainer` can still be processing a record after its concurrent parent is stopped. If the parent is immediately restarted and that old child then fails, its error handler receives the parent and can stop the newly started generation of children.

I reproduced this on current `main` at `6c80b79505f362c69d7e400c9905332c5f6f87fc`; the new regression test failed because the restarted parent changed from running to stopped.

## Change

- give each child error handler a parent-container view bound to the parent generation that created it;
- increment the generation whenever the concurrent parent starts a new child set;
- serialize the generation check and mutating operation with the existing lifecycle lock;
- ignore mutations from stale generations while delegating reads and current-generation operations to the parent; and
- keep the change internal, without new public API.

This is narrower than the closed #3537 approach: it does not subclass `AbstractMessageListenerContainer` or change its constructors/visibility. One observable internal-contract change is that an error handler receives a delegating `MessageListenerContainer` view instead of the identical concurrent-parent object; the view delegates the container API and preserves current-generation stop behavior.

## Tests

Failing before / passing after:

- `ConcurrentMessageListenerContainerTests.fencedChildFromPreviousRunDoesNotStopRestartedParent`

Compatibility:

- `ConcurrentMessageListenerContainerMockTests.testCorrectContainerForConsumerError` verifies the current-generation view exposes parent state and can stop the parent with callback completion.
- Complete `ConcurrentMessageListenerContainerMockTests` and `ConcurrentMessageListenerContainerTests`: 43 tests passed.
- `:spring-kafka:checkstyleMain` and `:spring-kafka:checkstyleTest`: passed.
- Full `./gradlew :spring-kafka:test --no-daemon`: passed in 4m06s.

All commits include the DCO `Signed-off-by` trailer.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ChildAwareMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ChildAwareMessageListenerContainer.java
@@ -1,0 +1,225 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.kafka.listener;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.TopicPartition;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.kafka.event.ConsumerStoppedEvent;
+
+/**
+ * A parent container view associated with one generation of child containers. Mutating
+ * operations from an old generation are ignored after the parent has restarted.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ * @author Sudhanshu Ratna Thakur
+ * @since 4.2
+ */
+final class ChildAwareMessageListenerContainer<K, V> implements MessageListenerContainer {
+
+	private final ConcurrentMessageListenerContainer<K, V> parent;
+
+	private final long generation;
+
+	ChildAwareMessageListenerContainer(ConcurrentMessageListenerContainer<K, V> parent, long generation) {
+
+		this.parent = parent;
+		this.generation = generation;
+	}
+
+	@Override
+	public void setupMessageListener(Object messageListener) {
+		this.parent.invokeIfGenerationCurrent(this.generation, () -> this.parent.setupMessageListener(messageListener));
+	}
+
+	@Override
+	public Map<String, Map<MetricName, ? extends Metric>> metrics() {
+		return this.parent.metrics();
+	}
+
+	@Override
+	public ContainerProperties getContainerProperties() {
+		return this.parent.getContainerProperties();
+	}
+
+	@Override
+	@Nullable
+	public Collection<TopicPartition> getAssignedPartitions() {
+		return this.parent.getAssignedPartitions();
+	}
+
+	@Override
+	@Nullable
+	public Map<String, Collection<TopicPartition>> getAssignmentsByClientId() {
+		return this.parent.getAssignmentsByClientId();
+	}
+
+	@Override
+	public void enforceRebalance() {
+		this.parent.invokeIfGenerationCurrent(this.generation, this.parent::enforceRebalance);
+	}
+
+	@Override
+	public void pause() {
+		this.parent.invokeIfGenerationCurrent(this.generation, this.parent::pause);
+	}
+
+	@Override
+	public void resume() {
+		this.parent.invokeIfGenerationCurrent(this.generation, this.parent::resume);
+	}
+
+	@Override
+	public void pausePartition(TopicPartition topicPartition) {
+		this.parent.invokeIfGenerationCurrent(this.generation, () -> this.parent.pausePartition(topicPartition));
+	}
+
+	@Override
+	public void resumePartition(TopicPartition topicPartition) {
+		this.parent.invokeIfGenerationCurrent(this.generation, () -> this.parent.resumePartition(topicPartition));
+	}
+
+	@Override
+	public boolean isPartitionPauseRequested(TopicPartition topicPartition) {
+		return this.parent.isPartitionPauseRequested(topicPartition);
+	}
+
+	@Override
+	public boolean isPartitionPaused(TopicPartition topicPartition) {
+		return this.parent.isPartitionPaused(topicPartition);
+	}
+
+	@Override
+	public boolean isPauseRequested() {
+		return this.parent.isPauseRequested();
+	}
+
+	@Override
+	public boolean isContainerPaused() {
+		return this.parent.isContainerPaused();
+	}
+
+	@Override
+	public boolean isAutoStartup() {
+		return this.parent.isAutoStartup();
+	}
+
+	@Override
+	public void setAutoStartup(boolean autoStartup) {
+		this.parent.invokeIfGenerationCurrent(this.generation, () -> this.parent.setAutoStartup(autoStartup));
+	}
+
+	@Override
+	@Nullable
+	public String getGroupId() {
+		return this.parent.getGroupId();
+	}
+
+	@Override
+	public String getListenerId() {
+		return this.parent.getListenerId();
+	}
+
+	@Override
+	@Nullable
+	public String getMainListenerId() {
+		return this.parent.getMainListenerId();
+	}
+
+	@Override
+	@Nullable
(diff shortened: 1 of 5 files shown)


LokeshAlamuri:
_I have not worked on this project recently. I have reviewed the PR at high level_. Changes are good. This change could fix the reported issue. I am ready to close the reported issue.  But, please consider following things.


**Problem**:

Even though the childContainer (kafkaMessageListenerContainer) is denied to take the control of it's parent container with the recent changes, I doubt it could still create few other issues during the exception flow of the childcontainer. Given changes are done to guard the parent container from fenced child container.

Initially, i thought of fixing the issue in this direction. But, it could further push for new issues.

**Root Cause:**

Reported issue occurred, when the ConcurrentMessageListenerContainer is started before the child containers are actually stopped. Issue is with the API not with the internal design. I accept changing the API is not recommended. 

**_Possible Fix:_**

1) Create 2 configurable mandatory/optional parameters related to starting the `ConcurrentMessageListenerContainer`. 
   a) Time to wait for previous child containers to completely stop before starting again ConcurrentMessageListenerContainer.
   b) fail/continue if previous child containers are not completely stopped with the given timeout value.
   c) Log message with critical warning if opted to continue even if child containers are not completely stopped.
   d) Update the API documentation about the possible error on starting the  `ConcurrentMessageListenerContainer` not properly.
   e) Provide some defaults that does not change the existing api behaviour by proving reasonable Timeout value and continue to start `ConcurrentMessageListenerContainer` even if child containers are not completely stopped.

sudhanshubliz:
Thanks for reviewing this and for outlining the alternative.
AbstractMessageListenerContainer.stop() already waits up to ContainerProperties.shutdownTimeout. The reproducer reaches this race after that wait expires: the parent is restarted while an old child is still unwinding. If restart continues after any additional timeout, the stale child can still affect the new parent generation unless its operations are fenced. Refusing the restart would avoid that, but would introduce a new lifecycle failure mode.
The current change therefore keeps existing stop/start timing and only prevents an old generation’s error-handler container from mutating the restarted parent. Operations from the current generation continue to delegate normally.
I agree that the wrapper surface and error-handler container identity deserve maintainer review. @sobychacko @artembilan, would you prefer the generation-fencing approach in this PR, or a start-time wait/refusal contract? I’m happy to adapt the implementation once you confirm the intended lifecycle semantics.

artembilan:
Hey, @LokeshAlamuri !

Welcome back and thank you for reviewing this for us!

@sudhanshubliz ,

thank you for fixing this complex task!
It is not clear yet if this cause a real problem or just academical research because we see that something is off 😄 .

I didn't look into code change, but that `generation-fencing approach` sounds interesting and probably the most robust.

Again: I don't feel like this is a major blocker, so bear with us when and how we are going to review this and merge (if any).

PR #4643, opened 2026-09-07
GH-4634: Prevent ShareConsumerAcknowledgment state regression to RENEW

### LINKED ISSUE #4634: ShareConsumerAcknowledgment state can regress from a terminal type to RENEW

`4.2.0-SNAPSHOT` (`main` at `ab8c2ec5`), and 4.1.x. 
The unconditional write dates from `7e645840`, the original share-consumer acknowledgment support, 
but it only became reachable once `b0bfbe4c` (GH-4322, 4.1.0-M2) added `RENEW`, 
because that is the first time two acknowledgments for one record can be in flight.

I found this by reading the `ShareConsumerAcknowledgment` state handling rather than from a production incident. The window is narrow (see below), so this is a report about the state machine being able to 
move backwards, and the reproduction below is unit level.

**Describe the bug**

In `ShareAckMode.MANUAL`, a record's acknowledgment state (`ShareConsumerAcknowledgment.acknowledgmentType`) is written in two places:

1. `acknowledgeInternal()`, on whichever thread the listener calls the acknowledgment from. This is the authoritative transition: it enforces the contract with `compareAndSet` (a terminal type is only accepted from `null` or `RENEW`, otherwise `IllegalStateException`) and then queues the acknowledgment for the consumer thread.

2. `notifyAcknowledged()`, called by `processQueuedAcknowledgments()` on the consumer thread when the queued entry is drained. It writes unconditionally, with the type the entry had when it was queued:

```java
void notifyAcknowledged(AcknowledgeType type) {
	this.acknowledgmentType.set(type);
}
```

The second write never moves the state forward, since (1) already did, but it can move it backwards. 
If a `RENEW` and a terminal acknowledgment are queued before the consumer thread drains:

```
listener : renew()        CAS null  -> RENEW    queue [RENEW]
listener : acknowledge()  CAS RENEW -> ACCEPT   queue [RENEW, ACCEPT]     state = ACCEPT
consumer : drain RENEW    notifyAcknowledged(RENEW)                       state = RENEW   <-- regressed
consumer : drain ACCEPT   notifyAcknowledged(ACCEPT)                      state = ACCEPT
```

While the state is regressed, a second terminal call from any thread that still holds the acknowledgment passes the `compareAndSet(RENEW, type)` guard and is queued, so an already-`ACCEPT`ed record 
gets a second `consumer.acknowledge(record, REJECT)`. 
The caller does not get the `IllegalStateException` the contract promises; 
the Kafka client's rejection is caught in the drain loop and surfaces only as an error log.

The window is the gap between the two drain iterations on the consumer thread, 
and hitting it needs a `renew()` followed by a terminal acknowledgment before a drain, 
plus a second terminal call from another thread in that gap. I do not claim this is common. 
The point is that `notifyAcknowledged()` can only ever be a no-op or a downgrade, 
so the guard in `acknowledgeInternal()` is not a reliable invariant.

`isAcknowledged()` is only used by `toString()`, so the regression is observable purely through the guard.

**To Reproduce**

Both tests fail on `main` today. The container is stopped before the assertions so the consumer thread is gone and every step is deterministic; `notifyAcknowledged(RENEW)` is invoked directly because 
it is exactly the call `processQueuedAcknowledgments()` makes for the stale `RENEW`, 
and the regressed state cannot be observed from outside a full drain.

```java
@Test
void terminalAcknowledgmentIsNotDowngradedByStaleRenewNotification() throws Exception {
	ShareAcknowledgment ack = captureManualAcknowledgment();   // MANUAL mode, mock ShareConsumer, container stopped
	ack.renew();
	ack.acknowledge();
	assertThat(KafkaTestUtils.getPropertyValue(ack, "acknowledgmentType", AtomicReference.class).get())
			.isEqualTo(AcknowledgeType.ACCEPT);

	ReflectionTestUtils.invokeMethod(ack, "notifyAcknowledged", AcknowledgeType.RENEW);

	assertThat(KafkaTestUtils.getPropertyValue(ack, "acknowledgmentType", AtomicReference.class).get())
			.isEqualTo(AcknowledgeType.ACCEPT);
	// expected: accept
	//  but was: renew
}

@Test
void secondTerminalAcknowledgmentIsStillRejectedAfterStaleRenewNotification() throws Exception {
	ShareAcknowledgment ack = captureManualAcknowledgment();
	ack.renew();
	ack.acknowledge();
	ReflectionTestUtils.invokeMethod(ack, "notifyAcknowledged", AcknowledgeType.RENEW);

	assertThatIllegalStateException().isThrownBy(ack::reject);
	// Expecting code to raise a throwable.
}
```

**Expected behavior**

A terminal acknowledgment is final. Draining a stale `RENEW` must not move the state back, 
and a second terminal call must keep throwing `IllegalStateException` regardless of what is still 
in the queue.

I have the fix ready (make `notifyAcknowledged()` refuse to overwrite a terminal state) together 
with the two tests above. Both tests fail before the change and pass after it, and the existing share consumer unit and integration tests still pass.

Arbaz4Sayyad:
### Summary
Fixes #4634

In `ShareAckMode.MANUAL`, when a `renew()` call is followed by a terminal acknowledgment (`acknowledge()`, `release()`, or `reject()`) before the consumer thread drains the queued acknowledgments, `notifyAcknowledged(RENEW)` on the consumer thread was unconditionally overwriting `acknowledgmentType`, regressing the state back to `RENEW`.

This regression allowed subsequent calls on the already-completed acknowledgment to pass the `compareAndSet(RENEW, type)` guard and queue duplicate terminal acknowledgments to Kafka.

### Changes
- Updated `notifyAcknowledged(AcknowledgeType type)` in `ShareConsumerAcknowledgment` to use `accumulateAndGet` guarded by `isTerminal(current)`.
- Added unit test `shareConsumerAcknowledgmentStateShouldNotRegressFromTerminalToRenew` to `ShareKafkaMessageListenerContainerUnitTests`.

### Checklist
- [x] Contributor Covenant Code of Conduct followed
- [x] Developer Certificate of Origin (DCO) Signed-off (`git commit -s`)
- [x] Unit tests added & passing (`./gradlew :spring-kafka:test`)
- [x] Code conventions adhered to (tabs, formatting)

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
@@ -796,7 +796,7 @@ public class ShareKafkaMessageListenerContainer<K, V>
 			private void acknowledgeInternal(AcknowledgeType type) {
 				if (type == AcknowledgeType.RENEW) {
 					AcknowledgeType current = this.acknowledgmentType.get();
-					if (current == AcknowledgeType.ACCEPT || current == AcknowledgeType.RELEASE || current == AcknowledgeType.REJECT) {
+					if (isTerminal(current)) {
 						throw new IllegalStateException(
 								String.format("Record at offset %d has already been terminally acknowledged with type %s",
 										this.record.offset(), current));
@@ -827,8 +827,13 @@ public class ShareKafkaMessageListenerContainer<K, V>
 						new PendingAcknowledgment<>(this.record, type));
 			}
 
-		void notifyAcknowledged(AcknowledgeType type) {
-				this.acknowledgmentType.set(type);
+			void notifyAcknowledged(AcknowledgeType type) {
+				this.acknowledgmentType.accumulateAndGet(type,
+						(current, next) -> isTerminal(current) ? current : next);
+			}
+
+			private static boolean isTerminal(@Nullable AcknowledgeType type) {
+				return type == AcknowledgeType.ACCEPT || type == AcknowledgeType.RELEASE || type == AcknowledgeType.REJECT;
 			}
 
 			boolean isAcknowledged() {
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
@@ -51,6 +51,7 @@ import org.springframework.kafka.event.ConsumerStoppedEvent;
 import org.springframework.kafka.event.ConsumerStoppedEvent.Reason;
 import org.springframework.kafka.event.ShareConsumerStoppingEvent;
 import org.springframework.kafka.listener.ContainerProperties.ShareAckMode;
+import org.springframework.kafka.support.ShareAcknowledgment;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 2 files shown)


sobychacko:
@Arbaz4Sayyad Thanks for the PR. But this is a duplicate of https://github.com/spring-projects/spring-kafka/pull/4635. Therefore, we need to close this one. Thanks again!

PR #4647, opened 2026-09-08
Apply nullability to `Serializer` subclass

ngocnhan-tran1996:
* Add `@Nullable` to `targetType`
* Remove `@SuppressWarnings("NullAway")`
* Add `@Nullable` to the return value of `serialize`
* Add `@Nullable` to `data`
* `Implement DelegatingByTopicSerializer#serialize`

See https://kafka.apache.org/43/javadoc/org/apache/kafka/common/serialization/Serializer.html

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicSerializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicSerializer.java
@@ -71,9 +71,13 @@ public class DelegatingByTopicSerializer extends DelegatingByTopicSerialization<
 		return instance instanceof Serializer;
 	}
 
+	@SuppressWarnings("unchecked")
 	@Override
 	public byte @Nullable [] serialize(String topic, @Nullable Object data) {
-		throw new UnsupportedOperationException();
+		if (data == null) {
+			return null;
+		}
+		return ((Serializer<Object>) findDelegate(topic)).serialize(topic, data);
 	}
 
 	@SuppressWarnings("unchecked")
--- a/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicSerializer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicSerializer.java
@@ -21,11 +21,16 @@ import java.util.regex.Pattern;
 
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.serialization.Serializer;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.util.Assert;
 
 /**
  * A {@link Serializer} that delegates to other serializers based on a topic pattern.
  *
  * @author Gary Russell
+ * @author Ngoc Nhan
+ *
  * @since 2.8
  *
  */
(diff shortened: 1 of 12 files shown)


spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
```diff
@@ -88,19 +90,18 @@ public void configure(Map<String, ?> configs, boolean isKey) {
...
 		}
 		Serializer<Object> delegate = findDelegate(data);
 		return delegate.serialize(topic, data);
 	}
 
-	@SuppressWarnings("NullAway") // Dataflow analysis limitation
 	@Override
-	public byte[] serialize(String topic, Headers headers, Object data) {
+	public byte @Nullable [] serialize(String topic, Headers headers, @Nullable Object data) {
+		Assert.notNull(headers, "'headers' cannot be null");
```
sobychacko: Do we need the Assert.notNull(headers, ...) here? Since the parameter is now non-null, NullAway should be satisfied without it. My only concern is that the assert changes behavior; previously null headers were tolerated, so this could break passive compatibility. If we do want to enforce the Kafka contract, would it make sense to split it into ts own commit with a release note, rather than including it in the nullability polish? Same thing applies to the other serializers here, so whatever we decide, let's apply it consistently across all of them.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicSerializer.java
```diff
@@ -66,14 +71,19 @@ protected boolean isInstance(Object instance) {
...
 
+	@SuppressWarnings("unchecked")
 	@Override
-	public byte[] serialize(String topic, Object data) {
-		throw new UnsupportedOperationException();
+	public byte @Nullable [] serialize(String topic, @Nullable Object data) {
+		if (data == null) {
+			return null;
+		}
+		return ((Serializer<Object>) findDelegate(topic)).serialize(topic, data);
```
sobychacko, line 78: This one used to throw `UnsupportedOperationException` and now it's implemented. That's a behavior change beyond the nullability polish, so could we call it out in the PR description / release notes? Also just to confirm: the asymmetry with DelegatingSerializer (which still throws here) is intentional, right? That one selects the delegate from a header so it can't work without headers, whereas by-topic and by-type can.

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/ToStringSerializer.java
```diff
@@ -81,22 +83,24 @@ else if (config instanceof String) {
 	}
 
 	@Override
-	public byte[] serialize(String topic, @Nullable T data) {
-		return serialize(topic, null, data);
-	}
+	public byte @Nullable [] serialize(String topic, @Nullable T data) {
```
sobychacko: Minor nit: could we drop the blank line here?

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/ToStringSerializer.java
```diff
@@ -81,22 +83,24 @@ else if (config instanceof String) {
...
 			return null;
 		}
 
-		if (this.addTypeInfo && headers != null) {
+		return data.toString().getBytes(this.charset);
+	}
+
+	@Override
+	public byte @Nullable [] serialize(String topic, Headers headers, @Nullable T data) {
+
```
sobychacko: Please remove the blank line here.

sobychacko (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTopicSerializer.java
ngocnhan-tran1996: Yes, `DelegatingSerializer` still throws `UnsupportedOperationException`.

I implemented `DelegatingByTopicSerializer` because, according to the documentation, the content does not require headers.
https://docs.spring.io/spring-kafka/reference/kafka/serdes.html#by-topic

spring-kafka/src/main/java/org/springframework/kafka/support/serializer/DelegatingByTypeSerializer.java
ngocnhan-tran1996: I'm not sure whether we should enforce the Kafka contract here. I'll revert the `Assert.notNull(headers, ...)` changes for now.

PR #4648, opened 2026-09-08
Apply nullability annotations

ngocnhan-tran1996:
* Apply nullability to `ListenerMetadata` and related classes
* Remove `@Nullable` from local variables
* Use `type @Nullable []` instead of `@Nullable type[]` for nullable arrays
* Apply nullability to the `ShareKafkaMessageListenerContainer` inner class
* Remove a redundant expression in the `ShareKafkaMessageListenerContainer` inner class

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
@@ -478,9 +478,8 @@ public class ShareKafkaMessageListenerContainer<K, V>
 			}
 
 			String[] topics = containerProperties.getTopics();
-			if (topics != null) {
-				this.consumer.subscribe(Arrays.asList(topics));
-			}
+			Assert.notNull(topics,  "topics cannot be null");
+			this.consumer.subscribe(Arrays.asList(topics));
 		}
 
 		/**
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
@@ -424,7 +424,7 @@ public class ShareKafkaMessageListenerContainer<K, V>
 
 		private final long ackTimeoutMs;
 
-		ShareListenerConsumer(GenericMessageListener<?> listener, String consumerClientId) {
+		ShareListenerConsumer(GenericMessageListener<?> listener, @Nullable String consumerClientId) {
 			this.genericListener = listener;
 			this.clientId = consumerClientId;
 			ContainerProperties containerProperties = getContainerProperties();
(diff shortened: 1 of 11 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -477,7 +477,10 @@ else if (!this.syncShareCommits) {
 				});
 			}
 
-			this.consumer.subscribe(Arrays.asList(containerProperties.getTopics()));
+			String[] topics = containerProperties.getTopics();
+			if (topics != null) {
```
sobychacko: If topics is null here the consumer starts up without subscribing to anything, which is harder to diagnose than the NPE it replaces. Would you mind using `Assert.notNull(topics,  "...")` instead? That way NullAway is happy and we still fail fast on a misconfigured container.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
```diff
@@ -424,7 +424,7 @@ private class ShareListenerConsumer implements Runnable {
 
 		private final long ackTimeoutMs;
 
-		ShareListenerConsumer(GenericMessageListener<?> listener, String consumerClientId) {
+		ShareListenerConsumer(GenericMessageListener<?> listener, @Nullable String consumerClientId) {
```
sobychacko, line 428: `determineClientId()` always returns a non-null String (it falls back to `getBeanName()`), so this parameter can never actually receive null. I'd keep it as `String consumerClientId` - the `@Nullable` on the field is fine, but advertising nullability on the parameter is a bit misleading.

sobychacko (changes requested):

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
ngocnhan-tran1996: If this is not intentional, `consumerClientId` should be non-null, so we don't need this null check.

https://github.com/spring-projects/spring-kafka/blob/f5cb53a3dbca0513b3dc51a3d955d838228a2e9b/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java#L194-L196

https://github.com/spring-projects/spring-kafka/blob/f5cb53a3dbca0513b3dc51a3d955d838228a2e9b/spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java#L236-L240

sobychacko:
@ngocnhan-tran1996 Please add your name an author to the classes you changed.

spring-kafka/src/main/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainer.java
sobychacko: Good point — you're right. I'd forgotten getBeanName() is itself `@Nullable`.

PR #4652, opened 2026-09-10
GH-4394: Add record converter to share factory

### LINKED ISSUE #4394: Add message converter support for ShareKafkaListenerContainerFactory
## Expected Behavior
`ShareKafkaListenerContainerFactory` should support message converters the same way `AbstractKafkaListenerContainerFactory` does — allowing users to configure a `RecordMessageConverter` for record listeners, wired through to the `MessagingMessageListenerAdapter` so that message conversion and result handling work correctly for Share consumer listeners.

## Current Behavior
`ShareKafkaListenerContainerFactory.createListenerContainer()` always passes null as the message converter when calling `endpoint.setupListenerContainer(instance, null)`. This means:

The `MessagingMessageListenerAdapter` is created with no converter — raw message conversion does not work for Share consumer listeners Return values from listener methods are silently discarded for Share consumers in `MessagingMessageListenerAdapter `(line 414) with the comment // For share consumers, we don't handle results yet

Features like @ SendTo and reply handling are completely unavailable for Share consumers

This is also explicitly documented as a known limitation in kafka-queues.adoc under Current Limitations.

## Context
The regular Kafka listener stack (`AbstractKafkaListenerContainerFactory`) has full message converter support — it selects between `batchMessageConverter` and `recordMessageConverter` based on the listener type and passes it through the entire chain down to the adapter. The Share consumer stack has none of this infrastructure yet.
`ShareKafkaListenerContainerFactory` does not extend `AbstractKafkaListenerContainerFactory` — it is a standalone implementation — so none of the converter fields or wiring is inherited. The gap needs to be filled explicitly.

## Proposed Approach
The fix follows the exact same pattern already established in AbstractKafkaListenerContainerFactory:

Add recordMessageConverter field to ShareKafkaListenerContainerFactory with a setter method
Replace the hardcoded null in createListenerContainer() with the converter:

```java
// Before
endpoint.setupListenerContainer(instance, null);


// After
endpoint.setupListenerContainer(instance, this.recordMessageConverter);
```
Handle result forwarding in MessagingMessageListenerAdapter for Share consumers once the converter is in place

The scope is intentionally kept minimal and focused — matching the existing patterns in the codebase without introducing new abstractions.

Happy to submit a PR once this approach is confirmed or any other approach that sounds suitable for solving the problem

sudhanshubliz:
## Summary

- add `RecordMessageConverter` configuration to `ShareKafkaListenerContainerFactory`
- propagate the configured converter to `KafkaListenerEndpoint` setup
- preserve and test the existing default behavior when no converter is configured

This is the bounded first increment discussed in the issue. Result forwarding and `@SendTo` support remain separate follow-up work.

Fixes #4394

## Validation

- Added regression coverage that fails before the change because the Share factory has no converter setter.
- `ShareKafkaMessageListenerContainerUnitTests`: 39 tests passed.
- `checkstyleMain` and `checkstyleTest`: passed.
- Full `:spring-kafka:test`: the affected tests passed; three unrelated embedded-broker timing tests failed under the parallel full-suite run and all three passed when rerun independently.

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/config/ShareKafkaListenerContainerFactory.java
@@ -33,6 +33,7 @@ import org.springframework.kafka.listener.ShareConsumerRecordRecoverer;
 import org.springframework.kafka.listener.ShareKafkaMessageListenerContainer;
 import org.springframework.kafka.support.JavaUtils;
 import org.springframework.kafka.support.TopicPartitionOffset;
+import org.springframework.kafka.support.converter.RecordMessageConverter;
 import org.springframework.util.Assert;
 
 /**
@@ -52,6 +53,7 @@ import org.springframework.util.Assert;
  *
  * @author Soby Chacko
  * @author Kumar Gaurav
+ * @author Sudhanshu Ratna Thakur
  *
  * @since 4.0
  */
@@ -70,6 +72,8 @@ public class ShareKafkaListenerContainerFactory<K, V>
 	private int concurrency = 1;
 
 	private @Nullable ShareConsumerRecordRecoverer recordRecoverer;
+
+	private @Nullable RecordMessageConverter recordMessageConverter;
 
 	@SuppressWarnings("NullAway.Init")
 	private ApplicationEventPublisher applicationEventPublisher;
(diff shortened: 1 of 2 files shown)


sobychacko:
Merged via https://github.com/spring-projects/spring-kafka/commit/51aa7edd8b1f409b8720a19f3aff3ed60618e02a. Thanks!

PR #4655, opened 2026-09-10
GH-4654: Allow configuring a KafkaAdmin on the container factory

### LINKED ISSUE #4654: AbstractKafkaListenerContainerFactory should allow to configure a KafkaAdmin
In https://github.com/spring-projects/spring-boot/issues/38830 we're trying to add support for configuring a `KafkaAdmin` on consumers and producers to help with retrieving the cluster ID. 

Looking at `AbstractKafkaListenerContainerFactory` there is no way to set the `KafkaAdmin`. The only way I found is to set a `ConsumerFactory`, which forces me to wrap the one that would have been defined and compose it. This looks too much code if setting a `KafkaAdmin` is something that we'd like to offer.

Can you please add a setter for `KafkaAdmin` and configure the instance accordingly? `AbstractMessageListenerContainer` has a setter for `KafkaAdmin` so that looks like a good fit.

sobychacko:
Fixes https://github.com/spring-projects/spring-kafka/issues/4654

Add `setKafkaAdmin(KafkaAdmin)` to `AbstractKafkaListenerContainerFactory` and apply it to each container it creates. Previously the admin could only be set on the containers individually.

### DIFF
--- a/spring-kafka/src/test/java/org/springframework/kafka/annotation/ContainerFactoryTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/annotation/ContainerFactoryTests.java
@@ -95,15 +95,14 @@ public class ContainerFactoryTests {
 		assertThat(container.getContainerProperties().getGroupId()).isEqualTo("myGroup");
 	}
 
-	@SuppressWarnings("unchecked")
 	@Test
 	void kafkaAdminTransferred() {
 		ConcurrentKafkaListenerContainerFactory<String, String> factory =
 				new ConcurrentKafkaListenerContainerFactory<>();
-		factory.setConsumerFactory(mock(ConsumerFactory.class));
-		KafkaAdmin kafkaAdmin = mock(KafkaAdmin.class);
+		factory.setConsumerFactory(mock());
+		KafkaAdmin kafkaAdmin = mock();
 		factory.setKafkaAdmin(kafkaAdmin);
-		ConcurrentMessageListenerContainer<String, String> container = factory.createContainer("foo");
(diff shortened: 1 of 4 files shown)


spring-kafka/src/test/java/org/springframework/kafka/annotation/ContainerFactoryTests.java
```diff
@@ -93,4 +95,16 @@ protected MessagingMessageListenerAdapter<String, String> createMessageListener(
 		assertThat(container.getContainerProperties().getGroupId()).isEqualTo("myGroup");
 	}
 
+	@SuppressWarnings("unchecked")
+	@Test
+	void kafkaAdminTransferred() {
+		ConcurrentKafkaListenerContainerFactory<String, String> factory =
+				new ConcurrentKafkaListenerContainerFactory<>();
+		factory.setConsumerFactory(mock(ConsumerFactory.class));
+		KafkaAdmin kafkaAdmin = mock(KafkaAdmin.class);
```
artembilan: Please, use `mock()` API without arguments.
Much easier to read code.

spring-kafka/src/test/java/org/springframework/kafka/annotation/ContainerFactoryTests.java
```diff
@@ -93,4 +95,16 @@ protected MessagingMessageListenerAdapter<String, String> createMessageListener(
...
+		factory.setKafkaAdmin(kafkaAdmin);
+		ConcurrentMessageListenerContainer<String, String> container = factory.createContainer("foo");
```
artembilan: Any chances that we don't use `foo/bar` language in new code, please?

artembilan (changes requested):

PR #4663, opened 2026-09-11
GH-4660: Require topics for share containers

### LINKED ISSUE #4660: ShareKafkaMessageListenerContainer NPEs at start() when built from a topic pattern or partition list
`AbstractShareKafkaMessageListenerContainer`'s constructor accepts `topicPattern` and `topicPartitions` container properties, not just `topics` (`AbstractShareKafkaMessageListenerContainer.java` lines 99-119):

```java
String @Nullable [] topics = containerProperties.getTopics();
if (topics != null) {
    this.containerProperties = new ContainerProperties(topics);
}
else {
    Pattern topicPattern = containerProperties.getTopicPattern();
    if (topicPattern != null) {
        this.containerProperties = new ContainerProperties(topicPattern);
    }
    else {
        TopicPartitionOffset @Nullable [] topicPartitions = containerProperties.getTopicPartitions();
        if (topicPartitions != null) {
            this.containerProperties = new ContainerProperties(topicPartitions);
        }
        else {
            throw new IllegalStateException("topics, topicPattern, or topicPartitions must be provided");
        }
    }
}
```

But `ShareKafkaMessageListenerContainer.ShareListenerConsumer`'s constructor ends with an unconditional:

```java
this.consumer.subscribe(Arrays.asList(containerProperties.getTopics()));
```

`getTopics()` is `null` when the container was built from a pattern or partition list, so this throws an `NPE` out of `start()` instead of failing at construction time with a clear message.

The underlying Kafka client cannot support either shape: `org.apache.kafka.clients.consumer.ShareConsumer` declares only

```java
void subscribe(Collection<String> topics);
void unsubscribe();
```

— no `subscribe(Pattern)`, no `assign(...)` (verified via `javap` against `kafka-clients-4.3.1`).

`ShareKafkaListenerContainerFactory` already anticipates this and rejects both up front:

```java
public ShareKafkaMessageListenerContainer<K, V> createContainer(TopicPartitionOffset... topicPartitions) {
    throw new UnsupportedOperationException("ShareConsumer does not support explicit partition assignment");
}

public ShareKafkaMessageListenerContainer<K, V> createContainer(Pattern topicPattern) {
    throw new UnsupportedOperationException("ShareConsumer does not support topic patterns");
}
```

But `AbstractShareKafkaMessageListenerContainer`'s constructor — the path used when constructing a container directly, which is exactly what the reference guide's "Basic Container Setup" example shows — does not, and instead defers the failure to `start()` with a bare NPE.

**Suggested fix:** have the constructor throw `UnsupportedOperationException` (reusing the factory's messages) when `topicPattern` or `topicPartitions` is supplied instead of `topics`, so this fails fast and clearly regardless of how the container was created.

kalayciburak:
## Summary

Share consumer containers only support explicit topic names. The constructor now requires `topics` via `Assert.notNull` and copies the rest of the properties onto a topics-based `ContainerProperties`. Pattern and partition shapes fail immediately with `IllegalArgumentException` instead of an NPE at `start()`.

Fixes #4660

## Test plan

Executed:

- `./gradlew :spring-kafka:test --tests org.springframework.kafka.listener.ShareKafkaMessageListenerContainerUnitTests` (39/0)
- `./gradlew :spring-kafka:checkstyleMain :spring-kafka:checkstyleTest`

### DIFF
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
@@ -17,7 +17,6 @@
 package org.springframework.kafka.listener;
 
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.regex.Pattern;
 
 import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
@@ -32,7 +31,6 @@ import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.ApplicationEventPublisherAware;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.kafka.core.ShareConsumerFactory;
-import org.springframework.kafka.support.TopicPartitionOffset;
 import org.springframework.util.Assert;
 
 /**
@@ -48,6 +46,7 @@ import org.springframework.util.Assert;
  *
  * @author Soby Chacko
  * @author Ngoc Nhan
+ * @author Burak Kalayci
  * @since 4.0
  */
 public abstract class AbstractShareKafkaMessageListenerContainer<K, V>
@@ -98,26 +97,9 @@ public abstract class AbstractShareKafkaMessageListenerContainer<K, V>
 		Assert.notNull(shareConsumerFactory, "'shareConsumerFactory' cannot be null");
 		this.shareConsumerFactory = (ShareConsumerFactory<K, V>) shareConsumerFactory;
 		String[] topics = containerProperties.getTopics();
-		if (topics != null) {
-			this.containerProperties = new ContainerProperties(topics);
-		}
-		else {
-			Pattern topicPattern = containerProperties.getTopicPattern();
-			if (topicPattern != null) {
-				this.containerProperties = new ContainerProperties(topicPattern);
-			}
-			else {
-				TopicPartitionOffset[] topicPartitions = containerProperties.getTopicPartitions();
-				if (topicPartitions != null) {
-					this.containerProperties = new ContainerProperties(topicPartitions);
-				}
-				else {
-					throw new IllegalStateException("topics, topicPattern, or topicPartitions must be provided");
-				}
-			}
-		}
-		BeanUtils.copyProperties(containerProperties, this.containerProperties,
-				"topics", "topicPartitions", "topicPattern");
+		Assert.notNull(topics, "'topics' must be provided");
+		this.containerProperties = new ContainerProperties(topics);
+		BeanUtils.copyProperties(containerProperties, this.containerProperties, "topics");
 	}
 
 	@Override
--- a/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
+++ b/spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
@@ -26,6 +26,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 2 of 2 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
```diff
@@ -48,6 +46,7 @@
  *
  * @author Soby Chacko
  * @author Ngoc Nhan
+ * @author Burak KALAYCI
```
artembilan: Why your last name is in upper case ?
I think we have discussed this with your before.
And I believe that I saw your last name in some commits in the normal international style.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
```diff
@@ -101,20 +100,14 @@ protected AbstractShareKafkaMessageListenerContainer(ShareConsumerFactory<? supe
 		if (topics != null) {
 			this.containerProperties = new ContainerProperties(topics);
 		}
+		else if (containerProperties.getTopicPattern() != null) {
+			throw new UnsupportedOperationException("ShareConsumer does not support topic patterns");
+		}
+		else if (containerProperties.getTopicPartitions() != null) {
+			throw new UnsupportedOperationException("ShareConsumer does not support explicit partition assignment");
```
artembilan: I don't think this is correct choice for an exception.
How is this an `operation`?
We just reject here some provided property, so `IllegalArgumentException` is the right choice.
See `org.springframework.util.Assert.isNull()` API instead.

spring-kafka/src/main/java/org/springframework/kafka/listener/AbstractShareKafkaMessageListenerContainer.java
```diff
@@ -101,20 +100,14 @@ protected AbstractShareKafkaMessageListenerContainer(ShareConsumerFactory<? supe
...
-			else {
-				TopicPartitionOffset[] topicPartitions = containerProperties.getTopicPartitions();
-				if (topicPartitions != null) {
-					this.containerProperties = new ContainerProperties(topicPartitions);
-				}
-				else {
-					throw new IllegalStateException("topics, topicPattern, or topicPartitions must be provided");
-				}
-			}
+			throw new IllegalStateException("topics, topicPattern, or topicPartitions must be provided");
```
artembilan: Please, give this code more engineering time.
We've just reject both `topicPattern` and `topicPartitions`.
What do they do in this exception message?

I even think the logic with the issue goal becomes very simple:
```
String[] topics = containerProperties.getTopics();
Assert.notNull(topics, "topics must be provided:);
this.containerProperties = new ContainerProperties(topics);
BeanUtils.copyProperties(containerProperties, this.containerProperties, "topics");
```

And we don't need to check for any other props like `topicPattern` or `topicPartitions`.
That would be ignored downstream anyway.
And what we really care is just `topics`.

spring-kafka/src/test/java/org/springframework/kafka/listener/ShareKafkaMessageListenerContainerUnitTests.java
```diff
@@ -184,6 +187,29 @@ void shouldRejectInvalidConcurrency() {
 				.withMessageContaining("concurrency must be greater than 0");
 	}
 
+	@Test
+	void shouldRejectTopicPatternAtConstruction() {
+		ContainerProperties containerProperties = new ContainerProperties(Pattern.compile("test-.*"));
+		containerProperties.setMessageListener(this.messageListener);
```
artembilan: Why do we need `listener` in this unit test scope?
What value does it bring or what aspect of the assertion below does it affect?

artembilan (changes requested):
Thank you!
If you agree with my review comments, just fix them and push.
I'll see our consensus in your commit change set.

artembilan (approved):

artembilan:
@kalayciburak ,

PR #4669, opened 2026-09-12
GH-4535: Permit configuring JsonMapper backing Jackson serdes

### LINKED ISSUE #4535: Permit configuring Deserializer/Serializer that is backing JacksonJsonDeserializer/Serializer
**Expected Behavior**

You can pass properties via binding producer/consumer properties that configure a backing serializer/deserializer for jackson json integration.

**Current Behavior**

Each `JacksonJsonDeserializer/Serializer` creates its own `JsonMapper` from a static method that does not permit passing any properties into it.

**Context**

When using `spring-kafka` with `spring-boot-cloud-stream` users can only provide a class name for the deserializer/serializer, which ends up not being managed by spring application context, and in turn there is no way to perform dependency injection into the created to pass in the properties used by the default spring jackson json mapper. This imposes an issue, where defaults changing between jackson versions affect what can or cannot be sent via spring-kafka integration.

So far my workaround was to create my own extension of the default deserializer/serializer and call my static method which builds my jackson mapper, just like the default implementation does (example below).

```java
class MyJacksonDeserializer extends JacksonJsonDeserializer {
    MyJacksonDeserializer(Class type, boolean useHeaders) {
        super(type, JsonMapper.builder().build(), useHeaders);
    }
}
```

This becomes cumbersome when you have conflicting requirements between topics.

More over, most delegating serializers/deserializers only ever look for default no arg constructor, which further exacerbates the issue. 

Instead of this I would like to pass in a full `spring.jackson` property tree to configure the mapper, or have the serializers/deserializers become managed beans by the context.

arnabnandy7:
### Summary

This pull request introduces configuration-driven and fluent customization for the `JsonMapper` backing `JacksonJsonSerializer`, `JacksonJsonDeserializer`, and `JacksonJsonSerde`.

Previously, customizing mapper configurations required creating explicit subclasses. In declarative and property-driven architectures (such as Spring Cloud Stream Kafka Binder or externalized client properties), subclassing adds unnecessary boilerplate and maintenance overhead.

This change enables declarative and programmatic configuration of the mapper while honoring the immutability design of Jackson 3:

- **Declarative Configuration**: Allows supplying or tailoring the mapper via standard Kafka configuration properties supporting mapper instances, `Supplier<JsonMapper>` types, `Consumer<JsonMapper.Builder>` customizers, or static factory method references.
- **Key & Value Scoping**: Supports topic-level or role-specific mappers via dedicated key and value configuration properties alongside global fallbacks.
- **Fluent API**: Exposes `jsonMapper(...)` configuration on serializer, deserializer, and serde builders.
- **Documentation & Verification**: Updates the reference guide and adds test coverage for all configuration pathways.

Fixes #4535

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/serdes.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka/serdes.adoc
@@ -115,11 +115,19 @@ They have no effect if you have provided `Serializer` and `Deserializer` instanc
 * `JacksonJsonDeserializer.TYPE_MAPPINGS` (default `empty`): See xref:kafka/serdes.adoc#serdes-mapping-types[Mapping Types].
 * `JacksonJsonDeserializer.KEY_TYPE_METHOD` (default `empty`): See xref:kafka/serdes.adoc#serdes-type-methods[Using Methods to Determine Types].
 * `JacksonJsonDeserializer.VALUE_TYPE_METHOD` (default `empty`): See xref:kafka/serdes.adoc#serdes-type-methods[Using Methods to Determine Types].
+* `JacksonJsonSerializer.MAPPER` / `JacksonJsonDeserializer.MAPPER` (`spring.json.mapper`): Custom `JsonMapper` instance, class (`Supplier<JsonMapper>`, `Consumer<JsonMapper.Builder>`, or `JsonMapper`), or static method reference.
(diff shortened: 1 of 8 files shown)


sobychacko:
Thanks for the PR, but we decided to not to merge it due to the complexities involved by adding it. See the issue (#4535) for more details on the reasonings. Thanks again.

PR #4674, opened 2026-09-12
GH-4272: Bound recovery failures per record

### LINKED ISSUE #4272: ConsumerRecordRecoverer failures cause infinite reprocessing
**Problem**: 
When a ConsumerRecordRecoverer implementation continuously fails inside Spring Kafka error handlers, the consumer repeatedly seeks and reprocesses the same offset indefinitely. Current controls (`setResetStateOnExceptionChange(false)` and `setSeekAfterError(false))` do not prevent infinite reprocessing: disabling reset avoids repeated error-handler retries on the next poll, but the consumer still seeks the same message and never progresses.

**Reproduction steps:**

1. Configure a `DefaultErrorHandler` with a `ConsumerRecordRecoverer` that throws on every invocation.
3. Produce a message that constantly fails
4. Observe that the same offset is repeatedly processed, and the consumer does not advance.

**Expected behavior:** 
When a recoverer fails repeatedly, the error handler should provide a configurable terminal behavior instead of repeatedly seeking the same offset forever. like move to next offset (skip/commit)

Please add an explicit configuration option on `DefaultErrorHandler` to skip such messages and continue processing the following messages. I can try to contribute if the changes make sense to you.

BK202503:
Fixes #4272

When a `ConsumerRecordRecoverer` keeps throwing for the same record, `DefaultErrorHandler` seeks back to it and the record is redelivered indefinitely; `resetStateOnRecoveryFailure` and `seekAfterError` change how it is redelivered, not whether. The consumer never advances past the offset.

This adds `FailedRecordProcessor.setMaxRecoveryFailures(int)` (so both `DefaultErrorHandler` and `DefaultAfterRollbackProcessor` get it). Once recovery has failed that many times for a record, the record is logged at ERROR level with the last recovery exception and treated as recovered: it is no longer included in the seeks and its offset is committed according to the ack mode. `RetryListener.recoveryFailed` is still called for every failure. Default is unlimited, i.e. the current behavior.

The count lives next to the retry state in `FailedRecordTracker` rather than inside `FailedRecord`, since that state is discarded on a recovery failure when `resetStateOnRecoveryFailure` is `true` (the default) and the no-retries path has no retry state at all. It is cleared on a successful recovery, when a different offset fails on the partition, and by `clearThreadState` / `clearThreadStateFor`.

Tests cover the default (unbounded), the no-retries path, the count surviving a back-off reset, `resetStateOnRecoveryFailure=false`, per-record/per-partition scoping, the handler-level seek behaviour, and the argument check. Docs and what's-new updated.

### DIFF
--- a/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
+++ b/spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
@@ -22,6 +22,15 @@ In a non-concurrent container with multiple partitions, the container calls this
 This fixes a bug where a healthy batch from one partition would reset the backoff retry counter for an unrelated failing partition, preventing broken records from ever reaching the dead-letter topic.
 Implementations of `CommonErrorHandler` that maintain per-partition state (such as custom subclasses of `FailedRecordProcessor`) may override this method to apply the same scoping.
 
+[[x42-error-handler-max-recovery-failures]]
+=== Bounded Recovery Failures in Error Handlers
+
+`DefaultErrorHandler` and `DefaultAfterRollbackProcessor` (via `FailedRecordProcessor`) now have a `setMaxRecoveryFailures(int)` property.
+When the recoverer keeps throwing for the same record (for example, the dead-letter topic is unavailable), the record was previously seeked to and redelivered indefinitely.
+With the limit set, once recovery has failed that many times for a record, it is logged at `ERROR` level and skipped as if it had been recovered.
+By default, there is no limit.
+See xref:kafka/annotation-error-handling.adoc#default-eh[Default Error Handler].
+
 [[x42-listener-observation-tags]]
 === Listener Observation Meter Tags
 
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
@@ -36,6 +36,7 @@ import org.springframework.util.backoff.FixedBackOff;
  * Common super class for classes that deal with failing to consume a consumer record.
  *
  * @author Gary Russell
+ * @author Bill Kim
  * @since 2.3.1
  *
  */
@@ -123,6 +124,20 @@ public abstract class FailedRecordProcessor extends ExceptionClassifier implemen
 	 */
 	public void setResetStateOnExceptionChange(boolean resetStateOnExceptionChange) {
 		this.failureTracker.setResetStateOnExceptionChange(resetStateOnExceptionChange);
+	}
+
+	/**
+	 * Set the maximum number of times recovery may fail for the same record before the
+	 * record is skipped as if it had been recovered: it is logged at ERROR level, it is
+	 * no longer included in the seeks, and its offset is committed according to the
+	 * container's ack mode. By default there is no limit; a record whose recovery keeps
+	 * failing is redelivered until the recoverer succeeds.
+	 * @param maxRecoveryFailures the maximum number of recovery failures; must be greater than 0.
+	 * @since 4.2
+	 * @see #setResetStateOnRecoveryFailure(boolean)
+	 */
+	public void setMaxRecoveryFailures(int maxRecoveryFailures) {
+		this.failureTracker.setMaxRecoveryFailures(maxRecoveryFailures);
 	}
 
 	/**
--- a/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
+++ b/spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
@@ -44,6 +44,7 @@ import org.springframework.util.backoff.BackOffExecution;
  * Track record processing failure counts.
  *
  * @author Gary Russell
+ * @author Bill Kim
  * @since 2.2
  *
  */
@@ -51,7 +52,11 @@ class FailedRecordTracker implements RecoveryStrategy {
 
 	private final Map<Thread, Map<TopicPartition, FailedRecord>> failures = new ConcurrentHashMap<>();
 
+	private final Map<Thread, Map<TopicPartition, RecoveryFailures>> recoveryFailures = new ConcurrentHashMap<>();
+
 	private final ConsumerAwareRecordRecoverer recoverer;
+
+	private final LogAccessor logger;
 
 	private final boolean noRetries;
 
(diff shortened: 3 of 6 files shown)


spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
```diff
@@ -125,6 +126,20 @@ public void setResetStateOnExceptionChange(boolean resetStateOnExceptionChange)
...
+	 * Set the maximum number of times recovery may fail for the same record before the
+	 * record is skipped as if it had been recovered: it is logged at ERROR level, it is
+	 * no longer included in the seeks, and its offset is committed according to the
+	 * container's ack mode. By default there is no limit; a record whose recovery keeps
+	 * failing is redelivered until the recoverer succeeds.
+	 * @param maxRecoveryFailures the maximum number of recovery failures; must be greater than 0.
+	 * @since 4.2
+	 * @see #setResetStateOnRecoveryFailure(boolean)
+	 */
+	public void setMaxRecoveryFailures(int maxRecoveryFailures) {
```
sobychacko, line 139: Small but important: when the limit is hit we call recoveryFailed but not recovered, even though the record is treated as recovered for the seek/commit. Anyone relying on recovered as their "this record is finally done" hook won't get it. Can we spell that out in the javadoc? Just want to save people a debugging session.

spring-kafka-docs/src/main/antora/modules/ROOT/pages/whats-new.adoc
```diff
@@ -22,6 +22,15 @@ In a non-concurrent container with multiple partitions, the container calls this
 This fixes a bug where a healthy batch from one partition would reset the backoff retry counter for an unrelated failing partition, preventing broken records from ever reaching the dead-letter topic.
 Implementations of `CommonErrorHandler` that maintain per-partition state (such as custom subclasses of `FailedRecordProcessor`) may override this method to apply the same scoping.
 
+[[x42-error-handler-max-recovery-failures]]
+=== Bounded Recovery Failures in Error Handlers
+
+`DefaultErrorHandler` and `DefaultAfterRollbackProcessor` (via `FailedRecordProcessor`) now have a `setMaxRecoveryFailures(int)` property.
+When the recoverer keeps throwing for the same record (for example, the dead-letter topic is unavailable), the record was previously seeked to and redelivered indefinitely.
+With the limit set, once recovery has failed that many times for a record, it is logged at `ERROR` level and skipped as if it had been recovered.
```
sobychacko, line 30: Should mention `recovered` isn't fired, just `recoveryFailed`.

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
```diff
@@ -224,16 +243,20 @@ BackOff determineBackOff(ConsumerRecord<?, ?> record, @Nullable Exception except
 		return backOffToUse != null ? backOffToUse : this.backOff;
 	}
 
-	private void attemptRecovery(ConsumerRecord<?, ?> record, @Nullable Exception exception, @Nullable TopicPartition tp,
+	private void attemptRecovery(ConsumerRecord<?, ?> record, @Nullable Exception exception, TopicPartition tp,
```
sobychacko, line 246: tp went from `@Nullable` to non-null and the no-retries path now passes a real partition. Is this intentional? a small comment would help.

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordProcessor.java
```diff
@@ -125,6 +126,20 @@ public void setResetStateOnExceptionChange(boolean resetStateOnExceptionChange)
 		this.failureTracker.setResetStateOnExceptionChange(resetStateOnExceptionChange);
 	}
 
+	/**
+	 * Set the maximum number of times recovery may fail for the same record before the
+	 * record is skipped as if it had been recovered: it is logged at ERROR level, it is
```
sobychacko, line 131: "before the record is skipped" reads like skip-on-the-third. With maxRecoveryFailures=2 we skip on the second failure. Can we reword so N-then-skip is unambiguous?

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
```diff
@@ -243,15 +266,53 @@ private void attemptRecovery(ConsumerRecord<?, ?> record, @Nullable Exception ex
...
+		if (failures == null || failures.getOffset() != record.offset()) {
+			failures = new RecoveryFailures(record.offset());
+			map.put(tp, failures);
+		}
+		int count = failures.increment();
+		if (count < this.maxRecoveryFailures) {
+			return false;
+		}
+		clearRecoveryFailures(tp);
+		this.logger.error(failure, () -> "Recovery failed " + count + " times for " + KafkaUtils.format(record)
```
sobychacko, line 285: maybe include the `maxRecoveryFailures` too in the messge?

spring-kafka/src/main/java/org/springframework/kafka/listener/FailedRecordTracker.java
```diff
@@ -318,4 +379,24 @@ void setLastException(@Nullable Exception lastException) {
 
 	}
 
+	static final class RecoveryFailures {
```
sobychacko, line 382: Might be good to have an explanation for this class as to why we need to preserve these states?

sobychacko (changes requested):
@BK202503 Nice PR! I added some feedback. Please take a look. One other thing - could you try to use your real name on DCO commits? Thanks.

sobychacko:
Hi @BK202503 - thanks for the PR. With `4.2.0-M2` going out soon I didn't want it to slip, so I picked up the review comments myself and merged it as https://github.com/spring-projects/spring-kafka/commit/edc1fae00e692161d6f51f09c57677eded4f46b1, with you as the author and me as co-author.

Closing in favor of the merged commit. Thanks again.

BK202503:
> Hi @BK202503 - thanks for the PR. With `4.2.0-M2` going out soon I didn't want it to slip, so I picked up the review comments myself and merged it as [edc1fae](https://github.com/spring-projects/spring-kafka/commit/edc1fae00e692161d6f51f09c57677eded4f46b1), with you as the author and me as co-author.
> 
> Closing in favor of the merged commit. Thanks again.

Checked, thanks!

PR #4678, opened 2026-09-13
GH-4677: Report async success-path failures to Observation

### LINKED ISSUE #4677: Async listener reply success path never reports failures to the receiver Observation
`MessagingMessageListenerAdapter.handleResult(...)` (async `CompletableFuture`/`Mono` reply path) has an asymmetry between its success and error branches when reporting to the deferred receiver `Observation`:

https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/listener/adapter/MessagingMessageListenerAdapter.java#L585-L603

```java
completableFutureResult.whenComplete((r, t) -> {
    try (var ignored = observation.openScope()) {
        if (t == null) {
            asyncSuccess(r, replyTopic, source, messageReturnType);   // <-- no catch
            if (isAsyncReplies()) {
                acknowledge(acknowledgment);                          // <-- no catch
            }
        }
        else {
            Throwable cause = t instanceof CompletionException ? t.getCause() : t;
            cause = cause == null ? t : cause;
            observation.error(cause);
            asyncFailure(request, acknowledgment, consumer, cause, source);
        }
    }
    finally {
        observation.stop();
    }
});
```

The error branch (`t != null`) correctly calls `observation.error(cause)` before handling the failure. The success branch has no equivalent `catch`: if `asyncSuccess(...)` (which builds and publishes the reply via `replyTemplate`) or `acknowledge(...)` throws, that exception is never passed to `observation.error(...)`. Because nothing chains onto (or joins) the `CompletableFuture` returned by this `whenComplete` call, the exception is effectively swallowed — and the `finally` still runs `observation.stop()` normally, so the Observation/span is reported as a clean success even though the reply publish or the ack actually failed.

**Failure scenario:** a `@KafkaListener` method returns a `CompletableFuture<?>`/`Mono<?>` with a `@SendTo` reply configured and observation enabled. The future completes successfully, but the reply publish inside `asyncSuccess` fails (e.g. `replyTemplate.send(...)` throws because the topic doesn't exist or the producer is misconfigured). The exception disappears silently, and any tracing/metrics backend shows the operation as successful.

**Suggested fix:** wrap the success branch (`asyncSuccess`/`acknowledge`) in a `try/catch` that calls `observation.error(ex)` and rethrows, mirroring the error branch — or use `Observation.scoped(Runnable)` (which already does exactly this: opens the scope, runs the action, and calls `error(...)` before propagating on any exception) instead of the hand-rolled `try (openScope()) { ... } finally`.

Found while comparing this method against a very similar Micrometer-observation gap in `spring-amqp`'s `AbstractAdaptableMessageListener.handleResult` (spring-projects/spring-amqp#3646) — that PR's reviewer discussion is where this was first noticed; filing separately here since the same pattern predates that PR and is independent of it.

Gaurav1112:
Fixes: #4677

`MessagingMessageListenerAdapter.handleResult(...)`'s async `whenComplete` callback reports listener failures to the deferred receiver Observation via `observation.error(cause)`, but the success branch had no equivalent: if `asyncSuccess(...)` (reply publish via the `replyTemplate`) or `acknowledge(...)` threw, the exception was captured by the dependent `CompletableFuture` and silently dropped, while the shared `finally` still ran `observation.stop()` — so tracing/metrics reported the operation as a clean success even though the reply publish or ack failed.

Per the issue's suggestion, this wraps the success branch in a `try/catch` that calls `observation.error(ex)` and rethrows, exactly mirroring the error branch. The alternative of switching to `Observation.scoped(Runnable)` would have restructured the shared open-scope/`stop()` frame, so the minimal try/catch was chosen.

One design point for review: the catch records and rethrows (per the issue text) rather than routing into `asyncFailure(request, acknowledgment, consumer, ex, source)` as the error branch does. If you'd prefer reply-publish/ack failures to also engage the `asyncFailure` error-handling path, happy to change it — the test covers either shape.

Adds a broker-free test in `MessagingMessageListenerAdapterTests` (following the existing `testAsyncResult` spy idiom, with `ObservationRegistry.create()` + an error-capturing handler as used elsewhere in the observation tests) that fails on the previous code — the exception surfaces from `whenComplete` while the observation records no error — and passes with the fix. `:spring-kafka:test` and Checkstyle pass.

This contribution was prepared with the assistance of Claude (Anthropic); the fix and the failing/passing test runs were verified by execution.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sobychacko:
@Gaurav1112 Please use DCO for your commits. https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring. Thanks.

PR #4688, opened 2026-09-16
Ignore stale asynchronous acknowledgments after rebalance

### LINKED ISSUE #4687: Suspend @KafkaListener: consumer permanently stops fetching partitions it retains across a rebalance
### Summary

When a rebalance lands while a `suspend fun` `@KafkaListener` still has records in flight, the consumer **permanently stops fetching the partitions it keeps**. Those partitions are never committed and never delivered again, for the lifetime of the consumer.

Nothing surfaces: no exception reaches a `CommonErrorHandler`, nothing is logged above `DEBUG`, heartbeats stay healthy, and no further rebalance is triggered. From the outside the consumer looks perfectly alive — it simply never touches those partitions again.

We hit this in production. One member held partitions 6-10 of a topic; after a rebalance it went 1.5 hours processing nothing while `LOG-END-OFFSET` kept growing on two of them and the other group members stayed healthy. The only thing that ever cleared it was another rebalance, which then wedged a different member.

### Versions

| | |
|---|---|
| spring-kafka | 4.0.7 (also observed in production on 4.0.5) |
| kafka-clients | 4.1.2 |
| Kotlin | 2.3.0 |
| kotlinx-coroutines | 1.10.2 |
| JDK | 21 |

### Reproducer

https://github.com/paulius-p-zedge/spring-kafka-async-ack-rebalance

```bash
./gradlew test
```

Two tests, identical except for the partition assignment strategy:

| Test | Assignor | Result |
|---|---|---|
| `EagerAssignorTest` | `RangeAssignor` (eager, the default) | **fails** |
| `CooperativeAssignorTest` | `CooperativeStickyAssignor` | passes |

The scenario:

1. Produce 4 records to each of 2 partitions. `max.poll.records` is 8, so one poll takes them all.
2. The listener blocks on a gate, so the entire poll is in flight with no ack completed.
3. Start a second listener in the same group, forcing a rebalance.
4. Release the gate. Records complete in **descending offset order**, so acks are parked in `deferredOffsets` behind the still-running head offset.
5. Once things settle, produce 3 more records to each partition.
6. Assert every partition's committed offset reached the log end.

Step 4 is the part that matters: the bug needs at least one ack that is not the head offset at the moment of revocation.

### Expected vs actual

Expected: both partitions committed at the log end.

Actual:

```
expected: {async.ack.rebalance-0=7L, async.ack.rebalance-1=7L}
 but was: {async.ack.rebalance-1=7L}

started=[0-0, 0-1, 0-2, 0-3, 1-0, 1-1, 1-2, 1-3,      first poll, 8 records held in flight
         0-0, 0-1, 0-2, 1-0, 0-3, 1-1, 1-2, 1-3,      all 8 redelivered after the rebalance
         1-4, 1-5, 1-6]                               second round: partition 1 only
```

Partition 0 is the one the first listener **kept** across the rebalance. It ends with no committed offset at all, and the three records produced to it afterwards are never delivered. Partition 1 moved to the second listener, which started with clean state and tracked the log end throughout.

So it is specifically the *retained* partitions that break; transferred ones are fine because the new owner builds its bookkeeping from scratch.

### Analysis

A `suspend fun` listener is `asyncReplies`, which forces `AckMode.MANUAL` in `ListenerConsumer#determineAckMode` and enables out-of-order commit tracking through `offsetsInThisBatch` / `deferredOffsets`.

In `ackInOrder`, an ack that is not the head offset is parked in `deferredOffsets`. `onPartitionsRevoked` then drops both map entries for every revoked partition:

```java
synchronized (ListenerConsumer.this) {
    Map<TopicPartition, List<Long>> pendingOffsets = ListenerConsumer.this.offsetsInThisBatch;
    if (pendingOffsets != null) {
        partitions.forEach(tp -> {
            pendingOffsets.remove(tp);
            Objects.requireNonNull(ListenerConsumer.this.deferredOffsets).remove(tp);
        });
        if (pendingOffsets.isEmpty()) {
            ListenerConsumer.this.consumerPaused = false;
        }
    }
}
```

An eager assignor revokes **all** partitions on every rebalance, including ones the same consumer is immediately handed back. The still-in-flight acks for those partitions then reach `ackInOrder` with no bookkeeping left and hit:

```java
throw new IllegalStateException("Unexpected ack for " + KafkaUtils.format(cRecord)
        + "; offsets list is empty");
```

That throw happens inside the `completableFutureResult.whenComplete(...)` callback in `MessagingMessageListenerAdapter#handleResult`. Nothing consumes the future that `whenComplete` returns, so the exception is swallowed — which is why no error handler ever sees it and nothing appears in the logs.

I have not pinned down precisely which piece of state is left inconsistent afterwards; what the reproducer establishes is the observable behaviour, which is that the consumer never fetches those partitions again. Given `doPauseConsumerIfNecessary` pauses while `offsetsInThisBatch` is non-empty and `doResumeConsumerIfNecessary` only clears `pausedForAsyncAcks` once it is empty, a stale or partially-cleared entry there looks like the likely culprit.

This looks like the same family as the previously-fixed #2410, #4258, #4265, #4269 and #4332, with revocation as a trigger that none of those cover.

### Why this is hard to detect

Worth calling out, because it cost us a long time to find:

- The `IllegalStateException` is swallowed, so no error handler and no log.
- The container logs pause/resume at `DEBUG` only.
- The consumer stays in the group with healthy heartbeats, so no rebalance or eviction follows.
- CPU and memory look normal.
- `spring.kafka.listener` timings look fine, because for a suspend listener `MicrometerHolder`'s sample stops when the continuation returns rather than when the coroutine completes.

The only external symptom is consumer lag with a floor that never drains, on a subset of one member's partitions.

### Workaround (partial — see the caveat)

```kotlin
consumerProps[ConsumerConfig.PARTITION_ASSIGNMENT_STRATEGY_CONFIG] =
    CooperativeStickyAssignor::class.java.name
```

Cooperative rebalancing only revokes partitions that actually change owner, which is why `CooperativeAssignorTest` passes — that test exercises a single clean join, where the retained partition is never revoked at all.

**It is not a complete workaround, and I want to correct that up front.** Cooperative rebalancing still calls `onPartitionsRevoked` for partitions that genuinely move, so a rolling deploy or an autoscaler event can still revoke a partition that has in-flight acks.

We have since observed exactly that in production with `cooperative-sticky` confirmed active on the running consumers.

So the assignor reduces how often the trigger is hit rather than removing it. Keeping the ack bookkeeping consistent across revocation looks necessary regardless of assignor.

dlwhdgus0810:
An asynchronous record listener can still complete records after a rebalance revokes and reassigns their partition. If the same offsets have already been redelivered, those old acknowledgments enter the new poll's deferred-offset lists. Mixing the two deliveries can consume offsets twice and leave the consumer paused indefinitely.

Retain the identity of the pending-offset list when creating each record acknowledgment. Under the same monitor used by offset capture and revocation, ignore acknowledgments whose list has been removed or replaced. This keeps an old completion from acknowledging a new delivery of the same offset.

The regression interleaves old and redelivered acknowledgments in descending offset order. It fails on the original code with `IndexOutOfBoundsException` and verifies that only current acknowledgments drain the new batch after the fix.

Validation:
- Full `:spring-kafka:test`: 1,203 tests, 1,197 passed and 6 existing skips; no failures/errors.
- `:spring-kafka:checkstyleMain` and `:spring-kafka:checkstyleTest` passed.
- The issue reporter's external Kotlin reproducer also passes with both eager and cooperative assignors against these sources. It was run through a temporary Gradle init script and is not included in this patch.

Fixes #4687.


sobychacko:
Merged via https://github.com/spring-projects/spring-kafka/commit/32334d6fa6d73134c80d873401bc93e00dfd3f7b.

